PR #47, opened 2016-03-14
Relational int domains in base

isabelmax:
### Summary:
Added the possibility of using relational int domains in the base analysis. When adding now for example `--set ana.int.relational true --set ana.int.equations true`, a new relational domain is used, which handles relations of the form `c1 * a + c2 + b + c3 = 0`.

### Explanation:
I wrote an interface for relational int domains, similar to the IntDomTuple in the IntDomain and created two relational int domains for it:
1. SimpleEquations is a relational domain, which handles relations of the form `c1 * a + c2 + b + c3 = 0` as `(Cil.varinfo * const) * (Cil.varinfo * const) * const`. An equation is always ordered alphabetically by the variable names and c1 is always 1.
2. The ApronDomain was adjusted, such that it can be used now from the poly and the base analysis. In order to handle global, as well as local variables, the local variable names get a space `" "` added. This can hardly be noticed by the user in the printed abstract value and spaces will never be part of a variable name in C.

### Comment:
Unfortunately, I had to use IntDomain.IntDomTuple everywhere, where ValueDomain.ID was used before, because else there was a circular build error (which I could not solve otherwise).

### How to test:
`goblint --html -o ../results/results_relational_int_own/  tests/regression/01-cpa/27-interval.c  --set ana.int.relational true --set ana.int.equations true --set ana.int.aprondomain true`
[...]

PR #48, opened 2016-04-08
Relational int domain in base

isabelmax:
As there is a build problem on the travis build server, we decided to start merging it to the apron branch.
The rest below is copied from the old pull request (see here: https://github.com/goblint/analyzer/pull/47).

## Summary:
Added the possibility of using relational int domains in the base analysis. When adding now for example `--set ana.int.relational true --set ana.int.equations true`, a new relational domain is used, which handles relations of the form `c1 * a + c2 + b + c3 = 0`.

## Explanation:
I wrote an interface for relational int domains, similar to the IntDomTuple in the IntDomain and created two relational int domains for it:
1. SimpleEquations is a relational domain, which handles relations of the form `c1 * a + c2 + b + c3 = 0 `as `(Cil.varinfo * const) * (Cil.varinfo * const) * const`. An equation is always ordered alphabetically by the variable names and c1 is always 1.
2. The ApronDomain was adjusted, such that it can be used now from the poly and the base analysis. In order to handle global, as well as local variables, the local variable names get a space `" "` added. This can hardly be noticed by the user in the printed abstract value and spaces will never be part of a variable name in C.

## Comment:
Unfortunately, I had to use `IntDomain.IntDomTuple` everywhere, where `ValueDomain.ID` was used before, because else there was a circular build error (which I could not solve otherwise).

## How to test:
`goblint --html -o ../results/results_relational_int_own/ tests/regression/01-cpa/27-interval.c --set ana.int.relational true --set ana.int.equations true --set ana.int.aprondomain true`
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/goblint/analyzer/pull/48?approve=1'></a>&nbsp;<a href='https://github.com/goblint/analyzer/pull/48'></a>
<a href='#crh-end'></a>

---
This change is [Reviewable](https://reviewable.io/reviews/goblint/analyzer/48)


vogler:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/goblint/analyzer/pull/48'>Refresh</a>.<a href='#crh-update-none'></a>

PR #60, opened 2016-10-19
handle casts correctly

### LINKED ISSUE #56: handle casts correctly

vogler:
This should fix #56. Ignore test script changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ git:
   submodules: false
 install: bash -ex scripts/travis-ci.sh
 before_script: ./make.sh headers
-script: ruby scripts/update_suite.rb -s -v -d
+script: ruby scripts/update_suite.rb -s -d
 notifications:
   webhooks:
     urls:
--- a/pgoblint.sh
+++ b/pgoblint.sh
@@ -9,6 +9,7 @@ if [ ! -e $file ]; then
   exit 1
 fi
 params="`grep -oP "PARAM: \K.*" $file`"
-cmd="./goblint --enable colors --enable dbg.debug --enable dbg.verbose --enable dbg.showtemps $params ${@:3} $file"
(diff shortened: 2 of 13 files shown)


vogler:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/goblint/analyzer/pull/60'>Refresh</a>.<a href='#crh-update-none'></a>

vogler:
Somehow it makes some race tests fail:

~~~
Expected norace, but registered race on arraycollapse_nr:54
09/13 regions/arraycollapse_nr failed!
Expected norace, but registered race on arrayloop_nr:54
09/17 regions/arrayloop_nr failed!
~~~

vesalvojdani:
I have been able to narrow this down slightly. Something happens during malloc calls, possibly the implicit cast generated by CIL, that trigger accesses to all members of a struct.

```
#include<pthread.h>
#include<stdlib.h>

struct s {
  int datum;
  struct s *next;
};

struct s *glob;

void *t_fun(void *arg) {
  struct s *node = malloc(sizeof(struct s));
  node->next = glob->next;
  return NULL;
}

int main () {
  pthread_t t1;
  pthread_create(&t1, NULL, t_fun, NULL);
  return 0;
}
```

Running this with default config (no args) will produce these accesses:
```
Memory location (struct s).datum (safe)
  read@13-arraycollapse_nr.c:14 {} -> {} (conf. 60)
  read@13-arraycollapse_nr.c:15 {} -> {} (conf. 60)
```
There are no accesses to the datum field in the program.

vogler:
I guess the read access comes from here because the type wasn't known before: https://github.com/goblint/analyzer/blob/56-casts/src/analyses/mCP.ml#L529
`Uk` stands for unknown?
What was the idea behind `reach`? It doesn't look like it's ever set to true.

PR #70, opened 2019-08-14
Incremental

vogler:
Incremental analysis with td3.


src/cdomains/arincDomain.ml
```diff
@@ -60,6 +60,8 @@ struct
   (* Printable.S *)
   (* let equal = Util.equals *)
   let equal x y = Pid.equal x.pid y.pid && Pri.equal x.pri y.pri && Per.equal x.per y.per && Cap.equal x.cap y.cap && Pmo.equal x.pmo y.pmo && PrE.equal x.pre y.pre && Pred.equal x.pred y.pred && Ctx.equal x.ctx y.ctx
+  (* Compare all fields with correspoding compare operators. TODO: make a "lazy" comparision *)
+  let compare x y = List.fold_left (fun acc v -> if acc = 0 && v <> 0 then v else acc) 0 [Pid.compare x.pid y.pid; Pri.compare x.pri y.pri; Per.compare x.per y.per; Cap.compare x.cap y.cap; Pmo.compare x.pmo y.pmo; PrE.compare x.pre y.pre; Pred.compare x.pred y.pred; Ctx.compare x.ctx y.ctx]
```
vogler: Why is this needed?

src/domains/setDomain.ml
```diff
@@ -515,8 +523,15 @@ struct
   let name () = "Set (" ^ E.name () ^ ")"
   (* let equal x y = try Map.equal (List.for_all2 E.equal) x y with Invalid_argument _ -> false *)
   let equal x y = leq x y && leq y x
-  let hash xs = fold (fun v a -> a + E.hash v) xs 0
-  let compare = compare
+  let hash = Hashtbl.hash
```
vogler: Why change to polymorphic hash?

vogler:
Todos:
- [ ] abort if compilation fails
  - currently the previous cilly result is then analyzed
    - [ ] why does it then take so long to analyze? Should be no change.
  - if compilation fails, we must not save any result, so that the changes over several commits will be accumulated until we reach a version that compiles and can be analyzed
  - [ ] make this the default (warn if compilation fails and ask if user wants to continue?)
  - [ ] make option `bench` for being strict and fail instead of asking
- [ ] what should be done if there are no (semantic) changes?
  - [x] always run solver and save results because of possible line number changes in the output
  - [ ] levels of sameness which we should report: 1. same CIL-output, 2. no semantic changes detected for function, 3. destabilized vars but ended up with same rho
- [ ] get `ana.hashcons` to work
  - Problem: weak references can't be marshalled
  - [HashconsLifter](https://github.com/goblint/analyzer/blob/master/src/framework/constraints.ml#L13) uses [BatHashcons](http://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatHashcons.html)
  - enough to `HM.iter` over `rho` in td3 and insert into polymorphic Hashtbl before marshalling if `ana.hashcons` is enabled? `iter: (key BatHashcons.hobj -> unit) ->
       t -> unit` but we want to insert `key`'s `obj` without `tag` and `hcode`.
  - if not, we need a module `HM` 
- [ ] do (un)marshalling outside of solver somewhere here in [control.ml](https://github.com/goblint/analyzer/blob/master/src/framework/control.ml#L321) since it's not td3-specific
  - every `Solver.solve` should get some `state` and return `rho, state` where state should be marshallable.
- [ ] make commit direction configurable
  - currently we analyze going forward in time, i.e. it looks for analysis results in older commits
  - could also analyze current commit and go backwards, i.e. look for analysis results in newer commits
  - alternative: take closest commit (beneficial for `git bisect` (binary search on commits))

Relevant for benchmarks (general issues, not incremental):
- [ ] check if `exp.widen-context` works reliably and make it the default if `exp.full-context` is off (too expensive otherwise (how much?))
- [ ] output benchmark-relevant data to some json-file (or database) so that we don't need to parse unstructured text from logs, and to make it easier to compare options
  - [ ] should be written whenever goblint terminates (i.e. also for exceptions, just with partial info then)
  - [ ] `printstats`(results from `Stats.time`)
  - [ ] `Generic.SolverStats` (vars = ... evals = ...)
  - [ ] for long running analyses, some intermediate info while solving, e.g. same as for `ctrl+\`, just every x minutes
    - [ ] vars, evals, var with most updates
    - [ ] some info on number of contexts per function. problem: solver just has `S.Var` and doesn't know about nodes/contexts.

jerhard:
As of now now, Goblint exits if the `make` used to combine the source files produces an exit code different from 0. In case of compiler warnings, the exit code of the make command is still 0 and Goblint therefore continues. **Cilly discards the `Werror` flag**, so it even produces a combined file in the presence of warnings and the `Werror` flag in the Makefile.

vogler:
It's not clear to me why `ana.hashcons` should fail. @jerhard What's the exception?
Marshalling from one utop to another worked fine:
~~~ocaml
utop # module HC = BatHashcons.MakeTable (struct include BatInt let hash = Hashtbl.hash end)
module HC :
  sig
    type key = int
    type t
    val create : key -> t
    val clear : t -> unit
    val hashcons : t -> key -> key BatHashcons.t
    val iter : (key BatHashcons.t -> unit) -> t -> unit
    val fold : (key BatHashcons.t -> 'a -> 'a) -> t -> 'a -> 'a
    val count : t -> key
  end

utop # let h = HC.create 13
val h : HC.t = <abstr>

utop # HC.count h
- : int = 0

utop # let x = HC.hashcons h 42
val x : int BatHashcons.t = {BatHashcons.obj = 42; tag = 1; hcode = 395478795}

utop # let y = HC.hashcons h 43
val y : int BatHashcons.t = {BatHashcons.obj = 43; tag = 2; hcode = 986401281}

utop # let z = HC.hashcons h 42
val z : int BatHashcons.t = {BatHashcons.obj = 42; tag = 1; hcode = 395478795}

utop # HC.count h
- : int = 2

utop # Marshal.to_bytes x []
- : bytes =
Bytes.of_string "\132\149\166\190\000\000\000\b\000\000\000\001\000\000\000\004\000\000\000\004\176jA\002\023\146\135\011"
~~~

New utop:
~~~ocaml
utop # module HC = BatHashcons.MakeTable (struct include BatInt let hash = Hashtbl.hash end)
utop # let h = HC.create 13
utop # let bx = Bytes.of_string "\132\149\166\190\000\000\000\b\000\000\000\001\000\000\000\004\000\000\000\004\176jA\002\023\146\135\011"
utop # let dump : int BatHashcons.t = Marshal.from_bytes bx 0
val dump : int BatHashcons.t =
  {BatHashcons.obj = 42; tag = 1; hcode = 395478795}
~~~

If `Printable.HConsed(B).htable` is not restored, it's just empty on the next run.
If hashcons was on, same values are shared (transparent to goblint), and sharing should be preserved by `Marshal`.
We load `rho` (with shared values) which contains things like `int BatHashcons.t = {BatHashcons.obj = 42; tag = 5; hcode = 848809375}` (except we have D.t and C.t instead of int).
These initial values are referenced by `rho` only, and not by `htable` since it's empty in the new run, so we might miss sharing for all values in the initial `rho` in the worst case.
However, if we compute new values during solving, they will populate `htable` and same values should be shared.
Since the tags are not pure but just incremented as a side-effect, we might end up with a value `{BatHashcons.obj = 42; tag = 5; hcode = ..}` from the old `rho` and `{BatHashcons.obj = 42; tag = 3; hcode = ..}` from the new run. Now `leq` and `hash` in `Lattice.S` will be fine, since they are based on the value, but `equal` and `compare` will say the values are different.
Apart from the missed sharing, this would lead to the solver not recognizing that the value for a variable x hasn't changed and it would unnecessarily destabilize.
To fix this we would need to make sure that the tags are the same or let equal compare the value instead of the tag, which would however destroy the performance benefit of not having to do structural comparison for the hashconsed values.

Aside: I noticed that `hashcons` is only called after each `bot, top, join, meet, widen, narrow`.
If a rhs does not call any of those, but results in the same value as `old` but with a different tag (since no call to `hashcons`, it will also not be recognized as equal and the solver will unnecessarily destabilize. I'm not sure whether this could be a reason for non-termination (if x destabilizes itself there should be a widening point which leads to the correct tag, so probably not), but it might negate the benefits of hashconsing.

vogler:
Copied from 2311bdf:

This hack is for fixing hashconsing.
If hashcons is enabled now, then it also was for the loaded values (otherwise it would crash). If it is off, we don't need to do anything.
HashconsLifter uses BatHashcons.hashcons on Lattice operations like join, so we call join (with bot) to make sure that the old values will populate the empty hashcons table via side-effects and at the same time get new tags that are conform with its state.
The tags are used for `equals` and `compare` to avoid structural comparisons. TODO is this really faster (if values are shared by hashcons they should be physically equal)?
We have to replace all tags since they are not derived from the value (like hash) but are incremented starting with 1, i.e. dependent on the order in which lattice operations for different values are called, which will very likely be different for an incremental run.
If we didn't do this, during solve, a rhs might give the same value as from the old rho but it wouldn't be detected as equal since the tags would be different.
In the worst case, every rhs would yield the same value, but we would destabilize for every var in rho until we replaced all values (just with new tags).
The other problem is that we would likely use more memory since values from old rho would not be shared with the same values in the hashcons table. So we would keep old values in memory until they are replace in rho and eventually garbage collected.

src/framework/control.ml
```diff
@@ -29,10 +29,13 @@ let get_spec () : (module Spec) =
...
-  (** The main function to preform the selected analyses. *)
-  let analyze (file: file) (startfuns, exitfuns, otherfuns: Analyses.fundecs)  (module Spec : Spec) =
+    (** The main function to preform the selected analyses. *)
+    let analyze (file: file) (startfuns, exitfuns, otherfuns: Analyses.fundecs)  (module Spec : Spec) (increment: increment_data) =
+  
+    let module Inc = struct let increment = increment end in
+
     (** The Equation system *)
-    let module EQSys = FromSpec (Spec) (Cfg) in
+    let module EQSys = FromSpec (Spec) (Cfg) (Inc) in
```
vogler: Why do we need this module?

src/util/defaults.ml
```diff
@@ -169,13 +169,14 @@ let _ = ()
       ; reg Experimental "exp.no-narrow"         "false" "Overwrite narrow a b = a"
       ; reg Experimental "exp.basic-blocks"      "false" "Only keep values for basic blocks instead of for every node. Should take longer but need less space."
       ; reg Experimental "exp.widen-context"     "false" "Do widening on contexts. Method depends on exp.full-context - costly if true."
-      ; reg Experimental "exp.solver.td3.term"  "true" "Should the td3 solver use the phased/terminating strategy?"
-      ; reg Experimental "exp.solver.td3.space" "false" "Should the td3 solver only keep values at widening points?"
-      ; reg Experimental "exp.solver.td3.space_cache" "true" "Should the td3-space solver cache values?"
-      ; reg Experimental "exp.solver.td3.space_restore" "true" "Should the td3-space solver restore values for non-widening-points? Needed for inspecting output!"
+      ; reg Experimental "exp.solver.wp.restore" "true" "Should the wp solver restore values for non-widening-points?"
```
vogler: The solver uses the new options on the left. This was merged the wrong way.

src/framework/control.ml
vogler: Why do we need this module?

src/framework/constraints.ml
```diff
@@ -374,7 +379,8 @@ struct
   module G = S.G
 
   let full_context = get_bool "exp.full-context"
-
+  (* Dummy module. No incremental analysis supported here*)
```
vogler: Why do we need this?

src/domains/setDomain.ml
vogler: Should use non-polymorphic hash.

src/domains/printable.ml
```diff
@@ -479,6 +528,8 @@ struct
 
   let lift x = `Lifted x
 
+  let compare = Pervasives.compare
```
vogler: Not needed.

src/domains/printable.ml
```diff
@@ -40,6 +39,7 @@ end
 module Blank =
 struct
   include Std
+  let compare = compare
```
vogler: Not needed.

src/domains/mapDomain.ml
```diff
@@ -73,6 +73,7 @@ struct
   (* And one less brainy definition *)
   let for_all2 = M.equal
   let equal = for_all2 Range.equal
+  let compare x y = if equal x y then 0 else M.compare Range.compare x y
```
vogler: Why `if equal x y then 0 else`? If they're not equal, we do the work twice (`equal`, then `compare`).

src/cdomains/intDomain.ml
```diff
@@ -359,6 +359,7 @@ struct
   let pretty_diff () (x,y) = Pretty.dprintf "%a instead of %a" pretty x pretty y
   let join x y = if Int64.compare x y > 0 then x else y
   let meet x y = if Int64.compare x y > 0 then y else x
+  let compare = Int64.compare
```
vogler: Not needed.

src/cdomains/intDomain.ml
```diff
@@ -583,7 +584,7 @@ struct
     | `Definite of Integers.t
     | `Bot
   ] [@@deriving to_yojson]
-
+  let compare = compare
```
vogler: Not needed.

src/cdomains/intDomain.ml
```diff
@@ -838,7 +839,7 @@ struct
   module I = CBigInt
   module C = CircularBigInt
   type t = I.t interval
-
+  let compare = compare
```
vogler: Not needed.

src/cdomains/intDomain.ml
```diff
@@ -1142,6 +1143,7 @@ struct
   type t = bool [@@deriving to_yojson]
   let hash = function true -> 51534333 | _ -> 561123444
   let equal_to i x = if x then `Top else failwith "unsupported: equal_to with bottom"
+  let compare = compare
```
vogler: Not needed.

src/cdomains/basetype.ml
```diff
@@ -168,6 +168,7 @@ struct
   type t = string [@@deriving to_yojson]
   let hash (x:t) = Hashtbl.hash x
   let equal (x:t) (y:t) = x=y
+  let compare x y = String.compare x y
```
vogler: Not needed.

src/framework/control.ml
vogler: So far only needed to access `S.increment` in `td3.ml`.
We should move most of the incremental code out from `td3.ml` into `control.ml`.
`solve` should just get the loaded state and some change information to know what to destabilize.

vogler:
Copied from d7bd36b:
The other problem are the tags for the context part of a S.Var.t.
This will cause problems when old and new vars interact or when new S.Dom values are used as context:
- destabilize before solve is fine since it runs only on keys from the old rho, so we should start solve with a consistant state.
- reachability is a problem since it marks vars reachable with a new tag, which will remove vars with the same context but old tag from rho.
- If we destabilized a node with a call, we will also destabilize all vars of the called function. However, if we end up with the same state at the caller node, without hashcons we would only need to go over all vars in the function once to restabilize them since we have the old values, whereas with hashcons, we would get a context with a different tag, could not find the old value for that var, and have to recompute all vars in the function (without access to old values).

vogler:
hashcons done, equal_debug did not complain, figlet: 5m19s (1s equal) 2165MB on vs. 7m22s (41s equal) 2652MB off -> with hashcons 72% runtime, 82% major heap

~~~
TOTAL                          310.762 s
  parse                           0.021 s
  convert to CIL                  0.018 s
  analysis                       310.723 s
    solving                        310.695 s
      S.Dom.equal                     1.047 s
Timing used
Memory statistics: total=1046383.53MB, max=177.08MB, minor=1046377.96MB, major=2165.49MB, promoted=2159.92MB
    minor collections=498997  major collections=128 compactions=0
~~~

~~~
TOTAL                          428.034 s
  parse                           0.026 s
  convert to CIL                  0.017 s
  analysis                       427.990 s
    solving                        427.965 s
      S.Dom.equal                    41.736 s
Timing used
Memory statistics: total=1494497.43MB, max=197.48MB, minor=1494493.31MB, major=2652.64MB, promoted=2648.52MB
    minor collections=712682  major collections=153 compactions=0
~~~

vogler:
Since this PR now works with hashcons, let's merge it and continue on master or create other PRs for #70 (comment)

PR #69, opened 2019-05-14
Octagon Analysis Implementation

ozeidan:

PR #71, opened 2019-09-03
More Sophisticated Array Domain

michael-schwarz:
This PR contains three changes:

- Adds a more sophisticated array domain that can be enabled with `exp.partition-arrays.enabled` and tuned with some options in `exp.partition-arrays.*`
- Fixes some issues with the Octagon domain introduced in #69 
- Addresses some issues that were discovered for path sensitive analyses during the development of the array domain (that was originally path-sensitive)


michael-schwarz:
The fact that all other analyses are affected by this PR is due to two things:

- `name` for analyses is now a function `unit -> string` to make the module "safe" to allow for the way `smart_join` is realized in the BaseDomain
- `should_join` was ignored earlier, now all path-sensitive analyses need to specify it themselves

PR #72, opened 2019-10-01
SV-Comp

vogler:
Witness generation and script for https://sv-comp.sosy-lab.org/2020.
Benchmarks: https://github.com/sosy-lab/sv-benchmarks


vogler:
~~~
$ cat 
int g[2] = {1, 2};

int main() {
  int a[2] = {1, 2};
  return a[0];
}
$ ./goblint -v --html --enable ana.int.interval global_array_init.c
~~~
Gives `g -> Array(Top)` now instead of `g -> Array([1,2])`.

Broken by 15a4cfe1efb0ce3c2d44d999e922cd7d9e47ab79
~~~ocaml
  (* run eval_rv from above, but change bot to top to be sound for programs with undefined behavior. *)
  (* Previously we only gave sound results for programs without undefined behavior, so yielding bot for accessing an uninitialized array was considered ok. Now only [invariant] can yield bot/Deadcode if the condition is known to be false but evaluating an expression should not be bot. *)
  let eval_rv (a: Q.ask) (gs:glob_fun) (st: store) (exp:exp): value =
    let r = eval_rv a gs st exp in
    if VD.is_bot r then top_value a gs st (typeOf exp) else r
~~~

I thought that would be enough, but I guess not. We should however do something like that, otherwise there's just a lot of dead code because the tests access uninitialized malloced structs/arrays as a means to simulate input.

michael-schwarz:
> I thought that would be enough, but I guess not. We should however do something like that, otherwise there's just a lot of dead code because the tests access uninitialized malloced structs/arrays as a means to simulate input.

Since globals are zero-initialized and CIL outputs code to explicitly do that, a quick improvement here would be to that not do this for globals.

PR #74, opened 2019-10-17
Allow vla in cil result

michael-schwarz:
Handle the changes https://github.com/goblint/cil/pull/7 and allow VLAs.
The test that are red, are the ones that were red on master before as well.

Modifies the following things:
-  Code to handle new CIL instruction `VarDecl` by evaluating the expression inside for VLA [] and setting the array length in the array domain accordingly
-  Changes CArrays.make to take an `int option` instead of `int` for length of an array
-  Remove deprecated array domains as some of them rely on the length being known


src/analyses/base.ml
```diff
@@ -753,17 +753,11 @@ struct
...
-            match ID.to_int n with
-            | Some n -> `Array (ValueDomain.CArrays.make (Int64.to_int n) (bot_value a gs st ai))
-            | _ -> default
-          end
-        | _ -> default
-      end
+    | TArray (ai, None, _) ->
+      `Array (ValueDomain.CArrays.make None (bot_value a gs st ai))
+    | TArray (ai, Some exp, _) ->
+      let l = Cil.isInteger (Cil.constFold true exp) in
```
vogler: Why replace `eval_rv` when `Cil.constFold` does strictly less?

src/analyses/base.ml
```diff
@@ -779,18 +773,11 @@ struct
...
-              match ID.to_int n with
-              | Some n -> `Array (ValueDomain.CArrays.make (Int64.to_int n) (if get_bool "exp.partition-arrays.enabled" then (init_value a gs st ai) else (bot_value a gs st ai)))
-              | _ -> default
-            end
-          | _ -> default)
-       | None -> default)
+    | TArray (ai, None, _) ->
+      `Array (ValueDomain.CArrays.make None (if get_bool "exp.partition-arrays.enabled" then (init_value a gs st ai) else (bot_value a gs st ai)))
+    | TArray (ai, Some exp, _) ->
+      let l = Cil.isInteger (Cil.constFold true exp) in
```
vogler: Why replace `eval_rv` when `Cil.constFold` does strictly less?

src/analyses/base.ml
```diff
@@ -805,18 +792,11 @@ struct
...
-              match ID.to_int n with
-              | Some n -> `Array (ValueDomain.CArrays.make (Int64.to_int n) (if get_bool "exp.partition-arrays.enabled" then (top_value a gs st ai) else (bot_value a gs st ai)))
-              | _ -> default
-            end
-          | _ -> default)
-       | None -> default)
+    | TArray (ai, None, _) ->
+      `Array (ValueDomain.CArrays.make None (if get_bool "exp.partition-arrays.enabled" then (top_value a gs st ai) else (bot_value a gs st ai)))
+    | TArray (ai, Some exp, _) ->
+      let l = Cil.isInteger (Cil.constFold true exp) in
```
vogler: Why replace `eval_rv` when `Cil.constFold` does strictly less?

src/cdomains/valueDomain.ml
```diff
@@ -223,12 +224,20 @@ struct
     | a, b -> a = b
 
   let cast_addr t a =
+    let rec patchTypeNotVLA t =
```
vogler: Why do we need this here? Better name? `stripVarLenArr`, `onlyConstLengthArrays`...

src/cdomains/valueDomain.ml
```diff
@@ -612,6 +624,15 @@ struct
         end
       | _ -> BinOp(MinusPP, exp, StartOf start_of_array_lval, intType)
     in
+    (* Create a typesig from a type, but drop the arraylen attribute *)
+    let typeSigWithoutArraylen t =
```
vogler: `typeSigWithoutArrayLen`?

src/util/cilfacade.ml
```diff
@@ -69,7 +69,10 @@ let do_preprocess ast =
   iterGlobals ast (function GFun (fd,_) -> List.iter (f fd) !visitors | _ -> ())
 
 let createCFG (fileAST: file) =
-  if not (get_bool "exp.basic-blocks") then end_basic_blocks fileAST;
+  (* After adding support for VLAs, there are new VarDecl instructions at the point where a VLA was declared and    *)
+  (* its declaration is not printed at the beginning of the function. For these VarDecl putting them into their own *)
+  (* BB causes the output CIL file to no longer compile. Therefore this is not done for justcil. *)
+  if not (get_bool "exp.basic-blocks") && not (get_bool "justcil") then end_basic_blocks fileAST;
```
vogler: Better fix `end_basic_blocks` instead.

src/framework/constraints.ml
```diff
@@ -612,6 +626,7 @@ struct
   let tf var getl sidel getg sideg edge d =
     begin match edge with
       | Assign (lv,rv) -> tf_assign var edge lv rv
+      | VDecl (v)      -> tf_vdecl var edge v
```
vogler: Introducing a new kind of edge just for VLAs is a bit heavy. Either solve it differently or fully embrace it and do it for all declarations. Should think about whether this has any (dis)advantages.

src/cdomains/valueDomain.ml
michael-schwarz: Since the attribute we are removing is actually `Attr("arraylen",_)`, I'd stick with non camelCase version here personally, but if you feel strongly about having it camelCase, I'm fine with that too

src/analyses/base.ml
michael-schwarz: Because eval_rv evaluates expression such as `x+17` for the current binding of `x`.
But this only yields the correct length if it is done in the correct place (i.e. the location of the declaration `int a[x+17];`), but we don't know where this bot_value is called.

src/analyses/base.ml
michael-schwarz: See above

src/cdomains/valueDomain.ml
michael-schwarz: Good point, I'll change the name!

It is needed because `bitsSizeOf`, which is called a few lines down, can not handle types that contain arrays of non-const size as it is not possible to determine their size in bits from just to the type.

vogler:
> * Remove deprecated array domains as some of them rely on the length being known

Why? This should be undone! They're only deprecated in the sense that we currently don't use them, but it's nice to have the code for experiments.

src/analyses/base.ml
vogler: So this means: for VLAs the length will be `None` here (and needs to be updated later) and for normal arrays it should be the static length.

src/analyses/base.ml
vogler: For a VLA with some known length, it might be that we overwrite the length here. Probably length should also be a lattice (interval), then there's no problem for VLAs/malloced arrays.

michael-schwarz: Changed it so the length is from `Idx` (domain also used for indices).

src/analyses/base.ml
michael-schwarz: Replaced it so the length is from `Idx`.

In `bot_value` and `init_value` we start with length `Idx.bot ()` as we are guaranteed to set the length later at some point (where the VLA is declared).

In `top_value` we use `Idx.top ()`

vogler:
Summary (cf. https://github.com/goblint/cil/pull/7#discussion_r336014870):
- `Cil.instr += VarDecl`
- Goblint:
  - `MyCFG.edge += VDecl`
  - `Analyses.Spec.vdecl` only overwritten in `Base`

Currently only used for VLAs, but could be used for any declaration, which has the advantage that the state would only contain variables that have already been declared so far. OTOH one should then introduce a dual transfer function for when a variable goes out of scope. Cil's renaming would probably have to be kept in case of shadowing.

PR #75, opened 2019-10-31
initialize locals with bot, warn on access (undefined behavior)

vogler:
Accessing uninitialized objects is undefined behavior and we should warn about it!
The access might happen in [get] or in [do_offs] in some sub-expression in the above [eval_rv].
Since bot propagates and we don't want to warn for all sub-expressions we only warn here.
Previously we only gave sound results for programs without undefined behavior, so yielding bot for accessing an uninitialized array was considered ok.
Now we want to warn about about any access to uninitialized memory and treat it as top/Unknown instead of bot/Deadcode.
We could just change bot to top here, but then we would not see which values are transitively uninitialzed (int x; int y = x; // y should also be bot).
So we only warn here, keep bot, and instead treat bot as top when branching. Locals are now initialized with bot instead of top.
This way bot for variables means uninitalized (before we only had bot inside arrays and blobs).
To be sound we adjust:
- [branch] such that it falls through and meets [invariant] in both branches instead of raising Deadcode if eval_rv cond is bot
- asserts to return UNKNOWN
-  calls of expressions that evaluate to bot TODO can this happen?


vogler:
As expected this breaks some tests (`undefined_behavior/access-uninitialized` is new):
image

vogler:
Notes:
- `06/{19,22} symbeq/*`: some problem in `varEq.ml`
- `25/02 vla/loop`: assert seems fine, i.e. remove `UNKNOWN`
- `22/01 partitioned_arrays/simple_array`: this should fail because the values are not in `char`? So probably a problem with cast on bot.
- `26/01 undefined_behavior/access-uninitialized`: problem with `is_bot` in ArrayDomain?

michael-schwarz:
For `25/02 vla/loop` the assert is not ok.

```
  for(int i = 2; i < 47; i++) {
    int n = 1;
    int a[1];

    if(i == 2) {
      a[0] = 42;
    }

    assert(a[0] == 42); //UNKNOWN
  }
```

This assert only holds in the iteration with `i=2`. It probably is some issue with the new `initial_state` for partitioned arrays, I will look into this.

michael-schwarz:
Also currently we warn about UB in too many cases:

```
  for(int i=0; i < 42; i++) {
    if(i ==2) {
      x = 7;
    }
  }
```

Leads to `[undefined behavior] Evaluating expression with uninitialized content: i == 2`.
At least for conditionals used to branch we should not warn if they have the value bottom.

michael-schwarz:
This change also has the following effect:

```
  int t;

  for(int i = 7; i++; i < 49) {
    if(srand()) {
      t = 8;
    }
  }

  assert(t == 8);
```

So we still make the no-UB assumption, just in a different way. I'm not sure if this is desirable.

vogler:
> Also, currently we warn about UB in too many cases:
> 
> ```
>   for(int i=0; i < 42; i++) {
>     if(i ==2) {
>       x = 7;
>     }
>   }
> ```
> 
> Leads to `[undefined behavior] Evaluating expression with uninitialized content: i == 2`.
> At least for conditionals used to branch we should not warn if they have the value bottom.

1. I don't get that warning.
2. The value for `i` shouldn't be bottom.
3. If it was we should still warn.

michael-schwarz:
1. Hmm, strange... Did you push your most recent changes? I am running it with
`./goblint --enable colors --enable dbg.debug --enable dbg.showtemps --enable dbg.verbose --enable dbg.earlywarn --sets result html --sets solver td3  --enable ana.int.interval --set ana.activated "['base','expRelation']"  --disable ana.int.trier --disable exp.fast_global_inits --enable exp.partition-arrays.enabled`
2. No, but  `i == 2` is bot in the first iteration.

vogler:
> This change also has the following effect:
> 
> ```
>   int t;
> 
>   for(int i = 7; i++; i < 49) {
>     if(rand()) {
>       t = 8;
>     }
>   }
> 
>   assert(t == 8);
> ```
> 
> This assert will succeed without any warnings.
> 
> So we still make the no-UB assumption, just in a different way. I'm not sure if this is desirable.

True, but at least we now warn for must-UB.
Simpler example:
~~~
int main() {
  int x, y;
  if (x) // WARN must-UB
    y = 1;
  return y; // NOWARN but should be WARN may-UB
}
~~~

michael-schwarz:
The partitioned arrays behave somewhat different to other locals at the moment because of [this](https://github.com/goblint/analyzer/blob/9ec72685db98fb87e2d8372a097acf1931c12d9e/src/cdomains/arrayDomain.ml#L102)

Without it, we lose most of the benefit of partitioning, as e.g. in
```
int x[42];
x[7] = 1;
x[8] = 2;
assert(x[6] == 1);
```
the assert would succeed.

src/analyses/base.ml
```diff
@@ -807,6 +809,9 @@ struct
     | TNamed ({ttype=t}, _) -> init_value a gs st t
     | _ -> `Top
 
+  (* We can't just set all to `Bot since some domains/analyses rely on the abstract value having the right type (would make 30 tests fail), so we create a bot of the right abstract type. *)
```
michael-schwarz: Why are we keeping the old `init_value` around? Does not seem to be used anywhere

vogler: This whole `{bot,top,init}_value` duplication is ugly. Once it's clear what to do, we want some option for whether we assume the program to be free from undefined behavior and set the initial value depending on that.
The question is how `init_value` is different from `top_value`:
~~~ocaml
    | t when is_mutex_type t -> `Top
    | TInt (ik,_) -> `Int (ID.(cast_to ik (top ())))
    | TPtr _ -> `Address (if get_bool "exp.uninit-ptr-safe" then AD.(join null_ptr safe_ptr) else AD.top_ptr)
~~~
That, and arrays contain `bot` instead of `top`.

vogler: I'm refactoring it.

michael-schwarz:
I cherry-picked some commits that we definitely want in the master branch (e.g. getting rid of wrongly returning `\bot` in `Interval32.eq`), even if we don't end up merging this.

michael-schwarz:
I merged master into this, so it doesn't become even staler than it already is/

michael-schwarz:
Closing as stale.

Issue #54, opened 2016-07-27
make sizes of primitive types configurable with current machine as default


vogler:
Cil's defaults are generated from `src/machdep-ml.c.in` and then stored in `_build/machdep.ml`.

- [ ] make `Machdep.theMachine` configurable in goblint
- [ ] provide default configurations {intel, arm} {32bit, 64bit} {linux, macOS}

vogler:
Some more context from other comments:

---

Current generated `machdep.ml` attached below.
Remove the msvc stuff, `gcc` -> `gcc_arm64_macos`, generate and add the same for `x86_64_linux` etc.
`theMachine` is already a reference, just need to set it from goblint to the current machine or option if provided (and then also pass it for preprocessing as mentioned by Michael above).

```ocaml
(* This module was generated automatically by code in Makefile and machdep-ml.c *)
type mach = {
  version_major: int;     (* Major version number *)
  version_minor: int;     (* Minor version number *)
  version: string;        (* gcc version string *)
  underscore_name: bool;  (* If assembly names have leading underscore *)
  sizeof_short: int;      (* Size of "short" *)
  sizeof_int: int;        (* Size of "int" *)
  sizeof_bool: int;       (* Size of "_Bool" *)
  sizeof_long: int ;      (* Size of "long" *)
  sizeof_longlong: int;   (* Size of "long long" *)
  sizeof_ptr: int;        (* Size of pointers *)
  sizeof_float: int;      (* Size of "float" *)
  sizeof_double: int;     (* Size of "double" *)
  sizeof_longdouble: int; (* Size of "long double" *)
  sizeof_floatcomplex: int;      (* Size of "float _Complex" *)
  sizeof_doublecomplex: int;     (* Size of "double _Complex" *)
  sizeof_longdoublecomplex: int; (* Size of "long double _Complex" *)
  sizeof_void: int;       (* Size of "void" *)
  sizeof_fun: int;        (* Size of function *)
  size_t: string;         (* Type of "sizeof(T)" *)
  wchar_t: string;        (* Type of "wchar_t" *)
  alignof_short: int;     (* Alignment of "short" *)
  alignof_int: int;       (* Alignment of "int" *)
  alignof_bool: int;      (* Alignment of "_Bool" *)
  alignof_long: int;      (* Alignment of "long" *)
  alignof_longlong: int;  (* Alignment of "long long" *)
  alignof_ptr: int;       (* Alignment of pointers *)
[...]

PR #80, opened 2020-03-02
Simmo's latest changes to witness generation

michael-schwarz:


src/util/defaults.ml
```diff
@@ -178,8 +178,10 @@ let _ = ()
       ; reg Experimental "exp.fast_global_inits" "false" "Only generate 'a[0] = 0' for a zero-initialized array a[n]. This is only sound for our flat array domain! TODO change this once we use others!"
       ; reg Experimental "exp.uninit-ptr-safe"   "false" "Assume that uninitialized stack-allocated pointers may only point to variables not in the program or null."
       ; reg Experimental "exp.ptr-arith-safe"    "false" "Assume that pointer arithmetic only yields safe addresses."
+      ; reg Experimental "exp.witness_path" "'witness.graphml'" "Witness output path"
       ; reg Experimental "exp.minwitness"        "false" "Try to minimize the witness"
       ; reg Experimental "exp.uncilwitness"      "false" "Try to undo CIL control flow transformations in witness"
+      ; reg Experimental "exp.architecture"      "'64bit'" "Architecture for analysis, currently for witness"
```
michael-schwarz: This should be tied in with the CIL flags concerning architecture, rather than being specifiable on its own. See also https://github.com/goblint/analyzer/issues/54

src/witness/myARG.ml
```diff
@@ -149,6 +165,12 @@ struct
   let next_opt _ = None
 end
 
+(* Exps can contain compinfo <-> fieldinfo loops in TComp typ which make using (=) impossible.
+ * Ideally should manually define equality for exps to define equality for typs.
+ * Should probably be used elsewhere besides partition_if_next as well.
+ * There is CompareAST.eq_exp but it has the same issue. *)
+let is_equal_exp exp1 exp2 = exp1 == exp2
```
michael-schwarz: How about `Expcompare.compareExp` from CIL?

src/witness/myARG.ml
```diff
@@ -259,16 +284,8 @@ struct
...
-sig
-  include Node
-
-  val move: t -> MyCFG.node -> t
-  val is_live: t -> bool
-end
-
-module Intra (Node: MoveNode) (ArgIntra: SIntraOpt) (Arg: S with module Node = Node):
-  S with module Node = Node =
+module Intra (ArgIntra: SIntraOpt) (Arg: S):
```
michael-schwarz: Is this not `Inter`?

sim642: Probably just because it connects the intraprocedural transformations with the ARG interface. But the latter is interproedural so it's confusing indeed.

If you have a better name suggestion, which makes the intra-inter connection clearer, it might be better indeed. Or maybe something like `ApplyIntra` or `MapIntra` as in applying/mapping the ARG with intraprocedural transformation.

sim642:
I should also say something about the `prev_node` business because it has some legacy.

My original implementation of `WitnessLifter` uses an auxiliary `VF` component in its domain to make the previous node of a transfer function accessible without modifying the signatures by having each transfer function put the current node in the domain, the previous value of which is accessible. The `prev_node` addition to the signatures was the beginning of an attempt to get rid of that stupidity in the domain by making the previous node directly accessible through `ctx`.

That transition, however, is not complete. `step_ctx` has a sanity-assert for checking that the two approaches indeed behave the same. This helped in finding some bugs in one or the other at some point. Ideally the `VF` component of the domain should be removed and `prev_node` should be directly used. I never got around to stripping it out. At some point I thought I realized it can't be stripped out because of some subtlety in `enter`/`combine`, like having to distinguish two previous nodes (the call site and the called function's return) but I never really tried to go through with it.

Issue #90, opened 2020-08-18
Fix branching on variables


jerhard:
As shown by the test case [here](https://github.com/goblint/analyzer/blob/3d54f9bab9214a20fc203d3532018d886288a8e6/tests/regression/01-cpa/35-intervals.c), branching on a variable (without an explicit comparison) currently restricts the abstract domain of the variable in question for the true branch to the abstract value [1;1] for intervals or the respective representation of 1 of the domain.
This is unsound.

vogler:
Probably a bug with the new `invariant` function or does it also happen with the old one?

vogler:
It's just strange that none of the existing tests caught it.
Should rethink the assert-based regression tests and replace it with something more expressive. Also, the test script should be able to check some tests with multiple configurations (e.g. all int domains separately).

vogler:
Seems to be related to https://github.com/goblint/analyzer/commit/f1c6e572906bea3f3bd4e6b34f3574be5844c3db.
There you changed `of_bool` for `Enums` but for `Interval32` it's still zero/one. For `DefExc` it already was zero/non_zero.
Initially all of them were zero/one - should check if some place relies on one instead of non_zero and fix it.

jerhard:
> Probably a bug with the new `invariant` function or does it also happen with the old one?

Yes, the issue does not occur when I always use the fallback

>Seems to be related to f1c6e57.

Well, in so far as the branching code caused some trouble with reaching the fixed point for enums. I initially thought I would only have to fix it there, but then it turned out that intervals were delivering spurious results, too.

> Initially all of them were zero/one - should check if some place relies on one instead of non_zero and fix it.

Yes. And there are multiple calls in Goblint that expect it to be zero/one, because they want the result of a comparsion.
But here, in the new `invariant` method, it should be zero/non zero:
https://github.com/goblint/analyzer/blob/baf07d1f10901f540ae21a69131dcb7f8d71b7ec/src/analyses/base.ml#L1495
So I guess we should add a new method `of_bool_inv_iking: ikind -> bool -> t` to the IntDomain and then use it in the `invariant`.

vogler:
Idk, seems messy, those `{is,of,to}_{bool,int}` need some rethinking.
The problem probably arose due to the discrepancy between `of_int` and `of_bool`.
`of_int` always gives a definite value, which is why `of_bool` also did.
`{is,to}_{bool,int}` also only return true/`Some` if it is definite.
[...]

PR #91, opened 2020-08-20
Fix branching/invariant code

jerhard:
This patch fixes the unsound behavior of `Base.invariant` (see  #90) while retaining accuracy for intervals.

To fix soundness, the abstract value that is passed to the `inv_exp` in case of `tv = true` now is now an abstract representation of  
the `set of non-zero integers`, instead of only `{1}`. 
As intervals cannot accurately represent the set of non-zero integers, information is lost in the existing implementation, when only intervals are turned on. To fix this, `invariant` now always uses `DefExc` during its computations, whether it is activated in the configuration or not.
To make this possible, I moved the code in `IntDomTuple` to a functor `IntDomTupleFunc` that takes a configuration of activated int analyses.   The `module IntDomTuple` then is created  using the provided configuration. A new `module IntDomBranchingTuple`, that always makes use of `DefExc`, is then used in the `invariant`. 

This code is based on the `def_exc_fix` branch.


vogler:
Better close this and make the PR against def_exc_fix since that's what it's based on.
Otherwise it will lump together its changes against master.

PR #92, opened 2020-08-20
Branching fix

jerhard:
This patch fixes the unsound behavior of `Base.invariant` (see  #90) while retaining accuracy for intervals.

To fix soundness, the abstract value that is passed to the `inv_exp` in case of `tv = true` now is now an abstract representation of  
the `set of non-zero integers`, instead of only `{1}`. 
As intervals cannot accurately represent the set of non-zero integers, information is lost in the existing implementation, when only intervals are turned on. To fix this, `invariant` now always uses `DefExc` during its computations, whether it is activated in the configuration or not.
To make this possible, I moved the code in `IntDomTuple` to a functor `IntDomTupleFunc` that takes a configuration of activated int analyses.   The `module IntDomTuple` then is created  using the provided configuration. A new `module IntDomBranchingTuple`, that always makes use of `DefExc`, is then used in the `invariant`.


src/analyses/base.ml
```diff
@@ -9,13 +9,15 @@ module Q = Queries
...
 module ID = ValueDomain.ID
+module IDB = IntDomain.IntDomBranchingTuple
 module IdxDom = ValueDomain.IndexDomain
 module IntSet = SetDomain.Make (IntDomain.Integers)
 module AD = ValueDomain.AD
 module Addr = ValueDomain.Addr
 module Offs = ValueDomain.Offs
 module LF = LibraryFunctions
 module CArrays = ValueDomain.CArrays
+module DefExc = IntDomain.DefExc
```
vogler, line 20: unused?

src/analyses/base.ml
```diff
@@ -9,13 +9,15 @@ module Q = Queries
 
 module GU = Goblintutil
 module ID = ValueDomain.ID
+module IDB = IntDomain.IntDomBranchingTuple
```
vogler, line 12: Push into `invariant` since it's only used there.

src/analyses/base.ml
```diff
@@ -139,6 +141,11 @@ struct
     | BNot -> ID.bitnot
     | LNot -> ID.lognot
 
+  let unop_IDB = function
```
vogler, line 144: Push into `invariant` since it's only used there.

src/analyses/base.ml
```diff
@@ -1496,7 +1498,13 @@ struct
     if eval_bool exp = Some tv then raise Deadcode
     else
       let ik = get_ikind (Cil.typeOf exp) in
-      Tuple3.map1 (fun _ -> inv_exp (ID.of_bool_ikind ik tv) exp) st
+      let tv_abs =
+        if tv
+          then IDB.of_excl_list ik [Int64.zero]
+          else IDB.of_int_ikind ik Int64.zero
+      in
+      Tuple3.map1 (fun _ -> inv_exp (tv_abs) exp) st
```
vogler, line 1506: `(tv_abs)` -> `tv_abs`

src/cdomains/intDomain.ml
```diff
@@ -1530,10 +1530,33 @@ module Enums : S = struct
...
+
+module IntDomTupleStdConf: IntDomTupleConf = struct
+  let lookup s = get_bool @@ "ana.int."^s
+  let def_exc () = lookup "def_exc"
+  let interval () = lookup "interval"
+  let cinterval () = lookup "cinterval"
+  let enums () = lookup "enums"
+end
+
+module IntDomTupleInvariantConf: IntDomTupleConf = struct
```
vogler, line 1548: Replace content with
~~~ocaml
include IntDomTupleStdConf
let def_exc () = true
~~~

src/cdomains/intDomain.ml
```diff
@@ -1530,10 +1530,33 @@ module Enums : S = struct
...
+  let cinterval () = lookup "cinterval"
+  let enums () = lookup "enums"
+end
+
 (* The above IntDomList has too much boilerplate since we have to edit every function in S when adding a new domain. With the following, we only have to edit the places where fn are applied, i.e., create, mapp, map, map2. *)
-module IntDomTuple = struct
+module IntDomTupleFunc (M: IntDomTupleConf) = struct
   include Printable.Std (* for default invariant, tag, ... *)
-
+  module Conf = M
```
vogler, line 1559: remove?

vogler (changes requested):

vogler:
> As intervals cannot accurately represent the set of non-zero integers, information is lost in the existing implementation, when only intervals are turned on.

This was the same before, no? Maybe one wants to run without exclusion sets to compare. Ok, here they're only temporarily used in `invariant`, but when does that help if it's converted back again in the end?
Do you have some examples where it makes a difference with intervals?

jerhard:
>Ok, here they're only temporarily used in invariant, but when does that help if it's converted back again in the end?

For intervals, `true` cannot possibly be recovered by `to_bool` from the abstract value in the following code snippet (from `master`):
https://github.com/goblint/analyzer/blob/a7e04ee071c56980ab651c9baff91d55acf42a2d/src/analyses/base.ml#L1365-L1388

This means, that when only intervals are used, the `invariant` cannot perform any restriction of the abstract value for the `true` branch. In the following example, the `assert` cannot be verified by Goblint, if only intervals are used:
https://github.com/goblint/analyzer/blob/c626efb15346c3408e12ebcabb788f2f0699cd03/tests/regression/01-cpa/36-interval-branching.c#L4-L7

Using `DefExc` helps `to_bool` to reconstruct `Some true` in the first code snippet, which in turn causes a more precise handling.

src/cdomains/intDomain.ml
jerhard: The `Conf` is used in the conversion function back to the regular `IntDomTuple` to check whether the `DefExc.t` value should be retained or not.
https://github.com/goblint/analyzer/blob/e5333b75677a22989c534cccf996999aa17aa12c/src/cdomains/intDomain.ml#L1690

src/analyses/base.ml
jerhard: Done, in e2c3e47715c9c3d4879a9d5dea1b583e3f767677.

src/analyses/base.ml
jerhard: Done, in b6997e81485c9e7f97903ec78eb8450e96afa6c9.

src/analyses/base.ml
jerhard: Done, in b6997e81485c9e7f97903ec78eb8450e96afa6c9.

src/analyses/base.ml
jerhard: Done, in 4976bb7a9a8f34e6fb9abe36aa3efdd04dd2902c.

src/cdomains/intDomain.ml
jerhard: Done, in e5333b75677a22989c534cccf996999aa17aa12c

src/cdomains/intDomain.ml
vogler: Ah, didn't see.

vogler:
Ok, the `{1}` is the right result to propagate for a true comparison, but it's not the right value to set for just an Lval.
If we handle the single Lval as an expression separately, there should be no need for adding DefExc.

vogler:
I now only use true=1 only for comparisons, and non_zero for everything else.
Maybe always adding exclusion sets would improve something for arithmetic operations, but I'm not sure if it's worth the added complexity.

Regarding `of_bool`, I changed the `DefExc` implementation to match the rest, which is in conformity with what the mli says:
https://github.com/goblint/analyzer/blob/93462fb7e4e00f040910e9cb85546705106e42a9/src/cdomains/intDomain.mli#L27-L29
Whether this is right/wrong is open for discussion. At least searching for `ID.of_bool` it looks like all those calls use it only for comparisons, which is fine.

vogler:
`x + 1 != 2`
For that example adding the exclusion sets would result in x != 1 but then once it's converted to an interval it's also top - so no gain.
Can you think of an example where it would improve something?

PR #93, opened 2020-08-25
DefExc: #89 without range for definite values + #95 better cast of exclusion range

vogler:
This should include all the fixes and tests from #89 minus the changes to add a range to `Def` in `IntDomain.DefExc`.
Might need some more clean up since I mostly accepted incoming changes during merge of the cherry-picked commits.

Edit: This now also includes #95.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -278,7 +278,7 @@ struct
       match op with
       | MinusA when equality () = Some true ->
         let ik = Cilfacade.get_ikind (Cil.typeOf exp) in
-        Some (`Int (ID.of_int_ikind ik 0L))
+        Some (`Int (ID.cast_to ik @@ ID.of_int 0L))
       | MinusPI
       | MinusPP when equality () = Some true -> Some (`Int (ID.of_int 0L))
       | MinusPI
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -829,8 +829,11 @@ struct
   (* run eval_rv from above, but change bot to top to be sound for programs with undefined behavior. *)
   (* Previously we only gave sound results for programs without undefined behavior, so yielding bot for accessing an uninitialized array was considered ok. Now only [invariant] can yield bot/Deadcode if the condition is known to be false but evaluating an expression should not be bot. *)
   let eval_rv (a: Q.ask) (gs:glob_fun) (st: store) (exp:exp): value =
-    let r = eval_rv a gs st exp in
-    if VD.is_bot r then top_value a gs st (typeOf exp) else r
+    try
+      let r = eval_rv a gs st exp in
+      if VD.is_bot r then top_value a gs st (typeOf exp) else r
+    with IntDomain.ArithmeticOnIntegerBot _ ->
+      top_value a gs st (typeOf exp)
 
   (* Evaluate an expression containing only locals. This is needed for smart joining the partitioned arrays where ctx is not accessible. *)
   (* This will yield `Top for expressions containing any access to globals, and does not make use of the query system. *)
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -279,7 +279,9 @@ struct
       in
       let ptrdiff_ikind = match !ptrdiffType with TInt (ik,_) -> ik | _ -> assert false in
       match op with
-      | MinusA
+      | MinusA when equality () = Some true ->
+        let ik = get_ikind (Cil.typeOf exp) in
+        Some (`Int (ID.of_int_ikind ik 0L))
       | MinusPI
       | MinusPP when equality () = Some true -> Some (`Int (ID.of_int 0L))
       | MinusPI
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -979,7 +979,11 @@ struct
         let e2_val = eval_rv ctx.ask ctx.global ctx.local e2 in
         match e1_val, e2_val with
         | `Int i1, `Int i2 -> begin
-            if ID.is_bot (ID.meet i1 i2) then
+            (* This should behave like == and also work on different int types, hence the cast (just like with == in C) *)
+            let e1_ik = get_ikind (Cil.typeOf e1) in
+            let e2_ik = get_ikind (Cil.typeOf e2) in
+            let ik= Cil.commonIntKind e1_ik e2_ik in
+            if ID.is_bot (ID.meet (ID.cast_to ik i1) (ID.cast_to ik i2)) then
               begin
                 (* Printf.printf "----------------------> NOPE may equality check for %s and %s \n" (ExpDomain.short 20 (`Lifted e1)) (ExpDomain.short 20 (`Lifted e2)); *)
                 `Bool(false)
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -100,9 +100,6 @@ struct
     | 0 -> Flag.compare x2 y2
     | x -> x
 
-  let get_ikind t = match Cil.unrollType t with TInt (ik,_) | TEnum ({ekind = ik; _},_) -> ik | _ ->
-    (* important to unroll the type here, otherwise problems with typedefs *)
-    M.warn "Something that we expected to be an integer type has a different type, assuming it is an IInt";  Cil.IInt
 
   (**************************************************************************
    * Initializing my variables
@@ -280,7 +277,7 @@ struct
       let ptrdiff_ikind = match !ptrdiffType with TInt (ik,_) -> ik | _ -> assert false in
       match op with
       | MinusA when equality () = Some true ->
-        let ik = get_ikind (Cil.typeOf exp) in
+        let ik = Cilfacade.get_ikind (Cil.typeOf exp) in
         Some (`Int (ID.of_int_ikind ik 0L))
       | MinusPI
       | MinusPP when equality () = Some true -> Some (`Int (ID.of_int 0L))
@@ -982,8 +979,8 @@ struct
         match e1_val, e2_val with
         | `Int i1, `Int i2 -> begin
             (* This should behave like == and also work on different int types, hence the cast (just like with == in C) *)
-            let e1_ik = get_ikind (Cil.typeOf e1) in
-            let e2_ik = get_ikind (Cil.typeOf e2) in
+            let e1_ik = Cilfacade.get_ikind (Cil.typeOf e1) in
+            let e2_ik = Cilfacade.get_ikind (Cil.typeOf e2) in
             let ik= Cil.commonIntKind e1_ik e2_ik in
             if ID.is_bot (ID.meet (ID.cast_to ik i1) (ID.cast_to ik i2)) then
               begin
@@ -1106,7 +1103,7 @@ struct
               | `Bool t -> Q.BD.to_bool t = Some true
               | _ -> false
             in
-            let ik = get_ikind (typeOf currentE') in
+            let ik = Cilfacade.get_ikind (typeOf currentE') in
             let newE = Basetype.CilExp.replace l' r' currentE' in
             let currentEPlusOne = BinOp (PlusA, currentE', Cil.kinteger ik 1, typeOf currentE') in
             if are_equal newE currentEPlusOne then
@@ -1228,7 +1225,11 @@ struct
       (* The true-branch where x == value: *)
       | Eq, x, value, true ->
         if M.tracing then M.tracec "invariant" "Yes, %a equals %a\n" d_lval x VD.pretty value;
-        Some (x, value)
+        (match value with
+        | `Int n ->
+          let ikind = Cilfacade.get_ikind (typeOf (Lval lval)) in
+          Some (x, `Int (ID.cast_to ikind n))
+        | _ -> Some(x, value))
       (* The false-branch for x == value: *)
       | Eq, x, value, false -> begin
           match value with
@@ -1237,7 +1238,8 @@ struct
               | Some n ->
                 (* When x != n, we can return a singleton exclusion set *)
                 if M.tracing then M.tracec "invariant" "Yes, %a is not %Ld\n" d_lval x n;
-                Some (x, `Int (ID.of_excl_list ILongLong [n]))
+                let ikind = Cilfacade.get_ikind (typeOf (Lval lval)) in
+                Some (x, `Int (ID.of_excl_list ikind [n]))
               | None -> None
             end
           | `Address n -> begin
@@ -1264,7 +1266,7 @@ struct
           let limit_from = if tv then ID.maximal else ID.minimal in
           match value with
           | `Int n -> begin
-            let ikind = get_ikind (typeOf (Lval lval)) in
+            let ikind = Cilfacade.get_ikind (typeOf (Lval lval)) in
             let n = ID.cast_to ikind n in
             let range_from x = if tv then ID.ending_ikind ikind (Int64.sub x 1L) else ID.starting_ikind ikind x in
             let limit_from = if tv then ID.maximal else ID.minimal in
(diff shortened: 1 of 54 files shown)


vogler:
Compiles but fails tests 01/10 and 27/01.
~~~
$ ./regtest 01 10 --trace eval
...
%%% eval: (tests/regression/01-cpa/10-posneg.c:14)
  eval_rv k + 1 = (Not {6}([-31,32]))
...
~~~
`eval_rv` gives the right result, but it's not set or lost.

michael-schwarz:
Btw, when comparing this with the `def_exc_fix` branch, don't rely on https://github.com/goblint/analyzer/compare/pr89_fixes...def_exc_fix, the comparison is somehow broken (e.g. shows that `27/01` in `pr89_fixes` does not have the additional tests for short and long which it does).
This just confused me for half an hour.

michael-schwarz:
>   eval_rv k + 1 = (Not {6}([-31,32]))

Notice the [-31,32]!
It is the cast to target type `IInt` that leads to this: The range of `(Not {6}([-31,32]))` is bigger than `[-31,31]` which means unsafe cast and hence `top ()`

michael-schwarz (commented):
LGTM in principle.
But I don't think we will ever get this to work completely without fixing `Def_Exc`, whether that's in the manner from #89 or some other.

michael-schwarz:
> It is the cast to target type `IInt` that leads to this: The range of `(Not {6}([-31,32]))` is bigger than `[-31,31]` which means unsafe cast and hence `top ()`.


https://github.com/goblint/analyzer/blob/f0e37b159c045fb5fed83a513330b8989c9311b6/src/cdomains/valueDomain.ml#L766-L772


When we remove this cast, it is only some asserts about `long` in `27/01` that fail, but I don't think it is what one should do.

```
Expected assert, but registered unknown on ints:34
    assert(xl != 1);
Expected assert, but registered unknown on ints:114
    assert(xl != one);
27/01 inv_invariants/ints failed!

1 tests failed: ["27/01 ints"]
``

**Comment by vogler** [comment]
> Btw, when comparing this with the `def_exc_fix` branch, don't rely on [pr89_fixes...def_exc_fix](https://github.com/goblint/analyzer/compare/pr89_fixes...def_exc_fix), the comparison is somehow broken (e.g. shows that `27/01` in `pr89_fixes` does not have the additional tests for short and long which it does).
> This just confused me for half an hour.

You need to use `..` instead of `...` (starting at a common ancestor of both): https://github.com/goblint/analyzer/compare/pr89_fixes..def_exc_fix

**Comment by vogler** [comment]
> LGTM in principle.
> But I don't think we will ever get this to work completely without fixing `Def_Exc`, whether that's in the manner from #89 or some other.

Yes, the idea was to see what fails without the range and if it's easy to fix.

File: src/cdomains/intDomain.ml [hunk 1]
```diff
@@ -743,13 +778,36 @@ struct
...
+  let lift2_special f x y = match x,y with
     (* The good case: *)
     | `Definite x, `Definite y -> (try `Definite (f x y) with | Division_by_zero -> top ())
-    (* If any one of them is bottom, we return top *)
-    | _ -> top ()
+    (* We don't bother with exclusion sets: *)
+    | `Excluded (_, r), `Definite _
+    | `Definite _, `Excluded (_, r)
+    | `Excluded (_, r), `Excluded (_, _) ->
+      `Excluded (S.empty (), r)
```
michael-schwarz, line 806: `Excluded (S.empty (), r)` was sound only because `r` was the full range of the left argument. When changing this, it is now unsound as well (on #95 as well, but added it here since #95 does not touch this piece of code).

```
x = 1;
if(top) {
   x =2;
}
// x→ Exc{0} [0,8]

y = x << 10; 
// y → Exc {} [0,8]
```

src/cdomains/intDomain.ml
```diff
@@ -813,20 +879,20 @@ struct
...
   let rem  = lift2 Integers.rem
-  let lt = lift2 Integers.lt
-  let gt = lift2 Integers.gt
-  let le = lift2 Integers.le
-  let ge = lift2 Integers.ge
+  let lt x y = lift2 Integers.lt x y
+  let gt x y = lift2 Integers.gt x y
+  let le x y = lift2 Integers.le x y
+  let ge x y = lift2 Integers.ge x y
   let bitnot = lift1 Integers.bitnot
```
michael-schwarz, line 886: Similar issues as for the `lift2_special`:
```
    int x = 1;
    if(top) {
        x =2;
    }
    // x -> Not {0}([0,8])

    x = ~x; 
    // x -> Not {-1}([0,8])
```

michael-schwarz:
The reason why it is difficult to see if the way ranges are implemented is ok is that they are hardly ever used during asserts: e.g.

```
    int top;
    int x = 1;
    int y;

    int x = 1;
    if(top) {
        x =2;
    }

    x = ~x;

    if(x < 0) {
        assert(5==5);
    }
```
Will happily go into the if branch, even though the ranges say it should be unreachable.

src/cdomains/intDomain.ml
vogler: Thanks. There might be several of those changes that need to be reverted.

michael-schwarz: E.g. `lift2` as well

src/cdomains/intDomain.ml
vogler: The range needs to be adjusted after every arithmetic operation on `Exc`. This is done in `lift2_inj`, not needed in `lift2`, and still missing in `lift1`.

michael-schwarz: Why is it not needed in `lift2`? It is the one you call for shifts for example?

https://github.com/goblint/analyzer/blob/1e9119542b806aef0e96a86ed733f37e95f4e512/src/cdomains/intDomain.ml#L780-L794

vogler:
> The reason why it is difficult to see if the way ranges are implemented is ok is that they are hardly ever used during asserts or when branching. E.g.

True, but as written above: The range needs to be adjusted after every arithmetic operation on `Exc`. This is done in `lift2_inj`, not needed in `lift2` since it just goes to top, and missing in `lift1`.
One could insert asserts that check that every excluded value is in range, or just abstract out the range computation and call it in those few places.

src/cdomains/intDomain.ml
vogler: Fixed. I meant there's no computation done on `Excluded` in `lift2`, so no need to adjust the range.

vogler:
#95 works now. I guess most of the `ID.cast_to` inserted here are no longer needed. Idk if we should keep them.

michael-schwarz:
Looking at some of the changes in #89  that might still be worth keeping:

I changed this:
https://github.com/goblint/analyzer/blob/52fb0ba7bb4f998be751a6eb2342b241c3d0c6c5/src/analyses/base.ml#L1319-L1325

to:

```
(* This is tricky: It it is not sufficient to check that ID.cast_to_ik v = v                *)
(* If there is one domain that knows this to be true and the other does not we *)	
(* should still impose the invariant. E.g. i -> ([1,5]; Not {0}[byte])                       *)	
if VD.is_safe_cast t1 (Cil.typeOf (Lval x)) then
```

I think this is probably still the better thing to do and I would include it.

michael-schwarz:
> most of the `ID.cast_to` are no longer needed. Idk if we should keep them.
I would be in favor of keeping them. Some future int domain might rely on things being cast to the correct ik. If we add such a domain, we would need to go through the trouble of finding the correct `ik` all over again.

src/analyses/base.ml
```diff
@@ -1399,7 +1422,7 @@ struct
         let m = ID.meet a b in
         (match op, ID.to_bool c with
         | Eq, Some true
-        | Ne, Some false -> both m (* def. equal *)
+        | Ne, Some false -> both m (* def. equal: if they compare equal, both values must be from the meet *)
         | Eq, Some false
         | Ne, Some true -> (* def. unequal *)
           (* Both values can not be in the meet together, but it's not sound to exlcude the meet from both. *)
```
michael-schwarz, line 1428: Typo: exlcude -> exclude

vogler:
> > most of the `ID.cast_to` are no longer needed. Idk if we should keep them.
> 
> I would be in favor of keeping them. Some future int domain might rely on things being cast to the correct ik. If we add such a domain, we would need to go through the trouble of finding the correct `ik` all over again.

Yea, probably. I'm not sure if they maybe can make something imprecise by casting too early. But I also don't feel like going through them and check. Other cons: readability, execution overhead.

> might rely on things being cast to the correct ik
Everything should be cast to the correct ik without those calls, just later.

vogler:
Apparently I already removed a bunch of the unnecessary casts while merging: https://github.com/goblint/analyzer/compare/pr89_fixes..def_exc_fix

Copied from https://github.com/goblint/analyzer/pull/95#discussion_r482974051 (these diffs should belong to this branch): 

Some things that are actually different:

base:
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R230
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R691
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1344
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1417-R1418
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1482-R1483
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R2050

valueDomain:
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R403
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R431
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R459
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R524

michael-schwarz:
**base:**
> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R230

I'd stick with `Int(ID.of_int z)` from this over `| Some z -> `Int(diff)` from #89. This way, we propagate the definite value to all domains.

> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R691

I'd go with #89 here. Should not change anything about the values we have, as we keep the smaller range if we do a cast that's safe.

> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1344

I'd go with #89. See here: https://github.com/goblint/analyzer/pull/93#issuecomment-686419797

> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1417-R1418

I'd suggest #89. Can't hurt to give the `ikind` here.

> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1482-R1483

#89 as well here.

> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R2050

#89 as well: I think it suffices if one domain knows it to be zero, it is not required that every domain knows it.

michael-schwarz:
**valueDomain:**
> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R403
> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R431
> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R459
> https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R524

Those lead to going to `VD.top ()` as soon as the field that was used to access a union is `Top`. It does not matter in the end, since we will read `Top` as soon as the field is `Lifted _` for unions in `eval_offset` anyway:

https://github.com/goblint/analyzer/blob/6f0f3b5fa4a1bda2347e6cef9ccc55dbeb94fb9a/src/cdomains/valueDomain.ml#L719-L725

But maybe it is nice to enforce this `Field = top () -> Value = VD.top ()`? Not sure.

vogler:
> **base:**
> 
> > https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R230
> 
> I'd stick with `Int(ID.of_int z)` from this over `| Some z -> `Int(diff)` from #89. This way, we propagate the definite value to all domains.

Included.

> > https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R691
> 
> I'd go with #89 here. Should not change anything about the values we have, as we keep the smaller range if we do a cast that's safe.

Included. What was the reason for it?

> > https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1344
> 
> I'd go with #89. See here: [#93 (comment)](https://github.com/goblint/analyzer/pull/93#issuecomment-686419797)

Included.

> > https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1417-R1418
> 
> I'd suggest #89. Can't hurt to give the `ikind` here.

Not needed due to meet.

> > https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1482-R1483
> 
> #89 as well here.

Not needed since it's always true.

> > https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R2050
> 
> #89 as well: I think it suffices if one domain knows it to be zero, it is not required that every domain knows it.

Included. I thought the problem was just that `of_int 0L` was missing the def. range in #89 and therefore `=` failed.

vogler:
Ok, LGTM, anything else before merge?

michael-schwarz:
Nope, let's merge it!

michael-schwarz:
I found one more issue from the random tests generated by CSmith that I started this morning:
After extracting it from the big program, this is what it boiled down to:

~~~c
int main ()
{
  int tmp;
  int p_9 = 60;

  tmp = (p_9 +1) % 0; // A

  if ((p_9 +1) % 0) { // B
    tmp = 1;
  }

  return (0);
}
~~~

We fail at B with:

```
Fatal error: exception Dune__exe__IntDomain.ArithmeticOnIntegerBot("Error int op 60")
Raised at file "src/cdomains/intDomain.ml", line 838, characters 6-90
Called from file "src/cdomains/intDomain.ml", line 1500, characters 62-69
Called from file "src/cdomains/intDomain.ml", line 1501, characters 44-78
Called from file "src/analyses/base.ml", line 1389, characters 45-53
Called from file "src/analyses/base.ml", line 1477, characters 23-52
Called from file "src/analyses/base.ml", line 1479, characters 19-32
```

This is of course UB, but we probably should:
- warn about the problem / fail with an understandable error message
- behave like we do if it is not in a condition and just continue (this is what happens at A)

vogler:
Yes, `warning: remainder by zero is undefined [-Wdivision-by-zero]`.
I guess check for `b = 0` in `Div` and `Mod` in `invariant`, continue with top and warn.

PR #95, opened 2020-09-01
DefExc: better cast of exclusion range

vogler:
Based on #93.
Test 01/29 fails because it is more precise now. All other tests pass.
The only change is that the exclusion range is kept small on injection/cast, i.e. it grows only by join/computation.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1432,7 +1432,10 @@ struct
           (* e.g. a=[0,1], b=[1,2], meet a b = [1,1], but (a != b) does not imply a=[0,0], b=[2,2] since others are possible: a=[1,1], b=[2,2] *)
           (* Only if a is a definite value, we can exclude it from b: *)
           let excl a b = match ID.to_int a with Some x -> ID.of_excl_list ik [x] | None -> b in
-          meet_bin (excl b a) (excl a b)
+          let a' = excl b a in
+          let b' = excl a b in
+          if M.tracing then M.tracel "inv" "inv_bin_int: unequal: %a and %a; ikind: %a; a': %a, b': %a\n" ID.pretty a ID.pretty b d_ikind ik ID.pretty a' ID.pretty b';
+          meet_bin a' b'
         | _, _ -> a, b
         )
       | Lt | Le | Ge | Gt as op ->
@@ -1457,10 +1460,6 @@ struct
     let eval e = eval_rv a gs st e in
     let eval_bool e = match eval e with `Int i -> ID.to_bool i | _ -> None in
     let set' lval v = Tuple3.first (set a gs st (eval_lv a gs st lval) v ~effect:false ~change_array:false ~ctx:(Some ctx)) in
-    let is_cmp = function
-      | BinOp ((Lt | Gt | Le | Ge | Eq | Ne), _, _, t) -> true
-      | _ -> false
-    in
     let rec inv_exp c = function
       | UnOp (op, e, _) -> inv_exp (unop_ID op c) e
       | BinOp(op, CastE (t1, c1), CastE (t2, c2), t) when (op = Eq || op = Ne) && typeSig t1 = typeSig t2 && VD.is_safe_cast t1 (typeOf c1) && VD.is_safe_cast t2 (typeOf c2) ->
@@ -1469,10 +1468,11 @@ struct
         if M.tracing then M.tracel "inv" "binop %a with %a %s %a == %a\n" d_exp e VD.pretty (eval e1) (show_binop op) VD.pretty (eval e2) ID.pretty c;
         (match eval e1, eval e2 with
         | `Int a, `Int b ->
-          let ik = Cilfacade.get_ikind @@ typeOf @@ if is_cmp e then e1 else e in
-          let a', b' = inv_bin_int (a, b) ik (ID.cast_to ik c) op in
-          let m1 = inv_exp (ID.cast_to (Cilfacade.get_ikind (Cil.typeOf e1)) a') e1 in
-          let m2 = inv_exp (ID.cast_to (Cilfacade.get_ikind (Cil.typeOf e2)) b') e2 in
+          let ik = Cilfacade.get_ikind @@ typeOf e1 in (* both operands have the same type! *)
+          let a', b' = inv_bin_int (a, b) ik c op in
+          if M.tracing then M.tracel "inv" "binop: %a, a': %a, b': %a\n" d_exp e ID.pretty a' ID.pretty b';
+          let m1 = inv_exp a' e1 in
+          let m2 = inv_exp b' e2 in
           CPA.meet m1 m2
         (* | `Address a, `Address b -> ... *)
         | a1, a2 -> fallback ("binop: got abstract values that are not `Int: " ^ sprint VD.pretty a1 ^ " and " ^ sprint VD.pretty a2))
@@ -1509,6 +1509,10 @@ struct
     in
     if eval_bool exp = Some (not tv) then raise Deadcode (* we already know that the branch is dead *)
     else
+      let is_cmp = function
+        | BinOp ((Lt | Gt | Le | Ge | Eq | Ne), _, _, t) -> true
+        | _ -> false
+      in
       let itv = (* int abstraction for tv *)
         if not tv || is_cmp exp then (* false is 0, but true can be anything that is not 0, except for comparisons which yield 1 *)
           ID.of_bool tv (* this will give 1 for true which is only ok for comparisons *)
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1431,7 +1431,7 @@ struct
           (* Both values can not be in the meet together, but it's not sound to exlcude the meet from both. *)
           (* e.g. a=[0,1], b=[1,2], meet a b = [1,1], but (a != b) does not imply a=[0,0], b=[2,2] since others are possible: a=[1,1], b=[2,2] *)
           (* Only if a is a definite value, we can exclude it from b: *)
-          let excl a b = match ID.to_int a with Some x -> ID.of_excl_list ILongLong [x] | None -> b in
+          let excl a b = match ID.to_int a with Some x -> ID.of_excl_list ik [x] | None -> b in
           meet_bin (excl b a) (excl a b)
         | _, _ -> a, b
         )
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -161,6 +161,7 @@ struct
 
   (* Evaluate binop for two abstract values: *)
   let evalbinop (op: binop) (t1:typ) (a1:value) (t2:typ) (a2:value): value =
+    if M.tracing then M.tracel "eval" "evalbinop %a %a %a\n" d_binop op VD.pretty a1 VD.pretty a2;
     (* We define a conversion function for the easy cases when we can just use
      * the integer domain operations. *)
     let bool_top () = ID.(join (of_int 0L) (of_int 1L)) in
@@ -271,7 +272,9 @@ struct
     let binop op e1 e2 =
       let equality () =
         match ask (Q.ExpEq (e1,e2)) with
-        | `Bool x -> Some x
+        | `Bool x ->
+          if M.tracing then M.tracel "query" "ExpEq (%a, %a) = %b\n" d_exp e1 d_exp e2 x;
+          Some x
         | _ -> None
       in
       let ptrdiff_ikind = match !ptrdiffType with TInt (ik,_) -> ik | _ -> assert false in
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1489,19 +1489,22 @@ struct
         let v = VD.meet oldv c' in
         if is_some_bot v then raise Deadcode
         else (
-          if M.tracing then M.tracel "inv" "improve lval %a = %a with %a (from %a), meet = %a\n" d_lval x VD.pretty oldv VD.pretty c' ID.pretty c VD.pretty v;
+          if M.tracing then M.tracel "inv" "improve lval %a from %a to %a (c = %a, c' = %a)\n" d_lval x VD.pretty oldv VD.pretty v ID.pretty c VD.pretty c';
           set' x v
         )
       | Const _ -> Tuple3.first st (* nothing to do *)
       | CastE ((TInt (ik, _)) as t, e) -> (* Can only meet the t part of an Lval in e with c (unless we meet with all overflow possibilities)! Since there is no good way to do this, we only continue if e has no values outside of t. *)
         (match eval e with
-        | `Int a ->
-          if ID.leq a (ID.cast_to ik a) then
+        | `Int i ->
+          if ID.leq i (ID.cast_to ik i) then
              match Cil.typeOf e with
-              | TInt(ik_e, _) -> inv_exp (ID.cast_to ik_e c) e
+              | TInt(ik_e, _) ->
+                let c' = ID.cast_to ik_e c in
+                if M.tracing then M.tracel "inv" "cast: %a from %a to %a: i = %a; cast c = %a to %a = %a\n" d_exp e d_ikind ik_e d_ikind ik ID.pretty i ID.pretty c d_ikind ik_e ID.pretty c';
+                inv_exp c' e
               | x -> fallback ("CastE: e did evaluate to `Int, but the type did not match" ^ sprint d_type t)
           else
-            fallback ("CastE: " ^ sprint d_plainexp e ^ " evaluates to " ^ sprint ID.pretty a ^ " which is bigger than the type it is cast to which is " ^ sprint d_type t)
+            fallback ("CastE: " ^ sprint d_plainexp e ^ " evaluates to " ^ sprint ID.pretty i ^ " which is bigger than the type it is cast to which is " ^ sprint d_type t)
         | v -> fallback ("CastE: e did not evaluate to `Int, but " ^ sprint VD.pretty v))
       | e -> fallback (sprint d_plainexp e ^ " not implemented")
     in
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1379,7 +1379,8 @@ struct
       Tuple3.first (invariant ctx a gs st exp tv)
     in
     (* inverse values for binary operation a `op` b == c *)
-    let inv_bin_int (a, b) ik c =
+    (* ikind is the type of a for limiting ranges of the operands a, b. The only binops which can have different types for a, b are Shiftlt, Shiftrt (not handled below; don't use ikind to limit b there). *)
+    let inv_bin_int (a, b) ikind c =
       let meet_bin a' b'  = ID.meet a a', ID.meet b b' in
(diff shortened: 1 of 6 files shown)


michael-schwarz:
For [df_comb](https://github.com/goblint/bench/blob/master/coreutils/df_comb.c) we get: 

```
About to crash on vasnprintf.c:4906
fd-reopen.c:34: Error: Cannot resolve variable tmp.
fd-reopen.c:34: Error: Expected lval for assignment. Got Const(Int64(0,int,None))
fd-reopen.c:34: Error: Cannot resolve variable tmp.
Fatal error: exception Dune__exe__IntDomain.ArithmeticOnIntegerBot("[-1,-1] op bottom")
Raised at file "src/cdomains/intDomain.ml", line 299, characters 27-111
Called from file "src/cdomains/intDomain.ml", line 1506, characters 62-69
Called from file "src/cdomains/intDomain.ml", line 1507, characters 80-114
Called from file "src/analyses/base.ml", line 1402, characters 51-65
Called from file "src/analyses/base.ml", line 1474, characters 23-65
Called from file "src/analyses/base.ml", line 1476, characters 19-83
Called from file "src/batTuple.ml", line 128, characters 24-27
Called from file "src/analyses/base.ml", line 1667, characters 16-65
Called from file "src/analyses/mCP.ml", line 441, characters 17-35
Called from file "src/analyses/mCP.ml", line 293, characters 46-72
```

michael-schwarz:
> For df_comb we get

I had the same issue on #89 at some point too, IIRC it was fixed by making sure to have the correct `ikind` everywhere, not sure if this is helpful here.

vogler:
Haven't checked before the merge, but now I get no exception, but FP problems:
~~~
$ ./goblint ../bench/coreutils/df_comb.c
...
Fixpoint not reached at node "! (amt != 0UL)" on human.c:346 (human.c:346)
...
Difference: Map: p =
 HoarePO: {buf___1[Unknown int([0,64])]} not leq {buf___1[Unknown int([-63,63])]}. 
~~~

vogler:
Ok, with these changes it reaches the FP.
WIP. Needs some more thought on overflows.
Range is now limited to int, but what one probably wants is to have `from` and `to` for casts in order to know if it's really a downcast or not.

For exclusion ranges that come from the join of definite values we want to start with the smallest range and go up until we reach the range of the static type.
For exclusion ranges from unknown values we start with the full range.
The question is what to do on operations that can overflow. Currently `lift_inj` treats case 2 as case 1, i.e., it increases the range, and then the cast has to handle potential overflows (hard to do for exclusion sets).

vogler:
I think this should be it. Let me know if you find something wrong or if we should add more comments.

The idea is as described above. Summary of what happens:
Top values start with the full range (as in #89).
Definite values don't need a range. On join of two definite values, we get an exclusion set with the minimal range.
On arithmetic operations, this range might grow.
Values are cast to their static type whenever they're set or if there is an actual cast. `cast_to` deals with overflows/wrap around (so far same as before).
In `cast_to`, the range can 1) still be smaller than the target range -> do nothing, 2) be bigger due to an operation but without a change of the static type, 3) be bigger due to a downcast.
For 2) we can just cast the set and crop the range.
For 3) I now also handle overflows in the exclusion set precisely instead of going to top. The idea is that overflowed excluded values can't coincide with a new included value that overflowed. I guess this only holds if sizes are divisible by each other, but that's the case in C.

src/cdomains/intDomain.ml
```diff
@@ -616,9 +617,8 @@ struct
     | `Bot
   ] [@@deriving to_yojson]
   let name () = "def_exc"
-  let top_of ik = `Excluded (S.empty (), size ik)
-  let top_size = Size.max `Signed
-  let top () = top_of top_size
+  let top_range = R.of_interval (-99L, 99L) (* Since there is no top ikind we use a range that includes both ILongLong [-63,63] and IULongLong [0,64]. Only needed for intermediate range computation on longs. Correct range is set by cast. *)
+  let top () = `Excluded (S.empty (), top_range)
```
michael-schwarz, line 621: Since we call `ID.top ()` in quite some places, values of range [-99,99] have the potential to propagate to a lot of places. Maybe it would be good to once again to the `cast_to ik` wherever we know the `ik` for an `ID.top ()` to limit this.

src/cdomains/intDomain.ml
```diff
@@ -769,7 +788,9 @@ struct
   (* Default behaviour for unary operators, simply maps the function to the
    * DefExc data structure. *)
   let lift1 f x = match x with
-    | `Excluded (s,r) -> `Excluded (S.map f s, r)
+    | `Excluded (s,r) ->
+      let s' = S.map f s in
```
michael-schwarz, line 792: Imo, this is still unsound:

- The range can also change if the set of excluded values is empty: Consider e.g. `neg`

~~~c
  int test = 0;
  int top;
  if(top) {
    test = 2;
  }
  int test2 = -test; 
  // We currently have test2 -> Excluded {} [0,8]
~~~

- I don't think changing the range so it fits whatever the excluded values map to is enough anyway: The values that are included might map to things that require a larger range than what we could exclude, right?

src/cdomains/intDomain.ml
```diff
@@ -759,7 +772,13 @@ struct
     | `Excluded (s,r) -> min_of_range r
     | `Bot -> None
 
-  let of_excl_list t l = `Excluded (List.fold_right S.add l (S.empty ()), size t)
+  (* calculates the minimal extension of range r to cover the exclusion set s *)
+  let extend_range r s = S.fold (fun i s -> R.join s (size @@ Size.min_for i)) s r
```
michael-schwarz, line 776: Could this also produce ranges that are not in C such as `[-7,8]`?
If yes, do we want such ranges?

michael-schwarz:
~~~c 
int test = 3;
int top;
if(top) {
    test = 2;
 }

 if(test != 5) {
   int test3 = -test;
   // test3 → Not {-5, 0}([-7,8])
 }
~~~

michael-schwarz: I guess this also happens in `join` and `meet`, so not only here. So I guess it's fine.

src/cdomains/intDomain.ml
```diff
@@ -648,12 +648,24 @@ struct
   | `Definite x -> if i = x then `Eq else `Neq
   | `Excluded (s,r) -> if S.mem i s then `Top else `Neq
 
-  let cast_to t = function
+  let top_of ik = `Excluded (S.empty (), size ik)
+  let top_if_not_in_int64 ik f x = try f x with Size.Not_in_int64 -> top_of ik
+  let cast_to ?torg ik = top_if_not_in_int64 ik @@ function
```
michael-schwarz, line 653: I think this `?torg` may be benefit from some comments, it's not completely clear what it means to me

src/cdomains/intDomain.ml
vogler: Yes, it can be seen as a poison value that should never make its way into results (at least for C integers).
This PR is based on #93, so it should still contain those cast calls.

src/cdomains/intDomain.ml
vogler: Right, idk why I did it differently for `lift1` - should just be the same as for `lift2_inj`, i.e., apply f on the range.

src/cdomains/intDomain.ml
michael-schwarz: I was comparing against #89 and there is a few more casts there, not sure which ones (if any) one would like to also have here:
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix

src/cdomains/intDomain.ml
vogler: > Could this also produce ranges that are not in C such as [-7,8]? If yes, do we want such ranges?

Yes, why not? You could also manage an interval for the actual values instead of the sizes, if you want it to be more precise.

src/cdomains/intDomain.ml
vogler: I'll put in in the .mli since that's where most comments are.
Aside: It's a bit annoying to have two places for it.
~~~sh
$ fd .mli   
src/analyses/libraryFunctions.mli
src/cdomains/arrayDomain.mli
src/cdomains/arrayDomain_deprecated.mli
src/cdomains/intDomain.mli
src/cdomains/structDomain.mli
~~~
Not that many left...

src/cdomains/intDomain.ml
```diff
@@ -648,12 +648,24 @@ struct
...
+        if R.leq r r' then (* upcast -> no change *)
+          s, r
+        else if torg = None then (* same static type -> no overflows for r, but we need to cast s since it may be out of range after lift2_inj *)
+          let s' = S.map (Integers.cast_to ik) s in
+          s', r'
+        else (* downcast: may overflow *)
+          let s' = S.map (Integers.cast_to ik) s in
+          (* We want to filter out all i in s' where (t)x with x in r could be i. *)
+          (* Since this is hard to compute, we just keep all i in s' which overflowed, since those are safe - all i which did not overflow may now be possible due to overflow of r. *)
+          S.diff s' s, r'
```
michael-schwarz, line 666: What if several different values in the original type (of which one was excluded) map to the same `x \in S'`? In this case we could not exclude `x`.

Are we sure this can not happen for casts in C? To me, it seems like this is bound to happen when casting something like an `unsigned long` into something much smaller such as an `unsigned char`?

I can try to come up with an example for this.

src/cdomains/intDomain.ml
vogler: Apparently none of them are needed. All those casts in base make the code pretty hard to read.

src/cdomains/intDomain.ml
michael-schwarz: It actually happens:
~~~c
#include "stdio.h"
int main() {
    unsigned long acc = 42;

    for(int i=0;i<10;i++) {
        unsigned char c = (char)acc;
        acc += 256;
        printf("%i\n", c);
    }
}
~~~~
Will print `42` very time.

So a case where our thing is unsound is e.g.:
~~~c
  unsigned long acc;

  if(acc != 511) {
      unsigned char c = (char)acc;
      // c → Not {255}([0,8])
  }
~~~

src/cdomains/intDomain.ml
vogler: Some things that are actually different:

base:
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R230
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R691
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1344
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1417-R1418
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R1482-R1483
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-a9a7366a202e75b5e023ecd14b20ce90R2050

valueDomain:
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R403
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R431
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R459
https://github.com/goblint/analyzer/compare/fix/ranges..def_exc_fix#diff-d1d6b7a9c4034aaf9a2f822ca2cf0b02R524

src/cdomains/intDomain.ml
vogler: Done.

src/cdomains/intDomain.ml
vogler: Let's go through those in https://github.com/goblint/analyzer/pull/93#issuecomment-686506489 since they're changes in the parent of this branch. Probably accept most of them?

src/cdomains/intDomain.ml
vogler: What's wrong with it?

~~~c
  unsigned long ul = 511;
  char c = ul;
  unsigned char uc = ul;
  printf("%lu %d %d\n", ul, c, uc);
  // 511 -1 255
~~~

michael-schwarz: Our analysis claims that c in the example above can not be `255`. Which is wrong.

michael-schwarz:
~~~c
  unsigned long acc = 7;

  if(top) {
    acc = 255;
  }


  if(acc != 511) {
      unsigned char c = (char)acc;
      assert(c != 255);
      int b = 42;
  }
~~~

vogler: > Yes, but just because acc can not be 511, it does not mean that acc can not be 255.

Yes, but we're talking about `c`, not `acc`. `c` can not be 255 since that's what 511 becomes after casting.

vogler: That's the whole point: when an excluded value overflows, it doesn't make a difference if it's excluded or definite.
It cannot be 'overwritten' by a value in the included range. The ones that don't overflow don't change, and the ones that do can't hit it since it also overflowed.
If they overflowed several times, they will also hit the same spot since `(n*t) % t` stays constant.

michael-schwarz: Assume a concrete execution where `top` is `true`.
`acc` therefore is `255`, which is not `511`, so we go into the if branch. Casting `255` to `unsigned char` yields `255`. Therefore, `c` is `255` and our assert fails. 

The analysis says it will succeed.

michael-schwarz: No, no better idea.

I think the only case in which we can do better than top is if we have an exclusion set that excludes all values that will map to a certain value in the smaller type. That's probably expensive to check and I also don't know when we would ever have such an exclusion set.

vogler: I wonder if there's something one could do. Test 21/03 fails now:
~~~c
  char a;
  if(a == 127) return;
  char b = a+1; // (char)128 = -128
  assert(b != -128);
~~~

src/analyses/base.ml
```diff
@@ -1469,10 +1468,11 @@ struct
         if M.tracing then M.tracel "inv" "binop %a with %a %s %a == %a\n" d_exp e VD.pretty (eval e1) (show_binop op) VD.pretty (eval e2) ID.pretty c;
         (match eval e1, eval e2 with
         | `Int a, `Int b ->
-          let ik = Cilfacade.get_ikind @@ typeOf @@ if is_cmp e then e1 else e in
-          let a', b' = inv_bin_int (a, b) ik (ID.cast_to ik c) op in
-          let m1 = inv_exp (ID.cast_to (Cilfacade.get_ikind (Cil.typeOf e1)) a') e1 in
-          let m2 = inv_exp (ID.cast_to (Cilfacade.get_ikind (Cil.typeOf e2)) b') e2 in
+          let ik = Cilfacade.get_ikind @@ typeOf e1 in (* both operands have the same type! *)
```
michael-schwarz, line 1471: Why? Not true e.g. for shift in CIL

michael-schwarz: Also, if it is a comparison, then `ik` is `CIl.IInt` regardless of the type of `e1` and `e2`

vogler: We don't handle shift. Is it the only one?
I can revert it to `let ik = Cilfacade.get_ikind @@ typeOf @@ if is_cmp e then e1 else e in`, but this would also be wrong for shift since there's just one `ik` and it's not that of the result `e`.

michael-schwarz:
```
          match bop with
            (Mult | Div) ->
              typeMatch t1 t2; checkArithmeticType tres;
              typeMatch t1 tres; tres
          | (Eq|Ne|Lt|Le|Ge|Gt) ->
              typeMatch t1 t2; checkArithmeticType t1;
              typeMatch tres intType; tres
          | Mod|BAnd|BOr|BXor ->
              typeMatch t1 t2; checkIntegralType tres;
              typeMatch t1 tres; tres
          | LAnd | LOr ->
              checkScalarType t1; checkScalarType t2;
              typeMatch tres intType; tres
          | Shiftlt | Shiftrt ->
              typeMatch t1 tres; checkIntegralType t1;
              checkIntegralType t2; tres
          | (PlusA | MinusA) ->
                typeMatch t1 t2; typeMatch t1 tres;
                checkArithmeticType tres; tres
          | (PlusPI | MinusPI | IndexPI) ->
              checkPointerType tres;
              typeMatch t1 tres;
              checkIntegralType t2;
              tres
          | MinusPP  ->
              checkPointerType t1; checkPointerType t2;
              typeMatch t1 t2;
              typeMatch tres !ptrdiffType;
              tres

```

so `LAnd | LOr | Shiftlt | Shiftrt | (PlusPI | MinusPI | IndexPI) | MinusPP `.



Probably, it should even be the type of `e1`, since we now use `ik` like this:

```
      | Eq | Ne as op ->
        let both x = x, x in
        let m = ID.meet a b in
        (match op, ID.to_bool c with
        | Eq, Some true
        | Ne, Some false -> both m (* def. equal: if they compare equal, both values must be from the meet *)
        | Eq, Some false
        | Ne, Some true -> (* def. unequal *)
          (* Both values can not be in the meet together, but it's not sound to excude the meet from both. *)
          (* e.g. a=[0,1], b=[1,2], meet a b = [1,1], but (a != b) does not imply a=[0,0], b=[2,2] since others are possible: a=[1,1], b=[2,2] *)
          (* Only if a is a definite value, we can exclude it from b: *)
          let excl a b = match ID.to_int a with Some x -> ID.of_excl_list ik [x] | None -> b in
          meet_bin (excl b a) (excl a b)
        | _, _ -> a, b
```

michael-schwarz: Which are probably all things we should add a comment about

vogler: > so LAnd | LOr | Shiftlt | Shiftrt | (PlusPI | MinusPI | IndexPI).

`LAnd | LOr` can't appear since CIL transforms them into statements.
The rest we don't handle in `invariant`.

> Probably, it should even be the type of e1, since we now use ik like this:

That's what it is right now. What was your idea about `ik`? It's the type of the operands - the type of c is not of interest.

Aside: You can use ~~~ocaml instead of ~~~ for proper highlighting on github.

michael-schwarz: > What was your idea about ik? It's the type of the operands - the type of c is not of interest.

I found it necessary to cast `c` to `ik` as well at some point for handling more complicated expressions that involve several levels of arithmetic operations. But I'm not sure if it still is required.

> Aside: You can use ~~~ocaml instead of ~~~ for proper highlighting on github.

I didn't know it works for OCaml to (and not only C), thx!

PR #96, opened 2020-09-10
Fix Enums (c.f. #93)

michael-schwarz:
This mirrors the changes made in #93 to `Def_Exc`:

- Change `cast_to` to handle casts correctly
- Introduce a new special range for top: `(-99L, 99)` bigger than any C-type

Changes to the functions for arithmetic operations were not needed here, as the domain does not compute with `Exc` values and instead goes to top.

What is a bit surprising is that enabling `enums` on top of `interval` and `def_exc` might sometimes lead to a solution that is less precise:

~~~C
  for(int x = 0; x < 10; x++){
    assert(x >= 0);  // UNKNOWN
    assert(x <= 9);
  }
~~~

This seems to be because the false branch remains unreachable for more iterations, causing the analysis to behave differently.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1650,6 +1650,18 @@ struct
   let branch ctx (exp:exp) (tv:bool) : store =
     Locmap.replace Deadcode.dead_branches_cond !Tracing.next_loc exp;
     let valu = eval_rv ctx.ask ctx.global ctx.local exp in
+    let refine () =
+      let res = invariant ctx ctx.ask ctx.global ctx.local exp tv in
+      if M.tracing then M.tracec "branch" "EqualSet result for expression %a is %a\n" d_exp exp Queries.Result.pretty (ctx.ask (Queries.EqualSet exp));
+      if M.tracing then M.tracec "branch" "CondVars result for expression %a is %a\n" d_exp exp Queries.Result.pretty (ctx.ask (Queries.CondVars exp));
+      if M.tracing then M.traceu "branch" "Invariant enforced!\n";
+      match ctx.ask (Queries.CondVars exp) with
+      | `ExprSet s when Queries.ES.cardinal s = 1 ->
+        let e = Queries.ES.choose s in
+        M.debug_each @@ "CondVars result for expression " ^ sprint d_exp exp ^ " is " ^ sprint d_exp e;
+        invariant ctx ctx.ask ctx.global res e tv
+      | _ -> res
+    in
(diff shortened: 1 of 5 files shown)


vogler:
> What is a bit surprising is that enabling enums on top of interval and def_exc might sometimes lead to a solution that is less precise:

Only with all three enabled?

>  This seems to be because the false branch remains unreachable for more iterations, causing the analysis to behave differently.

false branch of the for loop? How does that influence the first assertion?

michael-schwarz:
> Only with all three enabled?

Enabling `enums` on top of `interval` already causes this.

> false branch of the for loop? How does that influence the first assertion?

Yes. Seems crazy, yes. I looked for a bug for a while, but found none.
The value that `interval` has for `x` inside the loop changes from `[0,9]` to `[-2147483647,9]` with `enums`.

If you change `Some true` to `None` here, the analysis will consider the false branch live sooner, making the assert succeed:
https://github.com/goblint/analyzer/blob/ed1968e3067013c11d8c5a0c8ecd4a8af8ecf9d2/src/cdomains/intDomain.ml#L1457
I had it narrowed down so that the calls from `branch` were the only calls of `to_bool`.

michael-schwarz:
It works with `--sets solver WL`, fails with `TD3`, `SLR3`, `new`.

michael-schwarz:
Thanks for spotting this @jerhard, I had been staring at it for three hours yesterday!

jerhard:
So the issue originated from the fact that due to the additional information from the `Enums` domain, we get a definite value from `ID.is_bool value`  in `Base.branch` in the following code after the first widening occured. 

In particular, we have the information:
` x -> ([0,2147483647],{0, 1})` (for the interval and enum domain) at the widening point.
When we do the branching from this value into the body of the loop, we did not refine the abstract value for the interval domain, because we don't call `invariant` here. 
https://github.com/goblint/analyzer/blob/ed1968e3067013c11d8c5a0c8ecd4a8af8ecf9d2/src/analyses/base.ml#L1658-L1670

So now, we do the refinement with the `invariant` even when we have a definite value for the condition (https://github.com/goblint/analyzer/pull/96/commits/83546cd3d6bce4bd3b47bf14ba8c6e9742a9ed39).

vogler:
Maybe it's a good idea to generically do the reduced cross product after every transfer function, such that int abstractions can refine themselves. There may be examples where this leads to an overall better precision.

jerhard:
Yes, we can think about some mutual refinement, but I guess that would be separate issue/pr.

PR #97, opened 2020-09-11
DefExc: Define widen by join.

jerhard:
Previously, `DefExc` used the widening operator as implemented by `Lattice.StdCousot` -- which is not a valid widening operator (as of now), because it just returns the second argument:
https://github.com/goblint/analyzer/blob/0a88b34c8356d49c9889c42926a8532b0c100020/src/domains/lattice.ml#L35-L39

So now we define the `widen` simply by the `join` for `DefExc`. 

More generally speaking, there are 16 includes of `StdCousot` in the code base (not counting those in `arrayDomain_deprecated` and `intDomain_unused`). At some of these places, the `widen` might not be overridden and the resulting behavior might be unsound.

### DIFF
--- a/src/cdomains/arincDomain.ml
+++ b/src/cdomains/arincDomain.ml
@@ -32,7 +32,6 @@ module D =
 struct
   type t = process [@@deriving to_yojson]
   include Printable.Std
-  include Lattice.StdCousot
 
   (* printing *)
   let short w x = Printf.sprintf "{ pid=%s; pri=%s; per=%s; cap=%s; pmo=%s; pre=%s; pred=%s; ctx=%s }" (Pid.short 3 x.pid) (Pri.short 3 x.pri) (Per.short 3 x.per) (Cap.short 3 x.cap) (Pmo.short 3 x.pmo) (PrE.short 3 x.pre) (Pretty.sprint 200 (Pred.pretty () x.pred)) (Ctx.short 50 x.ctx)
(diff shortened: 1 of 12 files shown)


jerhard:
We maybe should consider to simply remove `widen` from `Lattice.StdCousot`, as to be sound, it would either have to rely on a passed `join` operator (given that our domains only have finite ascending chains), or always return `top`.

vogler:
Searching for `S.Dom.widen`, the only place where we don't do `S.Dom.widen old (S.Dom.join old tmp)` in the solver seems to be this:
https://github.com/goblint/analyzer/blob/0a88b34c8356d49c9889c42926a8532b0c100020/src/solvers/sLR.ml#L69
So, as it's used currently, it shouldn't be unsound (except for `slr3` maybe).

We should probably just remove `StdCousot` alltogether. In order to replace widen with `let widen = B.join`, it'd need to be functorized.

jerhard:
I tried to do this [here](https://github.com/goblint/analyzer/pull/97/commits/e5889b29180fe2fd4fd0a08cd3415cb3459557fa), but somehow it makes test [09/03](https://github.com/goblint/analyzer/blob/e5889b29180fe2fd4fd0a08cd3415cb3459557fa/tests/regression/09-regions/03-list2_rc.c) fail, because some races are not detected anymore.

jerhard:
Ok, now that @michael-schwarz fixed it, should we just merge this?

michael-schwarz:
Apparently this behavior was intended and the solvers are written in a way such that they always did a `join`, before calling `widen` (as Ralf mentioned above).
However, in today's Gobcon the consensus was that this should be abandoned and that every domain needs to specify widen in a way s.t. `join a b \leq widen a b`.

PR #98, opened 2020-09-11
Integer Domains: Fix unsound treatment of integer arithmetic

jerhard:
Much of the arithmetic on C integers is not injective in general, due to the possibility of collisions caused by overflows.
The `DefExc` domain makes an assumption of injectivity in some cases without taking this possibility of integer overflows into account.

The issue is in particular about the treatment of operations where one Operand is `Exluded s` and the other is `Definite v`.


jerhard:
#### Example:
Assume, for simplicity, we do arithmetic on 4 bit unsigned integers (i.e. we compute `mod 8`) 
We want to compute the abstract value for:
`z := x * y`
where we know that `x =  Excluded {3}` and `y = Definite 2`.
Now, with the current implementation, we would compute `Excluded {6}` as the abstract value for `z`. 
However, if `x==7`, then `z:=x*y==6` (`mod 8`).

Real world example [here](https://github.com/goblint/analyzer/blob/def_exc_sound_arith/tests/regression/01-cpa/40-arithm.c)

jerhard:
We could either just use `lift2` here instead of `lift2_inj`, but this would unnecessarily lose precision in some cases.
https://github.com/goblint/analyzer/blob/63ccd5a44d317400123280d89439222876eaa8f4/src/cdomains/intDomain.ml#L871-L876

In particular, when we can prove that the abstract values cannot lead to an overflow, we can still assume injectivity of the operation. A check whether an overflow might occur, mit use the range we have for the `Excluded` abstract value, the value of the `Definite` value and the `ikind` the operation is performed on.

jerhard:
Besides operations on an `Definite` and an `Exluded` abstract value, we also have to check whether the operations on two 
`Definite` values produce an overflow, and return `Top` if one occurs. This should also fix the problem for this test case [here](https://github.com/goblint/analyzer/blob/def_exc_sound_arith/tests/regression/27-inv_invariants/03-bot-mul-div.c).

For this test case, Goblint computes `Bottom` for parts of the expression in the if-statement. The reason for this is that due to an integer overflow in the `inv_bin_int` logic in the following snippet, a meet with the wrong value is computed, which results in `Bottom`:
https://github.com/goblint/analyzer/blob/63ccd5a44d317400123280d89439222876eaa8f4/src/analyses/base.ml#L1402-L1416

jerhard:
https://github.com/goblint/analyzer/blob/93450d2a0709989ee80218536841f73483e1e1e5/src/analyses/base.ml#L1400
This treatment of a multiplication in the `invariant` is unsound, unfortunately, as the following example shows:
https://github.com/goblint/analyzer/blob/93450d2a0709989ee80218536841f73483e1e1e5/tests/regression/27-inv_invariants/04-mul-arith.c#L6-L14 
Basically, the issue is that an overflow might have occurred during the multiplication that lead to the value `c`.
I discussed this with @michael-schwarz, and we came to the (preliminary) conclusion that one could possibly treat multiplication for unsigned integers along the following lines:
`a * b == c (mod m)` is equivalent to `a * b == c + k * m`. 
Thus, we could set `a == c / b + (k * m) / b` and `b == c / a + (k * m) / a`. 
Implementing this in Goblint would require some kind of strided intervals, though. 

It's not so clear to me how one would best treat signed integer multiplication in the `invariant`, though. I guess one first would have to check whether the abstract values for `a` and `b` might lead to an overflow and then give up if this is the case, and otherwise do the refinement as it is done now.

jerhard:
To properly deal with overflows in the C code we analyze, and to avoid overflows happening in the implementation of our integer domains, the idea would be to port all integer domains to use arbitrary precision integers internally. Additionally, we would add an `ikind` to the operations in the int domains, so they can do an appropriate overflow handling.

Currently, we have the integer domains `Intervall32`, `CircInterval`, `DefExc` and `Enums`. The question is whether we want to keep all of them? Is there a good reason to have both `DefExc` and `Enums`, and both `Intervall32` and `CircInterval`? Focusing on fewer domains might help with maintenance.

jerhard:
> https://github.com/goblint/analyzer/blob/93450d2a0709989ee80218536841f73483e1e1e5/src/analyses/base.ml#L1400
>This treatment of multiplication in the `invariant` is unsound, unfortunately, as the following example shows:
> https://github.com/goblint/analyzer/blob/93450d2a0709989ee80218536841f73483e1e1e5/tests/regression/27-inv_invariants/04-mul-arith.c#L6-L14

As @michael-schwarz and I just discussed, if one of the operands of the multiplication is a value that is not divisible by two, then we can still do this refinement for the other. This is according to the [lemma of Bezout](https://www.uni-weimar.de/fileadmin/user/fak/medien/professuren/Mediensicherheit/Teaching/WS1819/Diskrete_Strukturen/ds02.pdf).

vogler:
> Merge branch 'def_exc_fix' into def_exc_sound_arith

Is there a reason for this? It's closed and #93 contains the merged changes.
It'd be better to revert and just rebase this branch on master, then you don't include all the commits from the other PR.

jerhard:
>Is there a reason for this?

The idea behind merging the 'def_exc_fix' branch was that there were a couple of places, in particular in `base.ml` where the right `ikind` was already taken into account, that did not make it into the master branch yet.

vogler:
There are probably quite a few conflicts since #89 and #93 diverged. See the discussion in #93.
Also, #93 contains many cherry-picked commits which would result in duplication mess if we merge the original commits (at least if not squashed).
Are the `ikind` changes needed for this PR? If not, I would put them in a separate PR once needed by copying them from the changes https://github.com/goblint/analyzer/compare/pr89_fixes..def_exc_fix

jerhard:
>Are the ikind changes needed for this PR?

So if we want to have non-trivial sound integer domains that are valid for more than one type of integer, the domains have to be aware of the `ikind` of the integer they represent. This is basically due to the possibility of overflows.

The general idea would be to do the following:
1. Make all integer domains aware of the `ikind` of the integer
2. Use BigInt in the implementation of all integer domains
3. In the domains implementations of all operations, check whether the operations may/must overflow, taking the `ikind` into account. In case of signed overflows, go to top. In case of unsigned overflows, compute the correct value, which depends on the `ikind`.

With this, it would really make sense to add an `ikind` to the parameters of the  `of_int` function. Then we would not have the additional calls `cast_to` in `base.ml` introduced in the `def_exc_fix` branch, but instead would pass the `ikind` to `of_int`.
If proceeding like that, it might make sense to rewrite this from scratch instead of merging the commits, but yes, we IMHO still have to introduce code that passes information on the `ikind` to the integer domains in `base.ml`.

vogler:
Sounds good to me. But better rebase + rename this and make a new PR. What's actually in this PR if #89 is subtracted?
Also, we should identify again what the problem is with the current implementation.
As far as I remember regarding overflows the idea was (at least for intervals, the other domains might not all do this) that overflows in the implementation can not happen since it uses 64 bit but we only abstract 32 bit integers and everything bigger collapses into the 32 bit top. So this should be the same as using arb. prec. integers, just with a smaller range.
After an arithmetic operation the result may be out of range, but the overflow is then handled by casting to its range.

Issue #99, opened 2020-09-24
Fix and uncomment failing domain property tests


sim642:
This is a follow-up issue to PR #86. Before merging I commented out all the property testing of all the (integer) domains where something failed, so it doesn't cause CI to massively fail on the master branch but still brings the property testing infrastructure up to date to a usable place.

## Usage
* Compiling: `make domaintest`
* Running: `./goblint.domaintest`.
  See `--help` for other useful flags provided by qcheck, e.g. `-v` or `--long`.

## Current situation
That special standalone binary runs property tests on the domains listed in [`maindomaintest.ml`](https://github.com/goblint/analyzer/blob/08316c5560b1c1298d8b0b29af4b076f12de0eb5/src/maindomaintest.ml#L48-L74). The first list `domains` just tests lattice properties (see `DomainProperties`) and the second list `intDomains` is also intended to additionally check things like whether the integer domain is a valid abstraction of finite sets of integers (see `IntDomainProperties`).

## TODO
As you can see, most of them are currently commented out and possibly even in the wrong list. What needs to be done is commenting them out and seeing what fails and hopefully fixing the issue. For example for `DefExc` 9 lattice properties apparently fail:
```
--- Failure --------------------------------------------------------------------

Test def_exc: join assoc failed (187 shrink steps):

(Not {1}([-63,63]), 0, -1)

--- Failure --------------------------------------------------------------------

Test def_exc: join abs failed (61 shrink steps):

(Not {109}([-63,63]), Unknown int([-63,63]))

--- Failure --------------------------------------------------------------------

Test def_exc: meet leq failed (62 shrink steps):

(Unknown int([-63,63]), Not {-98}([-63,63]))

--- Failure --------------------------------------------------------------------

Test def_exc: meet idem failed (61 shrink steps):

Not {126}([-63,63])

--- Failure --------------------------------------------------------------------

Test def_exc: meet abs failed (61 shrink steps):

(Not {-118}([-63,63]), Unknown int([-63,63]))

--- Failure --------------------------------------------------------------------

Test def_exc: top meet failed (60 shrink steps):

Not {118}([-63,63])

--- Failure --------------------------------------------------------------------

Test def_exc: connect join failed (33 shrink steps):

(7144723426, Unknown int([-63,63]))

--- Failure --------------------------------------------------------------------

Test def_exc: connect meet failed (57 shrink steps):

(Unknown int([-63,63]), Not {100}([-63,63]))

--- Failure --------------------------------------------------------------------

[...]

PR #100, opened 2020-09-24
Def_exc: Fix issues in leq, join, narrow, meet with broken checking for in_range

michael-schwarz:
This fixes some of the issues quickcheck (see #99) found for `def_exc`.

- `leq`: between `Definite` and `Excluded` did not take the ranges into account
- `join`: was too eager to increase the range.
- `meet`: checked whether the excluded value on the right was in the range interval instead of checking if it is in the range represented by that interval

It also fixes an issue where `narrow` was changed by mistake in #97 causing a time-out on some of the regression tests.
It now runs the quickchecks but skips the test for associativity of join, as our join is not associative.

###  The join i still not associative

**Because of the ranges (fixed now)**
```
a: Unknown int([-63,63]) 
b: 0
c: 7317727852

join(a,b) : Unknown int([-63,63]) 
join(join(a,b), c): Unknown int([-63,63])

join(b,c): Unknown int([0,64])
join(join(b,c), a):  Unknown int([-63,64])
```

This problem comes from having to chose an exclusion range for a join of two definites. Whichever way one does this (so `long` vs. `unsigned long` here), this problem will crop up.
I guess one could go to the non-C range of `[0,n-1]` for values that fit both into the signed and unsigned type of size n.

**And more fundamentally**
```
a: Not {1}([-63,63]) 
b: 0
c: -1

join(a,b): Not {1}([-63,63]) 
join(join(a,b),c): Not {1}([-63,63])

join(b,c): Unknown int([-7,7]) 
join(a, join(b,c)): Unknown int([-63,63])
```

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -102,6 +102,25 @@ module Size = struct (* size in bits as int, range as int64 *)
     in
     M.tracel "cast_int" "Cast %Li to range [%s, %s] (%s) = %s (%s in int64)\n" x (string_of_big_int a) (string_of_big_int b) (string_of_big_int c) (string_of_big_int y) (if is_int64_big_int y then "fits" else "does not fit");
     try int64_of_big_int y with _ -> raise Not_in_int64
+
+  let min_range_sign_agnostic x =
+    let size ik =
+      let a,b = bits_i64 ik in
+      Int64.neg a,b
+    in
+    if sign x = `Signed then
+      size (min_for x)
+    else
+      let a, b = size (min_for x) in
+      if b <= 64L then
+        let upper_bound_less = Int64.sub b 1L in
+        let max_one_less = Int64.(pred @@ shift_left 1L (to_int upper_bound_less)) in
+        if x < max_one_less then
+          a, upper_bound_less
+        else
+          a,b
+      else
+        a, b
 end
 
 exception Unknown
@@ -734,15 +753,15 @@ struct
       if x = y then `Definite x
       (* Unless one of them is zero, we can exclude it: *)
       else
-        let a,b = Size.(min_for x, min_for y) in
-        let r = R.join (size a) (size b) in
+        let a,b = Size.min_range_sign_agnostic x, Size.min_range_sign_agnostic y in
+        let r = R.join (R.of_interval a) (R.of_interval b) in
         `Excluded ((if x = 0L || y = 0L then S.empty () else S.singleton 0L), r)
     (* A known value and an exclusion set... the definite value should no
      * longer be excluded: *)
     | `Excluded (s,r), `Definite x
     | `Definite x, `Excluded (s,r) ->
       if not (in_range r x) then
-        let a = size (Size.min_for x) in
+        let a = R.of_interval (Size.min_range_sign_agnostic x) in
         `Excluded (S.remove x s, R.join a r)
       else
         `Excluded (S.remove x s, r)
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -788,7 +788,8 @@ struct
       let s' = S.union x y |> S.filter (in_range r') in
       `Excluded (s', r')
 
-  let narrow = meet
+  let narrow x y = x
+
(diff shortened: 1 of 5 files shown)


vesalvojdani:
These exclusion sets have been explored in this [SAS paper](https://jorgenavas.github.io/papers/nonlattice-sas13.pdf). In particular, Example 6 "any minimal quasi-join for the donut domain over intervals is non-associative". I haven't taken a very deep look at what remedies they suggest, but I think a deep read of that paper may be useful.

michael-schwarz:
To recap some of the conclusion Vesal, Kalmer, Julian and I reached at today's Gobcon:

**It seems that `Def_Exc` as it is currently implemented (on top of not being a lattice) is also not a quasi-lattice as defined in the SAS '13 paper.**
For the quasi-lattice, they require the `join` to give a minimal upper bound on two values (but not necessarily a least upper bound, i.e. there may exists several minimal upper bounds that are incomparable).

For `Def_Exc`, this is violated for the join  `Definite a` `Definite b` values as we in this case go to `Excluded {} [r]` for some range R, but there exists a smaller element in the domain namely `Excluded { r | r in R, r !=a, r != b} [r]`. However, excluding everything in the range except `a` and `b` is the same as keeping a definite set of values, so this is definitely undesirable.

In order to make it at least into a quasi-lattice, it would be possible to limit the number of excluded values. But then these would be chosen randomly, so this is probably not desirable either.

**The interesting questions is if we leave `Def_Exc` as it is (and accept that is neither lattice nor quasi-lattice), can this cause solving to not terminate or some other weird issue?**

vesalvojdani:
Maybe the safest approach is to limit the content of these sets to only zero and constant values that occur in case statements (or all conditional guards). If we have a fixed finite subset of select values V after pre-processing, the least upper bound of definite values is expanded to all other values `a ⊔ b = V \ {a,b}`. This should be safe, but it can impact performance. We have to normalize the sets after operations, so one can't preserve `(x != 0)` after  `x++; x--` unless 1 is in V. I don't think it's that's a serious drawback, but I also don't know if non-termination is caused by this for the solvers we use. It would be good to know more certainly if this can cause non-termination for the TD3 solver.

michael-schwarz:
It seems like the cases from above where analysis does not terminate are unrelated to the arrays, we have the same behavior for this simple example as well:

https://github.com/goblint/analyzer/blob/037ac0d8eaa342a4275376841dc8f8c4a769b54f/tests/regression/02-base/33-backwards-loop.c#L1-L10

michael-schwarz:
Changing `starting x` and `ending x` to not exclude 0 leads to `2 tests failed: ["04/36 trylock_nr", "04/42 trylock_2mutex"]`.
Both of those work again if we enable intervals for them.

vesalvojdani:
A naive question: why does narrowing for exclusions sets always use the meet operation? That seems like an interval narrowing always improving the result. Maybe just meet if exclusion set is empty, but do not grow it if it contains any elements already. Do we actually need exclusion sets to be refined?

michael-schwarz:
> A naive question: why does narrowing for exclusions sets always use the meet operation? That seems like an interval narrowing always improving the result. Maybe just meet (use refined value) if exclusion set is empty, but do not grow it if it contains any elements already. Do we actually need exclusion sets to be refined?
> 

Not so much a naive question as a good one! Looks like we replaced `narrow x y = x` by `narrow = meet` when getting rid of StdCousot in https://github.com/goblint/analyzer/commit/e5889b29180fe2fd4fd0a08cd3415cb3459557fa.
This is obviously not a smart thing to do.

> [Edited] I forgot my main question. Simmo had a similar case in his path-sensitive domain where backward looping failed, but forward iteration worked. Are we handling negative numbers differently, or why is this only triggered for `x--`?

No, it does not have anything to do with negative numbers or iterating backwards per se, we have the same behavior in all of these cases here. It's just that the iterating backwards case is the one that appears most natural out of all of those.

https://github.com/goblint/analyzer/blob/d74e95e89d3e21b003572607c54d526a4b60ca3d/tests/regression/02-base/33-meet-def-exc.c

michael-schwarz:
I copied the parts that will still be relevant after merging this into #105

PR #108, opened 2020-10-08
Remove indented, compact, and (old) html output

### LINKED ISSUE #20: Remove toXML_f and toXML
For XML output `printXML` is used instead of `toXML` because `toXML` was too slow/memory intensive for large outputs.
The problem with removing it, is that it is used for `result` options `indented`, `compact`, `html`.

michael-schwarz:
This removes the `to_XML` and `to_XML_f` functions (see #20).

The output types `indented`, `compact`, and `html` are removed by this, the replacements are `fast_xml` or the `g2html` output via `fast_xml` respectively. In case anyone has any scripts that use any of these three old output formats, we output a message stating they were removed making it easy to understand what happened.

Having two ways to produce XML and two (three including the JSOO one https://github.com/michael-schwarz/goblint-webfrontend-jsoo) ways to get HTML output is confusing, so I'd argue that getting rid of  `indented`, `compact`, and `html` is a step towards cleaner code.

Last time there was talk about removing them was in 2015, hopefully the situation has since changed.

Closes #20.


michael-schwarz:
I also corrected some typos my IDE highlighted while doing this in the same commit.

vogler:
[src/util/xmldump.ml](https://github.com/goblint/analyzer/blob/remove_old_xml/src/util/xmldump.ml) is also dead now.

vogler:
I also removed `Analyses.RS` in https://github.com/goblint/analyzer/commit/a56c407bfdc49d841aa564c4e64e774c541a069c.
The only place that still needs the `xml-light` package (search for `Xml.`) is `src/witness/graphml.ml`.
Not that important but maybe it makes sense to use a more up-to-date package or get rid of the dependency? @sim642

sim642:
`GraphMlWriter` has intentionally been written in imperative style, so I should be able to pretty easily get rid of that dependency as well. It would probably speed up writing of larger witnesses anyway, so no reason not to.

vogler:
> `GraphMlWriter` has intentionally been written in imperative style, so I should be able to pretty easily get rid of that dependency as well. It would probably speed up writing of larger witnesses anyway, so no reason not to.

Can't see `xml-light` anymore, so this is done I guess 🎉

sim642:
Yes, this happened soon after this PR in https://github.com/goblint/analyzer/commit/3a61dda041a5543fc8a82a9078bd63f962254d04.

Issue #102, opened 2020-09-25
Submit verification tasks to SV-COMP's NoDataRace category


sim642:
The deadline for submitting verification tasks to SV-COMP 2021 is **9th October**.

I was looking at extracting data-race based test cases from our regression suite and placed them to [my fork of sv-benchmarks](https://github.com/sim642/sv-benchmarks/tree/goblint-regression). I took all non-skip, non-kernel, non-osek regression tests which contain `RACE` or `NORACE` and constructed SV-COMP task definition files (.yml) with the corresponding expected verdict for the new no-data-race property.

Since sv-benchmarks requires everything to compile (preferably without warnings), I had to make some small fixes here and there. Notably I had to replace the use of uninitialized ints for non-determinism (which causes compiler warnings) with `__VERIFIER_nondet_int()`.

## TODO
- [x] Are there any more of the data-race regression benchmarks that should be excluded from submission?
- [x] Are there any more of data-race benchmarks (not just regression) that should be included in the submission? I'm not really sure what https://github.com/goblint/bench has to offer.
- [x] [sv-benchmarks requires all submitted tasks to be accompanied with a license](https://github.com/sosy-lab/sv-benchmarks/blob/master/CONTRIBUTING.md#license). Since the regression tests are from this repo, I suppose [this LICENSE](https://github.com/goblint/analyzer/blob/master/LICENSE) should work. Or is there anything in the regression tests that comes from another source and should be excluded? I'm not familiar with the history of these.

sim642:
While replacing uninitialized local int variables with `__VERIFIER_nondet_int()` is straightforward, there were a couple of uninitialized local pointer variables, where I wasn't so sure. There is also `__VERIFIER_nondet_pointer()` which I currently went with but that may not have been the right thing to do.

### 04-mutex/31-uninitialized
This one does `pthread_mutex_lock` on an uninitialized mutex pointer, which I assume is intentional by the test name.
https://github.com/goblint/analyzer/blob/0610c520ca3fdda91dbc365f44a5e78c76906c6f/tests/regression/04-mutex/31-uninitialized.c#L7-L11
But I'm skeptical about submitting this test because wouldn't the locking of an uninitialized mutex be undefined behavior in the first place? I think SV-COMP doesn't want any undefined behavior.

### 09-regions/05-ptra_rc
This one uses an uninitialized struct pointer `sp` in different ways.
[...]

PR #109, opened 2020-10-08
Reviewed data race regression tests for SV-COMP submission

sim642:
In the regression tests suitable for #102, I did the following:
* Consistently used `RACE!` for actual races.
* Consistently used `NORACE` for the lack of races (as opposed to `NORACE!`, `NOWARN` or `NOWARN!`).
* Consistently used `RACE` for races Goblint finds due to imprecision which aren't actually races.
* Replaces uninitialized variables for nondeterminism with an extern `__VERIFIER_nondet_*`.
* (Includes automatic trailing whitespace deletion in the files I happened to touch.)

I only looked at data race tests which are not based on assertions and which make sense as independent programs (so ignored kernel ones and a few others).

Having these in our repo is good for consistency and also makes it easy to automatically convert them to sv-benchmarks.

### DIFF
--- a/tests/regression/04-mutex/03-munge_rc.c
+++ b/tests/regression/04-mutex/03-munge_rc.c
@@ -12,7 +12,7 @@ void munge(pthread_mutex_t *m) {
 }
 
 void *t_fun(void *arg) {
-  munge(&mutex2); // NOWARN
+  munge(&mutex2);
   return NULL;
 }
 
@@ -20,7 +20,7 @@ void *t_fun(void *arg) {
 int main(void) {
   pthread_t id;
   pthread_create(&id, NULL, t_fun, NULL);
-  munge(&mutex1); // NOWARN
+  munge(&mutex1);
   pthread_join (id, NULL);
   return 0;
 }
--- a/tests/regression/04-mutex/03-munge_rc.c
+++ b/tests/regression/04-mutex/03-munge_rc.c
@@ -12,7 +12,7 @@ void munge(pthread_mutex_t *m) {
 }
 
 void *t_fun(void *arg) {
-  munge(&mutex2); // NOWARN!
+  munge(&mutex2); // NOWARN
   return NULL;
 }
 
@@ -20,7 +20,7 @@ void *t_fun(void *arg) {
 int main(void) {
   pthread_t id;
   pthread_create(&id, NULL, t_fun, NULL);
-  munge(&mutex1); // NOWARN!
+  munge(&mutex1); // NOWARN
   pthread_join (id, NULL);
   return 0;
 }
--- a/tests/regression/04-mutex/35-trylock_rc.c
+++ b/tests/regression/04-mutex/35-trylock_rc.c
@@ -49,7 +49,7 @@ void *monitor_thread (void *arg) {
     if (status != EBUSY) {
       if (status != 0)
         err_abort (status, "Trylock mutex");
-      printf ("Counter is %ld\n", counter/SPIN); // RACE!
+      printf ("Counter is %ld\n", counter/SPIN); // RACE
       status = pthread_mutex_unlock (&mutex);
       if (status != 0)
         err_abort (status, "Unlock mutex");
--- a/tests/regression/04-mutex/41-pt_rwlock.c
+++ b/tests/regression/04-mutex/41-pt_rwlock.c
@@ -8,8 +8,8 @@ static pthread_rwlock_t rwlock = PTHREAD_RWLOCK_INITIALIZER;
 
 void *t_fun(void *arg) {
   pthread_rwlock_wrlock(&rwlock);
-  data1++;            //NORACE!
-  printf("%d",data2); //RACE!
+  data1++;            // NORACE
+  printf("%d",data2); // RACE
   pthread_rwlock_unlock(&rwlock);
   return NULL;
 }
@@ -19,8 +19,8 @@ int main(void) {
   pthread_create(&id, NULL, t_fun, NULL);
 
   pthread_rwlock_rdlock(&rwlock);
-  printf("%d",data1); //NORACE!
-  data2++;            //RACE
+  printf("%d",data1); // NORACE
+  data2++;            // RACE
   pthread_rwlock_unlock(&rwlock);
 
   pthread_join (id, NULL);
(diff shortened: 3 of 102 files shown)


tests/regression/04-mutex/03-munge_rc.c
```diff
@@ -12,15 +12,15 @@ void munge(pthread_mutex_t *m) {
 }
 
 void *t_fun(void *arg) {
-  munge(&mutex2); // NOWARN!
+  munge(&mutex2); // NOWARN
```
michael-schwarz, line 15: Why `NOWARN` here instead of `NORACE`?

michael-schwarz: There are a couple of places like this.

tests/regression/09-regions/26-alloc_region_rc.c
```diff
@@ -49,12 +49,13 @@ int main () {
   pthread_mutex_lock(&A_mutex);
   insert(p, &A);
   pthread_mutex_unlock(&A_mutex);
-  
+
   pthread_mutex_lock(&A_mutex);
   p = A;
-  while (p->next)
+  while (p->next) {
```
michael-schwarz, line 55: This changes the semantics of the test program. Is this on purpose?

tests/regression/09-regions/23-evilcollapse_rc.c
```diff
@@ -40,15 +42,15 @@ static inline void list_add(struct list_head *new, struct list_head *head) {
 }
 
 inline static struct list_head *lookup (int d) {
-  int hvalue;
+  int hvalue = __VERIFIER_nondet_int();
   struct list_head *p;
   p = c.slot[hvalue].next;
```
michael-schwarz, line 47: I guess if you are strict about it, such things are possibly undefined behavior, as `hvalue` might go beyond the array.
Now, we don't care about it, but I don't know if the SV-Comp folks are also so relaxed about UB.

michael-schwarz, line 47: I guess if you are strict about it, such things are possibly undefined behavior, as `hvalue` might go beyond the array.

We don't care about it, but I don't know if the SV-Comp folks are also so relaxed about UB.

tests/regression/04-mutex/35-trylock_rc.c
```diff
@@ -49,7 +49,7 @@ void *monitor_thread (void *arg) {
     if (status != EBUSY) {
       if (status != 0)
         err_abort (status, "Trylock mutex");
-      printf ("Counter is %ld\n", counter/SPIN); // RACE!
+      printf ("Counter is %ld\n", counter/SPIN); // RACE
```
michael-schwarz, line 52: I don't really understand why this is not an actual race. Maybe worth adding a comment.

tests/regression/04-mutex/41-pt_rwlock.c
```diff
@@ -19,8 +19,8 @@ int main(void) {
   pthread_create(&id, NULL, t_fun, NULL);
 
   pthread_rwlock_rdlock(&rwlock);
-  printf("%d",data1); //NORACE!
-  data2++;            //RACE
+  printf("%d",data1); // NORACE
+  data2++;            // RACE
```
michael-schwarz, line 23: Maybe it would be interesting to add a version where a second thread also does 

```
  pthread_rwlock_rdlock(&rwlock);
  printf("%d",data1); // NORACE
  data2++;            // RACE

```

Because the it would be `RACE!`

tests/regression/04-mutex/03-munge_rc.c
sim642: I left them like that because those are slightly unrelated to the question of races and comparing the `rc`/`nr` versions it's not where a race is happening anyway. I didn't want to remove them either because supposedly there's something we want to check, but that's more of a user experience question. Like, why do we insist on checking that Goblint doesn't tell the user where the racing function is called from.

tests/regression/09-regions/26-alloc_region_rc.c
sim642: This is on purpose because the original program was actually race-free, although for very complicated reasons (for Goblint). This loop in the main thread would've always moved to the last element of the singly linked list to access its data. The other thread would've modified the data of `p` which it prepended to the same singly linked list, but `p` could never be the last element of that list. We decided keeping that original program in our test suite would be worthless because we don't intend for Goblint to be able to do that complicated reasoning.

The first modification of this program also revealed what seems to be a different bug, so there's a skipped 09/33 test.

tests/regression/09-regions/23-evilcollapse_rc.c
sim642: I noticed it in a bunch of places too. I suppose I'll have to add some assertion/assumption after it, keeping only a valid range.

tests/regression/04-mutex/35-trylock_rc.c
sim642: If `pthread_mutex_trylock` above succeeds to lock, then that access of `counter` is protected by `mutex`. The access in the else branch where locking failed is the racy one. Comparing to 04/36 that's where the difference comes from.

tests/regression/04-mutex/41-pt_rwlock.c
sim642: I will add a few variations on this for submission but I think we aren't planning to add proper read-write lock support which would be able to rule this one out.

There is some slightly different handling of write and read lock locking currently implemented but that is not exactly it. It seems to consider _any_ write in a read lock section potentially unsafe. That is why the upper `NORACE` lines are correctly excluded because the writing happens in the thread with a write lock.

tests/regression/09-regions/23-evilcollapse_rc.c
sim642: I just committed a fix to this as well using `assume_abort_if_not` defined through `abort` which is how the rest of sv-benchmarks handle such thing.

tests/regression/04-mutex/03-munge_rc.c
vesalvojdani: No, there is nothing we want to check there. The 04-mutex category is probably the oldest regression test category and my understanding then was to very narrowly follow goblint's current behavior. It would be much better if the tests make sense independently of what goblint does (except TODO annotations).

michael-schwarz:
Looks like it's good to merge now!

vesalvojdani:
This can happen after merging, but some of the list tests should be fixed to work with arbitrary length lists to make sure it is some sort of heap abstraction being used and not just points-to information. Also, for some of these tests, we should also use an arbitrary number of threads. I would like to enable the uniqueness analysis, so it would be good if there were some tests that it does not wrongly consider things unique.

PR #112, opened 2020-10-13
Semantic Search

superbr4in:
This feature introduces a search provider for semantic code properties, based on Goblint's analysis results. For this purpose, a new transformation named `expeval` uses the built-in query system to evaluate specified arithmetic expressions (without actually transforming source code).

My corresponding bachelor thesis: _Extending a Syntactic Search in C Source Code with Analysis Results from the Static Analyzer Goblint_


michael-schwarz:
Thanks for the PR!
We'll delay merging this a bit until we have cut a release for the pre-run of SVCOMP 2020.

dune-project
```diff
@@ -22,7 +22,6 @@
     (ocaml (>= 4.04.1)) ; TODO check with travis env matrix as for goblint-cil
     dune
     zarith_stubs_js ; only needed for js_of_ocaml
-    goblint-cil ; TODO no way to define as pin-depends? Used goblint.opam.template to add it for now. https://github.com/ocaml/dune/issues/3231. Alternatively, removing this line and adding cil as a git submodule and `(vendored_dirs cil)` as ./dune also works. This way, no more need to reinstall the pinned cil opam package on changes. However, then cil is cleaned and has to be rebuild together with goblint.
```
vogler, line 25: This probably shouldn't be removed.

goblint.opam
```diff
@@ -45,6 +45,3 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/goblint/analyzer.git"
-pin-depends: [
```
vogler, line 48: This probably shouldn't be removed. Generated via `goblint.opam.template`.

goblint.opam.template
```diff
@@ -1,3 +0,0 @@
-pin-depends: [
```
vogler, line 1: This probably shouldn't be removed. We want the pinned version since we don't release to opam regularly.

vogler (changes requested):

michael-schwarz:
Before we can merge this, we need to decide what to do with: https://github.com/faddeenkov/syntactical_analyzer (after some more cleanup of that which I'm currently doing).
This project is the base for this feature, it does the syntactical part of the search.

I see three ways to do this:
- Merge https://github.com/faddeenkov/syntactical_analyzer into our CIL fork, as it is not Goblint specific
- Merge https://github.com/faddeenkov/syntactical_analyzer into Goblint together with this
- Keeping https://github.com/faddeenkov/syntactical_analyzer as a submodule as proposed in this PR

I'd favor either 1 or 2, I think there is no real reason to have this as a submodule going forward.

vogler:
https://github.com/faddeenkov/syntactical_analyzer does not exist. You mean https://github.com/faddeenkov/cil_fork?
By 2 you mean copy the CIL fork into Goblint? Depends how actively you want to change CIL going forward. What are the disadvantages of keeping them separate? You can have multiple opam packages in one repo, but it will increase build time and might make issues/PRs/search confusing unless you tag everything.
By 3 you also mean to merge it into our CIL fork and just use it as a submodule? Then the only difference between 1 and 3 is that 1 needs git push & opam update (& pin add/install?) to get CIL changes into goblint whereas 3 will just take the files but also slightly increase build times if CIL wasn't touched. However, for case 1 you can also just `opam pin` the local directory to speed it up a bit.

sim642:
I'm guessing https://github.com/faddeenkov/syntactical_analyzer is a private repo, which is why it appears to us as non-existent. If that's the case, then it's hard to decide what to do with it without knowing what's there.

michael-schwarz:
@sim642 is right, it is a private repository. I also can't change the visibility of my fork, so I'll have to ask my student to publish it. I invited both of you to my fork for now.

The question is whether to
- merge the code in that separate repo into CIL
- merge it into Goblint
- make it a submodule of Goblint

I don't want think we need to change anything about the way we include CIL.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "g2html"]
 	path = g2html
 	url = https://github.com/goblint/g2html.git
+[submodule "syntactical_analyzer"]
```
michael-schwarz, line 7: This is where we include Olga's code as a submodule for the moment.

michael-schwarz:
https://github.com/faddeenkov/syntactical_analyzer should be public now.

vogler:
No strong opinion. Doesn't look like it's a lot. Probably depends on whether we want it as a separate binary or if it can be sensibly integrated into the goblint CLI.

michael-schwarz:
I merged Olga's syntactic search into CIL as a new feature for now: https://github.com/goblint/cil/tree/feature/syntactic_search

@superbr4in can you adapt this PR to pin goblint-cil to that branch and remove the submodule and see if it works and what changes are required?

Once that works, we can merge https://github.com/goblint/cil/tree/feature/syntactic_search into the `develop` branch of `goblint-cil`, change the pin back and then merge this.

superbr4in:
It looks like the `JsonParser` module of _goblint-cil.syntacticsearch_ collides with another module of the same name. Example usages can be found in [gobConfig.ml](https://github.com/goblint/analyzer/blob/master/src/util/gobConfig.ml#L321) and [defaults.ml](https://github.com/goblint/analyzer/blob/master/src/util/defaults.ml#L281).

How could this be resolved?

vogler:
Could pull it out into a separate library. But it's maybe better to use yojson which we have as a dependency anyway for ppx_deriving_yojson.

michael-schwarz:
Alternatively, a solution would be renaming the module in the CIL submodule to something like QueryParser. 

JsonParser is a bit generic for what it actually does (define the needed types and call the functions generated by `ppx_deriving_yojson`) anyway.

https://github.com/goblint/cil/blob/feature/syntactic_search/src/ext/syntaticsearch/jsonParser.ml

superbr4in:
Alright, it works now. The `pin-depends` are back again.

src/transform/expressionEvaluation.ml
```diff
@@ -0,0 +1,199 @@
...
+                  if is_debug () then print_endline "| (Unreachable)";
+                  Some false
+                (* Valid location *)
+              | Some value_before ->
+                  (* Use the last listed matching statement; TODO: use Hashtbl.find_all to consider other matching statements *)
+                  let _, statement = Hashtbl.find statements location in
+                  (* Use the first evaluable successor; TODO: consider other successors *)
+                  let succeeding_statement = ref None in
+                  let successor_evaluation =
+                    List.find_map_opt
```
michael-schwarz, line 85: This seems to be a recent addition `OCaml >= 4.10.0`, but we aim to remain compatible with `4.07.1` (but only added checks for this to the CI in 46602ca1b696b138c6a7ae2030cd7749ecd0c453, this is why this PR still passes).

Can you make this compatible with `4.07.1`?

michael-schwarz:
I think as soon as this is made compatible with `OCaml 4.07.1`, we're good to merge!
Meanwhile, I merged the syntactic search into CIL and cut a new release (`goblint-cil-1.7.8`).

src/transform/expressionEvaluation.ml
superbr4in: The `4.07.1` checks seem to pass somehow, but I can't explain why.

Issue #120, opened 2020-10-21
Suspected Unsoundness in SV-COMP SoftwareSystems


sim642:
In my own run of Goblint on SV-COMP SoftwareSystems category, Goblint was unsound (`reach_error()` being unreachable: expected `false`, actual `true`) on the following benchmarks:

- [x] ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--staging--lustre--lustre--llite--llite_lloop.ko-entry_point.cil.out.i

sim642:
I looked at SV-COMP 2020 results to see if any other verifier has given a violation witness for this benchmark but all of them either timed out or gave an unknown result.

Oddly, the correctness witness Goblint produces is just a single entry node. Somehow region analysis manages to break witness generation because disabling region makes the witness be outputted in whole. While it's something I need to look into, it's not the real issue because in HTML output, `ldv_error` is dead code either way.

sim642:
After hunting around that benchmark I found the following:

https://github.com/sosy-lab/sv-benchmarks/blob/ac7cce1b1da7546521309eb9bb417ee994fe9439/c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--staging--lustre--lustre--llite--llite_lloop.ko-entry_point.cil.out.i#L9645-L9655:

```c
__inline static void *kmalloc_array(size_t n , size_t size , gfp_t flags )
{
  void *tmp ;
  {
  if (size != 0UL && 0xffffffffffffffffUL / size < n) {
    return ((void *)0);
  } else {
  }
  tmp = __kmalloc(n * size, flags);
  return (tmp);
}
```

This gets called with `n = 16`, `size = 4912` and according to Goblint, this always returns 0:
![Screenshot_2020-10-21_15-41-56](https://user-images.githubusercontent.com/378740/96720999-31793480-13b4-11eb-9955-c46d010b1708.png)
But with these values, the condition should actually be false. This benchmark should follow the LP64 data model and the typedefs in it say `size_t` is `unsigned long`.

The following regression test should model this issue:
```c
// PARAM: --enable ana.int.interval --enable ana.int.def_exc
#include <assert.h>

int main() {
    // should be LP64
    unsigned long n = 16;
    unsigned long size = 4912;

    assert(!(0xffffffffffffffffUL / size < n));
}
```

sim642:
[...]

PR #121, opened 2020-10-21
Add proper handling of integer-overflows

jerhard:
In order to implement proper handling of integer overflows in the analyzed C code, and to be able to deal with values outside the range of OCaml-`int64` (#120), we have to make several changes to the integer domains.

1. Add `ikind` information to the integer domains.
2. Use arbitrary precision integers in the implementation of the domains.
3. In the implementation of the abstract operations of the integer domains, add detection & handling of overflows.

Regarding 1., I added an functor `IntDomLifter`, that takes a module of signature `IntDomain.S` and adds a layer to store the ikind. 

TODOs:
- it is not clear how the `top` and `bot` should look like in the `IntDomLifter`. These either would have to be of some arbitrary `ikind` or some additional `Top` and `Bottom` element 
- 2 and 3. still have to be implemented


vogler:
Instead of copying `module type S` to `module type Z`, it'd be better to just include the common values.

vogler:
Instead of carrying `ikind` around we could add it as an argument for operations. Adding an indirection and additional 64 bit for `ikind` for every abstract int might be costly.
Since a lot of bot/top handling is shared, it's maybe a good idea anyway to merge operations into an `unop`/`binop` which gets the operation and `ikind`.
If we still want to keep the specialized functions like `add`, `sub` etc., they could be added by some functor.

src/cdomains/valueDomain.ml
```diff
@@ -288,7 +288,7 @@ struct
         | TInt (ik,_) ->
           `Int (ID.cast_to ?torg ik (match v with
               | `Int x -> x
-              | `Address x when AD.equal x AD.null_ptr -> ID.cast_to ?torg (ptr_ikind ()) @@ ID.of_int Int64.zero
+              | `Address x when AD.equal x AD.null_ptr -> ID.cast_to ?torg (ptr_ikind ()) @@ ID.of_int (Cilfacade.ptrdiff_ikind ()) Int64.zero
```
michael-schwarz, line 291: What is the reason behind making a `0` of `Cilfacade.ptrdiff_ikind ()` and then casting it to `ptr_ikind ()`?

src/cdomains/valueDomain.ml
```diff
@@ -389,8 +389,9 @@ struct
         | _ -> AD.join y AD.top_ptr)
     | (`Address x, `Address y) -> `Address (AD.join x y)
     | (`Struct x, `Struct y) -> `Struct (Structs.join x y)
-    | (`Union x, `Union y) -> `Union (Unions.join x y)
-    | (`Array x, `Array y) -> `Array (CArrays.join x y)
+    | (`Union (f,x), `Union (g,y)) -> `Union (match UnionDomain.Field.join f g with
+        | `Lifted f -> (`Lifted f, join x y) (* f = g *)
+        | x -> (x, `Top)) (* f <> g *)    | (`Array x, `Array y) -> `Array (CArrays.join x y)
```
michael-schwarz, line 394: Move `| (Array x, Array y) -> Array (CArrays.join x y)` down one line again?

jerhard:
>Instead of carrying ikind around we could add it as an argument for operations.
I think there are places besides the arithmetic operations where we want to know the `ikind`. E.g. if we perform a `widen` on ints, it would be useful to know what the `ikind` of the integers is that is being joined.

vogler:
> > Instead of carrying ikind around we could add it as an argument for operations.
> 
> I think there are places besides the arithmetic operations where we want to know the `ikind`. E.g. if we perform a `widen` on ints, it would be useful to know what the `ikind` of the integers is that is being widened.

Might be worthwhile to think about whether this would actually make a difference.
After `widen` an interval might become bigger than the range of its type, but this can then be recovered by first normalizing arguments in any operation. The only case where it could surface is widened values that are not operated on afterwards. However, those we could normalize to their type's range after the solver is done.

src/analyses/base.ml.orig
```diff
@@ -2172,32 +2176,31 @@ struct
...
-                                  (eval_lv ctx.ask gs st lv, `Address (AD.from_var_offset (heap_var, `Index (IdxDom.of_int (Cilfacade.ptrdiff_ikind ()) BI.zero, `NoOffset))))]
-=======
           let heap_var = heap_var ctx in
           let add_null addr =
             if get_bool "exp.malloc.fail"
             then AD.join addr AD.null_ptr (* calloc can fail and return NULL *)
             else addr in
-          set_many ctx.ask gs st [(add_null (AD.from_var heap_var), `Array (CArrays.make (IdxDom.of_int Int64.one) (`Blob (VD.bot (), eval_int ctx.ask gs st size)))); (* TODO why? should be zero-initialized *)
-                                  (eval_lv ctx.ask gs st lv, `Address (add_null (AD.from_var_offset (heap_var, `Index (IdxDom.of_int 0L, `NoOffset)))))]
+<<<<<<< HEAD
```
michael-schwarz, line 2191: this looks a bit like a merge conflict ;)

jerhard: Ups. I removed the `.orig` files now.

michael-schwarz:
This causes `sv-comp/sv-benchmarks/c/loops/linear_search.c` to unsoundly claim that the error is not reached.

michael-schwarz:
I pushed a problematic example where the ikinds don't line up:

~~~C
    signed char* s = malloc(10*sizeof(signed char));
    s[0] = -5;

    unsigned char* us = s;
    *us = 12;
~~~

This is valid C, and GCC also doesn't complain, but we currently fail with `- Fatal error: exception Failure("ikinds signed char and unsigned char are incompatible. Values: (-5) and (12)")`

This is this issue we discussed where it would be necessary to go to `supertop` for `Int` of different `ikind`.

Issue #118, opened 2020-10-20
Region analysis type mismatch failures in SV-COMP SoftwareSystems


sim642:
In my own run of Goblint on SV-COMP SoftwareSystems category, Goblint crashed with

```
exception Failure("Type mismatch!")
```

https://github.com/goblint/analyzer/blob/3164939e27b7908839fe083f036ca165eb54d7ea/src/cdomains/lval.ml#L508-L514

from region analysis domain (which uses `PartitionDomain` and `collapse` functions) on the following benchmarks:

- [ ] ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--arcnet--rfc1201.ko-entry_point.cil.out.i
- [ ] ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--ethernet--micrel--ks8842.ko-entry_point.cil.out.i
- [ ] ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wireless--ath--ath6kl--ath6kl_usb.ko-entry_point.cil.out.i
- [ ] ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wireless--libertas--usb8xxx.ko-entry_point.cil.out.i
- [ ] ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wireless--mwifiex--mwifiex_usb.ko-entry_point.cil.out.i
- [ ] ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wireless--rtl818x--rtl8180--rtl818x_pci.ko-entry_point.cil.out.i
- [ ] ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-32_7a-drivers--misc--sgi-gru--gru.ko-entry_point.cil.out.i
- [ ] ldv-linux-4.0-rc1-mav/linux-4.0-rc1---drivers--misc--sgi-gru--gru.ko.cil.i

sim642:
[...]

PR #124, opened 2020-11-03
Replace Vars component and sync with direct sideg in region analysis

sim642:
See https://github.com/goblint/analyzer/issues/118#issuecomment-720569170.

Also adds very basic `sideg` support to global initializers.


vesalvojdani:
Ok, let's merge this. I was not too happy about it because this means region collapses can never be improved, even in single-threaded mode. But then, if we want a more serious heap analysis, we should do that separately and let this one be as simple as possible.

PR #125, opened 2020-11-03
Remove unused must-equals component from region analysis

sim642:
The transfer functions did update it but other than that nothing read it. Thus nothing really depended on it either. No point in wasting CPU cycles for something completely irrelevant.

This is something I noticed a while ago but now that we're already simplifying region analysis, it might make sense to also do this. In the git history I have a commit where I also replaced the `Equ` component with `Unit` to easily prove a point that nothing used it.

I'm not sure what the original intention of the must-equals component might've been. Maybe it should be somehow used, then there's a bug that it isn't. Maybe the use of `region` in combination with `symb_locks` and `var_eq` automagically covers the originally required behavior.

### DIFF
--- a/src/analyses/region.ml
+++ b/src/analyses/region.ml
@@ -3,8 +3,6 @@
 open Prelude.Ana
 open Analyses
 
-(* module Equ = MusteqDomain.Equ *)
-module Equ = RegionDomain.Equ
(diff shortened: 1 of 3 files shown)


michael-schwarz:
LGTM, let's merge it!

Issue #119, opened 2020-10-20
Field on a non-compound errors in SV-COMP SoftwareSystems


sim642:
In my own run of Goblint on SV-COMP SoftwareSystems category, Goblint crashed with `Errormsg.Error` exception

```
Bug: typeOffset: Field on a non-compound
```

on the following benchmarks:

```
ldv-consumption/32_7a_cilled_linux-3.8-rc1-32_7a-drivers--usb--host--isp1362-hcd.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.i
ldv-consumption/32_7a_cilled_linux-3.8-rc1-32_7a-drivers--pcmcia--pcmcia.ko-ldv_main2_sequence_infinite_withcheck_stateful.cil.out.i
ldv-consumption/32_7a_cilled_linux-3.8-rc1-drivers--usb--host--r8a66597-hcd.ko-main.cil.out.i
ldv-linux-3.16-rc1/205_9a_array_safes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--arcnet--rfc1051.ko-entry_point.cil.out.i
ldv-linux-3.16-rc1/205_9a_array_safes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--arcnet--capmode.ko-entry_point.cil.out.i
ldv-linux-3.16-rc1/205_9a_array_safes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--arcnet--arc-rawmode.ko-entry_point.cil.out.i
ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-43_2a-drivers--mmc--host--rtsx_pci_sdmmc.ko-entry_point.cil.out.i
ldv-linux-4.0-rc1-mav/linux-4.0-rc1---drivers--scsi--megaraid.ko.cil.i
ldv-linux-3.0/module_get_put-drivers-bluetooth-btmrvl.ko.cil.out.i
ldv-consumption/32_7a_cilled_linux-3.8-rc1-drivers--isdn--mISDN--l1oip.ko-main.cil.out.i
ldv-consumption/32_7a_cilled_linux-3.8-rc1-32_7a-drivers--mmc--host--vub300.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.i
ldv-linux-4.0-rc1-mav/linux-4.0-rc1---drivers--usb--host--u132-hcd.ko.cil.i
[...]

PR #128, opened 2020-11-05
Cast ints on read instead of write in base analysis

sim642:
This is an experimental attempt towards fixing #119.

The only reason types are required in `VD.update_offset` is to cast ints to appropriate types when written to through pointers which have been cast to a different type. This moves that cast from write time to later read time. Thus eliminating the need for having (problematic) type info available when writing.

This cast in the `Var` case is actually consistent with the already existing cast in the `Mem` case.

The bigger impact of this simple change is unknown. It could also be problematic for #121.

---

#### TODO
- [x] Actually remove type argument from `VD.update_offset`.
- [x] Run SV-COMP SoftwareSystems benchmarks to see if that fixes all of the type calculation errors.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -645,7 +645,14 @@ struct
           let x = String.sub x 2 (String.length x - 3) in (* remove surrounding quotes: L"foo" -> foo *)
           `Address (AD.from_string x) (* `Address (AD.str_ptr ()) *)
         (* Variables and address expressions *)
-        | Lval (Var v, ofs) -> do_offs (get a gs st (eval_lv a gs st (Var v, ofs)) (Some exp)) ofs
+        | Lval ((Var v, ofs) as b) ->
+          let t = typeOfLval b in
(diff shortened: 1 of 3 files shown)


sim642:
I did another SV-COMP SoftwareSystems run on the commit exactly before this PR for comparison. The differences are:
* It fixes all the `Field on a non-compound` errors. One of them turns into a correct true answer, the rest become unknown or time out after 60 seconds.
* Understandably, now blindly casting even more things just to be sure adds a tiny performance penalty. Luckily this is relatively minor: only three benchmarks turn from a correct true answer to a 60 second timeout.

So SV-COMP-wise this doesn't change much, a few benchmarks this way or that way. Therefore I don't feel too strongly about merging this for SV-COMP but it's related to issues we'd still want to have fixed.

michael-schwarz:
My intuition would be merging this for `svcomp`, and then working on our more long-term, less ad-hoc solution after that. Let's discuss this tomorrow at the GobCon.

PR #129, opened 2020-11-06
Make fast_global_inits work again

### LINKED ISSUE #122: Make fast_global_inits work for Partitioned Arrays
Currently, we need to turn off `fast_global_inits` for running with partitioned arrays.
This mean executing `n*m` assigns to `0` for a zero-initialized global array `a[n][m]` which is expensive.
We can avoid this by adding a special function to initialize a complete array to `0` to the array domain.

michael-schwarz:
By default, we generate `n*m` assign edges for a global array of dimensions `array[n][m]` when dealing with global initializers.

As an improvement, @vogler added the option `exp.fast_global_inits` where, for zero-initialized global arrays, only one assign to zero is generated: `array[0][0] = 0;`. This only works for the flat array domain, so we had to disable it when using the partitioned arrays (which we also do in SVCOMP).

**This PR makes this option compatible with partitioned arrays. To this end, we**

- Introduce `Base.zero_init_value t` to make a zero-initialized abstract value of type `t`.
   -  (In this step, we also remove unused arguments from `Base.init_value`, `Base.top_value`, and `Base.bot_value`)
- Add a new transfer function `zeroinit` that takes a `varinfo`
- Make `MyCFG.getGlobalInits file` return a list of `varinfo` to be zero-initialized on top of a list of edges for initialization
- Make `exp.fast_global_inits` the default and remove `--disable exp.fast_global_inits` from all test explicitly setting it

We chose to still generate the `array[0][0] = 0` edge, even when `exp.fast_global_inits` is enabled, causing `zeroinit array` to be called later.
This is so that it is safe to do nothing on `zeroinit` in all analyses except `base`. They still get the same assignment they got when `fast_global_inits` was implemented in the old way, so if they were sound before this change to `fast_global_inits`, they are still.

**The performance impact is quite big for large arrays:**

```
int global_array[50][200][20];

int main(void) {
  for(int i =0; i < 50; i++) {
      assert(global_array[i][42][7] == 0);
  }
}
```
Without `fast_global_inits`:
```
TOTAL                          157.453 s
  analysis                       157.450 s
    solving                         0.008 s

Memory statistics: total=293813.44MB, max=166.58MB, minor=293798.83MB, major=806.48MB, promoted=791.88MB
    minor collections=140102  major collections=24 compactions=0
```

With `fast_global_inits`:
```
TOTAL                           0.186 s
  analysis                        0.185 s
    solving                         0.009 s

Memory statistics: total=235.24MB, max=4.39MB, minor=233.53MB, major=3.47MB, promoted=1.76MB
    minor collections=112  major collections=3 compactions=0
```

Closes #122

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1620,7 +1620,7 @@ struct
   (**************************************************************************
    * Simple defs for the transfer functions
    **************************************************************************)
-  let zeroinit ctx v =
+  let zero_init ctx v =
     set ctx.ask ctx.global ctx.local (AD.from_var v) (zero_init_value v.vtype)
 
   let assign ctx (lval:lval) (rval:exp):store  =
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1620,6 +1620,9 @@ struct
   (**************************************************************************
    * Simple defs for the transfer functions
    **************************************************************************)
+  let zeroinit ctx v =
+    set ctx.ask ctx.global ctx.local (AD.from_var v) (zero_init_value v.vtype)
+
   let assign ctx (lval:lval) (rval:exp):store  =
     let char_array_hack () =
       let rec split_offset = function
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -793,10 +793,10 @@ struct
           M.debug ("Failed evaluating "^str^" to lvalue"); do_offs AD.unknown_ptr ofs
       end
 
-  let rec bot_value a (gs:glob_fun) (st: store) (t: typ): value =
+  let rec bot_value (t: typ): value =
     let bot_comp compinfo: ValueDomain.Structs.t =
       let nstruct = ValueDomain.Structs.top () in
-      let bot_field nstruct fd = ValueDomain.Structs.replace nstruct fd (bot_value a gs st fd.ftype) in
+      let bot_field nstruct fd = ValueDomain.Structs.replace nstruct fd (bot_value fd.ftype) in
       List.fold_left bot_field nstruct compinfo.cfields
     in
     match t with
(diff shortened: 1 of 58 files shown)


src/analyses/base.ml
```diff
@@ -1585,13 +1613,16 @@ struct
...
     match v with
-    | `Top -> set ask gs st adr (top_value ask gs st (AD.get_type adr)) ?lval_raw ?rval_raw
+    | `Top -> set ask gs st adr (top_value (AD.get_type adr)) ?lval_raw ?rval_raw
     | v -> set ask gs st adr v ?lval_raw ?rval_raw
 
 
   (**************************************************************************
    * Simple defs for the transfer functions
    **************************************************************************)
+  let zeroinit ctx v =
```
vogler, line 1623: Probably should be `zero_init` as above it's `zero_init_value`.

vogler:
Seems a bit heavy to introduce a transfer function for it.
Maybe use a special `zero_init` expression that `assign` can react to? After all, it's only a special case of `assign`.

michael-schwarz:
> Maybe use a special `zero_init` expression that `assign` can react to? After all, it's only a special case of `assign`.

I also thought about that. My reasoning against it was that it introduces a new variant for CIL expression that we then need to handle everywhere we do anything with expressions (also inside CIL), and that we would have to modify the assign in every analysis.

The other thing I considered was adding some sort of boolean `zero_init` to the `assign` edge, but that also didn't look particularly nice.

src/analyses/base.ml
michael-schwarz: I'll refactor that tomorrow.

vogler:
I meant using something like `MyCFG.unknown_exp`, just as an int of the right type.
Maybe `MyCFG.all_indices` which is meant to abstract paths for all definite indices for the size of the type.
Every place that does not handle it will just fall back to top.

michael-schwarz:
> I meant using something like `MyCFG.unknown_exp`, just as an int of the right type.

I tried this in #133. But I am not convinced it is nicer than this one here.

michael-schwarz:
We fixed this the hacky way by mergingn #133 for now, when we want to support initializers properly (i.e. not just zero initialization), the way to go would probably be this one.

PR #130, opened 2020-11-10
Extract base analysis flag to separate analysis, refactor thread spawning

sim642:
This is my ongoing effort to extract the flag (containing a "boolean" of single-/multithreaded mode and current thread ID) from base analysis to a separate analysis.

Additionally, I'm trying to refactor thread spawning, which multiple analyses currently try to do in parallel in `special` in a very similar way. The idea is to add additional transfer functions (currently `threadenter` and `threadspawn`) which make other analyses aware of a thread spawned by base analysis without needing to duplicate the logic.

This should make it easier to handle thread joins (`pthread_join`) and in simpler cases go back to singlethreaded mode, avoiding races that come from the end of main thread after other threads have been joined.


michael-schwarz:
I think separately from this PR, we should probably at some point cleanup the regression tests that specify which specific analyses to run explicitly to:
- use the default ones, and add any additionally needed ones
-  and only specify exactly which ones to run where necessary

This will make future diffs smaller.

src/framework/constraints.ml
```diff
@@ -204,7 +213,7 @@ struct
 
   let startstate v = (S.startstate v, !start_level)
   let exitstate  v = (S.exitstate  v, !start_level)
-  let otherstate v = (S.otherstate v, !start_level)
+  (* let otherstate v = (S.otherstate v, !start_level) *)
```
michael-schwarz, line 216: Why still keep this around here ?

michael-schwarz:
> The idea is to add additional transfer functions (currently threadenter and threadspawn) which make other analyses aware of a thread spawned by base analysis without needing to duplicate the logic.

Could you elaborate a bit on when these transfer functions would be called specifically?
And also how does this relate to the `otherstate` which was somehow related to this if I see correctly?

src/analyses/threadFlag.ml
```diff
@@ -0,0 +1,87 @@
...
+
+open Prelude.Ana
+open Analyses
+
+let is_multi_ask (ask: Queries.ask): bool =
+  match ask Queries.SingleThreaded with
+  | `Bool b -> not b
+  | `Top -> true
+  | _ -> failwith "is_multi_ask"
+let is_multi ctx: bool = is_multi_ask ctx.ask
```
michael-schwarz, line 14: Why do we need both of these here? Any `ctx` would also contain a `ask`, so what is the downside of using `is_multi_ask ctx.ask` vs. `is_multi ctx`? Passing around the entire `ctx` here confused me a bit at first.

src/analyses/threadFlag.ml
```diff
@@ -0,0 +1,87 @@
...
+
+  let special ctx lval f args =
+    ctx.local
+
+  let query ctx x =
+    match x with
+    | Queries.SingleThreaded -> `Bool (Queries.BD.of_bool (not (Flag.is_multi ctx.local)))
+    | Queries.IsNotUnique -> `Bool (Flag.is_bad ctx.local)
+    (* This used to be in base but also commented out. *)
+    (* | Queries.IsPublic _ -> `Bool (Flag.is_multi ctx.local) *)
```
michael-schwarz, line 67: Uncommenting this would affect the privatization of `precious_globals` (which is in turn somehow related to `early_globs`). Also this should probably only return `Bool false` if `Flag.is_multi ctx.local` is `false` and `Top` otherwise. 

https://github.com/goblint/analyzer/blob/813b9ee72a73bdf486356063cf6f8c9546bf8cf3/src/analyses/base.ml#L36-L40

src/analyses/uninit.ml
```diff
@@ -28,7 +28,8 @@ struct
   let should_join x y = D.equal x y
 
   let startstate v : D.t = D.empty ()
-  let otherstate v : D.t = D.empty ()
+  let threadenter ctx f args : D.t = D.empty ()
+  let threadspawn ctx f args fctx = D.bot ()
```
michael-schwarz, line 32: `D.bot ()` and `D.empty ()` are the same for `ValueDomain.AddrSetDomain`, so maybe use `D.bot ()` here instead of `D. empty ()`

src/analyses/poly.ml
```diff
@@ -19,7 +19,7 @@ struct
   let val_of x = x
   let context x = if GobConfig.get_bool "exp.full-context" then x else D.bot ()
 
-  let otherstate _ = D.top ()
+  let threadenter ctx f args = D.top ()
```
michael-schwarz, line 22: Is `threadspawn` missing here? We don't build this file by default, so `dune` wouldn't complain.

michael-schwarz:
It is definitely a lot nicer to have these things separate now, good job!

sim642:
> I think separately from this PR, we should probably at some point cleanup the regression tests that specify which specific analyses to run explicitly to:
> 
>     * use the default ones, and add any additionally needed ones
> 
>     * and only specify exactly which ones to run where necessary
> 
> 
> This will make future diffs smaller.

Absolutely! I was thinking the same when i had to fix the merge conflicts in 100 regression test `PARAM`s.

src/framework/constraints.ml
sim642: I now remember that there should be a TODO there because `threadenter` doesn't go back to `!start_level` right now. Whatever difference that makes...

sim642:
> > The idea is to add additional transfer functions (currently threadenter and threadspawn) which make other analyses aware of a thread spawned by base analysis without needing to duplicate the logic.
> 
> Could you elaborate a bit on when these transfer functions would be called specifically?
> And also how does this relate to the `otherstate` which was somehow related to this if I see correctly?

The new `threadenter` is basically the old `otherstate` on steroids: it gets much more information as arguments than just the function `varinfo`, notably the access to `ctx.local` and `ctx.ask`. As before, this transfer function is used when a new thread's initial state is to be calculated after `ctx.spawn` has been done somewhere.

The new `threadspawn` is a bit like combine (I originally had it named `threadcombine` but it was misleading) but it combines the new thread's initial state (after all analyses have done their `threadenter` to get that) with the current thread's state. This has two main goals:
1. The thread analysis is made aware of a new thread that was spawned from the current thread, thus it can add that to its local created threads set. Previously the thread analysis had to duplicate and match base analysis logic to know when that happens. Now only base calls `ctx.spawn` and via these transfer functions everyone else is aware.
2. In #137 the base analysis can ask for the created thread ID (which is calculated by threadid analysis' `threadenter`) in order to set the local `pthread_t` variable to point to that.

The biggest peculiarity with `threadspawn` (or multiple) is that it happens _in parallel_ with a normal transfer function and those are joined, like how `ctx.split` works. This is because if base's `special` does a `ctx.spawn`, then thread's `special` can't be made aware of the fact that something spawned (unless we use `ctx.postsub` I guess, but that's so ugly). Also, since in some kernel cases base's `assign` may do `ctx.spawn`, this parallel `threadspawn` is nicely compatible with threads being spawned in every weird place or condition.

If only we had somewhere to nicely document this sort of things...

src/analyses/threadFlag.ml
sim642: I think it should be fine as is, because in the boolean lattice used for these results `Bool true` is the top of that type. I've had to carefully consider the behavior of the related boolean queries because of this to make sure that now with multiple analyses answering the thread-related ones, they meet in the right way.

As for this commented out query, I guess the idea at some point was that in single threaded mode everything should be private anyway. But that is checked in base's `is_private` now, so the query wasn't needed. I'll leave it commented out here though because I suppose it could be changed back (along with the precious globs check).

src/framework/constraints.ml
michael-schwarz: I guess that should be fixed then: The way I understand this lifter, it limits how deeply we descend into function calls. 
When we get below `0`, the `EvalFunvar` query will return `[]` and the call will be a nop.

In the case of a new thread, we should start this counter at `!start_level` again for the new thread, and not with the current one.

src/analyses/uninit.ml
sim642: I used `D.empty ()` for `threadenter` because the previous `otherstate` used that. And I used `D.bot ()` for `threadspawn` because that more clearly expresses the notion that when `threadspawn` is joined with its parallel transfer function, it will not affect that.

michael-schwarz: Alright!

sim642:
#### consumption/32_7a_cilled_linux-3.8-rc1-32_7a-sound--core--snd-rawmidi.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.i

Something quite funny is happening that this turns `unknown`. Apparently we end up in multithreaded mode which obviously makes us less precise. It's a SoftwareSystems program and doesn't use any threading though, so what gives? Apparently we're spawning a function through a function pointer given to an special unknown function (in this case the standard `free`). The program isn't doing something stupid like freeing a function pointer though but instead freeing a struct which happens to contain function pointers. The unknown function spawning logic uses `collect_funargs` to find all reachable function pointers from that and that's how `snd_rawmidi_input_event_work` ends up being spawned by `free`.

**Why didn't this happen before?** Probably because thanks to mallocWrapper the freed blob is more precise (a struct containing a function pointer, not top).

**Why didn't this happen immediately when mallocWrapper was added but just now with thread flag changes?** I didn't bother looking into it that deep but my guess is that the thread spawning logic, which previously was duplicated in multiple analyses, had weaker implementations in some (not using reachability for unknown specials). Now that this happens centrally in base where the logic is the most powerful, it now arises.

Obviously this is silly behavior, so I see two ways to avoid this from happening:
1. **Add a new config option to control whether unknowns specials spawn.** This can be enabled by default, keeping our current logic, but disabled for SV-COMP because there programs should be complete and not call unknown functions which may spawn threads.
2. **Make our `LibraryFunctions` invalidate actions mechanism more fine-grained.** Right now `free` is specified as `writesAll` because it mutates the pointed memory but that doesn't mean it may spawn threads from it. Even the unknown specials spawning logic has a concerned comment:
   ```ocaml
    (* why do we only spawn arguments that are written?? *)
   ```
   We'd probably be better off having a another invalidation action to specify which arguments of which special functions may spawn.

sim642:
The pull_request CI checks now seem to be failing because the added regression test 29/21 in master doesn't enable `threadid` and `threadflag` analyses, which would be a simple enough fix. Actually adding `threadid` is enough to fix it.

But it's weird because base analysis doesn't depend on `threadid` at all. The crash is
```
Fatal error: exception Failure("BUG: Empty set of start variables; may happen if enter_func of any analysis returns an empty list.")
```
What makes is weird that _adding_ analyses fixes the start variables, not removing problematic analyses.

PR #131, opened 2020-11-10
Fix arithmetic on bot failures for SV-COMP benchmarks

### LINKED ISSUE #115: ArithmeticOnIntegerBot exceptions in SV-COMP SoftwareSystems
In my own run of Goblint on SV-COMP SoftwareSystems category, Goblint crashed with `ArithmeticOnIntegerBot` exception on the following benchmarks:

- [ ] ldv-linux-3.0/usb_urb-drivers-vhost-vhost_net.ko.cil.out.i (line 7168, `bottom op [-2147483648,2147483647]`)
- [ ] ldv-linux-3.14/linux-3.14_complex_emg_linux-alloc-spinlock_drivers-net-wireless-orinoco-orinoco_usb.cil.i (line 7732, `[0,0] op bottom`)
- [ ] ldv-linux-4.0-rc1-mav/linux-4.0-rc1---drivers--usb--gadget--udc--net2272.ko.cil.i (line 6546, `[0,0] op bottom`)

After the `hidden` parse error fix (https://github.com/goblint/cil/issues/18, https://github.com/goblint/cil/commit/26847e87655fa7a5c8d9e14d9d9cee81eee6b290) another one appeared:
- [ ] ldv-consumption/32_7a_cilled_linux-3.8-rc1-drivers--vfio--pci--vfio-pci.ko-main.cil.out.i (line 4343, `Error int op 0`)

jerhard:
This should fix #115.
Three issues are handled by this:
- for the `Interval` domain, `starting n`, `ending n` now return a `top` value, if `n < 2^31 or > 2^31-1`
- in `evalbinop` and `evalunop` in `Base`, a cast to the type of the expression is now performed after executing the operation on the abstract values. This is necessary as the operations on the intdomains are not aware of the `ikind` as of now
- catch the `IntDomain.ArithmeticOnIntegerBot` in `inv_bin_int` and raise `Deadcode` in `inv_exp`, if both subexpressions of a binary operation were bottom


michael-schwarz:
LGTM!

tests/regression/29-svcomp/11_arithmetic_bot.c
```diff
@@ -0,0 +1,40 @@
+// PARAM: --conf=conf/svcomp21.json
```
sim642, line 1: Might be better to have the few specific parameters added directly in the tests.

sim642 (commented):
I re-ran this on SoftwareSystems and it does fix the crashes, so that's good!

There's five performance regressions (correct true -> TIMEOUT) though. The most extreme one of these is ldv-challenges/linux-3.14_linux-kernel-locking-spinlock_drivers-net-ethernet-amd-pcnet32.cil.i, which before gave a correct answer before in 0.4 seconds but now timeouts in 60 seconds. I'm wondering what about these changes can have such performance impact.

michael-schwarz:
> I'm wondering what about these changes can have such performance impact.

Probably cases where because of unsoundness some parts of the code were deemed unreachable that now need to be analyzed.

tests/regression/29-svcomp/11_arithmetic_bot.c
jerhard: Ok, done https://github.com/goblint/analyzer/pull/131/commits/b061cc03d3af59807ddc70773c510b81022f7d06

sim642:
> Probably cases where because of unsoundness some parts of the code were deemed unreachable that now need to be analyzed.

That makes very much sense! It also explains why these unsoundness issues didn't reveal them as incorrect answers: they just happened to be inside correct programs. So previously we gave the correct answer for wrong reasons, which is something witness validation is (at least ideally) supposed to catch.

Looks good then.

PR #132, opened 2020-11-10
Malloc wrappers

### LINKED ISSUE #123: Better handling of calls to `malloc` wrapped inside functions such as `ldv_malloc`
This is a follow-up to #119:

A lot of tests in `sv-comp` wrap calls to `malloc` into calls to the helper function `ldv_malloc`. We only have one abstract heap object for all memory allocated via `ldv_malloc` (as the `malloc` for these calls is always in the same place), which might be a big source of imprecision.

The easiest solution would be to treat `ldv_malloc` specially just like we do `malloc`, but according to @sim642 there are different versions of `ldv_malloc` with some of them also simulating failing `malloc` calls while normally we can assume them to succeed in `sv-comp`.

Another approach would be to have some limited form of context-sensibility specifically for memory allocated inside functions such as `ldv_malloc`.

ivanazuzic:
This should fix #123:

- A new `mallocWrapper` analysis is created for the purpose of handling `malloc` calls wrapped inside other functions
- The malloc wrapper analysis is now called from `base` analysis for `calloc` and `malloc`
- Hash table for heap variables was moved from `base` analysis to `mallocWrapper` analysis
- The functions which are interesting as malloc wrappers are stored in a configurable list `exp.malloc.wrappers`


src/analyses/base.ml
```diff
@@ -110,13 +110,16 @@ struct
   let return_var () = AD.from_var (return_varinfo ())
   let return_lval (): lval = (Var (return_varinfo ()), NoOffset)
 
-  let heap_var loc = AD.from_var (BaseDomain.get_heap_var loc)
+  let heap_var ctx = 
+    let mallocloc = match (ctx.ask Q.HeapVar) with
+      | `Varinfo (`Lifted vinfo) -> vinfo
+      | _ -> failwith("Sadly no heap variable found.") in
```
michael-schwarz, line 116: I would fail with something like `ran without a malloc analysis`

src/analyses/base.ml
```diff
@@ -2101,9 +2104,9 @@ struct
     | `Calloc size ->
       begin match lv with
         | Some lv -> (* array length is set to one, as num*size is done when turning into `Calloc *)
-          let heap_var = BaseDomain.get_heap_var !Tracing.current_loc in (* TODO calloc can also fail and return NULL *)
-          set_many ctx.ask gs st [(AD.from_var heap_var, `Array (CArrays.make (IdxDom.of_int Int64.one) (`Blob (VD.bot (), eval_int ctx.ask gs st size)))); (* TODO why? should be zero-initialized *)
-                                  (eval_lv ctx.ask gs st lv, `Address (AD.from_var_offset (heap_var, `Index (IdxDom.of_int 0L, `NoOffset))))]
+          let heap_var = heap_var ctx in (* TODO calloc can also fail and return NULL *)
```
michael-schwarz, line 2107: Maybe address this `TODO` and also fail if `exp.malloc.fail` is active?

src/analyses/base.ml
```diff
@@ -2101,9 +2104,9 @@ struct
     | `Calloc size ->
       begin match lv with
         | Some lv -> (* array length is set to one, as num*size is done when turning into `Calloc *)
-          let heap_var = BaseDomain.get_heap_var !Tracing.current_loc in (* TODO calloc can also fail and return NULL *)
-          set_many ctx.ask gs st [(AD.from_var heap_var, `Array (CArrays.make (IdxDom.of_int Int64.one) (`Blob (VD.bot (), eval_int ctx.ask gs st size)))); (* TODO why? should be zero-initialized *)
-                                  (eval_lv ctx.ask gs st lv, `Address (AD.from_var_offset (heap_var, `Index (IdxDom.of_int 0L, `NoOffset))))]
```
michael-schwarz, line 2106: Why remove the offset here?

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -0,0 +1,80 @@
...
+    with Not_found ->
+      let name = "(alloc@" ^ loc.file ^ ":" ^ string_of_int loc.line ^ ")" in
+      let newvar = Goblintutil.create_var (makeGlobalVar name voidType) in
+      Hashtbl.add heap_hash loc newvar;
+      newvar
+
+  let query ctx (q:Q.t) : Q.Result.t =
+    match q with
+    | Q.HeapVar -> 
+      let b = match ctx.local with
```
michael-schwarz, line 69: Why b?

sim642 (commented):
Seems to work as it should.

I'm now mostly wondering what's the best way to put together a good malloc wrapper functions list for SV-COMP. There were a couple names in a benchmark that originally lead to this issue here: https://github.com/goblint/analyzer/issues/123#issuecomment-717766048. Since the LDV benchmarks there seem to have similar framework (e.g `ldv_malloc`), then looking at some manually to be sure might be enough. It's very hard to be sure though.

src/util/defaults.ml
```diff
@@ -154,7 +154,8 @@ let _ = ()
       ; reg Experimental "exp.addr-context"      "false" "Ignore non-address values in function contexts."
       ; reg Experimental "exp.no-int-context"    "false" "Ignore all integer values in function contexts."
       ; reg Experimental "exp.no-interval32-context" "false" "Ignore integer values of the Interval32 domain in function contexts."
-      ; reg Experimental "exp.malloc-fail"       "false" "Consider the case where malloc fails."
+      ; reg Experimental "exp.malloc.fail"       "false" "Consider the case where malloc fails."
+      ; reg Experimental "exp.malloc.wrappers" "['myalloc','myalloc2','ldv_malloc','smalloc','no_malloc']"  "Loads a list of known malloc wrapper functions."
```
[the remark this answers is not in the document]
michael-schwarz: I would argue that we should have all of the things we map to `malloc` or `calloc` in https://github.com/goblint/analyzer/blob/master/src/analyses/libraryFunctions.ml in this list by default.

If we have any of these functions appears in a benchmark and we have a body for them (and don't use the `special` edge therefore), we probably want to treat them as wrappers around malloc.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -0,0 +1,80 @@
...
+      else D.top () in     
+    [(ctx.local, callectx)]
+
+  let combine ctx (lval:lval option) fexp (f:varinfo) (args:exp list) fc (au:D.t) : D.t =
+    ctx.local
+
+  let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    ctx.local
+
+  let startstate v = D.bot ()
```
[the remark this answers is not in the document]
michael-schwarz: This is also what we have my default in the `unit.ml` which is an "analysis specification for didactic purposes".
https://github.com/goblint/analyzer/blob/6d5af0bbf7bda9f6a8b8585672d7fabe9aae57a2/src/analyses/unit.ml#L37

src/util/defaults.ml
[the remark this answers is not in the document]
sim642: Seems reasonable indeed. It would then also cover the original intent of forcefully making `kzalloc` and `kmalloc` special just to keep their locations but without being unsound.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -0,0 +1,80 @@
...
+    ctx.local
+
+  let enter ctx (lval: lval option) (f:varinfo) (args:exp list) : (D.t * D.t) list =
+    let interestingfunctions = get_string_list "exp.malloc.wrappers" in
+    let calleofinterest = List.mem f.vname interestingfunctions in
+    let callectx = if calleofinterest then
+       if ctx.local = `Top then
+        `Lifted (MyCFG.getLoc ctx.node) 
+        else ctx.local
+      else D.top () in
```
sim642, line 43: The logic here is that a boring callee changes this to top, but wouldn't it be better to keep it at it's current value?
For example, if `ldv_malloc` that is listed as malloc wrapper calls some random wrapper `wrapper` that isn't listed as a malloc wrapper which then `malloc`s something. Right now, `wrapper` would forget the `ldv_malloc`s location and thus the inner `malloc` wouldn't have that precisely. Seems to me like we should remember the outermost malloc wrapper even if it goes through some things which we haven't listed as such.
Or does that cause some problem?

michael-schwarz: I think the reasoning was avoiding doing a callstring-sensitive analysis of other functions in case the malloc wrapper also calls some other (non malloc-wrapper) functions. In case we don't go to top there, we will analyze all things transitively called from a wrapper function for each caller of the wrapper. This could potentially become very expensive with not much gain.

sim642: My intuition would be that a malloc wrapper doesn't really call a whole bunch of different things that need lots of reanalysis. Right now, if multiple wrappers are used, we must list them all or we lose all the benefit. If the outermost wrapper was remembered through non-listed calls, then it's necessary to only have the outermost function listed. It would be prone against situations where another wrapper has been put into its implementation in some benchmarks.

We can leave it as is, but a comment in the code might be good to draw attention to the fact that it's supposed to currently forget wrappers in that case. If we have a somewhat sensible list of malloc wrappers for SV-COMP, I might want to do an experimental run the other way for comparison (of analysis times and precision).

src/analyses/base.ml
ivanazuzic: Summary of what was discussed in person: I removed it because of the type of heap_var, but the offset is necessary, so I put it back and  heap_var is now varinfo.

src/analyses/mallocWrapperAnalysis.ml
ivanazuzic: For now I wrote comments about this.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -0,0 +1,80 @@
...
+
+  let body ctx (f:fundec) : D.t =
+    ctx.local
+
+  let return ctx (exp:exp option) (f:fundec) : D.t =
+    ctx.local
+
+  let enter ctx (lval: lval option) (f:varinfo) (args:exp list) : (D.t * D.t) list =
+    let interestingfunctions = get_string_list "exp.malloc.wrappers" in
+    let calleofinterest = List.mem f.vname interestingfunctions in
```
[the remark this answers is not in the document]
ivanazuzic: I will write a comment about this, but I think the list won't be very long.

src/analyses/base.ml
```diff
@@ -2126,9 +2129,13 @@ struct
...
+          let heap_var = heap_var ctx in
+          let add_null addr =
+            if (get_bool "exp.malloc.fail")
```
michael-schwarz, line 2134: `()` are not needed

src/maingoblint.ml
```diff
@@ -110,8 +110,7 @@ let option_spec_list =
   ; "--osekisrsuffix"      , Arg.String (set_string "ana.osek.isrsuffix"), ""
   ; "--osekcheck"          , Arg.Unit (fun () -> set_bool "ana.osek.check" true), ""
   ; "--oseknames"          , Arg.Set_string OilUtil.osek_renames, ""
-  ; "--osekids"            , Arg.Set_string OilUtil.osek_ids, ""
-  ]
+  ; "--osekids"            , Arg.Set_string OilUtil.osek_ids, ""]
```
michael-schwarz, line 113: Why change this file?

src/util/defaults.ml
```diff
@@ -154,7 +154,8 @@ let _ = ()
       ; reg Experimental "exp.addr-context"      "false" "Ignore non-address values in function contexts."
       ; reg Experimental "exp.no-int-context"    "false" "Ignore all integer values in function contexts."
       ; reg Experimental "exp.no-interval32-context" "false" "Ignore integer values of the Interval32 domain in function contexts."
-      ; reg Experimental "exp.malloc-fail"       "false" "Consider the case where malloc fails."
+      ; reg Experimental "exp.malloc.fail"       "false" "Consider the case where malloc fails."
```
michael-schwarz, line 157: Description should probably also say that this now considers failing `calloc` as well

src/util/defaults.ml
```diff
@@ -154,7 +154,8 @@ let _ = ()
...
+      ; reg Experimental "exp.malloc.wrappers"   "['kmalloc','kmalloc','usb_alloc_urb','__builtin_alloca','kzalloc']"  "Loads a list of known malloc wrapper functions."
```
michael-schwarz, line 158: Maybe add a comment to `libraryFunctions.ml` that when adding something new that maps to `malloc` or `calloc`, they should be added here as well.

michael-schwarz:
LGTM!

src/util/defaults.ml
```diff
@@ -154,7 +154,8 @@ let _ = ()
       ; reg Experimental "exp.addr-context"      "false" "Ignore non-address values in function contexts."
       ; reg Experimental "exp.no-int-context"    "false" "Ignore all integer values in function contexts."
       ; reg Experimental "exp.no-interval32-context" "false" "Ignore integer values of the Interval32 domain in function contexts."
-      ; reg Experimental "exp.malloc-fail"       "false" "Consider the case where malloc fails."
+      ; reg Experimental "exp.malloc.fail"       "false" "Consider the case where malloc or calloc fails."
+      ; reg Experimental "exp.malloc.wrappers"   "['kmalloc','kmalloc','usb_alloc_urb','__builtin_alloca','kzalloc']"  "Loads a list of known malloc wrapper functions." (* When something new that maps to malloc or calloc is added to libraryFunctions.ml, it should also be added here.*)
```
sim642, line 158: `kmalloc` is twice on the list.

ivanazuzic: It should have been: 'kmalloc','__kmalloc'. Thank you for noticing.

sim642:
I suppose this can be merged now?

PR #133, opened 2020-11-11
Fast global inits without adding a new transfer function (See #129)

michael-schwarz:
**This is an alternative to #129 that also solves #122.**

As suggested, for this case we add a new expression 

~~~OCaml
let all_array_index_exp : exp = CastE(TInt(Cilfacade.ptrdiff_ikind (),[]), unknown_exp)`
~~~
 to `MyCFG` (with the semantics that an assign to this special index represents an assignment to all indices of an array) and in the assign in the partitioned array domain check if we are dealing with an assign to this expression.

This means touching less code, but I am not sure if it is nicer than #129. To me, it feels quite hacky. What do y'all think @vogler @sim642 @jerhard @vesalvojdani  ?

**Sidenote:** The diff of `src/cdomains/arrayDomain.ml` appears huge, Github somehow gets confused. The change is adding:

~~~OCaml
if i = `Lifted MyCFG.all_array_index_exp then
   (Expp.top(), (a, a, a))
else
~~~

The rest is indentation changing because of it.

### DIFF
--- a/src/util/defaults.ml
+++ b/src/util/defaults.ml
@@ -172,7 +172,7 @@ let _ = ()
       ; reg Experimental "exp.solver.td3.space" "false" "Should the td3 solver only keep values at widening points?"
       ; reg Experimental "exp.solver.td3.space_cache" "true" "Should the td3-space solver cache values?"
       ; reg Experimental "exp.solver.td3.space_restore" "true" "Should the td3-space solver restore values for non-widening-points? Needed for inspecting output!"
-      ; reg Experimental "exp.fast_global_inits" "false" "Only generate 'a[0] = 0' for a zero-initialized array a[n]. This is only sound for our flat array domain! TODO change this once we use others!"
+      ; reg Experimental "exp.fast_global_inits" "true" "Only generate 'a[0] = 0' for a zero-initialized array a[n]. This is only sound for our flat array domain! TODO change this once we use others!"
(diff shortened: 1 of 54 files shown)


src/util/defaults.ml
```diff
@@ -172,7 +172,7 @@ let _ = ()
       ; reg Experimental "exp.solver.td3.space" "false" "Should the td3 solver only keep values at widening points?"
       ; reg Experimental "exp.solver.td3.space_cache" "true" "Should the td3-space solver cache values?"
       ; reg Experimental "exp.solver.td3.space_restore" "true" "Should the td3-space solver restore values for non-widening-points? Needed for inspecting output!"
-      ; reg Experimental "exp.fast_global_inits" "false" "Only generate 'a[0] = 0' for a zero-initialized array a[n]. This is only sound for our flat array domain! TODO change this once we use others!"
+      ; reg Experimental "exp.fast_global_inits" "true" "Only generate 'a[0] = 0' for a zero-initialized array a[n]. This is only sound for our flat array domain! TODO change this once we use others!"
```
sim642, line 175: This description should then also say `a[all_array_index_exp] = 0` or something.

sim642 (approved):
I think I prefer this alternative, but that may just be because [I had a similar fix in mind](https://github.com/goblint/analyzer/issues/122#issuecomment-721156658). It is slightly hacky but it's just a variation of `unknown_exp` which we have anyway, so I'm not bothered by this. Especially since it's just the array domains that need to handle it.

---

By the way, in the PR files changed view there's a setting "Hide whitespace changes". Too bad it doesn't just stay enabled all the time.

PR #134, opened 2020-11-11
Interval: Don't compute on top values

jerhard:
This introduces a fix so we no longer compute on `top` values within the interval domain. 
The `top` in the interval is actually represented by the interval `[-2^31, 2^31-1]`. As the concrete values for a integer might lie outside this range, we cannot compute with this value, because this could yield unsound results. So instead, if one of the operands of a binary/unary operation is `top`, the result of the operation will be `top`, too.

Side note: Having a top value in an interval representation that is truly an over-approximation for all possible concrete values (including `-2^63`, `2^64-1`) would require an implementation with arbitrary precision integers.

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -361,15 +361,18 @@ struct
         meet (bit Int64.rem x y) range
 
   let mul x y =
-    on_top_return_top_bin x y @@
-    match x, y with
-    | None, None -> bot ()
-    | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (short 80 x) (short 80 y)))
-    | Some (x1,x2), Some (y1,y2) ->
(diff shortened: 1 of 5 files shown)


src/cdomains/intDomain.ml
```diff
@@ -266,7 +266,15 @@ struct
       let ur = if Int64.compare max_int x2 = 0 then y2 else x2 in
       norm @@ Some (lr,ur)
 
+  (* Returns top if one of a, b is top, otherwise return result *)
+  let on_top_return_top_bin a b result =
```
michael-schwarz, line 270: The way it is implemented, this swallows `ArithmeticOnIntegerBot` exceptions in case where one of the arguments is `\top` and the other one is `\bot`. Is this by design?

jerhard: Actually, it doesn't.

michael-schwarz:
LGTM!

PR #136, opened 2020-11-16
Add __VERIFIER_atomic_* support

### LINKED ISSUE #113: Add __VERIFIER_atomic_{begin,end} support
[SV-COMP rules](https://sv-comp.sosy-lab.org/2021/rules.php):

> **`__VERIFIER_atomic_begin()`**: For modeling an atomic execution of a sequence of statements in a multi-threaded run-time environment, those statements can be placed between two function calls `__VERIFIER_atomic_begin()` and `__VERIFIER_atomic_end()` (deprecated but still valid: those statements can be placed in a function whose name matches `__VERIFIER_atomic_`). The verifiers are instructed to assume that the execution between those calls is not interrupted. The two calls need to occur within the same control-flow block; nesting or interleaving of those function calls is not allowed.

---

Importantly, the two functions cannot simply correspond to locking and unlocking some special atomic lock, because such atomic block must also not interfere with any single statement outside of an atomic block.

sim642:
Closes #113.

Besides `__VERIFIER_atomic_begin` and `__VERIFIER_atomic_end`, it also implements the deprecated but still valid form of SV-COMP atomics by function name beginning with `__VERIFIER_atomic_`.

### DIFF
--- a/src/analyses/mutexAnalysis.ml
+++ b/src/analyses/mutexAnalysis.ml
@@ -197,14 +197,14 @@ struct
       | None -> ()
     end;
     (* deprecated but still valid SV-COMP convention for atomic block *)
-    if String.starts_with fundec.svar.vname "__VERIFIER_atomic_" then
+    if get_bool "ana.sv-comp.functions" && String.starts_with fundec.svar.vname "__VERIFIER_atomic_" then
       Lockset.remove (verifier_atomic, true) ctx.local
     else
       ctx.local
 
   let body ctx f : D.t =
     (* deprecated but still valid SV-COMP convention for atomic block *)
-    if String.starts_with f.svar.vname "__VERIFIER_atomic_" then
+    if get_bool "ana.sv-comp.functions" && String.starts_with f.svar.vname "__VERIFIER_atomic_" then
       Lockset.add (verifier_atomic, true) ctx.local
     else
       ctx.local
@@ -260,9 +260,9 @@ struct
       Lockset.remove (console_sem,true) ctx.local
     | _, "__builtin_prefetch" | _, "misc_deregister" ->
       ctx.local
-    | _, "__VERIFIER_atomic_begin" ->
+    | _, "__VERIFIER_atomic_begin" when get_bool "ana.sv-comp.functions" ->
       Lockset.add (verifier_atomic, true) ctx.local
-    | _, "__VERIFIER_atomic_end" ->
+    | _, "__VERIFIER_atomic_end" when get_bool "ana.sv-comp.functions" ->
       Lockset.remove (verifier_atomic, true) ctx.local
     | _, x ->
       let arg_acc act =
--- a/src/analyses/mutexAnalysis.ml
+++ b/src/analyses/mutexAnalysis.ml
@@ -193,13 +193,21 @@ struct
 
   let return ctx exp fundec : D.t =
     begin match exp with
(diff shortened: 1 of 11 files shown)


src/analyses/mutexAnalysis.ml
```diff
@@ -189,13 +193,21 @@ struct
...
-      | Some exp ->
-        access_one_top ctx false false exp;
-        ctx.local
-      | None -> ctx.local
-    end
+      | Some exp -> access_one_top ctx false false exp
+      | None -> ()
+    end;
+    (* deprecated but still valid SV-COMP convention for atomic block *)
+    if String.starts_with fundec.svar.vname "__VERIFIER_atomic_" then
```
michael-schwarz, line 200: Maybe only do this if `ana.sv-comp` is set?

src/analyses/mutexAnalysis.ml
```diff
@@ -248,6 +260,10 @@ struct
       Lockset.remove (console_sem,true) ctx.local
     | _, "__builtin_prefetch" | _, "misc_deregister" ->
       ctx.local
+    | _, "__VERIFIER_atomic_begin" ->
```
michael-schwarz, line 263: Same here, I'd suggest that this should happen only with `ana.sv-comp` on

src/analyses/mutexAnalysis.ml
sim642: I thought about it but that makes the corresponding regression tests weird: it then requires a property file and generates witnesses files in the middle of the tests directory when running the script.

sim642: I can make the SV-COMP option more granular to work around that somewhat, but there are still places where this can't be done conditionally. For example, `invalidate_actions` in `LibraryFunctions` could still in theory affect programs outside of whatever SV-COMP mode.

michael-schwarz: Ah, I see. Maybe we should have one flag that controls this and the `sv-comp` includes?

If some things can not easily be done conditionally, I guess it is fine to always include them (especially given that we want to give the library mechanism an overhaul in the long run anyway).

My main reason for having a flag to control this is that it is some sort of implicit documentation that is related to `sv-comp`, making it easier to find such places (as opposed to, e.g. the ARINC stuff that just appears everywhere).

src/util/defaults.ml
```diff
@@ -131,7 +131,8 @@ let _ = ()
       ; reg Analyses "ana.opt.equal"       "true"  "First try physical equality (==) before {D,G,C}.equal (only done if hashcons is disabled since it basically does the same via its tags)."
       ; reg Analyses "ana.restart_count"   "1"     "How many times SLR4 is allowed to switch from restarting iteration to increasing iteration."
       ; reg Analyses "ana.mutex.disjoint_types" "true" "Do not propagate basic type writes to all struct fields"
-      ; reg Analyses "ana.sv-comp"         "false" "SV-COMP mode"
+      ; reg Analyses "ana.sv-comp.enabled" "false" "SV-COMP mode"
+      ; reg Analyses "ana.sv-comp.functions" "false" "SV-COMP functions"
```
michael-schwarz, line 135: Maybe this should be a bit more verbose e.g.
`"Enable special handling for SV-COMP specific functions"`

michael-schwarz:
LGTM!

PR #137, opened 2020-11-17
Handle thread joins

sim642:
This is based on #130.

Firstly, it makes base analysis track which `pthread_t` variable corresponds to which of our own thread ID (fake `varinfo`). This is required, because `pthread_join` takes that as an argument to know which thread to join. The possible thread IDs a thread variable corresponds to are stored using address sets right now.
This is, however, problematic, because `pthread_t` is actually an integer type, so there's a disconnect between concrete value types and abstract value types. This itself shouldn't be a problem, because if a program treats a `pthread_t` as integer (e.g. does arithmetic), then we lose track of which thread IDs it corresponds to. The bigger issue is related to #128, which for this experiment is right now commented out (**which is why the test fails**) to prevent all thread ID address sets from being cast to unknown ints, making any joining impossible.

Secondly, based on the previous, the thread analysis, which currently keeps locally track of created threads as a set, now handles `pthread_join` by removing the joined thread from the set. This is only done if the joined thread is unique _and_ it has joined (by this same logic) all of its created threads when returning (i.e., its created thread set is empty at return node).
The thread analysis also answers to singlethreadedness queries: if it's the main thread and it's created thread set is empty, then it's back to being single threaded. That prevents accesses from being counted again, especially at the end of main thread after other threads have been joined.

This should also mutually improve the thread uniqueness analysis itself. For example, if a thread is created and joined in the same loop iteration, then it won't be considered to be repeatedly created there, because by the time it's created again, the previous instance has been removed from the set.

Thirdly, as a bonus, `pthread_join` can now also handle thread return values (either direct or via `pthread_exit`).


sim642:
As a further experiment, I tried to move most of these threading additions out of the base analysis as well using `ctx.assign` to have it still do the value tracking in the [`thread/join-assign`](https://github.com/goblint/analyzer/tree/thread/join-assign) branch. Unfortunately that has even bigger issues, so we can forget that for now.

sim642:
This behaves almost identically to #130 on SoftwareSystems, which isn't surprising because there threads aren't even used.

On NoDataRace it solves some extra cases that involve excluding races after joins but only 5. Maybe this still needs some additional improvements, I haven't looked into NoDataRace benchmarks that we still don't solve.

src/analyses/base.ml
```diff
@@ -2023,10 +2040,20 @@ struct
...
+      D.bot () (* actual results joined via threadspawn *)
     (* handling thread joins... sort of *)
     | `ThreadJoin (id,ret_var) ->
       begin match (eval_rv ctx.ask gs st ret_var) with
         | `Int n when GU.opt_predicate (BI.equal BI.zero) (ID.to_int n) -> cpa,dep
+        | `Address ret_a ->
+          begin match eval_rv ctx.ask gs st id with
+            | `Address a ->
+              (* TODO: is this type right? *)
+              set ctx.ask gs st ret_a (Cil.typeOf ret_var) (get ctx.ask gs st a None)
```
sim642, line 2054: This needs to be considered after re-adding type to `set`.

src/analyses/base.ml
```diff
@@ -2145,15 +2172,25 @@ struct
...
 
-  let threadspawn ctx (f: varinfo) (args: exp list) fctx: D.t =
-    D.bot ()
+  let threadspawn ctx (lval: lval option) (f: varinfo) (args: exp list) fctx: D.t =
+    match lval with
+    | Some lval ->
+      begin match ThreadId.get_current fctx.ask with
+        | `Lifted tid ->
+          (* TODO: is this type right? *)
+          set ctx.ask ctx.global ctx.local (eval_lv ctx.ask ctx.global ctx.local lval) (Cil.typeOfLval lval) (`Address (AD.from_var tid))
```
sim642, line 2188: This needs to be considered after re-adding type to `set`.

PR #138, opened 2020-11-19
Use fast_global_inits also for non-zero initialized global arrays

michael-schwarz:
I looked again at where we waste time before we start the actual solving: The only significant amount of time is spent computing the initial state (i.e. handling of global initializers).

As an improvement, in this PR we use `fast_global_inits` also for non-zero initialized arrays. This is faster whenever the same value is contained in an array multiple times. The only added costs are more lookups in the hash table.

For some of the `sv-comp` tests, this leads to a big decrease of the time spent doing global initializers. 
The most noticable drop was going **from >17000 assign edges to < 3500, and ~140s to ~13s spent on global initialization** for [this](https://github.com/sosy-lab/sv-benchmarks/blob/25191159b7ff6bb205f4107be9f83c0163073c1b/c/ldv-linux-3.14/linux-3.14_complex_emg_linux-alloc-spinlock_drivers-media-pci-saa7134-saa7134.cil.c) program. There are no examples where we are considerably slower.

**Downside** of this change is that we no longer know the exact value of the array at index zero.

~~~C
int global_array[5] = {0,0,0,0,42};
int main(void) {
  assert(global_array[0] == 0); // This will succeed without fast_global_inits and fail with it
}
~~~

On the other hand, this means that the array is unpartitioned after the global inits, so it might become partitioned somewhere else in the program, which could positively affect precision there.

### DIFF
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -295,7 +295,7 @@ struct
 
   let set_with_length length (ask:Q.ask) ((e, (xl, xm, xr)) as x) (i,_) a =
     if i = `Lifted MyCFG.all_array_index_exp then
-      (assert !Goblintutil.global_initialization; (* just joining with xr here assumes that all values will be set, which is guaranteed during inits *)
+      (assert !Goblintutil.global_initialization; (* just joining with xm here assumes that all values will be set, which is guaranteed during inits *)
       let r =  Val.join xm a in
       (Expp.top(), (r, r, r)))
     else
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -295,7 +295,8 @@ struct
 
   let set_with_length length (ask:Q.ask) ((e, (xl, xm, xr)) as x) (i,_) a =
     if i = `Lifted MyCFG.all_array_index_exp then
-      (Expp.top(), (a, a, a))
+      let r =  Val.join (join_of_all_parts x) a in
+      (Expp.top(), (r, r, r))
     else
       normalize @@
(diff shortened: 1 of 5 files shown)


sim642:
> **Downside** of this change is that we no longer know the exact value of the array at index zero.

And that's limited to just index 0, not any others?

In that case I'm thinking we could maybe keep that previous precision by first doing all the all-index inits like this now does but also adding the index 0 init after them or something? Even if that would work, I'm not sure we should because the speciality of index 0 seems more like our own oddity than something which is required to verify many practical programs.

michael-schwarz:
> And that's limited to just index 0, not any others?

Since we see all writes consecutively, in this case it will be index 0. If the order was somehow different, it would be a different position that we have exact information on.
This partitioned array domain has one value for everything to the left, an exact value at some index, and then everything to the right. This is mostly helpful for local variables, as this partitioning expression is often e.g. a loop index, allowing us to have different information for the position at the loop index, to the left and to the right. I can send you a copy of my Master's thesis if you're interested.
For globals this is mostly not really helpful though (except maybe in single-threaded programs).

> In that case I'm thinking we could maybe keep that previous precision by first doing all the all-index inits like this now does but also adding the index 0 init after them or something? 

Yes we could.

> Even if that would work, I'm not sure we should because the speciality of index 0 seems more like our own oddity than something which is required to verify many practical programs.

I agree, probably not really interesting for most programs, and in case of single-threaded programs, it might actually be nice to be able to partition them according to some other expression later.

michael-schwarz:
Unless anyone is opposed, I'd suggesting merging this.

sim642:
Looks good

Issue #141, opened 2020-11-23
Unsoundness in SV-COMP Prerun


michael-schwarz:
In the following cases we were unsound for the pre-runs:

- [x] bitvector/s3_clnt_3.BV.c.cil-2a.yml
  - Turns into a timeout with #121
- [x] float-benchs/nan_double.yml
  - Still happens after merging #121
- [x] float-benchs/nan_float.yml
  - Still happens after merging #121
- [x] float-newlib/double_req_bl_1210.yml
  - Still happens after merging #121
- [x] float-newlib/float_req_bl_1210.yml
  - Still happens after merging #121
- [x] ldv-linux-3.14-races/linux-3.14--drivers--usb--misc--iowarrior.ko.cil.yml
  - Still happens after merging #121 
  - See #143 
- [x] pthread-driver-races/char_generic_nvram_read_nvram_write_nvram.yml
  - Still happens after merging #121 
  - See #143
- [x] goblint-regression/28-race_reach_45-escape_racing.yml
  - Still happens after merging #121 
  - See #142 
- [x]  ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--staging--lustre--lustre--llite--llite_lloop.ko-entry_point.cil.out.yml
  - Fixed by https://github.com/goblint/analyzer/commit/7d0fbeefba2cdcaa66c1b53b8d2cc05827189f8f and `unknown` result on current master

michael-schwarz:
I'll check which of these are problematic also after merging #121

michael-schwarz:
The first two are variations of this:

~~~C
/* Assertion can be violated because uninitialized doubles can be NaN.  */

int main()
{
  double x;
  x = __VERIFIER_nondet_double();

  __VERIFIER_assert(x==x);
  return 0;
}
~~~

sim642:
Right now on master the last one gives `unknown` for me. This was fixed by https://github.com/goblint/analyzer/commit/7d0fbeefba2cdcaa66c1b53b8d2cc05827189f8f. Does #121 break it again or what?

michael-schwarz:
> Right now on master the last one gives `unknown` for me. This was fixed by [7d0fbee](https://github.com/goblint/analyzer/commit/7d0fbeefba2cdcaa66c1b53b8d2cc05827189f8f). Does #121 break it again or what?

It doesn't I was running with the wrong property :facepalm: 
[...]

PR #142, opened 2020-11-23
Fix var_eq analysis not invalidating escaped variable

sim642:
This is the source of unsoundness for SV-COMP goblint-regression/28-race_reach_45-escape_racing.yml (issue #141), which is actually from our regression suite. There it doesn't fail because it's run without `var_eq`.

`var_eq` keeps the equality `i == 0` after `i` escapes through the 4th argument of `pthread_create`. Changing `LibraryFunctions` to invalidate that argument also fixes this test, but breaks 05/17. Because of this I'm not sure if this is the right fix or maybe there's something better to do.

I suspect not because our invalidation and `var_eq` are quite ignorant about passing pointers. In one case `&i` is passed, in the other only `p`. The thing is that the argument (which is a pointer) itself wouldn't be changed by such `pthread_create` but only whatever the argument points to.

Another oddity is that `base` also doesn't invalidate `i` in that example. But only later when it is read (for the assertion), the global invariant is used instead because according to `escape` it has escaped.

### DIFF
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -394,7 +394,7 @@ let invalidate_actions = ref [
     "dev_driver_string", readsAll;
     "__spin_lock_init", writes [1];
     "kmem_cache_create", readsAll;
-    "pthread_create", writes [1;4];
+    "pthread_create", onlyWrites [1];
     "__builtin_prefetch", readsAll;
     "idr_pre_get", readsAll;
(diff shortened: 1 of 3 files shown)


sim642:
Hard-coding the `pthread_create` argument invalidation only to `var_eq` seems to fix the issue without introducing a regression.

I guess it wasn't set to invalidate in `LibraryFunctions` to avoid generating a spurious write access directly on `pthread_create` itself.

PR #143, opened 2020-11-23
Fix var_eq from keeping equalities with pointers to globals

sim642:
See https://github.com/goblint/analyzer/issues/141#issuecomment-732080297.

### DIFF
--- a/src/analyses/varEq.ml
+++ b/src/analyses/varEq.ml
@@ -353,6 +353,29 @@ struct
       | _ -> D.top ()
     *)
 
+  let rec is_global_var (ask: Queries.ask) x =
+    match x with
+    | SizeOf _
+    | SizeOfE _
+    | SizeOfStr _
+    | AlignOf _
+    | AlignOfE _
+    | UnOp _
+    | BinOp _ -> None
+    | Const _ -> Some false
+    | Lval (Var v,_) -> Some v.vglob
+    | Lval (Mem e, _) ->
+      begin match ask (Queries.MayPointTo e) with
+        | `LvalSet ls when not (Queries.LS.is_top ls) && not (Queries.LS.mem (dummyFunDec.svar, `NoOffset) ls) ->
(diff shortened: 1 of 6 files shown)


sim642:
The original pthread-driver-races/char_generic_nvram_read_nvram_write_nvram.i has following in the global invariant:
* `whoop_loff_t -> Address {&(alloc@...)}`
* `&(alloc@...) -> Blob (Int)`
I don't understand how `NULL` isn't in the address set of `whoop_loff_t` anymore, because it is zero-initialized to that.

In both of the minimized versions I've committed here, the `NULL` remains there and the malloc also adds unknown pointer:
* `whoop_loff_t -> Address {NULL, ?}`
* `&(alloc@...) -> Blob (Bot)`
The PR handles both cases but it'd be great to actually also test that but I'm very confused how the original behavior is even possible.

michael-schwarz:
> I don't understand how NULL isn't in the address set of whoop_loff_t anymore, because it is zero-initialized to that.

We don't have early globals on in `sv-comp`. And it only goes to multithreading after `whoop_loff_t` is set if I see it correctly? So this should be the expected behavior, no?

sim642:
But the 29/20 I committed should then behave the same way, by starting out single threaded as well, but for whatever reason it doesn't.

michael-schwarz:
> But the 29/20 I committed should then behave the same way, by starting out single threaded as well, but for whatever reason it doesn't.

True, that is indeed strange! I'll take a  look maybe, I can see what's going on.

michael-schwarz:
If you add `#include <stdlib.h>` to your example program, it behaves just like the big example.

michael-schwarz:
I think we can merge this now

sim642:
> If you add `#include <stdlib.h>` to your example program, it behaves just like the big example.

Of course, thanks for noticing that! I now remember fixing a similar thing in a previous regression test.

It's so annoying though that due to `malloc` being so specialized we don't have any warnings about it even lacking a _declaration_ and that `int` is implicitly assumed to be the return type, which really screws with base analysis.

PR #145, opened 2020-11-23
Fixes/calloc

### LINKED ISSUE #135: Problematic handling of `calloc`
While reviewing #132, I realized our current handling of `calloc` (heap-allocation with zeroing out memory) is problematic:

- For `calloc` we currently create an `Array(Blob(\bot, size),1)`, whereas for `malloc` we create just a `Blob(\bot,size)`. This probably stems from the fact that calloc takes `void *calloc(size_t nitems, size_t size)`.
If we want to stick to creating `Array(Blob(\bot,_),_)` it should probably at least be `Array(Blob(\bot,size),nitems)` instead of `Array(Blob(\bot,size*nitems),1)`.

- The more problematic thing is that we are potentially unsound:

~~~C
#include<stdlib.h>
#include<assert.h>
int main(void) {
    int *r = calloc(2,sizeof(int));
    r[0] = 3;
    // We compute (alloc@4) → `Array(`Blob(3,size:8),length:1)

    int z = r[1]; // Due to to a bug we read supertop here

    assert(z == 3);  // If we fix the bug above we would claim that this succeeds, 
                     //  but in the concrete it fails as z is 0 here
}
~~~

This is hidden by a bug that causes us to always read `supertop` from `calloc`ed blobs because we get confused with offsets.

To fix this, probably the `Blob` domain would have to distinguish if something is the result of a calloc call (as opposed to malloc). 

- **If it is, when writing into the blob:**
If the value is ⊥ and we are writing `a` into it (3 in this case), we should make a zero-initialized version of the same type as `a` (`0` in this case) and join a with it (`3 join 0` in this case). Then when we read, we also cover the possibility of reading somewhere where we have not written before. (If nobody has written before, we read `bot` which we then turn into `top` in `eval_rv`)
Ideally, we would have it so the abstract values for the blob would be initialized to zero, but we don't know at allocation which type we will write into the blob, and if we guess wrong there, any other write would immediately yield supertop.

- **For memory that has been malloced:**
We can keep doing what we are currently doing, reading memory that has not been initialized is undefined behavior, so we can only read things someone has explicitly written.

ivanazuzic:
This should fix #135:
- The created Blobs keep track of their creator (so-called `origin`, which can be either `Calloc` or `Malloc`)
- The idea is to know that the Blob created by `calloc`, which has a value `Bot` is actually supposed to be 0, because `calloc` zero-initializes the allocated place
- The type argument was returned to the signature of the `update_offset` function from the value domain because it's necessary to know the type of the zero which is produced by `calloc`
- `Calloc` used to take only one argument `n * size` and now those two are separated so `calloc` takes `n` and `size`
- Tests for `calloc` have been added
- Type was added as parameter to the `eval _offset `function of the value domain to make possible to read 0 when something was allocated by `calloc` and is not initialized yet
- Functions `bot_value`, `init_vaue` and `top_value` have been moved from `base` analysis to `valueDomain`. There a `top_value` function already existed and now, not to keep two `top_value` functions, a combination is kept in the `valueDomain` only (it resembles more to the one from the base analysis, which was more specific about arrays). The functions `is_mutex_type` and `is_immediate_type` have been moved together because they were only used by the functions that now went to the value domain.
- Partition arrays parameter has been added to the `calloc` tests.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2065,7 +2065,7 @@ struct
                                   (eval_lv ctx.ask gs st lv, `Address heap_var)]
         | _ -> st
       end
-    | `Calloc size ->
+    | `Calloc (n, size) ->
       begin match lv with
         | Some lv -> (* array length is set to one, as num*size is done when turning into `Calloc *)
           let heap_var = heap_var ctx in
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -806,9 +806,9 @@ struct
     try
       let r = eval_rv a gs st exp in
       if M.tracing then M.tracel "eval" "eval_rv %a = %a\n" d_exp exp VD.pretty r;
-      if VD.is_bot r then ValueDomain.Compound.top_value2 (typeOf exp) else r
+      if VD.is_bot r then ValueDomain.Compound.top_value (typeOf exp) else r
     with IntDomain.ArithmeticOnIntegerBot _ ->
-    ValueDomain.Compound.top_value2 (typeOf exp)
+    ValueDomain.Compound.top_value (typeOf exp)
 
   (* Evaluate an expression containing only locals. This is needed for smart joining the partitioned arrays where ctx is not accessible. *)
   (* This will yield `Top for expressions containing any access to globals, and does not make use of the query system. *)
@@ -1544,7 +1544,7 @@ struct
 
   let set_savetop ?lval_raw ?rval_raw ask (gs:glob_fun) st adr v : store =
     match v with
-    | `Top -> set ask gs st adr (ValueDomain.Compound.top_value2 (AD.get_type adr)) ?lval_raw ?rval_raw
+    | `Top -> set ask gs st adr (ValueDomain.Compound.top_value (AD.get_type adr)) ?lval_raw ?rval_raw
     | v -> set ask gs st adr v ?lval_raw ?rval_raw
 
 
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2197,7 +2197,7 @@ struct
       | TPtr (t, attr), `Address a
         when (not (AD.is_top a))
           && List.length (AD.to_var_may a) = 1
-          && not (ValueDomain.Compound.is_immediate_type t)
+          && not (VD.is_immediate_type t)
         ->
         let cv = List.hd (AD.to_var_may a) in
         "ref " ^ VD.short 26 (CPA.find cv es)
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -436,7 +436,7 @@ struct
       | `Top ->
         let typ = AD.get_type adr in
         let warning = "Unknown value in " ^ AD.short 40 adr ^ " could be an escaped pointer address!" in
-        if ValueDomain.Compound.is_immediate_type typ then () else M.warn_each warning; empty
+        if VD.is_immediate_type typ then () else M.warn_each warning; empty
       | `Bot -> (*M.debug "A bottom value when computing reachable addresses!";*) empty
       | `Address adrs when AD.is_top adrs ->
         let warning = "Unknown address in " ^ AD.short 40 adr ^ " has escaped." in
@@ -806,7 +806,7 @@ struct
     try
       let r = eval_rv a gs st exp in
       if M.tracing then M.tracel "eval" "eval_rv %a = %a\n" d_exp exp VD.pretty r;
-      if VD.is_bot r then ValueDomain.Compound.top_value (typeOf exp) else r
+      if VD.is_bot r then VD.top_value (typeOf exp) else r
     with IntDomain.ArithmeticOnIntegerBot _ ->
     ValueDomain.Compound.top_value (typeOf exp)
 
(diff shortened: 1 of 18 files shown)


src/analyses/base.ml
```diff
@@ -443,7 +436,7 @@ struct
       | `Top ->
         let typ = AD.get_type adr in
         let warning = "Unknown value in " ^ AD.short 40 adr ^ " could be an escaped pointer address!" in
-        if is_immediate_type typ then () else M.warn_each warning; empty
+        if ValueDomain.Compound.is_immediate_type typ then () else M.warn_each warning; empty
```
michael-schwarz, line 439: Why not VD.is_immediate_type ?

src/analyses/base.ml
```diff
@@ -871,9 +806,9 @@ struct
     try
       let r = eval_rv a gs st exp in
       if M.tracing then M.tracel "eval" "eval_rv %a = %a\n" d_exp exp VD.pretty r;
-      if VD.is_bot r then top_value a gs st (typeOf exp) else r
+      if VD.is_bot r then ValueDomain.Compound.top_value (typeOf exp) else r
```
michael-schwarz, line 809: Use `VD.top_value` here?

src/analyses/base.ml
```diff
@@ -2122,19 +2061,19 @@ struct
...
+    | `Calloc (n, size) ->
       begin match lv with
         | Some lv -> (* array length is set to one, as num*size is done when turning into `Calloc *)
           let heap_var = heap_var ctx in
           let add_null addr =
             if get_bool "exp.malloc.fail"
             then AD.join addr AD.null_ptr (* calloc can fail and return NULL *)
             else addr in
-          set_many ctx.ask gs st [(add_null (AD.from_var heap_var), `Array (CArrays.make (IdxDom.of_int Int64.one) (`Blob (VD.bot (), eval_int ctx.ask gs st size)))); (* TODO why? should be zero-initialized *)
+          set_many ctx.ask gs st [(add_null (AD.from_var heap_var), `Array (CArrays.make (IdxDom.of_int Int64.one) (`Blob (VD.bot (), eval_int ctx.ask gs st size, false)))); (* TODO why? should be zero-initialized *)
```
michael-schwarz, line 2076: Maybe remove the TODO and explain what we now do instead?

tests/regression/02-base/44-malloc_array.c
```diff
@@ -0,0 +1,11 @@
+// PARAM: --set ana.int.interval true --enable exp.partition-arrays.enabled
+#include<stdlib.h>
+#include<assert.h>
+
+int main(void) {
+    int *r = malloc(5 * sizeof(int));
+
+    r[3] = 2;
+
+    assert(r[4] == 2);
```
michael-schwarz, line 10: This should have an explanation why this succeeds, it does not necessarily do so in the concrete.

src/analyses/base.ml
```diff
@@ -1096,6 +1031,9 @@ struct
      * not include the flag. *)
     let update_one_addr (x, offs) (nst, fl, dep): store =
       let cil_offset = Offs.to_cil_offset offs in
+      let t = match t_override with
+        | Some t -> t
+        | None -> Cil.typeOf (Lval(Var x, cil_offset)) in
```
sim642, line 1036: Does this not cause us to have #119 again? (Until that's fixed by conditionally doing things for non-heap/heap variables.)

src/cdomains/valueDomain.ml
```diff
@@ -27,36 +29,48 @@ sig
...
   val size: t -> size
   val invalidate_value: Q.ask -> typ -> t -> t
 end
 
-module Blob (Value: S) (Size: IntDomain.S) =
+module AllocOrigin = IntDomain.MakeBooleans (
+  struct
+    let truename = "Malloc"
+    let falsename = "Calloc"
+  end)
```
sim642, line 56: Is this intended to be the boolean lattice, where `Calloc ⊑ Malloc`, and not a flat lattice?

src/cdomains/valueDomain.ml
```diff
@@ -673,20 +765,45 @@ struct
...
           let l', o' = shift_one_over l o in
-          do_eval_offset ask f (Blobs.value c) ox exp l' o' v
+          let ev = do_eval_offset ask f (Blobs.value c) ox exp l' o' v t in
+          if orig then (* This Blob came from malloc *) 
+            ev 
+          else (* This Blob came from calloc *)
+            if ev = `Bot then
+              zero_init_value t (* This should be zero initialized *)
+            else
+              ev (* This already contains some value *)
```
sim642, line 781: This logic is copy-pasted three times here. Might be better to have a helper function for it.

src/cdomains/valueDomain.ml
```diff
@@ -737,36 +854,53 @@ struct
...
           let l', o' = shift_one_over l o in
-          mu (`Blob (join x (do_update_offset ask x ofs value exp l' o' v), s))
+          let x = if orig then
+            x  (* the origin is malloc *)
+          else 
+            match x with  (* the origin is calloc *)
+              | `Bot ->  (* nothing was assigned before, so in the case of calloc there is an initial 0 *)
+                zero_init_value t
+              | _ -> x
+             in
```
sim642, line 873: The same logic seems repeated also here twice.

src/analyses/base.ml
ivanazuzic: No good reason, it was just more readable to me. I'll use VD in the future because that's the style everywhere.

src/analyses/base.ml
michael-schwarz: Yes, it does.
This PR should probably be only merged once I have a working patch addressing #119.

src/cdomains/valueDomain.ml
ivanazuzic: You are right, they should be incomparable. Now it's a fake lattice of booleans.

src/analyses/base.ml
michael-schwarz: Should work now.

PR #149, opened 2020-11-29
Support for checking NoOverflows property of SV-COMP

michael-schwarz:
This is an experimental first steps towards supporting NoOverflows for SV-COMP.

To do this, we extend Julian's sound arithmetic such that when we invoke the Overflow handling for something of signed type during the verifying phase, we set a bool from true to false. If this bool is true after verifying, the were no overflows during re-evaluation of all right-hand sides, and therefore no overflow.

TODO:

- [x] Test-run (w/ witness verification)
- [x] Investigate unsoundness
- [x] Fix witness generation
- [x] Figure out correct configuration for this property

### DIFF
--- a/conf/svcomp21.json
+++ b/conf/svcomp21.json
@@ -28,6 +28,7 @@
     "partition-arrays": {
       "enabled": true
     },
+    "lower-constants" : false,
     "region-offsets": true,
     "malloc": {
       "wrappers": [
--- a/conf/svcomp21.json
+++ b/conf/svcomp21.json
@@ -28,7 +28,6 @@
     "partition-arrays": {
       "enabled": true
     },
-    "lower-constants" : false,
     "region-offsets": true,
     "malloc": {
       "wrappers": [
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -492,9 +492,13 @@ struct
     | Some (a, b) ->
       if a = b && b = i then `Eq else if Ints_t.compare a i <= 0 && Ints_t.compare i b <=0 then `Top else `Neq
 
+  let set_overflow_flag ik =
+    if Cil.isSigned ik && !GU.in_verifying_stage then
+      Goblintutil.did_overflow := true
+
   let norm ik = function None -> None | Some (x,y) ->
-    if Ints_t.compare x y > 0 then None
-    else if Ints_t.compare (min_int ik) x > 0 || Ints_t.compare (max_int ik) y < 0 then top_of ik
+    if Ints_t.compare x y > 0 then (set_overflow_flag ik; None)
+    else if Ints_t.compare (min_int ik) x > 0 || Ints_t.compare (max_int ik) y < 0 then (set_overflow_flag ik; top_of ik)
     else Some (x,y)
 
   let leq (x:t) (y:t) =
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -497,7 +497,7 @@ struct
       Goblintutil.did_overflow := true
 
   let norm ik = function None -> None | Some (x,y) ->
-    if Ints_t.compare x y > 0 then (set_overflow_flag ik; None)
+    if Ints_t.compare x y > 0 then None
     else if Ints_t.compare (min_int ik) x > 0 || Ints_t.compare (max_int ik) y < 0 then (set_overflow_flag ik; top_of ik)
     else Some (x,y)
 
@@ -611,6 +611,16 @@ struct
       | Some x, Some y -> (try norm ik (of_int ik (f ik x y)) with Division_by_zero -> top_of ik)
       | _              -> top_of ik
 
+  let bitcomp f ik i1 i2 =
+    match is_bot i1, is_bot i2 with
+    | true, true -> bot_of ik
+    | true, _
+    | _   , true -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (short 80 i1) (short 80 i2)))
+    | _ ->
+      match to_int i1, to_int i2 with
+      | Some x, Some y -> (try norm ik (of_int ik (f ik x y)) with Division_by_zero -> top_of ik)
+      | _              -> (set_overflow_flag ik;  top_of ik)
+
   let bitxor = bit (fun _ik -> Ints_t.logxor)
   let bitand = bit (fun _ik -> Ints_t.logand)
   let bitor  = bit (fun _ik -> Ints_t.logor)
@@ -624,8 +634,8 @@ struct
       | _      -> top_of ik
 
   let bitnot = bit1 (fun _ik -> Ints_t.lognot)
-  let shift_right = bit (fun _ik x y -> Ints_t.shift_right x (Ints_t.to_int y))
-  let shift_left  = bit (fun _ik x y -> Ints_t.shift_left  x (Ints_t.to_int y))
+  let shift_right = bitcomp (fun _ik x y -> Ints_t.shift_right x (Ints_t.to_int y))
+  let shift_left  = bitcomp (fun _ik x y -> Ints_t.shift_left  x (Ints_t.to_int y))
 
   let neg ik = function None -> None | Some (x,y) -> norm ik @@ Some (Ints_t.neg y, Ints_t.neg x)
 
--- a/src/witness/svcomp.ml
+++ b/src/witness/svcomp.ml
@@ -6,6 +6,7 @@ struct
   type t =
     | UnreachCall of string
     | NoDataRace
+    | NoOverflow
 
   let of_string s =
     let s = String.strip s in
@@ -14,6 +15,8 @@ struct
       let global_not = Str.matched_group 1 s in
       if global_not = "data-race" then
         NoDataRace
+      else if global_not = "overflow" then
+        NoOverflow
       else
         let call_regex = Str.regexp "call(\\(.*\\)())" in
         if Str.string_match call_regex global_not 0 then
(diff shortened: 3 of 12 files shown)


src/cdomains/intDomain.ml
```diff
@@ -492,9 +492,13 @@ struct
...
       if a = b && b = i then `Eq else if Ints_t.compare a i <= 0 && Ints_t.compare i b <=0 then `Top else `Neq
 
+  let set_overflow_flag ik =
+    if Cil.isSigned ik && !GU.in_verifying_stage then
+      Goblintutil.did_overflow := true
+
   let norm ik = function None -> None | Some (x,y) ->
-    if Ints_t.compare x y > 0 then None
-    else if Ints_t.compare (min_int ik) x > 0 || Ints_t.compare (max_int ik) y < 0 then top_of ik
+    if Ints_t.compare x y > 0 then (set_overflow_flag ik; None)
```
michael-schwarz, line 500: Do we need to set the flag here in fact?

michael-schwarz:
I ran `NoOverflows-Other.set` with a `10s` timeout, and we answer `true` on `24/108` cases where the expected verdit is `true`. At least with this small timeout, there are no unsoudnesses.
For `NoOverflows-Bitvectors.set`, it is `59/127`. We however also have three cases in which we're unsound.

I'll take a look at the cases in which we are unsound and run it with a longer timeout again.

My feeling is that unsoundness there might also translate to unsoundness w.r.t. to reachability.

michael-schwarz:
All three unsoundness cases are due to CIL doing constant propagation and assuming wrap-arounds for ints.

conf/svcomp21.json
```diff
@@ -28,6 +28,7 @@
     "partition-arrays": {
       "enabled": true
     },
+    "lower-constants" : false,
```
michael-schwarz, line 31: This should only be enabled when checking for overflows, otherwise it is helpful

sim642: Instead of putting it into svcomp21 you could toggle the option here:

https://github.com/goblint/analyzer/blob/03c5da2f4fd913fd242ae48e10a96acda4f64ee9/src/witness/witness.ml#L277-L287

It might make more sense to fix the unsoundness in CIL itself, then we wouldn't need such toggle.

michael-schwarz: > Instead of putting it into svcomp21 you could toggle the option here:
I'll try that, it might be too late though, this needs to be set on the call to `initCil()` already.

>It might make more sense to fix the unsoundness in CIL itself, then we wouldn't need such toggle.

It is not really a unsoundness, CIL just assumes that the programs have no undefined behavior. This is once again this issue that what it means to be "sound" for programs with UB is hard to say, as these programs do not have any meaningful semantics from the C perspective.

This option is very helpful, because it allows, e.g. 
```
enum
{
  _ISupper = ((0) < 8 ? ((1 << (0)) << 8) : ((1 << (0)) >> 8)),
  _ISlower = ((1) < 8 ? ((1 << (1)) << 8) : ((1 << (1)) >> 8))
};
```

which, when this option is disabled will fail as CIL tries to generate `if` `else` statements which can not appear at a global scope (this actually causes us to fail with an exception on some of the `BusyBox-NoOverflow` tests when this toggle is on).

So, CIL itself would need to be able to check if a constant expression might overflow and in this case not do the lowering, but keep doing it otherwise. I don't think that it is realistic to get this working for sv-comp.

michael-schwarz:
It appears to be sound now, but now wel crash in a lot of the `BusyBox` cases, because~30 of them have this global:

~~~C
enum
{
  _ISupper = ((0) < 8 ? ((1 << (0)) << 8) : ((1 << (0)) >> 8)),
  _ISlower = ((1) < 8 ? ((1 << (1)) << 8) : ((1 << (1)) >> 8))
};
~~~

For **SoftwareSystems-BusyBox-NoOverflows.set**, and a 10s timeout we have:

```
NONE None, no expected_verdict for no-overflow: 2
NONE timeout, no expected_verdict for no-overflow: 1
UNKNOWN None, expected False: 22
UNKNOWN None, expected True: 24
UNKNOWN timeout, expected False: 12
UNKNOWN timeout, expected True: 13
--------------------------------------------------------------------------------
points: CORRECT{ True: 2, False: 1 }, INCORRECT{ True: -32, False: -16 }
total points: 0 (assuming witnesses are correct)
total points_must: 0 (assuming witnesses are correct, only counting true-unreach (could return UNKNOWN for may-info))
total time: 357.90 s
```

For **NoOverflows-BitVectors.set**, and a 10s timeout we have:

```
CORRECT True: 59
NONE None, no expected_verdict for no-overflow: 25
NONE True, no expected_verdict for no-overflow: 4
NONE timeout, no expected_verdict for no-overflow: 4
UNKNOWN None, expected False: 143
UNKNOWN None, expected True: 43
UNKNOWN timeout, expected False: 15
UNKNOWN timeout, expected True: 20
--------------------------------------------------------------------------------
points: CORRECT{ True: 2, False: 1 }, INCORRECT{ True: -32, False: -16 }
total points: 118 (assuming witnesses are correct)
total points_must: 118 (assuming witnesses are correct, only counting true-unreach (could return UNKNOWN for may-info))
total time: 419.30 s
```

For **NoOverflows-Others.set**, and a 10s timeout we have:

```
CORRECT True: 24
NONE None, no expected_verdict for no-overflow: 44
NONE True, no expected_verdict for no-overflow: 14
NONE timeout, no expected_verdict for no-overflow: 8
UNKNOWN None, expected False: 48
UNKNOWN None, expected True: 30
UNKNOWN timeout, expected False: 24
UNKNOWN timeout, expected True: 46
--------------------------------------------------------------------------------
points: CORRECT{ True: 2, False: 1 }, INCORRECT{ True: -32, False: -16 }
total points: 48 (assuming witnesses are correct)
total points_must: 48 (assuming witnesses are correct, only counting true-unreach (could return UNKNOWN for may-info))
total time: 882.96 s
```

conf/svcomp21.json
sim642: > I'll try that, it might be too late though, this needs to be set on the call to `initCil()` already.

It's being set after `initCIL ()` though already.

sim642:
The crashing on `if` statements in global inits is from our own simple global init logic or from within CIL?

Anyway, this would be yet another reason to move global inits into the constraint system and do them while solving to get the full features, unlike global inits right now where the transfer functions are called manually. I think it's too big of a change to do for SV-COMP.

michael-schwarz:
> The crashing on `if` statements in global inits is from our own simple global init logic or from within CIL?

Cil checks that the translation of a global initializer does not create statements.

```
Frontc is parsing /home/michael/Documents/goblint-cil/analyzer/goblint_temp_30b98f/yes-2.i
Converting CABS->CIL
../../sv-comp/sv-benchmarks/c/busybox-1.22.0/yes-2.i:86: Error: Constant expression tmp has effects
Error on A.ONLYTYPEDEF (Errormsg.Error)
Frontc is parsing /home/michael/Documents/goblint-cil/analyzer/goblint_temp_30b98f/sv-comp.c
Error: There were parsing errors in /home/michael/Documents/goblint-cil/analyzer/goblint_temp_30b98f/sv-comp.c
Fatal error: exception Errormsg.Error
````

> Anyway, this would be yet another reason to move global inits into the constraint system and do them while solving to get the full features, unlike global inits right now where the transfer functions are called manually.

Yes, I guess.

> I think it's too big of a change to do for SV-COMP.

Agreed.

**Comment by sim642** [comment]
> Cil checks that the translation of a global initializer does not create statements.

Oh that `enum` case is also not so trivial as I first throught because there isn't even anywhere to put the `if` statement. I suppose it would have to stay as a ternary operator and Goblint needs to also handle that sort of expression, even if it only ends up appearing in global initializers like that.

File: conf/svcomp21.json (hunk 2 above)
Comment (michael-schwarz) [reply conf/svcomp21.json:31]: True, definitely before CIL parses the file though.
I'll just try and see if works there, otherwise I'll find a different sport for it.

File: src/cdomains/intDomain.ml (hunk 1 above)
Comment (michael-schwarz) [reply src/cdomains/intDomain.ml:500]: Not setting it here causes us to produce correct results for more cases, but we also the become unsound for 3 cases. I'll see what's going on there.

File: src/witness/svcomp.ml [hunk 3]
```diff
@@ -92,6 +96,7 @@ struct
     | False (Some spec) ->
       let result_spec = match spec with
         | UnreachCall _ -> "unreach-call"
+        | NoOverflow -> "overflow"
```
sim642, line 99: This string needs to be `no-overflow` corresponding to the SV-COMP property name. Although it won't really matter right now if we never output `false`.

michael-schwarz: done

src/cdomains/intDomain.ml
michael-schwarz: Setting it here is not necessary, but it needs to be set in `shift` operations when we don't bother to find out the actual result and simply go to top.
This means we answer correctly on one fewer task, but is the right thing to do.

conf/svcomp21.json
michael-schwarz: Ok, we can indeed set it here.
The other question is if we don't want different config files for different properties in general? This would probably be cleaner. Do you know if it is allowed by the rules to have different configs per property?

michael-schwarz: I set this from `Maingoblint.main` now.

michael-schwarz:
The results also stay the same with a `60s` timeout except that for some progams we get a stack overflow.

sim642:
I did a BenchExec run on all NoOverflow categories and got the following results using Ultimate as witness validator:
Screenshot_2020-11-30_12-08-33

It seems to me that the benchmarks we manage to solve are simple enough for Ultimate to also reach the same conclusion without our witness. But that's irrelevant and our witnesses are technically confirmed.

The only unconfirmed witness is for `termination-numeric/LogRecursive.c` where we generate a massive inlined witness for a recursive program. The size of it probably hinders Ultimate more than it helps.

PR #150, opened 2020-11-30
IntervalSet + Assertions

Fayanzar:


michael-schwarz:
For asserts see #157. Interval sets won't be merged for now, probably require a re-implementation after adding `ikind` everywhere.

PR #154, opened 2020-12-09
Proposal: Delete code in *_deprecated modules and commented out code

michael-schwarz:
I think it is time to rid Goblint of some things that have collected over the years, namely:

- Old modules named `*_deprecated`
- Code that is commented out

I'd propose deleting such things.

There probably are things worth keeping among all of this code, but I would argue that for such things, we should:
- Uncomment and make compile with the current state
- Rename to something different than `*_deprecated`

Things that are purely relevant for historic reasons or because one might need something similar again at some point (while not worth the effort to make them compile with the current state), are safely stored in the Git history and can always be brought back.

What do y'all think?


sim642:
> Things that are purely relevant for historic reasons or because one might need something similar again at some point (while not worth the effort to make them compile with the current state), are safely stored in the Git history and can always be brought back.

I'd say the main difficulty with that is not being able to know which things have existed in the past but have since then been deleted. One needs to know that a particular thing existed at some point to know to even try to find it through the git history using some search over the entire history.

Nevertheless, I think it's a good idea to remove such things. It might be easier to reimplement something than adapt an old commented out solution since so many other things could've been refactored since.

Issue #148, opened 2020-11-27
Make threadflag/thread analysis path-sensitive


sim642:
The thread join support from PR #137 didn't improve our results on SV-COMP's NoDataRace benchmarks as much as expected. At least some of them create and join threads conditionally. This currently renders the added thread join support useless because joins in the thread analyses lose the necessary precision. Making threadflag and maybe even thread analysis path-sensitive would help with that.

I made a quick experiment but turns out this is problematic and cannot currently easily be done. The issue is the `S.context d` call here:
https://github.com/goblint/analyzer/blob/9fd3ccc9d19cabaf8bc139d00afe3d62b36803bf/src/framework/constraints.ml#L566-L579

If `d` is a path-sensitivity domain set that happens to contain more than one path, then `S.context` fails because it can just return a single context. It might be necessary to rework this mechanism to be more like our previous function call mechanism where `enter` returns a list, not a single element, which is then handled in `FromSpec` to apply `S.context` per-element.
(That in itself is a weird mechanism because the path-sensitivity handling isn't localized to the corresponding lifter but throughout the analysis specification hierarchy.)

I have my doubts about being able to achieve this by SV-COMP 2021.

sim642:
As I said in #155, there's an experimental fix in [`thread/path-sens`](https://github.com/goblint/analyzer/tree/thread/path-sens) branch.

sim642:
Key parts of the experiment have now been cherry-picked to `traces` branch since the `threadenter` path-sensitivity crash is also possible with mutex paths. The SV-COMP examples and configuration wasn't cherry-picked.

sim642:
Since the implementation parts already reached `master` via #183, then I've also cherry-picked the rest of `thread/path-sens`, which is the regression tests for this. Based on these, the feature works, so I'm closing the issue.

PR #155, opened 2020-12-15
Fix unsound query in path-sensitivity functor

sim642:
While thinking about #148 (which has an experimental fix in the [`thread/path-sens`](https://github.com/goblint/analyzer/tree/thread/path-sens) branch) and comparing the mechanism with normal function call mechanism, I discovered the following unsoundness bug.

If different elements of the path-sensitive set (e.g. made path-sensitive by different mutex sets) have different points-to sets for a function pointer address, then these were previously combined with `meet` in `PathSensitive2` when `FromSpec` uses the query from outside to determine possible functions:
https://github.com/goblint/analyzer/blob/e2a0555f8c30fdb7140cac1258e66b23c93671b5/src/framework/constraints.ml#L676-L679
So the intersection became empty and no functions were called at all! The right thing to do would be to join queries over all paths instead because paths are disjunct unlike multiple analyses which are conjunct.

This obviously isn't as precise as it could be because it forgets which functions may get called from which paths but there's no easy way around that right now. This is due to all the logic being on the top level at `FromSpec` instead of being somehow within `PathSensitive2` where each path could evaluate the function variable separately and handle them only along that path. But that's another story altogether...

### DIFF
--- a/src/analyses/mCP.ml
+++ b/src/analyses/mCP.ml
@@ -468,6 +468,7 @@ struct
         ; assign = (fun ?name _ -> failwith "Cannot \"assign\" in query context.")
         }
       in
+      (* meet results so that precision from all analyses is combined *)
       Queries.Result.meet a @@ S.query ctx' q
     in
     match q with
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -1034,7 +1034,7 @@ struct
     fold' ctx Spec.sync identity (fun (a,b) (a',b') -> D.add a' a, b'@b) (D.empty (), [])
 
   let query ctx q =
-    fold' ctx Spec.query identity (fun x f -> Queries.Result.meet x (f q)) `Top
+    fold' ctx Spec.query identity (fun x f -> Queries.Result.join x (f q)) `Bot
 
   let enter ctx l f a =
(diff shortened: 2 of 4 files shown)


michael-schwarz:
I was thinking if this causes any issues with analyses that can not answer some type of query at all and hence return `Top`. But since we still meet in `MCP2` and only do the join in the functor(s) that add path-sensitivity, this is fine: If the result is `Top` along some path, then ask (which needs to be path-insensitive here) should also return `Top`.

Maybe add some comment on why we `join` here instead of `meet` that we do everywhere else?

**So, LGTM!** (Really fascinating that such bugs can be hidden for such a long time.)

sim642:
> Maybe add some comment on why we `join` here instead of `meet` that we do everywhere else?

The "everywhere else" for `meet` is really just `MCP2` AFAIK, but I added comments to both now.

> (Really fascinating that such bugs can be hidden for such a long time.)

This isn't so surprising because `FromSpec` doing the `EvalFunvar` query from outside of `PathSensitive` is basically the only query done at such high level (except also the witness generation ones now, but they have a specific implementation in `PathSensitive`). Everything else goes only through `MCP2` and the activated analyses. And I guess conditional locking and their dependent function pointers aren't really used in practice, so this hasn't shown.

PR #159, opened 2021-01-18
Port the DefExc Domain to use BigInts

jerhard:
Use `BigInt` in the `DefExc` domain. 

Also add the option `ana.int.wrap_on_signed_overflow` to assume wrap-around behavior for signed integers according to the two-complement representation. An example where this may be helpful is contained in `tests/regression/01-cpa/10-posneg.c`.


tests/regression/01-cpa/10-posneg.c
```diff
@@ -1,3 +1,4 @@
+//PARAM: --enable ana.int.wrap_on_signed_overflow
```
michael-schwarz, line 1: Maybe add a comment on why this is needed here?

src/cdomains/intDomain.ml
```diff
@@ -1244,91 +1312,126 @@ struct
...
+    shift BigInt.shift_right
   (* TODO: lift does not treat Not {0} as true. *)
-  let logand = lift2 Integers.logand
-  let logor  = lift2 Integers.logor
-  let lognot = eq (of_int 0L)
+  let logand = lift2 BigInt.logand
+  let logor  = lift2 BigInt.logor
+  let lognot ik = eq ik (of_int ik BigInt.zero)
+
+  (* let invariant _ _ = failwith "invariant not implemented for DefExc. Use invariant_ikind intead." *)
```
michael-schwarz, line 1426: Maybe uncomment or remove?

tests/regression/01-cpa/10-posneg.c
jerhard: I added a comment to the test case.

src/cdomains/intDomain.ml
jerhard: I removed the commented out line.

src/cdomains/intDomain.ml
```diff
@@ -1062,43 +1100,78 @@ struct
...
+     *)
+    let norm ik (t :t) =
+      let should_wrap ik = not (Cil.isSigned ik) || GobConfig.get_bool "ana.int.wrap_on_signed_overflow" in
+      match t with
+      | `Excluded (s, r) ->
+        let r = if R.leq (size ik) r then size ik else r in
+        let min, max = range_min_max r in
+        if should_wrap ik then (
+          (* Perform a wrap-around for unsigned values and for signed values (if configured). *)
+          S.iter (fun v -> print_endline (BI.to_string v)) s;
```
michael-schwarz, line 1128: Is this debug code that was left over?

jerhard: Yes, it was. I removed it now.

michael-schwarz:
I started a run of this on the `sv-comp` benchmarks, unless we see anything strange, I think we can merge this!

jerhard:
I just realized that `01/10` now works even when `ana.int.wrap_on_signed_overflow`, and it should not.
So for exclusion sets representing a  signed integer, I need to add a check whether an overflow might have occurred (based on the range) and go to top in case `wrap_on_signed_overflow` is not activated.

jerhard:
>I just realized that 01/10 now works even when ana.int.wrap_on_signed_overflow, and it should not.
So for exclusion sets representing a signed integer, I need to add a check whether an overflow might have occurred (based on the range) and go to top in case wrap_on_signed_overflow is not activated.

This should be fixed for the implementation of `DefExc` with 4acbc8b.

michael-schwarz:
> This should be fixed for the implementation of `DefExc` with [4acbc8b](https://github.com/goblint/analyzer/commit/4acbc8bdf5befcb94cf096713f587b3490c67f8c).

Ok, I"ll restart the run on the server then.

michael-schwarz:
Nothing out of the ordinary in the sv-comp run,let's merge it

PR #162, opened 2021-01-19
Replace part_access transfer function with PartAccess query

sim642:
This PR makes the following changes:
1. **Replaces the `part_access` transfer function with `PartAccess` query.**
    This simplifies the `Spec` signature and allows removing dummy `part_access` implementations from `Spec` lifting functors. Especially as anything outside of `MCP2` doesn't influence access partitioning anyway. The `meet` of the new query result is chosen to coincide with the old `part_access` combination logic in `MCP2`, giving exactly the same behavior.
2. **Moves access handling out of `MCP2`.**
    It doesn't need to be there at all and mutex analysis is a suitable place for it anyway, considering it is the one to invoke the `Access` queries in the first place.
3. **Removes `Access` query.**
    Since mutex analysis can directly partition accesses using `PartAccess` query, the `Access` query, which was previously used to signal `MCP2`, is now unnecessary.

It could be also argued that the whole accessing logic doesn't really belong to mutex analysis either and could be extracted out to a separate analysis, but I thought I won't go that far immediately.

### DIFF
--- a/src/domains/access.ml
+++ b/src/domains/access.ml
@@ -60,7 +60,11 @@ struct
   let is_bot x = cardinal x = 1 && LSSet.is_empty (choose x)
   (* top & is_top come from SetDomain.Make *)
 
+  (* Since Queries.PartAccess and PartAccessResult are only used within MCP2,
+     these operations are never really called. *)
   let leq _ _ = raise (Lattice.Unsupported "LSSSet.leq")
+  (* meet (i.e. join in PartAccessResult) for PathSensitive query joining
+     isn't needed, because accesses are handled only within MCP2. *)
   let meet _ _ = raise (Lattice.Unsupported "LSSSet.meet")
   let widen _ _ = raise (Lattice.Unsupported "LSSSet.widen")
   let narrow _ _ = raise (Lattice.Unsupported "LSSSet.narrow")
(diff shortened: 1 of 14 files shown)


src/domains/access.ml
```diff
@@ -49,7 +49,25 @@ struct
...
+  (* TODO: is this actually some partition domain? *)
+  let join po pd =
+    let mult_po s = union (map (LSSet.union s) po) in
+    fold mult_po pd (empty ())
+  let bot () = singleton (LSSet.empty ())
+  let is_bot x = cardinal x = 1 && LSSet.is_empty (choose x)
+  (* top & is_top come from SetDomain.Make *)
+
+  let leq _ _ = raise (Lattice.Unsupported "LSSSet.leq")
+  let meet _ _ = raise (Lattice.Unsupported "LSSSet.meet")
```
michael-schwarz, line 64: Can this not happen as a result of having two different paths when path sensitivity is enabled?
Then the query results are `joined` i.e. `meet` is called for a reversed lattice.
If this can not happen, we should have a comment here why it can't.

sim642: `meet` (i.e. `join` of `PartAccessResult`) isn't really needed because these queries are only happening in `MCP2` and never go up to the path sensitivity functor level, i.e. `Queries.PartAccess` isn't ever made from outside. Just like `part_access` wasn't previously used or properly implemented for the `Spec` functors.

I also added that as a comment to the code.

michael-schwarz:
LGTM!

Issue #165, opened 2021-02-23
Fixpoint not reached when marshaling solution back in


michael-schwarz:
@keremc discovered that there are issues where after marshaling old results back in, Goblint fails to verify that the computed solution is in fact a post solution (despite having verified this before writing it out).

It seems to be somewhat related to arrays, these were small cases in which I was able to reproduce the problem:

~~~C
int main(void) {
  int arr[20];

  for(int i = 0; i < 20; i++)
  {
    arr[i] = 42;
  }
}
~~~

The messages are for example:

```
Fixpoint not reached at node 9 "i ++;" on 06-interprocedural.c:5 (tests/regression/22-partitioned_arrays/06-interprocedural.c:5)
 Solver computed:
 {[Reversed (Topped Set (Normal Lvals * booleans)):{},
   Topped Set (variables):{}, MT mode:Singlethreaded, Thread:main,
   value domain * array partitioning deps:(mapping {
                                             Local {
                                               arr -> (Array: Array (part. by i): (intervals:[42,42] -- intervals:[42,42] -- intervals:[-2147483648,2147483647]), ([20,20]))
                                               i -> ([0,19])
                                             }
                                           }, mapping {
                                                  Local {
                                                    i -> {arr}
                                                  }
                                                }),
   lifted proglines:Unknown line, Unit:()]}
 Right-Hand-Side:
 {[Reversed (Topped Set (Normal Lvals * booleans)):{},
   Topped Set (variables):{}, MT mode:Singlethreaded, Thread:main,
   value domain * array partitioning deps:(mapping {
                                             Local {
                                               arr -> (Array: Array (part. by i): (intervals:[42,42] -- intervals:[-2147483648,2147483647] -- intervals:[-2147483648,2147483647]), ([20,20]))
                                               i -> ([0,19])
                                             }
                                           }, mapping {
                                                  Local {
                                                    i -> {arr}
                                                  }
                                                }),
   lifted proglines:Unknown line, Unit:()]}
 Difference: Map: arr =
 partitioned array: Array: Array (part. by i): (intervals:[42,42] -- intervals:[-2147483648,2147483647] -- intervals:[-2147483648,2147483647]) not leq Array: Array (part. by i): (intervals:[42,42] -- intervals:[42,42] -- intervals:[-2147483648,2147483647]). 

```

[...]

PR #166, opened 2021-02-24
Fix Hashconsing & Marshaling of Results

michael-schwarz:
There are two issues with verifying results that are marshaled back in:

### Hashconsing-related issues
- TBD
- Fix likely similar to what @jerhard did for TD3

### Issues with Physical Equality on `varinfo` 

**These issues are solved via the commits I cherry-picked onto the master branch!**
- Because the result that was marshaled back in and the data CIL generated when parsing are different, they are not physically equal (e.g. `varinfo`s will have the same `vid` but not compare equal under `==`)
-  For a reason I don't really understand, CIL uses `==` in `Expcompare.compareLval` which the partitioned arrays use (via the `ExpRelation` analysis) **NB: Structural Equality would also not be solution here, as the `vtype` might be cyclic.**
https://github.com/goblint/cil/blob/5c21f059ede5e3af4374d7c596745a668b94d14f/src/expcompare.ml#L81
- Goblint has its own `Basetype.CilExp.compareExp` (that funnily enough I added but then did not use), that does not have this issue.
- I now replaced all uses of `Expcompare.compareLval` with `Basetype.CilExp.compareExp`.

There might be other places where CIL makes this assumption that `varinfo`s for the same variable compare physically equal fails, but since they would cause failing to verify results that are marshaled back in, I think we should just wait and see if such issues surface.

**References #165**


jerhard:
I think the issue that causes the remaining problems with hashconsing has to with the fact that we wrap a `HashconsContextLifter` in a `HashconsLifter`: 
https://github.com/goblint/analyzer/blob/6ca684670346c59e79c927ee4160dd479f0867bf/src/framework/control.ml#L12-L30

The `HashconsContextLifter` provides hashconsing on contexts (`S.C`); but when the result is wrapped in a `HashconsLifter` again, we set the `relift` function for the resulting `S.C` to be the identity function:
https://github.com/goblint/analyzer/blob/6ca684670346c59e79c927ee4160dd479f0867bf/src/framework/constraints.ml#L12-L21
since `Printable.HC.relift` is defined as follows:
https://github.com/goblint/analyzer/blob/6ca684670346c59e79c927ee4160dd479f0867bf/src/domains/printable.ml#L42-L45

jerhard:
One option to fix this would be to combine the functionality of `HashconsLifter` and `HashconsContextLifter`. 

Another option would be to expect a `SpecHC` in the `HashconsLifter`, so we can reuse the `relift` function of the passed `S.C`.
Probably several of the involved functors would have to be changed a bit for that to work.

michael-schwarz:
> One option to fix this would be to combine the functionality of HashconsLifter and HashconsContextLifter.

Apparently this used to be one Functor until it was split into two in https://github.com/goblint/analyzer/commit/aff80b16aeacad4ce8f773e752dc018434c0fd98.
Do you recall why  it was neccessary to split them @sim642?

michael-schwarz:
It also is indeed this issue, I was able to verify it works (on this separate branch https://github.com/goblint/analyzer/tree/fixes/old_hashcons), to not clutter history here if this splitting is in fact needed. Marshaling now works and also all tests still pass.

sim642:
> Apparently this used to be one Functor until it was split into two in [aff80b1](https://github.com/goblint/analyzer/commit/aff80b16aeacad4ce8f773e752dc018434c0fd98).
> Do you recall why it was neccessary to split them @sim642?

I used that for some kind of optimization in the next commit: https://github.com/goblint/analyzer/commit/d704c057d8e0b24389e3555ad3322584ef2e4666, where I wanted to just move the context hashconsing to the other side of path-sensitivity.

But it looks like they don't really need to be separated any more, because for witness generation both are applied inside as well:
https://github.com/goblint/analyzer/blob/6ca684670346c59e79c927ee4160dd479f0867bf/src/framework/control.ml#L20-L22.
So the SV-COMP conditional hashconsing should still be here: https://github.com/goblint/analyzer/commit/967279cb36d23251df7b71fce35091862dc9a107#diff-6463fdbeebd7b3b5d34874cea9c6c84b32977d3596ffcf601463c5a0f4be2b6dL20-L22.

sim642:
But the fact that the correct behavior of relifting relies crucially on that specific positioning of hashconsing in the `Spec` lifting still worries me, because any change or use of hashconsing anywhere else might still break marshalling silently. It might even break any SV-COMP run marshalling because relifting isn't propagated inwards to it.

michael-schwarz:
Does it make to apply hashconsing anywhere but on the top level anyway?
Does that not lead to the desired memory usage savings already? Or does the
witness lifter generate so many common things that are then not shared
among top level domain elements?

sim642:
As the current comment on top of `control.ml` also says, the witness functor mainly needs `tag`s from hashconsing to uniquely and without conflicts identify path-sensitivity set elements as integers, which is the only way to pass that through queries to the witness generation completely outside of the whole `Spec`.
I suppose the same can be achieved by implementing some `Tagged` functor there instead, which would construct a similar hashtable as hashconsing but not change the element type (so no `hobj` which wants to rely on physical equality). A naive implementation of that wouldn't help though:
1. The hashtable inside `Tagged` would need to use _weak reference arrays_, just like `BatHashcons` does to not prevent GC. Without that, there would be a massive memory leak.
2. It wouldn't have the normal benefits of hashconsing "for free".

Now thinking about it, it might even be useful to have each path element hashconsed (even outside SV-COMP) because that would more-or-less be a more generic solution to #164, where hashconsed path elements would avoid expensive `equal` and `hash` operations on the inner domain values for any analysis, not just the base analysis's `CPA`.

Even if that isn't the case, #164 still makes a compelling argument for some analysis possibly using hashconsing within itself. The current solution in #164 isn't hashconsing (but just lazy `hash`), but this is making me reconsider now.

A more broader conclusion I've drawn already from my SV-COMP-saga is that path-sensitivity set elements are just as important on their own as elements of the entire `Spec` domain. The implementation treats them as an afterthought, trying to treat things outside and inside of `PathSensitive` very uniformly, while they are not.

michael-schwarz:
I cherry-picked the non-hashcons related commits onto master, so we can focus on fixing hashcons here.

michael-schwarz:
This very stupid change seems to fix the problems, I verified that hashconsing is indeed called and that the debug assertions in hashcons do not fail. But this for some reason seems to be an overly simple fix, which genuinely surprises me. So would be good if someone else could also check that it is the right thing to do.

michael-schwarz:
Ok, looks like  it does not work nested applications of the ContextHashconsLifter yet

michael-schwarz:
I ran the `coreutils` tests, it makes no difference performance-wise. We'll run the tests to see if the solution can be verified after marshaling back in past solutions for all of the regression tests later this week.
If that works, we should be able to merge.

michael-schwarz:
So for our regression tests, marshaling back in does sometimes fail because polymorphic compare gets called on cyclic data structures. This was already the case before this change though when marshaling in past results:

To reproduce:

```
$ ./regtest.sh 05 17 --sets save_run run
$ ./regtest.sh 05 17 --conf run/config.json --sets load_run run -v
Raised by primitive operation at Dune__exe__Printable.Std.compare in file "src/domains/printable.ml", line 43, characters 16-23
Called from BatSet.Concrete.split in file "src/batSet.ml", line 261, characters 14-21
Called from BatSet.Concrete.diff in file "src/batSet.ml", line 488, characters 12-29
Called from BatSet.Make.diff in file "src/batSet.ml" (inlined), line 684, characters 14-71
Called from Dune__exe__SetDomain.ToppedSet.diff in file "src/domains/setDomain.ml", line 261, characters 26-38
Called from Dune__exe__MCP.MCP2.special.f in file "src/analyses/mCP.ml", line 727, characters 17-37
Called from Dune__exe__MCP.MCP2.map_deadcode.one_el in file "src/analyses/mCP.ml", line 298, characters 46-72
Called from Stdlib__list.fold_left in file "list.ml", line 121, characters 24-34
Called from Dune__exe__MCP.MCP2.map_deadcode in file "src/analyses/mCP.ml", line 299, characters 13-35
Called from Dune__exe__MCP.MCP2.special in file "src/analyses/mCP.ml", line 729, characters 15-52
Called from Dune__exe__Constraints.PathSensitive2.map.h in file "src/framework/constraints.ml", line 977, characters 16-36
Called from BatSet.Concrete.fold in file "src/batSet.ml", line 202, characters 35-56
Called from Dune__exe__Constraints.PathSensitive2.map in file "src/framework/constraints.ml", line 980, characters 12-43
Called from Dune__exe__Constraints.DeadCodeLifter.lift_fun in file "src/framework/constraints.ml", line 470, characters 13-29
Called from Dune__exe__Constraints.HashconsLifter.special in file "src/framework/constraints.ml", line 78, characters 14-43
Called from BatList.map in file "src/batList.mlv", line 236, characters 23-28
Called from Dune__exe__Constraints.FromSpec.tf_proc in file "src/framework/constraints.ml", line 644, characters 17-48
Called from Dune__exe__Constraints.FromSpec.tf in file "src/framework/constraints.ml", line 659, characters 26-60
Called from Dune__exe__Constraints.FromSpec.tf in file "src/framework/constraints.ml", line 656, characters 4-605
Called from BatList.fold_left2 in file "src/batList.mlv", line 657, characters 39-54
Called from Dune__exe__Constraints.FromSpec.tf in file "src/framework/constraints.ml", line 681, characters 4-107
Called from Dune__exe__Constraints.Verify2.verify.verify_var.verify_constraint in file "src/framework/constraints.ml", line 1198, characters 16-56
Called from Stdlib__list.iter in file "list.ml", line 110, characters 12-15
Called from Stdlib__hashtbl.iter.do_bucket in file "hashtbl.ml", line 155, characters 8-18
Called from Stdlib__hashtbl.iter in file "hashtbl.ml", line 161, characters 6-21
Re-raised at Stdlib__hashtbl.iter in file "hashtbl.ml", line 166, characters 4-13
Called from Dune__exe__Constraints.Verify2.verify in file "src/framework/constraints.ml", line 1206, characters 4-28
Called from Stats.repeattime.repeatf in file "src/ocamlutil/stats.ml", line 106, characters 10-15
Re-raised at Stats.repeattime.repeatf in file "src/ocamlutil/stats.ml", line 110, characters 1-8
Called from Dune__exe__Control.AnalyzeCFG.analyze.solve_and_postprocess in file "src/framework/control.ml", line 429, characters 8-48
Called from Dune__exe__Goblintutil.timeout in file "src/util/goblintutil.ml", line 378, characters 12-17
Called from Dune__exe__Control.AnalyzeCFG.analyze in file "src/framework/control.ml", line 513, characters 17-155
Called from Dune__exe__Control.AnalyzeCFG.analyze_loop in file "src/framework/control.ml", line 541, characters 6-33
Called from Dune__exe__Maingoblint.do_analyze.do_all_phases.do_one_phase in file "src/maingoblint.ml", line 311, characters 12-48
```

michael-schwarz:
The cases where marshalling still fails are the `OSEK2` (?) analysis in `osektransactionality.ml` that stores some things in a hashtable inside the analysis module which is of course not marshalled.

michael-schwarz:
**Ok, I think this is ready to for review & merging now:**

Changes are:
- Include an option `-m` to `update_bench.rb` to test the marshaling of results (except for tests annotated with `NOMARSHAL`: Currently tests for OSEK things and one ARINC test) and test this in CI
- Replace incorrect usage of `==` to compare things that need not be physically equal after marshaling in some analyses
- Remove incorrect usages of `Expcompare.compareExp`
- Remove polymorphic compare from `Printable.Std` and `Printable.Blank` and provide modules (`Printable.StdPolyCompare` and `Printable.BlankPolyCompare`) that supply it. The idea is that polymorphic compare should not be included by default, but including it becomes an active choice (if one is convinced that there is no cyclic data)
- Correct relifting (entails removing `SpecHC` and adding relift to normal `Spec`)
- Unrelated: Print diff for globals too when verify fails
- Fixes to Octagon that abused a function named `comapre` to compare only some elements of a tuple
- Set `name` to `lvalMapDomain` in `lvalMapDomain.ml`

src/analyses/uninit.ml
```diff
@@ -108,7 +108,7 @@ struct
     let rec is_offs_prefix_of pr os =
       match (pr, os) with
       | (`NoOffset, _) -> true
-      | (`Field (f1, o1), `Field (f2,o2)) -> f1 == f2 && is_offs_prefix_of o1 o2
+      | (`Field (f1, o1), `Field (f2,o2)) -> f1.fcomp.ckey = f2.fcomp.ckey && f1.fname = f2.fname && is_offs_prefix_of o1 o2
```
sim642, line 111: Here fields are compared both by `fcomp.ckey` and `fname`.

src/cdomains/lval.ml
```diff
@@ -533,7 +533,7 @@ struct
       match a,b with
       | `NoOffset , `NoOffset -> true
       | `Field (f1,o1), `Field (f2,o2) when f1.fname = f2.fname -> eq o1 o2
```
sim642, line 535: Here fields are compared just by `fname`.

Maybe we should have our own `compare`/`equal` functions for comparing fields in an uniform way. Not sure if there is any module for fields where this could be.

src/domains/lattice.ml
```diff
@@ -439,6 +439,13 @@ module LiftBot (Base : S) =
 struct
   include Printable.LiftBot (Base)
 
+  let compare x y =
+    match x,y with
+    | `Bot,`Bot -> 0
+    | (`Bot, _) -> -1
+    | (_, `Bot) -> 1
+    | (`Lifted x, `Lifted y) -> Base.compare x y
+
```
sim642, line 448: Why is this in `Lattice.LiftBot` as opposed to `Printable.LiftBot`?

src/cdomains/lval.ml
michael-schwarz: In both cases, the check for `fcomp.ckey` is not needed as this will be true because the variable ids compare equal. I removed the check in the other case too.

sim642 (approved):
I guess now you just need to resolve the conflicts :smile:

michael-schwarz:
The two tests that now fail for some strange reason only fail when running them via the ruby script, but not if I manually marshal out data and marshal it back in. :confused:

sim642:
The two tests are:
1. 00/14 sanity/startstate,
2. 04/48 mutex/assign_spawn.

I haven't looked at them but both sound like they could be related to some #183 changes that might give ideas what to look at:
1. `startstate` might have something to do with `entrystate_global` etc. Maybe missing `relift`?
2. `assign_spawn` is related to the changes regarding `ctx.spawn`, `threadenter` and `threadspawn`. Not sure what the problem there could be though.

michael-schwarz:
I don't really think it is related to #183 as the two failing tests work if I marshal the results out and read them again on the command line, and only fail when running through the regression test script.

What both failing tests have in common is that they are the only tests to do
~~~C
#include <linux/device.h>
~~~

This leads me to guess that it has something to do with the includes.

sim642:
I just looked at other kernel tests and they seem to use `mainfun`, but these two tests don't. If I add `--set mainfun[+] "'start'"` to startstate and `--set mainfun[+] "'misc_init'"` to assign_spawn, then the marshaling tests seem to pass with the script.

michael-schwarz:
> I just looked at other kernel tests and they seem to use mainfun, but these two tests don't. If I add --set mainfun[+] "'start'" to startstate and --set mainfun[+] "'misc_init'" to assign_spawn, then the marshaling tests seem to pass with the script.

Ah, nice catch! Still strikes me as odd why this is not needed when calling the tests directly?

sim642:
But there is also code that's supposed to add the `__init` attributed function as `mainfun` without explicitly having to list it in the options:
https://github.com/goblint/analyzer/blob/c398773746841ff7e51525f4911c831368b626ce/src/util/cilfacade.ml#L205-L206

Maybe something about that is flawed? Like why does it even do `set_string "mainfun[+]" mn;` when it calls `add_main def acc` anyway?

sim642:
Also I managed to reproduce the marshaled verification failure by both saving and loading the run from command line manually. Did you maybe forget the `--conf run/config.json` while loading?

I'm not sure why the script would have to do that though since it's passing the same options to Goblint anyway.

Maybe the problem is that when saving, the `mainfun` is added via the special kernel case, which changes the config, which is then written to `run/config.json`. When loading, the `mainfun` is added explicitly via the option, not the special kernel case. I'm not sure why this would cause a difference, but maybe that's mixing up some `fundec` records.

michael-schwarz:
Yes, I just found out that if I remove `--conf run/config.json` it works for the benchmark script, and I had indeed forgotten to pass that option on the command line as I had somehow assumed it was subsumed by `load_run`.

michael-schwarz:
It is somehow broken by memoization of options 5c7300840df30ef25baa1e4ef53b4d2ad55b3302. If I revert that, it works again.

sim642:
That seems the more likely reason indeed. I ended up looking at `Cilfacade.callConstructors` and `addConstructors`, which run before `Cilfacade.getFuns` and doesn't handle these kernel `__init` functions. So there's also a difference whether a kernel `__init` function is listed explicitly as `mainfun` or added later to the list by `getFuns`. In this case there are no constructors though, so that shouldn't be the fault.

(On another note, these "constructors" are yet another maniacal form of global initialization that we handle in yet another weird way: by adding them as calls into the main function to be performed during the solving, not during our global inits. Another thing for #178...)

sim642:
I just noticed that the regression testing script doesn't clean up the marshaled files. There's now a run subdirectory in each regression test group directory.

michael-schwarz:
`BatCache` does not allow for resetting the entire cache, one would have to manually delete all items.

It seems like it is a poor match for what we are trying to do here anyway. 
> This cache gives access to some internals of a memoized function f, called the generating function. **This function is usually pure**; always returning the same output for a given input. This module allows the results of complex functions to be cached so that the function does not need to be called again to get that result.

(from the [doc](https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatCache.html), emphasis mine)

michael-schwarz:
> I just noticed that the regression testing script doesn't clean up the marshaled files. There's now a run subdirectory in each regression test group directory.

Yes, which is ignored via the `.gitgnore`. I could also change it to remove them if we want to.

sim642:
> Yes, which is ignored via the `.gitgnore`. I could also change it to remove them if we want to.

They aren't though, that's why I noticed them existing at all. Maybe you didn't commit the `.gitignore` change?

Either way, I think it'd be nicer if running the tests didn't leave temporary files in the middle of the test suite.

sim642:
> `BatCache` does not allow for resetting the entire cache, one would have to manually delete all items.

Using `.enum` and `.del`-ing each one probably wouldn't be too bad, since it'd only have to happen once after loading a new `--conf`.

> It seems like it is a poor match for what we are trying to do here anyway.

Since the `set_*` functions do properly invalidate the cache, it shouldn't really be a problem. The only missed means of modifying `json_conf` is the missed configuration merging, that manipulates it directly.

What is poor is how we internally also want to modify the options and it's not clearly separate into two phases: loading-setting the options and reading them during the analysis, when they shouldn't change any more. If it were so clear-cut, the caching could be applied then. But as this kernel `__init` `mainfun` modification shows, we sometimes change things after the obvious point (e.g. `AfterConfig.run ()`).

michael-schwarz:
I now added a function to drop memoization when a new json config is loaded, but that does not seem to be enough. :thinking:

sim642:
Maybe deleting during iteration of its contents breaks things? Or is the forcing supposed to work around that? Maybe converting it to an explicit list, which is decoupled from the original structure and its enumeration would behave differently?

michael-schwarz:
No, I added asserts, so this is not the problem.

~~~OCaml
    assert (BatEnum.count (memo_int.enum ()) = 0);
    assert (BatEnum.count (memo_bool.enum ()) = 0);
    assert (BatEnum.count (memo_string.enum ()) = 0);
    assert (BatEnum.count (memo_list.enum ()) = 0)
~~~

vogler:
> > `BatCache` does not allow for resetting the entire cache, one would have to manually delete all items.
> 
> Using `.enum` and `.del`-ing each one probably wouldn't be too bad, since it'd only have to happen once after loading a new `--conf`.
> 
> > It seems like it is a poor match for what we are trying to do here anyway.
> 
> Since the `set_*` functions do properly invalidate the cache, it shouldn't really be a problem. The only missed means of modifying `json_conf` is the missed configuration merging, that manipulates it directly.
> 
> What is poor is how we internally also want to modify the options and it's not clearly separate into two phases: loading-setting the options and reading them during the analysis, when they shouldn't change any more. If it were so clear-cut, the caching could be applied then. But as this kernel `__init` `mainfun` modification shows, we sometimes change things after the obvious point (e.g. `AfterConfig.run ()`).

Agree. There's no better match: either your function is pure or you have to invalidate the cache.
All these `set_` calls should be moved into some `init` function which also does the merging. Or maybe call some function to seal the config and throw an exception if trying to set afterwards.
After this the config should be immutable. Being able to mutate it at any point seems like a source for hard-to-find bugs.
If you really need to mutate, use a reference - we have a lot of those anyway 😄

michael-schwarz:
I guess the immediate issue here seems to be that `set_string` does not account for modification of arrays via `[+]`, so I'll try to fix that for now.

michael-schwarz:
Ok, now I drop all memoization whenever some setting is modified, except for during the phase where they are set to their default values `!GU.build_config = true`. This should be fairly cheap, as we usually do not modify settings exhaustively, and makes the code less fragile and complicated than having special handling of `[+]`.

PR #167, opened 2021-03-04
Make poly analysis compile again

ivanazuzic:
The _ApronDomain_ was missing the following functions: _to_yojson_,  _invariant_, _tag_ and _arbitrary_, so they are added. They don't do anything at the moment, but they are necessary for _ApronDomain_ to compile. 

I added some comments to _ApronDomain_.

_Poly_ analysis was using _EvalIntSet_ and _EvalInterval_ which don't exist, so I commented them such that Poly can be compiled and used.


goblint.opam
```diff
@@ -19,7 +19,6 @@ depends: [
   "goblint-cil"
   "batteries"
   "qcheck"
-  # "z3"
```
michael-schwarz, line 22: Can you please revert this change?

src/analyses/poly.ml
```diff
@@ -113,15 +113,15 @@ struct
         | Some i -> `Int i
         | _ -> `Top
       end
-    | EvalIntSet e ->
+   (* | EvalIntSet e ->
```
michael-schwarz, line 116: Can we remove instead of comment out?

michael-schwarz:
After fixing these two issues, I think we are good to merge.

PR #169, opened 2021-03-08
Improve JS compatibility

keremc:
## Summary of changes

* Use `batteries.unthreaded` (for the same reason as in goblint/cil#26)
* Update `batteries` to 3.2.0 since this is the first version with `batteries.unthreaded`. Update the lock file.
* Use `qcheck-core` instead of `qcheck` (`qcheck` brings in `ounit2`, which causes some issues in the browser)
* Since signals are not available in the browser, use a no-op timeout implementation there.
* `expressionEvaluation.ml`: Change the visibility of `query` and add some global mutable variables as hook points for Gobview
* `dune`: Split the executable target into a library & executable target.
   * Most modules are now part of `goblint.lib`.
   * All `Main*` modules belong to the executable target (with the exception of `Maingoblint`, which has a few useful functions).
   * Move the entry point of `Maingoblint` to `Goblint`


goblint.opam
```diff
@@ -17,9 +17,8 @@ depends: [
   "dune"
   "zarith_stubs_js"
   "goblint-cil"
-  "batteries"
-  "qcheck"
-  # "z3"
```
sim642, line 22: This Z3 comment should stay. Although I guess it can be annoying to keep around if the opam file is autogenerated.

src/dune
```diff
@@ -3,13 +3,27 @@
...
+
 (executables
   (names goblint maindomaintest) ; TODO: separate domaintest executable?
   (public_names goblint -)
   (modes native js) ; https://dune.readthedocs.io/en/stable/dune-files.html#linking-modes
-  (modules :standard \ apronDomain poly violationZ3)
-  (libraries goblint-cil.all-features batteries zarith_stubs_js qcheck) ; z3
+  (modules goblint mainarinc maindomaintest mainspec)
+  (libraries goblint.lib)
   (preprocess (staged_pps ppx_import ppx_deriving.std ppx_deriving_yojson ppx_distr_guards ocaml-monadic))
```
sim642, line 25: Do the ppx-s still need to be here at the executable if they're already used for the library?

src/util/timeout.ml
```diff
@@ -0,0 +1,22 @@
...
+end
+
+module Js = struct
+  let timeout f arg _ _ = f arg
+  (* TODO: Implement this *)
+end
+
+let timeout = match Sys.backend_type with
+| Other "js_of_ocaml" -> Js.timeout
+| _ -> Unix.timeout
```
sim642, line 22: These lines should be indented.

sim642 (changes requested):

src/maingoblint.ml
```diff
@@ -457,5 +457,6 @@ let main =
       with Exit -> exit 1
     )
 
-let _ =
-  at_exit main
+(* Moving this to the automatically generated goblint.ml. This is for compatibility with Gobview. *)
+(* let _ = *)
+(*   at_exit main *)
```
michael-schwarz, line 462: Maybe change comment to sth like:
```
(* Actual entry point is in auto-generated goblint.ml and is: *)
(* let _ = at_exit main *)
``` 
?

goblint.opam
keremc: Oops! Yeah, I generated the opam file with `dune build @install` and removed that line accidentally.

sim642 (approved):

PR #174, opened 2021-03-26
Re-enable some int domain tests with fixed ikinds

sim642:
This restores some property-based testing possibilities for int domains that now have ikinds. It's a partial fix though: instead of properly sampling arbitrary ikinds for each test, which requires bigger changes and extra consistency assumptions in the tests, I just added a functor, which fixes an ikind on an int domain for property-testing purposes.

I had to disable many ikinds (unsigned and long sizes) for property-testing though because they showed failures in `Interval` and `DefExc`. It could still be the fault of `arbitrary` implementations, which don't actually know which ikind they're generating for. Consistency is just ensured by casting to the tested ikind in the testing functor.

### DIFF
--- a/src/maindomaintest.ml
+++ b/src/maindomaintest.ml
@@ -67,6 +67,7 @@ let domains: (module Lattice.S) list = [
 let nonAssocDomains: (module Lattice.S) list = []
 
 let intDomains: (module IntDomainProperties.S) list = [
+  (module IntDomain.Interval);
   (* (module IntDomain.Flattened); *)
   (* (module IntDomain.Interval32); *)
   (* (module IntDomain.Booleans); *)
--- a/src/maindomaintest.ml
+++ b/src/maindomaintest.ml
@@ -66,8 +66,8 @@ let domains: (module Lattice.S) list = [
 
 let nonAssocDomains: (module Lattice.S) list = []
 
-let intDomains: (module IntDomain.IkindUnawareS) list = [
-  (module IntDomain.Flattened);
+let intDomains: (module IntDomainProperties.OldS) list = [
+  (* (module IntDomain.Flattened); *)
   (* (module IntDomain.Interval32); *)
   (* (module IntDomain.Booleans); *)
   (* (module IntDomain.CircInterval); *)
@@ -75,7 +75,7 @@ let intDomains: (module IntDomain.IkindUnawareS) list = [
   (* (module IntDomain.IntDomTuple); *)
 ]
 
-let nonAssocIntDomains: (module IntDomain.IkindUnawareS) list = [
+let nonAssocIntDomains: (module IntDomainProperties.OldS) list = [
   (* (module IntDomain.DefExc) *)
 ]
 
@@ -95,14 +95,14 @@ let nonAssocTestsuite =
   |> List.flatten
 let intTestsuite =
   List.map (fun d ->
-      let module D = (val d: IntDomain.IkindUnawareS) in
+      let module D = (val d: IntDomainProperties.OldS) in
       let module DP = IntDomainProperties.All (D) in
       DP.tests)
     intDomains
   |> List.flatten
 let nonAssocIntTestsuite =
   List.map (fun d ->
-      let module D = (val d: IntDomain.IkindUnawareS) in
+      let module D = (val d: IntDomainProperties.OldS) in
       let module DP = IntDomainProperties.AllNonAssoc (D) in
       DP.tests)
     nonAssocIntDomains
--- a/src/maindomaintest.ml
+++ b/src/maindomaintest.ml
@@ -80,15 +80,16 @@ let nonAssocIntDomains: (module IntDomainProperties.S) list = [
   (module IntDomain.DefExc);
 ]
 
+(* TODO: make arbitrary ikind part of domain test for better efficiency *)
 let ikinds: Cil.ikind list = [
-  (* TODO: enable more, takes very long and much memory... *)
-  (* IChar; *)
-  (* ISChar; *)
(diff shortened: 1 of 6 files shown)


src/maindomaintest.ml
```diff
@@ -66,46 +66,72 @@ let domains: (module Lattice.S) list = [
...
+  |> List.map (fun (d, ik) ->
+      let module D = (val d: IntDomainProperties.S) in
+      let module Ikind = struct let ikind () = ik end in
+      (module IntDomainProperties.WithIkind (D) (Ikind): IntDomainProperties.OldS)
+    )
 let intTestsuite =
-  List.map (fun d ->
-      let module D = (val d: IntDomain.IkindUnawareS) in
+  old_intdomains intDomains
+  |> List.concat_map (fun d ->
```
michael-schwarz, line 124: Which version of OCaml are we aiming at now? This is `Ocaml >= 4.10.0`.
In the CI we are pretending to check (but failing)  support for `4.07.1` too.
Is support for `4.07.1` something we want to abandon?

sim642: I guess this file could also `open Prelude`, which is our own and opens `Batteries`, which gives `List = BatList`, which should have `concat_map` regardless.

sim642: > In the CI we are pretending to check (but failing) support for `4.07.1` too.

Also, what do you mean by that?

michael-schwarz: No, `BatList` from the latest version (`batteries.3.3.0`) also does not have `concat_map` for `OCaml < 4.10.0`.
https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatList.html

sim642: Damn, that's sure some weird choice by Batteries to limit unnecessarily availability also by OCaml version. I guess I can revert to `map` and `concat` then.

michael-schwarz: > Also, what do you mean by that?

As evidenced by this PR, we do not really compile with `4.07.1`. We specify it in the Github actions

https://github.com/goblint/analyzer/blob/16497fa06d2059f2c3bde130da0bbb89b5a4d6c3/.github/workflows/tests.yml#L17-L19

but then our CI script uses the `pinned` file, installing the pinned version of OCaml.

sim642: If we want to still support 4.07 we should somehow fix the CI then, otherwise it's super misleading and pointless. I don't know how that would work with locking though.

So it seems easier to just ditch all the OCaml version stuff from the CI and just test the locked version. I don't have a reason for still supporting 4.07, but if someone does, I guess the safe thing then would be to have 4.07 locked instead. There's no use to locking 4.11 if none of the new features can be used.

Or I guess a mix of the two is also possible with the following CI configurations:
* Install locked dependencies (including OCaml)
* Install unlocked dependencies and OCaml 4.07
* Install unlocked dependencies and OCaml 4.11

michael-schwarz: Yes, I agree we should fix this.

**General considerations**
Ideally I was hoping to go for a setup like I have for CIL, where we just do everything via `opam` and `dune`.

https://github.com/goblint/cil/blob/60e9c5f77a7267591622b94c9be73555cc2881f1/.github/workflows/tests.yml

This is by far the cleanest and would also give us an indication if some package in an updated version is incompatible with Goblint such that we can put a constraint on the maximum version number of some package. Additional advantage is that testing multiple versions of OCaml is a breeze.

Downside is that we would not have reproduceable builds as the version of dependencies might change between runs.

**Now, as for support of 4.07.1 is concerned:** 
@jerhard needed this at some point for `earlybird`, but I guess `earylbird` now supports later versions? So maybe we can drop support for 4.07.1 now? However, I would still want to keep compatibility with some older versions (e.g. 4.09.1 now), because in my experience it sometimes takes time for e.g. the VSCode plugin (and the language server) to be adapted to a new version of merlin and it is annoying when such things break but you can't go back because Goblint uses new features.

sim642: I'll just replace these `List.concat_map`s for now, so we can merge this while still being 4.07 compatible.

We should probably move this CI fixing to a new issue. Cleaning up the CI procedures is something worth doing regardless of what minimum version we want to keep for now, because it's still `travis-ci.sh`...

sim642:
For whatever reason `Enums` doesn't seem to pass a bunch of tests, so I'll leave its testing commented out for now, but it can be compiled at least.

src/maindomaintest.ml
jerhard: >@jerhard needed this at some point for earlybird, but I guess earylbird now supports later versions? So maybe we can drop support for 4.07.1 now?

I used OCaml 4.07.1 for some time, to be able to use it together with `earlybird`, yes. Currently I use 4.11.1, so I don't rely on the compatibility with 4.07.1 anymore.
The last time I tried, I did not get `earlybird` to work with Goblint and the 4.11.1 compiler, however.

michael-schwarz: The issue is now tracked as #184.

jerhard:
>For whatever reason Enums doesn't seem to pass a bunch of tests, so I'll leave its testing commented out for now, but it can be compiled at least.

I had a look at it and made couple of fixes in the [intdomaintest_enable_enums](https://github.com/goblint/analyzer/tree/intdomaintest_enable_enums) branch. The domaintests now seem to work reliably for Enums, thus I re-enabled them.

We can either make a separate PR for these changes or I just push them here.

jerhard (approved):

sim642:
Thanks! I think I'll merge them into `intdomaintest` and then merge this PR since the tests that pass are better than no tests.

Issue #160, opened 2021-01-18
Port Enums Domain to BigInt 


jerhard:
Use `BigInt` in the implementation of the  `Enums` domain and take the `ikind` of `int`s into account when performing abstract operations.

michael-schwarz:
While we are on this, we should also once again define what exactly we want the enum domain to do. I'm not sure if there is consent on this right now.

jerhard:
>define what exactly we want the enum domain to do

What are the open points? I guess we should discuss in particular the following things:
- should arithmetic operations try to keep information or always go to top
- when do we want to enforce the size bound for sets/exclusion sets:
    - in arithmetic operations (in case they don't go to top, see point above)
    -  in `join`s
    - in `meet`s
    - in `of_interval` and in `of_excl_list`?
- How should `DefExc` and `Enums` relate? In particular, if we keep precise information for arithmetic operations in `Enums`, is there still a need for `DefExc`? (In #52 there were plans to remove `DefExc` a.k.a. `Trier`)

michael-schwarz:
> should arithmetic operations try to keep information or always go to top

In order to not make them some sort of `DefExc` except that it is a `IncExc`, I would suggest making them go to top.

> when do we want to enforce the size bound for sets/exclusion sets

This is a good point: In DefExc we have no fixed bound on the number of excluded elements. Can we maybe get away with the same logic here? I guess the only real blow-up that happens is because of arithmetic operations?

I guess we should discuss this at the next GobCon maybe?

**Additional, tangentially related, point:** Right now, we don't make use of the fact that things that are an enum in the C program actually have type `TEnum` which has an `enuminfo` value.  https://people.eecs.berkeley.edu/~necula/cil/api/Cil.html#TYPEenuminfo.

I was initially hoping one could restrict the possible values of a variable of type `enum` to the ones appearing in the enumeration. Unfortunately, the standard allows any integer value that fits into the `integer type` chosen by the compiler to represent the enum, so we can't really gain anything there. However, it might still be interesting to warn when doing things that lead to values outside the members of an enumeration, as this could point to a bug.

sim642:
My vague memory from a GobCon discussion before SV-COMP is that `Enums` is supposed to handle possible values for enums (duh!) on which arithmetic operations don't really make sense. In that view, it would be fine for arithmetic operations to go to top.

[...]

PR #175, opened 2021-03-26
Port the Enums domain to bigint

jerhard:
This ports the `Enums` domain to use arbitrary precision integers for representing values, see #160, and taking the `ikind` of the operands into account. A `norm` function is used to go to top if an overflow might have occurred.


src/cdomains/intDomain.ml
```diff
@@ -1862,32 +1862,49 @@ module Booleans = MakeBooleans (
...
-  module I = Integers
+  module I = BigInt
   module R = Interval32 (* range for exclusion *)
 
   let range_ikind = Cil.IInt
   let size t = R.of_interval range_ikind (let a,b = Size.bits_i64 t in Int64.neg a,b)
   type e = I.t (* element *)
-  and t = Inc of e list | Exc of e list * R.t [@@deriving to_yojson] (* inclusion/exclusion set *)
-  type int_t = int64
+  and t = Inc of I.t list | Exc of I.t list * R.t [@@deriving to_yojson] (* inclusion/exclusion set *)
```
sim642, line 1873: I feel like we should be using actual sets here, not lists. It's probably annoying to rewrite though because singleton `Inc`s cannot be then pattern matched.

src/cdomains/intDomain.ml
```diff
@@ -1862,32 +1862,49 @@ module Booleans = MakeBooleans (
...
-  let bot_of ik = bot ()
-  let is_top x = x = top ()
+  let top () = failwith "top () not implemented for Enums"
+  let bot_of ik = Inc []
+
+  let min_int ik = I.of_bigint @@ fst @@ Size.range_big_int ik
+  let max_int ik = I.of_bigint @@ snd @@ Size.range_big_int ik
 
-  let equal a b = a = b (* Be careful: this works only as long as the Range/Interval implementation used does not use big integers *)
+  let equal a b = a = b
```
sim642, line 1886: Should this not delegate to inner `equal` functions?

src/cdomains/intDomain.ml
```diff
@@ -1941,79 +1958,111 @@ module Enums : IkindUnawareS = struct
...
 
-  let widen x y = join x y
-  let narrow x y = meet x y
+  let widen = join
+  let narrow = meet
 
-  let leq x y = join x y = y
+  let leq x y = (join_ignore_ikind x y) = y
 
   let max_elems () = get_int "ana.int.enums_max" (* maximum number of resulting elements before going to top *)
```
sim642, line 1990: This option seems unnecessary now that only singleton `Inc` arithmetic is possible. We shouldn't be limiting the number of enum values we can analyze, because only finitely many can exist in a given program as they must literally be present in the code.

Singleton `Inc` arithmetic doesn't seem to require this bound for termination either. For example, `for (int i = 0; true; i++);`:
1. Initially: `Inc [0]`
2. First iteration gives `Inc [0] + Inc [1] = Inc [1]`
3. First iteration join gives `Inc[0] join Inc [1] = Inc [0, 1]`
4. Second iteration gives `Inc [0, 1] + Inc [1] = top`
5. Second iteration join gives `Inc [0, 1] join top = top`
6. Third iteration gives `top + Inc [1] = top`
7. Third iteration join gives `top join top = top`, fixpoint reached.

Or is there still some case that does require the upper bound for termination?

src/cdomains/intDomain.ml
jerhard: With 471f833356b253f7f5730f1c6cc8f8855d3bf0b3 the domain is now using sets, but this caused a termination problem of the solver in `01/39`. I am not sure what caused this. Termination on the tests work when not doing narrowing for Enums (see  96bb4a08d07f38bcb9be83ab0fa9e02018f59bc3).

src/cdomains/intDomain.ml
jerhard: Done.

src/cdomains/intDomain.ml
jerhard: I removed the precise arithmetic for unary operations (unary minus, bitwise negations) on inclusion sets with multiple elements (2bc92b0e58156a058ccf0268444a1987eebc88a8), which then allows to remove the cardinality limit.

src/cdomains/intDomain.ml
michael-schwarz: It would be very interesting to see where this happens, it is a bit surprising given that just switching from representing things as lists to representing them as sets should not cause any changes in termination. :thinking: 

I'd somehow expect that issue to be tied to the loop in the program
https://github.com/goblint/analyzer/blob/20f3a7d6285a35c7c7d5c2f6206e280a73b83c25/tests/regression/01-cpa/39-interval-and-enums.c#L52-L58

michael-schwarz:
Do you think it is in a shape where it is reasonable for me to run `csmith` on it again?

src/cdomains/intDomain.ml
jerhard: Yes, the solver is looping when analyzing this `for`-loop. 

The non-termation was introduced by the changes to [Enums here](https://github.com/goblint/analyzer/pull/175/commits/471f833356b253f7f5730f1c6cc8f8855d3bf0b3#diff-7c2aff56950cfd3ab8421eed789e8ee63632f69397ffbe53db74ae4729641fa3); I don't see what could possibly cause this.

jerhard: I think I found the problem. The issue is caused because the `hash` and `compare` for `Enums` now have to be implemented differently; probably because structural equality does not work for `Set`.

jerhard:
>Do you think it is in a shape where it is reasonable for me to run csmith on it again?

Now with the fix for the `narrow`,`hash` and `compare`, I think it's ready for `csmith`.

src/cdomains/intDomain.ml
```diff
@@ -1861,193 +1861,247 @@ module Booleans = MakeBooleans (
...
       else (* downcast: may overflow *)
-        Exc ([], r')
-    |  Inc x -> Inc (List.map (Integers.cast_to ik) x)
+        Exc (ISet.empty, r')
+    |  Inc x -> Inc (ISet.map (BigInt.cast_to ik) x)
+
+  let of_int ikind x = cast_to ikind (Inc (ISet.singleton x))
 
-  let of_interval ik (x,y) = (* TODO this implementation might lead to very big lists; also use ana.int.enums_max? *)
+  let of_interval ik (x,y) = (* TODO this implementation might lead to very big lists; use some configurable limit? *)
```
sim642, line 1961: When does `of_interval` actually get used? It could be a problem if somehow we take a large interval based on the interval analysis and then try to re-express it into enums. In such case it might only make sense to handle singleton intervals exactly and everything else as top.

src/cdomains/intDomain.ml
```diff
@@ -1861,193 +1861,247 @@ module Booleans = MakeBooleans (
...
+
+  let hash = function
+    | Inc x ->
+      let m = match BI.to_int (ISet.min_elt x) with n -> n | exception Not_found -> 41 | exception Failure _ -> 61 in (* Failure occurs when the conversion to int did not work, Not_found occurs when the set is empty *)
+      let n = match BI.to_int (ISet.max_elt x) with n -> n | exception Not_found -> 43 | exception Failure _ -> 67 in
+      3  * ISet.cardinal x + 7 * m + 13 * n
+    | Exc (x, r) ->
+      let m = match BI.to_int (ISet.min_elt x) with n -> n | exception Not_found  -> 53 | exception Failure _ -> 71 in
+      let n = match BI.to_int (ISet.max_elt x) with n -> n | exception Not_found  -> 59 | exception Failure _ -> 73 in
+      R.hash r + 5  * ISet.cardinal x + 11 * m + 19 * n
```
sim642, line 1920: Is there a reason why this is so ad-hoc and involved? If it's because `ISet.hash` doesn't exist, one could just use `SetDomain.Make` instead, like `DefExc`.

src/cdomains/intDomain.ml
```diff
@@ -1861,193 +1861,247 @@ module Booleans = MakeBooleans (
...
+  type int_t = BI.t
   let name () = "enums"
-  let top_range = R.of_interval range_ikind (-99L, 99L) (* Since there is no top ikind we use a range that includes both ILongLong [-63,63] and IULongLong [0,64]. Only needed for intermediate range computation on longs. Correct range is set by cast. *)
-  let bot () = Inc []
-  let top_of ik = Exc ([], size ik)
-  let top () = Exc ([], top_range)
-  let top_of ik = top ()
-  let bot_of ik = bot ()
-  let is_top x = x = top ()
+  let to_yojson x = failwith @@ "to_yojson unimplemented for " ^ (name ())
```
sim642, line 1880: Not sure if we really need this for anything anymore, but maybe `[@@deriving to_yojson]` would work here also.

src/cdomains/intDomain.ml
jerhard: The `of_interval` function in Enums is actually never called. 
Only the `of_interval` function of the interval domain gets actually called, in particular to construct ranges in `DefExc` and `Enums`.

src/cdomains/intDomain.ml
jerhard: Thanks, I am now using `SetDomain.Make`.

src/cdomains/intDomain.ml
jerhard: Yes, this works now when `SetDomain.Make` is used.

src/cdomains/intDomain.ml
jerhard: I changed the implementation now anyway to only produce singleton sets or top.

sim642 (approved):

jerhard:
@michael-schwarz Do you first want to run additional tests, or should we merge this PR?

michael-schwarz:
I'll start the `csmith` with it, leave it running for a while, and then we should be good to merge this evening!

michael-schwarz:
I found some more nice ikind issues :disappointed: 

~~~C
int main (int argc, char* argv[])
{
    signed char f2 = 7;
    signed char l_1857 = ((0xFFBD4A17L && f2) | f2);
}
~~~

jerhard:
> I found some more nice ikind issues disappointed
> (But apparently these fail for intervals already, too.)
> 
> ```c
> int main (int argc, char* argv[])
> {
>     signed char f2 = 7;
>     signed char l_1857 = ((0xFFBD4A17L && f2) | f2);
> }
> ```

Ah, interesting! I think this is actually caused by a Cil bug that gives the wrong type for a sub expression.
Cil generates the following code:
```
int main(int argc , char **argv ) 
{ 
  signed char f2 ;
  signed char l_1857 ;
  {
#line 5
  f2 = (signed char)7;
#line 6
  l_1857 = (signed char )((f2 != (signed char)0) | (int )f2);
#line 7
  return (0);
}
}
```
According to Cil, the subexpression `f2 != (signed char)0` (which is evaluated by `rval`) has the typ `signed char`; but the type should be `int`.

michael-schwarz:
> According to Cil, the subexpression f2 != (signed char)0 (which is evaluated by rval) has the typ signed char; but the type should be int.

Interesting, then we should probably make a bug against CIL. But once we have an issue in the CIL repo so we don't forget it, I think it should not black us from merging this PR.

sim642:
I agree, it's not an issue with enums themselves, but rather CIL's attempt to be smart.

src/cdomains/intDomain.ml
vogler: Yes, we used sorted lists because it should be cheaper than sets which have no cheap `equal`/`compare` (O(1) vs. O(n log n)?).
And because it allows for easier pattern matching.
Did you check if it hashconsing still works for equal sets with different repr.?
What speaks for using sets?

sim642: `DefExc` already uses sets, so if there's any problem with that, it existed already there. So if we want to optimize, we'd probably want to optimize both simultaneously. It wouldn't be very difficult to give a sorted list based `SmallSet` implementation of the same signature.

src/cdomains/intDomain.ml
vogler: Was there a problem with doing arithmetic on multiple elements?
The idea behind Enums was to not lose information on join/widen but on arithmetic operations (so it that it's precise for enums).
Non-termination can only arise due to new values from arithmetic operations - that's what the `ana.int.enums_max` was for.
What's the difference between this singleton `Inc` arithmetic and `ana.int.enums_max = 1`?

src/cdomains/intDomain.ml
vogler: Yes, probably makes sense. At the time I implemented Enums I didn't want to change DefExc.
Maybe first check these assumptions:
- physical and structural equality for lists are both O(1)
- set equality is done by fold mem which is O(n * log n)
- we call `equal`/`compare` way more often than we insert elements (which has the same complexity for insertion sorted lists)
- we have cases where we get sets that are equal but not structurally equal - how? Just inserting elements in different order shouldn't be enough since it's binary trees - combinations of insert/remove?

PR #180, opened 2021-04-07
Remove unsound option dopartial

### LINKED ISSUE #179: Remove unsound option `dopartial`
The option `dopartial` does a run of CIL's partial evaluation before starting Goblint.
**Unfortunately this partial evaluation of CIL is unsound (as uncovered by @keremc), even for simple programs:** 

```
int main(void)
{
    int x = 42;
    int *p = &x;
    *p = 0;
    int y = x;
    assert(y == 42); // FAIL!
}
```

is transformed into

```
int main(void) 
{ 
  int x;
  int *p;
  int y;

  x = 42;
  p = & x;
  *p = 0;
  y = 42;

  assert(1);

  return (0);
}
```

I suggest we remove `dopartial` from Goblint, as nobody really uses it anyway and thus not worth fixing. This does not influence CIL's constant folding, that is also not unsound.
References https://github.com/goblint/cil/issues/28

michael-schwarz:
Closes #179

### DIFF
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -274,8 +274,6 @@ let merge_preprocessed cpp_file_names =
     | xs -> Cilfacade.getMergedAST xs |> Cilfacade.callConstructors
   in
 
-  (* using CIL's partial evaluation and constant folding! *)
-  if get_bool "dopartial" then Cilfacade.partial merged_AST;
   Cilfacade.rmTemps merged_AST;
 
   (* create the Control Flow Graph from CIL's AST *)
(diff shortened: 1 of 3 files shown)


sim642 (approved):
Seems reasonable.

I'm not familiar with dopartial at all, but does it do any kind of sound constant folding that our base analysis wouldn't?

michael-schwarz:
> I'm not familiar with dopartial at all, but does it do any kind of sound constant folding that our base analysis wouldn't?

No, not really. Ideally, it would be cheaper than using the base analysis because it is not context-sensitive, but given that it is broken and no-one uses it, I don't want to invest the time to fix it.

PR #182, opened 2021-04-08
Make known functions in unknown AD still reachable

sim642:
When determining reachable addresses, e.g. for the purpose of an `extern` unknown function spawning everything given as arguments as threads, there's the case of an unknown address. Previously if the set contained that, nothing was spawned at all! With this change, at least the known addresses would be spawned. Obviously this may still be technically unsound, but it's a bit closer to reasonable behavior.

Code-wise it's a very small change, but the impact might potentially be huge. Therefore instead of quietly changing this, I want to know if anyone sees a problem with this. Or has any idea, why this hasn't been the behavior to far.

### DIFF
--- a/tests/regression/02-base/46-spawn-global-funptrs.c
+++ b/tests/regression/02-base/46-spawn-global-funptrs.c
@@ -2,11 +2,11 @@
 #include <assert.h>
 
 void foo() {
-  assert(1);
+  assert(1); // assert reachable
 }
 
 void bar() {
-  assert(1);
+  assert(1); // assert reachable
 }
 
 void (*funs[2])() = {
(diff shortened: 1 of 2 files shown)


tests/regression/02-base/46-spawn-global-funptrs.c
```diff
@@ -0,0 +1,33 @@
...
+extern void magic2(void (*funs[])());
+
+void *t_fun(void *arg) {
+  // just for going to multithreaded mode
+  return NULL;
+}
+
+int main() {
+  pthread_t id;
+  pthread_create(&id, NULL, t_fun, NULL); // enter multithreaded mode
```
michael-schwarz, line 28: Should `magic2(funs)` not already go to MT mode?

https://github.com/goblint/analyzer/blob/0df380bc67e1df90533f64d44d62a905f0ca3687/src/util/defaults.ml#L164

michael-schwarz:
Seems reasonable to me!

tests/regression/02-base/46-spawn-global-funptrs.c
sim642: The caveat with that is it only goes multithreaded when it actually spawns something. Arguably this should happen regardless, but it's hardly useful in the current state anyway where we don't have well-defined semantics and configuration for what unknown functions may and may not do.

But actually you're right that this explicit thread creation is unnecessary for the test case. Calling `magic1` in singlethreaded mode already invalidates `funs` to add unknown to it. I'll simplify the test.

vogler:
We should have the larger benchmarks run on some (maybe local) CI, to have some record of how performance changes for things like this.
I'll use >= this commit for rerunning `bench` and update.

Issue #178, opened 2021-04-07
Move extern inits, global inits and otherfuns spawning into constraint system


sim642:
Currently extern inits, global inits and otherfuns spawning is done by manually simulating transfer function calls with degenerate `ctx`:
https://github.com/goblint/analyzer/blob/20f3a7d6285a35c7c7d5c2f6206e280a73b83c25/src/framework/control.ml#L175-L349

In my experience, time and time again, this has been causing issues because:
1.`ctx`s are degenerate – some outright crash, others have behavior which doesn't match the one available during solving;
2. it somewhat duplicates what `FromSpec` does;
3. it misses some nontrivial things that would happen during actual solving – e.g. `main` is unaware of otherfuns having been spawned at all;
4. it's extremely error prone and hard to maintain.

The sensible thing would be to do all these things within the constraint system and let the solver evaluate them.

sim642:
Turns out there is yet another maniacal form of global initialization that we handle in yet another weird way: by adding them as calls into the main function to be performed during the solving, not during our global inits. No idea if any sane person uses them, since we have some support, their handling should also be made uniform with the rest.

https://github.com/goblint/analyzer/blob/c398773746841ff7e51525f4911c831368b626ce/src/util/cilfacade.ml#L136-L149

https://github.com/goblint/analyzer/blob/c398773746841ff7e51525f4911c831368b626ce/src/util/cilfacade.ml#L109-L121

vogler:
Are these C++ class constructors or some other feature in C?

sim642:
It's a C thing. From [GCC manual](https://gcc.gnu.org/onlinedocs/gcc-10.3.0/gcc/Common-Function-Attributes.html#index-constructor-function-attribute):

> The `constructor` attribute causes the function to be called automatically before execution enters `main ()`. [...] Functions with these attributes are useful for initializing data that is used implicitly during the execution of the program.

PR #183, opened 2021-04-08
Traces article implementation

sim642:
I'm opening this humongous draft pull request already now to have a place for keeping some notes on all the changes `traces` branch contains, including many general improvements etc.
Eventually these would get merged into `master`.

### Overview of changes
1. Many implementations for privatization via `exp.privatization` option.
    1. The new default privatization is `protection-read` aka `ProtectionBasedPriv` with additional readable lockset check aka Protection-Based Reading with additional readable lockset check. The additional check is necessary to handle many of our regression tests.
    2. Regression tests which require more precise privatizations have `// TODO` comments to be useful with regression testing script.
    3. Privatization extracted and abstracted from `base.ml` to `basePriv.ml`.
    4. `base` analysis local domain has third privatization-dependent component (by @vesalvojdani).
    5. `base` analysis global domain is privatization-dependent.
    6. Add `reason` argument to `sync` transfer function to allow analyses choose if they want to do computations at every node, or only at some necessary places (like join).
2. Changes to `mutex` analysis.
    1. Write-based protecting locksets computed by `mutex` analysis (by @vesalvojdani).
    2. Number of new mutex-related queries needed by different privatizations.
3. Event system (`ctx.emit` and `event` transfer function) for deduplicating thread creation, lock, unlock and escape logic in different analyses.
    1. Use separate `MCPSpec` for analyses inside `MCP`.
    2. Replace `ctx.split`'s explicit branch arguments with arbitrary events.
4. Comparison of privatization precisions with `privPrecCompare`.
    1. Based on dumping (marshalling) all `read_global` results via `exp.priv-prec-dump` option.
5. General improvements.
    1. Optimization of large CPAs that occur in some privatizations (issue #164).
    2. Partial fix of path-sensitive thread creation (issue #148).
    3. Use right value of `exp.earlyglobs` already during global initialization.
    4. Move global invariant simulation in `Control` earlier to allow `enter_with` and `otherstate` to also `sideg` (issue #178).
    5. Add new `sem` (semantics) category of options for configuring what unknown functions etc may and may not do.
    6. Fix unknown functions not invalidating unprivatized globals.
    7. Fix `Deadcode` behavior of `invariant` on binary operators.
6. General utilities.
    1. `AfterConfig` module for running initialization things after config has been loaded.
    2. `RichVarinfo` module for abusing `varinfo`s to represent richer data, particularly as global constraint variables. 

### Things broken
- [x] OSEK regression tests fail with new privatizations. Defaulting to `protection-vesal` for them now, but `old` passes as well.
- [ ] Witness generation is untested after significant `MapDomain` interface changes from #164. Domain tests fail for `HoareMap`.


michael-schwarz:
I'll take a look at the code when I'm back at the office, too huge to look at it on my laptop screen!

**Quick question: Is there somewhere a mapping of the names we used here to the names we proposed in the paper?
It would be convenient to have a mapping that says (for the setting described in the paper):**

My guess would be:
Miné's 2012 analysis adapted to our setting -> `???`
Protection-Based -> `global` ?
Lock-Centered -> `mine-lazy`?
Write-Centered -> `???`
Combined -> `global-history`

sim642:
The mapping for the paper is the following:
| Traces paper name | `exp.privatization` option | OCaml module name |
| --- | --- | --- |
| Protection-Based Reading | `global` | `PerGlobalPriv` |
| Lock-Centered Reading | `mine-lazy` | `MineLazyPriv` |
| Write-Centered Reading | `global-history` | `PerGlobalHistoryPriv` |
| Combined | `mine-global` | `MinePerGlobalPriv` |
| Miné 2012 | `mine-W` | `MineWPriv` |

---

The complete mapping from `exp.privatization` option to OCaml modules is at the end of `basePriv.ml`:
https://github.com/goblint/analyzer/blob/53efb01f8bb33dac3b46998335d94b29c5dadc00/src/analyses/basePriv.ml#L1640-L1653

OCaml module documentation also has the paper names:
https://github.com/goblint/analyzer/blob/53efb01f8bb33dac3b46998335d94b29c5dadc00/src/analyses/basePriv.ml#L1305-L1306

src/framework/control.ml
```diff
@@ -260,7 +273,8 @@ struct
       WResult.init file; (* TODO: move this out of analyze_loop *)
 
     GU.global_initialization := true;
-    GU.earlyglobs := false;
+    (* GU.earlyglobs := false; *)
+    GU.earlyglobs := get_bool "exp.earlyglobs";
```
michael-schwarz, line 277: If we already set it to `exp.earlyglobs` here, do we still need to set it again in Line 368?

michael-schwarz:
Great, thanks for the table!


Traces paper name | exp.privatization option | OCaml module name
-- | -- | --
Combined | mine-global | MinePerGlobalPriv

I'd suggest renaming `mine-global`, as it is quite different from the analysis proposed by him in 2012. Still having the option for it called `mine-*` is confusing, as I would expect that to be one of his analyses. How about `combined-lock-write` or something?

`mine-lazy` for Lock-Centered however is more justified, as it is indeed a modified and improved version of Miné 2012.

src/framework/control.ml
sim642: Nah, this just stayed around from when I was experimenting with this. That way it was easier to comment this first one out.

sim642:
I agree `mine-global` has the most confusing name out of all of them. If that's the only one we're going to rename now, calling it `lock` and `write` is inconsistent with the others. Then it might make sense to also rename the others to use those terms. And then `mutex-*` are in a weird position, because they're also Lock-Centered in a way, but have nothing to do with the Lock-Centered stuff based on Miné.

How about just `combo-global-mine`?

michael-schwarz:
> If that's the only one we're going to rename now,

I would be very wary of introducing names different from the paper at all where it is not required.
Realistically, that only confuses the reviewers of the artifact and potentially ourselves (or future Goblint developers as the paper will likely be the best documentation we will ever write for this feature).

I would suggest we prefix all analyses that have separate unknowns per global with a `g-`. So we would e.g. have
`g-old`, `g-protection`, `g-protection-read`, `g-vesal`, `g-mine-weak`, `g-mine-nothread`, `g-mine-W`, `g-mine-W-init`, `g-lock`, `g-write`, and `g-lock-write`.
The `per-mutex` ones would then be called `m-oplus` and `m-meet`.

What do you think?

sim642:
> I would suggest we prefix all analyses that have separate unknowns per global with a `g-`. So we would e.g. have
> `g-old`, `g-protection`, `g-protection-read`, `g-vesal`, `g-mine-weak`, `g-mine-nothread`, `g-mine-W`, `g-mine-W-init`, `g-lock`, `g-write`, and `g-lock-write`.

The separate unknowns per global is just a choice made in the paper. For example, the implementation of Lock-Centered Reading does this:
https://github.com/goblint/analyzer/blob/7518783515f2635a04b88c7567186b19b7a22fba/src/analyses/basePriv.ml#L707

There the unknowns are essentially mutexes, which correspond to the unlock-lock communication. So it might even be more appropriate to think of this as per-mutex as well, while the background lockset just provides a further partitioning of the `CPA` stored at that mutex.

And that's also the case for all the other analyses in the paper except Protection-Based. Of course one can factor the flat tuple unknowns in the paper also first by global variable in an equally valid way, but that's just a question of preference. It seemed the most appropriate here to do it by mutex since then each unlock does one `sideg` and each lock does one `getg` as opposed to an unknown number based on the total number of global variables. Although relationality is out of the scope of this paper, who's saying that you can't try using an octagon instead of `CPA` under the partitioned global unknown for a mutex?

This is why I'm hesitant to call them all per-global

sim642:
The easiest naming scheme following the paper would be something like: `protection`, `lock`, `write`, `write+lock` and `mine-W`. And the rest changed to somehow match this scheme of prefixes or staying with their current names, since they're irrelevant for the paper.

michael-schwarz:
> It seemed the most appropriate here to do it by mutex since then each unlock does one sideg and each lock does one getg as opposed to an unknown number based on the total number of global variables. Although relationality is out of the scope of this paper, who's saying that you can't try using an octagon instead of CPA under the partitioned global unknown for a mutex?

Yes, I agree, this is the better way to implement it, and calling it per-global is a bit confusing then!

> The easiest naming scheme following the paper would be something like: `protection`, `lock`, `write`, `write+lock` and `mine-W`.

Yes, then we should probably do this! And yes, the other names probably are not so relevant anyway for now.

src/analyses/baseUtil.ml
```diff
@@ -0,0 +1,19 @@
+open Prelude.Ana
+open GobConfig
+module Q = Queries
+
+let is_global (a: Q.ask) (v: varinfo): bool =
+  v.vglob || match a (Q.MayEscape v) with `MayBool tv -> tv | _ -> false
+
+let is_static (v:varinfo): bool = v.vstorage == Static
+
+let is_always_unknown variable = variable.vstorage = Extern || Ciltools.is_volatile_tp variable.vtype
```
michael-schwarz, line 10: Why are we using physical equality for `vstorage` once and structural equality the other time?

sim642: No reason I guess, I just extracted those form the current top of `base.ml`. Given that storage is just a basic enum, they should be equivalent in compiled code, so I suppose structural equality should be fine for both and consistency.

src/analyses/baseUtil.ml
```diff
@@ -0,0 +1,19 @@
+open Prelude.Ana
+open GobConfig
+module Q = Queries
+
+let is_global (a: Q.ask) (v: varinfo): bool =
+  v.vglob || match a (Q.MayEscape v) with `MayBool tv -> tv | _ -> false
```
michael-schwarz, line 6: Should this not default to `true` in case `threadEscape` is off?

src/util/cilfacade.ml
```diff
@@ -17,7 +17,7 @@ let init () =
   print_CIL_Input := true
 
 let currentStatement = ref dummyStmt
```
michael-schwarz, line 19: Since we touched this anyway, maybe also make `current_statement`?
Then naming is consistent at least locally here

src/maingoblint.ml
```diff
@@ -431,6 +432,7 @@ let main =
         Stats.reset Stats.SoftwareTimer;
         parse_arguments ();
         check_arguments ();
+        AfterConfig.run ();
 
         (* Cil.lowerConstants assumes wrap-around behavior for signed intger types, which conflicts with checking
           for overflows, as this will replace potential overflows with constants after wrap-around *)
```
michael-schwarz, line 438: Maybe we should warn here if we change the settings?
Not really related to this PR, just saw this now

michael-schwarz: And probably move to `check_arguments` too

src/analyses/baseUtil.ml
sim642: That would be (more) sound indeed. The consequence of that is all variables become global, which is hardly ever the desired outcome. I guess the past intent was that disabling escape analysis shouldn't have that outcome. I can make this change and add a warning about it too.

This also requires updating a bunch of tests to make them actually list escape analysis.

src/framework/control.ml
```diff
@@ -359,6 +374,7 @@ struct
     in
 
     let entrystates = List.map (fun (n,e) -> (MyCFG.FunctionEntry n, Spec.context e), e) startvars in
+    let entrystates_global = GHT.to_list gh in
```
michael-schwarz, line 377: Could you explain why this works with `exp.earlyglobs` on, whereas it did not when `GHT.to_list gh` to list was called in `do_global_inits`?
This strikes me as odd

src/analyses/baseUtil.ml
sim642: I just committed the fix, but that just caused all partitioned array tests to fail. I guess it tries to use `is_global` with some degenerate `ctx` somewhere?

michael-schwarz: >  I can make this change and add a warning about it too.

Yes, I think that is a good idea!

src/cdomains/valueDomain.ml
```diff
@@ -2,10 +2,7 @@ open Cil
 open Pretty
 open GobConfig
 
-module ID = IntDomain.IntDomTuple
-module IndexDomain = IntDomain.IntDomWithDefaultIkind (ID) (IntDomain.PtrDiffIkind)
-module AD = AddressDomain.AddressSet (IndexDomain)
-module Addr = Lval.NormalLat (IndexDomain)
+include PreValueDomain
```
michael-schwarz, line 5: Why did we move this?

src/domains/events.ml
```diff
@@ -0,0 +1,18 @@
+(* open Deriving.Cil *)
```
michael-schwarz, line 1: Remove commented out open?

src/framework/control.ml
sim642: I think it might still work if left there, but since I moved `gh` out of `do_global_inits` for `enter_with`, then it also made sense to move the final transformation of `gh` out of there. Now both `entrystates` are defined in the same location.

src/cdomains/valueDomain.ml
sim642: I wanted queries to refer to the same `Addr` without having to repeat all the preliminary definitions. If not extracted out to a third module, there's a circular dependency `ValueDomain -> Queries -> ValueDomain`.

src/framework/constraints.ml
```diff
@@ -395,7 +395,8 @@ module WidenContextLifterSide (S:Spec)
 =
 struct
   module B = WidenContextLifter (S) (* can't just include this since type of ctx and some tf functions are different; TODO some generic functor to lift functions with conv, inj, proj? *)
```
michael-schwarz, line 397: comment outdated?

sim642: This comment is merged in from a recent change on master:
https://github.com/goblint/analyzer/commit/67c9c7dd96b030ea980a73dbc56af1c88bab4b96#diff-353a05f938a5047a918e48ea3a800661cd3a72304470b731f4d9bbe77bf0de96R397

src/cdomains/baseDomain.ml
```diff
@@ -58,42 +62,129 @@ struct
...
+end
+
 module type ExpEvaluator =
 sig
-  val eval_exp: CPA.t * PartDeps.t ->  Cil.exp -> IntOps.BigIntOps.t option
+  type t
+  val eval_exp: t  ->  Cil.exp -> IntOps.BigIntOps.t option
 end
 
 (* Takes a module specifying how expressions can be evaluated inside the domain and returns the domain *)
```
michael-schwarz, line 158: Comment should reflect it takes two modules

src/analyses/symbLocks.ml
```diff
@@ -145,23 +145,6 @@ struct
     | _ ->
       ust
 
-  let may_race (ctx1,ac1) (ctx,ac2) =
```
michael-schwarz, line 148: Could you explain why we removed this?

src/analyses/poly.ml
```diff
@@ -20,7 +20,7 @@ struct
   let context x = if GobConfig.get_bool "exp.full-context" then x else D.bot ()
 
   let threadenter ctx lval f args = D.top ()
-  let threadspawn ctx lval f args fctx = D.bot ()
+  let threadspawn ctx lval f args fctx = ctx.local
```
michael-schwarz, line 23: Could you explain to me why `threadspawn` should return `ctx.local` rather than `D. bot ()`?
I am bit lost here, this used to be `otherstate` iirc?

src/analyses/baseUtil.ml
michael-schwarz: > I guess it tries to use is_global with some degenerate ctx somewhere?
I'll take a look, that is my code after all :smile:

src/analyses/poly.ml
sim642: `threadenter` used to be `otherstate`. `threadspawn` is called where the thread was created from (by any analysis).

src/analyses/symbLocks.ml
sim642: This was removed by @vesalvojdani early on: https://github.com/goblint/analyzer/commit/1ce6e7889419f8e96acc551537752ea485b23ebc#diff-5a9240ccb6206e2ebe2d89a00ae618eea08946c5569e44e0a63a2c9456784ce8L146-L162. I think it was a leftover from some older-style race checking that wasn't actually used any more.

src/analyses/poly.ml
michael-schwarz: Thanks! Should we maybe document such things in `analyses.ml` where we define the `module type Spec`?

src/analyses/baseUtil.ml
michael-schwarz: >  should we move this fix to a separate branch and PR?

Yes, let's do that!

src/analyses/poly.ml
sim642: I'll add that there but the entire act of spawning threads has become very nuanced. It's impossible to understand without looking at how `FromSpec` and `MCP2` put the transfer functions together and on which context which thing is called in what order. This needs a figure...

src/analyses/baseUtil.ml
sim642: I'll revert them here and cherry-pick to new branch then.

src/analyses/basePriv.ml
```diff
@@ -0,0 +1,1664 @@
...
+  {st with cpa = CPA.bot (); priv = startstate ()}
+
+module OldPrivBase =
+struct
+  include NoInitFinalize
+  module D = Lattice.Unit
+
+  let startstate () = ()
+
+  let lock ask getg cpa m = cpa
```
michael-schwarz, line 60: Why not name it `st` here too?

src/analyses/basePriv.ml
```diff
@@ -0,0 +1,1664 @@
...
+          (st,acc)
+      in
+      if M.tracing then M.traceu "globalize" "Done!\n";
+      res
+    in
+    (* We fold over the local state, and collect the globals *)
+    CPA.fold add_var st.cpa (st, [])
+end
+
+module OldPriv: S =
```
michael-schwarz, line 115: Maybe one should remark somewhere that it is unsound?!
And maybe even remove it from the official list of options for this listed in `defaults.ml`

michael-schwarz:
Should we rename `exp.precious_globs` to something more useful?
E.g., we could have `exp.earlyglobs.enabled` for `exp.earlyglobs` and `exp.earlyglobs.exclude` for `exp.precious_globs`?

sim642:
> Should we rename `exp.precious_globs` to something more useful?
> E.g., we could have `exp.earlyglobs.enabled` for `exp.earlyglobs` and `exp.earlyglobs.exclude` for `exp.precious_globs`?

I thought about it when changing something related, but then found that "precious globs" are also "my favorite things" that exclude things from invalidation:
https://github.com/goblint/analyzer/blob/85536ab5c9cc0ffe5088af1c01eef34be27fa00e/src/analyses/base.ml#L1799-L1803

This is independent of earlyglobs.

These two different exclusions should probably be separated then. But I'd defer it off this branch.

src/analyses/basePriv.ml
```diff
@@ -0,0 +1,1664 @@
...
+  val unlock: Q.ask -> (varinfo -> G.t) -> (varinfo -> G.t -> unit) -> BaseComponents (D).t -> LockDomain.Addr.t -> BaseComponents (D).t
+
+  val sync: Q.ask -> (varinfo -> G.t) -> BaseComponents (D).t -> [`Normal | `Join | `Return | `Init | `Thread] -> BaseComponents (D).t * (varinfo * G.t) list
+
+  val escape: Q.ask -> (varinfo -> G.t) -> (varinfo -> G.t -> unit) -> BaseComponents (D).t -> EscapeDomain.EscapedVars.t -> BaseComponents (D).t
+  val enter_multithreaded: Q.ask -> (varinfo -> G.t) -> (varinfo -> G.t -> unit) -> BaseComponents (D).t -> BaseComponents (D).t
+  val threadenter: Q.ask -> BaseComponents (D).t -> BaseComponents (D).t
+
+  (* TODO: better name *)
+  val is_private: Q.ask -> varinfo -> bool
```
michael-schwarz, line 34: What is the purpose of this now?

src/analyses/basePriv.ml
sim642: I'm doing the renaming of all of these locally and have added a comment here about its unsoundness.

michael-schwarz:
> I thought about it when changing something related, but then found that "precious globs" are also "my favorite things" that exclude things from invalidation:

The `precious` names we give things in Goblint are sometimes very `extraspecial`, one of `my_favorite_things` :wink:

src/analyses/basePriv.ml
sim642: I keep asking myself this question each time I see this function. Especially since it makes no sense for most implementations. But I'm also not sure about the implications of removing it since there is one use in `base.ml`:
https://github.com/goblint/analyzer/blob/4b0fa27033db36447f04b5b016c61ce8b26cc88e/src/analyses/base.ml#L1042-L1048

I think it somehow matters for protection-based privatizations (or maybe just the `sync`-based ones like `old` and `global-vesal`?) because it wants to avoid spurious side-effects when setting in `invariant`. For others we've already established that `invariant` would have to cause the thread to write and this isn't used I guess.

Maybe we should just make `effect` or something similar to be the argument of `Priv.write_global` so it would be internal to a `Priv` what to do about it.

src/analyses/basePriv.ml
```diff
@@ -0,0 +1,1664 @@
...
+  module P =
+  struct
+    (* Note different Map order! *)
+    include MapDomain.MapTop_LiftBot (Basetype.Variables) (MinLocksets)
+    let name () = "P"
+
+    (* TODO: change MinLocksets.exists/top instead? *)
+    let find x p = find_opt x p |? MinLocksets.singleton (Lockset.empty ()) (* ensure exists has something to check for thread returns *)
+
+    (* TODO: no longer right? remove? *)
```
michael-schwarz, line 1154: Should be removed, is indeed not right anymore

src/analyses/basePriv.ml
```diff
@@ -0,0 +1,1664 @@
...
+  (* weak: G -> (S:2^M -> (W:2^M -> D)) *)
+  (* sync: M -> (S:2^M -> (W:2^M -> (G -> D))) *)
+  include AbstractLockCenteredGBase (GWeakW) (GSyncW)
+end
+
+module WriteCenteredD =
+struct
+  open Locksets
+
+  module W =
```
michael-schwarz, line 1139: Why does W and P have a different order here? The order should be the same, the only difference is the start value:
For `W` we start with `\bot` = `{}`, whereas for `P` we start with `\top` = `{{}}`.

But the order is always the same `the set of all upward-closed subsets of M, ordered by subset inclusion.`

michael-schwarz: Or this exactly what we achieve here, i.e., non-present bindings are `bot` in W, and `top` in P?

sim642: > the set of all upward-closed subsets of M, ordered by subset inclusion

This gives the order of `MinLocksets`.



> Why does `W` and `P` have a different order here?

The order for the map domain itself determines what happens when two branches are joined but one contains the key while the other doesn't. The `MapBot` vs `MapTop` determines what the default value for missing keys (and by extension behavior of the join with those missing keys) is. And hence the map domain orderings match up with the assumed start values of the inner `MinLocksets`.

michael-schwarz:
In general, I'm very surprised by how modular the different privatisations turned out, that is a very nice touch!

src/analyses/basePriv.ml
michael-schwarz: > Maybe we should just make effect or something similar to be the argument of Priv.write_global so it would be internal to a Priv what to do about it.

Yes, I think that sounds reasonable!

src/analyses/basePriv.ml
michael-schwarz: Alright, thanks for clarifying!

src/analyses/basePriv.ml
sim642: I remembered another worry I had with `is_private`. It used to be used for `make_entry`:
https://github.com/goblint/analyzer/blob/85536ab5c9cc0ffe5088af1c01eef34be27fa00e/src/analyses/base.ml#L1834

At some point I just removed that extra filtering on `traces` to get something else working and probably didn't see a change on the regression tests, but there might've been a reason for that use of `is_private` as well.

src/analyses/basePriv.ml
```diff
@@ -0,0 +1,1664 @@
...
+    let is_in_G x _ = is_global ask x in
+    let side_cpa = CPA.filter is_in_G st.cpa in
+    let side_cpa = CPA.mapi (fun x v ->
+        let v = distr_init getg x v in
+        v
+      ) side_cpa
+    in
+    sideg (mutex_addr_to_varinfo m) (G.create_sync (GSync.singleton s side_cpa));
+    (* m stays in v, l *)
+    (* TODO: why is it so imprecise now? *)
```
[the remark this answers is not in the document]
sim642: Adding the removal back now doesn't seem to make a difference to the regression tests at least. I suppose it did when I added the TODO, but that must've then be related to something else.

I'll just remove it. I suppose whatever extra precision it seemed to have might've been unsound in general.

src/analyses/basePriv.ml
sim642: Anyway, the `make_entry` use case wouldn't be enough of a reason to keep `is_private` around anyway, because the later change of making `make_entry` behave differently for `threadenter` via `Priv` should allow handling whatever the intended use case was internally in the `Priv`, just like the extra argument of `write_global` does, so I'm just going to make this change.

The case for the old privatization and `invariant` was even more subtle: `write_global` there doesn't `sideg`, but just adds to `CPA` and only the following `sync` does the side effecting. So `invariant` prevented the `CPA` change under some conflicting conditions where we just traced?

michael-schwarz:
Maybe we should also explicitly document under what circumstances globals become part of `cpa`?
I'm assuming that is the case for single-threaded mode (unless `exp.earlygob` is on)? In all other circumstances, they are

michael-schwarz:
Sorry, about the mess, I clicked on `Close without comment` instead of discarding my comment by mistake. :upside_down_face:

michael-schwarz:
Is there still anything blocking us from merging?

sim642:
Not from my side, so I suppose I'll do the merge?

Since nobody else would probably try or review this branch anyway, we won't find out about possible regressions until it's on `master` and everyone is using the changes.

PR #191, opened 2021-04-19
Fix unsoundness when thread escape analysis is disabled 

sim642:
It was discovered during the review of #183 that `is_global` handles escaped local variables unsoundly when escape analysis is disabled: https://github.com/goblint/analyzer/pull/183#pullrequestreview-638795674.

The fix attempts from there have been cherry-picked to this PR, since some things seem to rely on this unsound behavior. Probably within degenerate `ctx`s.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -19,7 +19,7 @@ module CArrays = ValueDomain.CArrays
 module BI = IntOps.BigIntOps
 
 let is_global (a: Q.ask) (v: varinfo): bool =
-  v.vglob || match a (Q.MayEscape v) with `MayBool tv -> tv | _ -> false
+  v.vglob || ThreadEscape.has_escaped a v
 
 let is_static (v:varinfo): bool = v.vstorage == Static
 
--- a/src/analyses/threadEscape.ml
+++ b/src/analyses/threadEscape.ml
@@ -6,12 +6,16 @@ open Analyses
 module M = Messages
 
 let has_escaped (ask: Queries.ask) (v: varinfo): bool =
-  match ask (Queries.MayEscape v) with
-  | `MayBool b -> b
-  | `Top ->
-    M.warn "Without thread escape analysis all variables are considered escaped, i.e. global";
-    true
-  | _ -> failwith "ThreadEscape.has_escaped"
+  assert (not v.vglob);
+  if not v.vaddrof then
+    false (* Cannot have escaped without taking address. Override provides extra precision for degenerate ask in base eval_exp used for partitioned arrays. *)
+  else
+    match ask (Queries.MayEscape v) with
+    | `MayBool b -> b
+    | `Top ->
+      M.warn "Without thread escape analysis all variables are considered escaped, i.e. global";
+      true
+    | _ -> failwith "ThreadEscape.has_escaped"
 
 
 module Spec =
--- a/src/analyses/threadEscape.ml
+++ b/src/analyses/threadEscape.ml
@@ -5,6 +5,15 @@ open Analyses
 
 module M = Messages
 
+let has_escaped (ask: Queries.ask) (v: varinfo): bool =
+  match ask (Queries.MayEscape v) with
+  | `MayBool b -> b
+  | `Top ->
+    M.warn "Without thread escape analysis all variables are considered escaped, i.e. global";
+    true
+  | _ -> failwith "ThreadEscape.has_escaped"
+
+
 module Spec =
 struct
   include Analyses.DefaultSpec
--- a/tests/regression/02-base/09-ambigpointer.c
+++ b/tests/regression/02-base/09-ambigpointer.c
@@ -1,4 +1,4 @@
-// PARAM: --set ana.activated "['base','threadid','threadflag','mallocWrapper']"
+// PARAM: --set ana.activated "['base','threadid','threadflag','escape','mallocWrapper']"
 #include<pthread.h>
 #include<assert.h>
 
--- a/tests/regression/02-base/47-no-threadescape.c
+++ b/tests/regression/02-base/47-no-threadescape.c
@@ -1,4 +1,4 @@
-// PARAM: --set ana.activated "['base','mallocWrapper','threadflag']"
+// PARAM: --sets ana.activated[-] escape
 #include <pthread.h>
 int g = 10;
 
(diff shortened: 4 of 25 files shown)


sim642:
Ok, this is interesting, the same fix on top of master passes all the tests, but failed partitioned array ones on `traces`.

michael-schwarz:
The problem is that `eval_exp` when it gets called to make `smart_join` from `VD.join` does not have access to `ctx` and hence uses a variant of `ctx.ask` that always returns `Top`.

https://github.com/goblint/analyzer/blob/687348903d6e162c9b917e79037f7261be6a7c33/src/analyses/base.ml#L793-L802

Therefore, `ThreadFlag.is_multi a` and `is_global a x` both return `true` and we go into the branch where we do `Priv.read_global a gs st x`. 
https://github.com/goblint/analyzer/blob/687348903d6e162c9b917e79037f7261be6a7c33/src/analyses/base.ml#L338-L344

(That `ThreadFlag.is_multi` also returns `true` here is a consequence of #130 that I did not consider at the time. Before, that flag was part of the base analysis and thus also available here.)

`Priv.Read_global` then checks if the global is protected (which it is not because the query functions will say `Top` everywhere), then uses `getg` to access its value, which will return `Top` and hence `smart_join` will not know that it must be equal to zero (or max index), and hence is not able to do a smart join, and the partitioned arrays don't work anymore.

https://github.com/goblint/analyzer/blob/687348903d6e162c9b917e79037f7261be6a7c33/src/cdomains/arrayDomain.ml#L467-L475

---
Before traces, on the other hand, we did:

https://github.com/goblint/analyzer/blob/1cee7911a16bb3a689f664f99a481f9d89732ba3/src/analyses/base.ml#L370-L390

and simply got the value from the local state.

michael-schwarz:
To address this issue, the most reasonable thing would be to make `ctx` accessible in the joins, then we could use the proper `ask` function there.

sim642:
Thanks for the detailed breakdown!

It looks to me like this issue isn't necessarily specific to `traces`, but could also arise in the old (`master`) implementation if all the following holds:
1. `is_multi` is sound (returns `true` when query doesn't say otherwise).
2. `is_global` is sound (returns `true` when query doesn't say otherwise about not escaping) as the result of this PR.
3. The local state doesn't contain the variable (it's not privatized).

Then `get_global` would also end up using `getg`, no?

A quick hack could be to make `eval_exp`'s degenerate `ask` to pretend to answer all escape queries with `false`. That would restore the previously used behavior, but only under `eval_exp`. It doesn't seem morally right though and might bury the unsoundness just deeper.

Is it somewhere explained why this "smartness" of joining is needed or how it fits together? I'm skeptical about making `ctx` or even just `ask` accessible to `join`, because it doesn't make much sense. Even from a purely technical point of view: `join` takes two domain elements, but which one would be used for `ctx.local` and by extension for `ctx.ask` (which may use `ctx.local`, like in the case of escape analysis)?

michael-schwarz:
~~~OCaml
let has_escaped (ask: Queries.ask) (v: varinfo): bool =
  if not (v.vglob || v.vaddrof) then
    false
  else
    match ask (Queries.MayEscape v) with
    | `MayBool b -> b
    | `Top ->
      M.warn_each "Without thread escape analysis all variables are considered escaped, i.e. global";
      true
    | _ -> failwith "ThreadEscape.has_escaped"
  ~~~
  
 If the invariant that a local that never had its address taken can not escape (which as far as I can see should still hold),
 this would fix it at least for the regression tests without requiring any bigger chances. 
 
 In other cases, it will only be less precise than one would want, but never unsound.

sim642:
I think that makes sense and we could stay with that solution for now.

The condition with `vglob` confused me though because `is_global` would call `has_escaped` only for non-syntactic globals anyway. So the wider question would be: what should `has_escaped` (and by extension escape analysis) do about syntactic globals? Maybe the appropriate behavior for now would be to `assert (not v.vglob)` and leave it at that until someone wants to determine escapedness of globals for whatever reason outside of `is_global`.

michael-schwarz:
> Then get_global would also end up using getg, no?

Yes, leading to unnecessary imprecision, which is not problematic per se.


> Is it somewhere explained why this "smartness" of joining is needed or how it fits together?

It is a hack to avoid having to make the base analysis path-sensitive as was the case in the version described in my thesis.[1]

The idea is that

~~~C
int i = 0;
int a[42];

while(i < 0) {
  a[i] = 17;
  i++;
}
~~~

at the loop head the join between

```
s1: { a-> (-,T), i-> [0,0] }
s2: { a->((i,[17,17],T,T)) i -> [1,1] }
```

should be `{a->(i,([17,17],T,T)), i -> [0,1]}` instead of `{a-> (-,T), i-> [0,1]}` and we preserve information about the array to the left of `i`.
The reason this is sound is because in s1 `i =_{must} 0`  and therefore, there are no elements to the left of `i` in the array `a` in s1.


> I'm skeptical about making `ctx` or even just `ask` accessible to join, because it doesn't make much sense. Even from a purely technical point of view: `join` takes two domain elements, but which one would be used for `ctx.local` and by extension for `ctx.ask` (which may use `ctx.local`, like in the case of escape analysis)?

There, one would need to make both ctx available, like `lctx` for the left argument, and `rctx` for the right argument. I agree it would be quite ugly to do that.

---
[1] For people with VCS access: https://versioncontrolseidl.in.tum.de/teaching/goblint-related-ba-or-ma-theses/-/blob/master/MA_2019_Sound-Yet-Precise-Modelling-of-Arrays_Michael-Schwarz.pdf

sim642:
Interesting, it reminds me of the after-join `sync Join` that we still have to do in #183 for branched thread creation. But the difference for partitioned arrays is that such `sync` wouldn't work, because that happens after the normal join, where this information is already completely lost. But the other way would work: push `sync Join` inside the domain's `join`.

Another interesting idea is that since `FromSpec` does the joining of all incoming edges, it could be calling a special `join` transfer function instead with entire contexts available. But it would be different from the domain's `join`, which would happen elsewhere (like the solver?) or would `Spec.D.join` not be used anywhere else?

It might not be the craziest idea, but we'd need to figure out the details, because `Lattice.join` shouldn't need to have additional arguments.

michael-schwarz:
>  Maybe the appropriate behavior for now would be to assert (not v.vglob) and leave it at that until someone wants to determine escapedness of globals for whatever reason outside of is_global.

Yes, this is what is the cleanest probably. I'll commit it later

sim642:
Heh, I just got around to it myself and tested it together with #183 to make sure it works there as well.

vogler:
I extended the warning a bit in cc4202984f8318af4c25b2fa012692abd836d179.
Feel free to adjust.

PR #194, opened 2021-04-20
Remove references to unmaintained CIL extensions

keremc:
CIL extensions partial, simplify & oneret will be removed. See goblint/cil#28


michael-schwarz:
According to the discussion in https://github.com/goblint/cil/issues/31, we don not need the `vidVisitor` after all, so I think we can remove it here too :smiley:

sim642:
In #204 I left it commented out with an explanation why, so nobody is tempted to add it back.

PR #196, opened 2021-04-22
Make octagon based on Apron useable

ivanazuzic:
- `OctApron` analysis and `OctApronDomain` are added
- `Update_suite.rb` distinguishes `UNKNOWN` (can be unknown) and `UNKNOWN!` (must be unknown)
- Base analysis takes into account the assertion results from other analysis that answer to `AssertionResult`
- `AssertionResult` is added as a new query
- Function signatures regarding ranges of data types are added to the interface of `IntDomain.Size`
- zip is moved from `base.ml` to `goblintutil.ml` because it's also used from `octApron.ml`


tests/regression/01-cpa/06-pointers.c
```diff
@@ -1,3 +1,5 @@
+// PARAM: --set ana.activated "['base' ,'mallocWrapper']"
```
sim642, line 1: The regression test for `01/06-pointers` fails because `escape` analysis is deactivated. I suppose you meant to have `--sets ana.activated[-] octApron` here as well?

src/analyses/base.ml
```diff
@@ -1898,7 +1894,38 @@ struct
...
-    match check_assert e ctx.local with
+    let meet_results a b = 
+        match (a, b) with
+        | (a, `Bot) -> `Bot
+        | (`Bot, b) -> `Bot
+        | (a, `Top) -> a
+        | (`Top, b) -> b
+        | (`True, `False) -> `Bot
+        | (`False, `True) -> `Bot
+        | (a, b) ->  a
```
sim642, line 1905: This seems quite annoying. I think we already have a flat boolean lattice somewhere (`Queries.AR` seems to be what you added) that has this operation. I guess the issue here was the special `True` and `False` instead of `Lifted true` and `Lifted false`. It might make more sense to switch from the former to the latter, so we don't have to duplicate the logic and convert between them below.

src/cdomains/apronDomain.ml
```diff
@@ -263,7 +263,7 @@ struct
         that come from the expression we wish to assert. *)
         A.meet_lincons_array Man.mgr d ea
       | None -> d
-    with Invalid_argument "cil_exp_to_lexp" -> d
+    with Invalid_argument _ -> d
```
sim642, line 266: Did you remove these strings just because the OCaml compiler warned about them or was there an actual reason?

Also, is `ApronDomain` actually used for anything here? If not and `OctApronDomain` is used instead, then I guess there's no reason to change (or even compile) this file at all.

src/cdomains/octApronDomain.ml
```diff
@@ -0,0 +1,611 @@
...
+    let ret = if is_bot x then
+      y
+    else if is_bot y then
+      x
+    (* This would make sense, but causes:
+      Fatal error: exception Apron.Manager.Error(_) *)
+    (* else if is_top x || is_top y then
+      top () *) 
+    else
+      A.join (Man.mgr) x y in
```
sim642, line 62: Why is it necessary to special case the `is_bot` and `is_top` things? Can `A.join` not handle them even though `A.is_bottom` and `A.is_top` also come from Apron?

src/cdomains/octApronDomain.ml
```diff
@@ -0,0 +1,611 @@
...
+  open Texpr1
+  (* Apron linear constraints of level 1 *)
+  open Lincons1
+
+  let typesort =
+    let f (is,fs) v =
+      if isIntegralType v.vtype then
+        if GobConfig.get_bool "ana.oct_no_uints" then 
+          match v.vtype with
+          | TInt(IUInt, i) -> (is, fs)
```
sim642, line 113: I guess other unsigned types should also be here?

src/cdomains/octApronDomain.ml
```diff
@@ -0,0 +1,611 @@
...
+    | _ -> 
+      let () = match check_assert e octa with
+        | `False ->
+          warn ~annot:"FAIL" ("{red}Assertion \"" ^ expr ^ "\" will fail.")
+        | `True ->
+          warn ("{green}Assertion \"" ^ expr ^ "\" will succeed");
+        | `Top ->
+          warn ~annot:"UNKNOWN" ("{yellow}Assertion \"" ^ expr ^ "\" is unknown.")
+      in
+      octa
```
sim642, line 396: If there is now the assertion query, isn't it unnecessary to duplicate the base analysis assertion logic here?

src/cdomains/octApronDomain.ml
```diff
@@ -0,0 +1,611 @@
...
+
+  (* Converts CIL expressions to Apron expressions of level 1 *)
+  let cil_exp_to_apron_texpr1 env exp =
+    (* ignore (Pretty.printf "exptotexpr1 '%a'\n" d_plainexp x); *)
+    Texpr1.of_expr env (cil_exp_to_cil_lhost exp)
+  
+  let is_chosen (v:string) =
+    let oct_vars =  List.map Json.jsonString (GobConfig.get_list "octagon_vars") in 
+    if List.length oct_vars == 0 then
+      true
```
sim642, line 406: Since the empty list surprisingly means all variables instead of no variables, it should at least be mentioned in the description of `octagon_vars` in `Defaults`.

src/domains/queries.ml
```diff
@@ -232,6 +247,15 @@ struct
...
       | (`PartAccessResult x, `PartAccessResult y) -> `PartAccessResult (PartAccessResult.join x y)
+      | (`AssertionResult x, `AssertionResult y) -> 
+        if AR.equal x y then
+          `AssertionResult x
+        else if AR.leq x y then
+          `AssertionResult y
+        else if AR.leq x y then
+          `AssertionResult x
+        else
+          `Bot
```
sim642, line 258: `AR.join` should already be able to handle this. Also the final case is probably wrong for joining, `Top` seems more appropriate instead. But using already implemented modules and functions would be better for avoiding bugs.

src/domains/queries.ml
```diff
@@ -250,6 +274,15 @@ struct
...
       | (`PartAccessResult x, `PartAccessResult y) -> `PartAccessResult (PartAccessResult.meet x y)
+      | (`AssertionResult x, `AssertionResult y) -> 
+        if AR.equal x y then
+          `AssertionResult x
+        else if AR.leq x y then
+          `AssertionResult x
+        else if AR.leq x y then
+          `AssertionResult y
+        else
+          `Bot
```
sim642, line 285: `AR.meet` should already be able to handle this.

src/maingoblint.ml
```diff
@@ -57,6 +57,11 @@ let create_temp_dir () =
 let remove_temp_dir () =
   if not (get_bool "keepcpp") then ignore (Goblintutil.rm_rf !Goblintutil.tempDirName)
 
+
+let rec print_list = function 
+  [] -> ()
+  | e::l -> print_int e ; print_string " " ; print_list l
+
```
sim642, line 64: Is this still needed for anything or just left over from some debugging?

sv-comp/benchexec/my-bench/goblint.xml
```diff
@@ -9,6 +9,7 @@
   <option name="--disable">ana.int.def_exc</option>
   <option name="--enable">ana.int.enums</option>
   <option name="--enable">ana.int.interval</option>
+  <option name="--disable">ana.oct_no_uints</option>
```
sim642, line 12: I wouldn't bother modifying this file, since it's not really used as we put all the SV-COMP configuration into `svcomp21.json` instead.

tests/regression/31-ikind-aware-ints/07-defexc-large.c
```diff
@@ -1,4 +1,4 @@
-//PARAM: --disable ana.int.interval --enable ana.int.def_exc --disable ana.int.enums
```
sim642, line 1: Since this test wants to explicitly test def_exc regardless of whether we may change defaults in the future, I'd be better not to remove these options.

sim642 (changes requested):
Some general big question that need to be decided:
1. **This PR currently adds Apron dependencies to our usual build. Is this something we want to do?**
    It'd be one of the notable dependencies we have while not necessary for most Goblint use cases so far. I'd have to check in a fresh opam switch, but I think Apron isn't as annoying (read: time-consuming) to install as Z3 was, so this probably isn't a major issue. I'm not against having it as a normal dependency, because having it available would keep the existing Apron code from rotting (as nobody is using such build) and would simplify the future integration of more Apron-based analyses.
2. **This PR currently adds `octApron` as a default analysis. Is this something we want to do?**
    I think not.
3. **There are numerous regression tests that fail with `octApron` and thus it has been deactivated (from the defaults which activate it). Should these failures be fixed before merging, even if the defaults are changed?**
    I think yes, because enabling additional (more precise) analyses shouldn't break what we already manage to analyze.

---

In conclusion, in the exact current shape this PR probably wouldn't be a good idea to have directly merged. But if we get it into a state where it doesn't impact current Goblint usage, then I'd be for it.

vogler:
Haven't looked through all the changes, but is there a reason for introducing `AssertionResult` instead of answering the existing `EvalInt` (and querying it in `Base...eval_rv_pre`)?

ivanazuzic:
> Haven't looked through all the changes, but is there a reason for introducing `AssertionResult` instead of answering the existing `EvalInt` (and querying it in `Base...eval_rv_pre`)?

I just thought that `EvalInt` has a different purpose than asserting. I thought it tells about the value of some expression and in my case I don'T know the exact value but the constraints about it and I can say whether something is impossible or possible, but not always the exact value. am I wrong?

tests/regression/01-cpa/06-pointers.c
ivanazuzic: But isn't it the same if I write:
// PARAM: --set ana.activated "['base' ,'mallocWrapper']" --set ana.activated[-] "'octApron'"
because 
// PARAM: --set ana.activated "['base' ,'mallocWrapper']" 
already means that activated analysis are only base and mallocWraper?

tests/regression/31-ikind-aware-ints/07-defexc-large.c
ivanazuzic: I agree.

src/maingoblint.ml
ivanazuzic: Leftover from debugging, sorry.

src/util/defaults.ml
```diff
@@ -135,6 +136,7 @@ let _ = ()
       ; reg Analyses "ana.sv-comp.functions" "false" "Handle SV-COMP __VERIFIER* functions"
       ; reg Analyses "ana.specification"   "" "SV-COMP specification (path or string)"
       ; reg Analyses "ana.wp"              "false" "Weakest precondition feasibility analysis for SV-COMP violations"
+      ; reg Analyses "ana.oct_no_uints"    "false"  "Use OctApron without tracking unsigned integers."
```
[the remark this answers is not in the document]
ivanazuzic: I thought the problem is that now they will much easier go to top because I forget about them if they might be out of the bounds for their type, but just forgetting about them by default is not any better. Am I wrong?

vogler:
If you can tell whether the assertion holds, you can tell the value of the expression, no?
`assert(exp);`: `typeof exp = int`
If `exp` is a comparison you return 1 if it must hold, 0 if it must not hold, and [0,1] if you don't know.
Other expressions may evaluate to more than [0,1].

The more general question is: `base` does a lot of stuff that would profit from more precise integers, so why only use the information for assert?

sim642:
I guess Ralf meant using the entire assertion expression, e.g. `x > y` in `assert(x > y);` as the argument for `EvalInt`. Then that in octApron should evaluate to 0 if false, non-0 if true and unknown int if unknown. Int domains have a bunch of functions for dealing with booleans as integers.

It might be a good idea to go that way, because suppose you have the same expression somewhere else, e.g. `foo(x > y);` then base analysis might also benefit from octApron knowing the definite result there to get a more precise context.

tests/regression/01-cpa/06-pointers.c
sim642: I mean you don't need the `--set ana.activated "['base' ,'mallocWrapper']"` part at all. At least previously it worked fine with mutex, escape, etc defaults being activated. So if it fails just because of octApron right now, you can use your added syntax to deactivate it without having an incomplete analysis list which breaks the test.

src/analyses/octApron.ml
```diff
@@ -0,0 +1,182 @@
...
+    if D.is_bot ctx.local then D.bot () else
+      let vars = D.typesort f.slocals in
+      D.add_vars ctx.local vars
+
+  let assign ctx (lv:lval) e =
+    if D.is_bot ctx.local then D.bot () else
+      match lv with
+      | Var v, NoOffset when isArithmeticType v.vtype && (not v.vglob) -> 
+        D.assign_var_handling_underflow_overflow ctx.local v e
+      | _ -> D.topE (A.env ctx.local)
```
michael-schwarz, line 158: Maybe one wants to handle more things here? As it is, any assignment to a variable that is not a local variable without offset results in the entire octagon going to `top`?

Also, how do you handle escaping local variables?

src/cdomains/apronDomain.ml
michael-schwarz: Actually, that was my suggestion here. I think we should aim to treat warnings as errors.

src/cdomains/octApronDomain.ml
michael-schwarz: Same for `meet` and `widen`. Or is this a performance optimization?

michael-schwarz (changes requested):

src/analyses/octApron.ml
michael-schwarz: One could, e.g., limit oneself to only locals of arithmetic type that do not have their addresses taken. Then one can safely ignore all other assigns.

src/cdomains/apronDomain.ml
sim642: The sensible solution is then to make a new `exception` type that `cil_exp_to_lexp` raises and catch that specifically. Catching arbitrary `Invalid_argument` is dangerous, because many standard library functions also raise those, so if there's accidentally such coding error inside `cil_exp_to_lexp`, then it would be swallowed here.

goblint.opam
```diff
@@ -13,6 +13,7 @@ homepage: "https://goblint.in.tum.de"
 doc: "https://goblint.github.io/analyzer"
 bug-reports: "https://github.com/goblint/analyzer/issues"
 depends: [
+  "apron"
```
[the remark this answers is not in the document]
sim642: I think that's the right solution here, I wasn't aware optional dependencies being a thing in opam. It'd be especially nice if they worked automagically together with dune as well, i.e. the Apron dependent modules are only compiled when the opam optional dependency is present. That way we don't have to manually modify some dune etc files to enable their compilation.

src/util/defaults.ml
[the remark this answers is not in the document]
sim642: If the value `false` is sound and doesn't cause any other issues, then it's fine. I thought I noticed some other place, where this option was explicitly set to `true`, but I might be wrong.

tests/regression/01-cpa/06-pointers.c
ivanazuzic: It works fine with defaults, regardless of the octApron, so I removed the params.

src/cdomains/octApronDomain.ml
ivanazuzic: This logic originates from poly. I thought it might be some performance optimization, but if I just return `A.join (Man.mgr) x y` then I get `Fatal error: exception Apron.Manager.Error(_)`, which is not explaining the exception very well. The problem is when the apron manager tries to join something with bottom. If it joins top and bottom, only then it works.

sim642: According to Apron documentation there should be some information with that exception: https://antoinemine.github.io/Apron/doc/api/ocaml/Manager.html#EXCEPTIONError. I suppose you could try catching it and using `print_exclog` to display it. Maybe that explains something.

ivanazuzic: Thank you for the suggestion. Printing the exclog explained the reason for the Error. the error message was: `"The two abstract values have different environments."` That exception comes from [here](https://github.com/antoinemine/apron/blob/28e9c36fe6f64c2365b2535ca28f55673d7858e9/apron/ap_abstract1.c#L106). The environments are compared [here](https://github.com/antoinemine/apron/blob/28e9c36fe6f64c2365b2535ca28f55673d7858e9/apron/ap_environment.c#L657). It happens in `32-widen-context/01-on.c` in line 12 where an octagon with constraints (its environment is filled with variables) is joined with an octagon which is `bottom` (has an empty environment). I've talked about it with Michael and the conclusion was that this happens when we join with dead code.
This same if-based structure with `is_top` and `is_bottom` is there for `join`, `meet`, `widen`, `narrow`, `equal`, `leq` for the same reason: to avoid doing an operation with octagons who have different environments. This overhead code could be:

1. removed and only the Apron function would be used, but then some sneaky hacks need to be made to have equal environments (maybe adding the context from myCFG.current_node to bottom before joining), so the code would be shorter, but maybe less readable/logical
2. moved to so util function to shorten the code at least

Any advice about what to do @sim642 ?

sim642: Since it's necessary, I think it's fine to keep it this way, but there should at least be a comment above explaining why (`bot` has an octagon with incompatible environment or whatever).

Also a good idea might be to register the special Apron exception printer with OCaml's exception printing: https://ocaml.org/learn/tutorials/error_handling.html#Printing. That way it should be possible to make unhandled Apron exceptions automatically print our all their information without having to put any try-with constructs all over these operations.

ivanazuzic:
> Haven't looked through all the changes, but is there a reason for introducing `AssertionResult` instead of answering the existing `EvalInt` (and querying it in `Base...eval_rv_pre`)?

@vogler Asking about `EvalInt` form `eval_int_pre` causes an infinite recursion because `base` also asks itself about `EvalInt` which calls `eval_rv` which calls `eval_int_pre` and that repeats and repeats. even though `eval_int_pre` was meant to serve the purpose of asking other analysis about their useful knowledge, simply asking for `EvalInt` there is not a good solution.
Since there is no option to tell `ask` to ignore a response of certain analysis, there could be the following solutions:

1. Giving an expression and a boolean to `EvalInt` and the boolean would be a flag that indicates if it's allowed to ask `base`. I don't like this because I'd have to change over 30 places where `EvalInt` is used. I also don't think it would be clear what that boolean is to a person who doesn't know this particular problem. Then the boolean could be replaced by a named type (type dont_ask_base=bool) but that would look bad as well.
2. `eval_rv` could get a new optional parameter that disallows ask and this way a cycle would not happen. When `base` gets asked about `EvalInt` it will only answer what it knows if `eval_rv` has a true there. In other places where we have calls to `eval_rv` the argument would have a default value false. I'll implement this solution and make a commit. If you know a better one, please propose it.

sim642:
>     1. I also don't think it would be clear what that boolean is to a person who doesn't know this particular problem. Then the boolean could be replaced by a named type (type dont_ask_base=bool) but that would look bad as well.

This is why I've made the more complex queries use inline records in their variant, which allows naming the components in a clear way: https://github.com/goblint/analyzer/blob/11dcafdfd30fac8ba089840a633461ee08ddf3d1/src/domains/queries.ml#L43-L45

But that's not too relevant since solution 2 seems nicer and more contained anyway. Having a query refer to a particular analysis (e.g. `dont_ask_base`) doesn't look so good in a query system, which shouldn't really care about which analysis is answering.

---

This is getting off-topic, but just something to think about for the future: if each query were also a constraint system variable, i.e. (node, context, query), which stores the result of the query and is computed on demand, then cycles in queries would be naturally handled by our generic solvers. And it would reuse the computed result if multiple analyses make the same query during the same transfer function.

vogler:
Does it ever make sense for an analysis to answer itself a query?
If not, maybe this can be excluded here by storing `n` of the `ask`ing analysis in `ctx'` and yielding top instead of calling `query`.
https://github.com/goblint/analyzer/blob/11dcafdfd30fac8ba089840a633461ee08ddf3d1/src/analyses/mCP.ml#L527-L536

sim642:
> Does it ever make sense for an analysis to answer itself a query?
> If not, maybe this can be excluded here by storing `n` of the `ask`ing analysis in `ctx'` and yielding top instead of calling `query`.

I can't immediately think of an example, so maybe not for the query of the same type, but it might still be useful if the response to one query depends on a different query, which can be answered by both the same analysis or another one. This wouldn't necessarily be a cycle. But it might be quite confusing if the same query could yield different results depending on which analysis is making the query.

vogler:
For each analysis' `ctx` we could have a set of `ask`ed queries and a cache for the result. It would first try the cache, then return top if in the set or otherwise evaluate.
That should allow self-querying and break up cycles.

src/cdomains/octApronDomain.ml
ivanazuzic: The comment has been added and the suggestion about registering a printer is moved to #237.

sim642:
OctApron fails two new tests which I extracted from existing non-apron tests:
```
Expected unknown!, but registered fail on branch:11
    assert(i == 2); // UNKNOWN!
24/19 octagon/branch failed!

Expected unknown!, but registered success on float:16
    assert(top == top); //UNKNOWN!
24/20 octagon/float failed!
```

### branch
In the branch test, the octApron state in the if is flipped from what it should be! It states `[|i-1.>=0; -i+1.>=0|]` which essentially is `i-1==0`. In general the `branch` function isn't flipped (removing a `not` makes other octApron tests fail). But I think the issue is in this code:
https://github.com/goblint/analyzer/blob/134fd40fab9bc305d47676a8c0451d66af786f96/src/cdomains/apron/octApronDomain.apron.ml#L272-L278
The `!= 0` extension shouldn't just be done to single variables, but whatever expression. Making it unconditional causes some stack overflow though. Maybe it needs to be somehow conditional like: if it's already a "boolean" expression, don't do it, otherwise do.

This is quite crucial because it's unsound. By constructing the linear constraint from `i-1` it by default means `i-1==0` if I understand the Apron stuff correctly.


### float
I think octApron also tries to handle floats because Apron supposedly does. But it seems like it's oblivious to NaN, causing the same failure as varEq in #141.

sim642 (approved):
I think it should be in mergeable condition now. The optional Apron dependency and build seems to work. Octagon tests are skipped by default to not cause failures for anyone who doesn't build with Apron. Nevertheless, they can be explicitly executed (using the `group` "feature") and the CI also builds with Apron and runs them.

I realized just now that because Apron is included in the opam lockfile, such that the CI would install it for testing, it will also be installed thanks to the lockfile via `make dev`. So any new developers would still have Apron. But the optional dependency means that this PR at least doesn't cause build failures for existing developers. Not sure if we want to change that (remove apron from lockfile and install apron manually in CI script)?

src/util/defaults.ml
```diff
@@ -116,6 +117,7 @@ let _ = ()
       ; reg Analyses "ana.osek.flags"      "[]"    "List of global variables that are flags."
       ; reg Analyses "ana.osek.def_header" "true"  "Generate TASK/ISR macros with default structure"
       ; reg Analyses "ana.int.wrap_on_signed_overflow" "false" "Whether to assume wrap-around behavior on signed overflow. If set to true, assumes two's complement representation of signed integers. If set to false, goes to top on signed overflow."
+      ; reg Analyses "ana.int.no_signed_overflow" "true" "Assume there will be no signed overflow."
```
michael-schwarz, line 120: Why is this the default?

michael-schwarz:
Something else that springs to mind:

What is the relationship with the other octagon analysis made by a student when it comes to the tests?
The tests in folder 24 used to be for that, but now `octApron` somehow reuses them removing any tests for the old one, and also skipping some of them?
Should the tests for `octApron` not be their own thing (in a new folder) that leave the other tests alone? (At least until we decide to remove the other octagons?)

vogler:
> What is the relationship with the other octagon analysis made by a student when it comes to the tests?
> The tests in folder 24 used to be for that, but now `octApron` somehow reuses them removing any tests for the old one?
> Should the tests for `octApron` not be their own thing (in a new folder) that leave the other tests alone? (At least until we decide to remove the other octagons?)

Probably yes.
We should have a way to check the same test with multiple configurations to avoid duplicating files.
Would also be good to run several solvers on the same test. E.g. https://github.com/goblint/analyzer/blob/master/tests/regression/34-localization/01-nested.c should now also be tested with `td3` after https://github.com/goblint/analyzer/commit/cf11c4b0b3602f417a5573e56853b81940384998.

We could just have several lines starting with `PARAM: ` or something like `PARAM+:` which just appends/overrides the previous config.

sim642:
> You mean `make deps` (and `make setup`)?
> Does Apron have any extra system dependencies?
> If it doesn't complicate installation, it's probably fine to have it in the lockfile.

Sorry, yes, `make setup`. It has the extra `libmpfr-dev` system dependency, but that's also listed with `make setup` (and installed in CI), so that shouldn't be a problem.


> So it's only optional for a release to opam or if someone installs without `--locked`?
> For CI and benchmarks `--locked` makes sense, but for user installations we could also do without. That would also test if the package constraints are set right. But maybe not worth dealing with if everyone just installs it in a local switch without problems anyway.

Yes. User installations being without lockfile is an interesting idea too. Because if everyone's using the lockfile, then no dependency will ever get updated and we end up with unnecessarily ancient (and possibly more buggy) versions of things.


> What is the relationship with the other octagon analysis made by a student when it comes to the tests?
> The tests in folder 24 used to be for that, but now `octApron` somehow reuses them removing any tests for the old one?
> Should the tests for `octApron` not be their own thing (in a new folder) that leave the other tests alone? (At least until we decide to remove the other octagons?)

Yes, right now all of them are just used to test octApron instead. I can make the octApron tests a separate copy if that's desired. But realistically I think we won't end up maintaining that octagon analysis anyway. And currently the old apronDomain/poly stuff is also around, but given how octApronDomain/octApron is more or less a fixed copy, the old one should probably get removed as well (and the fixed one generalized to any Apron domain eventually).


> We should have a way to check the same test with multiple configurations to avoid duplicating files.

I've thought about this for a while. But the octApron case is further complicated by the fact that the availability and testability of its configuration depends on whether apron is installed.

michael-schwarz:
> I can make the octApron tests a separate copy if that's desired. 

I would go with this for now, given that afaik we have not tested the octApron for anything but toy examples?

> But realistically I think we won't end up maintaining that octagon analysis anyway. 

I agree. Once we are happy with octApron we probably want to remove the other one.

src/util/defaults.ml
michael-schwarz: I am not suggesting that we should not have this for octagons. However, this is decidedly not what all the other domains (e.g. intervals) do, where we very pain-stakingely made sure to go to top in such cases (which imo is the philosophically sounder thing (whatever sound means for programs with UB))

In general for all of our domains, I am against making the assumption of no signed overflows by default.

For the octagons, one might want this behavior though. As currently this is only used by octagons, I would suggest renaming the option to reflect that.

src/analyses/apron/octApron.apron.ml
```diff
@@ -0,0 +1,192 @@
...
+              | Some fnc -> let () = invalidate ctx.local (fnc `Write  args) in ctx.local
+              | None -> D.topE (A.env ctx.local)
+            in
+              st
+          end
+      end
+
+  let branch ctx e b =
+    if D.is_bot ctx.local then
+      D.bot ()
```
michael-schwarz, line 126: why not `raise Deadcode`?

sim642: To be honest, every one of the transfer functions (already in poly) starts with `if D.is_bot ctx.local then D.bot ()` for some reason. I don't know why this is because transfer functions shouldn't be called on already dead states anyway. For example, we don't do that sort of extra check in base.

Maybe @ivanazuzic knows why (or whether at all) these are necessary.

michael-schwarz (approved):

src/util/defaults.ml
sim642: > For the octagons, one might want this behavior though. As currently this is only used by octagons, I would suggest renaming the option to reflect that.

I'll rename this option (and other octapron options) into an option group like `ana.octapron`.

PR #203, opened 2021-04-22
Use variant type group in `MapDomain.Groupable` with `deriving show`

vogler:
Follow up of comments in https://github.com/goblint/analyzer/commit/0d818dcf34b1630812d1e8ee9ca90ab21f506921.
Instead of `int` this uses a `type group` with [ppx_deriving](https://github.com/ocaml-ppx/ppx_deriving).

Old:
~~~ocaml
let classify x = match x with
  | x when x.vglob -> 2
  | x when x.vdecl.line = -1 -> -1
  | x when x.vdecl.line = -3 -> 5
  | x when x.vdecl.line = -4 -> 4
  | _ -> 1
let class_name n = match n with
  |  1 -> "Local"
  |  2 -> "Global"
  |  4 -> "Context"
  |  5 -> "Parameter"
  | -1 -> "Temp"
  |  _ -> "None"
~~~
New:
~~~ocaml
type group = Temp | Local | Global | Context | Parameter [@@deriving show { with_path = false }]
let to_group = Option.some @@ function
  | x when x.vglob -> Global
  | x when x.vdecl.line = -1 -> Temp
  | x when x.vdecl.line = -3 -> Parameter
  | x when x.vdecl.line = -4 -> Context
  | _ -> Local
~~~

Since it's in `Groupable`, I got rid of the `class*` names. Also, easier to search for `to_group` since there are many `LibraryFunctions.classify` which have nothing to do with this.

Tests pass, but I don't know if this changed some output in a wrong way.
I assume in the worst case, some things might not be grouped anymore.
However, with `include Printable.Std` it should complain if something is only overridden partially.

### DIFF
--- a/src/domains/mapDomain.ml
+++ b/src/domains/mapDomain.ml
@@ -58,11 +58,9 @@ end
 module type Groupable =
 sig
   include Printable.S
-  type group (* use [@@deriving show { with_path = false }, enum] *)
+  type group (* use [@@deriving show { with_path = false }] *)
   val show_group: group -> string
-  val group_to_enum: group -> int
-  val group_of_enum: int -> group option
-  val to_group: t -> group
+  val to_group: (t -> group) option
   val trace_enabled: bool (* Just a global hack for tracing individual variables. *)
 end
 
@@ -143,12 +141,12 @@ struct
 
   let pretty_f short () mapping =
     let groups =
-      let group_fold key itm gps =
-        let cl = Domain.to_group key |> Domain.group_to_enum in
-        match gps with
-        | (a,n) when cl <>  n -> ((cl,(M.add key itm M.empty))::a, cl)
-        | (a,_) -> ((fst (List.hd a),(M.add key itm (snd (List.hd a))))::(List.tl a),cl) in
-      List.rev (fst (fold group_fold mapping ([],min_int)))
+      let h = Hashtbl.create 13 in
+      let opt_apply k = Option.bind Domain.to_group (fun to_group -> Some (to_group k)) in
+      iter (fun k v -> BatHashtbl.modify_def M.empty (opt_apply k) (M.add k v) h) mapping;
+      let cmpBy f a b = Stdlib.compare (f a) (f b) in
+      (* sort groups (order of constructors in type group)  *)
+      BatHashtbl.to_list h |> List.sort (cmpBy fst)
     in
     let f key st dok =
       if ME.tracing && trace_enabled && !ME.tracevars <> [] &&
@@ -159,11 +157,11 @@ struct
                   dprintf "%a -> \n  @[%a@]\n") Domain.pretty key Range.pretty st
     in
     let group_name a () = text (Domain.show_group a) in
-    let pretty_group  map () = fold f map nil in
-    let pretty_groups rest map =
-      match (fst map) with
-      | -1 ->  rest ++ pretty_group (snd map) ()
-      | a -> rest ++ dprintf "@[%t {\n  @[%t@]}@]\n" (group_name (Domain.group_of_enum a |> Option.get)) (pretty_group (snd map)) in
+    let pretty_group map () = fold f map nil in
+    let pretty_groups rest (group, map) =
+      match group with
+      | None ->  rest ++ pretty_group map ()
+      | Some g -> rest ++ dprintf "@[%t {\n  @[%t@]}@]\n" (group_name g) (pretty_group map) in
     let content () = List.fold_left pretty_groups nil groups in
     dprintf "@[%s {\n  @[%t@]}@]" (short 60 mapping) content
 
--- a/src/domains/mapDomain.ml
+++ b/src/domains/mapDomain.ml
@@ -162,7 +162,7 @@ struct
     let pretty_group  map () = fold f map nil in
     let pretty_groups rest map =
       match (fst map) with
-      | 0 ->  rest ++ pretty_group (snd map) ()
+      | -1 ->  rest ++ pretty_group (snd map) ()
       | a -> rest ++ dprintf "@[%t {\n  @[%t@]}@]\n" (group_name (Domain.group_of_enum a |> Option.get)) (pretty_group (snd map)) in
     let content () = List.fold_left pretty_groups nil groups in
     dprintf "@[%s {\n  @[%t@]}@]" (short 60 mapping) content
--- a/src/domains/mapDomain.ml
+++ b/src/domains/mapDomain.ml
@@ -58,7 +58,7 @@ end
 module type Groupable =
 sig
   include Printable.S
-  type group (* use [@@deriving show, enum] *)
+  type group (* use [@@deriving show { with_path = false }, enum] *)
   val show_group: group -> string
   val group_to_enum: group -> int
   val group_of_enum: int -> group option
--- a/src/domains/mapDomain.ml
+++ b/src/domains/mapDomain.ml
@@ -63,7 +63,7 @@ sig
   val group_to_enum: group -> int
   val group_of_enum: int -> group option
   val to_group: t -> group
-  val trace_enabled: bool
+  val trace_enabled: bool (* Just a global hack for tracing individual variables. *)
 end
 
 module PMap (Domain: Groupable) (Range: Lattice.S) : PS with
--- a/src/domains/mapDomain.ml
+++ b/src/domains/mapDomain.ml
@@ -60,7 +60,7 @@ sig
   include Printable.S
   type group (* use [@@deriving show { with_path = false }] *)
   val show_group: group -> string
-  val to_group: (t -> group) option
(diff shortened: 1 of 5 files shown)


vogler:
https://github.com/goblint/analyzer/blob/9def498985ca2b10a9841133ee8fac9999bcc546/src/domains/mapDomain.ml#L61-L64

Since variants are ordered, the enum stuff could be removed if the following is rewritten to not rely on `int`:

https://github.com/goblint/analyzer/blob/9def498985ca2b10a9841133ee8fac9999bcc546/src/domains/mapDomain.ml#L145-L151

src/domains/mapDomain.ml
```diff
@@ -58,19 +58,14 @@ end
...
   val trace_enabled: bool
 end
 
-module StripClasses (G: Groupable) =
-struct
-  include G
-  let classify x = 0
-end
-
-(* Just a global hack for tracing individual variables. *)
```
sim642, line 72: I've always wondered what this comment is referring to, but now I think it's about the `Groupable.trace_enabled`. Maybe it should just be moved there?

src/domains/mapDomain.ml
```diff
@@ -149,7 +144,7 @@ struct
   let pretty_f short () mapping =
     let groups =
       let group_fold key itm gps =
-        let cl = Domain.classify key in
+        let cl = Domain.to_group key |> Domain.group_to_enum in
         match gps with
         | (a,n) when cl <>  n -> ((cl,(M.add key itm M.empty))::a, cl)
         | (a,_) -> ((fst (List.hd a),(M.add key itm (snd (List.hd a))))::(List.tl a),cl) in
```
sim642, line 150: I tried to understand what it's doing, but it doesn't make sense as it is. The second case then has `n = cl` essentially. So the second component of the fold pair will always be the previous key's group. But then:
1. What about keys from alternating groups? Nothing about the groups is tied to the ordering (key `compare`) of the map, so it would only group consecutively?
2. Why is the second case consing a pair with `fst (List.hd a)` instead of just `cl` (the key's group)? How can it be anything other than `cl`?

src/domains/mapDomain.ml
```diff
@@ -163,12 +158,12 @@ struct
         dok ++ (if Range.isSimple st then dprintf "%a -> %a\n" else
                   dprintf "%a -> \n  @[%a@]\n") Domain.pretty key Range.pretty st
     in
-    let group_name a () = text (Domain.class_name a) in
+    let group_name a () = text (Domain.show_group a) in
     let pretty_group  map () = fold f map nil in
     let pretty_groups rest map =
       match (fst map) with
       | 0 ->  rest ++ pretty_group (snd map) ()
```
sim642, line 165: If I'm not mistaken, then this `fst map` comes from `groups` as some group number. But then group number `0` is special because it's printed differently?

Previously `Basetype.Variables` didn't have group 0, but now it has and changes the output?

src/domains/mapDomain.ml
vogler: Probably. Moved.

src/domains/mapDomain.ml
vogler: Yes, this code is way too convoluted and should be rewritten.
Is there some test that has enough grouped output that can be used to test it?

src/domains/mapDomain.ml
vogler: Mh, yes, seems like 0 means no group. Since `group_to_enum` starts with 0, I'll change this to -1 in `Printable.Std`.

src/domains/mapDomain.ml
sim642: Printing a `CPA` that contains a few local variables, global variables and function parameters should be observable from a fairly simple example, but I haven't actually tried. Also, I'm not sure how to get context and temp groups from a program, but for testing the grouping, a few of the above should be sufficient.

I've definitely seen them grouped in some tracing output. But I don't remember groups alternating, so maybe it currently handles proper grouping but I just didn't understand how.

src/domains/mapDomain.ml
sim642: If we want to get rid of `int`-based groups and use `deriving ord` instead, then I suppose `to_group` would have to return `group option` instead.

src/domains/mapDomain.ml
vogler: I just inserted a print in `MapDomain.LiftTop()().add` and it seems like it's working.
The `fold` traverses it in a fixed order such that the groups are consecutive. Maybe it relies on the keys' compare to ensure this.
The `group_to_enum` values don't affect the order - maybe it should be sorted by them afterwards?

src/domains/mapDomain.ml
vogler: Yes. However, like this it's an ok trade-off I think. We only have this `let group_to_enum () = -1` in `Printable.Std` - places that use groups can derive it. If we change `to_group` to `group option` you have the problem that users can return both `Some` and `None` which makes no sense. (And lots of `Some` in the default override case.)

vogler: Don't need `deriving ord` to get rid of `deriving enum` since constructors are `compare`d in order of definition.
Just need a reduce that starts with the first element instead of the fold.

src/domains/mapDomain.ml
vogler: Rewrote it. Should work with non-consecutive groups and sorts by the variant type's order.

src/domains/mapDomain.ml
vogler: Got rid of `enum` and changed it to `val to_group: (t -> group) option` which is safe and not too cumbersome to write.

vogler:
Got rid of `enum` and rewrote grouping in https://github.com/goblint/analyzer/pull/203/commits/8ad8625d3b8f8cdc72e99e18ab984a76646d0fcf.

src/domains/mapDomain.ml
sim642: The type `(t -> group) option` seems a bit weird. Why not have it as `t -> group option`? It would both be easier to work with and would match the current possibilities of groups, i.e. you can have ungrouped keys in a map, which may also contain some grouped things. I think none of the existing cases do that and it's probably not terribly useful, but the optional function is quite unusual too.

And then in the `Std` case, we could have `type group = |` ([empty variant](https://ocaml.org/releases/4.12/htmlman/emptyvariants.html)), which has no values. That would remove the weird alternative `Std.to_group`, which would give the unit group instead of no group.

src/domains/mapDomain.ml
sim642: While working on https://github.com/goblint/cil/issues/31 I think I figured out why the old implementation seemed to not have multiple non-consecutive copies of the same group in the output. The key ordering of `CPA` is based on `vid`s and by using `Partial.globally_unique_vids` those just happened to be renumbered in a way that all the parameters were numbered consecutively and then all the local variables were numbered consecutively etc.

sim642: I just checked that indeed the old implementation doesn't group non-consecutively. For example here the global `main` was introduced by Goblint (after all the CIL stuff) for the return value of the thread (not the function) and ends up in a duplicate block:
```
mapping {
  Global {
    g -> (36)
  }
  Temp {
    RETURN -> (0)
  }
  Global {
    main -> (0)
  }
}
```

After this PR the globals are actually together which makes much more sense IMO.

src/domains/mapDomain.ml
vogler: The question is if we want to allow having ungrouped mixed with grouped.
Would look like this (`None` instead of `Some Temp`):
~~~
mapping {
  tmp -> (0)
  tmp___0 -> (0)
  tmp___1 -> (1)
  tmp___2 -> (0)
  tmp___3 -> (1)
  tmp___4 -> (Unknown int([-31,31]))
  tmp___5 -> (Unknown int([-31,31]))
  tmp___6 -> (Unknown int([-31,31]))
  Global {
    __stdinp ->
      {NULL, ?}
    __stdoutp ->
      {NULL, ?}
    __stderrp ->
      {NULL, ?}
    sys_nerr -> (Unknown int([-31,31]))
    sys_errlist ->
      (Array: {}, (Unknown int([-63,63])))
    glob1 -> (5)
    glob2 -> (9)
  }
  Local {
    i -> (1)
    j -> (0)
    k -> (1)
  }
}
~~~
No strong opinion.

> the optional function is quite unusual too

In our codebase yes. It's just the apply that's harder to understand, the definitions don't differ much.

Changed it in https://github.com/goblint/analyzer/pull/203/commits/8ad23513351c985eda799cbf2c7cc642f713cbfd

sim642: I think it wouldn't hurt to have the possibility of mixing them, especially since it doesn't make the code any more sophisticated to have the flexibility.

sim642 (approved):

Issue #205, opened 2021-04-26
non-terminating benchmarks


vogler:
The following programs in [bench/single-thread] don't terminate:
- 400.perlbench_comb.c
- 445.gobmk_comb.c
- wget-1.12_comb.c

[bench/single-thread]: https://github.com/goblint/bench/tree/master/single-thread

The programs in [bench/coreutils] all terminate.

[bench/coreutils]: https://github.com/goblint/bench/tree/master/coreutils

Configuration:
`./goblint -v single-thread/wget-1.12_comb.c --enable exp.solver.td3.term --enable exp.earlyglobs --enable ana.int.interval --disable ana.int.enums --disable ana.int.def_exc --enable exp.widen-context --sets exp.privatization none`

With that configuration, all programs should terminate. The question is why they don't and how to debug it.

The following only terminate with `--enable exp.widen-context` and are therefore likely recursive:
- coreutils/sort_comb.c
- single-thread/458.sjeng_comb.c
- single-thread/482.sphinx_livepretend_comb.c
- single-thread/duff-0.5_comb.c
- single-thread/maradns-1.4.06_comb.c

michael-schwarz:
Still the case also with current Goblint.

michael-schwarz:
@SchiJoha would maybe be interesting to find out if your strategies from the new PR can enforce termination here?

michael-schwarz:
The non-termination could potentially also be due to paths here, for which we don't have any mechanism to bound their number.

Issue #172, opened 2021-03-18
Dubious dbg.timeout behavior


sim642:
I was running some benchmarks from [goblint/bench](https://github.com/goblint/bench) and noticed some running far longer than they should. The benchmark script there uses `--set dbg.timeout 900`, which I suppose should make Goblint time out after 900 seconds (15 minutes). On multiple occasions I saw no progress for a long time. When I looked at the `goblint` process in `htop` to kill it myself, I noticed it having used over 30 minutes of CPU time. The builtin timeout hadn't worked properly!

Looking at the code, there's some weird logic surrounding timeout:

1. Timeout is set (using SIGALRM, which I assume should work reliably) to call `Messages.waitWhat` after the given time:
https://github.com/goblint/analyzer/blob/956b9faaef2514cb530e6d9b8711106b3e3a6ab3/src/framework/control.ml#L514-L515
2. This function "stops the world" by raising `StopTheWorld` and also sets a boolean flag (to be sure?):
https://github.com/goblint/analyzer/blob/956b9faaef2514cb530e6d9b8711106b3e3a6ab3/src/util/messages.ml#L98-L103
3. It would all be great except transfer functions in `FromSpec` actually catch that exception and completely ignore the timeout having happened by evaluating to bottom:
https://github.com/goblint/analyzer/blob/956b9faaef2514cb530e6d9b8711106b3e3a6ab3/src/framework/constraints.ml#L691-L692
4. There is some extra logic in `common_ctx` of `FromSpec` that checks the boolean flag on every transfer function evaluation (why use the language's exception handling mechanism if we can just do it ourselves C-style, right?) and re-raises the exception:
https://github.com/goblint/analyzer/blob/956b9faaef2514cb530e6d9b8711106b3e3a6ab3/src/framework/constraints.ml#L530-L532

I see two issues with this:
1. When the timeout happens to occur exactly in the middle of transfer function evaluation, it gets ignored and everything continues running. I really don't understand why the exception is being caught there at all.
2. The re-raising in `common_ctx` appears as if it's there to work around the first issue, but it's completely broken: since `common_ctx` is called within `tf`. The exception it re-raises gets caught and ignored by the same handler. So not only does the analysis continue running after timeout, it can also make no progress and gets stuck in a loop, because the boolean flag causes the exception to be re-raised and ignored every time the solver evaluates a transfer function.

[...]

PR #207, opened 2021-04-26
Statistics in case of timeout, automatic colors, backtrace on SIGUSR2, `dbg.solver-stats-interval`

vogler:
To debug non-terminating benchmarks (see #205), it would be good to have some statistics in case of a timeout/abort.
Either catch the `Timeout` exception, print statistics and re-raise it, and/or output statistics in a configurable interval, e.g. every 30s. At the moment, one has to hit `Ctrl+C` to print statistics.


vogler:
Is there a reason to have colors off by default?

michael-schwarz:
> Is there a reason to have colors off by default?

At least none that I know of.

sim642:
> At the moment, one has to hit `Ctrl+C` to print statistics.

I've wondered about why statistics printing is Ctrl+C by default. It makes it quite annoying to kill Goblint because it breaks the convention of Unix programs. Also the regression testing script assumed that to stop Goblint until I fixed that in https://github.com/goblint/analyzer/commit/5ff22558904855e1459065a05ddd861d5231e279.

The most appropriate default would maybe be SIGUSR1.

sim642:
> Is there a reason to have colors off by default?

Having colorful output wouldn't hurt, but it's can be annoying when trying to pipe long Goblint output (e.g. tracing) to a file.

Maybe we should have one of those automatic color modes that detects whether output is piped and then disables.

src/maingoblint.ml
```diff
@@ -459,7 +459,7 @@ let main =
         | Exit ->
           exit 1
         | Timeout ->
-          Printexc.print_backtrace BatInnerIO.stderr;
+          (* Printexc.print_backtrace BatInnerIO.stderr; *)
```
sim642, line 462: The reason I added this line when updating the timeout mechanism is because it used to print backtrace also before #172. I thought it can be useful for understanding timeouts because it at least tells which step (initializing, solving, verifying, witness, etc) the timeout happens, which _may_ be relevant.

But I also see how the entire backtrace can be annoying because of solver recursion making it uselessly deep.

vogler:
> Maybe we should have one of those automatic color modes that detects whether output is piped and then disables.

Yes. I thought about that too, but didn't know what to search for and if it can be done in OCaml or requires some shell script around it.

sim642:
I think [`Unix.isatty`](https://ocaml.org/api/Unix.html#VALisatty) can be used to check that.

vogler:
> I've wondered about why statistics printing is Ctrl+C by default. It makes it quite annoying to kill Goblint because it breaks the convention of Unix programs. Also the regression testing script assumed that to stop Goblint until I fixed that in [5ff2255](https://github.com/goblint/analyzer/commit/5ff22558904855e1459065a05ddd861d5231e279).
> 
> The most appropriate default would maybe be SIGUSR1.

The idea was that you should be able to print the statistics with some key combination, and be able to kill it with some key combination.
The options I found ([wiki]) were
- `Ctrl+C` (SIGINT),
- `Ctrl+\` (SIGQUIT),
- `Ctrl+Z` (SIGSTP, useful for suspend and then `fg`),
- `Ctrl+T` (SIGINFO, prints the load and running time).

[wiki]: https://en.wikipedia.org/wiki/Signal_(IPC)#Sending_signals

The description of `SIGINT` sounded more appropriate than `SIGQUIT`:
"The SIGINT signal is sent to a process by its controlling terminal when a user wishes to interrupt the process."

`pkill` by default sends `SIGTERM`.

If we get some timer based interrupt, we could get rid of the `Ctrl+C` one.

vogler:
> I think [`Unix.isatty`](https://ocaml.org/api/Unix.html#VALisatty) can be used to check that.

Works. For reference: `$ cat tty.ml`
~~~ocaml
Printf.printf "stdin: %b, stdout: %b\n" (Unix.(isatty stdin)) (Unix.(isatty stdout))
~~~

~~~sh
$ ocaml unix.cma tty.ml
stdin: true, stdout: true

$ echo | ocaml unix.cma tty.ml
stdin: false, stdout: true

$ ocaml unix.cma tty.ml | cat
stdin: true, stdout: false
~~~

src/maingoblint.ml
vogler: Yes. I think by default this should be off since it's not an exception but expected (the timeout).
Maybe we get the output clean enough such that the step it was in is easily visible/searchable.
Maybe it's a good idea to have a signal for `Printexc.print_backtrace` to see what it's currently doing. Seems like it can be called without a raised exception.

vogler: I added `print_backtrace` on SIGUSR1 in https://github.com/goblint/analyzer/pull/207/commits/bb747e3115d6cfbe847db24fdc3210812c3226d2.
However, it only shows a trace until the re-raise here: https://github.com/goblint/analyzer/blob/bb747e3115d6cfbe847db24fdc3210812c3226d2/src/maingoblint.ml#L312-L313

vogler: Ok, with `Printexc.get_callstack 999` it shows more. https://github.com/goblint/analyzer/pull/207/commits/cd6184349049ac274eb1a5deae5133f724f32796

src/util/defaults.ml
```diff
@@ -78,7 +78,7 @@ let _ = ()
       ; reg Std "solverdiffs"     "false"        "Print out solver differences."
       ; reg Std "allfuns"         "false"        "Analyzes all the functions (not just beginning from main). This requires exp.earlyglobs!"
       ; reg Std "nonstatic"       "false"        "Analyzes all non-static functions."
-      ; reg Std "colors"          "false"        "Colored output."
+      ; reg Std "colors"          "'auto'"       "Colored output (via ANSI escape codes). 'auto': enabled if stdout is a terminal (instead of a pipe); 'always', 'never'."
```
michael-schwarz, line 81: In the scripts, we should probably just remove the `--enable colors` argument then.

https://github.com/goblint/analyzer/blob/75a14da737510a406ecc10ef93a2462a6ca97e98/regtest.sh#L12

https://github.com/goblint/analyzer/blob/75a14da737510a406ecc10ef93a2462a6ca97e98/scripts/dgoblint#L11-L12

https://github.com/goblint/analyzer/blob/75a14da737510a406ecc10ef93a2462a6ca97e98/scripts/spec/check.sh#L18

(I think the last one also needs more updates since `--sets result html` is gone as of #108 )

michael-schwarz:
Apart from the issue with the scripts, LGTM!

src/util/defaults.ml
vogler: We can probably remove `scripts/dgoblint` or does someone use it?

sim642 (approved):

src/util/defaults.ml
vogler: Removed 06bcba2

vogler:
Summary:
-  automatic `colors`: 'auto': enabled if stdout is a terminal (instead of a pipe); 'always', 'never'.
- `do_stats` on timeout, also warn about reached timeout before exit
- restore default behavior of Ctrl-C, `solver-signal` on `sigusr1`, `backtrace-signal` on `sigusr2`
- `dbg.timeout`: use profiling timer (user + system) instead of real time (wall clock)
  -> timeout can also be manually triggered with `sigprof`

sim642:
I wouldn't hurt because the use case for Ctrl-C is to kill a long-running and possibly non-terminating run.

PR #209, opened 2021-04-28
Refactor Set and Hoare domains

sim642:
### Changes
1. Move all Hoare domains into separate file for better organization.
2. Reduce duplication by defining `HoareDomain.Set_LiftTop` via `HoareDomain.Set` (without top) – no copy-paste Hoare order logic.
3. Reduce duplication by defining `HoareDomain.Set_LiftTop` via `SetDomain.LiftTop` (extracted from `SetDomain.ToppedSet`) – no copy-paste lifting logic.
4. Use polymorphic variants `Top` and `Lifted` for `SetDomain.LiftTop` instead of non-polymorphic `All` and `Set`. Although the naming is a bit less set-specific, it's now consistent with most of the other lifting, including `MapDomain` liftings.
5. Reduce duplication by defining `SetDomain.LiftTop` via `Lattice.LiftTop` – no copy-paste lifting logic, possible thanks to now-matching variants.
6. Enforce signatures for set domains.


sim642:
I realized one thing about our set domains. By default `SetDomain.Make` defines:
* `leq` as `subset`,
* `join` as `union`,
* `meet` as `inter`.
For basic sets, this is right. But then `SetDomain.LiftTop`:
* lifts `subset` and defines lifted `leq` as lifted `subset`,
* lifts `union` and defines lifted `join` as lifted `union`,
* lifts `inter` and defines lifted `meet` as lifted `inter`.
Again, if you think about lifting basic sets, this is right.

The problem is the following: `LiftTop` doesn't lift `leq` and `subset` independently, but implicitly assumes they are the same operation. Same for the other pairs. So it seems reasonable that `SetDomain.S` should explicitly state that they are the same operation. But that isn't always the case! `SetDomain.Reverse` reverses `leq`, `join` and `meet` but keeps `subset`, `union` and `inter` the same. It makes sense: you might want to literally `union` two reversed (i.e. must) sets without thinking that you actually need to use `meet` for it. But then the lifting should lift them separately to be completely correct.

Meanwhile, `HoareDomain.Set` defines them the other way around:
* `subset` as `leq`,
* `union` as `join`,
* `inter` as `meet`.
So there's no consistency about it.

Moreover, while `Reverse` preserves the underlying set functions and only redefines the lattice, `HoareDomain.Set` doesn't preserve the underlying set functions. This begs the question, whether it should also keep set operations literal but just redefine the lattice. The confusing thing just is, what possible meaning or use would that have.

I'm starting to think that the literal `subset` relation of Hoare ordering isn't literally very meaningful (and hence should maybe not even be exposed by its signature?). It only obtains a meaning when you consider the Hoare set as a maximal-elements representation of the set of downward-closed subsets. Then the `subset` relation is meaningful in the downward-closed subsets sense directly and happens to coincide with `leq` on the maximal-elements representation.

vogler:
> I'm starting to think that the literal `subset` relation of Hoare ordering isn't literally very meaningful (and hence should maybe not even be exposed by its signature?).

Makes sense. I guess we don't use it anyway or could replace those calls.
Maybe we should have a set signature that hides those operations that depend on the order.

sim642:
> Maybe we should have a set signature that hides those operations that depend on the order.

Splitting signatures like this makes functors, e.g. `SetDomain.LiftTop`, more annoying. Because then one needs a `LiftTop` that just lifts those functions (e.g. to implement `HoareDomain.Set_LiftTop` where the lifted set doesn't have `subset`), but in general you might want to also lift `subset`. So there'd need to be another `LiftTop` that handles that one additionally.

Functors aren't nicely composable like that, so in other places we just have base implementations that raise exceptions for the missing operations. Not as nice static typing wise, but more convenient for functors.

vogler:
Can't you have one functor to lift all and then hide some lifted functions you don't want to expose with some composite signature?

sim642:
It's a question of the functor's argument signature, which is mandatory. For example, `HoareDomain.Set` might not have `subset`, but then you can't use `SetDomain.LiftTop` to lift it.

vogler:
Ok, if the argument is also a problem you could nest them:

~~~ocaml
module type S1 = sig val a: unit end
module type S2 = sig include S1 val b: unit end
module F1 (M1: S1) = struct let a = M1.a end
module F2 (M2: S2) : S2 = struct include F1 (M2) let b = M2.b end
~~~

sim642:
Sure, but it seemed overkill in this case, I just made the `HoareDomain.Set` ones raise an exception for the few functions where it might matter. Didn't seem worth the hassle, especially given that we do the same for domains without `top`, which in an ideal world would also deserve a subsignature.

PR #211, opened 2021-04-30
Remove `sideg` list from `sync` return value

sim642:
Previously `sync` transfer functions returned both the updated state (like every other transfer function), but also an assoc list of `sideg`s to perform. Since `sync` may just use `sideg` directly to make the same side effects to global unknowns, the returned list provides no additional value. It's there for historical reasons.

Therefore I see no value keeping it around. Its removal simplifies a lot of `sync` code, especially in `BasePriv` and `Constraints`.


sim642:
One might have the performance concern that by doing `sideg` directly instead of collecting them causes some solver side effect functions to fire immediately, not at the end of the `sync` transfer. However, this doesn't actually happen because `MCP2` for whatever reason collects `sideg`s anyway and does them together afterwards, including ones from different analyses:
https://github.com/goblint/analyzer/blob/06eb5ae4a63642da0f321b03a74d5b63fde9bc52/src/analyses/mCP.ml#L837-L868

src/analyses/basePriv.ml
```diff
@@ -103,22 +103,23 @@ struct
         {st with cpa = CPA.add x v st.cpa}
     )
 
-  let sync ask getg (st: BaseComponents (D).t) reason =
+  let sync ask getg sideg (st: BaseComponents (D).t) reason =
     (* For each global variable, we create the diff *)
```
michael-schwarz, line 107: This comment seems outdated? Maybe also it should not be `add_var` anymore but `side_effect_var` or sth?

src/analyses/basePriv.ml
```diff
@@ -155,25 +156,26 @@ struct
         {st with cpa = CPA.add x v st.cpa}
     )
 
-  let sync ask getg (st: BaseComponents (D).t) reason =
+  let sync ask getg sideg (st: BaseComponents (D).t) reason =
     let privates = sync_privates reason ask in
     let module BaseComponents = BaseComponents (D) in
     if M.tracing then M.tracel "sync" "OldPriv: %a\n" BaseComponents.pretty st;
     (* For each global variable, we create the diff *)
```
michael-schwarz, line 163: Same here

PR #212, opened 2021-05-04
Clean up `Printable.S`

### LINKED ISSUE #210: Possibly excessive functions in `Printable.S`
The `Printable.S` signature has some functions whose necessity isn't clear to me. I want to know which of these do we really need and which we could get rid of. The signature is massive and thus super annoying to implement. So the fewer functions it contains, the better.

---

### `isSimple`
There are only two actual (non-delegating) calls of this function and both just use it to determine whether to avoid printing an extra `\n`:
https://github.com/goblint/analyzer/blob/08b43365752efd3e4fb3e8d20732e911c60e8330/src/domains/mapDomain.ml#L155-L156 https://github.com/goblint/analyzer/blob/08b43365752efd3e4fb3e8d20732e911c60e8330/src/util/hash.ml#L118-L119

### `pretty_f`
This seems to be only used for defining `pretty`. Might as well define `pretty` directly.

### `short`
This seems to just be used for debug printing and warning without `Pretty`. The massive downside of this function is its "width" argument:
* Many definitions outright ignore it and can produce longer string.
* Every call of `short` passes some completely random and different number as width, making the code full of weird magic constants.
* Often you find ridiculously high constants like 800 because there's actually no need to trim the large textual representation but instead we care about some information it may contain. It'd argue it's even harmful for debugging to see truncated strings because you can't tell if `<foo, ...>` and `<foo, ...>` are the same thing or not.
* Some calls even use width 0, which total nonsense and a sign that the width is meaningless.
* Many of these calls are also in commented out debugging code that tries to use `Printf` not `Pretty` or tracing.
* Mutex, symbolic locks and thread ID analyses use `short` to construct our fabulous string-based access partitions and lock sets, which are used to determine races: https://github.com/goblint/analyzer/blob/08b43365752efd3e4fb3e8d20732e911c60e8330/src/analyses/mutexAnalysis.ml#L92-L93 https://github.com/goblint/analyzer/blob/08b43365752efd3e4fb3e8d20732e911c60e8330/src/analyses/symbLocks.ml#L178 https://github.com/goblint/analyzer/blob/08b43365752efd3e4fb3e8d20732e911c60e8330/src/analyses/threadId.ml#L76-L77
  What if two different locks have insanely long names but share the 80-character prefix?

### `to_yojson`
This is just needed for "json" value for "result". What actually consumes these JSON results? If I understand correctly from git history, this was added to save results into MongoDB: https://github.com/goblint/analyzer/commit/8a4c722b97d710e251785951c6cc91bd186b47de. The MongoDB related code now is just around commented out.

Not that this is completely separate from our JSON configuration, which we have our own from-scratch JSON implementation for anyway.

sim642:
Closes #210.

### DIFF
--- a/src/analyses/arinc.ml
+++ b/src/analyses/arinc.ml
@@ -118,7 +118,7 @@ struct
     let node = Option.get !MyCFG.current_node in
     (* determine if we are at the root of a process or in some called function *)
     let fundec = MyCFG.getFun node in
-    let curpid = match Pid.to_int d.pid with Some i -> i | None -> failwith @@ "get_env: Pid.to_int = None inside function "^fundec.svar.vname^". State: " ^ D.short 100 d in
+    let curpid = match Pid.to_int d.pid with Some i -> i | None -> failwith @@ "get_env: Pid.to_int = None inside function "^fundec.svar.vname^". State: " ^ D.short d in
     let pname = match get_by_pid curpid with Some s -> s | None -> failwith @@ "get_env: no processname for pid in Hashtbl!" in
     let procid = Process, pname in
     let pfuns = funs_for_process (Process,pname) in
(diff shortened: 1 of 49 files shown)


vogler:
Too late now, but I guess here I tried to do the same: https://github.com/goblint/analyzer/compare/printable_name

Issue #208, opened 2021-04-27
High-level documentation


sim642:
I've been thinking about this for a while that Goblint has no high-level documentation that describes its overall structure and functioning. Some parts of the source code are commented (and few have OCaml documentation comments) but neither is useful for users and new developers, because you already have to know where to find something or which terms to grep for.

### Possible platforms
1. **GitHub Wiki**.
    * Pro: Accessible right here on GitHub with this repository.
    * Pro: Uses Markdown (among other possibilities).
    * Con: No good organization and structuring.
    * Con: Separate from git repository (has its own git repository under the hood though, but the GitHub website hardly exposes the git functionality of it).
2. **OCaml documentation** (odoc).
    * Pro: Together with source code.
    * Con: Uses obscure OCaml documentation markup language. Probably difficult to integrate any figures etc.
    * Con: Meant to just document _public_ API of a _library_ (only works now because `goblint-lib` is separated as a Dune library, Dune executables simply cannot have documentation).
    * Con: Organization isn't super flexible? Our unqualified module access means it's just a long list of modules.
    * Con: Must be hosted, although GitHub Pages would work.
3. [**GitBook**](https://www.gitbook.com/).
    * Pro: Uses Markdown.
    * Pro: Looks modern and easy to use.
4. [**Readthedocs**](https://readthedocs.org/).
    * Pro: Uses Markdown (or reStructuredText).
5. ...?

sim642:
I see @michael-schwarz has just started documenting _something_ for the SAS21 artifact (https://github.com/goblint/analyzer/tree/sas21_artifact_description), but we should probably have a broader strategy for this.

ivanazuzic:
I strongly support the idea of documenting Goblint. :) I definitely want to help with this.
We should pick the strategy that looks modern, is easy to look through (well structured) and at the same time not causing use too much overhead to write and maintain. GitBook seems to me like a good solution.

michael-schwarz:
> I see @michael-schwarz has just started documenting something for the SAS21 artifact

Yes, we need to provide with the artifact a description of how to reproduce our results (for `Validated` badge) and a description of the structure of the source code and how to extend it (for the `Extensible` badge).

---

Generally, the goal to provide comprehensive documentation of Goblint is very laudable, but I fear that goal is very ambitious and we will fall short of it and our documentation will become out-of-sync with the code very soon.

Maybe something more realistic to aim for is the following three things:

[...]

PR #216, opened 2021-05-06
Add documentation on how to debug Goblint with VS Code

### LINKED ISSUE #215: Add description on how to debug Goblint with VS Code
Add a description on how to use VS Code with `hackwaly.ocamlearlybird` to debug Goblint. This could be done in a Markdown file in a `docs` directory. Also add a template `launch.json`. This should make it easier for people to get started with debugging and developing Goblint. Related to #208 .

jerhard:
Add a Mardown description on how to debug Goblint with VS Code, using the `hackwaly.ocamlearlybird` extension. Closes #215. Related to #208.


docs/Debugging.md
```diff
@@ -0,0 +1,58 @@
+# Debugging Goblint
+
+## Building a Debuggable Goblint Executable
+
+To build a Goblint executable with debug information, run the following command within the `analyzer` directory.
+
+`make debug`
```
sim642, line 7: I haven't gotten around to trying earlybird but when I read about it, it required OCaml bytecode compilation. I see this debug target does that:
https://github.com/goblint/analyzer/blob/11dcafdfd30fac8ba089840a633461ee08ddf3d1/make.sh#L61-L63

Unlike the normal native build, this one still uses ocamlbuild instead of dune. Although this isn't necessarily about the documentation, I'm wondering if we could easily change this target to also build using dune.

docs/Debugging.md
```diff
@@ -0,0 +1,58 @@
...
+To debug OCaml programs, you can use the command line interface of `ocamldebug` or make use of the Visual Studio Code
+integration provided by `hackwaly.ocamlearlybird`. In the following, we describe the steps necessary to set up this VS Code extension to
+debug Goblint.
+
+### Setting-up Earlybird
+
+Install the [`hackwaly.ocamlearlybird` extension](https://marketplace.visualstudio.com/items?itemName=hackwaly.ocamlearlybird) in your installation of Visual Studio Code.
+To be able to use this extension, you additionally need to install `ocamlearlybird` on the opam switch you use for Goblint. To do this, run the following command in the `analyzer` directory:
+
+`opam install earlybird`
```
sim642, line 20: Maybe this should also be part of `make dev`?
https://github.com/goblint/analyzer/blob/11dcafdfd30fac8ba089840a633461ee08ddf3d1/make.sh#L98-L101

docs/Debugging.md
```diff
@@ -0,0 +1,58 @@
...
+        "arguments": [
+          "tests/regression/00-sanity/01-assert.c",
+          "--enable", "ana.int.interval",
+        ],
+        "stopOnEntry": false,
+      }
+    ]
+}
+```
+Note that the individual strings in `arguments` cannot contain spaces.
```
sim642, line 45: I guess they can, but then such argument would be passed as if it was quoted in the shell to contain a literal space.

sim642 (commented):
Since it's not mentioned anywhere, I'm guessing earlybird doesn't require a particular OCaml version anymore? Or if it does, then it should be mentioned here.

sim642 (approved):

sim642:
I just noticed something slightly annoying now. Having the bytecode build around makes regression testing script ask this every time: https://github.com/goblint/analyzer/blob/a71e9ed37d1367f069f9fcf95c35d37002ff0491/scripts/update_suite.rb#L43-L47

Maybe we should make it an option of the script instead?

michael-schwarz:
> Maybe we should make it an option of the script instead?

Maybe yes, but I would go with this behavior of asking as the default. If I have both around, it is not clear which one is up-to-date, and as a creature of habit one tends to run `make test`.

But maybe we could provide flags `-b` for bytecode and `-n` for native code?

jerhard:
>Since it's not mentioned anywhere, I'm guessing earlybird doesn't require a particular OCaml version anymore? Or if it does, then it should be mentioned here.

The latest version of earlybird does work with OCaml 4.11 and 4.12. As OCaml 4.11.1 is the default switch for the Goblint setup right now, it should not be necessary to specify the OCaml version here.

sim642:
I added some `./docs` on master now, which are also browsable via Read the Docs. You can merge in master and move your stuff at the end of `./docs/developer-guide/debugging.md` (while adjusting heading levels appropriately). Or I can take care of this myself after this gets merged.

PR #217, opened 2021-05-07
Remove Cygwin Support & Delete Old Scripts

michael-schwarz:
Cygwin support is probably outdated, and also no longer needed since the addition of WSL to Windows, which can indeed be used to run Goblint. This removes cygwin support.

I would also suggest deleting the following old scripts, which seem outdated and/or unmaintained. Any opinions?

- https://github.com/goblint/analyzer/blob/master/scripts/old/fixperms.sh
- https://github.com/goblint/analyzer/blob/master/scripts/old/make_dist.sh
- https://github.com/goblint/analyzer/blob/master/scripts/old/mit_Trier_vergleichen.rb
- https://github.com/goblint/analyzer/blob/master/scripts/old/monmake.sh
- https://github.com/goblint/analyzer/blob/master/scripts/old/runalltests.sh


michael-schwarz:
The failing tests are due to the new `readthedocs` CI action.

vogler:
Agree.
https://github.com/goblint/analyzer/blob/master/scripts/old/localization.sh also doesn't look important.

michael-schwarz:
> https://github.com/goblint/analyzer/blob/master/scripts/old/localization.sh also doesn't look important.

And then I guess https://github.com/goblint/analyzer/tree/master/tests/localization as well? The scripts is for running these tests.
Was this localization of widening ever implemented? And if yes what happened to the `cmpslr3vs2` solver?

vogler:
> Was this localization of widening ever implemented? And if yes what happened to the `cmpslr3vs2` solver?

I think it was for one of Kalmer's papers. Good question, it's not in master and I don't see any branch that might be it.
Probably worth to keep the examples around, but should add some reference to the paper, and move them into `regression` if we find the solver and it's actually testable.

michael-schwarz:
We discussed this at Gobcon but nobody quite recalls all the details, so I think merging them into the regression tests is the way to go.

Issue #28, opened 2015-12-22
Cleanup options


vogler:
There are 32 'experimental' options. Maybe those could be better classified or cleaned up.
 The number of experimental options should be kept at a minimum.
Todo:
- [x] Introduce additional categories:
   - [x] witness
   - [x] solver
   - [x] context
   - [x] incremental
   - [x] output
 - [x] Sort experimental options into the categories
 - [x] Rename `defaults.ml` to `options.ml`
 - [ ] Finalize options, so that they cannot be mutated after the solver started
 - [x] `ana.warnings` is used
 - [x] `warnstyle` only refers to race warnings

 Related #192

michael-schwarz:
@sim642: What is the state of `dbg.test.domain`? Right now it fails right away with

```
domain testing analysis...: PathSensitive2(MCP2 context hashconsed) lifted hashconsed
About to crash! (:-1)
Fatal error: exception Failure("HConsed lifted PathSensitive (Set ([expRelation:(Unit), mallocWrapper:(lifted proglines), base:(value domain * array partitioning deps * P), threadid:(Thread), threadflag:(MT mode), escape:(top or Set (variables)), mutex:(Reversed (top or Set (Normal Lvals * booleans)))])): no arbitrary")
```

Do we want this option to do domain testing when running on a single program? If yes we should aim to fix it?
Or is this superfluous since we introduced the `make domaintest` target?

sim642:
`dbg.test.domain` is separate from domaintest. The latter is testing the single small "leaf" domains that are explicitly listed for direct testing. The former is for testing `Spec.D` itself, which cannot be reasonably assembled in the separate domaintest executable without all the options and valid CIL values around (because some domains actually use them, but writing an `Arbitrary` instance for `varinfo` etc is problematic, if I remember correctly).

Since I never implemented `arbitrary` for every single domain in Goblint, `dbg.test.domain` has never tested the complete extent of the domains. Although when it initially was added, it at least didn't crash but just used trivial `arbitrary` implementations for some things.

So I'd keep it because it should allow testing the actual domain of the analysis, which depends on all the options, not just a handful we can test separately, for example all the domain functors. I'll have to see what the failing piece is right now.

michael-schwarz:
Closed in favor of #653.

PR #219, opened 2021-05-07
Introduce option `warn` with options `post | early | never`

### LINKED ISSUE #88: options for warnings/post-processing
There's some weirdness with the options and code for warnings:
- [ ] `Goblintutil.may_narrow` is only used to decide whether warnings/post-processing should be done and misused for it in the case of `dbg.earlywarn` (solver runs with `may_narrow = false`).
https://github.com/goblint/analyzer/blob/adeb01fe3c50cdb8c3d921091088e95d2c41d6c7/src/util/goblintutil.ml#L61-L63
https://github.com/goblint/analyzer/blob/adeb01fe3c50cdb8c3d921091088e95d2c41d6c7/src/framework/control.ml#L342-L344
- [ ] Warnings/post-processing are done during the verification phase. With `--enable noverify` there are no warnings. This should be decoupled.
- [ ] `--enable noverify` is weird (we have some `no-...` options for contexts, but those are actively changing something instead of not doing something).
- [ ] The code in [messages.ml](https://github.com/goblint/analyzer/blob/adeb01fe3c50cdb8c3d921091088e95d2c41d6c7/src/util/messages.ml) could use some cleanup/comments.

michael-schwarz:
The behavior is as follows:

- default: Warn in verifying phase, run verification regardless of `verify`, do not report problems if `verify` is `false`
- early: Same as above except that during solving warnings are already produced
- none: Do not produce warnings. Verification is run iff `verify` is true

Also the unused option `ana.warnings` was removed, and the config for incremental was fixed.

**`warn` set to `early` replaces `dbg.earlywarn`.**

Closes #88. References #28

### DIFF
--- a/src/util/defaults.ml
+++ b/src/util/defaults.ml
@@ -87,7 +87,7 @@ let _ = ()
       ; reg Std "save_run"        "''"           "Save the result of the solver, the current configuration and meta-data about the run to this directory (if set). The data can then be loaded (without solving again) to do post-processing like generating output in a different format or comparing results."
       ; reg Std "load_run"        "''"           "Load a saved run. See save_run."
       ; reg Std "compare_runs"    "[]"           "Load these saved runs and compare the results. Note that currently only two runs can be compared!"
-      ; reg Std "warn"            "'default'"    "Output warnings: 'default'. 'never': Do not produce warnings, 'early'. For debugging. Outputs warnings already while solving (may lead to spurious warnings/asserts that would disappear after narrowing)."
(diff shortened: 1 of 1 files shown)


michael-schwarz:
> post as in post-solving?

I renamed it thus.

> Maybe we need a new name for the verification phase, if it can be used either for verifying the solution or for reporting warnings.

I would want to stick with verify as that also makes sense in purely the fixpoint sense, especially since `verify` is likely not set to `false` too often, so it is very seldom that is used not only for verifying and only for warning.

sim642 (approved):

PR #220, opened 2021-05-09
Liveness analysis of multi-threaded C with POSIX threads

denis631:


src/util/defaults.ml
```diff
@@ -135,6 +135,9 @@ let _ = ()
       ; reg Analyses "ana.sv-comp.functions" "false" "Handle SV-COMP __VERIFIER* functions"
       ; reg Analyses "ana.specification"   "" "SV-COMP specification (path or string)"
       ; reg Analyses "ana.wp"              "false" "Weakest precondition feasibility analysis for SV-COMP violations"
+      ; (let open PthreadAnalysis.Flags in
+         let register f = reg Analyses (show f) (string_of_bool @@ default_value f) (description f) in
+         List.iter register all)
```
sim642, line 140: The two options should probably be defined here directly.

src/cdomains/pthreadDomain.ml
```diff
@@ -0,0 +1,96 @@
...
+
+
+  (** let hash = Hashtbl.hash *)
+  let hash x = Hashtbl.hash (Tid.hash x.tid, Pred.hash x.pred, Ctx.hash x.ctx)
+
+  let make tid pred ctx = { tid; pred; ctx }
+
+  let bot () = { tid = Tid.bot (); pred = Pred.bot (); ctx = Ctx.bot () }
+
+  let is_bot x = x = bot ()
```
sim642, line 73: Should probably be like `is_top` instead of hoping that structural equality does the right thing.

src/cdomains/pthreadDomain.ml
```diff
@@ -0,0 +1,96 @@
...
+  include Printable.Std
+
+  type t = domain [@@deriving to_yojson]
+
+  (** printing *)
+  let show x =
+    Printf.sprintf
+      "{ Tid=%s; pred=%s; ctx=%s }"
+      (Tid.show x.tid)
+      (Pretty.sprint 200 (Pred.pretty () x.pred))
```
sim642, line 38:
```suggestion
      (Pred.show x.pred)
```

src/cdomains/pthreadDomain.ml
```diff
@@ -0,0 +1,96 @@
...
+
+  let string_of_elt = Basetype.ProgLocation.show
+end
+
+type domain =
+  { tid : Tid.t
+  ; pred : Pred.t
+  ; ctx : Ctx.t
+  }
+[@@deriving to_yojson]
```
sim642, line 26: Why defined outside of `D`? Right now `to_yojson` is being derived twice for this type.

src/analyses/pthreadAnalysis.ml
```diff
@@ -0,0 +1,1359 @@
...
+    ; ("pthread_cond_wait", CondVarWait)
+    ]
+
+
+  let error_code = -1
+
+  let supported = List.map fst funs
+
+  let from_string s = Option.map snd @@ List.find (( = ) s % fst) funs
+end
```
sim642, line 89: This is somewhat duplicating `LibraryFunctions`.

src/analyses/pthreadAnalysis.ml
```diff
@@ -0,0 +1,1359 @@
...
+
+module Spec : Analyses.MCPSpec = struct
+  module M = Messages
+  module List = BatList
+
+  (* Spec implementation *)
+  include Analyses.DefaultSpec
+
+  (** Domains *)
+  include PthreadDomain
```
sim642, line 925: This is slightly unusual in our codebase. We're usually more explicit:
```suggestion
  module D = PthreadDomain.D
```
This also avoids including unnecessary stuff into this `Spec`, which could be just `open`ed for convenient access instead.

src/cdomains/pthreadDomain.ml
```diff
@@ -0,0 +1,96 @@
+open Prelude
+
+(** Thread ID *)
+module Tid = IntDomain.Flattened
+
+(** Context hash for function calls *)
+module Ctx = IntDomain.Flattened
```
sim642, line 7: Identifying contexts by hash isn't really sound due to hash collisions. This is why witness generation uses `tag` from hashconsing instead.

But I guess it's not worth fixing here because other extract analyses seem to do the same.

vogler: Is it? I thought it's just potentially less precise since contexts would not be kept apart but thrown together.
But yes, tags or this `ArincUtil.SymTbl` would be better.

src/analyses/pthreadAnalysis.ml
denis631: This is just a wrapper for easier pattern matching in the `special` analysis method. 
Do you propose extending `LibraryFunctions`? Since `LibraryFunctions` sees condition variable calls as lock/unlock, however, we wanna treat it differently here due to Promela Action extraction

src/cdomains/pthreadDomain.ml
sim642: I guess it might be fine, I'm not familiar with how the extraction things works.

I was just thinking of a case where a hash is used with `Hashtbl.replace` or similar. Then it replaces (which may cause unsoundness by overwriting something) instead of joining values together.

src/analyses/pthreadAnalysis.ml
sim642: The classification part of `LibraryFunctions` doesn't classify `pthread_cond_*` as `Lock`/`Unlock` right now. It classifies them just as `Unknown`, so it seems sensible to just extend that classification in general:
https://github.com/goblint/analyzer/blob/cf11c4b0b3602f417a5573e56853b81940384998/src/analyses/libraryFunctions.ml#L8-L81

The only mentions of `pthread_cond_*` in `LibraryFunctions` just state that calling them doesn't invalidate any of their arguments, but that has nothing to do with locking/unlocking:
https://github.com/goblint/analyzer/blob/cf11c4b0b3602f417a5573e56853b81940384998/src/analyses/libraryFunctions.ml#L245-L248

src/cdomains/pthreadDomain.ml
sim642: I didn't check how these values get used, so you're probably right that it's sound but in a rare case might just be less precise.

Seeing these extraction mechanisms now, at some point it might be worth generalizing and combining these extractions and witness generation, which both want to do this sort of thing.

src/cdomains/pthreadDomain.ml
sim642: This is still to be done.

src/analyses/extractPthread.ml
```diff
@@ -0,0 +1,1281 @@
...
+
+  let is_top tid var =
+    if Variable.is_global var
+    then
+      !table
+      |> Hashtbl.values
+      |> List.of_enum
+      |> List.exists (Set.exists (( = ) (Top var)))
+    else
+      Set.exists (( = ) (Top var)) @@ Hashtbl.find_default !table tid Set.empty
```
sim642, line 425: This is polymorphic equality on `varinfo`s inside `Variable.t` constructors.

tests/extraction/test.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+mkdir -p run
+cd run
+
+../../../goblint --set ana.activated[+] extract-pthread ../00-sanity.c && mkdir -p ../../../pml-result &&  mv pml-result/pthread.pml ../../../pml-result/pthread.pml && spin -a ../../../pml-result/pthread.pml && cc -o pan pan.c && ./pan -a &> out.txt
```
sim642, line 5: This is incredibly messy. Why do the test outputs have to be copied outside of the test tree instead of just copying the base model here once?

Also, this is precisely what #787 would be for.

michael-schwarz: Yes, I agree it is a bit ugly, maybe I try copying the model into the appropriate folder.

I mean a cram test would depend on all the output of `spin` and it prints a whole ton of things, whereas this tries to identify just the verdict.

sim642 (approved):

PR #224, opened 2021-05-11
Cleanups/unused options

michael-schwarz:
- Rename `ana.restart_count` to `exp.solver.slr4.restart_count` as it only applies for SLR4
- Remove options for circular intervals that were removed with #188 

References #28


jerhard (approved):
Looks good!

PR #225, opened 2021-05-11
Cast values for threadreturns to void*

jerhard:
This PR adds a cast for values of threadreturns to void-pointer-type. 

Background:

When multiple possible values for threadreturns are joined that are integers of different `ikind`, the result will be `top`.
Casting the values, before they are stored as threadreturns, avoids this issue.

Side note:
In particular, joining together the return values of different functions in the threadreturn happens when the `threadreturn` analysis is deactivated (which it is by default). @sim642 Maybe we should consider activating this analysis by default, as it does not seem to be too expensive?


jerhard:
>I suppose the same cast might be necessary here too?

As long as Cil knows that the expression passed as an argument to `pthread_exit` has type `void*`, eval_rv should already perform the cast if necessary. So this works in my testing even without an additional the cast there.

jerhard:
>Also, I'm all for making threadreturn activated by default. 

I now added it to the defaults analyses.

src/analyses/base.ml
```diff
@@ -2013,7 +2013,7 @@ struct
         | [exp] ->
           begin match ThreadId.get_current ctx.ask with
             | `Lifted tid ->
-              let rv = eval_rv ctx.ask ctx.global ctx.local exp in
+              let rv = eval_rv ctx.ask ctx.global ctx.local exp |> VD.cast ~torg:(typeOf exp) Cil.voidPtrType in
```
sim642, line 2016: I'm a bit confused now, because here you added the cast also to `pthread_exit`. Was this an accidental change?

jerhard: >Was this an accidental change?

Yes, thanks for noting! I will remove the cast on master.

Issue #31, opened 2016-01-11
use ppx to reduce boilerplate code


vogler:
Use type-driven code generation to get rid of boilerplate (see [ppx_deriving](https://github.com/whitequark/ppx_deriving)):

- `show`, `yojson` for output of domains values
- `eq`, `ord`, `map`, `fold`
- create ppx drivers for own ideas (e.g. names/lifting of modules)

sim642:
At some point I found [ppx_yojson](https://github.com/NathanReb/ppx_yojson), which attempts to simplify manual creation of Yojson values, e.g. by converting OCaml record-like syntax to a JSON object. It's not for deriving anything but just avoids having to construct Yojson values directly using its constructors.

This might be useful for manual implementations of `to_yojson`, which @keremc is adding more of. Although I think all those manual implementations are simple enough (e.g. a single string) that this isn't really worth it, but I'm still mentioning it here so I can close the tab...

sim642:
I happened upon [ppx_type_directed_value](https://github.com/janestreet/ppx_type_directed_value/), which Jane Street just has, but I've never heard of anyone using. Its idea being to make custom deriving plugins easier to define by not having to write a ppx deriver from scratch but just define the behavior for product and sum types and have it automagically work for all tuples, records, etc.

It sounds like a nice way to define our own lattice derivers, but it adds the Jane Street Base dependency and has notable performance implications:
> The usage of this PPX does incur a runtime cost proportional to the size of the type (e.g. number of fields/constructors/elements in a record/variant/tuple). In particular, if the type-directed value is a function type, there will be a runtime cost on _every_ invocation of the function.

In their benchmarking example they reimplement equality deriving, which via this ppx is ~4× slower than a direct equality deriver. So I'm a bit skeptical if it'd be fast enough to replace our currently hand-written lattice definitions.

PR #227, opened 2021-05-12
Reduce equal & compare boilerplate using deriving eq & ord

sim642:
Issue #31.

I went through the `equal` and `compare` definitions and replaced suitable ones with deriving `eq` and `ord`.

Some more boilerplate could be removed by moving some CIL-specific implementations that only compare IDs into `Deriving.Cil`. Right now their comparisons are duplicated all over the place.


sim642:
> Some more boilerplate could be removed by moving some CIL-specific implementations that only compare IDs into `Deriving.Cil`. Right now their comparisons are duplicated all over the place.

It kind of annoys me how such comparisons are spread around. For example `.vid\s+=` yields 39 different places where `varinfo`s are checked for equality. It would actually be better organized if all of those just delegated to `Basetype.Variables.equal`. Similarly with all the other modules we have for CIL types in `Basetype` (which might be better called `CilType` or something).

Then one doesn't need to define `Deriving.Cil.equal_varinfo` etc and make sure `Deriving.Cil` is opened everywhere. In fact, it could even make `Deriving.Cil` completely redundant because as seen in #210, we don't actually want the derived `to_yojson` functions anyway.

michael-schwarz:
https://github.com/goblint/analyzer/blob/274d41dc4befc4cb7f1dda5fe16b7fe1daa51790/src/cdomains/basetype.ml#L217-L239

https://github.com/goblint/analyzer/blob/274d41dc4befc4cb7f1dda5fe16b7fe1daa51790/src/cdomains/lval.ml#L525-L534


These and some other such places seem problematic in light of 
~~~
 If an ord comparator returns a value outside -1..1 range, the behavior is unspecified.
~~~
from https://github.com/ocaml-ppx/ppx_deriving#plugins-eq-and-ord 

One should probably patch all such places to return values from [-1,1].

src/solvers/topDown_deprecated.ml
```diff
@@ -18,20 +18,15 @@ module TD2 =
 
     module VI =
     struct
-      type t = Var.t * int
-      let compare (x,n) (y,m) =
-        match compare n m with
-        | 0 -> Var.compare x y
-        | n -> n
+      type t = Var.t * int [@@deriving ord]
```
michael-schwarz, line 21: Was this a deliberate choice to call compare on `x` and `y` only in case `n <> m`? One could imagine this would be more efficient?

sim642:
> One should probably patch all such places to return values from [-1,1]?

I noticed that at some point too, but chose to ignore it, because it worked regardless. And [`ppx_deriving_ord`](https://github.com/ocaml-ppx/ppx_deriving/blob/master/src_plugins/ord/ppx_deriving_ord.cppo.ml) actually nowhere makes any assumptions about the range, so I don't think it's an issue. But I'll go ahead and wrap those with `Stdlib.compare` anyway, so this stay open ended.

src/solvers/topDown_deprecated.ml
sim642: Good catch! Although I wouldn't bother changing it: that's inside a deprecated `TD2` solver, which at the end of the file isn't even registered. Even if it was some kind of optimization, I think it's irrelevant because TD3 doesn't use such thing anyway.

src/solvers/generic.ml
```diff
@@ -32,13 +32,8 @@ end
...
 module ExtendInt (B:Analyses.VarType) : Analyses.VarType with type t = B.t * int =
 struct
-  type t = B.t * int
+  type t = B.t * int [@@deriving eq, ord]
   let relift x = x
-  let compare ((u1,u2):t) (v1,v2) =
-    match Stdlib.compare u2 v2 with
-    | 0 -> B.compare u1 v1
-    | n -> n
-  let equal ((u1,u2):t) (v1,v2) = u2=v2 && B.equal u1 v1
```
sim642, line 41: Actually this place also used to compare second components first, so I'll have to look what the reason is.

sim642: This `ExtendInt` is actually used by some `MoreVars` SLR variants, so I'll revert it just in case.

src/solvers/topDown_deprecated.ml
sim642: I decided to revert this one along with the other place I found, just in case. And I added comments pointing this out to avoid anyone else tempting this change.

Issue #226, opened 2021-05-11
Fix comparison functions for Enums with multiple possible values


jerhard:
Currently, the Enums domain is not able to deliver precise information for comparisons, if at least one of the operands has multiple possible values.

See the following failing example. Enums infers the correct set of possible values for x at the assert, but fails at providing the precise value for the comparison itself.

```C
int main(){
    int top = rand();
    int x,y;

    if(top){
        x = 1;
    } else{
        x = 0;
    }

    assert(x<2); // x in {0,1}
    return 0;
}
```

jerhard:
Side Note: For DefExc, the range information could also be used for comparisons (but is currently not).

sim642:
Makes sense to have this because unlike arithmetic binary operators comparisons have only a few fixed possible results which cannot grow the value sets unboundedly.

PR #229, opened 2021-05-14
Enums: Make comparison functions more precise

jerhard:
This PR fixes the issue that the abstract comparison functions for `Enums` could only deliver exact results when both operands were singleton sets (see #226).

`lt`, `le` now look at the minium and maximum values of both abstract values to compute the result. `gt` is defined using `lt` with swapped arguments, analogous for `ge` and `le`.

Additionally, Enums now raises an `ArithmeticOnIntegerBot` exception, when exactly one of the operands of a binary arithmetic or logical operation is bottom. This behavior has already been implemented for `Interval` and `DefExc`.

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -412,6 +412,19 @@ module Size = struct (* size in bits as int, range as int64 *)
           a,b
       else
         a, b
+
+  (* From the number of bits used to represent a positive value, determines the maximal representable value *)
+  let max_from_bit_range pos_bits =
+    match pos_bits with
+    | Some i when i < 64L -> Some(BI.(pred @@ shift_left BI.one (to_int (BI.of_int64 i)))) (* things that are bigger than (2^63)-1 can not be represented as int64 *)
+    | _ -> None
+
+  (* From the number of bits used to represent a non-positive value, determines the minimal representable value *)
+  let min_from_bit_range neg_bits =
+    match neg_bits with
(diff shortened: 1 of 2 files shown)


jerhard:
The comparison functions `le`,`lt`,`gt`,`ge` could be extracted into a functor, and be defined in terms of `maximal`, `minimal`, `of_bool` and an `IntOps` module. I refrained from doing that for now as adding another functor will make things more convoluted.

src/cdomains/intDomain.ml
```diff
@@ -1750,8 +1755,57 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    else
+      match to_bool x with
+       | Some b -> of_bool ik (not b)
+       | None -> top_bool
+
+  let logand = lift2 I.logand
+  let logor  = lift2 I.logor
+  let maximal = function
+    | Inc xs when not (ISet.is_empty xs) -> Some (ISet.max_elt xs)
+    | Exc (_,r) -> Size.max_from_bit_range (R.maximal r)
```
michael-schwarz, line 1771: Is it sound to ignore the excluded elements here?

jerhard: The semantics of the `maximal` function should be to return an upper bound of the the concrete values, right? So this is sound, but possibly imprecise, because it might not be the smallest upper bound.

jerhard: I added some check in ac4d7d4817ce498ccb994a009b26811c0b3f5977 to make the values more precise in case the maximum given by the range is in the excluded set. 
I assume that this will happen pretty rarely and I am not sure whether we should do this?

michael-schwarz: One case where this might be interesting is if it allows excluding an overflow of sorts. Not sure if this realistically happens much

PR #230, opened 2021-05-17
Fix for serialization issue in HashCached

### LINKED ISSUE #228: Serialization and incremental runs fail with `--disable ana.opt.hashcons`
The use of `Lazy.t` in [`HashCached`](https://github.com/goblint/analyzer/blob/25f8eb4eb39438e3bee23618d8a9724d1a320069/src/domains/printable.ml#L174) causes objects of type `HashCached.t` to be unserializable unless the lazy value has been computed already. Presumably, `Lazy.t` is implemented as something like `type 'a t = Ready of 'a | Not_ready of (unit -> 'a)`, and the serialization fails when it encounters the second variant. By default, hash consing causes all hash values to be computed, but with `--disable ana.opt.hashcons`, hashes are left uncomputed, and serialization fails.

Possible solutions:
* @michael-schwarz suggested iterating over all entries in the result tables `lh` & `gh` and computing the hash values anyway when hash consing is disabled and `save_run` is specified
* Another way is to replace `Lazy.t` with a mutable field of `option` type

Note: Js_of_ocaml doesn't support [`Marshal.Closures`](https://ocaml.org/api/Marshal.html#VALto_channel), which is why I didn't list it as an option.

Example:
```
kerem@desktop:~/repos/analyzer$ ./goblint tests/regression/00-sanity/01-assert.c --sets save_run run
runtime: 00:00:00.037
vars: 11, evals: 13
[...]
Memory statistics: total=10.59MB, max=4.39MB, minor=8.88MB, major=3.54MB, promoted=1.83MB
    minor collections=4  major collections=1 compactions=0
kerem@desktop:~/repos/analyzer$ ./goblint tests/regression/00-sanity/01-assert.c --sets save_run run --disable ana.opt.hashcons
About to crash! (initfun:0)

runtime: 00:00:00.029
vars: 11, evals: 13
[...]
Fatal error: exception Invalid_argument("output_value: functional value")
```

keremc:
Fixes #228

### DIFF
--- a/src/domains/printable.ml
+++ b/src/domains/printable.ml
@@ -166,15 +166,17 @@ end
 
 module HashCached (M: S) =
 struct
+  module LazyHash = Goblintutil.LazyEval (struct type t = M.t type result = int let eval = M.hash end)
+
   let name () = "HashCached " ^ M.name ()
 
   type t =
     {
       m: M.t;
-      lazy_hash: int Lazy.t;
+      lazy_hash: LazyHash.t;
     }
 
-  let lift m = {m; lazy_hash = lazy (M.hash m)}
+  let lift m = {m; lazy_hash = LazyHash.make m}
   let unlift {m; _} = m
 
   let lift_f f x = f (unlift x)
@@ -184,7 +186,7 @@ struct
 
   let equal = lift_f2 M.equal
   let compare = lift_f2 M.compare
-  let hash x = Lazy.force x.lazy_hash
+  let hash x = LazyHash.force x.lazy_hash
   let show = lift_f M.show
 
   let pretty () = lift_f (M.pretty ())
--- a/src/util/goblintutil.ml
+++ b/src/util/goblintutil.ml
@@ -445,3 +445,25 @@ let signal_of_string = let open Sys in function
   | s -> failwith ("Unhandled signal " ^ s)
 
 let self_signal signal = Unix.kill (Unix.getpid ()) signal
+
+module LazyEval (M : sig
+  type t
+  type result
+  val eval : t -> result
+end) : sig
+  type t
+  val make : M.t -> t
+  val force : t -> M.result
+end = struct
+  type t = { user_data : M.t; mutable value : M.result option }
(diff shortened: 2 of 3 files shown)


src/domains/printable.ml
```diff
@@ -166,15 +166,17 @@ end
 
 module HashCached (M: S) =
 struct
+  module LazyHash = Goblintutil.LazyEval (struct type t = M.t type result = int let eval = M.hash end)
```
sim642, line 169: I like how drop-in this change is.

.github/workflows/tests.yml
```diff
@@ -59,3 +59,4 @@ jobs:
         run: |
           ./make.sh
           ruby scripts/update_suite.rb -m
+          gobopt='--disable ana.opt.hashcons' ruby scripts/update_suite.rb -m
```
michael-schwarz, line 62: One should probably put these into separate groups with appropriate titles, so one can see fast if it failed or nor

src/util/goblintutil.ml
```diff
@@ -445,3 +445,25 @@ let signal_of_string = let open Sys in function
...
+module LazyEval (M : sig
+  type t
+  type result
+  val eval : t -> result
+end) : sig
+  type t
+  val make : M.t -> t
+  val force : t -> M.result
+end = struct
+  type t = { user_data : M.t; mutable value : M.result option }
```
sim642, line 458: I doubt this makes any real difference, so probably not worth changing, but `user_data` isn't actually needed after the proper `value` has been computed. So it could just be
```ocaml
type t = { mutable value : (M.t, M.result) Either.t }
```
where `Left` holds the argument before evaluation and `Right` holds the result of the evaluation.

.github/workflows/tests.yml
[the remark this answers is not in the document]
keremc: OK. I added a test case for this bug in a new test group for marshaling related issues.

sim642 (approved):

PR #232, opened 2021-05-18
show widening points

vogler:
There's no way to show at which places the solver widened.
This would be good to have in order to see how solvers differ.
For now I added an option `dbg.print_wpoints` which just lists the widening points after solving.
`td3` only adds, but e.g. `slr3` also removes from `wpoint` during solving.
Of course one could add some tracing for it, but ideally this would somehow be indicated in the html output (e.g. by shape of node, color, *).

### DIFF
--- a/src/solvers/sLR.ml
+++ b/src/solvers/sLR.ml
@@ -169,6 +169,12 @@ module SLR3 =
       reachability vs;
       stop_event ();
 
+      if GobConfig.get_bool "dbg.print_wpoints" then (
+        Printf.printf "\nWidening points:\n";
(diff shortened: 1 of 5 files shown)


sim642 (approved):

vogler:
> Of course one could add some tracing for it, but ideally this would somehow be indicated in the html output (e.g. by shape of node, color, *).

Should we create an issue for this?
In general, it would be nice to have some way of seeing what the solver is doing.
In `td3` there's tracing output, but inspecting it is pretty cumbersome, and it's not consistent across solvers, so it's hard to compare.
An alternative to adding the same tracing to other solvers is having some more solver-specific tests like https://github.com/goblint/analyzer/tree/master/tests/regression/34-localization with comments how they differ (or add it in docs?).

PR #234, opened 2021-05-19
Extract `Printable` implementations for CIL types into separate `CilType` module

sim642:
This implements the thoughts I had in https://github.com/goblint/analyzer/pull/227#issuecomment-839698783 and furthers #31.

Instead of having equality, comparison and printing for CIL values duplicated all over the place whereever more sophisticated printables and lattices are defined, this moves those common implementations into `CilType`. For example, `CilType.Varinfo` has signature `Printable.S with type t = varinfo` and provides `equal`/`compare` based on just `vid`, which is how we compare them almost everywhere. For other types, e.g. `fieldinfo`, this provides proper implementations that don't forget to distinguish fields with the same name in different structs, etc. Whereever one really needs to override these sensible defaults, one can just define a new module to that use case specifically and include from `CilType`.

This also allows more ppx deriving to be used, which I tried to do as much as possible here. By defining new module types using `CilType.Varinfo.t` instead of `varinfo` directly, you can derive eq and ord, which will automatically use the sensible defaults for `varinfo` comparison from that module. This is cleaner than defining all these `equal_varinfo`, `compare_varinfo`, etc functions standalone in `Deriving.Cil` and having to remember to open that on top of `Cil`. The extra benefit is being able to use `CilType.Varinfo` directly as a `MapDomain` key module.

### DIFF
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -466,7 +466,7 @@ struct
         let over_all_x2 = op (op xl2 xm2) xr2 in
         let e1e_in_state_of_x2 = x2_eval_int e1e in
         let e2e_in_state_of_x1 = x1_eval_int e2e in
-        let e1e_is_better = (not (Cil.isConstant e1e) && Cil.isConstant e2e) || Basetype.CilExp.compareExp e1e e2e < 0 in
+        let e1e_is_better = (not (Cil.isConstant e1e) && Cil.isConstant e2e) || Basetype.CilExp.compare e1e e2e < 0 in (* TODO: why does this depend on exp comparison? *)
         if e1e_is_better then (* first try if the result can be partitioned by e1e *)
           if must_be_zero e1e_in_state_of_x2  then
             (e1, (xl1, op xm1 over_all_x2, op xr1 over_all_x2))
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -601,7 +601,7 @@ struct
   let update_length newl (x, l) = (x, newl)
 
   let printXml f (x,y) =
-    BatPrintf.fprintf f "<value>\n<map>\n<key>\n%s\n</key>\n%a<key>\n%s\n</key>\n%a</map>\n</value>\n" (Goblintutil.escape (Base.name ())) Base.printXml x "length" Idx.printXml y
+    BatPrintf.fprintf f "<value>\n<map>\n<key>\n%s\n</key>\n%a<key>\n%s\n</key>\n%a</map>\n</value>\n" (XmlUtil.escape (Base.name ())) Base.printXml x "length" Idx.printXml y
 end
 
 
@@ -645,7 +645,7 @@ struct
   let update_length newl (x, l) = (x, newl)
 
   let printXml f (x,y) =
-    BatPrintf.fprintf f "<value>\n<map>\n<key>\n%s\n</key>\n%a<key>\n%s\n</key>\n%a</map>\n</value>\n" (Goblintutil.escape (Base.name ())) Base.printXml x "length" Idx.printXml y
+    BatPrintf.fprintf f "<value>\n<map>\n<key>\n%s\n</key>\n%a<key>\n%s\n</key>\n%a</map>\n</value>\n" (XmlUtil.escape (Base.name ())) Base.printXml x "length" Idx.printXml y
 end
 
 module FlagConfiguredArrayDomain(Val: LatticeWithSmartOps) (Idx:IntDomain.Z):S with type value = Val.t and type idx = Idx.t =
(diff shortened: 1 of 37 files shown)


src/cdomains/arrayDomain.ml
```diff
@@ -449,7 +449,7 @@ struct
         let over_all_x2 = op (op xl2 xm2) xr2 in
         let e1e_in_state_of_x2 = x2_eval_int e1e in
         let e2e_in_state_of_x1 = x1_eval_int e2e in
-        let e1e_is_better = (not (Cil.isConstant e1e) && Cil.isConstant e2e) || Basetype.CilExp.compareExp e1e e2e < 0 in
+        let e1e_is_better = (not (Cil.isConstant e1e) && Cil.isConstant e2e) || Basetype.CilExp.compare e1e e2e < 0 in (* TODO: why does this depend on exp comparison? *)
```
michael-schwarz, line 452: This a very good question, I think the idea is to bias it towards simpler expressions.

src/cdomains/basetype.ml
```diff
@@ -46,7 +46,7 @@ struct
   let copy x = x
   let equal (x,a,_) (y,b,_) = ProgLines.equal x y && MyCFG.Node.equal a b (* ignores fundec component *)
   let compare (x,a,_) (y,b,_) = match ProgLines.compare x y with 0 -> MyCFG.node_compare a b | x -> x (* ignores fundec component *)
-  let hash (x,a,f) = ProgLines.hash x * f.svar.vid * MyCFG.Node.hash a
+  let hash (x,a,f) = ProgLines.hash x * f.svar.vid * MyCFG.Node.hash a (* FIXME: doesn't agree with equal, considers fundec component *)
```
michael-schwarz, line 49: I guess we should remove it here then?

michael-schwarz:
Apart from these two comments, looks good! I think it changes behavior in some cases, but in all of them, the new one seems to be the more reasonable one.
Should we maybe do a run of SV-Comp to see that we are still sound after these changes?

src/cdomains/arrayDomain.ml
sim642: Good point! I added a few comments about this.

sim642:
> Should we maybe do a run of SV-Comp to see that we are still sound after these changes?

Sounds reasonable. Am I correct that you've done some SV-COMP runs more recently, so maybe it's easier for you to do this?

michael-schwarz:
> Am I correct that you've done some SV-COMP runs more recently, so maybe it's easier for you to do this?

I started a run. Unfortunately our new sever is still not here and this testing framework by the student is not quite stable. It would be so nice if we had automation for this.

vogler:
> I started a run. Unfortunately our new sever is still not here and this testing framework by the student is not quite stable. It would be so nice if we had automation for this.

I didn't get the benefit of this testing framework over just putting a line in crontab or better having some server with a job queue checking every commit (https://github.com/goblint/analyzer/settings/hooks).

michael-schwarz:
> or better having some server with a job queue checking every commit (https://github.com/goblint/analyzer/settings/hooks).

Something along these lines was supposed to be the outcome. Basing it on this benchexec framework has the advantages that it is the same setup for SV-Comp so all those tests work out of the box and our own tests can be integrated without too many issues. Also this `tablegen` tool would in theory give us a nice diff of what changed between runs (or configurations).

One probably wants some glue code so that this is not all shell scripts but a bit more robust. But the idea was exactly this.

> checking every commit 

This is a bit optimistic given that one of these runs will likely take >12h (at least for SV-Com) even on the new hardware.

sim642:
> I started a run. Unfortunately our new sever is still not here and this testing framework by the student is not quite stable. It would be so nice if we had automation for this.

Any concerning results to report or can I go ahead with the merge?

michael-schwarz:
> Any concerning results to report or can I go ahead with the merge?

I'll take a look at the results this afternoon after teaching.

michael-schwarz:
The run aborted for some reason I don't really understand, so it's unfortunately not helpful at all.

michael-schwarz:
Comparing to an old run from January, not much changed: It succeeded on a few more benchmarks.

**However**, it also produces one incorrect result now for `pthread-atomic/qrcu-2.yml`.
This however, seems to be more likely to be linked to #183 than to this, so I am merging this now.

sim642:
> **However**, it also produces one incorrect result now for `pthread-atomic/qrcu-2.yml`.
> This however, seems to be more likely to be linked to #183 than to this, so I am merging this now.

Should be fixed by https://github.com/goblint/analyzer/commit/b6c8ae9e4b1e4100e095f259cd2eb267e0033a68 now by disabling some override behavior of the SV-COMP atomic lock in mutex analysis. I'm slightly confused about what the previous logic was there, so I left it commented out in case this makes it too imprecise again. There might be some middle ground where write locksets should be handled differently from readwrite locksets, but it's something to come back to when rethinking the atomics, which we don't extensively support anyway.

Issue #206, opened 2021-04-26
segfault for some benchmarks -> due to stack overflow


vogler:
The benchmark [single-thread/482.sphinx_livepretend_comb.c] terminates in 1s with the following configuration:
`./goblint -v single-thread/482.sphinx_livepretend_comb.c --enable exp.solver.td3.term --enable exp.earlyglobs --enable ana.int.interval --disable ana.int.enums --disable ana.int.def_exc --enable exp.widen-context --sets exp.privatization none`
but with only 701 vars solved.

When removing `--sets exp.privatization none`, i.e. with the default `'protection-read'` it crashes after ~18min:
~~~
...
Bug: typeOffset: Index on a non-array
Bug: typeOffset: Index on a non-array
timeout: the monitored command dumped core
Command terminated by signal 11
▸       Command being timed: "timeout --foreground 5h ./goblint -v single-thread/482.sphinx_livepretend_comb.c --enable exp.    solver.td3.term --enable exp.earlyglobs --enable ana.int.interval --disable ana.int.enums --disable ana.int.def_exc --          disable exp.full-context --disable exp.widen-context --disable exp.widen-context-partial --disable exp.solver.td3.              space_restore --sets save_run out/single-thread/482.sphinx_livepretend_comb.c.enable-term"
▸       User time (seconds): 1082.92
...
~~~ 

[single-thread/482.sphinx_livepretend_comb.c]: https://github.com/goblint/bench/blob/master/single-thread/482.sphinx_livepretend_comb.c

Signal 11 = SIGSEGV, segmentation violation

This was already the case before #183 was merged.
It also happens for `single-thread/duff-0.5_comb.c` in certain configurations:
| exp.privatization | exp.solver.td3.term | result |
| -------------|-------------|-----|
| none     | enable |  Command terminated by signal 11 |
| none     | disable |  Fatal error: exception Stack overflow |
| protection-read | enable |  Fatal error: out of memory |
| protection-read | disable |  Command terminated by signal 11 |

vogler:
Yes, something changed. At 6477d2758e147789a48b5397b6bc7465c3ef3d8b with `--sets exp.privatization none`:
`vars = 34477    evals = 298095`
`1m 45s`

jerhard:
As it seems to be resolved, I'll close this issue.

vogler:
The following still segfault (`ag 'terminated by signal 11' out-*`):
~~~
[...]

PR #235, opened 2021-05-19
Documentation: Add basic description on how to run Goblint and on need for context widening 

jerhard:
In https://github.com/goblint/analyzer/issues/206#issuecomment-842326410, @vogler  noted that stack overflows that occur in calls for C-functions result in Goblint being killed with `signal 11`. 

This PR adds a very basic documentation on how to run Goblint, with a remark to enable `exp.widen-context` when the analysis of recursive function leads to a stack overflow / `signal 11`.

Related to #208.

### DIFF
--- a/docs/user-guide/running.md
+++ b/docs/user-guide/running.md
@@ -25,21 +25,6 @@ For a list of all options and their possible configurations, run:
 ./goblint --print_all_options
 ```
 
-## Executing Regression Tests
-The regression tests contained in `./tests/regression` usually come with some specific options that should be passed be Goblint when executing them. The script `regtest.sh` can be used to execute Goblint on an individual regression test.
-It passes the options that are specified in a test file to Goblint.
-The options for a regression test are specified in a single line comment starting with `//PARAM:`.
-
-#### Example:
-The following command executes Goblint on a regression test. The test is identified using the digit-prefix of the folder and the source file.
-```
-./regtest.sh 01 02
-```
-To execute all regression tests, run:
-```
-make test
-```
-
 ## Analyzing Recursive Programs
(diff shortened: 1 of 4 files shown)


docs/user-guide/running.md
```diff
@@ -0,0 +1,47 @@
+# Running Goblint
```
sim642, line 1: This file also needs to be listed in `mkdocs.yml` for navigation.

michael-schwarz:
Should we also explain the Makefile support here? Or is this best saved for a different page?

docs/user-guide/running.md
```diff
@@ -0,0 +1,47 @@
...
+
+#### Example:
+The following command executes Goblint on a regression test. The test is identified using the digit-prefix of the folder and the source file.
+```
+./regtest.sh 01 02
+```
+To execute all regression tests, run:
+```
+make test
+```
```
sim642, line 41: I think this information shouldn't really be in the user guide and it's duplicating what's already written under "Testing". Some missing details could just be added there.

sim642:
> Should we also explain the Makefile support here? Or is this best saved for a different page?

I think this page would be fine for that until it grows too big. Although often these Read the Docs style sites have quite long pages, but with sensible heading structure, the sidebar makes even long pages navigable.

Same goes for these common options/configurations explanations, e.g. `exp.widen-context` in this case: in the future if we already more of them documented, then we might have to move them to a separate page.

jerhard:
>Should we also explain the Makefile support here? Or is this best saved for a different page?

We can put that into the same file, yes. It would need an a little bit more extensive explanation though, because it e.g. requires having a setup where the `cilly` executable is in the PATH – which it is not by default, IIRC. So I would move documenting this to another PR/Issue then.

PR #249, opened 2021-06-14
HoarePO.merge_join: Merge smaller map into larger one to improve performance

jerhard:
This was already an commented out performance tweak. In the library analysis uncommenting this code improves performance quite significantly, but this heavily relies on joining addresses. I assume this tweak should still have a positive impact on performance with the regular base analysis, although probably not as significant in most cases.

### DIFF
--- a/src/domains/hoareDomain.ml
+++ b/src/domains/hoareDomain.ml
@@ -68,7 +68,7 @@ struct
   (* join all elements from the smaller map into their bucket in the other one.
    * this doesn't need to go over all elements of both maps as the general merge above. *)
   let merge_join f x y =
-    (* let x, y = if Map.cardinal x < Map.cardinal y then x, y else y, x in *)
+    let x, y = if Map.cardinal x < Map.cardinal y then x, y else y, x in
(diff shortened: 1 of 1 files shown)


jerhard:
`cardinal` does not have constant time complexity as it is implemented in `BatMap`:

https://github.com/ocaml-batteries-team/batteries-included/blob/87c96084313d60333cece3daa6d9478ace5e0276/src/batMap.mlv#L503-L509

We are running the SV-Comp benchmarks to see how this impacts performance overall.

sim642:
There's particular place where this optimization is really useful, right? I think I noticed some base `reachable` changes on the library analysis branch.

If we don't want the extra cardinality comparisons generally, then it might be worth to just optimize that part as well. For example, by using a lifting functor which adds constant cardinality data with all operations (a la the lazy hash caching of maps). Or just using a `Hashtbl` there instead of joining together many big sets, like repeated string concatenation vs a mutable string builder.

jerhard:
We saw the tendency that this indeed improves performance on the SV-Comp benchmarks a little bit.

PR #251, opened 2021-06-15
Preparations for removing the Assert  query

ivanazuzic:
- Asked queries are tracked to prevent cycles, this seems to not fully work yet.
- `Queries.ml` now has `BigInt` in the integer domain.
- `IntDomain.Integers` are now a functor which takes `IntOps`. This way we can choose if we want them with `int64` or `BigInt`.

### DIFF
--- a/src/analyses/apron/poly.apron.ml
+++ b/src/analyses/apron/poly.apron.ml
@@ -110,7 +110,7 @@ struct
     | EvalInt e ->
       begin
         match D.get_int_val_for_cil_exp d e with
-        | Some i -> ID.of_int i
+        | Some i -> Queries.ID.of_int i
         | _ -> `Top
       end
     | MustBeEqual (e1, e2) ->
--- a/src/analyses/arinc.ml
+++ b/src/analyses/arinc.ml
@@ -398,15 +398,15 @@ struct
       | "LAP_Se_SetPartitionMode", [mode; r] -> begin
           match ctx.ask (Queries.EvalInt mode) with
           | `Lifted i ->
-            let pm = partition_mode_of_enum @@ Int64.to_int i in
-            if M.tracing then M.tracel "arinc" "setting partition mode to %Ld (%s)\n" i (show_partition_mode_opt pm);
-            if mode_is_multi (Pmo.of_int i) then (
+            let pm = partition_mode_of_enum @@ IntOps.BigIntOps.to_int i in
+            if M.tracing then M.tracel "arinc" "setting partition mode to %Ld (%s)\n" (IntOps.BigIntOps.to_int64 i) (show_partition_mode_opt pm);
+            if mode_is_multi (Pmo.of_int (IntOps.BigIntOps.to_int64 i)) then (
               let tasks = ctx.global tasks_var in
               ignore @@ printf "arinc: SetPartitionMode NORMAL: spawning %i processes!\n" (Tasks.cardinal tasks);
               Tasks.iter (fun (fs,f_d) -> Queries.LS.iter (fun f -> ctx.spawn None (fst f) []) fs) tasks;
             );
             add_action (SetPartitionMode pm)
-            |> D.pmo (const @@ Pmo.of_int i)
+            |> D.pmo (const @@ Pmo.of_int (IntOps.BigIntOps.to_int64 i))
           | _ -> D.top ()
         end
       | "LAP_Se_GetPartitionStatus", [status; r] -> todo () (* != mode *)
@@ -499,6 +499,9 @@ struct
         begin match name, entry_point, pri, per, cap with
           | `Lifted name, ls, `Lifted pri, `Lifted per, `Lifted cap when not (Queries.LS.is_top ls)
                                                                    && not (Queries.LS.mem (dummyFunDec.svar,`NoOffset) ls) ->
+            let pri = (IntOps.BigIntOps.to_int64 pri) in
+            let per = (IntOps.BigIntOps.to_int64 per) in
+            let cap = (IntOps.BigIntOps.to_int64 cap) in
             let funs_ls = Queries.LS.filter (fun (v,o) -> let lval = Var v, Lval.CilLval.to_ciloffs o in isFunctionType (typeOfLval lval)) ls in (* do we need this? what happens if we spawn a variable that's not a function? shouldn't this check be in spawn? *)
             if M.tracing then M.tracel "arinc" "starting a thread %a with priority '%Ld' \n" Queries.LS.pretty funs_ls pri;
             let funs = funs_ls |> Queries.LS.elements |> List.map fst |> List.unique in
@@ -527,7 +530,7 @@ struct
       | "LAP_Se_Suspend", [pid; r] ->
         add_actions @@ List.map (fun pid -> Suspend pid) (eval_id pid)
       | "LAP_Se_SuspendSelf", [timeout; r] ->
-        let t = eval_int timeout in
+        let t = IntOps.BigIntOps.to_int64 (eval_int timeout) in
         add_action (SuspendSelf (env.procid, t))
       | "LAP_Se_Resume", [pid; r] ->
         add_actions @@ List.map (fun pid -> Resume pid) (eval_id pid)
@@ -564,7 +567,7 @@ struct
       | "LAP_Se_DisplayBlackboard", [bbid; msg_addr; len; r] ->
         add_actions @@ List.map (fun id -> DisplayBlackboard id) (eval_id bbid)
       | "LAP_Se_ReadBlackboard", [bbid; timeout; msg_addr; len; r] ->
-        let t = eval_int timeout in
+        let t = IntOps.BigIntOps.to_int64 (eval_int timeout) in
         add_actions @@ List.map (fun id -> ReadBlackboard (id, t)) (eval_id bbid)
       | "LAP_Se_ClearBlackboard", [bbid; r] ->
         add_actions @@ List.map (fun id -> ClearBlackboard id) (eval_id bbid)
@@ -576,9 +579,9 @@ struct
         (* create resource for name *)
         let sid' = Semaphore, eval_str name in
         assign_id sid (get_id sid');
-        add_action (CreateSemaphore Action.({ sid = sid'; cur = eval_int cur; max = eval_int max; queuing = eval_int queuing }))
+        add_action (CreateSemaphore Action.({ sid = sid'; cur = (IntOps.BigIntOps.to_int64 (eval_int cur)); max = (IntOps.BigIntOps.to_int64 (eval_int max)); queuing = (IntOps.BigIntOps.to_int64 (eval_int queuing)) }))
       | "LAP_Se_WaitSemaphore", [sid; timeout; r] -> (* TODO timeout *)
-        let t = eval_int timeout in
+        let t = IntOps.BigIntOps.to_int64 (eval_int timeout) in
         add_actions @@ List.map (fun id -> WaitSemaphore (id, t)) (eval_id sid)
       | "LAP_Se_SignalSemaphore", [sid; r] ->
         add_actions @@ List.map (fun id -> SignalSemaphore id) (eval_id sid)
@@ -595,7 +598,7 @@ struct
       | "LAP_Se_ResetEvent", [eid; r] ->
         add_actions @@ List.map (fun id -> ResetEvent id) (eval_id eid)
       | "LAP_Se_WaitEvent", [eid; timeout; r] -> (* TODO timeout *)
-        let t = eval_int timeout in
+        let t = IntOps.BigIntOps.to_int64 (eval_int timeout) in
         add_actions @@ List.map (fun id -> WaitEvent (id, t)) (eval_id eid)
       | "LAP_Se_GetEventId", [name; eid; r] ->
         assign_id_by_name Event name eid; d
@@ -603,7 +606,7 @@ struct
       (* Time *)
       | "LAP_Se_GetTime", [time; r] -> todo ()
       | "LAP_Se_TimedWait", [delay; r] ->
-        add_action (TimedWait (eval_int delay))
+        add_action (TimedWait (IntOps.BigIntOps.to_int64 (eval_int delay)))
       | "LAP_Se_PeriodicWait", [r] ->
         add_action PeriodicWait
       (* Errors *)
@@ -632,8 +635,11 @@ struct
     let d = ctx.local in
     match q with
     | Queries.Priority _ ->
-      if Pri.is_int d.pri then
-        Queries.ID.of_int @@ Option.get @@ Pri.to_int d.pri
+      if Pri.is_int d.pri then Queries.ID.of_int @@ IntOps.BigIntOps.of_int64 @@ Option.get @@ Pri.to_int d.pri
+        (*match Option.get @@ Pri.to_int d.pri with
+          | `Lifted i -> Queries.ID.of_int @@ IntOps.BigIntOps.of_int64 i
+          | `Top -> Queries.Result.top q
+          | `Bot -> Queries.Result.bot q *)
       else if Pri.is_top d.pri then Queries.Result.top q else Queries.Result.bot q (* TODO: remove bot *)
     (* | Queries.MayBePublic _ -> *)
     (*   `Bool ((PrE.to_int d.pre = Some 0L || PrE.to_int d.pre = None) && (not (mode_is_init d.pmo))) *)
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -11,7 +11,7 @@ module Q = Queries
 module GU = Goblintutil
 module ID = ValueDomain.ID
 module IdxDom = ValueDomain.IndexDomain
-module IntSet = SetDomain.Make (IntDomain.Integers)
+module IntSet = SetDomain.Make (IntDomain.Integers (IntOps.Int64Ops))
 module AD = ValueDomain.AD
 module Addr = ValueDomain.Addr
 module Offs = ValueDomain.Offs
@@ -274,7 +274,7 @@ struct
 
   (* evaluate value using our "query functions" *)
   let eval_rv_pre (ask: Q.ask) exp pr =
-    let _other_analsyis_result = ask.f (Q.EvalInt exp) in
+    (*let _other_analsyis_result = ask.f (Q.EvalInt exp) in *) (* This is here in case we want to create acycle on purpose *)
     let binop op e1 e2 =
       let equality () =
         (* TODO: just return bool? *)
@@ -811,7 +811,6 @@ struct
   (* interpreter end *)
 
   let query ctx (type a) (q: a Q.t): a Q.result =
-    let to_int = BI.to_int64 in
     match q with
     | Q.EvalFunvar e ->
       begin
@@ -821,7 +820,7 @@ struct
       end
     | Q.EvalInt e -> begin
         match eval_rv (Analyses.ask_of_ctx ctx) ctx.global ctx.local e with
-        | `Int i when ID.is_int i -> Queries.ID.of_int (to_int (Option.get (ID.to_int i)))
+        | `Int i when ID.is_int i -> Queries.ID.of_int (Option.get (ID.to_int i)) 
         | `Bot   -> Queries.Result.bot q (* TODO: remove *)
         | v      -> M.warn ("Query function answered " ^ (VD.show v)); Queries.Result.top q
       end
@@ -836,7 +835,7 @@ struct
           let alen = List.filter_map (fun v -> lenOf v.vtype) (AD.to_var_may a) in
           let d = List.fold_left ID.join (ID.bot_of (Cilfacade.ptrdiff_ikind ())) (List.map (ID.of_int (Cilfacade.ptrdiff_ikind ()) %BI.of_int) (slen @ alen)) in
           (* ignore @@ printf "EvalLength %a = %a\n" d_exp e ID.pretty d; *)
-          (match ID.to_int d with Some i -> Queries.ID.of_int (to_int i) | None -> Queries.Result.top q)
+          (match ID.to_int d with Some i -> Queries.ID.of_int i | None -> Queries.Result.top q)
         | `Bot -> Queries.Result.bot q (* TODO: remove *)
         | _ -> Queries.Result.top q
       end
@@ -847,8 +846,8 @@ struct
         | `Address a ->
           let r = get ~full:true (Analyses.ask_of_ctx ctx) ctx.global ctx.local a  None in
           (* ignore @@ printf "BlobSize %a = %a\n" d_plainexp e VD.pretty r; *)
-          (match r with
-           | `Blob (_,s,_) -> (match ID.to_int s with Some i -> Queries.ID.of_int (to_int i) | None -> Queries.Result.top q)
+          (match r with 
+           | `Blob (_,s,_) -> (match ID.to_int s with Some i -> Queries.ID.of_int i | None -> Queries.Result.top q)
            | _ -> Queries.Result.top q)
         | _ -> Queries.Result.top q
       end
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -274,6 +274,7 @@ struct
 
   (* evaluate value using our "query functions" *)
   let eval_rv_pre (ask: Q.ask) exp pr =
+    let _other_analsyis_result = ask.f (Q.EvalInt exp) in
     let binop op e1 e2 =
       let equality () =
         (* TODO: just return bool? *)
--- a/src/analyses/flagModes.ml
+++ b/src/analyses/flagModes.ml
@@ -32,12 +32,12 @@ struct
               ctx.local
             end else begin
               match eval_int (Analyses.ask_of_ctx ctx) rval with
-              | Some ex -> D.add f (`Lifted (false,true,ex)) ctx.local
+              | Some ex -> D.add f (`Lifted (false,true, Option.get(IntDomain.FlatPureIntegers.to_int (IntOps.BigIntOps.to_int64 ex)))) ctx.local
               | _ -> D.remove f ctx.local
             end
           end else begin
             match eval_int (Analyses.ask_of_ctx ctx) rval with
-            | Some ex -> D.add f (`Lifted (false,true,ex)) ctx.local
+            | Some ex -> D.add f (`Lifted (false,true, IntOps.BigIntOps.to_int64 ex)) ctx.local
             | _ -> D.remove f ctx.local
           end
         end
@@ -60,6 +60,7 @@ struct
           let temp = eval_int (Analyses.ask_of_ctx ctx) ex in
           match temp with
           | Some value -> begin (*guard == value = (true true value*)
+              let value = IntOps.BigIntOps.to_int64 value in
               try
                 match (D.find f ctx.local) with
                 | `Lifted (false,_,old_val) -> if value <> old_val then D.add f `Bot ctx.local else ctx.local
@@ -100,6 +101,7 @@ struct
           let temp = eval_int (Analyses.ask_of_ctx ctx) ex in
           match temp with
           | Some value -> begin
+              let value = IntOps.BigIntOps.to_int64 value in
               try
                 match (D.find f ctx.local) with
                 | `Lifted (false,_,old_val) -> if value <> old_val then ctx.local else D.add f `Bot ctx.local
--- a/src/analyses/mCP.ml
+++ b/src/analyses/mCP.ml
@@ -527,6 +527,7 @@ struct
     (Queries.Result.top q) 
   else
     (* let () = Printf.printf "Size 1: %d\n" (QuerySet.cardinal asked) in *)
+    let asked = QuerySet.add (Any(q)) asked in
     let module Result = (val Queries.Result.lattice q) in
     let f a (n,(module S:MCPSpec),d) =
       let ctx' : (S.D.t, S.G.t, S.C.t) ctx =
@@ -536,7 +537,7 @@ struct
         ; control_context = ctx.control_context
         ; context = (fun () -> ctx.context () |> assoc n |> obj)
         ; edge   = ctx.edge
-        ; ask    = (fun (type b) (q: b Queries.t) -> query' (QuerySet.add (Any(q)) asked) ctx q)
+        ; ask    = (fun (type b) (q: b Queries.t) -> query' asked ctx q)
         ; emit   = (fun _ -> failwith "Cannot \"emit\" in query context.")
         ; presub = filter_presubs n ctx.local
         ; postsub= []
--- a/src/analyses/mCP.ml
+++ b/src/analyses/mCP.ml
@@ -4,6 +4,8 @@ open Prelude.Ana
 open GobConfig
 open Analyses
 
+module QuerySet = Set.Make(Queries.Query)
+
 type spec_modules = { spec : (module MCPSpec)
                     ; dom  : (module Lattice.S)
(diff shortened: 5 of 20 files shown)


src/analyses/apron/poly.apron.ml
```diff
@@ -110,7 +110,7 @@ struct
     | EvalInt e ->
       begin
         match D.get_int_val_for_cil_exp d e with
-        | Some i -> ID.of_int i
+        | Some i -> Queries.ID.of_int i
```
michael-schwarz, line 113: Not needed, we open queries.

src/analyses/arinc.ml
```diff
@@ -398,15 +398,15 @@ struct
       | "LAP_Se_SetPartitionMode", [mode; r] -> begin
           match ctx.ask (Queries.EvalInt mode) with
           | `Lifted i ->
-            let pm = partition_mode_of_enum @@ Int64.to_int i in
-            if M.tracing then M.tracel "arinc" "setting partition mode to %Ld (%s)\n" i (show_partition_mode_opt pm);
-            if mode_is_multi (Pmo.of_int i) then (
+            let pm = partition_mode_of_enum @@ IntOps.BigIntOps.to_int i in
```
michael-schwarz, line 401: Maybe define an alias for IntOps.BigIntOps?

src/cdomains/intDomain.ml
```diff
@@ -102,8 +102,9 @@ end
 
 module type IkindUnawareS =
 sig
-  include B with type int_t = int64
-  include Arith with type t:= t
+  type int_t
+  include B with type int_t := int_t
```
sim642, line 106: Is this separation of `int_t` still necessary? I know we added it at some point, but then also undid something?

src/analyses/arinc.ml
```diff
@@ -499,6 +499,9 @@ struct
         begin match name, entry_point, pri, per, cap with
           | `Lifted name, ls, `Lifted pri, `Lifted per, `Lifted cap when not (Queries.LS.is_top ls)
                                                                    && not (Queries.LS.mem (dummyFunDec.svar,`NoOffset) ls) ->
+            let pri = (IntOps.BigIntOps.to_int64 pri) in
```
michael-schwarz, line 502: The `(` `)` are superfluous here

src/analyses/arinc.ml
```diff
@@ -632,8 +635,11 @@ struct
     let d = ctx.local in
     match q with
     | Queries.Priority _ ->
-      if Pri.is_int d.pri then
-        Queries.ID.of_int @@ Option.get @@ Pri.to_int d.pri
+      if Pri.is_int d.pri then Queries.ID.of_int @@ IntOps.BigIntOps.of_int64 @@ Option.get @@ Pri.to_int d.pri
+        (*match Option.get @@ Pri.to_int d.pri with
```
michael-schwarz, line 639: why the commented out code?

src/domains/queries.ml
```diff
@@ -28,39 +28,38 @@ module Unit = Lattice.Unit
...
-  | MayBeLess: exp * exp -> MayBool.t t (* may exp1 < exp2 ? *)
+  | MustBeEqual: CilType.Exp.t * CilType.Exp.t -> MustBool.t t (* are two expression known to must-equal ? *)
+  | MayBeEqual: CilType.Exp.t * CilType.Exp.t -> MayBool.t t (* may two expressions be equal? *)
+  | MayBeLess: CilType.Exp.t * CilType.Exp.t -> MayBool.t t (* may exp1 < exp2 ? *)
   | HeapVar: VI.t t
-  | IsHeapVar: varinfo -> MayBool.t t (* TODO: is may or must? *)
-  | Assert: exp -> AR.t t
-
+  | IsHeapVar: CilType.Varinfo.t -> MayBool.t t (* TODO: is may or must? *)
+  | Assert: CilType.Exp.t -> AR.t t
```
sim642, line 62: Given that deriving ord didn't really work for a GADT, it might be more readable to revert these back.

src/analyses/flagModes.ml
```diff
@@ -32,12 +32,12 @@ struct
               ctx.local
             end else begin
               match eval_int (Analyses.ask_of_ctx ctx) rval with
-              | Some ex -> D.add f (`Lifted (false,true,ex)) ctx.local
+              | Some ex -> D.add f (`Lifted (false,true, Option.get(IntDomain.FlatPureIntegers.to_int (IntOps.BigIntOps.to_int64 ex)))) ctx.local
```
michael-schwarz, line 35: This seems complicated

src/domains/queries.ml
```diff
@@ -156,3 +155,127 @@ struct
     | PartAccess _ -> PartAccessResult.top ()
     | Assert _ -> AR.top ()
 end
+
+type any_query = Any: 'a t -> any_query
+
+module Query = struct
```
sim642, line 161: This module might be better named `Any` or something, because it's not for the actual GADT queries.

src/domains/queries.ml
```diff
@@ -156,3 +155,127 @@ struct
     | PartAccess _ -> PartAccessResult.top ()
     | Assert _ -> AR.top ()
 end
+
+type any_query = Any: 'a t -> any_query
```
sim642, line 159: This could be directly defined in the following module.

src/analyses/mCP.ml
```diff
@@ -519,8 +521,13 @@ struct
     let d = do_emits ctx !emits d in
     if q then raise Deadcode else d
 
-  (* Explicitly polymorphic type required here for recursive GADT call in ask. *)
-  and query: type a. (D.t, G.t, C.t) ctx -> a Queries.t -> a Queries.result = fun ctx q ->
+  and query': type a. QuerySet.t -> (D.t, G.t, C.t) ctx -> a Queries.t -> a Queries.result = fun asked ctx q ->
```
michael-schwarz, line 524: The comment still applies, right?

src/transform/expressionEvaluation.ml
```diff
@@ -129,7 +129,7 @@ module ExpEval : Transform.S =
         method private try_ask location expression =
           match ~? (fun () -> (ask location).Queries.f (Queries.EvalInt expression)) with
             (* Evaluable: Definite *)
-          | Some (`Lifted value) -> Some (Some (value <> Int64.zero))
+          | Some (`Lifted value) -> Some (Some (value <> IntOps.BigIntOps.zero))
```
sim642, line 132: This comparison should probably delegate to `BigIntOps.equal` or whatever instead.

I suspect there may be similar issues around elsewhere as well.

sim642 (changes requested):
This PR now contains both the generalization of `Queries.ID` from `int64` to big ints and the `query` cycle check, which isn't actually used.

There are three options:
1. Also do assert query removal and extend this PR to merge it all together.
2. Remove the `MCP2.query` change for now and merge the domain change on its own.
3. Keep it as is. I suppose the query cycle detection is also working but not used yet, so I think this would also be fine.

src/analyses/flagModes.ml
sim642: There's a very similar use below, where this is done more directly.

src/analyses/arinc.ml
ivanazuzic: I forgot to remove it. Done.

ivanazuzic:
> This PR now contains both the generalization of `Queries.ID` from `int64` to big ints and the `query` cycle check, which isn't actually used.
> 
> There are three options:
> 
> 1. Also do assert query removal and extend this PR to merge it all together.
> 2. Remove the `MCP2.query` change for now and merge the domain change on its own.
> 3. Keep it as is. I suppose the query cycle detection is also working but not used yet, so I think this would also be fine.

I vote for option 1.

src/cdomains/intDomain.ml
ivanazuzic: It's still necessary, without it I'm getting: "Multiple definition of the type name int_t. Names must be unique in a given structure or signature." for `include B`.

src/transform/expressionEvaluation.ml
ivanazuzic: I only found this one case when some value is compared to an IntOps.BigIntOps. 
If I find more similar cases later, I'll change them too.

src/domains/queries.ml
```diff
@@ -154,5 +153,130 @@ struct
...
+        | Any(CondVars _) -> 22
+        | Any(PartAccess _) -> 23
+        | Any(IterPrevVars _) -> 24
+        | Any(IterVars _) -> 25
+        | Any(MustBeEqual _) -> 26
+        | Any(MayBeEqual _) -> 27
+        | Any(MayBeLess _) -> 28
+        | Any(HeapVar) -> 29
+        | Any(IsHeapVar _) -> 30
+        (*| Any(Assert _) -> 31*)
```
michael-schwarz, line 198: We should probably also remove `Assert` here then

src/domains/queries.ml
```diff
@@ -154,5 +153,129 @@ struct
...
+          else
+            CilType.Exp.compare e2 e4
+        | Any(MayBeLess (e1, e2)), Any(MayBeEqual (e3, e4)) -> 
+        let r2 = CilType.Exp.compare e1 e3 in
+          if r2 <> 0 then
+            r2
+          else
+            CilType.Exp.compare e2 e4 
+        | Any(IsHeapVar v1), Any(IsHeapVar v2) -> CilType.Varinfo.compare v1 v2
+        (*| Any(Assert a1), Any(Assert a2) -> CilType.Exp.compare a1 a2 *)
```
michael-schwarz, line 279: Here too :)

ivanazuzic: Sorry, yes.

src/cdomains/apron/octApronDomain.apron.ml
```diff
@@ -593,4 +562,287 @@ struct
...
+  include CilType.Varinfo
+  let var_name g = g.vname
+end
+
+module GV =
+struct
+  include VarMetadataTbl (GVM)
+
+  let make g = make_var g
+end
```
sim642, line 848: I think you cherry-picked too much stuff here, this isn't necessary here, is it?

src/analyses/base.ml
```diff
@@ -274,44 +274,75 @@ struct
...
+      | Question (_, _, _, _)
+      | CastE (_, _)
+      | StartOf (_, _) ->
+        let ans =  ask.f (Q.EvalInt exp) in
+        (match ans with
+          | `Top ->  evaluate
+          | `Bot -> None
+          | `Lifted z -> 
+            let ik = Cilfacade.get_ikind (Cil.typeOf exp) in Some (`Int (ID.of_int ik z) )
+        )
```
sim642, line 343: Why is the `Const` case separate here? It looks like the same exact code.

src/analyses/apron/octApron.apron.ml
```diff
@@ -163,13 +163,6 @@ struct
...
-    | Assert e ->
-      begin match D.check_assert e ctx.local with
-        | `Top -> `Top
-        | `True -> `Lifted true
-        | `False -> `Lifted false
-        | _ -> `Bot
-      end
     | EvalInt e ->
       begin
         match D.get_int_val_for_cil_exp d e with
```
sim642, line 168: This `EvalInt` can only evaluate (linear) expressions like `x - y`, but not constraints like `x < y`. So if there's an `assert(x < y);` and base queries the expression, this will just return `Top`, no?

I think this needs to distinguish the two cases and either evaluate an expression or check a constraint (via `check_assert` or whatever). How do the octApron tests even pass right now?

sim642: I think I figured out why the tests even pass right now. It's because the few tests that we have are from the octagon analysis and have assertion in weird form like `assert(X - N == 0);`. The base analysis also queries the subexpression `X - N`, which octApron can evaluate as a linear expression, not a constraint, and returns the bounds `[0, 0]` for. And then the base analysis is doing the comparison between 0 and 0.

If it was `assert(X == N);` instead, then it wouldn't work, although it really should, because Apron can easily handle such constraint. It fails because `X == N` isn't handled (it's a constraint, not an expression) and `X` and `N` separately have no bounds, so base cannot compare any subexpressions either.

sim642:
I fixed my latest review comments myself now. The separation of constraints (comparisons) and expressions in `EvalInt` was straightforward but it revealed a subtle issue. Namely, Apron doesn't apply non-equality constraints very precisely (at all?): https://github.com/antoinemine/apron/issues/37.

As suggested, one should assert the less-than and greater-than constraints instead and join the results. I guess this is what @ivanazuzic attempted to do at one point in `assert_inv`, but that attempt to gain back this precision was unsound. The trick with the join is that if both sides are bottom, the entire thing is also bottom. But more subtly, if only one of the sides is bottom, it also refines the result to the other. I think this helps in cases where non-equality is used to exclude one endpoint of some interval exactly.

sim642 (approved):

sim642:
I've tuned this enough that I think I'm satisfied, but there is one potential aspect that might be considered.

Namely, `eval_rv_pre` is called by recursive `eval_rv` calls on every subexpression and the `EvalInt` query is made if it has integer type. So not only is the entire asserted expression passed to octApron, but smaller subexpressions are as well if the outer query didn't already give a definite result. In theory this could be useful since octApron still may know a value in a subexpression that base doesn't (e.g. in 36/50). And that applies to any evaluation in base, not just assertions.

I haven't made any benchmarking, but it sounds like this could be somewhat expensive. Even when octApron isn't activated, base would do the `EvalInt` query to itself, which would do a duplicate `eval_rv` that can only return a lifted integer via the query result, but not interval. Maybe we'll have to also implement the logic of `ctx.ask` not calling the same analysis's `query`.

michael-schwarz:
I can start a benchmarking run for this later today :smile: Then we can see what sort of impact it has.

michael-schwarz:
Sorry for the delay, I started the run now.

michael-schwarz:
When running without octagons and comparing to ga06a611c, we get 18 correct answers that turn into 1 FAILED, 2 TIMEOUT, and the rest unknown. On the other hand, we succeed on 4 additional tests. 
[rm_assert_query_no_oct.zip](https://github.com/goblint/analyzer/files/6753394/rm_assert_query_no_oct.zip)

I am currently also running with octApron so we can see what happens there.

michael-schwarz:
Also, we are additionally unsound on `pthread-divine/one_time_barrier_2t.yml` and `pthread-divine/one_time_barrier_3t.yml`

sim642:
Hmm yeah, there's a bunch of other concerning things there:
- [ ] nla-digbench-scaling benchmarks go from 0.1s (unknown) to TIMEOUT.
- [ ] recursive benchmarks go from TIMEOUT to SEGMENTATION FAULT.
- [ ] seq-mthreaded-reduced go from unknown to ERROR (verify).

sim642:
Ugh, there's something else going on here. The new commit in the diff results is `1c298caf`, but that doesn't even exist in the repo. I'm guessing you merged this with master locally?

Anyway, the first unsoundness case is sound on this branch, but actually fails on the current master 95d75cc8. Bisect seems to tell me that purely on master, it becomes unsound after https://github.com/goblint/analyzer/commit/cd33c35f1b4856869f21ebea00df26deaf904ef7.

michael-schwarz:
>  but that doesn't even exist in the repo. I'm guessing you merged this with master locally?

Yes, it is this PR merged into master

vogler:
> **EDIT:** @vogler Running
> 
> ```
> ./goblint --conf conf/svcomp21.json --sets ana.specification ../sv-benchmarks/c/properties/unreach-call.prp ../sv-benchmarks/c/pthread-divine/one_time_barrier_2t.i --sets exp.solver.td3.side_widen unstable_called
> ```
> 
> is fine on current master, but removing the `side_widen` and using the new default, it becomes unsound.

Ok, interesting. We just came up with that yesterday. It should be more precise since it creates fewer widening points for side-effected variables. No idea how it could cause unsoundness.
How do I see that/what is unsound? The only difference in output is `SV-COMP result: unknown`/`true`.

sim642:
> How do I see that/what is unsound? The only difference in output is `SV-COMP result: unknown`/`true`.

According to dead code output, in the unsound case there's one more dead branch, which probably is ruling out the key `reach_error()` call:
```
Dead code: the then branch over expression '! (b->seen < b->thread_count)' is dead! (../sv-benchmarks/c/pthread-divine/one_time_barrier_2t.i:715)
```

vogler:
Any chance the benchmark is wrong and it should be dead? 😄 
~~~
sound: --sets exp.solver.td3.side_widen unstable_called
|rho|=232
|stable|=249
|infl|=479
|wpoint|=10
~~~

~~~
unsound: --sets exp.solver.td3.side_widen sides
|rho|=223
|stable|=249
|infl|=470
|wpoint|=1
~~~

michael-schwarz:
It is always an option.

However, given that CMBC and Divine both produce accepted violation witnesses in SVCOMP'20 for this test that can be verified by other tools, I think it is more likely that we have a bug somewhere than both of them (and the verifiers) are buggy.

michael-schwarz:
Something that comes to mind is how we handle `pthread_cond_wait(&b->sig, &b->lock)` in the presence of privatization? Does it correctly release `&b->lock` and reacquire it?

Otherwise we might read only our stale privatized value.

vogler:
> It might however be the case, that the problem is not widening less but something unrelated, that just shows after this change.

That seems plausible. However the only difference should be having fewer widening points.

Sorry for committing to master. Seemed like a simple change and our reg. tests were ok.
Should I revert it on master and put it in a PR?

michael-schwarz:
> Should I revert it on master and put it in a PR?

I don't think is is necessary, I think I have identified a likely culprit: #267

sim642:
It might make sense to redo the comparison benchmark regardless of master, so from merge base `3690e97b968298cf7854fe12875647743bf2428b` to this branch (unmerged).

This way there's no wait for #267 and TD3 changes are also isolated. Just like the unsoundness reveal, other changes (without octApron) might've also been simply due to TD3 changes, not these.

michael-schwarz:
This is the diff between 3690e97 and 2c0369e (no octagons). There is one test where we newly succeed in this branch, however there are still 15 tests where we go from a correct answer to unknown in this branch (and 4 new timeouts for what used to be correct). Also, there is quite a number of timeouts where we had unknown before.


[rm_assert_query_new.zip](https://github.com/goblint/analyzer/files/6767608/rm_assert_query_new.zip)

sim642:
Nevermind the changes in true results, the timeouts are still significant. Especially those which go from 0.1s to 60s timeout, like https://github.com/sosy-lab/sv-benchmarks/blob/99d37c5b4072891803b9e5c154127c912477f705/c/nla-digbench-scaling/ps6-ll_valuebound5.c.

I think I know what the issue is here: some of the integer expressions are very large (and thus deep), so evaluating `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y + 12*x == 0` goes like:
* Ask in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y + 12*x == 0`
  * Answer in base
    * Ask in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y + 12*x == 0`
      * Query cycle, return top
    * Eval in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y + 12*x == 0`
      * Eval in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y`
        * Ask in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y`
          * Answer in base
            * Ask in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y`
              * Query cycle, return top
            * Eval in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y`
              * Ask in base  `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y`
              * ...
  * Eventually the query just returns top anyway
* Eval in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y + 12*x == 0`
    * Eval in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y`
      * Ask in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y`
        * Answer in base
          * Ask in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y`
            * Query cycle, return top
          * Eval in base `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y + y*y`
            * Ask in base  `-2*y*y*y*y*y*y - 6 * y*y*y*y*y - 5 * y*y*y*y`
            * ...

So the fact, that at every level base first asks and as fallback evaluates itself, doesn't just double the work but also doubles all the underlying work. So it's exponential at minimum although I suspect it might even be worse than that. Because you see, asking base evaluates the expression exponentially and then the fallback also does the whole exponential thing itself, just without the outermost wrapping query.

I think this all could be avoided, if base didn't need to fall back, but could just return intervals etc through `Queries.ID`.

As a quick workaround I tried the suggested thing of not querying itself (by adding an excluding argument to queries in `MCP`). It fixes this example and makes it run instantly again. But it also breaks 66 regression tests, so self-querying is pretty crucial after all.

michael-schwarz:
> I think this all could be avoided, if base didn't need to fall back, but could just return intervals etc through Queries.ID.

Yes, the downside then is that it introduces the dependency with `IntDomTuple` everywhere, so if we go that route, I guess we need to supply convenience functions to wrap values from a specific ID into such an `IntDomTuple`.

sim642:
I tried what @ivanazuzic initially suggested: adding an argument to `eval_rv` to forbid the `EvalInt` query and making it true when base itself is answering `EvalInt`. It fixes the timeout but the runtime is still 2s instead of 0.1s.

This is because when base evaluates each subexpression itself (not when responding to `EvalInt`), it will still query each one. And base will respond to each of those. Thanks to the excluding argument, it will not go for a cyclic query in `eval_rv_pre`, but it will still do the fallback to its own normal `eval_rv` logic just for the purposes of the query. And then that ends up returning top, so the outermost evaluation in base falls back and also does the same `eval_rv` logic on all the same subexpressions repeatedly.

So this extra `eval_rv` argument does avoid one of the exponential processes but not the other.

The annoyance is still that `EvalInt` may just return flat integers, so when base is answering that, it will go through all its normal trouble and return top if the value was not a definite integer. Only to then fall back at the outer level to do all the same work again and this time actually use the result directly because it doesn't go through `Queries.ID`. 

Note that some kind of query result cache wouldn't solve this problem because it's `eval_rv` in base that's done multiple times, not the `EvalInt` query. So really it would be the best if base's `EvalInt` could return the non-definite value (e.g. interval) through the query and the outer `eval_rv` doesn't have to fall back to anything because through the query it already gets the most precise value it would also compute by falling back. Hence avoiding the second exponential duplication component.

sim642:
> Yes, the downside then is that it introduces the dependency with `IntDomTuple` everywhere, so if we go that route, I guess we need to supply convenience functions to wrap values from a specific ID into such an `IntDomTuple`.

I don't think any additional convenience functions are required. Since `Queries.ID` is flat integers right now, the only values being embedded are definite integers, so just using `IntDomTuple.of_int i` instead of `Lifted i` (which is `Queries.ID.of_int` anyway) suffices.

The more problematic aspect is that `IntDomTuple` requires ikinds from every analysis responding to `EvalInt`. I don't know if it's just as easy as `get_ikind (typeOf e)` or not.

sim642:
I think I managed to finally switch `Queries.ID` to `IntDomTuple` and make base's `eval_rv` go through `EvalInt` for integer expressions without having to exponentially fall back to the normal case. It solves the issue of deep expressions.

The result of all this refactoring of `eval_rv` is that it no longer even requires the query cycle check. For best possible performance, when base answers `EvalInt`, it directly jumps to `eval_rv_no_ask_evalint` to avoid asking for the exact same query, that due to cycle would return a useless top anyway. Still, we can leave the query cycle mechanism in.

I think we can do a new comparison benchmark to see if this hopefully behaves the same as the old master.

michael-schwarz:
> I think we can do a new comparison benchmark to see if this hopefully behaves the same as the old master.
I'll starte one :smile:

michael-schwarz:
I guess this looks a lot more reasonable now :smile: 

[rm_assert_query_3.zip](https://github.com/goblint/analyzer/files/6786031/rm_assert_query_3.zip)

sim642:
Indeed, although I might try to look into the 15 benchmarks that go from true to unknown. Seems a bit surprising that it's less precise.

Could also be that we were unsound before. Maybe due to sub-asks in `patched_ask` still being based off the old local state or something that I might have fixed.

sim642:
I just tried one locally and it still outputted true, so I was confused about the unknown result in the table.
But apparently the output can sometimes get mixed up like this:
```
../sv-benchmarks/c/ldv-consumption/32_7a_cilled_linux-3.8-rc1-32_7a-drivers--net--arcnet--arcnet.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.iSV-COMP result: true
:6576: Bug: typeOffset: Field on a non-compound
```
The benchexec tool-info module for Goblint doesn't account for this possibility. I'm not sure it should either because there's no reason `SV-COMP result: true` couldn't be split up by arbitrary garbage output as well.

In this case it seems like we now get tons of these typeOffset errors while previously we didn't, so there might be something to still fix here.

sim642:
Actually, maybe not. They just appear because I do an extra `typeOf` call to determine whether `EvalInt` should be queried: https://github.com/goblint/analyzer/pull/251/files#diff-969b2ca0b7175f19daa5deae1308a02718e8cd1927a4d8e160541a35e57c31dbR565-R577.

sim642:
Honestly, I don't know how to fix this other than fixing the complete trash error handling strategy in CIL. It prints error text directly and raises an argumentless `Errormsg.Error`. So it's impossible _not_ to have the output because catching the exception doesn't undo printing. The sane thing would be to not print errors directly but return the text as an argument to the exception (you know, like `Failure` does). But I think that's well beyond the scope of this PR.

I guess I could call our copy-pasted `Cilfacade.typeOf` instead, which doesn't print garbage:
https://github.com/goblint/analyzer/blob/e9e71ae5259975c4bdfe2ff8644c595a06299d79/src/util/cilfacade.ml#L249-L317
But every use of `typeOf` in Goblint uses `Cil.typeOf` directly (except a single one in osek) and this problem would in no way be limited to the `typeOf` call I added here (base has 69 others, which all could just as easily cause this problem) as any other CIL functions may do the same crap error handling.

michael-schwarz:
> I guess I could call our copy-pasted `Cilfacade.typeOf` instead

That funnily enough slightly differs from `Cil.typeOf` e.g. in that it doesn't handle `Real` and `Imag`and handles `Question` different.

Also, the entire `typeOf` logic in CIL seems to be in need of a rewrite, e.g. it uses two slightly different mechnaism of reporting errors `E.s (bug "typeOfLval: Mem on a non-pointer (%a)" !pd_exp addr)` vs `E.s (E.bug "typeOffset: Index on a non-array")`
https://github.com/goblint/cil/blob/d61cfa6a403f4d278c86344091bc07dccd8b549b/src/cil.ml#L1963

________________

Maybe we can work some super dirty workaround that temporarily sets `Errormsg.logChannel ` to something different (assuming we patch the uses of `bug` to be uses of `E.bug`)?

sim642:
> That funnily enough slightly differs from `Cil.typeOf` e.g. in that it doesn't handle `Real` and `Imag`and handles `Question` different.

I noticed that too when I copied the CIL implementation over `Cilfacade`'s to see if it's out of date.


> Maybe we can work some super dirty workaround that temporarily sets `Errormsg.logChannel ` to something different (assuming we patch the uses of `bug` to be uses of `E.bug`)?

I tried to do that, but it didn't work: there was still output. Now I know why: `Cil.bug` doesn't use `logChannel`. I didn't even notice this difference, I just thought `Errormsg` was opened there...
Also, there's no null `out_channel` AFAIK, so one would have to `open_out "/dev/null"`, which is just so bad...

---

Given that we already have `Cilfacade.typeOf`, I think I'll update it and change all errors to a custom descriptive exception types (instead of `Not_found`) and make all `Cil.typeOf` calls go to `Cilfacade.typeOf` instead. I'll do this in a separate PR because this one is already overwhelming enough and otherwise seems to now behave as needed.

src/analyses/base.ml
```diff
@@ -781,14 +832,29 @@ struct
...
   (* Wherever possible, don't use this but the query system or normal eval_rv instead. *)
-  let eval_exp x (exp:exp) =
+  let eval_exp st (exp:exp) =
     (* Since ctx is not available here, we need to make some adjustments *)
-    let knownothing = { Queries.f = fun (type a) (q: a Queries.t) -> Queries.Result.top q } in (* our version of ask *)
-    let gs = fun _ -> G.top () in (* the expression is guaranteed to not contain globals *)
-    match (eval_rv knownothing gs x exp) with
+    let rec query: type a. a Queries.t -> a Queries.result = fun q ->
+      match q with
+      | EvalInt e -> query_evalint ask gs st e (* mimic EvalInt query since eval_rv needs it *)
```
michael-schwarz, line 853: I'll leave this comment here on how it works for more complicated expressions, because it took me a while to follow the logic there.

```
E.g. EvalInt ((x+y)+z)
-> query_evalint ((x+y)+z)
   -> eval_rv_no_ask_evalint ((x+y)+z)
     -> eval_rv_ask_mustbeequal ((x+y)+z) (*)
       -> eval_rv_base ((x+y)+z)
         -> eval_rv (x+y) (**)
         -> eval_rv (z)
```

(*) and (**) use the ask that we provide that falls through to `eval_rv_base` all the time.

sim642:
It seems like this might finally be in merge-able condition although there is one more subtlety about this change regarding the extra `eval_rv` wrapper which turns `Bot` and `ArithmeticOnIntegerBot` into top values.
Previously this was applied also to `EvalInt` responses.
Now it's not applied to those, only direct `eval_rv` calls in base itself. I could change the new `EvalInt` to also do that, but then it would also do that conversion at every subexpression level, whereas previously it was only done at the outermost level, which would change some intermediate values from `Bot` to top.

I don't completely understand this weird conversion mechanism (seems non-monotonic), so I don't really know what the expected behavior should be.

sim642:
In the process of making #278 work, I answered my own question: the query response also needs to change bottoms into tops to supply sound values to other analyses. That then happens at every subexpression, not just the outermost, and nothing seemed to break because of this, so hopefully it's good.

src/cdomains/intDomain.ml
```diff
@@ -786,69 +786,68 @@ module IntIkind = struct let ikind () = Cil.IInt end
...
   let ne n1 n2 = of_bool (n1 <> n2)
-  let bitnot = Int64.lognot
-  let bitand = Int64.logand
-  let bitor  = Int64.logor
-  let bitxor = Int64.logxor
-  let shift_left  n1 n2 = Int64.shift_left n1 (Int64.to_int n2)
-  let shift_right n1 n2 = Int64.shift_right n1 (Int64.to_int n2)
+  let bitnot = Ints_t.lognot
+  let bitand = Ints_t.logand
+  let bitor  = Ints_t.logor
```
michael-schwarz, line 837: Is this correct ?!

For `int64`:

```
 val logor : int64 -> int64 -> int64  (* Bitwise logical or. *)
 ```
 
But for Ints_t we have `bitnot` etc.

Issue #248, opened 2021-06-12
Unsoundness of OctApron (& other results from a run)


michael-schwarz:
I ran SV-Comp with OctApron, for the following tests we reported **true** contrary to how the task is labelled:

- [x] bitvector-regression/implicitunsignedconversion-1.yml
- [x] bitvector-regression/signextension-1.yml
- [x] bitvector-regression/signextension2-2.yml
- [x] ldv-linux-3.7.3/main15_drivers-usb-core-usbcore-ko--32_7a--linux-3.7.3.yml

The base analysis does not have that problem.

Apart from that the changes are:
- 14 tasks where OctApron+Base succeed where Base alone doesn't :tada: 
- 1 additional case where we fail during verify
- 29 APRON Manager Exception
- 404 tasks with Exception(Failure)
- 3 FAILED
- 32 new OUT_OF_MEMORY (where runtime for base was minimal)
- (59 tests that work with base only timeout now)

I'll link to the result in a comment on Slack, GitHub doesn't let me add HTML files.

sim642:
From the results table, this is with octApron that's currently on master (https://github.com/goblint/analyzer/commit/a06a611c), not traces-relational.

Apparently we have an octApron regression test that doesn't actually work with octApron, but is based on signextension-1: https://github.com/goblint/analyzer/blob/847352e74a1f388f5a23074dd9bd16c47f85c547/tests/regression/36-octapron/08-problem-signed-unsigned-overflow.c.

michael-schwarz:
> From the results table, this is with octApron that's currently on master (a06a611), not traces-relational.

Yes, probably easier to first fix these bugs as I guess they are easier to debug than those that depend on privatization. 

Those that fail with `Exception(Failure)` are also related to octApron directly it seems;
`Fatal error: exception Failure("Environment.dim_of_var: unknown variable in the environment")`

sim642:
I looked at the first Apron Manager exception (missing variable in environment) one and it boils down to this:
```c
double copysign_double(double x, double y) {
  return x;
}

int main() {
  // oct is top here, so entry state of copysign_double is also top (doesn't contain x')
  copysign_double(0, 0); // works when combine has no lval
  // double z = copysign_double(0, 0); // doesn't work when combine has lval
}
```

Many of the Failure exceptions might be for related reasons, just on different operations, which raise a different exception.

[...]

PR #253, opened 2021-06-16
Remove floats from octApron

sim642:
Follow-up for https://github.com/goblint/analyzer/issues/248#issuecomment-860471640.

Since Goblint doesn't even analyze floats non-relationally in base, we shouldn't try to make octApron do it either. The old implementation from poly tried to support them to some extent because Apron supposedly does, but it doesn't know about NaN, which causes problems. Therefore it's the easiest to remove their support in octApron for now.

It also simplifies a bunch of logic in octApron because one doesn't have to constantly manage the integer and floating-point variables in the Apron environment separately (and have to think about weird situations where they might be named the same etc).

### DIFF
--- a/src/analyses/apron/octApron.apron.ml
+++ b/src/analyses/apron/octApron.apron.ml
@@ -54,7 +54,7 @@ struct
       let f = Cilfacade.getdec f in
       let is = D.typesort f.sformals in
       let is = is @ List.map (fun x -> x^"'") is in
-      let newd = D.add_vars ctx.local (is,[]) in
+      let newd = D.add_vars ctx.local is in
       let formargs = Goblintutil.zip f.sformals args in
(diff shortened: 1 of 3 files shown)


ivanazuzic:
> Follow-up for [#248 (comment)](https://github.com/goblint/analyzer/issues/248#issuecomment-860471640).
> 
> Since Goblint doesn't even analyze floats non-relationally in base, we shouldn't try to make octApron do it either. The old implementation from poly tried to support them to some extent because Apron supposedly does, but it doesn't know about NaN, which causes problems. Therefore it's the easiest to remove their support in octApron for now.
> 
> It also simplifies a bunch of logic in octApron because one doesn't have to constantly manage the integer and floating-point variables in the Apron environment separately (and have to think about weird situations where they might be named the same etc).

I looked at what you changed. Seems ok. I think you can merge it.

PR #255, opened 2021-06-17
Rework system outputting warnings 

### LINKED ISSUE #201: Rework system outputting warnings
Categories should be introduced to create a better understanding of the warnings.

#198 #199 #200 
@vandah
@edincitaku

Wherekonshade:
This pull request closes #201
We introduced a must or may tag and the following categories:

        - Behavior
            * Undefined
		* ArrayOutOfBounds
			     * PastEnd
			     * BeforeStart
			     * Unknown
		* NullPointerDereference
		* UseAfterFree
		
        - Implementation
        - Machine

        -Integer
            * Overflow
            * Div_by_zero

        -Cast
            * Type_mismatch (bug)

        -Race

~~We introduced the warning as mywarn, which can be renamed to the normal warn once all warnings have been replaced.~~

The user can enable/disable the following flags(enabled by default) to filter warnings:

- dbg.warn.may
- dbg.warn.must
- dbg.warn.behavior
- dbg.warn.integer
- dbg.warn.cast
- dbg.warn.race
- dbg.warn.array

The structure inside messages.ml is the following:
module LogEvent with two fields:
    • certainty: May | Must (optional, if it's None, then it's always printed -> this is used for debugging)
    • event_type: Behavior | Integer | ... (these can bear additional information but it needs to be determined what information is useful for which warning type)
also has methods .may and .must to shorten the creation of may/must warnings (then it only takes the event_type  as an argument)

Calling the warnings should be rather intuitive but here are some examples:
`M.warn @@ M.Warning.Unknown "Extra info can be passed here";`
`M.warn_each ~must:true @@ warn_type M.ArrayOOB.PastEnd;`

Path sensitivity was not taken into account as it was ruled out of scope.

But there are still some things open to discussion in our opinion:
~~1. The flags would also make sense in Std in our opinion~~ We will introduce a new category as suggested by @sim642 
2.  What other categories should be introduced ?  
3. Should the user be able to enable/disable subcategories ? Right now he can enable/disable the behavior category but not the subcategory of NullPointerDereference for example


src/cdomains/valueDomain.ml
```diff
@@ -324,7 +324,8 @@ struct
               a (* probably garbage, but this is deref's problem *)
               (*raise (CastError s)*)
             | SizeOfError (s,t) ->
-              M.warn_each("size of error: " ^  s ^ "\n");
+              M.warn_each (M.LogEvent.may (M.EventType.Unknown ("size of error: " ^  s)));
+              (* old warning: M.warn_each("size of error: " ^  s ^ "\n");*)
```
sim642, line 328: Why is this commented out old version still here?

src/util/defaults.ml
```diff
@@ -228,6 +228,15 @@ let _ = ()
...
       ; reg Debugging "dbg.test.domain"     "false" "Test domain properties"
+      ; reg Debugging "dbg.warn.behavior"        "true"  "undefined behavior warnings"
+      ; reg Debugging "dbg.warn.integer"        "true"  "integer (Overflow, Div_by_zero) warnings"
+      ; reg Debugging "dbg.warn.cast"        "true"  "Cast (Type_mismatch(bug) warnings"
+      ; reg Debugging "dbg.warn.race"        "true"  "Race warnings"
+      ; reg Debugging "dbg.warn.array"        "true"  "Array (Out_of_bounds of int*int) warnings"
+      ; reg Debugging "dbg.warn.unknown"        "true"  "Unknown (of string) warnings"
+      ; reg Debugging "dbg.warn.debug"        "true"  "Debug (of string) warnings"
+      ; reg Debugging "dbg.warn.may"        "true"  "Enable or disable may warnings"
+      ; reg Debugging "dbg.warn.must"        "true"  "Enable or disable must warnings"
```
sim642, line 239: It probably makes sense to have a separate category for warnings instead of shoving them all under debugging.

src/util/messages.ml
```diff
@@ -3,6 +3,137 @@ open Pretty
...
+  let should_warn e =
+    get_bool ("dbg.warn." ^ (to_string e))
+
+  let show c =
+    match c with
+    | May -> "[MAY]"
+    | Must -> "[MUST]"
+end
+
+module LogEvent =
```
sim642, line 115: Why is this called `LogEvent`?

src/analyses/base.ml
```diff
@@ -1324,7 +1324,7 @@ struct
     let inv_bin_int (a, b) ikind c op =
       let warn_and_top_on_zero x =
         if GU.opt_predicate (BI.equal BI.zero) (ID.to_int x) then
-          (M.warn "Must Undefined Behavior: Second argument of div or mod is 0, continuing with top";
+          (M.warn (M.LogEvent.may (M.EventType.Unknown ("Must Undefined Behavior: Second argument of div or mod is 0, continuing with top")));
```
sim642, line 1327: What's the point of defining `Certainty.Must` and `BehaviorEvent` if they're not properly used? Same goes for many other `EventType`s.

src/util/messages.ml
```diff
@@ -3,6 +3,137 @@ open Pretty
...
+
+module CastEvent =
+struct
+  type t = TypeMismatch
+end
+
+module ArrayEvent =
+struct
+  type t = OutOfBounds of int*int
+end
```
sim642, line 61: It's quite clear that these aren't properly used if they don't have `show` implemented and everything is still fine.

src/util/messages.ml
```diff
@@ -3,6 +3,137 @@ open Pretty
...
+    | Integer _ -> "[Integer]"
+    | Race -> "[Race]"
+    | Array _ -> "[Array]"
+    | Cast _ -> "[Cast]"
+    | Unknown msg -> Printf.sprintf "[Unknown] %s" msg
+    | Debug msg -> Printf.sprintf "[Debug] %s" msg
+end
+
+module Certainty = struct
+  type t = May | Must
```
sim642, line 99: I don't know where, but somewhere we must it _very_ clear that this `Must` doesn't truly mean "must" as commonly understood in program analysis. Not only does this have to be clear to every Goblint developer in the code, but somehow also to any user just running Goblint and reading the output.

Maybe it just needs a new name like `MustIfReach`.

sim642 (commented):
In https://github.com/goblint/analyzer/issues/201#issuecomment-858764024 it was said:
> 3. Reports and warnings should be merged with this new approach.

Yet `Messages.report` is still a thing.

src/util/messages.ml
Wherekonshade: We would be happy to change it, if you have a different suggestion ? :)

src/util/messages.ml
Wherekonshade: We discussed putting this disclaimer into the readme or the docs but a new name as suggested sounds like the best solution to us. We'll do that, thank you

src/analyses/region.ml
```diff
@@ -27,7 +27,7 @@ struct
       let ev = Reg.eval_exp exp in
       let to_exp (v,f) = (v,Lval.Fields.to_offs' f) in
       List.map to_exp (Reg.related_globals ev (part,reg))
-    | `Top -> Messages.warn "Region state is broken :("; []
+    | `Top -> Messages.warn (Messages.LogEvent.may (Messages.EventType.Unknown ("Region state is broken :("))); []
```
vogler, line 30: This seems too verbose. Not specific to this, but in general.
Now with the non-polymorphic variants, there's no way around the path, but maybe change all those calls to 
`Messages.(warn (LogEvent.may (EventType.Unknown ("Region state is broken :("))))`?

src/util/messages.ml
vogler: Why are any of those called Event? IMO sth like `Warn.Integer.Overflow` would be more readable than `M.EventType (M.EventType.Integer M.IntegerEvent.Overflow)`.
Of course if you want to filter `Integer` as a group, it can't be flat but you need some run-time representation. Could define some shortcuts or even generate them.

vogler: Also see https://github.com/goblint/analyzer/issues/201#issuecomment-859452423
I would also nest those modules to reflect the hierarchy of the types. But then you have problems with the different paths w/o shortcuts.

michael-schwarz: Maybe this should be called `Warning`?

vandah: `Warning` or `Warn` sound good :+1: 
I originally named it LogEvent since I associate `warn`, `debug` etc with logging.

src/analyses/base.ml
vandah: We have currently only changed our newly added warnings to their proper types. These changes with the unknowns are for backwards compatibility to make it compile as we were not sure what exact types should be which of these lines. This line you commented on is especially tricky since the text says _Undefined behavior_ but I think in the categories it should be _Integer.DivByZero_ ?

src/util/messages.ml
vandah: yes, these are just prepared here to be extended as we weren't exactly sure where to use them or what information would be relevant to pass into them. Plus for example, we now have _Array out of bounds_ under _Undefined behavior_ and I actually don't know which one should it be - whether array out of bounds should be in this Array event or whether it should belong under the undefined behavior which it also is.

src/util/messages.ml
vandah: Wouldn't `MustIfReach` imply that whenever a line is reached then this event occurs? Whereas in the following example the line with the array access is reached multiple times but only once it's a `Must` out of bounds access.

```
void f(int n) {
  int arr[5];
  arr[n] = 1; // this line is reached 3 times and in one case it's in bounds, in one certainly out of bounds and in the last one it could be anything
}
int main() {
  f(3); // -> this results in an in bounds access
  f(10); // -> this results in a Must Warning in the function f
  int top;
  f(top); // -> this results in a May Warning in the function f 
}
``` 

What I'd like to see here ideally would be something like

```
arr[n] = 1; is out of bounds following the function call f(10);
```

src/analyses/region.ml
vandah: I think in most modules `Messages` was imported as just `M`. Would maybe some similar shortcuts help with the verbosity too? Or bring some more of the types a bit higher like `M.warn_may` and `M.warn_must` or `may_unknown`, `must_integer` and `may_integer` ?

src/analyses/base.ml
vogler: > it should be `Integer.DivByZero`

Which is undefined behavior (same goes for `Integer.Overflow`).
Those `Event` modules need some cleanup/reordering.
If each of those modules just consists of a `type t` and `show` for it, you could also just put all types in one module and name the types accordingly.

src/util/messages.ml
vogler: I would put it in undefined behavior. Most of our warnings prob. are about UB. Which ones aren't besides asserts?
If we had some warnings in `Array` and `UB.Array`, it'd be cumbersome to filter for both.
If we have such cases, we could let `UB.Array` imply `Array` for filtering.

src/analyses/region.ml
vogler: Yes. You can also just define aliases instead of pulling them out. Or as I wrote above, just put all types in one module. If there's only going to be the type and `show` for it, modules for each just makes it unnecessarily verbose (I assume they're not going to be put in some functor).

> I think in most modules Messages was imported as just M.

I meant the repeated `M.x M.y M.z` instead of `M.(x y z)`.

src/util/defaults.ml
vandah: It is now a separate category, but unfortunately just `warn` is already used in `Std` so I named the option `warn_filter`.

src/util/messages.ml
vogler: Yes, there's `dbg.warn_with_context` which is off by default and currently only supplied the context for asserts.
If you turn it on and add `~ctx`, the warning should include the context of `f(10)`.
Ideally, the result viewer would then take care to indicate such things.

src/cdomains/valueDomain.ml
vandah: I didn't touch the commented out code as that's not something included in the compilation so if I made a typo in there it would be just a mess if somebody wanted to use it later. I'm not even sure why these commented out are there anyway?

Should the commented out lines be also changed then? There's only a few with the warnings.

src/util/defaults.ml
vogler: `warn` should not be needed as often, so it could be renamed to `warn_at` or sth similar which is more descriptive anyway.

src/util/messages.ml
vandah: okay I changed `LogEvent` to `Warning`, `EventType` to `WarnType` and the others to the version without `Event` too - such as `IntegerEvent` to `Integer`. Is it okay like this?

src/cdomains/valueDomain.ml
vogler: From the diff it looks like the comment is an added 'backup copy' of the removed line.

src/analyses/region.ml
vogler: Another way to make it less verbose would be to have the certainty as an optional argument for `warn` and `warn_each` as it is already done with `?ctx`:
`M.(warn Integer "...")`
`M.(warn ~must:true Integer "...")`
I assume we produce more `may` than `must`.

src/util/defaults.ml
vandah: alright, I renamed `warn` to `warn_at` and put all the options for our warnings into just `warn` instead of the `warn_filter`

sim642: Seems to me like the options could even be `warn.at` and `warn.filter` also in the warning options category.

src/util/messages.ml
```diff
@@ -128,8 +259,20 @@ let with_context msg = function
...
+
+(*
+let warn_each_ctx ctx msg = (* cyclic dependency... *)
+  if not @@ GobConfig.get_bool "dbg.warn_with_context" then warn_each msg else
+  (* let module S = (val Control.get_spec ()) in *)
+  (* warn_each (msg ^ " in context " ^ S.C.short 99999 (Obj.obj ctx.context ())) *)
+  (* warn_each (msg ^ " in context " ^ IO.to_string S.C.printXml (Obj.obj ctx.context ())) *)
+  warn_each (msg ^ " in context " ^ string_of_int (Hashtbl.hash (Obj.obj ctx.context ())))
+*)
+
```
jerhard, line 271: Maybe you can delete this?

src/util/defaults.ml
vogler: Maybe `warn.on` since `warn.filter` can be confused as filter out or only those.

src/util/messages.ml
```diff
@@ -149,19 +293,10 @@ let warn_each ?ctx msg =
...
-  if not @@ GobConfig.get_bool "dbg.warn_with_context" then warn_each msg else
-  (* let module S = (val Control.get_spec ()) in *)
-  (* warn_each (msg ^ " in context " ^ S.C.short 99999 (Obj.obj ctx.context ())) *)
-  (* warn_each (msg ^ " in context " ^ IO.to_string S.C.printXml (Obj.obj ctx.context ())) *)
-  warn_each (msg ^ " in context " ^ string_of_int (Hashtbl.hash (Obj.obj ctx.context ())))
-*)
-
 let debug msg =
-  if (get_bool "dbg.debug") then warn ("{BLUE}"^msg)
+  if (get_bool "dbg.debug") then warn (Warning.debug ("{BLUE}"^msg))
```
jerhard, line 297: We discussed this yesterday in the meeting. I suggest you leave the `debug` and `debug_each` as you have implemented it right now.

src/analyses/region.ml
vandah: The optional `must` parameter is now there. I also added some nesting and helper methods to build the types from the bottom so now it should be possible to do:

```
M.warn (M.Warning.Integer.overflow ());
M.warn (M.Warning.Behavior.Undefined.nullpointer_dereference ());
M.warn ~must:true (M.Warning.Behavior.Undefined.ArrayOutOfBounds.before_start ());
```

also the variants are now not hidden inside modules so the top level variants can be used directly as:

```
M.warn (M.Unknown "my message");
```

src/util/messages.ml
vandah: I felt like the combination of `WarnType` and `Warning` was a little confusing after all and changed it to `Warning` and `WarningWithCertainty`. The `warn` and `warn_each` functions now accept `Warning` and an optional bool parameter `must`. Then the `WarningWithCertainty` is constructed from that internally. So the users should never have to type this long module name and only use the `Warning` module.

src/util/messages.ml
vandah: it's removed now :+1:

src/analyses/region.ml
jerhard: @vandah Previously, there was the `Certainty.t option`, which would allow for `None` in case of "Debug" warnings. How is this handled now?

vandah: This remained there. The `debug` function still works the same.

Before, I just called `warn` from the `debug` function. Now there's a function `warn_internal` which does what `warn` did previously and is called from both `warn` and `debug`. This `warn_internal` takes the type `WarningWithCertainty.t` which still has the `Certainty.t option` inside.

src/util/messages.ml
```diff
@@ -3,6 +3,173 @@ open Pretty
...
+    | Debug msg -> "debug"
+    | Analyzer -> "analyzer"
+
+  let should_warn e =
+    get_bool ("warn." ^ (to_string e))
+
+  let show e =
+    match e with
+    | Behavior behavior -> Printf.sprintf "[Behavior] %s" (Behavior.show behavior)
+    | Integer _ -> "[Integer]"
```
jerhard, line 128: Is there a reason why the parameter is not used for `Integer` and `Cast`?

src/util/messages.ml
```diff
@@ -3,6 +3,173 @@ open Pretty
...
+type cast = TypeMismatch
+
+type warning =
+  | Behavior of behavior
+  | Integer of integer
+  | Race
+  | Cast of cast
+  | Unknown of string
+  | Debug of string
+  | Analyzer
```
jerhard, line 32: What is the difference in meaning for `Debug` and `Analyzer`?

src/util/messages.ml
```diff
@@ -3,6 +3,173 @@ open Pretty
...
+    | Behavior _ -> "behavior"
+    | Integer _ -> "integer"
+    | Race -> "race"
+    | Cast _ -> "cast"
+    | Unknown msg -> "unknown"
+    | Debug msg -> "debug"
+    | Analyzer -> "analyzer"
+
+  let should_warn e =
+    get_bool ("warn." ^ (to_string e))
```
jerhard, line 123: It probably makes sense to define `to_string` within `should_warn`, as it is only used there.

Also, if you don't use a matched value, you should use `_` instead of a variable name (see the `msg` in your code).

src/util/messages.ml
```diff
@@ -3,6 +3,173 @@ open Pretty
...
+module Certainty = struct
+  type t = May | Must
+
+  let to_string e =
+    match e with
+    | May -> "may"
+    | Must -> "must"
+
+  let should_warn e =
+    get_bool ("warn." ^ (to_string e))
```
jerhard, line 145: Same applies here.

src/util/messages.ml
```diff
@@ -149,19 +318,19 @@ let warn_each ?ctx msg =
...
 let debug msg =
-  if (get_bool "dbg.debug") then warn ("{BLUE}"^msg)
+  if (get_bool "dbg.debug") then warn_internal (WarningWithCertainty.debug ("{BLUE}"^msg))
 
 let debug_each msg =
-  if (get_bool "dbg.debug") then warn_each ("{blue}"^msg)
+  if (get_bool "dbg.debug") then warn_internal_with_loc (WarningWithCertainty.debug ("{blue}"^msg))
+
+let mywarn () =
+  warn (Warning.Behavior.Undefined.nullpointer_dereference ())
```
jerhard, line 334: Please remove this.

src/util/messages.ml
vandah: just not finished yet

src/util/messages.ml
vandah: I expect `Analyzer` to have subtypes later as I think that most current warnings are probably Analyzer warnings? Or at least I feel like most doesn't fall into any other categories.

src/util/messages.ml
vandah: they're now used

vesalvojdani:
I added to our GobCon agenda that we should discuss the follow-up work to this pull request next meeting. There is still a lot of open questions about the warning system, so it would probably be best to merge this pull request, and then proceed with somewhat smaller pull requests and discussions on all the remaining questions.

PR #256, opened 2021-06-17
Nullpointer dereference

### LINKED ISSUE #199: Detect Null-Pointer Dereferencation
**Current situation**

Null-Pointer Dereferencation

```
// Program to demonstrate 
//Null-Pointer Dereferencation
int main(int argc, char ** argv){
  char buf[255];
  char * ptr = NULL;
  if (argc>1) {
    ptr = argv[1];
  }
  strcpy(str,ptr);
  return 0;
}

Output :
Runtime Error : Segmentation Fault (SIGSEGV)

```

Goblint can detect Null-Pointer Dereferencation

@vandah
@EdinCitaku

EdinCitaku:
Fixes #199 
In this PR we add detection for nullpointer dereference and add regression tests for this.
We differentiate between may and must, the detection happens inside of the base analysis and it is under the flag `dbg.warn.behavior`.
The detection happens inside of `eval_lv`.
The must warning gets printed when there exists at least one path and context where the nullpointer dereferenciation must happen when reaching the program point it in the corresponding context and with a value described by the predecessor.
The may warning gets printed when we are not sure that the pointer is not null.


src/analyses/base.ml
```diff
@@ -657,7 +660,6 @@ struct
               VD.top () (* upcasts not! *)
           in
           let v' = VD.cast t v in (* cast to the expected type (the abstract type might be something other than t since we don't change addresses upon casts!) *)
-          M.tracel "cast" "Ptr-Deref: cast %a to %a = %a!\n" VD.pretty v d_type t VD.pretty v';
```
sim642, line 660: Why is this removed? If anything, it should actually have `if M.tracing then ` added to it.

src/analyses/base.ml
```diff
@@ -760,8 +762,14 @@ struct
      * evaluate [(\*exp).subfield]. We first evaluate [exp] to { (x,field) }
      * and then add the subfield to it: { (x,field.subfield) }. *)
     | Mem n, ofs -> begin
+      let warn_type = (M.EventType.Behavior (M.BehaviorEvent.Undefined (M.UndefinedBehavior.NullPointerDereference))) in
         match (eval_rv a gs st n) with
-        | `Address adr -> do_offs (AD.map (add_offset_varinfo (convert_offset a gs st ofs)) adr) ofs
+        | `Address adr -> 
+          if GobConfig.get_bool "ana.nullptr" then 
+            if (AD.is_null adr) then M.warn_each (M.LogEvent.may warn_type)
+            else if( AD.may_be_null adr) then M.warn_each (M.LogEvent.must warn_type);
```
sim642, line 770: Aren't may and must switched here?

src/analyses/base.ml
```diff
@@ -783,7 +791,7 @@ struct
 
   (* Evaluate an expression containing only locals. This is needed for smart joining the partitioned arrays where ctx is not accessible. *)
   (* This will yield `Top for expressions containing any access to globals, and does not make use of the query system. *)
-  (* Wherever possible, don't use this but the query system or normal eval_rv instead. *)
+  (* Whereve  r possible, don't use this but the query system or normal eval_rv instead. *)
```
sim642, line 794: Accidental typo.

src/util/defaults.ml
```diff
@@ -134,6 +134,7 @@ let _ = ()
       ; reg Analyses "ana.sv-comp.functions" "false" "Handle SV-COMP __VERIFIER* functions"
       ; reg Analyses "ana.specification"   "" "SV-COMP specification (path or string)"
       ; reg Analyses "ana.wp"              "false" "Weakest precondition feasibility analysis for SV-COMP violations"
+      ; reg Analyses "ana.nullptr"        "false"  "Null Pointer dereferencing check"
```
sim642, line 137: Is it really necessary to have this option here? The warning options already control whether to show this (among unknown pointer dereferencing warnings, which don't have such corresponding option) and it's just one extra check.

tests/regression/26-undefined_behavior/09-nullpointer-dereference-struct.c
```diff
@@ -0,0 +1,21 @@
+// PARAM: --enable ana.nullptr --enable dbg.debug
```
sim642, line 1: Don't regression tests have `dbg.debug` universally enabled?

tests/regression/26-undefined_behavior/12-nullpointer-dereference-array.c
```diff
@@ -0,0 +1,23 @@
...
+
+#include <stdio.h>
+// source base:
+// https://stackoverflow.com/questions/4007268/what-exactly-is-meant-by-de-referencing-a-null-pointer
+int main() {
+  int*  arr[] = {NULL, NULL, NULL, NULL, NULL, NULL};
+  int a, b, c, x; // some integers
+  int *pi;        // a pointer to an integer
+  int *ok;
+  a = arr[2];
```
sim642, line 11: `arr[2]` has type `int*` while `a` has type `int`, is this intentional?

tests/regression/26-undefined_behavior/12-nullpointer-dereference-array.c
```diff
@@ -0,0 +1,23 @@
...
+  ok = &a;
+  pi = &a; // pi points to a
+  b = *pi; // b is now 5
```
sim642, line 14: Where does the 5 come from?

sim642 (commented):
Are you sure that the single check in `eval_lv` covers all the possible ways a NULL pointer can be accessed?

For example, `reachable_from_value` currently just silently removes `NullPtr` without emitting any warnings.

src/analyses/base.ml
EdinCitaku: good catch!

EdinCitaku:
> Are you sure that the single check in `eval_lv` covers all the possible ways a NULL pointer can be accessed?
> 
> For example, `reachable_from_value` currently just silently removes `NullPtr` without emitting any warnings.

It covers all regression tests we created. I don`t know about `reachable_from_value`, do u maybe know a case where it is called? Maybe a short snippet in C where our test would not detect a nullpointer dereference?

src/util/defaults.ml
EdinCitaku: sure I can remove the flag!

tests/regression/26-undefined_behavior/09-nullpointer-dereference-struct.c
EdinCitaku: for me they fail if I dont have this enabled

sim642: Hmm, apparently it depends on some weird conditions in the regression testing script:
https://github.com/goblint/analyzer/blob/3690e97b968298cf7854fe12875647743bf2428b/scripts/update_suite.rb#L139-L186

Namely, if the program contains an `assert` anywhere in it, then `dbg.debug` is enabled. Most tests contain some assertions so that's why it seemed to me like that's universally the case.

I'm not sure why this needs to be so complicated, if it's up to me, I'd just make the script always enable it, but maybe @michael-schwarz has a different thought?

sim642:
> It covers all regression tests we created. I don't know about `reachable_from_value`, do u maybe know a case where it is called? Maybe a short snippet in C where our test would not detect a nullpointer dereference?

One way it's called is from `invalidate`, which is used when a library function, which modifies its arguments, is passed some pointers. For example in this case:
```c
// PARAM: --enable ana.nullptr  --enable dbg.debug
#include <string.h>

int main() {
  memset(NULL, 42, 10); // WARN
  return 0;
}
```

tests/regression/26-undefined_behavior/10-nullpointer-dereference-simple.c
```diff
@@ -0,0 +1,12 @@
+// PARAM: --enable ana.nullptr  --enable dbg.debug
+#include <stdio.h>
+int main(){    
+    int *pi;     // a pointer to an integer
+    int *t;
+    int a = *t; // WARN
+    t = &pi;
```
jerhard, line 7: An `int**` is assigned to an `int*`. Is this intended?

tests/regression/26-undefined_behavior/09-nullpointer-dereference-struct.c
michael-schwarz: > I'd just make the script always enable it, but maybe @michael-schwarz has a different thought?
No, I agree!

jerhard:
Maybe you can add some test cases where a possible NULL-pointer is dereferenced when it is passed as an argument to a function call?

src/util/defaults.ml
EdinCitaku: removed the flag! :)

tests/regression/26-undefined_behavior/09-nullpointer-dereference-struct.c
EdinCitaku: alright, fixed it for all regression tests!

EdinCitaku:
> Maybe you can add some test cases where a possible NULL-pointer is dereferenced when it is passed as an argument to a function call? (E.g.: `foo(*p)`)

Created a new regression test for that!

EdinCitaku:
>One way it's called is from invalidate, which is used when a library function, which modifies its arguments, is passed some pointers. For example in this case:

@sim642 We had a discussion with the tutors (@michael-schwarz  and @jerhard) and came to the conclusion that this would be out of scope for the PR. Hope that's ok for you!

src/analyses/base.ml
```diff
@@ -764,9 +764,11 @@ struct
...
         match (eval_rv a gs st n) with
         | `Address adr -> 
-            if AD.is_null adr then M.warn_each (M.LogEvent.must warn_type)
-            else if AD.may_be_null adr then M.warn_each (M.LogEvent.may warn_type);
-          do_offs (AD.map (add_offset_varinfo (convert_offset a gs st ofs)) adr) ofs
+          if AD.is_null adr 
+            then M.warn_each (M.LogEvent.must warn_type)
+          else if AD.may_be_null adr 
+            then M.warn_each (M.LogEvent.may warn_type);
+            do_offs (AD.map (add_offset_varinfo (convert_offset a gs st ofs)) adr) ofs
```
jerhard, line 771: The `do_offs` is not only executed when the `else if` is executed, right? So the indentation is a bit misleading.

EdinCitaku: No it always is executed, Ill add brackets to make it more clear!

PR #257, opened 2021-06-17
Detect Array out of bounds

### LINKED ISSUE #198: Detect Array out of bounds
**Current situation**

-Array-Bounds (Index out of bounds access; Source: [https://www.geeksforgeeks.org/accessing-array-bounds-ccpp/](https://www.geeksforgeeks.org/accessing-array-bounds-ccpp/) )

There are 2 general ways this can happen: 
1. Access non allocated location of memory: The program can access some piece of memory which is owned by it:
```c

// Program to demonstrate 
// accessing array out of bounds
#include <stdio.h>
int main()
{
    int arr[] = {1,2,3,4,5};
    printf("arr [0] is %d\n", arr[0]);
      
    // arr[10] is out of bound
    printf("arr[10] is %d\n", arr[10]);
    return 0;
}

Output :

arr [0] is 1
arr[10] is -1786647872
```
2.  Segmentation fault: The program can access some piece of memory which is not owned by it, which can cause crashing of program such as segmentation fault

```
// Program to demonstrate 
// accessing array out of bounds
#include <stdio.h>
int main()
{
    int arr[] = {1,2,3,4,5};
    printf("arr [0] is %d\n",arr[0]);
    printf("arr[10] is %d\n",arr[10]);
      
    // allocation memory to out of bound 
    // element
    arr[10] = 11;
    printf("arr[10] is %d\n",arr[10]);
    return 0;
}

Output :
Runtime Error : Segmentation Fault (SIGSEGV)
```

Goblint can detect this undefined Behavior. This will be optional.

@vandah
@edincitaku

Wherekonshade:
This PR fixes #198 
We introduced detection for out of bounds accesses in arrays. This works for both statically and dynamically sized arrays. With interval analysis enabled we can also properly detect out of bounds accesses inside for loops or other constructs taken into account by the interval analysis. 
We added regression tests for the following cases:

- basic
- pointer
- multidimensional
- dynamically sized
- pointer to arrays of different sizes
- arrays within structures

The warnings fall into the `behavior` category and in the new warning system, they can be disabled with `--disable dbg.warn.behavior` along with other Behavior warnings.
The warnings are also tagged as May/Must according to the certainty with which we know that the situation occurs.
The warning can also specify if the access happened PastEnd, BeforeStart or Unknown


src/analyses/fileUse.ml
```diff
@@ -11,6 +11,7 @@ struct
   module D = FileDomain.Dom
   module C = FileDomain.Dom
   module G = Lattice.Unit
+  module M = Messages
```
sim642, line 14: Why? This doesn't come from #255.

tests/regression/26-undefined_behavior/02-array-out-of-bounds.c
```diff
@@ -0,0 +1,17 @@
+// PARAM: --set ana.activated "['base','threadid','threadflag','escape','mallocWrapper']" --set dbg.debug true --enable ana.arrayoob --enable ana.int.interval --enable ana.int.enums
```
sim642, line 1: Is it really necessary to explicitly list the activated analyses here?

tests/regression/26-undefined_behavior/02-array-out-of-bounds.c
```diff
@@ -0,0 +1,17 @@
...
+#include <stdio.h>
+//This is the most basic case
+int main()
+{
+    int arr[] = {1, 2, 3, 4, 5, 6};
+    arr[2] = 0; //NOWARN
+    arr[6] = 10; //WARN
+    arr[-1] = 10; //WARN
+    for (int i = 0; i < 5; ++i)
+
```
sim642, line 11: What is this?

src/cdomains/arrayDomain.ml
```diff
@@ -552,6 +552,29 @@ struct
...
   let update_length _ x = x
 end
+(* This is the main array out of bounds check*)
+let array_oob_check ( type a ) (module Idx: IntDomain.Z with type t = a) (x, l) (e, v) =
+  if GobConfig.get_bool "ana.arrayoob" then (* The purpose of the following 2 lines is to give the user extra info about the array oob *)
+    let idx_before_end = Idx.to_bool (Idx.lt v l) (* check whether index is before the end of the array *)
+    and idx_after_start = Idx.to_bool (Idx.ge v (Idx.of_int Cil.ILong BI.zero)) in (* check whether the index is non-negative *)
+(* For an explanation of the warning types check the Pull Request #255 *)
+    let warn_type t = (M.EventType.Behavior (M.BehaviorEvent.Undefined (M.UndefinedBehavior.ArrayOutOfBounds t))) in
+    let () =
```
sim642, line 562: This unit let is unnecessary around here.

src/util/defaults.ml
```diff
@@ -135,6 +135,8 @@ let _ = ()
       ; reg Analyses "ana.specification"   "" "SV-COMP specification (path or string)"
       ; reg Analyses "ana.wp"              "false" "Weakest precondition feasibility analysis for SV-COMP violations"
       ; reg Analyses "ana.octapron.no_uints"    "false"  "Use OctApron without tracking unsigned integers."
+      ; reg Analyses "ana.arrayoob"        "false"        "Array out of bounds check"
+      ; reg Analyses "ana.octapron.no_floats"   "true"  "Use OctApron without tracking floats."
```
sim642, line 139: Why is `ana.octapron.no_floats` being added here?

Wherekonshade: This was probably caused by a weird merge order from our side.

vandah: it should be resolved now

docs/developer-guide/testing.md
```diff
@@ -23,6 +23,11 @@ To pass additional options to Goblint with `update_suite.rb`, use the `gobopt` e
 gobopt='--sets exp.privatization write+lock' ./scripts/update_suite.rb
 ```
 
+### Writing
+* If you introduce a new warning message, you need to manually add it to `./scripts/update_suite.rb` in the declaration at line 386, otherwise it won`t be recognized as a warning
+    * Example: ` when /Array out of bound/        then "warn"`
+* Add parameters to a regression test in the first line: `//PARAM: --set dbg.debug true`
+* Annotate lines inside the regression test with comments: `arr[9] = 10; //WARN`
```
jerhard, line 30: Maybe "Annotating tests" would be a better title for this section? Also the option `--set dbg.debug true` should already be activated by default by the regression test script, so maybe choose another example.

tests/regression/26-undefined_behavior/07-arrays-within-structures.c
```diff
@@ -0,0 +1,31 @@
...
+    float marks;
+  };
+  // Structure array declaration
+  struct student stu[] = {
+      {"vandah", 12, 89.5f},
+      {"edin", 15, 98.0f},
+      {"david", 17, 90.0f},
+  };  
+  // FIXME: this should be OK? but generates an AST error
+  // stu[6].name = "thisisareallylongstringjkdsfjbkslagjbfdghfdnvbfsöajgfklgjfgjlfdsjhhgjgfbldfhghlfdgjöjsadfjhsagjflghfkdsgjhsgblkjsadgfkjabgjlsagdhlshagljhdasghsadgjh"; // WARN
```
jerhard, line 18: An assignment of a string to a char array is not possible in this manner, so I would remove the commented-out line.

tests/regression/26-undefined_behavior/07-arrays-within-structures.c
```diff
@@ -0,0 +1,31 @@
...
+  stu[0].roll = 2; //NOWARN
+  stu[-1].roll = 10;    // WARN
+  stu[1].marks = 90.5f; //NOWARN
+  stu[20].marks = 89.5f; //WARN
+  //stu[1].name = "Edin"; //NOWARN
```
jerhard, line 23: Also here.

src/cdomains/arrayDomain.ml
```diff
@@ -563,6 +563,27 @@ struct
 
   let update_length _ x = x
 end
+(* This is the main array out of bounds check*)
+let array_oob_check ( type a ) (module Idx: IntDomain.Z with type t = a) (x, l) (e, v) =
+  if GobConfig.get_bool "ana.arrayoob" then (* The purpose of the following 2 lines is to give the user extra info about the array oob *)
+    let idx_before_end = Idx.to_bool (Idx.lt v l) (* check whether index is before the end of the array *)
+    and idx_after_start = Idx.to_bool (Idx.ge v (Idx.of_int Cil.ILong BI.zero)) in (* check whether the index is non-negative *)
```
jerhard, line 570: Maybe use `Cilfacade.ptrdiffikind ()` here instead of `Cil.ILong`? `Idx` uses that in the definition of its default `ikind`.

docs/developer-guide/testing.md
sim642: > Also the option `--set dbg.debug true` should already be activated by default by the regression test script, so maybe choose another example.

I thought that too here https://github.com/goblint/analyzer/pull/256#discussion_r660474588, but apparently it depends on some intricate conditions I'm not sure are necessary.

vogler: won`t -> won't
Also, 'at line 386' is fragile, better use a permalink which can also show the code inline.

PR #258, opened 2021-06-17
Detect use after free

### LINKED ISSUE #200: Generalized Use-After-Free in a Multithreaded setting
**Current situation**

Use-After-Free

There are 2 general ways this can happen: 
1. Runs alright but returns a random value 
```c

#include <stdio.h>
#include <stdlib.h>

int main() {
  int *a = malloc(sizeof(int));
  if (a != NULL) {
    *a = 1;
    int *b = a;
    free(b); // detects the free even when the pointer is aliased to b
    return *a; // use after ‘free’ of ‘a’
  }
  return 0;
}

Output:
```
2.  First couple values are somewhat random and the last six are left at a set value

```
#include<stdio.h>
#include<stdlib.h>

int main() {
  int* a = malloc(10*sizeof(int));
  for(int i = 0; i < 10; i++) {
    a[i] = 0xff;
  }

  free(a);

  for(int i = 0; i < 10; i++) {
    printf("%d ", a[i]); // use after free
  }
  printf("\n");
  return 0;
}

Output:

1530302590 5 1762123792 21939 255 255 255 255 255 255
```

Goblint can detect Use-After-Free

@vandah
@edincitaku

Wherekonshade:
This PR fixes #200 for the basic case.
We introduced a new spec file https://github.com/goblint/analyzer/commit/9f55a619757520a8ead7cfdd7aaa8f10d0ef90e0, which can detect use after free.
In the same folder regression test: tests/regression/26-undefined_behavior/14-Use_after_free.c is for the most basic case and works.
We also implemented the warnings from spec files into our reworked warning system (https://github.com/vandah/analyzer/commit/b57e08f5f5bd44554df68366c312cf7af2260b97)


sim642 (commented):
The added regression tests activate the file analysis. How is that related to the spec analysis?

vogler:
Also, the spec file is almost the same as the existing https://github.com/goblint/analyzer/blob/3690e97b968298cf7854fe12875647743bf2428b/tests/regression/19-spec/malloc.spec.
Maybe this should be in one place at the end.

Wherekonshade:
> The added regression tests activate the file analysis. How is that related to the spec analysis?

Sorry, we misunderstood the activation.

>Also, the spec file is almost the same as the existing https://github.com/goblint/analyzer/blob/3690e97b968298cf7854fe12875647743bf2428b/tests/regression/19-spec/malloc.spec.
Maybe this should be in one place at the end.

Yes, we used that as a base to understand the specification language. Now the spec file is a new one: https://github.com/goblint/analyzer/pull/258/commits/9f55a619757520a8ead7cfdd7aaa8f10d0ef90e0 :)

vogler:
> Yes, we used that as a base to understand the specification language. Now the spec file is a new one: [9f55a61](https://github.com/goblint/analyzer/commit/9f55a619757520a8ead7cfdd7aaa8f10d0ef90e0) :)

Ok, better for this PR. Just need to remember to combine it into malloc.spec in the end, since that's where it belongs.

Wherekonshade:
> Why is there only a transition for writes and not reads?

We tried read transitions but couldn't find one which would detect something.

vogler:
> We tried read transitions but couldn't find one which would detect something.

`3 -w1> 3 _ = *$p` does not work?

But yea, would need to add something to the spec parser to denote evaluation of a subexpression since it could appear anywhere, e.g. `f(1 + *p * 2);`.
That's probably why any 'read' warnings are missing 😄

michael-schwarz:
I would propose to close this PR without merging, as after merging the new warning system this only contains the new spec, that also is not sound in that it only recognizes writes after free.

If we want to do this properly again in the future, we could probably start a separate PR. What do you think @sim642 @vogler?

PR #260, opened 2021-06-17
Add congruence integer domain and enable mutual refinement of the integer domains

### LINKED ISSUE #197: Add Congruence Analysis
### Current situation
The current version of Goblint doesn't support analyzing congruence relations e.g., integer valued X is congruent to c modulo m. 
###  Motivation
The congruence analysis can be used to improve the results of other analyses e.g., constant propagation and evenness analysis.
###  Concrete steps to fix this issue
[to be filled in] 
###  Team collaborators
@felihong, @hydrogenoxide, @MartinWehking, @nicokn

### LINKED ISSUE #202: Enable mutual refinement of integer domains
### Motivation

The results of different analyses can be used to refine output of each other. E.g., if an interval analysis says an integer variable x is in [4, 6] and a congruence analyisis finds that x has to be congruent to 0 modulo 3,  x must be 6.

### Steps to fix this issue

Fix #197, [to be filled in]

### Team collaborators

@Felihong, @hydrogenoxide, @MartinWehking, @nicokn

hydrogenoxide:
WIP. There are still ~~many~~ some errors and the refinement only uses the integer and congruence domain at the moment.

This PR

- Adds a congruence integer domain, including regression tests  (fixes #197)
- Enables the mutual refinement of the integer domains (fixes #202)

The new domain can be enabled with `--enable ana.int.congruence`. By default, the analysis assumes that an overflow can occur so returns top in many cases. If the interval domain is enabled, it will use this information to check if no overflow happens which makes the result more precise.

`./goblint --html --enable ana.int.congruence  --enable ana.int.interval tests/regression/37-congruence/01-simple.c`


src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1876,470 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    | Some (c1,m1), Some (c2,m2) -> if (m1 =: Ints_t.zero) && (m2 =: Ints_t.zero) then
+        if c1 <: c2 then of_bool ik true else of_bool ik false
+      else top_bool
+
+  let lt ik x y =
+    let res = lt ik x y in
+    if M.tracing then  M.trace "mything" "Cong. less than : %a %a becomes %a \n" pretty x pretty y pretty res;
+    res
+
+  let invariant c x = failwith "unimplemented"
```
michael-schwarz, line 2301: Please patch these invaraints to return `None` (for now, actually it would be good to implement them). 
Otherwise, I can't run your tests with SV-COMP

michael-schwarz:
@Wherekonshade I can't assign you as a reviewer for some reason, consider yourself assigned though :wink:

michael-schwarz:
I did a run of sv-comp with your congruences analysis enabled. There are six cases where your analysis is unsound, and quite a lot of cases where the fixpoint can not be verified (usually hinting at a bug with `leq`). For the ones where the verification of a fixpoint worked before this PR.

I will send you a diff table between an old run and a run with your new domains enabled on Slack, I somehow can't attach it here.
It also includes all the logfiles Goblint produced during the run.
The input programs can be found at: https://github.com/sosy-lab/sv-benchmarks/

What we usually do is extract a minimal example from the failing test and then add that to the regression tests, so it is checked when running the regression tests.

make.sh
```diff
@@ -1,8 +1,9 @@
-#! /bin/bash
+#!/usr/bin/env bash
```
EdinCitaku, line 1: Are u sure u want this to be merged to master?
Wdy think @michael-schwarz

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1876,470 @@ module Enums : S with type int_t = BigInt.t = struct
...
+  let gt ik x y = match x, y with
+    | None, None -> bot_of ik
+    | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
+    | Some (c1,m1), Some (c2,m2) -> if (m1 =: Ints_t.zero) && (m2 =: Ints_t.zero) then
+        if c1 >: c2 then of_bool ik true else of_bool ik false
+      else top_bool
+
+  let gt ik x y =
+    let res = gt ik x y in
+    if M.tracing then  M.trace "mything" "Cong. greater than : %a %a becomes %a \n" pretty x pretty y pretty res;
```
EdinCitaku, line 2286: I think its best to remove tracing once ur finished

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1876,470 @@ module Enums : S with type int_t = BigInt.t = struct
...
+  let refine_with_congruence a b = a
+  let refine_with_excl_list a b = a
+  let refine_with_incl_list a b = a
 end
 
+module Congruence = CongruenceFunctor (BI)
+
+(* module Congruence32 =
+ *   IntDomWithDefaultIkind
+ *     (IntDomLifter (CongruenceFunctor (IntOps.Int64Ops))) (IntIkind) *)
```
EdinCitaku, line 2342: again here, its best to remove commented out lines before merging to master

src/util/defaults.ml
```diff
@@ -119,6 +119,7 @@ let _ = ()
       ; reg Analyses "ana.int.def_exc"      "true"  "Use IntDomain.DefExc: definite value/exclusion set."
       ; reg Analyses "ana.int.interval"    "false" "Use IntDomain.Interval32: (int64 * int64) option."
       ; reg Analyses "ana.int.enums"       "false" "Use IntDomain.Enums: Inclusion/Exclusion sets. Go to top on arithmetic operations (except for some easy cases, e.g. multiplication with 0). Joins on widen, i.e. precise integers as long as not derived from arithmetic expressions."
+      ; reg Analyses "ana.int.congruence"  "false" "Use IntDomain.Congruence: TBD"
```
EdinCitaku, line 122: whats the status of this?

tests/regression/37-congruence/04-casting.c
```diff
@@ -0,0 +1,18 @@
...
+#include <stdio.h>
+
+int main(){
+  int c = 128;
+  for (int i = 0; i < 1; i++) {
+      c = c - 150;
+   }
+  char k = (char) c;
+  printf ("k: %d", k);
+  assert (k == -22); //UNKNOWN
```
Wherekonshade, line 13: Missing "!"

EdinCitaku:
Also dont forget to remove Commented out lines before you are finished, there are quite a few so I didnt comment on any specifically

tests/regression/37-congruence/01-simple.c
```diff
@@ -0,0 +1,28 @@
...
+
+		a = b - a;
+		c = e + d;
+	}
+
+	a = d / 2;
+	b = d % 2;
+
+	assert (c == 4); // UNKNOWN!
+	assert (d == 12); // UNKNOWN
```
Wherekonshade, line 23: Missing "!"

MartinWehking: This question came up a few weeks ago.
To our knowledge the difference is that 'UNKNOWN!' expects the variable to be definitely unknown (which means it should not have a precise value like 'c = 4' or in the congruence domain: c = {4} ((4,0) in our implementation). In the example above we figured that there is no way that c ever becomes a concrete val after a fixed-point iteration.
You're right though, we should definitely check the other tests again as our implementations and test cases changed quite frequently.

make.sh
hydrogenoxide: I asked during one meeting if this change was ok. Without it I can't run the scripts on my system. 
@michael-schwarz I could put this change in a separate pull request if you want.

src/cdomains/intDomain.ml
hydrogenoxide: Do you have any specific reasons for that? In other places there are tracing instructions too and it was immensely useful for fixing some problems. If you don't explicitly activate it, it doesn't change the behavior of the program in any way.

EdinCitaku: It is my personal preference, but I understand why one would want to keep it. But in any case I would at least change the flag for tracing from "mything" to something more fitting.

hydrogenoxide: > But in any case I would at least change the flag for tracing from "mything" to something more fitting.
That we'll definitely do!

make.sh
```diff
@@ -73,14 +74,19 @@ rule() {
...
-    #   rm doclist.odocl;
-    #   ln -sf _build/doclist.docdir doc
-    # ;; tag*)
-    #   otags -vi `find src/ -iregex [^.]*\.mli?`
+    ;; docs)
+      rm -rf doc;
+      ls src/**/*.ml | egrep -v $EXCLUDE  | sed 's/.*\/\(.*\)\.ml/\1/' > doclist.odocl;
+      ocb -ocamldoc ocamldoc -docflags -charset,utf-8,-colorize-code,-keep-code doclist.docdir/index.html;
+      rm doclist.odocl;
+      ln -sf _build/doclist.docdir doc
```
sim642, line 82: What's the reason for reintroducing this? It's using the old ocamlbuild instead of dune, which can build documentation with a single command: https://dune.readthedocs.io/en/stable/documentation.html.

make.sh
```diff
@@ -73,14 +74,19 @@ rule() {
...
+    ;; tag*)
+      otags -vi `find src/ -iregex [^.]*\.mli?`
+    ;; poly)
+      echo "open ApronDomain" >> $TARGET.ml
+      echo "open Poly" >> $TARGET.ml
+      ocb -no-plugin -package apron -package apron.polkaMPQ -package apron.octD $TARGET.native &&
+      cp _build/$TARGET.native goblint
```
sim642, line 89: No reason to reintroduce these here.

src/util/defaults.ml
```diff
@@ -119,6 +119,9 @@ let _ = ()
       ; reg Analyses "ana.int.def_exc"      "true"  "Use IntDomain.DefExc: definite value/exclusion set."
       ; reg Analyses "ana.int.interval"    "false" "Use IntDomain.Interval32: (int64 * int64) option."
       ; reg Analyses "ana.int.enums"       "false" "Use IntDomain.Enums: Inclusion/Exclusion sets. Go to top on arithmetic operations (except for some easy cases, e.g. multiplication with 0). Joins on widen, i.e. precise integers as long as not derived from arithmetic expressions."
+      ; reg Analyses "ana.int.congruence"  "false" "Use IntDomain.Congruence: (c, m) option, meaning congruent to c modulo m"
+      ; reg Analyses "ana.int.congruence_no_overflow" "false" "TBD"
+      ; reg Analyses "ana.int.refinement"   "false" "TBD"
```
sim642, line 124: These descriptions are still missing.

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1890,479 @@ module Enums : S with type int_t = BigInt.t = struct
...
+  include Std (struct type nonrec t = t let name = name let top_of = top_of let bot_of = bot_of let show = show let equal = equal end)
+
+  let equal_to i = function
+    | None -> failwith "unsupported: equal_to with bottom"
+    | Some (a, b) when b =: Ints_t.zero -> if a = i then `Eq else `Neq
+    | Some (a, b) -> if i %: b = a then `Top else `Neq
+
+  let set_overflow_flag ik =
+    if Cil.isSigned ik && !GU.in_verifying_stage then
+      Goblintutil.did_overflow := true
```
sim642, line 1973: How is congruence analysis detecting overflow?

src/cdomains/intDomain.mli
```diff
@@ -231,6 +234,13 @@ module type S =
 sig
   include B
   include ArithIkind with type t:= t
+
+  val add : ?no_ov:bool -> Cil.ikind ->  t -> t -> t
+  val sub : ?no_ov:bool -> Cil.ikind ->  t -> t -> t
+  val mul : ?no_ov:bool -> Cil.ikind ->  t -> t -> t
+  val div : ?no_ov:bool -> Cil.ikind ->  t -> t -> t
+  val cast_to : ?torg:Cil.typ -> ?no_ov:bool -> Cil.ikind -> t -> t
```
sim642: What is this `no_ov` supposed to mean?

make.sh
michael-schwarz: Yes, putting these into a separate PR would clarify things here :+1:

src/cdomains/intDomain.ml
```diff
@@ -1853,27 +2395,156 @@ module IntDomTupleImpl = struct
...
+      BatOption.
+        ( map (r.fp2 (module I1)) a
+        , map (r.fp2 (module I2)) b
+        , map (r.fp2 (module I3)) c
+        , map (r.fp2 (module I4)) d )
+
+  let flat f x = match to_list_some x with [] -> None | xs -> Some (f xs)
+
+  let to_excl_list x = mapp2 { fp2 = fun (type a) (module I:S with type t = a and type int_t = int_t) -> I.to_excl_list } x |> flat List.concat
+  let to_incl_list x = mapp2 { fp2 = fun (type a) (module I:S with type t = a and type int_t = int_t) -> I.to_incl_list } x |> flat List.concat
```
jerhard, line 2501: Why are the inclusion lists concatenated? If all the inclusion lists are a sound overapproximation, one could take the meet of the inclusion lists, no?
One would have to be careful to consider that some implementations of `to_incl_list` return `None` even when the abstract value is not bottom, but when it is only impractical to create the  complete inclusion list (e.g. the interval domain does this.)

src/cdomains/intDomain.ml
```diff
@@ -779,6 +804,44 @@ struct
...
+             else if Ints_t.compare rcx lcy = 0 then Some (rcx, rcx)
+             else Some (rcx, lcy)
+    | _ -> intv
+
+  let refine_with_congruence x y =
+    let refn = refine_with_congruence x y in
+    if M.tracing then M.trace "refine" "int_refine_with_congruence %a %a -> %a\n" pretty x pretty y pretty refn;
+    refn
+
+  let refine_with_interval a b = a
```
michael-schwarz, line 833: Not sure how meaningful this is, but one could set this to the meet?

src/cdomains/intDomain.ml
```diff
@@ -779,6 +804,44 @@ struct
...
+
+  let refine_with_excl_list (intv : t) (excl : (int_t list) option) : t =
+    match intv, excl with
+    | None, _ -> intv
+    | _, None -> intv
+    | Some(l, u), Some(ls) ->
+        let l' = (Ints_t.add l (Ints_t.of_int(Bool.to_int(List.mem l ls)))) in
+        let u' = (Ints_t.sub u (Ints_t.of_int(Bool.to_int(List.mem u ls)))) in Some(l', u')
+
+  let refine_with_incl_list a b = a
```
michael-schwarz, line 843: Please add this, it should be fairly straightforward

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
 
+    let refine_with_congruence a b =
+      let contains c m x = if BI.equal m BI.zero then BI.equal c x else (BI.rem (BI.sub x c) m) == BI.zero in
+      match a with
+      | Inc e -> ( match b with
+                | None -> Inc e
+                | Some (c, m) -> Inc (ISet.filter (contains c m) e))
+      | a -> a
+  let refine_with_interval a b = a
+  let refine_with_excl_list a b = a
```
michael-schwarz, line 1912: The inclusion set contained in an enum can be refined by an exclusion set

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+  let (<:) x y = Ints_t.compare x y < 0
+  let (>:) x y = Ints_t.compare x y > 0
+  let (<=:) x y = Ints_t.compare x y <= 0
+  let (>=:) x y = Ints_t.compare x y >= 0
+  (* a divides b *)
+  let ( |: ) a b = (b %: a) =: Ints_t.zero
+
+  let to_yojson t = failwith "to yojson unimplemented"
+
+  let rec gcd x y = (*TODO Double Check*)
```
michael-schwarz, line 1939: Is this double checked?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    | None -> None
+    | Some (c, m) -> if m =: Ints_t.zero then Some (c, m) else Some (c %: (abs m), (abs m))
+
+  let min_int ik = Ints_t.of_bigint @@ fst @@ Size.range_big_int ik
+
+  let max_int ik = Ints_t.of_bigint @@ snd @@ Size.range_big_int ik
+  let top () = Some (Ints_t.zero, Ints_t.one)
+  let top_of ik = Some (Ints_t.zero, Ints_t.one)
+  let bot () = None
+  let bot_of ik = bot () (* TODO: improve  *)
```
michael-schwarz, line 1956: What is there to be improved here?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+
+  let is_top x = x = top ()
+
+  let show = function ik -> match ik with
+    | None -> "∅"
```
michael-schwarz, line 1961: Can we use `⟂` here?

src/cdomains/intDomain.ml
michael-schwarz: We should not set the overflow flag here, setting it leads to reporting an overflow right away, however this is usually quite imprecise in the congruence domain.
If we are more serious about overflow checking, we would need to check for each subexpression if an active domain can exclude the overflow, and set the overflow flag only id no active domain can

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    if c <: min_int ik   || c >: max_int ik || m <: min_int ik || m >: max_int ik then (set_overflow_flag ik; top_of ik)
+    else Some (c,m)
+
+  let leq (x:t) (y:t) =
+    match x, y with
+    | None, _ -> true
+    | Some _, None -> false
+    | Some (c1,m1), Some (c2,m2) when m2 =: Ints_t.zero && m1 =: Ints_t.zero -> c1 =: c2
+    | Some (c1,m1), Some (c2,m2) when m2 =: Ints_t.zero -> c1 =: c2 && m1 =: Ints_t.zero
+    | Some (c1,m1), Some (c2,m2) -> m2 |: (gcd (c1 -: c2) m2)
```
michael-schwarz, line 1995: Does this need to be so complicated?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    let res = leq x y in
+    if M.tracing then M.trace "congruence" "leq %a %a -> %a \n" pretty x pretty y pretty (Some(Ints_t.of_int (Bool.to_int res), Ints_t.zero)) ;
+    res
+
+  let join ik (x:t) y =
+    match x, y with
+    | None, z | z, None -> z
+    | Some (c1,m1), Some (c2,m2) ->
+      let m3 = gcd m1 (gcd m2 (c1 -: c2)) in
+      norm ik @@ normalize (Some (c1, m3))
```
michael-schwarz, line 2007: what is the difference between `norm` and `normalize`. They should be renamed to more obvious names

src/cdomains/intDomain.ml
```diff
@@ -1853,27 +2405,156 @@ module IntDomTupleImpl = struct
...
+  let refine ((a, b, c, d ) : t ) : t =
+    if not (GobConfig.get_bool "ana.int.refinement") then (a, b, c, d )
+    else
+      let dt = ref (a, b, c, d) in
+      dt := refine_with_excl_list !dt (to_excl_list (a, b, c, d));
+      let quit_loop = ref false in
+      while not !quit_loop do
+        let old_dt = !dt in
+        List.iter (fun f -> dt := f !dt) refine_functions;
+        quit_loop := old_dt = !dt;
```
jerhard, line 2521: Use `equal` here instead of the structural equality `=`.

src/cdomains/intDomain.ml
```diff
@@ -1853,27 +2405,156 @@ module IntDomTupleImpl = struct
...
+        if M.tracing then M.trace "cong-refine-loop" "old: %a, new: %a\n" pretty old_dt pretty !dt;
+      done;
```
jerhard, line 2523: You can check in each step whether some of the domains became `bot`, and return the appropriate bottom value of `IntDomTuple` in this case.

Besides that, there should be an option to perform the refinement exactly once (so no further looping). The option ` "ana.int.refinement"` could thus be turned into a string, instead of a boolean, with possible values `"never"`, `"once"`, and `"fixpoint"`.

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    let res = meet ik x y in
+    if M.tracing then M.trace "congruence" "meet %a %a -> %a\n" pretty x pretty y pretty res;
+    res
+
+  let is_int = function Some (c, m) when m =: Ints_t.zero -> true | _ -> false
+
+  (* TODO: change to_int signature so it returns a big_int *)
+  let to_int = function Some (c, m) when m =: Ints_t.zero -> Some c | _ -> None
+  let of_int ik (x: int_t) = Some (x, Ints_t.zero)
+  let of_pair ik p = normalize (Some p)
```
michael-schwarz, line 2045: Maybe this should be closer to where it is used?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+
+  let widen ik x y =
+    let res = widen ik x y in
+    if M.tracing then M.trace "congruence" "widen %a %a -> %a\n" pretty x pretty y pretty res;
+    res
+
+  let narrow = meet
+
+  let log f ik i1 i2 =
+    match is_bot i1, is_bot i2 with
```
michael-schwarz, line 2094: Can we pull this handling of `bot` out so we have it in only one place?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    | _   , true -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show i1) (show i2)))
+    | _ ->
+      match to_int i1, to_int i2 with
+      | Some x, Some y -> (try norm ik (of_int ik (f ik x y)) with Division_by_zero | Invalid_argument _ -> top_of ik)
+      | _              -> (set_overflow_flag ik;  top_of ik)
+
+  let is_power_of_two x = x >: Ints_t.zero && Ints_t.of_int 2 |: x
 
+  (* Not very pretty. A proper log operation might be better*)
+  let rec log2 c k = if float_of_int 2 **  float_of_int k = float_of_int c then k else log2 c (k + 1)
```
michael-schwarz, line 2129: Why are we going via floats here?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+
+  (*let shift_right ik x y = match x, y with
```
michael-schwarz, line 2131: What is the state of this?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+
+  let rem ik x y = let res = rem ik x y in
+    if M.tracing then  M.trace "congruence" "rem : %a %a -> %a \n" pretty x pretty y pretty res;
+    res
+
+  let div ?(no_ov=false) ik x y =
+    match x,y with
+    | None, None -> bot ()
+    | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
+    | _, x when leq zero x -> top () (*TODO warn about undefined behaviour when x equals zero*)
```
michael-schwarz, line 2255: At least warn if we actually have must ub

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+  let div ?no_ov ik x y =
+    let res = div ?no_ov ik x y in
+    if M.tracing then
+      M.trace "congruence" "div : %a %a -> %a \n" pretty x pretty y pretty
+        res ;
+    res
+
+  let ne ik i1 i2 = if meet ik i1 i2 = None then of_bool ik true else top_bool
+
+  let eq ik (i1: t) (i2: t) = if meet ik i1 i2 <> None then top_bool else of_bool ik false
```
michael-schwarz, line 2271: What about the case where they must be equal?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
+    | Some (c1,m1), Some (c2,m2) -> if (m1 =: Ints_t.zero) && (m2 =: Ints_t.zero) then
+        if c1 <: c2 then of_bool ik true else of_bool ik false
+      else top_bool
+
+  let lt ik x y =
+    let res = lt ik x y in
+    if M.tracing then  M.trace "congruence" "less than : %a %a -> %a \n" pretty x pretty y pretty res;
+    res
+
```
michael-schwarz, line 2320: There seems to be a lot of duplication here, can we pull most of this into a helper function maye?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+       let open Cil in
+       let (c, m) = BatTuple.Tuple2.mapn (fun a -> kintegerCilint ik @@ Cilint.Big (Ints_t.to_bigint a)) (c, m) in
+       (try
+          Invariant.of_exp (BinOp (Eq, (BinOp (Mod, l, m, TInt(ik,[]))), c, TInt(ik,[])))
+        with e -> None)
+    | None -> None
+
+  let arbitrary () =
+    let open QCheck in
+    let ik = Cil.ILongLong in
```
michael-schwarz, line 2337: Why always `I.LongLong`? Maybe it is better to use an `arbitrary` `ikind`.

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1901,483 @@ module Enums : S with type int_t = BigInt.t = struct
...
+  let refine_with_congruence a b = a
+  let refine_with_excl_list a b = a
+  let refine_with_incl_list a b = a
 end
 
+module Congruence = CongruenceFunctor (BI)
+
+(* module Congruence32 =
+ *   IntDomWithDefaultIkind
+ *     (IntDomLifter (CongruenceFunctor (IntOps.Int64Ops))) (IntIkind) *)
```
michael-schwarz, line 2380: What is this for?

src/cdomains/intDomain.ml
```diff
@@ -1853,27 +2405,156 @@ module IntDomTupleImpl = struct
...
     let f n g = if get_bool ("ana.int."^n) then Some (g x) else None in
-    f "def_exc" @@ r.fi2 (module I1), f "interval" @@ r.fi2 (module I2), f "enums" @@ r.fi2 (module I3)
+    f "def_exc" @@ r.fi2 (module I1), f "interval" @@ r.fi2 (module I2), f "enums" @@ r.fi2 (module I3), f "congruence" @@ r.fi2 (module I4)
+
+  let no_overflow ik r =
+    if GobConfig.get_bool "ana.int.congruence_no_overflow" then true
+    else let ika, ikb = Size.range_big_int ik in
+      match I2.minimal r, I2.maximal r with
+        | Some ra, Some rb -> ika < ra || rb < ikb
+        | _ -> false
```
michael-schwarz, line 2423: Can you explain what exactly this is supposed to do?

tests/regression/37-congruence/01-simple.c
```diff
@@ -0,0 +1,34 @@
+//PARAM: --enable ana.int.congruence  --enable ana.int.congruence_no_overflow --disable ana.int.def_exc --disable ana.int.enums
```
michael-schwarz, line 1: Please add a comment to make it clear that this example comes straight from Granger's paper, in order to give proper credit to him.

tests/regression/37-congruence/01-simple.c
```diff
@@ -0,0 +1,34 @@
...
+
+		a = b - a;
+		c = e + d;
+	}
+
+	a = d / 2;
+	b = d % 2;
+
+    // c is unknown
+	assert (c == 4); // UNKNOWN!
```
michael-schwarz, line 23: `//UNKNOWN!` has the following meaning:
- The analysis claiming that this must fail is unsound
- The analysis claiming that this must succeed is unsound

Or to phrase it differently, this means that **in the concrete** there are both executions where this holds and where it doesn't. In this case, `//UNKNOWN` is the correct annotation, as this holds in the concrete and it just the analysis that can not show it.

michael-schwarz: I think in the remaining tests, you used it correctly.

tests/regression/38-int-refinements/01-interval-congruence.c
```diff
@@ -0,0 +1,26 @@
...
+        r = 7;
+    }
+
+    // At this point r in the congr. dom should be 2 + 5Z
+    int k = r;
+    if (k >= 3) {
+
+        // After refinement with congruences, the lower bound should be 7 as the numbers 3 - 6 are not in the congr. class
+        // TODO fix refinement and uncomment the following line
+        assert (k < 7); // FAIL
```
michael-schwarz, line 19: Is `//FAIL` the right annotation here?

tests/regression/38-int-refinements/01-interval-congruence.c
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+//    if (r >= -11 && r <= -4) {
+//        assert (r == -8);
+//    }
```
jerhard, line 24: What's up with this commented-out code?

src/cdomains/intDomain.ml
michael-schwarz: This might be wrong! Consider e.g. 5 mod 13 \leq 5 mod 7

tests/regression/37-congruence/03-interval-overflow.c
```diff
@@ -0,0 +1,44 @@
...
+    r = -68;
+  } else {
+    r = -63;
+  }
+
+  char k = r - 80;
+  assert (k == 0); //UNKNOWN!
+
+  char non_ov = r - 10;
+  assert (non_ov == -78); //UNKNOWN
```
michael-schwarz, line 19: Is `UNKNOWN` the right annotation here?

src/cdomains/intDomain.ml
```diff
@@ -128,6 +138,11 @@ sig
   val of_interval: Cil.ikind -> int_t * int_t -> t
   val is_top_of: Cil.ikind -> t -> bool
   val invariant_ikind : Invariant.context -> Cil.ikind -> t -> Invariant.t
+
+  val refine_with_congruence: Cil.ikind -> t -> (int_t * int_t) option -> t
+  val refine_with_interval: Cil.ikind -> t -> (int_t * int_t) option -> t
```
michael-schwarz, line 143: Can we use `Interval.t` and `Congruences.t` here instead of `(int_t * int_t) option`? Or does that lead to issues with cyclic builds?

src/cdomains/intDomain.ml
```diff
@@ -1826,9 +1925,467 @@ module Enums : S with type int_t = BigInt.t = struct
...
+  | _, None -> a
+  | a , Some (ls) -> meet ik a (of_excl_list ik ls)
+  let refine_with_incl_list ik a b = match a, b with
+  | _, None -> a
+  | Inc x, Some (ls) -> meet ik (Inc x) (Inc (ISet.of_list ls))
+  | a, _ -> a
 
 end
 
+module CongruenceFunctor(Ints_t : IntOps.IntOps): S with type int_t = Ints_t.t and type t = (Ints_t.t * Ints_t.t) option =
```
michael-schwarz, line 1950: I guess this doesn't need to be parametric in `Ints_t`, we could just use `BigIntOps` here.

tests/regression/37-congruence/03-interval-overflow.c
```diff
@@ -0,0 +1,44 @@
+// PARAM: --enable ana.int.interval --enable ana.int.congruence --disable ana.int.def_exc
+// Overflow information should be passed from the interval domain to congruences
+#include <assert.h>
+#include <stdio.h>
+
+int main(){
+  char r;
```
michael-schwarz, line 7: Maybe switch to `signed char` in this example, it is not really clear from the C standard if char should be signed or not (it is signed for GCC).

src/cdomains/intDomain.ml
hydrogenoxide: I wanted to do that at first. This is in the definition of type S. At this point the module Congruences is not yet defined. But we can't put its definition before that as Congruences has type S. Same situation with Interval.

tests/regression/37-congruence/03-interval-overflow.c
MartinWehking: No, it should have been UNKNOWN! due to the branching above. I fixed it

src/cdomains/intDomain.ml
```diff
@@ -779,6 +804,69 @@ struct
...
+    | _ -> intv
+
+  let refine_with_congruence ik x y =
+    let refn = refine_with_congruence ik x y in
+    if M.tracing then M.trace "refine" "int_refine_with_congruence %a %a -> %a\n" pretty x pretty y pretty refn;
+    refn
+
+  let refine_with_interval ik a b =
+    match b with
+    | None -> a
```
michael-schwarz, line 839: Why is this `a` if `b` is `bot`? Do we have the same problem as with the lists here?

src/cdomains/intDomain.ml
hydrogenoxide: This is again so because we initially copied the code from the interval domain

make.sh
michael-schwarz: Pulled out into #276

src/cdomains/intDomain.ml
MartinWehking: It would make more sense to return `bot` instead. The reason why I put `a` there was because I incorrectly assumed that `refine_with_interval `and `refine_with_congruence` would also be called if interval/congruence analysis were disabled and parameter b would therefore become `bot`. Instead this is filtered by the `maybe` function. I checked the code again quickly and saw that this problem occurs in a few domains. I guess we can simply return `bot` everywhere where `refine_with_interval` /`refine_with_congruence` are called with param b = `bot`.

src/cdomains/intDomain.ml
```diff
@@ -1887,15 +2444,198 @@ module IntDomTupleImpl = struct
...
+           quit_loop := equal old_dt !dt;
+           if is_bot !dt then dt := bot_of ik; quit_loop := true;
+           if M.tracing then M.trace "cong-refine-loop" "old: %a, new: %a\n" pretty old_dt pretty !dt;
+         done;
+      | _ -> ()
+    in
+    !dt
+
+  let no_overflow ik r =
+    if GobConfig.get_bool "ana.int.congruence_no_overflow" then true
```
michael-schwarz, line 2578: `ana.int.congruence_no_overflow` should only apply to **signed** overflows, overflows fro variables of unsigned integer types are well-defined and may exist in programs without Undefined Behavior.

michael-schwarz:
With `congruences` on, refinement on `fixpoint` and `ana.int.congruence_no_overflow` we are currently unsound for the following SV-COMP benchmarks. I'll fix the issue with the flag affecting the behavior for unsigned types too, this might fix some of these issues.

- [] bitvector/s3_clnt_3.BV.c.cil-2a.yml
- [] bitvector-regression/implicitunsignedconversion-1.yml
- [] bitvector-regression/integerpromotion-3.yml
- [] bitvector-regression/signextension-1.yml
- [] bitvector-regression/signextension2-2.yml
- [] seq-mthreaded/pals_STARTPALS_Triplicated.1.ufo.BOUNDED-10.pals.yml
- [] seq-mthreaded/pals_STARTPALS_Triplicated.1.ufo.UNBOUNDED.pals.yml
- [] seq-mthreaded/pals_STARTPALS_Triplicated.2.ufo.BOUNDED-10.pals.yml
- [] seq-mthreaded/pals_STARTPALS_Triplicated.2.ufo.UNBOUNDED.pals.yml
- [] pthread/queue.yml
- [] pthread/queue_longer.yml
- [] pthread/queue_longest.yml
- [] ldv-linux-3.14-races/linux-3.14--drivers--spi--spi-tegra20-slink.ko.cil.yml
- [] aws-c-common/aws_byte_buf_write_from_whole_buffer_harness-1.yml
- [] dv-linux-3.12-rc1/linux-3.12-rc1.tar.xz-144_2a-drivers--input--touchscreen--usbtouchscreen.ko-entry_point.cil.out.yml

michael-schwarz:
The `bitvector` ones seem deeper than an issue with no overflow.
If `def_exc` is enabled in addition, at the branch we fail because of a call to an `assert false` in the branching function. This also hold with refinement disabled.

https://github.com/nicokn/analyzer/blob/dbc62f00421915cc1934369dfca4b0708b523ae5/src/analyses/base.ml#L1648-L1652

src/cdomains/intDomain.ml
hydrogenoxide: I think the reason why I blankly applied it to all integers is that we lack a coherent theory of how the congruence classes change when overflowing.
For example (0 + 5ℤ) +(1 + 5ℤ) = (1 + 5ℤ). But, assuming we're working with 4 bit to represent unsigned integers, 15 + 1 = 0 ∈ (0 + 5ℤ). So unless we're working with singletons, I think it's still necessary to know if an overflow might occur.

michael-schwarz: >  So unless we're working with singletons, I think it's still necessary to know if an overflow might occur.

Yes, I absolutely agree that we need to know if an overflow can occur. However, currently if `ana.int.congruence_no_overflow` is enabled, it just claims that there are NEVER any overflows. This is justified for signed types (as overflows there are undefined behavior, and one may want to assume that programs are free of that).

However, unsigned types have well-defined overflows, so one can not just claim that they never overflow. Instead for unsigned types - even if I assume no UB - I need to use the logic to check for overflows via Intervals that is also used when `ana.int.congruence_no_overflow` is disabled.

Or am I misunderstanding your point?

hydrogenoxide:
> **So it seems that the cast operation is not handled properly in congruences.**

Regression test 21 01 might be of interest too. I noticed that some assertions at the end are suddenly unknown when enabling congruences. The first cast `b = (schar) a` already doesn't seem to be handled by the `cast_to'` in congruences (defexc says it's -128 but in the congruence domain it's top).

Concerning the implementation of `cast_to`:
Again I used the implementation of other domains as a reference. But I didn't include the case that's commented with `(* same static type -> no overflows for r, but we need to cast s since it may be out of range after lift2_inj *)` (e.g., for defexc, line 1187) as I didn't completely understand it and assumed that returning top in such unhandled cases is still sound.

src/cdomains/intDomain.ml
hydrogenoxide: Ah, then I misunderstood the semantics the flag should have. I thought of it as ignoring any and all integer over- or underflows.

Of course if this should only ignore undefined overflows, this completely makes sense.

michael-schwarz:
>  as I didn't completely understand it and assumed that returning top in such unhandled cases is still sound

Yes, returning top in unhandled cases is always sound. The problem in the other example is that congruences unsoundly claims that `signedtosigned → congruences:65535`. If it were top, that would be sound (even if maybe unnecessarily imprecise)

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1925,467 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    | Some (x, y) when y =: Ints_t.zero -> Some x
+    | _ -> None
+
+  let minimal t = match t with
+    | Some (x,y) when y =: Ints_t.zero -> Some x
+    | _ -> None
+
+  (* cast from original type to ikind, set to top if the value doesn't fit into the new type *)
+  let cast_to ?torg ?(no_ov=false) t x =
+    if no_ov then x
```
michael-schwarz, line 2096: `no_ov` does not imply that the cast did not affect the value.

~~~C
    int i = -1; // Interval [-1.-1]
    unsigned int j = i; // Interval [4294967295,4294967295]
~~~

Here the `no_ov` flag would be set as the value int the interval domain is not top.
However, the value would still need to be changed

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
+        pretty res ;
+    res
+
+  let bitnot ik x = match x with
+    | None -> None
+    | Some (c, m) -> if (Cil.isSigned ik) then sub ik (neg ik x) one else Some (Ints_t.sub (max_int ik) c, m)
+
+  (** The implementation of the bit operations is based on the master’s thesis
+      'Abstract Interpretation and Abstract Domains' written by Stefan Bygde.
+      see: https://www.dsi.unive.it/~avp/domains.pdf *)
```
michael-schwarz, line 2241: This is missing the weakening operation ξ from the paper, and hence the
`Ints_t.shift_left Ints_t.one (Ints_t.to_int (min m m')))` can be by ridiculous amounts which cause the program to run out of memory.

michael-schwarz:
After removing the bit operations for now, the changes for running SV-COMP look reasonable (if a little underwhelming):
[292b0c7.zip](https://github.com/goblint/analyzer/files/6808706/292b0c7.zip)

Adding bit operations back is a good idea, but I would propose doing that in a separate PR and merging this now.

Anything we should still change before merging this PR @sim642 @vogler @jerhard ?

sim642:
I quickly looked at the results diff and there are a bunch of cases where the runtime of an unknown benchmark goes up by 2..4 times (at least) to TIMEOUT. For example:
* seq-mthreaded-reduced/pals_lcr-var-start-time.3.2.ufo.BOUNDED-6.pals.c.v+sep-reducer
* ldv-linux-3.4-simple/32_1_cilled_ok_nondet_linux-3.4-32_1-drivers--media--common--tuners--qt1010.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out

Seems a bit suspicious to increase this much.

vogler:
What's the summary of these results?
It's more precise for 2 out of 1420 tests and it ran 182000s-174000s = 2.22h or 4.6% longer?
Why is there 8648 total instead of 8649?

michael-schwarz:
> It's more precise for 2 out of 1420 tests 
In aggregate, yes. If one looks at `results.2021-07-13_13-56-38.diff.html` the results are as follows:

- 7 correct answers in the old config turn into timeouts with a 60s timeout (3 out of which would succeed with a 61s timeout)
- 9 tests succeed in the new config.

> and it ran 182000s-174000s = 2.22h or 4.6% longer?

Yes, but it is comparing different configurations. With congruences and refinement off, I would not expect a measurable difference. An increase in runtime when enabling refinement and an additional integer domain is to be expected imo.

> Why is there 8648 total instead of 8649?

Benchexec has a tendency to hang on the last task, so I aborted that one manually. I don't think that this one is makes the difference either way


> Seems a bit suspicious to increase this much.

I'm not sure, I am not so surprised by this result. 
- It now performs refinement on the int domain until a fixpoint is reached. This is the most expensive setting.
- We have a new intDomain enabled.
- Congruences can act as a set of two definite values, which might drive up the numbers of contexts.

michael-schwarz:
The runs with the same seed succeed on the branch before merging (and the runs with other seeds succeed here too https://github.com/nicokn/analyzer), so there seems to be some interaction with changes to the master that happened after this diverged. :thinking:

sim642:
>    * It now performs refinement on the int domain until a fixpoint is reached. This is the most expensive setting.

This is what I was the most worried about: this int refinement fixpoint not terminating or taking a large number of iterations. Or are there some theoretical guarantees with the int domains that we have, that their mutual refinement terminates quickly enough?

michael-schwarz:
> Or are there some theoretical guarantees with the int domains that we have, that their mutual refinement terminates quickly enough?

Yes, this is indeed a fascinating question. Off the top of my head, I don't know that we should have such guarantees. While one might be able to artificially construct abstract values for which such a refinement takes a long time, I am a bit skeptical that we would encounter it in real world examples. We should thoroughly consider this again if we ever want to enable refinement by default.

michael-schwarz:
> I quickly looked at the results diff and there are a bunch of cases where the runtime of an unknown benchmark goes up by 2..4 times (at least) to TIMEOUT. For example:

> seq-mthreaded-reduced/pals_lcr-var-start-time.3.2.ufo.BOUNDED-6.pals.c.v+sep-reducer
> ldv-linux-3.4-simple/32_1_cilled_ok_nondet_linux-3.4-32_1-drivers--media--common--tuners--qt1010.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out

I tried to get an understanding for them, but these tests are highly complicated and in the case of the `-reduced` are somehow automatically extracted from other benchmarks by CPAChecker, so I am not sure if they are meant to be understood.

If there was really an explosion of runtimes, I would expect to at least see some indication of this also for benchmarks where we succeed. For these, the runtime is up by 14% (6220s vs 5420s).

Re: The not too impressive gain in precision: Currently there is no refinement on branching that involves a modulo operation. #268. I would expect this to maybe make this domain pay off more. Also, one could re-add bit operations in a sound manner.

My proposal would be merging and leaving the default settings as they are and then improving incrementally via PRs that are so small they are easier to review.

sim642:
> If there was really an explosion of runtimes, I would expect to at least see some indication of this also for benchmarks where we succeed. For these, the runtime is up by 14% (6220s vs 5420s).

Could be that those handful of benchmarks contain some particularly complex kind of expressions that involve some expensive congruence calculation? I'm not familiar with the algorithms for this, but could be something like large primes or large numbers with many divisors, which don't occur in most benchmarks.

michael-schwarz:
> Could be that those handful of benchmarks contain some particularly complex kind of expressions that involve some expensive congruence calculation?

Not that I can tell, but given that they are several thousands of lines, it is really hard to say what is going on. There seems to be tons of branching and tons of constant values and also ungodly amounts of variables, but apart from that is very hard to say.

conf/svcomp21.json
```diff
@@ -7,7 +7,10 @@
     "int": {
       "def_exc": true,
       "enums": false,
-      "interval": true
+      "interval": true,
+      "congruence": false,
+      "congruence_no_overflow": false,
+      "refinement": "never"
```
sim642, line 13: Is there a reason to change our svcomp21 conf after the fact, even though these are the defaults anyway?

src/cdomains/intDomain.ml
```diff
@@ -779,6 +804,66 @@ struct
...
+  let modulo n k =
+    let result = Ints_t.rem n k in
+    if Ints_t.compare result Ints_t.zero >= 0 then result
+    else Ints_t.add result  k
+
+  let refine_with_congruence ik (intv : t) (cong : (int_t * int_t ) option) : t =
+    match intv, cong with
+    | Some (x, y), Some (c, m) ->
+        if (m = Ints_t.zero) && (Ints_t.compare c x < 0 || Ints_t.compare c y > 0) then None
+        else if m = Ints_t.zero then Some (c, c)
```
sim642, line 820: Should probably use `Ints_t.equal` instead of `=` since other comparisons are delegated.

src/cdomains/intDomain.ml
```diff
@@ -779,6 +804,66 @@ struct
     in
     QCheck.(set_shrink shrink @@ set_print show @@ map (*~rev:BatOption.get*) (of_interval ik) pair_arb)
   let relift x = x
+
+  let abs x = let r = Ints_t.neg x
+   in if Ints_t.compare x r < 0 then r else x
```
sim642, line 809: Is it just me or is this implementation of `abs` a bit weird? The following seems more natural and common:
```ocaml
let abs x =
  if Ints_t.compare x Ints_t.zero < 0 then
    Ints_t.neg x
  else
    x
```

src/cdomains/intDomain.ml
```diff
@@ -779,6 +804,66 @@ struct
...
+        if (m = Ints_t.zero) && (Ints_t.compare c x < 0 || Ints_t.compare c y > 0) then None
+        else if m = Ints_t.zero then Some (c, c)
+        else let rcx =
+             if Ints_t.equal x (min_int ik) then x else
+             Ints_t.add x (modulo (Ints_t.sub c x) (abs(m))) in
+             let lcy =
+              if Ints_t.equal y (max_int ik) then y else
+              Ints_t.sub y (modulo (Ints_t.sub y c) (abs(m))) in
+             if Ints_t.compare rcx lcy > 0 then None
+             else if Ints_t.compare rcx lcy = 0 then norm ik @@ Some (rcx, rcx)
```
sim642, line 828:
```suggestion
             else if Ints_t.equal rcx lcy then norm ik @@ Some (rcx, rcx)
```

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
+                | Some (c, m) -> Inc (ISet.filter (contains c m) e))
+      | a -> a
+  let refine_with_interval ik a b = a
+  let refine_with_excl_list ik a b = match a, b with
+  | _, None -> a
+  | a , Some (ls) -> meet ik a (of_excl_list ik ls)
+  let refine_with_incl_list ik a b = match a, b with
+  | _, None -> a
+  | Inc x, Some (ls) -> meet ik (Inc x) (Inc (ISet.of_list ls))
+  | a, _ -> a
```
sim642, line 1936: Indentation is a bit messed up here.

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
 end
 
+module Congruence : S with type int_t = BI.t and type t = (BI.t * BI.t) option =
+struct
+  let name () = "congruences"
+  module Ints_t = BI
+  type int_t = Ints_t.t
+
+  (* represents congruence class of c mod m *)
+  type t = (Ints_t.t * Ints_t.t) option
```
sim642, line 1947: Maybe also comment what `None` represents then.

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
+  let (=:) = Ints_t.equal
+  let (<:) x y = Ints_t.compare x y < 0
+  let (>:) x y = Ints_t.compare x y > 0
+  let (<=:) x y = Ints_t.compare x y <= 0
+  let (>=:) x y = Ints_t.compare x y >= 0
+  (* a divides b *)
+  let ( |: ) a b =
+    if a =: Ints_t.zero then false else (b %: a) =: Ints_t.zero
+
+  let to_yojson t = failwith "to yojson unimplemented"
```
sim642, line 1963: This is unnecessary if `include Std` follows anyway, no?

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
+
+  let rec gcd x y =
+    if y =: Ints_t.zero then x else gcd y (x %: y)
+
+  let abs x = let r = Ints_t.neg x
+    in if x <: r then r else x
```
sim642, line 1969: Same thing about this `abs` implementation.

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    | None, None -> true
+    | Some (a, b), Some (c, d) -> a =: c && b =: d
+    | _, _ -> false
+
+  include Std (struct type nonrec t = t let name = name let top_of = top_of let bot_of = bot_of let show = show let equal = equal end)
+
+  let equal_to i = function
+    | None -> failwith "unsupported: equal_to with bottom"
+    | Some (a, b) when b =: Ints_t.zero -> if a = i then `Eq else `Neq
+    | Some (a, b) -> if i %: b = a then `Top else `Neq
```
sim642, line 2005: Polymorphic `=`.

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    | Some (c, m) when m =: Ints_t.zero ->
+      let c' = Ints_t.of_bigint @@ BigInt.cast_to t (Ints_t.to_bigint c) in
+      (* When casting into a signed type and the result does not fit, the behavior is implementation-defined. (C90 6.2.1.2, C99 and C11 6.3.1.3) *)
+      (* We go with GCC behavior here: *)
+      (*  For conversion to a type of width N, the value is reduced modulo 2^N to be within range of the type; no signal is raised. *)
+      (*   (https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html)   *)
+      (* Clang behaves the same but they never document that anywhere *)
+      Some (c', m)
+    | _ -> match torg with
+      | (Some (Cil.TInt (ikorg, _)) ) when ikorg = t || (max_int t >= max_int ikorg && min_int t <= min_int ikorg) -> if M.tracing then M.trace "cong-cast" "some case"; x
```
sim642, line 2102: Polymorphic comparisons on `max_int` and `min_int`.

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
+     res
+
+  let shift_left ik x y =
+    (* Naive primality test *)
+  let is_prime n =
+    let n = Ints_t.to_int (abs n) in
+    let rec is_not_divisor d =
+      d * d > n || (n mod d <> 0 && is_not_divisor (d + 1)) in
+    n <> 1 && is_not_divisor 2
+    in
```
sim642, line 2168: Messed up indentation.

Also, why is this all done `int`? Does it not fail with large values then?

src/cdomains/intDomain.ml
```diff
@@ -1825,9 +1919,468 @@ module Enums : S with type int_t = BigInt.t = struct
...
+    if M.tracing then  M.trace "congruence" "rem : %a %a -> %a \n" pretty x pretty y pretty res;
+    res
+
+  let div ?(no_ov=false) ik x y =
+    match x,y with
+    | None, None -> bot ()
+    | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
+    | _, x when leq zero x -> top ()
+    | Some(c1, m1), Some(c2, m2) when not no_ov && m2 = Ints_t.zero && c2 = Ints_t.neg Ints_t.one -> top ()
+    | Some(c1, m1), Some(c2, m2) when m1 =: Ints_t.zero && m2 = Ints_t.zero -> Some(c1 /: c2, Ints_t.zero)
```
sim642, line 2273: Couple of polymorphic `=`.

src/cdomains/intDomain.ml
```diff
@@ -1887,15 +2439,198 @@ module IntDomTupleImpl = struct
...
+      match dom with Some y -> reffun ik domtup y | _ -> domtup
+    in
+    [(fun (a, b, c, d) -> refine_with_excl_list ik (a, b, c, d) (to_excl_list (a, b, c, d)));
+     (fun (a, b, c, d) -> refine_with_incl_list ik (a, b, c, d) (to_incl_list (a, b, c, d)));
+     (fun (a, b, c, d) -> maybe refine_with_interval ik (a, b, c, d) b);
+     (fun (a, b, c, d) -> maybe refine_with_congruence ik (a, b, c, d) d)]
+
+  let refine ik ((a, b, c, d ) : t ) : t =
+    let dt = ref (a, b, c, d) in
+    let _ = match GobConfig.get_string "ana.int.refinement" with
```
sim642, line 2555: Unnecessary `let _ = `.

src/cdomains/intDomain.ml
```diff
@@ -1887,15 +2439,198 @@ module IntDomTupleImpl = struct
...
+         done;
+      | _ -> ()
+    in
+    !dt
+
+  let no_overflow ik r =
+    if GobConfig.get_bool "ana.int.congruence_no_overflow" && Cil.isSigned ik then true
+    else let ika, ikb = Size.range_big_int ik in
+         match I2.minimal r, I2.maximal r with
+         | Some ra, Some rb -> ika < ra || rb < ikb
```
sim642, line 2576: Polymorphic comparisons?

src/cdomains/intDomain.mli
```diff
@@ -196,6 +196,9 @@ sig
   val is_excl_list: t -> bool
   (* Checks if the element is an exclusion set. *)
 
+  val to_incl_list: t -> int_t list option
+  (* Gives a list representation of the included values if possible. *)
```
sim642, line 200: This (and actually some of the comments above) aren't actually documentation comments which should start with `(**`.

src/cdomains/intDomain.ml
michael-schwarz: Yeah, I stumbled on the same thing, I think it somehow saves one operation in the true case but lets' use your suggestion that seems simpler

src/cdomains/intDomain.ml
michael-schwarz: Switched to `Ints_t` now and exposed the end recursion more clearly in the hope the compiler takes the hint

michael-schwarz: I annotated it as such. The short circuit version is also tail-recursive, so I switched back to that.

sim642:
After merging #286 into master, 37/05 fails there on two `UNKNOWN`s. If were made more precise at some point than that test expects, then that test should be updated to reflect that. Otherwise if they're supposed to be actually unknown, they should be changed to `UNKNOWN!` and the unsoundness should be fixed.

michael-schwarz:
> If were made more precise at some point than that test expects, then that test should be updated to reflect that.

Goblint indeed is more precise than expected, I removed the annotation.

PR #261, opened 2021-06-18
Improve to_yojson output

keremc:
This PR makes the output of `to_yojson` a bit easier to read. For variants, I use Yojson's `Variant` - an extension to JSON - and encode these as lists when printing. This way, the output is still valid JSON. Unfortunately, unless I've missed something, variant serializers have to be handwritten in an error-prone manner.


vogler:
>  Unfortunately, unless I've missed something, variant serializers have to be handwritten in an error-prone manner.

What's the problem with deriving variants?
https://github.com/ocaml-ppx/ppx_deriving_yojson#semantics

keremc:
@vogler The derived functions use lists to encode variants. So, it is not possible to distinguish an actual list of 2 strings from an encoded variant. The deserializer can figure it out because it knows what types to expect, but I don't have any context.

sim642:
https://github.com/janestreet/ppx_yojson_conv handles variants the same way (doesn't use `Variant`) by the way. If we actually wanted `Variant`s to be derived, then we'd have to fork one of the derivers and make a minor modification, but it's probably not worth the hassle for now.

vogler:
I'm also missing context 😃
You can't use the generated deserializer? For what?

keremc:
@vogler My use case is this: I unmarshal the result table produced by Goblint and wish to display them in a readable format.

However, for an object like `Lifted "x"`, the default implementation of `to_yojson` generates something like `List [ "Lifted"; "x" ]` which is indistinguishable from the serialization of an actual list `[ "Lifted"; "x" ]`. This ends up being an issue because I don't know whether I should display this as a list or not. I've had a few ideas like treating lists that have "Lifted," "Top" or "Bot" as the first element as special cases but such heuristics can easily fail (e.g., when handling a list of `varinfo`s where the first variable has a "special" name).

As for the generated deserializers, I was just mentioning that this doesn't affect them because they always know the structure of the type they want to deserialize. Unfortunately, in my use case, I don't.

@sim642 This is something I've considered as well and want to do at some point. I think a fork is not even necessary because this might be useful for others as well. They would probably accept a PR.

sim642:
> @sim642 This is something I've considered as well and want to do at some point. I think a fork is not even necessary because this might be useful for others as well. They would probably accept a PR.

Probably not. They're non-standard JSON constructs, so making the deriver suddenly emit them would break so much code and fixing it to get previously well-established behavior would require everyone to implement the recursive conversion function to replace them with normal JSON constructs. This is an additional recursive pass over the constructed JSON response, which some projects are likely against for performance reasons.

The fact that Yojson has the non-standard constructors is some historical relic and they plan on eventually removing them altogether: https://github.com/ocaml-community/yojson/issues/104. This makes it even less likely that such PR is accepted into either deriver. The development on Yojson is very slow, so such removal is too far in the future for us to have to worry about for now though.

src/cdomains/valueDomain.ml
```diff
@@ -992,6 +992,17 @@ struct
...
+    | `Int n -> `Variant ("Int", Some (ID.to_yojson n))
+    | `Address n -> `Variant ("Address", Some (AD.to_yojson n))
+    | `Struct n -> `Variant ("Struct", Some (Structs.to_yojson n))
+    | `Union n -> `Variant ("Union", Some (Unions.to_yojson n))
+    | `Array n -> `Variant ("Array", Some (CArrays.to_yojson n))
+    | `Blob n -> `Variant ("Blob", Some (Blobs.to_yojson n))
+    | `List n -> `Variant ("List", Some (Lists.to_yojson n))
+    | `Bot -> `Variant ("Bot", None)
+    | `Top -> `Variant ("Top", None)
+
```
sim642, line 1005: Based on this, gobview will show all abstract values in base's CPA tagged with their variant name now? I'm just wondering because the currently g2html, which uses the above `printXml` just directly serializes the internal representation and doesn't add a tag to the viewable output. Does this not make the output too verbose?

src/domains/printable.ml
```diff
@@ -87,13 +87,14 @@ struct
...
   let printXml f x = BatPrintf.fprintf f "<value>\n<data>\n%s\n</data>\n</value>\n" (XmlUtil.escape (P.show x))
+  let to_yojson x = `String (P.show x)
 end
 
 
 module type Name = sig val name: string end
 module UnitConf (N: Name) =
 struct
-  type t = unit [@@deriving eq, ord, yojson]
+  type t = unit [@to_yojson fun () -> `String N.name] [@@deriving eq, ord, to_yojson]
```
sim642, line 97: Better just make this a normal `let to_yojson` because it practically is manually defining it anyway.

src/domains/printable.ml
```diff
@@ -433,6 +453,8 @@ struct
     BatPrintf.fprintf f "<value>\n<map>\n";
     loop 0 xs;
     BatPrintf.fprintf f "</map>\n</value>\n"
+
+  let to_yojson xs = `List (BatList.map Base.to_yojson xs)
```
sim642, line 457: Is this already not the default derived implementation for lists?

src/domains/printable.ml
```diff
@@ -442,7 +464,7 @@ end
 
 module Chain (P: ChainParams): S with type t = int =
 struct
-  type t = int [@@deriving eq, ord, yojson]
+  type t = int [@to_yojson fun x -> `String (P.names x)] [@@deriving eq, ord, to_yojson]
```
sim642, line 467: Same here, normal `let to_yojson` would be more readable.

src/domains/printable.ml
```diff
@@ -530,7 +560,7 @@ end
 
 module Strings =
 struct
-  type t = string [@@deriving eq, to_yojson]
+  type t = string [@@deriving eq, yojson]
```
sim642, line 563: Why does this need to also derive `of_yojson`?

src/framework/analyses.ml
```diff
@@ -183,8 +183,16 @@ struct
       | MyCFG.Function g      -> BatPrintf.fprintf f "ret%d" g.vid
       | MyCFG.FunctionEntry g -> BatPrintf.fprintf f "fun%d" g.vid
     in
+    (* Remove Yojson's non-standard extensions to JSON *)
+    let rec transform = function
+      | `Assoc l -> `Assoc (BatList.map (fun (n, v) -> (n, transform v)) l)
+      | ( `List l | `Tuple l ) -> `List (BatList.map transform l)
+      | `Variant (n, None) -> `List [ `String n ]
+      | `Variant (n, Some v) -> `List [ `String n; transform v ]
+      | _ as x -> x
```
sim642, line 192: Just `| x -> x` should do.

src/framework/analyses.ml
```diff
@@ -288,7 +296,7 @@ struct
         | MyCFG.Function g      -> fprintf f "\"ret%d\"" g.vid
         | MyCFG.FunctionEntry g -> fprintf f "\"fun%d\"" g.vid
       in
-      let p_fun f x = fprintf f "{\n  \"name: \"%s\",\n  \"nodes\": %a\n}" x (p_list p_node) (SH.find_all funs2node x) in
+      let p_fun f x = fprintf f "{\n  \"name\": \"%s\",\n  \"nodes\": %a\n}" x (p_list p_node) (SH.find_all funs2node x) in
```
sim642, line 299: Nice catch!

src/cdomains/valueDomain.ml
keremc: I use the variant names for additional formatting and skip over them otherwise. So, they only make the JSON output more verbose. However, I think the only variant names that I treat differently are `Bot`, `Top`, and `Set`, the last of which was used in the set domain and doesn't seem to exist anymore. So, I guess I might just remove most of these `Variant`s.

src/domains/printable.ml
keremc: You're right!

keremc:
@sim642 I was thinking of making it an option, like the `[@encoding]` directive, but since they are planning to remove `Variant`, it doesn't matter. Thanks for linking that issue!

src/domains/printable.ml
keremc: It doesn't need to. This was probably accidental.

Issue #245, opened 2021-06-07
Running without base analysis


sim642:
Recently @vesalvojdani told me about not being able to run without base analysis, e.g. just constants analysis or something else didactical, so all assertions don't pass by virtue of base. At some point I also wanted to do this with octApron to see how it's responding to a simple assertion as well.

I think there are two issues:
- [x] `FromSpec.tf_proc` queries the inner analyses using `EvalFunvar` to get possible functions to call: https://github.com/goblint/analyzer/blob/826cf00a48797d73030e4887004939e96381fd22/src/framework/constraints.ml#L675-L680
   This is only answered by base, so if that's deactivated, the query returns top and folding over top raises an exception.
   It should be possible to fix it for simple cases at least, where the called functions are statically known (so no function pointer call) and `FromSpec` could handle it directly. Then base analysis would only be necessary when function pointer calls are needed.
- [x] `PathSensitive2.map` converts bottom state into dead code: https://github.com/goblint/analyzer/blob/826cf00a48797d73030e4887004939e96381fd22/src/framework/constraints.ml#L999-L1005
   Normally the inner state isn't bot because some default analysis has a non-bot local state, but only running constants analysis, it would be, and thus immediately converted to dead code here by exception. This shouldn't actually happen because dead code is actually indicated by `DeadCodeLifter`'s bot value, which is brought about explicitly via `raise Deadcode`, not implicitly by the inner state just happening to be bot.

Other than these, I don't know of any reason why running without base shouldn't be possible. There shouldn't be an inherent need for it in a generic framework.

sim642:
I thought I'd go ahead and fix this. The first task is straightforward (https://github.com/goblint/analyzer/commit/1f97eb347f10a0c9993346aeb2dc02ddc1dc1d8e), but the second one has me puzzled now.

I'm not even sure why I thought that would be an issue without having fixed the first one, because having done that, running constants analysis alone (without base) worked just fine. Even when the local state of constants analysis is actually empty (so bottom), like in this example:
https://github.com/goblint/analyzer/blob/a271f879b59ff5601183c79e89319bb6dab984d6/tests/regression/33-constants/04-empty-not-dead.c#L3-L9

Turns out the whole reason it's fine is because `D.add` on the Hoare set domain used for paths allows adding inner `bot` values into the set, so it doesn't become empty (i.e. `D.is_bot`) to trigger the problem that I thought it would.
[...]

PR #262, opened 2021-06-22
Simple tutorial for getting started

vesalvojdani:
This was tried by one student who had already figured out most of the stuff one week before, so maybe not the most representative sample. I will try it on a some more new students before we merge this, but feel free to comment already if you have any opinions on how this first analysis thing should be written.

There is one question of whether one should change constraint.ml to not use the query system to resolve function calls if the call is completely known. Right now, I am not turning off the base analysis, but this can cause some confusion about what the student implemented and what goblint discovered on its own.


vesalvojdani:
Ah, nice, it can already be viewed here:
https://goblint--262.org.readthedocs.build/en/262/developer-guide/firstanalysis/

docs/developer-guide/firstanalysis.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+If you run goblint out of the box on this example it will not work: `./goblint --enable dbg.debug example.c` will claim that the assertion in unknown.
+Goblint could verify that this asertion does hold using interval analysis (`--enable ana.int.interval`), but here we will implement a simple sign analysis instead.
+
+We begin by registering and running a new analysis called "signs":
+
+1. Create a copy of the unit analysis (*src/analyses/unit.ml*) into the file *src/analyses/signs.ml*.
+2. Inside the file, change the name from "unit" to "signs".
+3. Try to run goblint with the new analysis enabled: `./goblint --sets "ana.activated[+]" signs --enable dbg.debug example.c`. The result will still be that nothing is verified.
```
sim642, line 43: No need for `""` around `ana.activated[+]`.

docs/developer-guide/firstanalysis.md
```diff
@@ -0,0 +1,144 @@
...
+struct
+  type t = Neg | Zero | Pos [@@deriving eq, to_yojson]
+  let name () = "signs"
+  let show x = match x with
+  | Neg -> "-"
+  | Pos -> "+"
+  | Zero -> "0"
+
+  (* We need the following to generate output... *)
+  include Printable.StdPolyCompare
```
sim642, line 62: Shouldn't recommend `StdPolyCompare`, but just `Std` and use `[@@deriving ord]` instead.

docs/developer-guide/firstanalysis.md
```diff
@@ -0,0 +1,144 @@
...
+```
+let eval (d: D.t) (exp: exp): SL.t = match exp with
+| Const (CInt64 (i, _, _)) -> SL.of_int i
+| Lval (Var x, NoOffset) -> D.find x d
+| _ -> SL.top ()
+
+let assign ctx (lval:lval) (rval:exp) : D.t =
+  let d = ctx.local in
+  match lval with
+  | Var x, NoOffset -> D.add x (eval d rval) d
```
sim642, line 119: We shouldn't teach unsound analyses, so this should check `vaddrof` like the constants analysis.

docs/developer-guide/firstanalysis.md
```diff
@@ -0,0 +1,144 @@
...
+module G = Lattice.Unit
+module C = D
+```
+
+We will not care about the globals and the context, but it's convenient to have the same context as the domain itself.
+
+
+## Implementing the sign analysis
+
+The key part now is to define transfer functions for assignment. Here is a simple version that will suffice for our example. Replace the definition of `assign` in your file with the following:
```
sim642, line 108: The other default definitions in the unit analysis do undesired things like passing the map into functions.

docs/developer-guide/firstanalysis.md
vogler: They're needed when using `zsh`.

docs/developer-guide/firstanalysis.md
vesalvojdani: Mmm... I wanted it really basic. Granted, adddding a `when` statement here isn't that complicated, but one should then explain it maybe. Anyway,  I'm not sure we necessarily have to insist on this analysis being sound or meaningful in any way. The goal is simply to get people started. They can learn about soundness later. :)

docs/developer-guide/firstanalysis.md
```diff
@@ -0,0 +1,144 @@
...
+With this in place, we should have sufficient information to tell Goblint that the assertion does hold. We need to answer assertion queries as follows:
+
+```
+let assert_holds (d: D.t) (e:exp) = match e with
+| BinOp (Gt, e1, e2, t) -> SL.gt (eval d e1) (eval d e2)
+| _ -> false
+
+let query ctx (type a) (q: a Queries.t): a Queries.result =
+  let open Queries in match q with
+  | Assert e when assert_holds ctx.local e -> `Lifted true
```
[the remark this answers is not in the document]
vesalvojdani: I needed something that works now, but what are the plans once this is removed? Anyway, it's not clear what would be the best way to integrate the analysis into getting something that visibly does something. The student was a little bit overwhelmed by the types here, so maybe the query system isn't the best thing to involve in a "first tutorial" anyway.

docs/developer-guide/firstanalysis.md
vesalvojdani: Mmm... again, I don't actually care, but as part of our official docs, this (and the soundness) is problematic in the sense that if people use it as a template to actually write their own analyses. My intention was more to have a throwaway analysis to get people started on goblint, not a template for how to start on a new analysis that should be used.

docs/developer-guide/firstanalysis.md
vesalvojdani: This is a simple fix for this tutorial, but why haven't we deleted StdPolyCompare? I just copy-pasted this from somewhere. :)

docs/developer-guide/firstanalysis.md
sim642: Thanks for pointing that out. Make sense to leave it as is then. I'm now a bit worried that there could be other places where it's been used without quotes, but if they ever come up they can be fixed. Most of our scripts use a bash shebang so I guess that also has avoided such issues from appearing.

docs/developer-guide/firstanalysis.md
sim642: An alternative might be to just have a comment like "This is unsound if x is modified via pointers!".

docs/developer-guide/firstanalysis.md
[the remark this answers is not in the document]
sim642: > I needed something that works now, but what are the plans once this is removed?

The plan is that for an assertion `EvalInt` query is made instead and one can answer whether it's true or false or unknown through its integer value, which it ultimately has in C anyway.

> Anyway, it's not clear what would be the best way to integrate the analysis into getting something that visibly does something. The student was a little bit overwhelmed by the types here, so maybe the query system isn't the best thing to involve in a "first tutorial" anyway.

Good point. Maybe avoid queries here indeed and just implement `special` to handle the `Assert` case to print the result out itself, e.g. "Assertion x < y holds in signs analysis". It also makes it clear that the assertion holds just due to the signs analysis, not due to something base did itself.

It would also remove the dependency on base analysis here, because whichever query is used, that depends on the base analysis doing the query for an assertion and printing the output, so it wouldn't be possible to run without base at all to see any output.

sim642:
> There is one question of whether one should change constraint.ml to not use the query system to resolve function calls if the call is completely known.

By the way, this is tracked by #245.

docs/developer-guide/firstanalysis.md
```diff
@@ -0,0 +1,144 @@
...
+We will implement a very simple sign analysis.
+You need to install goblint for development.
+
+## First test
+
+Create a C file in the Goblint root directory for testing.
+Here we call it `example.c`. This is the first test:
+
+
+```
```
sim642, line 13: The code blocks should specify language (`c`/`ocaml`). I think it uses some generic tokenizer and highlighter without them (or tries to guess the language?).

docs/developer-guide/firstanalysis.md
michael-schwarz: Yes, `StdPolyCompare` and `BlankPolyCompare` being gone is great! Less possibility for bugs!

docs/developer-guide/firstanalysis.md
vesalvojdani: Actually, adding and explaining vaddrof turned out to be a single sentence... so I'll include it.

docs/developer-guide/firstanalysis.md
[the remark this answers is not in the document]
vesalvojdani: But on the other hand, it is nice to showcase the query system and that these extra analyses can help improve stuff and not just work separately. Maybe, I just switch to the Int query and then if it really is a problem for students one could change it.

vesalvojdani:
I should maybe not have done that cleanup in this same pull request... or I may as well do something about not being able to run without base. Although, it's not clear if that's a good idea. It might actually be nice to show how new analyses can actually help the base analysis, rather than everything being separated.

docs/developer-guide/firstanalysis.md
vesalvojdani: The soundness issue was more serious, but these unnecessary stuff happening on features that are beyond the scope of the tutorial --- is it harmful to ignore it?

sim642: If global variables are also excluded, then I suppose it might be fine to pass the map into the callee. Currently globals are added and could be modified in the callee, but `combine` goes back to the values of globals before the call.

docs/developer-guide/firstanalysis.md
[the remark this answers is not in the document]
sim642: The query system (and probably the event system too) might deserve separate documentation anyway, that also shows how to use `ctx.ask`, because using existing queries is probably more common than implementing new ones. Also, the explanation of the latter needs to go into some details like how may queries are different from must queries.

sim642:
>  It might actually be nice to show how new analyses can actually help the base analysis, rather than everything being separated.

This could also be a separate section in the tutorial, where first the assertion checking is done just in signs, but then something about queries is explained in order to benefit base.

michael-schwarz:
What are our plans with this one? What is still stopping us from merging?

sim642:
> I will try it on a some more new students before we merge this, but feel free to comment already if you have any opinions on how this first analysis thing should be written.

The original description at least says this, not sure about the timeline for that.

src/framework/analyses.ml
```diff
@@ -481,3 +481,34 @@ struct
...
+  let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) : D.t =
+    au
+
+  let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    ctx.local
+
+  let startstate v = D.bot ()
+  let threadenter ctx lval f args = [D.top ()]
+  let threadspawn ctx lval f args fctx = ctx.local
+  let exitstate  v = D.top ()
```
sim642, line 513: I'm not sure how correct these even are and if these same defaults could be used for any other analysis except constants. Like why is `startstate` bottom but the thread entry state top?

PR #264, opened 2021-06-28
Serialize additional data for Gobview

keremc:
* Update `goblint-cil` to commit `94ba9835b86977e73fb9b6ea0c09adc43f716a43` (to be able to use goblint/cil#38)
* Add a new boolean option "gobview".
* Serialize the MCP analysis table, the CIL file, the Cabs2cil environment, the Goblint warning table, and the GC/runtime stats if this flag is set.

Example usage:
```bash
$ ./goblint --enable gobview main.c # Implies --sets save_run run
$ ./goblint --sets save_run data --enable gobview main.c
```

### DIFF
--- a/goblint.opam
+++ b/goblint.opam
@@ -49,5 +49,5 @@ build: [
 ]
 dev-repo: "git+https://github.com/goblint/analyzer.git"
 pin-depends: [
-  [ "goblint-cil.1.7.8" "git+https://github.com/goblint/cil.git#5fdf44c42b194130fefabbbe8a7a29818ad3f539" ]
+  [ "goblint-cil.1.8.1" "git+https://github.com/goblint/cil.git#d61cfa6a403f4d278c86344091bc07dccd8b549b" ]
 ]
--- a/goblint.opam.locked
+++ b/goblint.opam.locked
@@ -26,7 +26,7 @@ depends: [
   "cppo" {= "1.6.6"}
   "dune" {= "2.7.1"}
   "easy-format" {= "1.3.2"}
-  "goblint-cil" {= "1.7.8"}
+  "goblint-cil" {= "1.8.1"}
   "mlgmpidl" {= "1.2.13"}
   "num" {= "1.3"}
   "ocaml" {= "4.11.1"}
@@ -68,7 +68,7 @@ build: [
 ]
 dev-repo: "git+https://github.com/goblint/analyzer.git"
 pin-depends: [
-  [ "goblint-cil.1.7.8" "git+https://github.com/goblint/cil.git#5fdf44c42b194130fefabbbe8a7a29818ad3f539" ]
+  [ "goblint-cil.1.8.1" "git+https://github.com/goblint/cil.git#d61cfa6a403f4d278c86344091bc07dccd8b549b" ]
 ]
 name: "goblint"
 version: "dev"
--- a/goblint.opam.template
+++ b/goblint.opam.template
@@ -1,3 +1,3 @@
 pin-depends: [
-  [ "goblint-cil.1.7.8" "git+https://github.com/goblint/cil.git#5fdf44c42b194130fefabbbe8a7a29818ad3f539" ]
+  [ "goblint-cil.1.8.1" "git+https://github.com/goblint/cil.git#d61cfa6a403f4d278c86344091bc07dccd8b549b" ]
 ]
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -383,23 +383,24 @@ struct
 
     let solve_and_postprocess () =
       (* handle save_run/load_run *)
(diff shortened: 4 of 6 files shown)


sim642:
The CI seems to fail because of this:
```
goblint-cil is now pinned to git+https://github.com/goblint/cil.git#94ba9835b86977e73fb9b6ea0c09adc43f716a43 (version 1.8.0)
The following dependencies couldn't be met:
  - goblint → goblint-cil >= 1.8.0 → ocaml = 4.10.0
      base of this switch (use `--unlock-base' to force)
```

michael-schwarz:
@keremc I guess if you now update the pin it should work?

sim642:
I'm confused why the CI works now. Why was it necessary to release a new CIL version to get the opam pin working? And why was it necessary to remove the opam lock file for CIL? Pinning worked fine with the lock file present before.

michael-schwarz:
The last pin was at https://github.com/goblint/cil/tree/5fdf44c42b194130fefabbbe8a7a29818ad3f539, where there was no lock file.
The problem is that it tries to use the lockfile for all dependencies as well, so we would have to keep these lockfiles in sync between Goblint and CIL, which is probably not worth the hassle.

sim642:
Ok, I didn't realize the pin was so much out of date all this time, because there had been changes to CIL since then. I thought Goblint depended on some of them at least, but it then really didn't?

Also, do none of our other dependencies have a lock file then? Or do opam repository dependencies not supply them but pinned dependencies do?

michael-schwarz:
> I thought Goblint depended on some of them at least, but it then really didn't?

I think it was only the fork for Gobview that depended on these changes. We immediately updated CIL accordingly, but only now start to depend on these changes to CIL in Goblint as we upstream things from this fork.

Another issue that this fixes is the one we had with https://github.com/goblint/analyzer/blob/8ed4ba7a9dc8816afff424701f86015937c00b1b/tests/regression/31-ikind-aware-ints/15-strange.c#L1-L9
as the new pin includes the fix for  https://github.com/goblint/cil/issues/29, so we could remove the skip annotation there.

____________

> Also, do none of our other dependencies have a lock file then? Or do opam repository dependencies not supply them but pinned dependencies do?

From here: https://opam.ocaml.org/doc/man/opam-pin.html

> In commands that use opam files found from pinned sources, if a variant of the file with an added .locked extension is found (e.g. foo.opam.locked besides foo.opam), that will be used instead.

So I guess it is the latter.

sim642:
Ugh, I feel like I ask this every time the pin gets updated but how am I supposed to update it locally?

Right now on master I get this compilation error:
```
File "src/framework/control.ml", line 446, characters 33-42:
446 |               Serialize.marshal (Stats.top, Gc.quick_stat ()) stats
                                       ^^^^^^^^^
Error: Unbound value Stats.top
```

I tried `opam install goblint-cil` but it just outputs the following:
```
><> Synchronising pinned packages ><><><><><><><><><><><><><><><><><><><><><><>
[goblint-cil.1.7.8] no changes from git+https://github.com/goblint/cil.git#5fdf44c42b194130fefabbbe8a7a29818ad3f539

[NOTE] Package goblint-cil is already installed (current version is 1.7.8).
```

The commit hash is the new one but it still says 1.7.8 and nothing is updated. There must be a way to do it without wiping the entire opam switch...

sim642:
Just `upgrade` or `reinstall` of goblint-cil did the same, but the following two command seemed to work:

1. `opam install --deps-only .`:
    ```
    Nothing to do.
    ```

2. `opam upgrade goblint-cil`:
    ```
    <><> Synchronising pinned packages ><><><><><><><><><><><><><><><><><><><><><><>
    [goblint-cil.1.8.1] no changes from git+https://github.com/goblint/cil.git#d61cfa6a403f4d278c86344091bc07dccd8b549b
    
    The following actions will be performed:
      ↗ upgrade goblint-cil 1.7.8 to 1.8.1*
    
    <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
    ⊘ removed   goblint-cil.1.7.8
    ∗ installed goblint-cil.1.8.1
    Done.
    ```

@michael-schwarz @vogler Is there any simpler way this should be done? If not, then I'll add this to our documentation so everyone knows how to fix their master build now without completely wiping the opam switch...

vogler:
Haven't checked, but I remember it needed some unexpected invocation to actually update.
`opam pin add goblint-cil git+https://github.com/goblint/cil.git` maybe?

sim642:
That's just adding the pin manually to the master branch of the goblint-cil repo instead of taking the one with the specific version and commit hash from the opam dependencies file. It would be completely pointless to have to manually re-add the exact thing that's in the file.

michael-schwarz:
@sim642 I can't hink of anything better than what you propose.

The problem with `opam pin add goblint-cil git+https://github.com/goblint/cil.git` is that it will sometimes break when we have updated CIL but have not yet adapted Goblint.

PR #270, opened 2021-07-02
Change `enter`, `combine`, `FunctionEntry`, `Function` argument `varinfo` -> `fundec`

sim642:
Many of them use `Cilfacade.getdec` to look up the `fundec` anyway, e.g. to check `sformals`. `FromSpec` already looked up the existence of `fundec` as well to determine whether normal or special function call should be performed, but it ignored the `fundec` instead of passing it into `enter` and `combine` for normal calls. This makes normal calls more consistent because their `body` and `return` already have `fundec` arguments.

`special` will still have to have a `varinfo` argument because those functions have no `fundec`, but they don't do `enter` and `combine` either.

## TODO
There's a few related changes I'd do:
- [x] Change `FunctionEntry` and `Function` nodes to contain `fundec` instead of `varinfo`?
- [x] ~~Change `threadenter` and `threadspawn` similarly?~~ (see comment)

### Questions
Both are complicated by one factor: it's possible to spawn threads with `extern` (i.e. unknown) functions as the thread body. As I started looking into this, I realized that our current support for such thing is broken and accidental: base analysis properly calls `ctx.spawn` for them (with a `varinfo`) and `FromSpec` calls `threadenter` and `threadspawn` correspondingly. The code surrounding `threadenter` treats it always as if it's a normal function: side effects to `FunctionEntry` and demands (and ignores) `Function` to force the function to be analyzed. This causes warnings like:
```
Creating a thread from unknown function magic (tests/regression/02-base/52-otherfun-special.c:9)
Calculated state for undefined function: unexpected node FunctionEntry magic (initfun:0)
Calculated state for undefined function: unexpected node Function magic (initfun:0)
```
These nodes and their values are complete nonsense: because there's no definition, there's no CFG and thus there's no edges even between them. We thereby assume that an unknown thread doesn't do anything (which isn't exactly sound).

So the big question is, what should we do when unknown threads get spawned. What may it do? This is slightly similar to calling an unknown function in a defined thread: currently we assume it invalidates globals but doesn't lock, unlock, spawn, etc.

This determines what part of that can be changed. I think `threadspawn` will have to take `varinfo` because the spawner should be aware of a thread having been spawned regardless (to go multithreaded, etc). But what should `threadenter` be for, only entering defined threads (like `enter`) or also for entering unknown threads (if so, then what would be done with that state)? Maybe there also needs to be `threadspecial` (like `special`) that does the invalidation?

(There's a related even more obscure case: `otherfun`s are also "spawned" using `threadenter`. So what if you add an extern function to `otherfun`? Apparently values in `otherfun` etc without corresponding definitions are silently ignored...)

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1814,14 +1814,13 @@ struct
         Some (lval, v, args)
       with Not_found ->
         if LF.use_special f.vname then None (* we handle this function *)
-        else if isFunctionType v.vtype then (
-          M.warn_each ("Creating a thread from unknown function " ^ v.vname);
+        else if isFunctionType v.vtype then
           let args = match arg with
             | Some x -> [x]
             | None -> []
           in
-          Some (lval, v,  args)
-        ) else (
+          Some (lval, v, args)
+        else (
           M.warn_each ("Not creating a thread from " ^ v.vname ^ " because its type is " ^ sprint d_type v.vtype);
           None
         )
@@ -1907,7 +1906,9 @@ struct
         newst
       end
 
-  let special_unknown_invalidate ctx ask gs st args =
+  let special_unknown_invalidate ctx ask gs st f args =
+    (if not (CilType.Varinfo.equal f dummyFunDec.svar) && not (LF.use_special f.vname) then M.warn_each ("Function definition missing for " ^ f.vname));
+    (if CilType.Varinfo.equal f dummyFunDec.svar then M.warn_each ("Unknown function ptr called"));
     let addrs =
       if get_bool "sem.unknown_function.invalidate.globals" then (
         M.warn_each "INVALIDATING ALL GLOBALS!";
@@ -2098,17 +2099,14 @@ struct
         let st =
           match LF.get_invalidate_action f.vname with
           | Some fnc -> invalidate ~ctx (Analyses.ask_of_ctx ctx) gs st (fnc `Write  args)
-          | None -> (
-              (if not (CilType.Varinfo.equal f dummyFunDec.svar) && not (LF.use_special f.vname) then M.warn_each ("Function definition missing for " ^ f.vname));
-              (if CilType.Varinfo.equal f dummyFunDec.svar then M.warn_each ("Unknown function ptr called"));
-              special_unknown_invalidate ctx (Analyses.ask_of_ctx ctx) gs st args
-              (*
-               *  TODO: invalidate vars reachable via args
-               *  publish globals
-               *  if single-threaded: *call f*, privatize globals
-               *  else: spawn f
-               *)
-            )
+          | None ->
+            special_unknown_invalidate ctx (Analyses.ask_of_ctx ctx) gs st f args
+            (*
+             *  TODO: invalidate vars reachable via args
+             *  publish globals
+             *  if single-threaded: *call f*, privatize globals
+             *  else: spawn f
+             *)
         in
         (* invalidate lhs in case of assign *)
         let st = match lv with
@@ -2179,7 +2177,7 @@ struct
     | exception Not_found ->
       (* Unknown functions *)
       let st = ctx.local in
-      let st = special_unknown_invalidate ctx (Analyses.ask_of_ctx ctx) ctx.global st args in
+      let st = special_unknown_invalidate ctx (Analyses.ask_of_ctx ctx) ctx.global st f args in
       [st]
 
   let threadspawn ctx (lval: lval option) (f: varinfo) (args: exp list) fctx: D.t =
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1907,6 +1907,25 @@ struct
         newst
       end
 
+  let special_unknown_invalidate ctx ask gs st args =
+    let addrs =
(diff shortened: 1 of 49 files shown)


sim642:
I tried doing the change with `FunctionEntry` and `Function` nodes nevertheless and besides the `threadenter` problem, there's another one in arinc and extract_arinc:
https://github.com/goblint/analyzer/blob/95d75cc881c800630c1072fdc1d4a2f21165c089/src/analyses/arinc.ml
Inside `special` it's recording the predecessor (?) as `Function f`, where `f` is the `varinfo` for an unknown function (`LAP_Se_CreateProcess`).

I suppose this could be worked around by creating a fake `fundec` for that function just for this `Pred` usage. Although I'm slightly confused how this `Function LAP_Se_CreateProcess` is actually useful for anything because it itself has no predecessor because it's not defined.

sim642:
I decided not to change `threadenter` and `threadspawn because `threadenter` is what assigns a thread ID to the create thread (which should happen for unknown threads as well) and `threadspawn` informs the creator about the new thread (which should happen for unknown threads as well).

This allows `threadenter` to do the `Cilfacade.getdec` lookup itself to distinguish the two cases and do globals invalidation when the created thread is an unknown function. I didn't do anything to fix the above mentioned `otherfun`s case because `otherfun` itself is void of clear semantics (they are entered like new threads but the main thread is still in singlethreaded mode).

src/analyses/base.ml
```diff
@@ -2168,11 +2167,31 @@ struct
...
+              | _ -> acc
+            ) []
+        )
+        else
+          []
+      in
+      (* TODO: what about escaped local variables? *)
+      (* invalidate non-static globals for unknown functions *)
+      let st = invalidate ~ctx (Analyses.ask_of_ctx ctx) ctx.global st addrs in
+      [st]
```
jerhard, line 2194: Can one deduplicate this code?

src/framework/constraints.ml
```diff
@@ -554,10 +554,15 @@ struct
...
-          if not full_context then sidel (FunctionEntry f, c) d;
-          ignore (getl (Function f, c))
+          match Cilfacade.getdec f with
+          | fd ->
+            let c = S.context d in
+            if not full_context then sidel (FunctionEntry fd, c) d;
+            ignore (getl (Function fd, c))
+          | exception Not_found ->
+            (* unknown function *)
+            ()
```
jerhard, line 565: Should we produce some kind of warning here?

sim642: Well, spawning threads of unknown functions already warns, but just in a different place:
https://github.com/goblint/analyzer/blob/262b6dab31938fafbc3cd9237f40b134812bd27d/src/analyses/base.ml#L1819

There's no point in warning twice, but I could move the warning here since this is a bit more general place for it (not dependent on base).

src/analyses/base.ml
sim642: Should be doable, but the bigger question is actually how much of it should be shared. `special` also has special more precise invalidation for `LibraryFunctions` and a bunch of special implementations. One option might even be just delegating to `special` here, but I don't know if that has any value. Spawning a thread from `memcpy` or `malloc` seems like complete nonsense.

src/framework/constraints.ml
sim642: I moved this warning. By warning at the suggested location, it also happens if some analysis other than base spawns a thread.

src/analyses/base.ml
sim642: I extracted just this part out. Calling `special` from `threadenter` is problematic because it would want to recursively spawn threads, but that's not currently allowed in `threadenter`.

sim642:
I added a few more fixes related to thread creation invalidation.

src/analyses/base.ml
```diff
@@ -1815,10 +1814,13 @@ struct
         Some (lval, v, args)
       with Not_found ->
         if LF.use_special f.vname then None (* we handle this function *)
-        else if isFunctionType v.vtype then (
-          M.warn_each ("Creating a thread from unknown function " ^ v.vname);
```
michael-schwarz, line 1819: Maybe add a comment on where the warning happens?

src/framework/constraints.ml
```diff
@@ -554,10 +554,16 @@ struct
...
-          if not full_context then sidel (FunctionEntry f, c) d;
-          ignore (getl (Function f, c))
+          match Cilfacade.getdec f with
+          | fd ->
+            let c = S.context d in
+            if not full_context then sidel (FunctionEntry fd, c) d;
+            ignore (getl (Function fd, c))
+          | exception Not_found ->
+            (* unknown function *)
+            M.warn_each ("Created a thread from unknown function " ^ f.vname)
```
michael-schwarz, line 565: Maybe we should phrase this differently because base might decide to not create a thread here depending on the type.

michael-schwarz:
LGTM!

src/framework/constraints.ml
sim642: If base decides to not create it based on the type, this `ctx.spawn` is never called anyway.

PR #271, opened 2021-07-05
Allow running without base analysis

### LINKED ISSUE #245: Running without base analysis

sim642:
Closes #245.

The explanation for path-sensitivity and Hoare set changes is here: https://github.com/goblint/analyzer/issues/245#issuecomment-873976513. Is it really the case that the Hoare set incorrectly filtered out `bot` values? I'm especially surprised since at some point for the traces paper we discussed these bottom and top things w.r.t Hoare domains and never questioned this.

### DIFF
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -986,7 +986,7 @@ struct
 
   let exitstate  v = D.singleton (Spec.exitstate  v)
   let startstate v = D.singleton (Spec.startstate v)
-  let morphstate v d = D.map_noreduce (Spec.morphstate v) d
+  let morphstate v d = D.map (Spec.morphstate v) d
 
   let call_descr = Spec.call_descr
 
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -676,8 +676,15 @@ struct
   let tf_proc var edge prev_node lv e args getl sidel getg sideg d =
     let ctx, r, spawns = common_ctx var edge prev_node d getl sidel getg sideg in
     let functions =
-      let ls = ctx.ask (Queries.EvalFunvar e) in
-      Queries.LS.fold (fun ((x,_)) xs -> x::xs) ls []
+      match e with
+      | Lval (Var v, NoOffset) ->
+        (* Handle statically known function call directly.
+           Allows deactivating base. *)
+        [v]
+      | _ ->
+        (* Depends on base for query. *)
+        let ls = ctx.ask (Queries.EvalFunvar e) in
+        Queries.LS.fold (fun ((x,_)) xs -> x::xs) ls []
     in
     let one_function f =
       let has_dec = try ignore (Cilfacade.getdec f); true with Not_found -> false in
--- a/tests/regression/33-constants/03-empty-not-dead-branch.c
+++ b/tests/regression/33-constants/03-empty-not-dead-branch.c
@@ -1,0 +1,16 @@
+//PARAM: --set ana.activated '["constants"]'
+
+int g;
+
+int main() {
+    // state: {bot}, because no locals
+    if (g) {
+        g = 1; // state: {bot}, because Hoare set add (in PathSensitive2 map) keeps bot, while reduce would remove
+    }
+    else {
+        g = 2; // state: {bot}, because Hoare set add (in PathSensitive2 map) keeps bot, while reduce would remove
+    }
+    // state: {}, because reduce is applied after join and that removes bot
+    g = 3; // state: {} and PathSensitive2 map raises Deadcode
+    return 0;
+}
--- a/src/domains/hoareDomain.ml
+++ b/src/domains/hoareDomain.ml
@@ -207,9 +207,12 @@ struct
         let evil = choose (filter p s1) in
         dprintf "%a:\n" B.pretty evil
         ++
-        fold (fun other acc ->
-            (dprintf "not leq %a because %a\n" B.pretty other B.pretty_diff (evil, other)) ++ acc
-          ) s2 nil
+        if is_empty s2 then
+          text "empty set s2"
+        else
+          fold (fun other acc ->
+              (dprintf "not leq %a because %a\n" B.pretty other B.pretty_diff (evil, other)) ++ acc
+            ) s2 nil
       with _ ->
         dprintf "choose failed b/c of empty set s1: %d s2: %d"
         (cardinal s1)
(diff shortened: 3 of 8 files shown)


src/framework/constraints.ml
```diff
@@ -676,8 +676,15 @@ struct
   let tf_proc var edge prev_node lv e args getl sidel getg sideg d =
     let ctx, r, spawns = common_ctx var edge prev_node d getl sidel getg sideg in
     let functions =
-      let ls = ctx.ask (Queries.EvalFunvar e) in
-      Queries.LS.fold (fun ((x,_)) xs -> x::xs) ls []
+      match e with
+      | Lval (Var v, NoOffset) ->
```
michael-schwarz, line 680: I was a bit skeptical here, because of code such as this one:
~~~C
int f(int x) {
  return x;
}

int main() {
  int (*ptr)(int) = &f;
  ptr(3);
}
~~~

But CIL adds the implicit dereference, so it also works here.

~~~C
  (*ptr)(3);
~~~

Are there any similar cases we should have a look at?

tests/regression/33-constants/03-empty-not-dead-branch.c
```diff
@@ -0,0 +1,16 @@
+//PARAM: --set ana.activated '["constants"]'
+
+int g;
+
+int main() {
+    // state: {bot}, because no locals
```
michael-schwarz, line 6: Is this not some sort of bot and top confusion in the constant analysis here? One would assume that we start out with top and not bot

sim642: Constants starts with a bottom `MapBot`, which is an empty map with missing values implicitly bottom:
https://github.com/goblint/analyzer/blob/262b6dab31938fafbc3cd9237f40b134812bd27d/src/analyses/constants.ml#L16-L17
https://github.com/goblint/analyzer/blob/262b6dab31938fafbc3cd9237f40b134812bd27d/src/analyses/constants.ml#L98

But the base analysis does the exact same thing:
https://github.com/goblint/analyzer/blob/262b6dab31938fafbc3cd9237f40b134812bd27d/src/cdomains/baseDomain.ml#L11-L12
https://github.com/goblint/analyzer/blob/262b6dab31938fafbc3cd9237f40b134812bd27d/src/analyses/base.ml#L51

So by the same logic our base analysis is wrong too because the same empty (`bot`) map would be if you just run base (without threadid and maybe others which always have non-bottom state that prevents this issue from occurring) on the same example as well.

---

Bottom of one analysis doesn't inherently mean dead code, `raise Deadcode` means dead code and that's handled by `DeadcodeLifter` and it's special bottom.
Just think about the unit analysis: it's only possible value is `()`, which is also the bottom of the unit lattice. So if you were to just activate the unit analysis, which is the stupidest possible thing, it's bottom everywhere. But that doesn't mean it's soundly analyzed all the code to be dead. The code is only considered dead when some analysis (hopefully soundly) raises `Deadcode` to explicitly indicate that.

src/framework/constraints.ml
sim642: That's a great example I didn't think of, but I guess we're lucky CIL does that. I was a bit unsure if this would always be correct.

I think if you follow `EvalFunvar (Lval (Var v, NoOffset))` from `query` through `eval_rv` and `eval_lv`, then it ends up returning a singleton address set with `(v, NoOffset)` in it. So it should be the maximal inlining of that logic, which also depends on CIL adding the dereference (i.e. `Mem` constructor).

I guess it'd be possible to add some extra conditions here, like `v` needs to be the `svar` of some `fundec` but AFAIK base doesn't do anything like that at any step. And it would be less general because no special functions (e.g. `assert`) can be called without base activated.

sim642: I think in CIL the explicit dereference is being added here: https://github.com/goblint/cil/blob/94ba9835b86977e73fb9b6ea0c09adc43f716a43/src/frontc/cabs2cil.ml#L4203-L4214.

tests/regression/33-constants/03-empty-not-dead-branch.c
```diff
@@ -0,0 +1,16 @@
...
+    if (g) {
+        g = 1; // state: {bot}, because Hoare set add (in PathSensitive2 map) keeps bot, while reduce would remove
+    }
+    else {
+        g = 2; // state: {bot}, because Hoare set add (in PathSensitive2 map) keeps bot, while reduce would remove
+    }
+    // state: {}, because reduce is applied after join and that removes bot
+    g = 3; // state: {} and PathSensitive2 map raises Deadcode
```
michael-schwarz, line 14: Can we enhance this test in some way so we would notice if this became bot again through some changes?

michael-schwarz: Also, the comment seems outdated?

sim642: I'll fix the comments because, yes, they correspond to the previous problematic state.

Not sure how we could enhance this test, because there's no way to test reachability here. Normally when base is active, `assert(1)` works for that purpose but constants analysis doesn't do anything about asserts. And the same test doesn't work with base, because it would have `g` in the local state, so it wouldn't be bottom.

An option that I've thought about is extracting an assert output analysis from base into a standalone analysis which just queries others (just like base does right now and after #251). Then I guess constants can run just with assert analysis and respond to `EvalInt` queries with the constant value (which constants can evaluate).

PR #272, opened 2021-07-06
Merge relevant commits from SAS '21 artifact description

michael-schwarz:
We need to go through changes again to see what we want in master and what not


michael-schwarz:
What do we want to do with the artifact description? Does it fit nicely into the docs in some sense?

michael-schwarz:
Maybe having it in the docs folder but not being served is in the docs (as it currently is), is exactly what we want?

sim642:
Besides the artifact badge that we're supposed to get, could there also be some textual review of the artifact that might ask us to improve something in the artifact description? If so, then we should do that before merging and calling it final.

sim642:
> What do we want to do with the artifact description? Does it fit nicely into the docs in some sense?

> Maybe having it in the docs folder but not being served is in the docs (as it currently is), is exactly what we want?

I think it makes sense to have this information in the docs, even in the readthedocs version. For now I will just put it into a separate artifact descriptions section, so these general descriptions aren't some kind of secret documentation. We don't have too much documentation anyway, so it should be fine for now.

michael-schwarz:
>  could there also be some textual review of the artifact that might ask us to improve something in the artifact description?

I don't know, I emailed Mukherjee yesterday, so I think we should find out soon enough.

sim642:
Review 1 of the artifact suggests the following point for improving the description:

> On the other side,  I found the precision comparison output a bit confusing and not well documented. For example, the precision comparison of the the multi-thread program pfscan,
> outputs the following text (these are just the first few lines):
> 
> ```
> pfscan_comb.c:1269 pqb: protection more precise than mine-W
>   protection: <{...}, def_exc:Unknown int([0,7]); enums:{0, 1}, ...> {
>                 buf ->
>                   {&(alloc@pfscan_comb.c:1204)[def_exc:0; enums:{0}]}
>                 closed -> (Unknown int([0,7]),{0, 1})
>                 less ->
>                   (__data, <def_exc:0; enums:{0}, def_exc:0; enums:{0}, ...> {
>                              __broadcast_seq -> (0,{0})
>                              __futex -> (0,{0})
>                              __lock -> (0,{0})
>                              __mutex -> (0,{0})
>                              __nwaiters -> (0,{0})
>                              __total_seq -> (0,{0})
>                              __wakeup_seq -> (0,{0})
>                              __woken_seq -> (0,{0})
>                            })
>                 more ->
>                   (__data, <def_exc:0; enums:{0}, def_exc:0; enums:{0}, ...> {
>                              __broadcast_seq -> (0,{0})
>                              __futex -> (0,{0})
>                              __lock -> (0,{0})
>                              __mutex -> (0,{0})
>                              __nwaiters -> (0,{0})
>                              __total_seq -> (0,{0})
>                              __wakeup_seq -> (0,{0})
>                              __woken_seq -> (0,{0})
>                            })
>                 mtx ->
>                   (__data, <(__list, <{NULL}>), def_exc:0; enums:{0}, ...> {
>                              __annonCompField1 ->
>                                (__list, <{NULL}> {
>                                           __next ->
>                                             {NULL}
>                                         })
>                              __count -> (0,{0})
>                              __kind -> (0,{0})
>                              __lock -> (0,{0})
>                              __nusers -> (0,{0})
>                              __owner -> (0,{0})
>                            })
>                 nextin -> (Unknown int([-31,31]),not {} ([-31,31]))
>                 nextout -> (Unknown int([-31,31]),not {} ([-31,31]))
>                 occupied -> (Not {0}([-31,31]),not {} ([-31,31]))
>                 qsize -> (Unknown int([-31,31]),not {} ([-31,31]))
>               }
> ...
> ```
> 
> which should states that protection analysis is more precise than mine-W.
> Also the few lines describing this part in the paper (lines 31-37 of page 17) do not help, which is why I suggest the authors to improve the documentation on this point
> (that is, step 3.2 in the 'Step by Step Instructions' section of file 'sas21-artifact-description.md').

sim642:
I'll add something about it, but it will have to be a separate branch from this, because here we already undid some stuff and moved things around. Then we can make the updated final artifact still from the correct place in the history.

michael-schwarz:
> will obtain the Available badge if the authors upload the _submitted_ version of the artifact to the Zenodo SAS 2021 public repository.

The way I understood it, to obtain the available badge, we need to submit the same version that we also submitted for evaluation, so I am not sure if we can update the description. But still worth fixing for the documentation at least.

sim642:
If that's really the case, then it indeed won't matter for the artifact since we'd want the Available badge.

Just seems a bit weird that a reviewer would suggest an improvement when there's no separate final submission (unlike for the papers themselves) into which to make the improvement.

sim642 (approved):
I added that additional explanation now at least for our own documentation then.

PR #274, opened 2021-07-07
Remove unnecessary Skip and Neg(1) edges from CFG

### LINKED ISSUE #231: remove superfluous edges in CFG
The CFG for
https://github.com/goblint/analyzer/blob/93741af0fa57f981c9ffa8c1b9b0251e95c62d6d/tests/regression/34-localization/01-nested.c#L8-L14
looks like this:
image

Would be nicer if those orange dead nodes with `skip` edges were gone, as well as the `Neg(1)` edges where possible.
The relevant cases should be `Loop` and `Goto` here:
https://github.com/goblint/analyzer/blob/93741af0fa57f981c9ffa8c1b9b0251e95c62d6d/src/framework/myCFG.ml#L274-L296
Maybe it's better to adjust our CIL to emit these `Neg(1)` edges and thereby guarantee that loops are always connected.

sim642:
Closes #231.

I tried to very carefully get rid of some of the Skip and Neg(1) edges we insert, while making sure that all the extreme special cases here (the reason why the Skips are added in the first place) still have correct CFGs: https://github.com/goblint/analyzer/blob/47cd7893cd662a520fbb58e76a3a0cb1ec99ccaf/tests/sv-comp/cfg/main_goto_loop_true-unreach-call.c.

Includes some additional fixes like better use of infinite loop successor instead of pseudo return and correct handling of `if (0)`.
I also added an extra check which makes sure that `FunctionEntry` is connected to `Function`.

All the regression tests still pass, but it's quite difficult to be sure that too many of those edges aren't now removed. So everyone's welcome to try to find cases where this excludes too much.

### TODO
- [x] Simplify some parts of `MyCFG`.
- [x] Add more explanation comments to `MyCFG`.

### Example 1 (from #231)
https://github.com/goblint/analyzer/blob/93741af0fa57f981c9ffa8c1b9b0251e95c62d6d/tests/regression/34-localization/01-nested.c#L8-L14

| Before | After |
| --- | --- |
| image | image |

**Change:** All dead Skip nodes and edges removed. Both Neg(1) edges removed.

### Example 2 (from #231)
https://github.com/goblint/analyzer/blob/47cd7893cd662a520fbb58e76a3a0cb1ec99ccaf/tests/regression/03-practical/22-nested-infinite-loops.c#L3-L11

| Before | After |
| --- | --- |
| main | image |

**Change:** All dead Skip nodes and edges removed.

### Example 3
https://github.com/goblint/analyzer/blob/47cd7893cd662a520fbb58e76a3a0cb1ec99ccaf/tests/regression/00-sanity/10-loop_label.c#L3-L6

| Before | After |
| --- | --- |
| image | image |

**Change:** Actually contains the self-loop it should! Correctly connect existing return instead of introducing pseudo return and leaving existing return without predecessors.

### Example 4
https://github.com/goblint/analyzer/blob/596cd371eab5e2a5a7112f6fe62adc68e3527526/tests/regression/00-sanity/20-if-0-realnode.c#L8-L17

| Before | After |
| --- | --- |
| image | image |

**Change:** `if (0)`, which has an else branch isn't completely ignored. What the hell was this 12 years old code trying to achieve? https://github.com/goblint/analyzer/blob/1b9808d46621351b26f6f85b95c7ba1fda9c558d/src/framework/myCFG.ml#L264-L273


src/framework/myCFG.ml
```diff
@@ -181,37 +181,52 @@ let createCFG (file: file) =
...
+          | exception (Failure _) -> if is_entry then stmt else raise Not_found
+          | next -> next
+        in
+        let sid = stmt.sid in
+        match stmt.skind with
+        | Block _ -> realnode is_entry (sid::visited) (next ())
+        | Goto _ -> realnode is_entry (sid::visited) (next ())
+        | Instr [] -> begin
+            let next = next () in
+            if next.sid == stmt.sid
```
michael-schwarz, line 200: This is probably correct here, but I think we should aim to eliminate usage of physical equality unless this is explicitly what is needed.

sim642: This was fine because `sid`s are `int`s, but I heavily refactored this function as well and that check wasn't even necessary (the usual cycle detection handles it as well).

src/framework/myCFG.ml
```diff
@@ -165,138 +165,214 @@ let createCFG (file: file) =
...
-            | Instr [] -> begin
-                if next.sid == stmt.sid
-                then stmt
-                else realnode is_entry (sid::visited) next
-              end
-            | Loop _ -> realnode is_entry (sid::visited) next
-            | If (exp,_,_,_) -> if isZero exp then realnode is_entry (sid::visited) next else stmt
-            | _ -> stmt
+  let addEdge fromNode edge toNode = addEdges fromNode [edge] toNode in
+  let addEdge_fromLoc fromNode edge toNode = addEdge fromNode (getLoc fromNode, edge) toNode in
```
michael-schwarz, line 179: These two things both called `edge` are of different types, right? Maybe we should make this clear here somehow?

src/framework/myCFG.ml
```diff
@@ -165,138 +165,214 @@ let createCFG (file: file) =
...
+          addEdge newst_node (fd_loc, Ret (None, fd)) (Function fd);
           newst_node
         )
         in
+        let loop_head_neg1 = NH.create 3 in
         (* So for each statement in the function body, we do the following: *)
         let handle stmt =
           (* Please ignore the next line. It creates an index of statements
            * so the Eclipse plug-in can know what function a given result
            * belongs to. *)
```
michael-schwarz, line 273: We should update this comment, we no longer have an Eclipse plugin.
This is actually used by `analyses.ml` now (via `getFun` -> `get_containing_function`).

src/framework/myCFG.ml
```diff
@@ -165,138 +165,214 @@ let createCFG (file: file) =
...
-                    realnode true (List.nth stmt.succs 1)
-                  with Failure _ -> realnode true (List.hd stmt.succs) in
-                mkEdge stmt (Test (exp, true )) true_stmt;
-                mkEdge stmt (Test (exp, false)) false_stmt
+            let edges = List.map edge_of_instr instrs in
+            let add_succ_node succ_node = addEdges (Statement stmt) edges succ_node in
+            begin match real_succs () with
+              | [] -> add_succ_node (Lazy.force pseudo_return) (* stmt.succs can be empty if last instruction calls non-returning function (e.g. exit), so pseudo return instead *)
+              | [succ] -> add_succ_node (Statement succ)
+              | _ -> failwith "MyCFG.createCFG: >1 non-empty Instr succ"
```
michael-schwarz, line 306: Non empty? Or not empty nor singleton?

src/framework/myCFG.ml
```diff
@@ -308,9 +384,22 @@ let createCFG (file: file) =
         let loop_heads = find_loop_heads_fun (module TmpCfg) fd in
         let reachable_return = find_backwards_reachable (module TmpCfg) (Function fd) in
```
michael-schwarz, line 385: `reachable_from_return`?

michael-schwarz (approved):
Apart from some comments on comments and variable names, LGTM!
It is nice that we are getting rid of all these annoying spurious edges. IIRC, this is helpful for Sarah and @jerhard 's incremental things.

src/framework/myCFG.ml
sim642: Indeed they are slightly different things. We use `MyCFG.edge` variants for the CFG edge labels, that's what most of Goblint sees. But inside the CFG modules, they always are paired with locations. Since this is the CFG construction, all of this code considers edges in the latter sense.

I would have wanted to even get rid of `addEdge_fromLoc` but there are a few special cases here, where the desired location doesn't come directly from some pattern matching but must be looked up specially.

src/framework/myCFG.ml
sim642: Indeed, but I'll leave it for another PR because I want to consistently refactor all of these things like `MyCFG.getFun`, `Cilfacade.getFun`, `Cilfacade.getdec` and possibly something else I currently forgot. This shouldn't even be intertwined with CFG construction but be a separate pass that builds the helper mappings.

src/framework/myCFG.ml
sim642: There's two different things here: the list of `instr`s in `Instr` and the list of the successors of `Instr`. The "non-empty" is about the `Instr` argument while the `>1` is about the number of successors.

If my study of CIL's CFG preparation is correct, these failures should never happen, but I wanted to make them all really explicit, just in case. Previously they were just mapped over as lists, as if there could be arbitrarily many successors that duplicate the same instructions on the edge, which doesn't really make sense. And indeed it shouldn't really happen, but I want to have all of this code very clear about what CIL gives us and what it doesn't give us.

src/framework/myCFG.ml
sim642: This was already here before, but the "backwards reachable from return" is just the optimal way to compute "nodes from which return is reachable" (one graph search instead of N). Naming it `reachable_from_return` would be misleading because everywhere else we think of reachability to mean forward reachability and nothing is forward reachable from the return node.

src/framework/myCFG.ml
sim642: I'll track this in #290.

PR #276, opened 2021-07-09
Update scripts to use more portable shebang

michael-schwarz:
> Using `#!/usr/bin/env bash` instead of `#!/bin/bash` as shebang makes the scripts usable on systems where bash is not available under `/bin`, e.g., NixOS.

This PR is by the team that added congruences in the practical. I pulled it out of #260 so the diff there becomes more focused.

Any opinions @sim642 @vogler?

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/usr/bin/env bash
 set -e # exit immediately if a command fails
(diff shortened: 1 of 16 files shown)


sim642 (approved):

vogler:
But `#!/usr/bin/env` is always at that location and NixOS doesn't just symlink `/bin/bash`? Weird, but ok.

PR #277, opened 2021-07-09
Replace `Cil.typeOf` with `Cilfacade.typeOf` which uses sensible exceptions

sim642:
The motivation for this is here: https://github.com/goblint/analyzer/pull/251#issuecomment-877068090.
It fixes the scrambled output of a couple of SV-COMP benchmarks from there.

Currently this PR is opened with #251 as base to avoid major conflicts, but we don't have to merge it there, just wait until #251 is in master and change the base for this.

### DIFF
--- a/src/util/cilfacade.ml
+++ b/src/util/cilfacade.ml
@@ -246,6 +246,20 @@ let rec get_ikind t =
 
 let ptrdiff_ikind () = get_ikind !ptrdiffType
 
+
+(** Cil.typeOf, etc reimplemented to raise sensible exceptions
+    instead of printing all errors directly... *)
+
+type typeOfError =
+  | RealImag_NonNumerical (** unexpected non-numerical type for argument to __real__ *) (* TODO: or __imag__? *)
+  | StartOf_NonArray (** typeOf: StartOf on a non-array *)
+  | Mem_NonPointer of exp (** typeOfLval: Mem on a non-pointer (exp) *)
+  | Index_NonArray (** typeOffset: Index on a non-array *)
+  | Field_NonCompound (** typeOffset: Field on a non-compound *)
+
+(* TODO: add exception printer *)
+exception TypeOfError of typeOfError
+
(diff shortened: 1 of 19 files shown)


michael-schwarz:
Would be worth hooking in this implementation in `deriving.ml` too?

Then one would automatically have the correct thing as soon as one  does `open Deriving.Cil` which I guess one has to do to use the derived `to_yojson` and so on anyway...

sim642:
> Would be worth hooking in this implementation in `deriving.ml` too?
> 
> Then one would automatically have the correct thing as soon as one does `open Deriving.Cil` which I guess one has to do to use the derived `to_yojson` and so on anyway...

Except we don't really want the derived `to_yojson` functions for CIL stuff anyway, but the versions in `CilType`, etc, which have a reasonable level of detail. So it seems to me that eventually we'd totally get rid of `Deriving.Cil` anyway.

It might be nice to be able to `open Cilfacade` after `Cil` to automatically use our overridden versions, but that requires some reoganization of `Cilfacade` (and others) altogether to separate the crap that shouldn't be opened.

michael-schwarz (approved):

src/util/cilfacade.ml
```diff
@@ -251,15 +251,27 @@ let ptrdiff_ikind () = get_ikind !ptrdiffType
     instead of printing all errors directly... *)
 
 type typeOfError =
-  | RealImag_NonNumerical (** unexpected non-numerical type for argument to __real__ *) (* TODO: or __imag__? *)
+  | RealImag_NonNumerical (** unexpected non-numerical type for argument to __real__/__imag__ *) (* TODO: CIL's own output forgets __imag__ *)
```
michael-schwarz, line 254: `__imag__` as added by ourselves, we should file a bug against CIL for this.

sim642: I opened the issue https://github.com/goblint/cil/issues/40 and will remove the TODO from here then.

PR #278, opened 2021-07-12
Move assert result output from base to separate assert analysis

sim642:
Inspired by https://github.com/goblint/analyzer/pull/271#discussion_r666743864 and some points in the GobCon agenda.

Currently this PR is opened with #251 as base to work via `EvalInt`, but we don't have to merge it there, just wait until #251 is in master and change the base for this.


vogler:
Looking at the changes, it's a bit ugly that we have so many reg. tests using `--set ana.activated` instead of `--sets ana.activated[+]`. I guess most just add some analysis.
Maybe sth to change in a sep. PR?

src/analyses/apron/octApron.apron.ml
```diff
@@ -97,15 +97,22 @@ struct
...
     if D.is_bot ctx.local then D.bot () else
       begin
         match LibraryFunctions.classify f.vname args with
-        | `Assert expression -> ctx.local
         | `Unknown "printf" -> ctx.local
-        | `Unknown "__goblint_check" -> ctx.local
-        | `Unknown "__goblint_commit" -> ctx.local
-        | `Unknown "__goblint_assert" -> ctx.local
+        (* | `Assert (e, _, change) -> assert_fn ctx e change *)
+        | `Assert (e, _, _) -> assert_fn ctx e true (* TODO: make assert() always change *)
```
michael-schwarz, line 115: Why? Should this not be consistent across base and octApron?

src/analyses/libraryFunctions.ml
```diff
@@ -9,7 +9,7 @@ type categories = [
   | `Malloc       of exp
   | `Calloc       of exp * exp
   | `Realloc      of exp * exp
-  | `Assert       of exp
+  | `Assert       of exp * bool * bool (* exp, should_warn?, change? *) (* TODO: change to non-poly variants to inline record can be used *)
```
michael-schwarz, line 12: Would it not make sense to split this into some sort of `GoblintAssert` that is e.g. the result of classify `__goblint_check`, and keep this completely separate from `Assert` and use `Assert` only for actual C asserts?

Then one could have a separate anlaysis for these "regression test" `__goblint_check` (which is the analysis you started in this PR). For actual `assert` the default behavior would always be `change` and there might be a separate analysis that also does the warnings for C-level `asserts` that do not hold.

michael-schwarz: This would also make it easy to warn for programs that contain `__goblint_check` but are run without this "regression assert" analysis, whereas the behavior changing a lot depending on whether the assert analysis is active or not might be confusing for some.

sim642:
> Looking at the changes, it's a bit ugly that we have so many reg. tests using `--set ana.activated` instead of `--sets ana.activated[+]`. I guess most just add some analysis.
> Maybe sth to change in a sep. PR?

Indeed and we've come across this countless times in the last year adding numerous default analyses. In some cases we probably deactivate a default analysis for a particular test instead. Previously this required a manual listing but since #171 it's also possible to deactivate one, so that's no longer a restriction.

One just needs to go over all these cases and determine, which analyses is the explicit list enabling or disabling compared to the default. Better to leave it as a separate task indeed.

src/analyses/apron/octApron.apron.ml
sim642: Indeed it should be and I think it would make sense for `assert()` to change by default (if needed we can have an option to disable that). I don't see the reason not to, but I wanted to wait until the GobCon discussion of this before making that change and finalizing this PR.

src/analyses/libraryFunctions.ml
sim642: Splitting `assert` from `__goblint_*` might be a bit cleaner indeed, but I'll postpone those changes also to after GobCon to have decided what's our plan with all these things.

Splitting this analysis itself also in two (C `assert()` analysis and our `__goblint_*` assert analysis) might be too much hair splitting. We have the `dbg.regression` option, so the same assert analysis could also warn if that's disabled but `__goblint_*` is called.

src/analyses/apron/octApron.apron.ml
sim642: I changed `assert()` refinement to follow `sem.assert.refine` option which is true by default.

sim642:
After making the `assert()` refine by default, running regression tests reveals which tests actually rely on that behavior and which don't. If contradictory conditions are sequentially checked that shouldn't refine under the old behavior, then the contradictory refinements lead to dead code, which then causes "registered nothing" instead of seeing any desired regression testing output.

Thus the `sem.assert.refine` option is disabled just in those tests. And parts of those could be switched to use `__goblint_check` instead if wanted.

sim642:
> So if we improve precision somewhere (e.g. by enabling interval which is not on by default), we might get lots of spurious fails because it depends on the old semantics of assert but we didn't identify them here.

If we improve precision of something that isn't already marked as `// TODO`, then it would fail anyway and require manual attention.

sim642:
I have now split and rebased this as discussed at GobCon:
1. Assertion check result printing is now done by a separate analysis.
2. All `assert`s in tests have been replaced with `__goblint_check`.
3. `assert` (and `__goblint_assert`) under the hood are now independent of `dbg.debug`: they always warn and refine when `sem.assert.refine` is enabled (which it is by default).
4. Apron analysis now also refines on compatible kinds of assertions. (This will probably allow simplifying some apron tests that previously had to use `if`s to set up desired abstract states, but that's now done here.)
5. Fixes `update_suite.rb` to actually check `__goblint_check`!
6. Adds `goblint.h`, which is included by default, with the function declarations. This is necessary such that CIL would insert appropriate cast for the argument (if it isn't integral).

Enjoy reviewing this PR!

michael-schwarz (approved):

michael-schwarz:
> Enjoy reviewing this PR!

Wasn't so bad given that the changes to the tests all were just mechanical!

Issue #279, opened 2021-07-12
OCaml 4.11.1 not available on Apple M1


vogler:
I then changed it to 4.12.0 and could install everything besides `earlybird`.
Should we make 4.12.0 the default? Who uses `earlybird`?

```console
$ opam --version
2.0.8

$ make setup
Make sure you have the following installed: opam >= 2.0.0, git, patch, m4, autoconf, libgmp-dev, libmpfr-dev
For the --html output you also need: javac, ant, dot (graphviz)
For running the regression tests you also need: ruby, gem, curl
For reference see ./Dockerfile or ./scripts/travis-ci.sh
+ opam init -y -a --bare

User configuration:
  Updating ~/.bash_profile.
[NOTE] Make sure that ~/.bash_profile is well sourced in your ~/.bashrc.

+ opam update

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><>  🐫
[default] no changes from https://opam.ocaml.org
+ opam switch -y create . --deps-only ocaml-base-compiler.4.11.1 --locked
[ERROR] No compiler matching 'ocaml-base-compiler.4.11.1' found, use 'opam switch list-available'
        to see what is available, or use '--packages' to select packages explicitly.
make: *** [setup] Error 5

$ opam switch list-available
# Listing available compilers from repositories: default
# Name                                 # Version                          # Synopsis
ocaml-option-afl                       1
          Set OCaml to be compiled with afl-fuzz instrumentation
ocaml-option-bytecode-only             1
          Compile OCaml without the native-code compiler
ocaml-option-default-unsafe-string     1
          Set OCaml to be compiled without safe strings by default
ocaml-option-flambda                   1
          Set OCaml to be compiled with flambda activated
ocaml-option-fp                        1
          Set OCaml to be compiled with frame-pointers enabled
ocaml-option-musl                      1
          Set OCaml to be compiled with musl-gcc
ocaml-option-nnp                       1
          Set OCaml to be compiled with --disable-naked-pointers
ocaml-option-no-flat-float-array       1
          Set OCaml to be compiled with --disable-flat-float-array
ocaml-option-spacetime                 1
          Set OCaml to be compiled with spacetime activated
ocaml-option-static                    1
          Set OCaml to be compiled with musl-gcc -static
ocaml-options-only-afl                 1
          Ensure that OCaml is compiled with AFL support enabled, and no other custom options
ocaml-options-only-flambda             1
[...]

Issue #184, opened 2021-04-09
Decide what version of OCaml to support & Fix CI accordingly


michael-schwarz:
**Current situation**
We currently pretend to build Goblint with Ocaml 4.07.1 and 4.11.1.
However, we do not really compile with `4.07.1`. We specify it in the Github actions

https://github.com/goblint/analyzer/blob/16497fa06d2059f2c3bde130da0bbb89b5a4d6c3/.github/workflows/tests.yml#L17-L19

but then our CI script uses the `pinned` file, installing the pinned version of OCaml.

**General considerations**
Ideally I was hoping to go for a setup like I have for CIL, where we just do everything via opam and dune.

https://github.com/goblint/cil/blob/60e9c5f77a7267591622b94c9be73555cc2881f1/.github/workflows/tests.yml

This is by far the cleanest and would also give us an indication if some package in an updated version is incompatible with Goblint such that we can put a constraint on the maximum version number of some package. Additional advantage is that testing multiple versions of OCaml is a breeze.
Downside is that we would not have reproducible builds as the version of dependencies might change between runs.

**Which OCaml version to support**
We needed 4.07.1 at some point for earlybird, but earlybird now supports later versions. So maybe we can drop support for 4.07.1 now? However, I would still want to keep compatibility with some older versions (e.g. 4.09.1 now), because in my experience it sometimes takes time for e.g. the VSCode plugin (and the language server) to be adapted to a new version of merlin and it is annoying when such things break but you can't go back because Goblint uses new features.

sim642:
> Downside is that we would not have reproducible builds as the version of dependencies might change between runs.

As I suggested in https://github.com/goblint/analyzer/pull/174#discussion_r609915934, we could have a separate CI job that doesn't do multiple versions but does use the lock file. That would at least give reproducible builds with the version of OCaml that we have locked.

---

In terms of the OCaml version, I just remembered that when I tried to use a newer than 4.07 feature in `traces` branch, it didn't work anyway because some ppx dependencies still can only handle 4.07 AST:
https://github.com/goblint/analyzer/blob/e91bdb4b3dfada91c8a729dccc22d5a023b61898/src/analyses/basePriv.mli#L2.
The particular `ppx_import` issue linked in that comment has since been fixed, but I think there was still some other ppx that didn't, so I gave up.

Therefore I'm not really opposed to having 4.07 support as long as the CI nicely checks it.

sim642:
We should probably address this issue before it falls below our radar.

[...]

PR #280, opened 2021-07-12
OCaml 4.11.1 -> 4.12.0

vogler:
Change the default OCaml version to 4.12.0 (setup & CI) since 4.11.1 is not available on Apple M1.
Fixes https://github.com/goblint/analyzer/issues/279.


vogler:
Should we merge this and continue with the general issue in #184?

PR #281, opened 2021-07-12
OCaml's `let*` instead of `ocaml-monadic` ppx

vogler:
I thought we could drop the [ocaml-monadic](https://github.com/zepalmer/ocaml-monadic) dependency since OCaml introduced definable [binding operators in 4.08.0](https://ocaml.org/manual/bindingops.html#let-operator).
However, not having something like `;%bind` makes the code a bit ugly to read.
I only changed `pml_arinc.ml` which compiles - `pml_osek.ml` is the same.

We can close this if there's no better solution.


michael-schwarz:
I merged master into this now such that we can decide to merge or close. I personally don't have strong feelings either way, but I think it is a good idea to reduce the number of old open PRs and issues.

sim642:
> I only changed `pml_arinc.ml` which compiles - `pml_osek.ml` is the same.

Ah, I guess after #735 that doesn't need to be done.

I would be for doing this change since the slight extra ugliness is just in one obscure module and `let* () =` isn't much more code than `;bind`, but it is more standard now, especially since we use `let*` in some other places as well.
The ocaml-monadic dependency could now be dropped from dune-project, etc entirely though.

sim642:
Actually, the `;bind`/`let* ()` could maybe be replaced with an appropriate infix operator definition for `>>` (i.e. a version of monadic bind that ignores the first value). Although `Pml` uses some other weird definition of `>>`, which is completely unrelated, so it's probably easiest to skip this and stay with the changes already made in this PR.

sim642 (approved):

PR #283, opened 2021-07-13
Remove ppx_import

sim642:
ppx_import causes issues here: https://github.com/goblint/analyzer/issues/279#issuecomment-878866381. And we don't really depend on ppx_import (and the derived `to_yojson` implementations): https://github.com/goblint/analyzer/pull/277#issuecomment-877218709.

So this PR does the following:
1. Removes imported CIL types and `to_yojson` derivation. We only cared about the manual implementations there, which are now in `CilType` for more convenient use anyway.
2. Removes imported `ref` type and its derivations. Nothing seems to depend on that?
3. Removes ppx_import dependency.
4. Removes `Deriving.Map.Make`, which additionally provides manually implemented `to_yojson`. This is useless because the only place `Deriving.Map.Make` is used is in `MapDomain` and there we have a different manual implementation of `to_yojson` now anyway.
5. Removes `Deriving` altogether because nothing is left of it.

In a handful of places I replaced a derived CIL `to_yojson` with one based on an already existing manual `show` implementation, particularly for `Cil.location`. This changes the `to_yojson` output of some locations from a JSON object of the location components (file, line) to some kind of string that's in the XML output anyway like that. @keremc, GobView doesn't depend on that, does it? If so, I'll just add a manual implementation of that to match the old behavior.

### DIFF
--- a/src/analyses/basePriv.mli
+++ b/src/analyses/basePriv.mli
@@ -1,5 +1,6 @@
 open Cil
 (* Cannot use local module substitutions because ppx_import is still stuck at 4.07 AST: https://github.com/ocaml-ppx/ppx_import/issues/50#issuecomment-775817579. *)
+(* TODO: try again, because ppx_import is now removed *)
 
 module type S =
 sig
--- a/_tags
+++ b/_tags
@@ -1,5 +1,5 @@
 <src> or <src/**>: include
-true: package(findlib), package(ppx_import), package(goblint-cil.all-features), package(batteries.unthreaded), package(ppx_deriving.std), package(ppx_deriving_yojson), package(ppx_distr_guards), package(qcheck-core.runner)
+true: package(findlib), package(goblint-cil.all-features), package(batteries.unthreaded), package(ppx_deriving.std), package(ppx_deriving_yojson), package(ppx_distr_guards), package(qcheck-core.runner)
 <unittest/**>: include, package(oUnit)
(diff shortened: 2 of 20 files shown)


vogler:
> 2. Removes imported `ref` type and its derivations. Nothing seems to depend on that?

No, it was only required at some point for deriving the rest in `Deriving.Cil`.

src/analyses/basePriv.mli
```diff
@@ -1,5 +1,6 @@
 open Cil
 (* Cannot use local module substitutions because ppx_import is still stuck at 4.07 AST: https://github.com/ocaml-ppx/ppx_import/issues/50#issuecomment-775817579. *)
+(* TODO: try again, because ppx_import is now removed *)
```
vogler, line 3: Just a comment to make this TODO visible in the PR Conversation...

sim642: I might try this again after #280 since it'll upgrade a bunch of other ppx dependencies too.

sim642:
> No, it was only required at some point for deriving the rest in `Deriving.Cil`.

Ahh, yes, makes sense. Didn't notice that since I started removing things from the end.

keremc:
@sim642 Gobview doesn't make many assumptions about the shape of the JSON object, so this shouldn't cause any issues.

sim642:
Thanks for confirming that! I'll merge this now.

Issue #110, opened 2020-10-08
Make assertion code consistent and remove asserting unknown


vesalvojdani:
I want to make some semi-brutal changes to how assertions are handled for our regression tests.

- Instead of exclamation marks, I would use "TODO" or "SKIP" annotations. These will be tracked by the script and we can give a positive warning once such a TODO starts working. I want all our test annotations to be about the program, and goblint failures should explicitly marked. (I already made this change in the test script for the witness branch).
- I want to eliminate UNKNOWN for assertions. The main problem with asserting unknown is that we have some tests that now fail if we make the analysis path sensitive. For each path the assertion is either known to be true or known to be false. I want to only allow no annotation = definitely succeeds and "MAY FAIL". 
- The goblint assertion code behaves differently for regression tests by not improving the value. Without the debug flag, it does refine the value based on the assertion. This is precisely the sort of thing that can be very painful when trying to understand why regression tests behave differently from running it normally.

This is a fairly easy fix that I plan to make soon after we submit the SV-COMP benches, but I thought I would check if there are any opinions about it.

michael-schwarz:
> I want to eliminate UNKNOWN for assertions. The main problem with asserting unknown is that we have some tests that now fail if we make the analysis path sensitive. For each path the assertion is either known to be true or known to be false. I want to only allow no annotation = definitely succeeds and "MAY FAIL".

I would argue against this. We quite often had the situation where the regression tests were passing, but there still was some unsoundness somewhere. Having such `UNKNOWN!` assertions to me seems the best way to express this.
Also, I think these `UNKNOWN!` assertions can be implemented in a way that does not interact badly with path-sensitivity, if this is not already the case.

vesalvojdani:
Can you think of any example where we really need uncertainty in both directions? Most of the time, we just want to make sure goblint isn't certain that some value is still the same when some other indirect access or thread could change it. For these cases, it suffices to say that the assertion may fail. I've seen only a few real examples, like join after thread creation, where I'd actually need to make sure goblint doesn't take the value from either branch. I would rather handle these rare cases by explicitly doing a new non-det branch in the test file and make two assertions that may fail. 

I do not think you can easily implement "unknown" assertions in a way that interacts well with path-sensitivity. The unknown assertion is making claims over multiple paths, so it would have to be handled within the path-sensitivity functor where one would have to rejoin split paths just for "asserts" or collect generated warnings from different paths and check if all agree. 

[...]

PR #286, opened 2021-07-14
Undo update_suite ignoring precision difference for UNKNOWN

sim642:
In #196 among many other changes we merged this innocent-looking change: https://github.com/goblint/analyzer/commit/0cb2c9c42c660b2c0b6889581b659ef7d213d105. This PR reverts that.

During #278 I looked over all the UNKNOWN asserts to see if making `assert` refine the state would cause problems. I saw cases where problems should have appeared but didn't due to the reverted change, which completely silenced all the cases where we still have an `UNKNOWN` annotation but suddenly output success instead. I only realized this when I used regtest (which awfully does similar expected comparisons within Goblint...) and saw different output.

Because we haven't completely gone through the entire regression test suite to recategorize each `UNKNOWN` to either `UNKNOWN!` or `TODO`, then we still have cases where `UNKNOWN` is used with the meaning that it must be unknown. Therefore the script should report if we're suddenly unsoundly reporting success or fail. See #110.

### DIFF
--- a/scripts/update_suite.rb
+++ b/scripts/update_suite.rb
@@ -166,8 +166,6 @@ regs.sort.each do |d|
         debug = true
         if obj =~ /FAIL/ then
           tests[i] = "fail"
-        elsif obj =~ /UNKNOWN!/ then
-          tests[i] = "unknown!"
         elsif obj =~ /UNKNOWN/ then
           tests[i] = "unknown"
         else
@@ -425,10 +423,8 @@ File.open(theresultfile, "w") do |f|
         end
       }
       case type
-      when "unknown"
-        check.call ["deadlock", "race", "fail", "unknown", "noterm", "term", "warn", "success"].include? warnings[idx]
-      when "deadlock", "race", "fail", "noterm", "unknown!", "term", "warn"
-        check.call warnings[idx] == type.tr('!', '')
+      when "deadlock", "race", "fail", "noterm", "unknown", "term", "warn"
+        check.call warnings[idx] == type
       when "nowarn"
         check.call warnings[idx].nil?
       when "assert"
--- a/tests/regression/00-sanity/19-update_suite.c
+++ b/tests/regression/00-sanity/19-update_suite.c
@@ -1,0 +1,10 @@
+// SKIP (just for manually testing that update_suite works)
+#include <assert.h>
+
+int main() {
+  int x = 42;
+  // Should fail with: Expected unknown, but registered success
+  assert(x == 42); // UNKNOWN
+
+  return 0;
+}
--- a/tests/regression/31-ikind-aware-ints/13-intervals-large.c
+++ b/tests/regression/31-ikind-aware-ints/13-intervals-large.c
@@ -9,8 +9,7 @@ int main(){
     if(x > 18446744073709551612ull){
         a = 1;
     }
-    // The following line should succeed, but is unknown for now
-    assert(a); // UNKNOWN
+    assert(a);
 
     unsigned long long y = x + 4;
     // Unsigned overflow -- The following assertion should succeed, but is unknown for now


vogler:
> I only realized this when I used regtest (which awfully does similar expected comparisons within Goblint...) and saw different output.

I agree with 'awful' concerning duplication, but IMO it'd be better for goblint to handle these things instead of some ruby script - ideally we'd get rid of it in favor of an OCaml executable or some minimal shell script (`parallel`) with goblint doing most of the work.

michael-schwarz:
@jerhard  and I were tossing this back and forth over lunch and something we discussed at some point was using something along the lines of `__goblint_check(a == 42, __GOBLINT_UNKNOWN)` or `__goblint_check(x ==17, __GOBLINT_SUCCESS)` in ehe regression tests instead of having an `assert(...) //UNKNOWN`.

Then, the expected result is immediately obvious to Goblint itself, and Goblint can either output things for all asserts or only those where something failed without having to somehow do a regex on the comments.

Also, once could use the syntactic search to identify those places where `__goblint_check(...)` is called that are unreachable and then also warn there that the assert is unreachable. (These are the cases where warnings are missing in `./regtest.sh` vs the ruby script).

The question that is still open is how to integrate warnings different from asserts into this setting.

Opinions?

vogler:
> Also, once could use the syntactic search to identify those places where `__goblint_check(...)` is called that are unreachable and then also warn there that the assert is unreachable. (These are the cases where warnings are missing in `./regtest.sh` vs the ruby script).

Good point to keep in mind.
This (together with the rest?) could be done as post-processing in `finalize` of the assert-analysis.

> using something along the lines of `__goblint_check(a == 42, __GOBLINT_UNKNOWN)` or `__goblint_check(x ==17, __GOBLINT_SUCCESS)` in ehe regression tests instead of having an `assert(...) //UNKNOWN`.

Pro:
- type-safe (constant instead of comment)
- available in CIL instead of having to read the comment from source

Con:
- can't compile and execute it anymore without including another file?
- `assert` is easily understood and portable

sim642:
Tbh, I don't really like the idea of this regression testing check (`dbg.regression` for regtest) being inside Goblint itself at all. It's ideologically wrong to have the testing logic right in the middle of Goblint (which would be both the tester and the testee simultaneously) instead of being independent. And having to write `__goblint_check(x == 17, __GOBLINT_SUCCESS)` is so much more verbose than a standard `assert(x == 17)`.

But that's all beside the point for this issue of update_suite not reporting possible unsoundness cases. If we want to discuss completely redoing the regression testing architecture, it's best to have another issue for that. And that won't solve the fact that we still have ambiguous uses of `UNKNOWN`.

vogler:
> And that won't solve the fact that we still have ambiguous uses of `UNKNOWN`.

https://github.com/goblint/analyzer/issues/288#issuecomment-879877330
If we did something like this, we'd at least see those diffs.

```console
$ grep -E 'UNKNOWN$' -r tests/regression | wc -l
313
```
All these need to be replaced with `UNKNOWN!` xor `TODO`?

sim642:
> All these need to be replaced with `UNKNOWN!` xor `TODO`?

This is the idea of #110, yes. Although we also have some cases like this:
https://github.com/goblint/analyzer/blob/cd6cf51463f61d57848bfaafd5b87c79449065c9/tests/regression/01-cpa/32-earlyglobs.c#L7-L9
It's not `UNKNOWN!` because in the concrete semantics, it is known. It's not `TODO` because it shouldn't be fixed, because it's the whole point of `earlyglobs`. So probably one needs a third `UNKNOWN!`-like category for things which should remain unknown but only by our own choice of imprecision, not by concrete semantics.

sim642:
This revealed an issue with a congruences test now: https://github.com/goblint/analyzer/pull/260#issuecomment-883235096.

Issue #254, opened 2021-06-16
should not `eval` unread globals


vogler:
Why do we `eval` and thereby `add_infl` `earlyglobs` variables which are written but not read?
In the example, `g2` is read, so the `eval` is fine, but afterwards in the trace it also evaluates `g1` which adds an unwanted dependency.
Where does this come from? sync/mutex stuff? Can it be removed?
This might cause a lot of unnecessary recomputation.

~~~c
// ./goblint -v add_infl.c --enable exp.earlyglobs --enable ana.int.interval --html --trace sol2
int g1, g2;

int main() {
  g1 = g2 + 1;
}
~~~

~~~
%%% sol2: solve node 1 "g1 = g2 + 1;" on add_infl.c:4, called: false, stable: true
%%% sol2: init node 1 "g1 = g2 + 1;" on add_infl.c:4
%%% sol2: add_infl node 1 "g1 = g2 + 1;" on add_infl.c:4 node 2 "return (0);" on add_infl.c:5
%%% sol2: eval node 2 "return (0);" on add_infl.c:5 ## g2 on add_infl.c:1
%%% sol2: simple_solve g2 on add_infl.c:1 (rhs: false)
%%% sol2: init g2 on add_infl.c:1
%%% sol2: add_infl g2 on add_infl.c:1 node 2 "return (0);" on add_infl.c:5
%%% sol2: eval node 2 "return (0);" on add_infl.c:5 ## g1 on add_infl.c:1
%%% sol2: simple_solve g1 on add_infl.c:1 (rhs: false)
%%% sol2: init g1 on add_infl.c:1
%%% sol2: add_infl g1 on add_infl.c:1 node 2 "return (0);" on add_infl.c:5
~~~

---

The example above is extracted from this which has the same issue (also with `--sets exp.privatization none`):

~~~c
// ./goblint -v side_cycle.c --enable exp.earlyglobs --enable allfuns --enable ana.int.interval --trace sol2
#include <pthread.h>
pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;

int g1, g2;

void f() {
  pthread_mutex_lock(&mutex);
  g2 = g1 + 1;
  pthread_mutex_unlock(&mutex);
}

int main() {
  pthread_mutex_lock(&mutex);
  g1 = g2 + 1;
  pthread_mutex_unlock(&mutex);
}
~~~

vogler:
[...]

PR #289, opened 2021-07-14
Fix base `invariant` worsening precision via ambiguous pointer

### LINKED ISSUE #285: Base `invariant` can worsen precision
During #278 the following bug with base's `invariant` was revealed, where refining actually worsens precision:
https://github.com/goblint/analyzer/blob/6ebe8168c19687c6227ac41127f1e7c02a46e4d9/tests/regression/27-inv_invariants/09-invariant-worsen.c#L5-L21

Branching on `*x != 0` where x is `{&a, &b}` sets them both unnecessarily to `Not {0}` instead of keeping their previous exact values 1 and 2 respectively.

sim642:
Closes #285.

Base `invariant` handles the refinement of lvalues monolithically:
https://github.com/goblint/analyzer/blob/dd4dee67a81961aff191f9d70a1c00b8fecd0fcd/src/analyses/base.ml#L1543-L1549

In the problematic example, it first evaluates `*x` to `Not {0}` (join of 1 and 2). The invariant expression `*x != 0` gives for `*x` the restriction to also `Not {0}`. These are met without change. Finally `*x` is `set` to the value `Not {0}`, which is written through the pointer to `a` and `b`, making them less precise. No meet happens on a per target address basis.

Fixing this in `invariant` directly would require duplicating `set`, which first evaluates `*x` to the address set `{&a, &b}` and then updates each one individually with the given value `Not {0}`. Since `set` has an `effect` argument (which is false for `invariant`), I changed this per-address met to happen inside `set` conditioned on that. This happens on just `ValueDomain` elements, so no possibly problematic meeting of `Priv` components is happening. Hopefully this is right!

This can easily be done in `set` because it reads variables before using `VD.update_offset` on them and writing. Following this change fixing #254 gets another condition when it still needs to read before writing globals: when coming from `invariant`.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1067,6 +1067,14 @@ struct
               M.warn ("Cilfacade.typeOfLval failed Could not obtain the type of "^ sprint d_lval (Var x, cil_offset));
               lval_type
       in
+      let update_offset old_value =
+        let new_value = VD.update_offset a old_value offs value lval_raw ((Var x), cil_offset) t in
+        if not effect then
+          (* without this, invariant for ambiguous pointer might worsen precision for each individual address to their join *)
+          VD.meet old_value new_value
+        else
+          new_value
+      in
       if M.tracing then M.tracel "setosek" ~var:firstvar "update_one_addr: start with '%a' (type '%a') \nstate:%a\n\n" AD.pretty (AD.from_var_offset (x,offs)) d_type x.vtype D.pretty st;
       if isFunctionType x.vtype then begin
         if M.tracing then M.tracel "setosek" ~var:firstvar "update_one_addr: returning: '%a' is a function type \n" d_type x.vtype;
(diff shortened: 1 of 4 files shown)


sim642:
The logic in `invariant` also does the following on the met value:
```ocaml
if is_some_bot v then raise Deadcode
```

This currently isn't done in the `meet` inside `set` for a single address. It would be wrong to directly do it, but it would be possible to gain extra precision from something similar: if the refined value doesn't hold for a particular address, then `invariant` could indirectly refine the address set to not point to such address any more. I've added that as a TODO test because it's not so simple to achieve: probably would have to make the per-address update logic explicit inside `invariant`.

michael-schwarz:
Should we in this step also unify `~effect` and `~chnage_array`? It seems like they do the same thing.

michael-schwarz (approved):

sim642:
> Should we in this step also unify `~effect` and `~change_array`? It seems like they do the same thing.
> Maybe we should just name it `~guard`?

I thought about renaming and flipping `~effect` to `~invariant`, which is what I used for `Priv.write_global` to indicate that it's a "write" coming from `invariant`. I think I'll do that and also use it instead of `~change_array`.

jerhard:
>Since set has an effect argument (which is false for invariant), I changed this per-address met to happen inside set conditioned on that

Ok, so the soundness of this relies on the usual join with the old state and the current state if the cardinality of the address set was greater than one:
 https://github.com/goblint/analyzer/blob/666647cdc62916ebd74731a4f387c3cf70d3e104/src/analyses/base.ml#L1175

Does it make sense to mention this in the comment in the `update_offset` function you define locally? So one gets not confused about the soundness.

sim642:
The soundness of this should simply be that it's sound for `invariant` to meet with the previous value. That join of old and new isn't just for this use case but any ambiguous pointer writes in general.

I saw that join too when looking where to easiest make this fix and actually found it quite confusing in general (not just the `invariant` case I added). Because on the previous line it folds over all the possible addresses and assigns to them one by one after one another instead of happening "in parallel". Conceptually what makes more sense is that all of the possible writes are done on `st.cpa` and those results are instead joined together. I guess this is some kind of optimization to just do one join instead of `AD.cardinal lval` joins and relies on the non-relationality (and maybe something else) of `CPA`.

But the soundness of this `invariant` improvement I think is independent of how those results are joined, as long as that's in itself sound.

Issue #259, opened 2021-06-17
OctApron refactorings


sim642:
There's a bunch of things to refactor in `OctApron` and `OctApronDomain`:
- [x] Remove `cil_exp_to_lexp`, `cil_exp_to_apron_linexpr1` and `cil_exp_to_apron_linecons`?
      We already have `cil_exp_to_cil_lhost`, which actually converts a `Cil.exp` into `Apron.Texpr1.expr` (tree expressions). I don't understand why we also need the first functions that convert `Cil.exp` into `Apron.Linexpr1.t` (linear expressions, i.e. list of variables with coefficients to be summed). Apron can already do linearization of tree expressions AFAIK. Is our conversion somehow supposed to be better than Apron?
- [x] Use `Apron.Var.t` for interacting with `OctApronDomain` instead of an arbitrary mix of `string` and `varinfo`.
- [ ] Figure out where to apply `ana.octapron.vars` filtering. Right now it's half-baked into `OctApronDomain`, but this doesn't make much sense IMO. Because if we're going to be adding all sorts of auxiliary variables (e.g. `g#in`, `g#out`, `g#prot`, `g#unprot` etc) corresponding to a single `varinfo`, then the simple string filtering doesn't actually apply to them. Seems to me like this should be pulled up into the analysis, which actually knows what is what.
- [x] Use bigints instead of `int` for the conversions from CIL expressions to Apron ones. Apron uses GMP, so it should be able to deal with anything just fine. We don't have to go through `int`.

sim642:
I did the first one on `traces-relational` and it seems fine on the existing octApron tests at least. If there ever comes a need I guess we can revive that linear conversion.

Also, it's probably better to do all these on `traces-relational` directly because things have and will deviate there more. It's easier to refactor all the priv stuff incrementally than later in a massive conflict.

sim642:
I also did the conversion of the `OctApronDomain` API to `Var` (as much as possible) and removed the chosen variable stuff while at it, because I didn't want to introduce duplicate functions (one version checks chosen, other ignores it). There's already a large number of functions with all the imperative `_with` variants.

sim642:
I'll consider this done after merging #291. The option `ana.octapron.vars` was removed there and not added back but once we need variable filtering/clustering, we'll have to think again how to best make it compatible with everything else.

Issue #273, opened 2021-07-07
Add global option to assume no overflows for signed types


michael-schwarz:
There should be some global switch that assumes no overflow for signed types for all int domains.
This would probably subsume `ana.int.wrap_on_signed_overflow` and `ana.octapron.no_signed_overflow`.

E.g., we could have `sem.int.signed_overflow` with values `'assume_none' | 'assume_top' | 'assume_wraparound'`. This would also make it clear that going to top still is an assumption and does not actually cover everything a program that has UB might do.

This would allow for better performance e.g. in SV-Comp where for tasks where reachability is checked, one may assume that there is no UB and therefore no signed overflows.

Should also not be hard to implement as we always have the `ikind` of ints in the `IntDomTuple` anyway.

sim642:
Tbh, I'm skeptical about `ana.octapron.no_signed_overflow` in general because it only does the bound check at the outermost level (e.g. entire right-hand side of an assignment), but not any of the intermediates, which may overflow during actual execution, but don't overflow when simplified by Apron, e.g. `y = (x + x) - x`.

Having a general option under `sem` is sensible, but we'll have to see if octApron should actually be doing any of that or leaving overflow checks to just normal intervals in base.

jerhard:
>Tbh, I'm skeptical about ana.octapron.no_signed_overflow in general because it only does the bound check at the outermost level (e.g. entire right-hand side of an assignment), but not any of the intermediates, which may overflow during actual execution, but don't overflow when simplified by Apron, e.g. y = (x + x) - x.

Could one implement an overflow check for every subexpression?

sim642:
> Could one implement an overflow check for every subexpression?

That's not how octApron evaluates expressions though. It doesn't recurse into subexpressions but just gives the entire expression to Apron to evaluate with its best relational ability.

Also, I think there's no point in implementing subexpression overflow checks into octApron because base and the interval domain already do it and I think there's no extra precision to be gained. After #251, base queries `EvalInt` on every expression anyway, so if some subexpression has a more precise value according to octApron (which would rule out overflow), then base will know about it and the tighter interval domain bounds could avoid going into the overflow case there.

sim642:
I suppose the bigger problem is that octApron doesn't actually know when the overflow case happens in the interval domain in base. So octApron wouldn't be invalidating some constraints it knows, which I guess is what you want from `assume_top`?

[...]

PR #291, opened 2021-07-15
Relational traces implementation, octApron fixes & refactoring

sim642:
Although the work on relational traces isn't yet complete, it would make sense to already merge the currently working stuff into master. The `traces-relational` branch contains lots of general fixes and refactoring for the octApron analysis, so it would be best to avoid it from deviating from `master` too much.

### Changes
1. Refactoring of `OctApronDomain` (#259).
   1. Replace `Apron.Linexpr1` based stuff with more general `Apron.Texpr1`.
   2. Consistently use `Apron.Var` for function arguments.
   3. Remove `ana.octapron.vars` (see below).
   4. Reorganize an simplify all `OctApronDomain` functions accordingly.
2. Add `OctApronDomain` version `D2` which supports operations on abstract values with different Apron environments.
3. Add utilify modules for managing metadata (e.g. `varinfo`) for string-based Apron variables. This required the removal of `ana.octapron.vars` since auxiliary Apron variables have no one-to-one mapping to program variables.
4. Fix some unsoundness bugs in octApron (#248).
5. Return intervals from `EvalInt` query (thanks to #251).
5. Add globals direct globals support in singlethreaded mode.
6. Add globals support in multithreaded mode using `OctApronPriv`, with two implementations: `protection` and `mutex-meet`. These heavily use auxiliary Apron variables and their metadata mapping. The interface for these is very similar to #183.


src/cdomains/apron/octApronDomain.apron.ml
```diff
@@ -56,504 +314,466 @@ struct
...
+          (* Signed overflows are undefined behavior, so octagon goes to top if it might have happened. *)
+          top_env (A.env oct)
+      else
+        new_oct
+    else if check_max <> `True || check_min <> `True then
+      (* Unsigned overflows are defined, but for now
+          the variable in question goes to top if there is a possibility of overflow. *)
+      forget_vars oct [Var.of_string v.vname]
+    else
+      new_oct
```
sim642, line 461: I'm really skeptical about this function because of https://github.com/goblint/analyzer/issues/273#issuecomment-875369259. I'm not sure it's worth handling overflows in octApron at all if it's incomplete like that.

Its signature also doesn't fit the rest because it's the only function here that requires a `varinfo` instead of `Apron.Var.t`. So I've really wanted to refactor this but I'm unsure what to do about this overflow stuff.

michael-schwarz:
Tbh, I'm having a hard time reviewing this as it is huge and I do not have intimate knowledge of Apron.
But I think we should go ahead and merge this PR to ensure it does not go stale.

If there are any issues, we can probably address them in separate new PRs. WDYT?

michael-schwarz (approved):

michael-schwarz:
After merging this, Goblint no longer compiles in my switch without `Apron` installed.
Does `octApronDomain.ml` not need the same treatment as all other apron things?
(So have a `-no-apron-` and an `-apron-` version?)

sim642:
You mean `octApronPriv.ml`? I think it might indeed because it depends on `Apron` and the others. Sorry I forgot about it! Feel free to commit that fix directly on master then.

michael-schwarz:
> octApronPriv.ml
Yeah, this is the one I meant.

Unfortunately, we will also need two different versions of `defaults.ml` as `ana.path_sens` contains `octApron` which Goblint will complain about if it does not find it at runtime.

michael-schwarz:
> Unfortunately, we will also need two different versions of defaults.ml 

I fixed this now by having `octApron` add itself to the list of path-sensitive analyses. Maybe not the nicest thing, but the cleanest one I could think of.

sim642:
> I fixed this now by having `octApron` add itself to the list of path-sensitive analyses. Maybe not the nicest thing, but the cleanest one I could think of.

Good point. I think that's fine for now, but I think the path-sensitivity situation also needs some rethinking. Because I don't really see the point of `ana.path_sens` if path-sensitivity must really be implemented by each analysis' `should_join` anyway. Right now it can be quite confusing if you override `should_join` but it doesn't work just because you forgot to also add the analysis to `ana.path_sens`.

PR #294, opened 2021-07-20
Clean up `MyCFG` and refactor CIL helper mappings

### LINKED ISSUE #290: Refactor CIL helper mappings to be consistent
_Originally posted by @sim642 in https://github.com/goblint/analyzer/pull/274#discussion_r670213389_

We have a bunch of helper mappings for CIL stuff that are defined and populated all over the place inconsistently:
- [x] `MyCFG.getFun` looks up for `node` the `fundec` it's in.
- [x] `Cilfacade.getFun` looks up for `string` the `fundec` with that name.
- [x] `Cilfacade.getdec` looks up for `varinfo` the `fundec` with that as function variable.
- [x] `InvariantCil.var_find_fundec` looks up for `varinfo` the `fundec` it's declared in (as local or formal).
- [x] `InvariantCil.var_find_original_name` looks up for `varinfo` the `string` containing its original name in the input file (before CIL renamed it).

And possibly something else I currently forgot.

These should be organized to one clear place and constructed in consistent manner instead of being intertwined with CFG construction or each one doing a separate `Cil.iterGlobals` (or even worse, doing that on each lookup without storing a mapping).

sim642:
### Changes
1. Clean up `Node`.
2. Extract `Edge` module from `MyCFG` (like `Node` was previously done).
3. Move large implementations from `MyCFG` to new `CfgTools` module and just keep declarations in `MyCFG`.

    I decided to do this after working on `MyCFG.createCFG` in #274. Every tiny change in the CFG construction implementation caused most of Goblint to be recompiled because `MyCFG` is opened all over the place. This was completely unnecessary because none of the opened type definitions actually changed.

    By moving all the large implementations into a separate module, recompilation times are much improved if those are changed. One day one might even think about replacing all `MyCFG.node` (and `MyCFG.edge`) uses with `Node.t` (and `Edge.t`) directly, but I decided not go for so many changes for now.

4. Refactor CIL helper mappings consistently and move all of them to `Cilfacade` (closes #290).

    Now they all use `Hashtbl` (some used `Map` before), are all constructed with `Lazy` (one was constructed in `MyCFG.createCFG`, one was constructed using manual lazy construction using auxiliary mutable variables, one used `iterGlobals` on each call).

    I also tried to give them longer descriptive names, but I'm not too happy about them, so feel free to suggest even clearer names.

5. Deduplicate CFG to `.dot` printing code.


src/util/cilfacade.ml
```diff
@@ -410,3 +380,89 @@ let countLoc fn =
...
+    let h = StringH.create 111 in
+    iterGlobals !current_file (function
+        | GFun (fd, _) ->
+          StringH.replace h fd.svar.vname fd
+        | _ -> ()
+      );
+    h
+  )
+
+let find_name_fundec name = StringH.find (Lazy.force name_fundecs) name (* name argument must be explicit, otherwise force happens immediately *)
```
michael-schwarz, line 433: Maybe `find_fundec_by_name` and `find_fundec_by_varinfo` make it clearer that these things return a `fundec`?

sim642: There's also `find_scope_fundec` with the type `varinfo -> fundec option` right now, but that finds the surrounding `fundec` for a local variable's `varinfo`, while `find_varinfo_fundec` finds the `fundec` which has the `varinfo` as its own name. This is the especially confusing pair of functions since the former could also be called `find_fundec_by_varinfo` or whatnot to indicate that the scope is a `fundec`.

I thought about being even more verbose like `find_function_varinfo_fundec` and `find_local_varinfo_fundec` etc, but it gets so extremely verbose to use. At least compared to `getFun` or `getdec` that we had before, so I wasn't sure how far to go with it.

src/util/cilfacade.ml
```diff
@@ -410,3 +380,89 @@ let countLoc fn =
   let res = Hashtbl.length locs in
   Hashtbl.clear locs;
   res
+
+
+
+module StmtH = Hashtbl.Make (CilType.Stmt)
```
michael-schwarz, line 386: All these things work under the assumption that we do not add new globals to `current_file`, while the old implementations did not assume this. I think currently we do not rely on creating new globals, but we should document this here.

src/util/cilfacade.ml
michael-schwarz: > There's also `find_scope_fundec` with the type `varinfo -> fundec option`

Maybe just `find_scope` would suffice here, since in CIL there is only function scope and globals?
I guess this would somewhat resolve the confusion.

src/framework/control.ml
```diff
@@ -154,9 +154,9 @@ struct
 
       (* Adding the state at each system variable to the final result *)
       let add_local_var (n,es) state =
-        let loc = Tracing.getLoc n in
+        let loc = UpdateCil.getLoc n in
```
michael-schwarz, line 157: There should be a comment on why this is `UpdateCil.getLoc` here.

src/framework/edge.ml
```diff
@@ -0,0 +1,60 @@
...
+let pretty_kind () = function
+  | Assign (lv,rv) -> dprintf "Assign"
+  | Proc (_  ,f,ars) -> dprintf "Proc"
+  | Entry f -> dprintf "Entry %s" f.svar.vname
+  | Ret (r,fd) -> dprintf "Ret"
+  | Test (p,b) -> dprintf "Test"
+  | ASM _ -> text "ASM"
+  | Skip -> text "Skip"
+  | VDecl _ -> text "VDecl"
+  | SelfLoop -> text "SelfLoop"
```
michael-schwarz, line 60: If it unused should it be removed?

src/framework/cfgTools.ml
```diff
@@ -0,0 +1,540 @@
...
+
+let do_the_params (fd: fundec) =
+  (* This function used to create extra variables, but now it just sets the
+   * vdecl to -3, lovely... *)
+  let create_extra_var (p: varinfo): unit =
+    match p.vtype with
+    | TPtr (t,_) -> p.vdecl <- {p.vdecl with line = -3 }
+    | _ -> p.vdecl <- {p.vdecl with line = -3 }
+  in
+  List.iter create_extra_var fd.sformals
```
michael-schwarz, line 18: In the long run, I would want to get rid of this setting locations to -3 to signal parameter. Maybe it would be nicer to have a hashtable of parameter `vid`s?

Also, is `line = -4` with meaning `Context` ever used anywhere.

michael-schwarz: In the mean time, why does this distinguish by `vtype` and then does nothing with it?

src/util/cilfacade.ml
sim642: They did though, all of them except `Cilfacade.getFun` constructed some kind of mapping once at the beginning, either during `MyCFG.createCFG`, using manually written laziness of `getdec` or `lazy`.

They all iterate over `GFun`s, so even if someone added those later, there would be bigger problems first: like no CFG for them, because `MyCFG.createCFG` works the same way.

src/framework/control.ml
sim642: I added a comment although I'm highly suspicious that there are other places where this needs to be done as well, since this only affects locations in the XML result, not anything else. So some of the other `Node.location` calls which bypass the `UpdateCil` override mapping might still end up in some terminal output.

But given the discussion in #290, it might be a better to change some other things altogether to remove the need for this distinction (and use overrides everywhere).

src/framework/edge.ml
sim642: I might do that once I figure out the similar situation in `Node`, where at least four different output formats (some with multiple copy-pasted implementations) exist:

https://github.com/goblint/analyzer/blob/ef44a3c988cee5a930f984d180b6f96db773e57a/src/framework/node.ml#L16-L26
https://github.com/goblint/analyzer/blob/ef44a3c988cee5a930f984d180b6f96db773e57a/src/framework/analyses.ml#L47-L51
https://github.com/goblint/analyzer/blob/ef44a3c988cee5a930f984d180b6f96db773e57a/src/framework/cfgTools.ml#L390-L394

I could move all the variants together into `Node` there to avoid identical duplicate implementations but it still seems excessive. Seems like we should have one consistent format, but I don't know which one it should be. Maybe the last CFG node naming one is the best: just `sid` for `Statement`, `f()` for `FunctionEntry f` and `return of f()` for `Function f`.
Others seem a bit more confusing, like `call of f()` meaning the return node.

Also other `Statement` ones include the statement itself, which might sometimes be useful in terminal output instead of just a node ID that needs to be looked up in the HTML output. But statement printing is a bit weird as well: for `Instr`s it's just one line for sure, but for `If` that's actually the whole if statement with both branches printed by `Cil`, which is multiline and super verbose.
And tbh, I've always found statements on nodes extremely confusing because what the hell do they even mean? Statements correspond to the edges of the CFG, not nodes between them. The node itself isn't executing any statement. So it's unclear whether the statement printed for a node corresponds to an edge before or after that node.
I believe the way our CFG is constructed, the statement in the node is the same as the edge going out of it. But then during analysis, the transfer function has `ctx.node` which corresponds to the node being evaluated while `ctx.prev_node` is the one `ctx.local` corresponds to. So actually the statement in `prev_node` is the one being evaluated and the statement in `node` has nothing to do with the current transfer function.

src/framework/cfgTools.ml
sim642: > Also, is `line = -4` with meaning `Context` ever used anywhere?

At some point I looked into it and also couldn't find it being used.

> In the long run, I would want to get rid of this setting locations to -3 to signal parameter. Maybe it would be nicer to have a hashtable of parameter `vid`s?

Indeed, these magic constants are spread around all over the place. No surprise that -4 has been removed from some places but still exists in others. And instead of a hashtable of `vid`s, it's much nicer to have a hashtable of `CilType.Varinfo` to let that take care of the identity automatically, not spreading the `vid` usage everywhere.

Actually, this could be related to `Cilfacade.find_scope_fundec`. Namely, if one generalizes that to not just store `fundec`s as values but variants like `Formal of fundec | Local of fundec | Global`, then the grouping info could be done using the same mapping that stores the scopes. But I'll leave the grouping cleanup for a separate PR.

src/framework/edge.ml
sim642: Since there's no clear best way to output nodes, I just moved all the possible variants into `Node` after all. But at least gave them sensible names and documentation.

The unused `Edge` thing is also removed now.

src/util/cilfacade.ml
sim642: I added documentation comments to all those functions, that should clarify the differences.

PR #295, opened 2021-07-21
Add column to location

sim642:
Compatibility for https://github.com/goblint/cil/pull/42.

### TODO
- [x] Update goblint-cil opam pin after it's merged there.

### DIFF
--- a/goblint.opam
+++ b/goblint.opam
@@ -48,5 +48,5 @@ build: [
 ]
 dev-repo: "git+https://github.com/goblint/analyzer.git"
 pin-depends: [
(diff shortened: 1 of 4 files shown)


michael-schwarz:
LGTM! I think we can update the pin and merge this.

sim642:
I'll go ahead and do that, but it's still open how we want to use and expose this extra information. Because I didn't `Cil.d_loc` to include the column number in the output, it's not outputted anywhere. So there's still things to be done, but I'll open a separate issue to discuss that.

PR #302, opened 2021-07-25
Fix unsoundness of Modulo Operations involving negative numbers #301

### LINKED ISSUE #301: Unsoundness of modulo operations involving negative numbers
Most of our domains seem to be unsound w.r.t. modulo operations involving negative numbers:

~~~C
#include "assert.h"
int main() {
    int x = -1;
    int z = x % 5;
    assert(z == -1);
}
~~~
For GCC and Clang the value of `z` is `-1`. Our domains have for `z` :
- `4` for `DefExc` 
- `bottom` for Interval. 

C99 Standard: http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf
> If the quotient a/b is representable, the expression (a/b)*b + a%b shall equal a.

If we set `b=5` and `x=a=-1`, also according to the standard `z` needs to be -1.

michael-schwarz:
Replaces incorrect usage of `Big_int_Z.mod_big_int` which computes the Euclidian Modulus with `Big_int_Z.sub_big_int a (mul b (div a b))` which computes the remainder (as does C and the `mod` operation on normal ints).

This necessitated a fix to the `normalize` function in the congruence domain, as the `c` in `c+mℤ` is no longer guaranteed to be non-negative leading to two different representations of the same congruence class (one with `c` and one with an equivalent `c'` of different sign).

Closes #301

### DIFF
--- a/src/util/intOps.ml
+++ b/src/util/intOps.ml
@@ -21,6 +21,9 @@ sig
   val sub : t -> t -> t
   val mul : t -> t -> t
   val div : t -> t -> t
+
+  (* This should be the remainder, not the Euclidian Modulus *)
+  (* -1 rem 5 == -1, whereas -1 Euclid-Mod 5 == 4 *)
   val rem : t -> t -> t
 
   (* Bitwise *)
--- a/src/util/intOps.ml
+++ b/src/util/intOps.ml
@@ -192,10 +192,14 @@ struct
   let mul = Big_int_Z.mult_big_int
 
   (* If the first operand of a div is negative, Zarith rounds the result away from zero.
-    We thus always transform this into a divison with a non-negative first operand.
+     We thus always transform this into a division with a non-negative first operand.
   *)
   let div a b = if Big_int_Z.lt_big_int a zero then Big_int_Z.minus_big_int (Big_int_Z.div_big_int (Big_int_Z.minus_big_int a) b) else Big_int_Z.div_big_int a b
-  let rem = Big_int_Z.mod_big_int
+
+  (* Big_int_Z.mod_big_int computes the Euclidian Modulus, but what we want here is the remainder, as returned by mod on ints
+     -1 rem 5 == -1, whereas -1 Euclid-Mod 5 == 4
+  *)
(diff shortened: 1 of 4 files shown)


src/util/intOps.ml
```diff
@@ -192,10 +192,14 @@ struct
...
   (* If the first operand of a div is negative, Zarith rounds the result away from zero.
-    We thus always transform this into a divison with a non-negative first operand.
+     We thus always transform this into a division with a non-negative first operand.
   *)
   let div a b = if Big_int_Z.lt_big_int a zero then Big_int_Z.minus_big_int (Big_int_Z.div_big_int (Big_int_Z.minus_big_int a) b) else Big_int_Z.div_big_int a b
-  let rem = Big_int_Z.mod_big_int
+
+  (* Big_int_Z.mod_big_int computes the Euclidian Modulus, but what we want here is the remainder, as returned by mod on ints
+     -1 rem 5 == -1, whereas -1 Euclid-Mod 5 == 4
+  *)
```
sim642, line 201: I assume this is already the case with other `IntOps` implementations, right? Maybe this should also be stated for `val rem` in its signature documentation to make it clear that we want this behavior consistently for all of them.

sim642 (approved):

sim642:
Slightly off-topic remark but, I just noticed that there's `Big_int_Z.gcd_big_int` and `Big_int_Z.abs_big_int`. It might make sense to add `gcd` and `abs` to `IntOps` as well, so `Congruence` doesn't have to reimplement them abstractly, which is likely less efficient than Zarith letting GMP do them directly.

michael-schwarz:
> It might make sense to add `gcd` and `abs` to `IntOps` as well

`abs for sure, for `gcd` we would have to carefully verify it is the same as the one Granger describes. He makes the point in his paper, that many GCD implementations are subtly wrong, and not suitable for use in the domain.

src/util/intOps.ml
michael-schwarz: I added a comment. Yes, all other implementations compute the remainder.

PR #303, opened 2021-07-25
Make use of congruences in invariant

### LINKED ISSUE #268: Make use of congruences in `invariant`
Once the practical course has finished, and #260 is merged, we should also exploit this new domain in the `invariant` function in base.ml. 

This allows the analysis to profit from guards such as the following one in a more meaningful way:

~~~C
if(x % 5 ==0) {
   // we should have `x = 0 mod 5` in here
}
~~~

One can at least do something for `x % b =c` whenever `c` and `b` are definite values, and probably even in more cases. The place to do this is here:

https://github.com/goblint/analyzer/blob/95d75cc881c800630c1072fdc1d4a2f21165c089/src/analyses/base.ml#L1369-L1390

michael-schwarz:
This adds some refinement to the `invariant` function for expressions of the form `a % b == c`.

https://github.com/goblint/analyzer/blob/8b1e63d9eb92c7d8c4de601a7e12cb44f434ec74/src/analyses/base.ml#L1458-L1462

**If both `b` and `c` are definite values, then `a` is in `c+bℤ`.**

This is added on top of the other refinements already present in the `mod` case, so they keep working when congruences are disabled. As explained in #268, the refinement works even without this change when signed types are used and `ana.int.congruence_no_overflow` is enabled. After this change, it works also if the flag is off and for unsigned types.

To achieve this, the IntDomains now provide an `of_congruence` function similar to `of_interval` or `of_excl_list`.

This closes #268.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1453,7 +1453,15 @@ struct
             ID.meet a' (ID.sub (ID.mul (ID.div a b) b) c)
           else a'
         in
-        meet_bin a'' b'
+        let a''' =
+          (* if both b and c are definite, we can get a precise value in the congruence domain *)
+          if ID.is_int b && ID.is_int c then
+            (* a%b == c  -> a: c+bℤ *)
+            let t = ID.of_congruence ikind ((BatOption.get @@ ID.to_int c), (BatOption.get @@ ID.to_int b)) in
+            ID.meet a'' t
+          else a''
+        in
+        meet_bin a''' b'
       | Eq | Ne as op ->
         let both x = x, x in
         let m = ID.meet a b in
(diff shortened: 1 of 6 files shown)


michael-schwarz:
Slightly off-topic remark but, I just noticed that there's `Big_int_Z.gcd_big_int` and `Big_int_Z.abs_big_int`. It might make sense to add `gcd` and `abs` to `IntOps` as well, so `Congruence` doesn't have to reimplement them abstractly, which is likely less efficient than Zarith letting GMP do them directly.

_Originally posted by @sim642 in https://github.com/goblint/analyzer/issues/302#issuecomment-886448846_

michael-schwarz:
I pulled `abs` and `gcd` into IntOps.
The latter requires and additional `abs` to behave consistently, otherwise the sign of `gcd(n,0)` changes depending on `n` for negative `n`. I guess the implementation slightly differs between `GMP` and the bigger (i.e., still negative but smaller magnitude) values handled directly in `OCaml`.

PR #304, opened 2021-07-25
Add global option to assume no overflows for signed types

### LINKED ISSUE #273: Add global option to assume no overflows for signed types

michael-schwarz:
- [x] Replace `ana.int.congruence_no_overflow` and `ana.int.wrap_on_signed_overflow` with `sem.int.signed_overflow`
- [x] Add support for ignoring overflows to
  - [x] DefExc
  - [x] Enums
  - [x] Congruences 
  - [x] Intervals
- [x] Figure out interaction with overflow detection mechanism
- [x] Deduplicate `should_ignore_overflow` and `should_overflow`
- [x] Figure out what to do about `ana.octapron.no_signed_overflow`
- [x] Do a run of SV-Comp to see if this improves our results

Closes #273.


src/cdomains/intDomain.ml
```diff
@@ -524,7 +524,32 @@ struct
...
+        if Ints_t.compare resdiff diff > 0 then
+          top_of ik
+        else
+          let l = Ints_t.of_bigint @@ Size.cast_big_int ik (Ints_t.to_bigint x) in
+          let u = Ints_t.of_bigint @@ Size.cast_big_int ik (Ints_t.to_bigint y) in
+          if Ints_t.compare l u <= 0 then
+            Some (l, u)
+          else
+            (* Interval that wraps around (begins to the right of its end). We can not represent such intervals *)
+            top_of ik
```
michael-schwarz, line 546: I would be very interested in feedback on this. It seems very complicated, but the simpler things I tried all turned out to be unsound.

src/cdomains/intDomain.ml
```diff
@@ -524,7 +524,30 @@ struct
 
   let norm ik = function None -> None | Some (x,y) ->
     if Ints_t.compare x y > 0 then None
-    else if Ints_t.compare (min_int ik) x > 0 || Ints_t.compare (max_int ik) y < 0 then (set_overflow_flag ik; top_of ik)
+    else if Ints_t.compare (min_int ik) x > 0 || Ints_t.compare (max_int ik) y < 0 then (
+      set_overflow_flag ik;
+      if should_wrap ik then
+        (* We can only soundly wrap if at most one overflow occurred, otherwise the minimal and maximal values of the AP interval *)
```
sim642, line 530: `AP interval`?

src/util/gobConfig.ml
```diff
@@ -84,6 +84,14 @@ sig
...
   (** Add a schema to the conf*)
   val addenum_sch: jvalue -> unit
+
+  (** Whether for a given ikind, we should compute with wrap-around arithmetic.
+    *  Always for unsigned types, for signed types if 'sem.int.signed_overflow' is 'assume_wraparound'  *)
+  val should_wrap: Cil.ikind -> bool
+
+  (** Whether for a given ikind, we should assume there are no overflows.
+    * Always false for unsigned types, true for signed types if 'sem.int.signed_overflow' is 'assume_none'  *)
+  val should_ignore_overflow: Cil.ikind -> bool
```
sim642, line 94: I think these C and analysis specific things shouldn't be in `GobConfig`, which otherwise implements an independent config system. These should be in `IntDomain` or some `IntUtil`.

src/cdomains/intDomain.ml
```diff
@@ -2589,7 +2637,7 @@ module IntDomTupleImpl = struct
     ); !dt
 
   let no_overflow ik r =
-    if GobConfig.get_bool "ana.int.congruence_no_overflow" && Cil.isSigned ik then true
+    if Cil.isSigned ik  && GobConfig.get_string "sem.int.signed_overflow" = "assume_none" then true
```
sim642, line 2640: This and the two following changes could just be `should_ignore_overflow` calls.

sim642:
>  Figure out what to do about `ana.octapron.no_signed_overflow`

It might make sense to do this after #291.

src/cdomains/intDomain.ml
michael-schwarz: Arbitrary Precision, I borrowed some LLVM terminology here without intending to.

PR #305, opened 2021-07-26
Fix more CFG display issues

### LINKED ISSUE #292: Empty while loops are displayed as dead
Compare the empty goto loop with the empty while loop:

| | Empty goto loop | Empty while loop |
| --- |:---:|:---:|
| Code | https://github.com/goblint/analyzer/blob/130e764d61e4de8002e9725471f556f6ebaa98b8/tests/regression/00-sanity/21-empty-loops.c#L30-L34 | https://github.com/goblint/analyzer/blob/130e764d61e4de8002e9725471f556f6ebaa98b8/tests/regression/00-sanity/21-empty-loops.c#L36-L39 |
| CFG | image | image |
| Loop state | image |  image |

Although they have an identical CFG and thus the analysis state, their output differs. The empty while loop is displayed as if it's dead and clicking on it shows that node's XML completely missing (clicking on other dead nodes at least shows the context).

### Problem
That node being completely missing from the output is due to this code excluding the output of all nodes with unknown locations:
https://github.com/goblint/analyzer/blob/130e764d61e4de8002e9725471f556f6ebaa98b8/src/framework/control.ml#L155-L171

The empty goto loop gets a location from CIL's `Goto`:
https://github.com/goblint/analyzer/blob/130e764d61e4de8002e9725471f556f6ebaa98b8/src/framework/myCFG.ml#L350-L354
The empty while loop on the other hand doesn't have an explicit `Goto`, but simply contains an `Instr []` that has itself as the successor:
https://github.com/goblint/analyzer/blob/130e764d61e4de8002e9725471f556f6ebaa98b8/src/framework/myCFG.ml#L279-L291
In this case there is nowhere to take an appropriate location from because locations are attached to all the individual instructions in `Instr`, but not the `Instr` constructor as a whole. So if it's empty, it has absolutely no location information.

### Impossible solutions
It's not as simple as removing the `locUnknown` comparison: that will cause g2html to crash because there's a node with no corresponding file (let alone a function or a line).
Also, changing the `Instr []` case to use `fd_loc` (function definition's own location) instead of `locUnknown` doesn't actually help. That location is just attached to the edge (which we maybe only use for setting tracing locations in `FromSpec`?) but not to the node and CIL stmt which are used for determining this output. This in itself is very confusing because both nodes (which are points between statements) and edges (which are the statements) have locations, and AFAIK they're not necessarily related or required to match.

Moreover, it's not possible to mutate the stmt to have a better location because there's nowhere to attach the location: stmt has no location, stmtkind has no location for the entire `Instr` and there's no instr elements.

sim642:
## Empty while loops appear dead
Closes #292.

As described in the issue, such loop bodies contain `Instr []` that has no locations to use, thus they're not properly reported in the results. Turns out that the instruction statement still has a label inserted by CIL (loop continue) and that has a location. So I implemented `Cilfacade.get_stmtLoc` (which actually works on `stmt`, not `stmtkind`) that takes the location from labels in such corner case.

A question is whether to keep it in Goblint's `Cilfacade` or move this fix up into goblint-cil. That means changing the type of a CIL function.

## Duplicate Neg(1) edges
https://github.com/goblint/analyzer/blob/0e6698e178e9b33ace993e2cb8ed4b93c160bdb1/tests/regression/34-localization/02-hybrid.c#L8-L17

| Before | After |
| --- | --- |
| image | image |

**Change:** A duplicate Neg(1) edge 6 → 22 is removed.

This was due to using `add` instead of `replace` on a `Hashtbl` to record loop heads. Loop heads reachable via multiple back edges were added multiple times and a Neg(1) edge was added for each one.

### DIFF
--- a/src/framework/cfgTools.ml
+++ b/src/framework/cfgTools.ml
@@ -172,7 +172,7 @@ let createCFG (file: file) =
         let loop_head_neg1 = NH.create 3 in
         (* So for each statement in the function body, we do the following: *)
         let handle stmt =
-          if Messages.tracing then Messages.trace "cfg" "Statement %d at %a.\n" stmt.sid d_loc (get_stmtLoc stmt.skind);
+          if Messages.tracing then Messages.trace "cfg" "Statement %d at %a.\n" stmt.sid d_loc (Cilfacade.get_stmtLoc stmt);
 
           let real_succs () = List.map (find_real_stmt ~parent:stmt) stmt.succs in
 
--- a/src/framework/cfgTools.ml
+++ b/src/framework/cfgTools.ml
@@ -26,9 +26,9 @@ let find_loop_heads_fun (module Cfg:CfgForward) (fd:Cil.fundec): unit NH.t =
   (* DFS *)
   let rec iter_node path_visited_nodes node =
     if NS.mem node path_visited_nodes then
-      NH.add loop_heads node ()
+      NH.replace loop_heads node ()
     else if not (NH.mem global_visited_nodes node) then begin
-      NH.add global_visited_nodes node ();
+      NH.replace global_visited_nodes node ();
       let new_path_visited_nodes = NS.add node path_visited_nodes in
       List.iter (fun (_, to_node) ->
           iter_node new_path_visited_nodes to_node
@@ -285,6 +285,7 @@ let createCFG (file: file) =
         let reachable_return = find_backwards_reachable (module TmpCfg) (Function fd) in
         NH.iter (fun node () ->
             if not (NH.mem reachable_return node) then (
+              if Messages.tracing then Messages.tracei "cfg" "unreachable loop head %a\n" pretty_short_node node;
               let targets = match NH.find_all loop_head_neg1 node with
                 | [] -> [Lazy.force pseudo_return]
                 | targets -> targets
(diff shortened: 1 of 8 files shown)


michael-schwarz (approved):

michael-schwarz:
> A question is whether to keep it in Goblint's Cilfacade or move this fix up into goblint-cil

In general, I would be all for pulling this up into goblint-cil. The question just is whether it is ok to use this modified function everywhere inside CIL where the other one is used. There seem to be only two usages inside CIL, one in `cil.ml` and one in `dataflow.ml`.

The one in `dataflow.ml` branches on this value and sets `currentLoc` to the return value depending on whether it is `locUnknown` or not. So I'm not sure if it sound to change this function here.

sim642:
> In general, I would be all for pulling this up into goblint-cil. The question just is whether it is ok to use this modified function everywhere inside CIL where the other one is used. There seem to be only two usages inside CIL, one in `cil.ml` and one in `dataflow.ml`.

All the current `get_stmtLoc` uses in CIL itself could simply be replaced with `get_stmtkindLoc`, which is what the current function really is, without altering any of the behavior. That can be done anytime in the future as well, so I'll merge it here for now.

Issue #236, opened 2021-05-20
ocp-indent pre-commit hook


sim642:
`make dev` sets up a Git pre-commit hook, which prevents committing any files with indentation that the tool considers "wrong":
https://github.com/goblint/analyzer/blob/f99164f777b5ce2c19835e73e991f0deb73f8a91/make.sh#L102-L110

I deleted this hook locally already years ago and I guess many of you have done the same (or never even set it up). It was impossible to make tiny correctly indented changes because it also forced to fix the indentation of the rest of the file, which hadn't been touched for a long time (I guess ocp-indent cannot be told to just complain about the modified changes). This is clearly still an issue: https://github.com/vandah/analyzer/commit/4e8840c19fe54b65755c92ad3fc8c4786e347421.

It forces the awful practice of doing logic changes together with completely unrelated formatting changes in the same commit. This obfuscates the Git history and makes Git blame more difficult to use, as one cannot just ignore formatting commits if they also contain logic changes: https://github.com/goblint/analyzer/blob/master/.git-blame-ignore-revs. Our suggested `make dev` setup should not be encouraging this practice!

I'm wondering what everyone's thoughts about this are, because it doesn't seem great that we recommend a pre-commit hook, which will immediately hinder development and just push everyone to remove the hook, which in itself leads again to more misindented code being committed.

One thought I have is to remove the hook, so locally you can make all the changes without being forced to combine unrelated formatting changes. Instead we could have a GitHub Actions job that runs ocp-indent. That would of course require fixing all the indentation problems that already are around and also agreeing that ocp-indent is the _gold standard_ for our code style.

jerhard:
I think having a common, automatable indentation style would be a good thing. From what I saw, the changes that ocp-indent makes to indentation are mostly minor, and it e.g. does not seem to insert line breaks depending on the syntax, but it makes the indentation more consistent.

>That would of course require fixing all the indentation problems that already are around and also agreeing that ocp-indent is the gold standard for our code style.

Are there other options?

vogler:
> It forces the awful practice of doing logic changes together with completely unrelated formatting changes in the same commit. This obfuscates the Git history and makes Git blame more difficult to use ...

Yes, true, also deleted it. Was too lazy to make it only apply to the diff.
Changing the `xargs` arg from
https://github.com/goblint/analyzer/blob/f99164f777b5ce2c19835e73e991f0deb73f8a91/scripts/hooks/pre-commit#L52
to
~~~sh
[...]

PR #308, opened 2021-07-27
check indentation of changes on GitHub

vogler:
Stumbled upon this: https://pre-commit.ci
Not sure if we want automatic commits that fix indentation in PRs.
Also, https://pre-commit.com does not list OCaml, so we might as well just add another GitHub workflow for it.

_Originally posted by @vogler in https://github.com/goblint/analyzer/issues/236#issuecomment-887556635_

Setting up OCaml just to install `ocp-indent` to check the code seems like a lot overhead, but maybe this can be cached.

### DIFF
--- a/.github/workflows/indentation.yml
+++ b/.github/workflows/indentation.yml
@@ -1,0 +1,30 @@
+name: indentation
+
+on: [ push, pull_request]
+
+jobs:
+  indentation:
+    strategy: # remove?
+      matrix:
+        os:
+          - ubuntu-latest
+        ocaml-compiler:
+          - 4.12.x
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      # reuse tests.yml or depend on it? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow
+      - name: Set up OCaml ${{ matrix.ocaml-compiler }}
+        uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: ${{ matrix.ocaml-compiler }}
+
+      - name: Install ocp-indent
+        run: opam install -y ocp-indent
+
+      - name: Run pre-commit hook
+        run: git reset --soft HEAD~ && ./scripts/hooks/pre-commit
--- a/.github/workflows/indentation.yml
+++ b/.github/workflows/indentation.yml
@@ -33,9 +33,9 @@ jobs:
           key: opam-ocp-indent-${{ runner.os }}-${{ matrix.ocaml-compiler }}
 
       - name: Set up OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v2
+        uses: ocaml/setup-ocaml@v1
         with:
-          ocaml-compiler: ${{ matrix.ocaml-compiler }}
+          ocaml-version: ${{ matrix.ocaml-compiler }}
 
       - name: Install ocp-indent
         run: opam install -y ocp-indent
--- a/.github/workflows/indentation.yml
+++ b/.github/workflows/indentation.yml
@@ -4,14 +4,7 @@ on: [ push, pull_request]
 
 jobs:
   indentation:
-    strategy: # remove?
-      matrix:
-        os:
-          - ubuntu-latest
-        ocaml-compiler:
-          - 4.12.0 # setup-ocaml@v1 does not support 4.12.x for ocaml-version
-
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


vogler:
Action failed with
```console
Run git reset --soft HEAD~ && ./scripts/hooks/pre-commit
fatal: ambiguous argument 'HEAD~': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

sim642:
This only checks the last commit, right? So if you push multiple commits, then it won't check for misindentations introduced by earlier commits that this workflow didn't run for.

vogler:
> This only checks the last commit, right? So if you push multiple commits, then it won't check for misindentations introduced by earlier commits that this workflow didn't run for.

Ah, yes, good point.
Confusing docs, maybe reset to `${{ github.event.push.before }}` then.
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#push
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push

vogler:
Ok, no, that's empty: `git reset --soft  `

vogler:
Ok, `${{ github.event.before }}` works.

Setup OCaml + ocp-indent still takes 1m21s + 30s.
So it seems the caching by `ocaml/setup-ocaml@v2` is not that good (or missing some argument).
Something to consider for #296.

vogler:
With `ocaml/setup-ocaml@v1` and manual caching the run-time decreased from ~2m to 31-35s.

vogler:
Ok, with `apt install ocp-indent` instead of via `opam install ocp-indent` it's only 12s instead of 31s.
(ubuntu-latest is 20.04 LTS which has ocp-indent 1.7.0 instead of 1.8.1 from opam.)

At least this detour was helpful for some opam cache experiments for #296.

michael-schwarz:
In the run https://github.com/goblint/analyzer/runs/3199968802, we have the error message

```
ocp-indent file: src/analyses/arinc.ml, lines: 122
ocp-indent: internal error, uncaught exception:
            Failure("Config file not found - neither /etc/ocamlfind.conf nor the directory /etc/ocamlfind.conf.d")
```

vogler:
So the apt package is broken? What does it need `ocamlfind` for? Strange.
I'll just revert to `opam` and merge.

sim642:
It's still somehow broken, cannot find `ocp-indent`: https://github.com/goblint/analyzer/runs/3201066169?check_suite_focus=true#step:6:8.

vogler:
I added a `eval $(opam env)` before in 0f86542f188a6cc67e405c6ef977336baa33b4e8 - hope that helps. I thought ocaml-setup should take care of modifying `.profile`/`.bashrc`...

I can't resurrect this PR anymore? I committed it on its branch, but it's not showing those commits anymore after the merge...

michael-schwarz:
https://github.com/goblint/analyzer/pull/310/checks?check_run_id=3212969178 fails with

```
Run git reset --soft 0000000000000000000000000000000000000000 && eval $(opam env) && ./scripts/hooks/pre-commit
  git reset --soft 0000000000000000000000000000000000000000 && eval $(opam env) && ./scripts/hooks/pre-commit
  shell: /usr/bin/bash -e {0}
  env:
    OPAMJOBS: 4
    OPAMYES: 1
fatal: Could not parse object '0000000000000000000000000000000000000000'.
Error: Process completed with exit code 128.
```

vogler:
Strange that it's all 0.
`${{ github.event.before }}`: 'The SHA of the most recent commit on `ref` before the push.'([doc](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push)).
Looks like you pushed 2 commits on Aug 1, and there's a commit before on Jun 16, correct? So it should evaluate to this commit instead of all 0.
I could change it to fallback to `HEAD~` or prob. better `HEAD^` if it's all 0, but it'd be better to know what's the reason for it first.

michael-schwarz:
> branch rename

Yes, I renamed the branch, but that was entirely local before it was published to GitHub. I'm not sure if this is visible to GitHub even, one would think the name of the local branch should not matter...
But maybe this is the cause

sim642:
If the branch was new on GitHub, then there was no most recent commit on it (the `ref`) before the push. Technically there were the commits on master that it merges, but maybe the actions events are directly branch-based. Or maybe they just don't follow merges because it would make the value ambiguous.

This is probably more work but there's also this:

| Key | Type | Description |
| -- | -- | -- |
| `commits[][distinct]` | `boolean` | Whether this commit is distinct from any that have been pushed before. |

vogler:
Just confirmed on https://github.com/goblint/analyzer/runs/3230967934 that the first push on a new branch has all 0.

vogler:
https://github.com/goblint/analyzer/commit/0ce4a7dadf2fd2c644ec1bab3e8321d4498f2fcc#r55854725

https://github.com/goblint/analyzer/blob/7879abe801d3bf42067fac9a6470bd526b8ae80d/.github/workflows/indentation.yml#L46

Did you try `git reset --soft HEAD^` if it's all 0?

sim642:
> This means the first push on a new branch won't be checked but only the following changes?

I suppose so. I just wanted those zero commit failures to go away because they made the indentation workflow too easy to ignore for false alarms.

> Did you try `git reset --soft HEAD^` if it's all 0?

That's not fully accurate either because you might push multiple commits to a new branch, but this would only consider the last of those.

Issue #252, opened 2021-06-16
Modification via pointers to locals in recursive functions


michael-schwarz:
After some discussion at GobCon yesterday, we were able to come up with an example where locals "escape" to recursive calls.

https://github.com/goblint/analyzer/blob/2569701a2bdeee6811b90b6968192e387a10f402/tests/regression/01-cpa/50-escaping-recursion.c#L1-L33

- The program has one call to `f(0,&t)`.
   - This call recursively calls `f(5,&x)` passing a reference to its local x (Let us call this variable **x0** for convenience).
       - The call to `f(5,&x)` then sets its x (**x5**) to 31. The following `*ptr` has `&x` and this is understood to refer to **x5**, where **x0** would be correct
       - After the nondeterministic branch `ptr` may point to **x0** and **x5**
       - The assignment modifies one, meaning that the asserts should be unknown
       - The call returns, having potentially modified x of its caller (**x0**)
   - After `f(5,&x)` returns x (**x0**) may have the values 17 or 12, the assert should fail.
   
@jerhard and I discussed how one could fix this: Our conclusion was to modify the enter such that `ptr` would point to a different copy of x (the one from the caller). In combine, one would then take the value of this outer copy in the return state of the function as the new value for the local x. This ought to be enough, as one can only refer to the outer copies of variables via pointers, and references to the variable directly refer to the instance in the current stackframe.

michael-schwarz:
Related to #246  in a way, but I would think at the core, these are separate issues.

sim642:
I don't think this is as "easy" as just creating an auxiliary `varinfo` for passing the outer reachable variable into the call. Everything reachable from the arguments (through address sets, structs, unions, arrays) is forwarded to the callee and there might be multiple levels of indirection (including cycles). So the normal `varinfo` must be substituted with the outer `varinfo` in all the passed abstract values as well.

Moreover, these `varinfo`s synthesized by base could confuse other analyses, e.g. when they query base for address sets that end up containing them. For example, the pointer you're passing is to a mutex, which the callee then locks and holds in its local lockset. If it's a wrapper for locking, such that the callee doesn't also unlock it, then its return state's lockset contains that synthetic lock. If you then unlock it directly outside, then you'd be trying to unlock something that's not in the lockset even. Suddenly the mutex analysis also needs to be able to undo the transformation of base.

sim642:
[...]

PR #310, opened 2021-08-01
Fix "escaping" via recursion

michael-schwarz:
The issue we discussed in #252 is that pointers to the locals of some invocation of the same function further down the callstack may be passed to a function, which may in turn modify this value.

https://github.com/goblint/analyzer/blob/5e05144dbf9361e1ab1b115277a9ce9710f4aa93/tests/regression/01-cpa/50-escaping-recursion.c#L1-L42

In #252, @jerhard  and I proposed a very elaborate solution, that would introduce a new variable `x_outer` for such a local variable when it is passed to the function where it is defined. Then, all references to the original `x` in the whole `CPA` would be patched to point to `x_outer`. This would then be undone in the combine.
If the situation where an `x_outer` is already reachable for a function call to its defining function, `x_outer` would be marked as multiple, and weak updates would be performed to it.

**This PR takes a somewhat more simple approach: Enhance CPA with a set of variables that must be weakly updated**, because an outer instance of this variable may also be reachable. This way, one avoids introducing `x_outer` and patching address sets, as the variable `x` now abstracts all instances of this variable in the whole call stack insofar as they might be reachable from the current invocation.

It works as follows:
- In `[[enter]]#` for a fundec:
    -  add those locals of fundec that are reachable via an argument to the set of values to weakly update in the callee
    
    https://github.com/goblint/analyzer/blob/5e05144dbf9361e1ab1b115277a9ce9710f4aa93/src/analyses/base.ml#L1872-L1875
    
- When updating the value for a variable: 
  - Check if it needs to be weakly updated, if yes, join new and previous value
  https://github.com/goblint/analyzer/blob/5e05144dbf9361e1ab1b115277a9ce9710f4aa93/src/analyses/base.ml#L1070-L1079

- In `[[return]]#`
  - Instead of removing all locals from the CPA, only remove those that do not represent multiple concrete variables (i.e. that do not need to be weakly updated)
 
  https://github.com/goblint/analyzer/blob/5e05144dbf9361e1ab1b115277a9ce9710f4aa93/src/analyses/base.ml#L1772
  

- In `[[combine]]#` 
     - For those locals that are still in the return value of the callee: Update the local copy to their value
     - Proceed with the set of variables to be weakly updated from the caller
 
 
**This is a bit less precise than the other solution that was proposed in #252**, as it assumes that any writes to the local may also be writes via pointers to the outer (reachable) values. 
https://github.com/goblint/analyzer/blob/5e05144dbf9361e1ab1b115277a9ce9710f4aa93/tests/regression/01-cpa/50-escaping-recursion.c#L30-L32

On the other hand, I think it is simpler as it does not require complicated patching of the address sets.

The code obviously still needs cleanup, but I think we can already discuss if we want that solution, before investing time to make it pretty.
What do you think @sim642 @vogler @vesalvojdani @jerhard ?


michael-schwarz:
It seems like the issue is here is that this data structure sometimes also escapes to globals via the `CPA` for some of the more elaborate privatization schemes. So I'll reimplement it a level higher in the base domain.

src/analyses/base.ml
```diff
@@ -1866,7 +1891,11 @@ struct
     let reachable = List.concat (List.map AD.to_var_may (reachable_vars (Analyses.ask_of_ctx ctx) (get_ptrs vals) ctx.global st)) in
     let reachable = List.filter (fun v -> CPA.mem v st.cpa) reachable in
     let new_cpa = CPA.add_list_fun reachable (fun v -> CPA.find v st.cpa) new_cpa in
-    {st' with cpa = new_cpa}
+    (* Identify locals of this fundec for which an outer copy (from a call down the callstack) is reachable *)
+    let reachable_other_copies = List.filter (fun v -> try (match Cilfacade.find_scope_fundec v with | Some fdec -> fdec.svar.vid = fundec.svar.vid |_ -> false) with _ -> false) reachable in
```
sim642, line 1895: This is quite difficult to read, maybe something like this instead:
```suggestion
    let is_outer v = match Cilfacade.find_scope_fundec v with
      | Some scope -> CilType.Fundec.equal scope fundec
      | None
      | exception Not_found -> false
    in
    let reachable_other_copies = List.filter is_outer reachable in
```

Although when does `find_scope_fundec` fail? It has an `option` return type, so it really shouldn't. Or if it does, then maybe a fix should be done in it.

src/cdomains/baseDomain.ml
```diff
@@ -19,8 +19,8 @@ struct
     let module VS = Set.Make (Basetype.Variables) in
     let rec context vs = {c with
         deref_invariant=(fun vi offset lval ->
-          let v = find vi m in
-          key_invariant_lval vi offset lval v vs
+                          let v = M.find vi m in
+                          key_invariant_lval vi offset lval v vs
```
sim642, line 23: No need to change this any more.

src/cdomains/baseDomain.ml
```diff
@@ -34,8 +34,7 @@ struct
 
     let key_invariant k v =
       key_invariant_lval k NoOffset (var k) v VS.empty in
-
-    fold (fun k v a ->
+    M.fold (fun k v a ->
```
sim642, line 37: No need to change this any more.

src/domains/queries.ml
```diff
@@ -92,6 +92,7 @@ type _ t =
   | MayBeLess: exp * exp -> MayBool.t t (* may exp1 < exp2 ? *)
   | HeapVar: VI.t t
   | IsHeapVar: varinfo -> MayBool.t t (* TODO: is may or must? *)
+  | MayBeMultiple: varinfo -> MayBool.t t
```
sim642, line 95: I find this query quite confusing. I see the only use for this right now is to allow vareq to ask base about its `WeakUpdates` set.

The issue of a single `varinfo` representing multiple concrete variables is highly specific to base (and I guess vareq) because they pass reachable variables into called functions, allowing for this situation to arise. However, this isn't fully generic in the sense that every analysis should be considering `MayBeMultiple` because naively one might also expect octApron to consider this information to make weak updates. But it doesn't have to (AFAIK) because it never passes local state into called functions in the same way as base does, so the confusion doesn't arise. Also, it somehow assumes that vareq passes reachable locals into called functions in exactly the same way as base, but is that really the case? So it seems to me like this notion is specific to each analysis wrt how much it passes down.

Moreover, if it were fully generic, I'd expect it to be possible to fix #316 using the same query by having mallocWrapper claim all of the `varinfo`s that it has returned to represent multiple concrete values (until we implement more precise uniqueness for mallocs). But again, this query wouldn't really work for that: mallocWrapper may answer the query for a malloced `varinfo` with `true`, but that wouldn't be in base's `WeakUpdates` set, so base answers with `false` and the meet of `MayBool` yields `false`. So for that kind of general use, it should be `MustBool` instead for a negated query like `MustBeUnique`, so if it's either in base's `WeakUpdates` or created by mallocWrapper, it wouldn't be unique any more.

But that doesn't really solve the first problem, that what is considered to be non-unique by one analysis isn't necessary non-unique for another.

michael-schwarz: >  is highly specific to base (and I guess vareq) because they pass reachable variables into called functions, allowing for this situation to arise

I don't really think it is. As soon as the base analysis treats something as Multiple all other analyses must also do so or not  answer `EvalInt` queries. This is because the points-to information that is used from base does not distinguish between inner and outer instances

> because naively one might also expect octApron to consider this information to make weak updates

I am fairly certain it needs to. At least it currently is unsound (both asserts succeed).

~~~C
// PARAM: --sets ana.activated[+] "var_eq" --sets ana.activated[+] "octApron"
int rec(int i,int* ptr) {
    int top;
    int x = 17;

    if(i == 0) {
        rec(5,&x);
        // Recursive call may have modified x
        assert(x == 17); //UNKNOWN!
    } else {
        x = 31;

        // ptr points to the outer x, it is unaffected by this assignment
        // and should be 17
        assert(*ptr == 31); //UNKNOWN!
    }
    return 0;
}

int main() {
    int t;
    rec(0,&t);

    return 0;
}
~~~

The question is just whether to fix all analyzes in this PR or make separate ones. The issue is that the octApron has `x = 31` after the assignment, the address domain in base has `ptr -> {&x}`  and asks for the value of `x` via the query system, to which octApron answers `[31,31]`.

michael-schwarz: Actually, after recompiling :upside_down_face:, it does seem to work with octApron, as it does not have `x` as a tracked local variable? :thinking:

sim642: Something else must be wrong with octApron in this case because `x` has its address taken in `rec(5,&x)`, so octApron shouldn't be tracking `x` at all.

src/analyses/base.ml
michael-schwarz: >  Although when does find_scope_fundec fail?

E.g. in the case of vars created by `malloc(...)` that were not in the globals hashtable when `varinfo_roles` was computed:

~~~C
void init (int *p, int x) {

}

int main () {
  int *p = malloc(sizeof(int));
  init(p,9);

  return 0;
}
~~~

sim642: Ahh yes, of course. Then to be safe it indeed makes sense to change `find_scope_fundec` to return `None` also for those because other uses of that function could hit the same issue.

src/domains/queries.ml
michael-schwarz: > There should probably be a comment in octApron then, explaining that due to vaddrof it doesn't need to consider it at function calls etc.

I'll ad a comment

> Still, I think it would make sense to flip the query to make it possible for mallocWrapper to also indicate the alloc variables as multiple/non-unique.

Sure, I'll flip it!

> Actually, would it even be possible to pull this WeakUpdates stuff into a separate analysis instead of baking it into base, if other analyses should be considering it regardless of what base is doing? It

That is very complicated, it requires the may-point-to sets at function entry. And if we query the base analysis for that, it is not too different from just embedding it.

sim642: I was thinking about this query again and how it would behave if mallocWrapper also answered it such that any varinfo that it created is not unique, but somehow it still doesn't make sense the way I thought it would.

#### MayBeMultiple (before)
| Variable | Base | MallocWrapper | Meet (and, `MayBool`) |
| --- | --- | --- | --- |
| Escaped | true (in `WeakUpdates`) | true (not created by mallocWrapper, top of `MayBool`) | true (:heavy_check_mark:) |
| Malloced | false (not in `WeakUpdates`) | true (created by mallocWrapper) | false (:x:) |

#### MustBeUnique (now)
| Variable | Base | MallocWrapper | Meet (or, `MustBool`) |
| --- | --- | --- | --- |
| Escaped | false (in `WeakUpdates` negated) | false (not created by mallocWrapper, top of `MustBool`) | false (:heavy_check_mark:) |
| Malloced | true (not in `WeakUpdates` negated) | false (created by mallocWrapper) | true (:x:) |

#### MustBeMultiple (should be?)
| Variable | Base | MallocWrapper | Meet (or, `MustBool`) |
| --- | --- | --- | --- |
| Escaped | true (in `WeakUpdates`) | false (not created by mallocWrapper, top of `MustBool`) | true (:heavy_check_mark:) |
| Malloced | false (not in `WeakUpdates`) | true (created by mallocWrapper) | true (:heavy_check_mark:) |

It seems to me now like the query should be `MustBeMultiple` because otherwise one of the `varinfo` multiplicity/uniqueness analyses needs to somehow subsume the other: base's `WeakUpdates` should also automagically contain all malloced variables?

sim642:
Funny how GitHub claimed that there's a conflict with `octApron.apron.ml` that is so complicated it cannot even show the conflict in the web interface. On this branch you added a comment, on master the file was renamed to `apronAnalysis.apron.ml`, which I guess is the conflict. But somehow locally merging works automatically and your comment is properly merged into `apronAnalysis.apron.ml` just fine.

src/domains/queries.ml
michael-schwarz: Interesting!

`MustBeMultiple` is also somehow not the right thing though. If something is in this `WeakUpdates` set, this just means that an outer copy **may** be reachable not that it must be. 
So the name `MustBeMultiple` would be very confusing, since we need to weakly update all things that **may** be multiple, not just those that must be.
Regardless of this, I agree that MustBool is the correct lattice. I'll see how to make this change while not making it confusing. (Maybe some comment?)

sim642: Maybe calling it `IsMultiple` is the right thing to do? It doesn't demand anything about must-multiplicity, but rather just asks each analysis whether it has made the `varinfo` multiple. And then `MustBool` is just the right lattice to correctly combine the answers of different analyses.

It muddies the water though because at some point we wanted all the queries to clearly be one or the other by name.

michael-schwarz: I went with the solution from the last comment now!

src/domains/queries.ml
```diff
@@ -127,7 +127,7 @@ struct
     | MustBeProtectedBy _ -> (module MustBool)
     | MustBeAtomic -> (module MustBool)
     | MustBeSingleThreaded -> (module MustBool)
-    | MustBeUniqueThread -> (module MustBool)
+    | IsMultipleThread -> (module MustBool) (* see https://github.com/goblint/analyzer/pull/310#discussion_r700056687 on why this needs to be MustBool *)
```
sim642, line 130: Besides renaming `MustBeUnique` -> `IsMultiple` you also renamed `MustBeUniqueThread` -> `IsMultipleThread`, which I think has nothing to do with this issue or discussion. Was renaming this query accidental?

michael-schwarz:
Yup, it was accidental. Feel free to revert or I'll do it when I'm back at
the office next week

PR #311, opened 2021-08-03
fixes for Apple M1

vogler:
Related issues:
- [ ] https://github.com/goblint/cil/issues/41
  - [ ] TODO fix `__int128` in CIL, then change the definition in `arm64.h` which currently uses `unsigned long long` instead
- [ ] add `linux-headers/arch/arm64` since we  only have `x86` which fails:
```
goblint/analyzer/linux-headers/arch/x86/include/asm/pvclock.h:56:2: error: #error implement me!
  -    56 | #error implement me!
  -       |  ^~~~~
  - Goblint: Preprocessing failed.
```

### DIFF
--- a/includes/arm64.h
+++ b/includes/arm64.h
@@ -1,0 +1,9 @@
+// see https://github.com/goblint/cil/issues/41
+// If goblint pre-processes a file including stdlib.h on an Apple M1 machine,
+// it will include /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/mach/arm/_structs.h
+// which uses __uint128_t which is somehow not defined before.
+
+#if defined(__arm64__)
+  // typedef unsigned __int128 uint128_t; // https://stackoverflow.com/a/34588884 TODO __int128 exists in CIL but fails; see https://github.com/goblint/cil/issues/41
+  typedef unsigned long long __uint128_t; // TODO wrong! use the above once fixed. This the definition for __uint64_t
+#endif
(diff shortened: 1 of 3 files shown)


michael-schwarz:
Do you want to address these related issues before merging this or should we merge it first (to get something working on M1 machines) and then tackle these separately?

vogler:
There are actually two goals:
- running on arm64 and analyzing as x86_64 requires https://github.com/goblint/cil/issues/41 and `-D__x86_64__` added to `cppflags`
- running on any and analyzing arm64 requires #312 

I haven't tested how different preprocessing results for `arm64` vs. `x86_64` are (can be done on any machine).
To have comparable benchmarks, it might make sense to analyze as `x86_64` even on other architectures.
OTOH it may conflict with the sizes determined by CIL in `machdep.ml` which would depend on #54.

With https://github.com/goblint/cil/issues/41 fixed, we could merge this to cover the first goal.
I'll add `-D__x86_64__` on `arm64` (could do it for any arch I guess). #54 can then introduce an option `arch` or sth similar.

michael-schwarz:
I think unlocked regression tests fail because the `goblint.opam` file was not regenerated in this PR and still has the old pin. We just don't see those failures yet on master because they only run nightly.

sim642:
> Maybe we should add some dune rule for generating or at least checking `goblint.opam` and `goblint.opam.locked` when `goblint.opam.template` changes.

This would happen if we build everything with `dune build` directly, instead of having `make` just compile the `.exe` target. But I think then it also tries to do the JS build, which last I tried failed for me (JS dependencies aren't declared?).

PR #313, opened 2021-08-04
Make Spec & Inc arguments of the functor AnalyzeCFG

keremc:
This way, module definitions like `EQSys`, `LHT`, `GHT` & `Slvr` and functions like `solver2source_result` & `print_dead_code` become accessible from other modules. I also pulled `analyze_loop` out of `AnalyzeCFG` and modified it a bit to keep the recursive restarting behavior.

### DIFF
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -584,6 +584,11 @@ struct
     Result.output (lazy local_xml) gh make_global_fast_xml file
 end
 
+(* This function was originally a part of the [AnalyzeCFG] module, but
+   now that [AnalyzeCFG] takes [Spec] as a functor parameter,
+   [analyze_loop] cannot reside in it anymore since each invocation of
+   [get_spec] in the loop might/should return a different module, and we
+   cannot swap the functor parameter from inside [AnalyzeCFG]. *)
 let rec analyze_loop (module CFG : CfgBidir) file fs change_info =
   try
     let (module Spec) = get_spec () in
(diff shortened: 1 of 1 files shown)


src/framework/control.ml
```diff
@@ -586,23 +582,21 @@ struct
...
-    with Refinement.RestartAnalysis ->
-      (* Tail-recursively restart the analysis again, when requested.
-         All solving starts from scratch.
-         Whoever raised the exception should've modified some global state
-         to do a more precise analysis next time. *)
-      (* TODO: do some more incremental refinement and reuse parts of solution *)
-      analyze_loop file fs change_info
 end
 
+let rec analyze_loop (module CFG : CfgBidir) file fs change_info =
```
sim642, line 587: I think there should be a comment here about why `analyze_loop` cannot be part of `AnalyzeCFG` any more. Took me a while to figure out it's simply because each call of `get_spec` in the loop should return a different `Spec` module, so obviously the functor argument cannot change within the functor itself.

keremc: Good point. I remember being confused by `analyze_loop` myself because I thought `get_spec` would always have the same result.

sim642: It's a quite obscure and hacky thing anyway. I only added this mechanism for refining with witnesses, where the `ana.activated` option is modified before restarting, causing `MCP2` inside `get_spec ()` to be extended.

sim642 (approved):

PR #315, opened 2021-08-05
Use fork of Zarith

keremc:
Use [this fork](https://github.com/goblint/Zarith/tree/goblint) of Zarith to support the deserialization of small ints in the browser. You can read [this document](https://github.com/goblint/Zarith/raw/goblint/goblint/main.pdf) to learn more about the complete rationale behind the changes I made.

For the release branch of this fork, I bundle all my changes in [a single patch file](https://github.com/goblint/Zarith/blob/goblint-release-1.12/goblint.patch). Opam applies it only when Js_of_ocaml appears to be available in the opam environment. As a result, switching to this fork should be fairly safe because, in most cases, the code will not differ in any way from the original Zarith code.

### DIFF
--- a/goblint.opam
+++ b/goblint.opam
@@ -50,4 +50,5 @@ dev-repo: "git+https://github.com/goblint/analyzer.git"
 pin-depends: [
   [ "goblint-cil.1.8.1" "git+https://github.com/goblint/cil.git#208d2a2f9e51a42ee0a036f4587624ac7ac23ccb" ]
   [ "ppx_deriving.5.2.1" "git+https://github.com/ocaml-ppx/ppx_deriving.git#0a89b619f94cbbfc3b0fb3255ab4fe5bc77d32d6" ]
+  [ "zarith.1.12-gob0" "git+https://github.com/goblint/Zarith.git#goblint-release-1.12" ]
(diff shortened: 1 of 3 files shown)


michael-schwarz (approved):
LGTM!

sim642 (commented):
It's an impressive technical report of the issue and the fix!

I'm just thinking maybe our fork's opam package name should be `goblint-zarith`, which would define a conflict with `zarith` and still provide the `Z`/`Big_int_Z` module. Just like how `goblint-cil` conflicts with `cil` and still provides the `Cil` module (among others).

**EDIT:** Nevermind, I guess opam wouldn't let setting up a pin for `zarith` to point to a repository containing the `goblint-zarith` package?

But still, shouldn't we be pinning version `1.12-gob0` instead then?

sim642:
> Opam does let you set up such a pin, but it appears to ignore the package file unless the names match because I can no longer get it to trigger a recompile when `js_of_ocaml` is installed.

Ahh, so it fails silently. That's super user-friedly, maybe deserves an issue on opam itself if there already isn't one about it.

sim642 (approved):

PR #321, opened 2021-08-09
OctApron: Support branched thread creation

### LINKED ISSUE #320: OctApron: Mutex-Meet Privatization unsound for threads being created in a branch
A run of SV-Comp has uncovered the following issue: With `Mutex-Meet` as privatization for apron, the assert succeeds even though it shouldn't.

~~~C
// SKIP PARAM: --sets ana.activated[+] octApron --enable ana.int.interval --sets exp.octapron.privatization mutex-meet
#include<pthread.h>
int global = 0;

void *t_fun(void *arg) {
  global = 5;
  assert(1);
}

int main(void) {
  pthread_t t;

  for (int i=0; i<1; i++) {
    pthread_create(&t, ((void *)0), t_fun, ((void *)0));
  }

  assert(global == 0); //UNKNOWN!

  return 0;
}
~~~

Curiously, if one changes
~~~C
  for (int i=0; i<1; i++) {
    pthread_create(&t, ((void *)0), t_fun, ((void *)0));
  }
~~~

to

~~~C
pthread_create(&t, ((void *)0), t_fun, ((void *)0));
~~~

the program is well-behaved.

michael-schwarz:
This handles branched thread creation (the same way it is done for the different base privatizations).

- When joining two threads, side-effect globals to special intializer mutex
- Removing all globals that are unprotected after the join (as we  keep apart locksets, this refers to the globals that were in the local state on the single-threaded path leading to the join)

Closes #320


src/analyses/apron/octApronPriv.apron.ml
```diff
@@ -416,8 +416,27 @@ struct
...
+      else
+        let oct = st.oct in
+        let g_vars = List.filter (fun var ->
+            match V.find_metadata var with
+            | Some (Global _) -> true
+            | _ -> false
+          ) (AD.vars oct)
+        in
+        let oct_side = AD.keep_vars oct g_vars in
+        sideg (mutex_inits ()) oct_side;
```
sim642, line 431: Inspired by `enter_multithreaded` I suppose it makes sense to also side effect to `mutex_inits` here, but curiously the analogous case in basePriv doesn't do this:
https://github.com/goblint/analyzer/blob/9fd923b76a443e2f9e08d1f68f124d6195ed0264/src/analyses/basePriv.ml#L370-L383

Should basePriv mutex-meet also be doing that then?

sim642: This is a broader issue tracked by #327.

sim642 (approved):

PR #322, opened 2021-08-09
Remove `exp.full_context` and `val_of`

michael-schwarz:
I have always found the name `exp.full-context` to be more confusing than it is helpful:
- It somehow implies that enabling this gives more precision, which it does not do over default options. (i.e., if I am unhappy with the results when using one of the partial context options or the defaults, I would expect this option to give more precision).
- The name is only tangentially related to what it actually does: 
  - Instead of side-effecting the initial value for a function call in context `c` to `[c, FunctIonEntry f]`  it generates that value via `val_of: C.t -> D.t`: `[c, FunctionEntry] \sqsupseteq (Spec.val_of c)`
- Only benefit of the name ` full_context`  is that it makes it immediately obvious that it is incompatible with any of the partial context options.

Especially in the light of the work on interactive things likely wanting to expose options for contexts more prominently, I think it is a good time to rename `exp.full-context`.

I would propose to rename `exp.full-context` to `exp.side-entries` with default value `true`. What do you think? (Especially the more senior people @vesalvojdani @vogler?)


sim642:
The verdict from GobCon was to completely remove this option and always use side effects to `FunctionEntry` nodes. In turn, this would allow further cleanup:
- [ ] Remove `val_of` from the `Spec` signature. A function `C -> D` makes no sense in general (when contexts could be partial) and only is proper with full contexts.
- [ ] Remove `WidenContextLifter` and keep `WidenContextLifterSide`.

The use of `exp.full-context` in octApron/poly should be replaced with an analysis-specific option, e.g. `ana.octapron.no-context` (flipped from full context to match other context options in base, which are also negated).

sim642:
The removal of `val_of` only became possible after removing the option `exp.widen-context-partial`. We deemed the removal of this feature fine, because it seems to cause weird behavior:
https://github.com/goblint/analyzer/blob/afe92c44f81abfffd6506f1e757c32b42b30b8ab/src/framework/constraints.ml#L371-L375

When the context returned by `S.context` is partial, then `S.val_of` returned some partial local state. So not only was such `v_new` used to create the new widened context (using `S.context` in `FromSpec`), but the very same `v_new` (without passing through `S.context` in `FromSpec`) also became the local value for the function entry node.

So not only were the contexts widened, but also the local states were made much more imprecise. Normally partial contexts get their non-partial local values for entry nodes and these simply get joined (or widened) together by the side effect, which isn't necessarily as imprecise as a top in the partialized local state.

@vogler What was the reasoning for this? Is there any reason to keep it?

vogler:
> @vogler What was the reasoning for this? Is there any reason to keep it?

I only introduced it to see if it helps for the non-terminating benchmarks #205 (which it did not).
I guess it can be removed.

vogler:
Agree that `exp.side-entries` is a better name.
Re. removing the option: did you discuss what this implies?
Do we have/want solvers that do not support side-effects? For example, for the td3 paper, I ran all benchmarks with `exp.full_context` since side-effects were only introduced as one variation of the solver.
Besides allowing partial contexts and widen-context, when is side-effecting entries beneficial?
I assume there are no cases where `exp.full_context` is better?

sim642:
> Do we have/want solvers that do not support side-effects?

All of the solvers we still have support side effects I think and all our constraint system signatures are also for side-effecting systems. A non-side-effecting solver would be quite limited overall because besides function entry, many analyses use side effects for other things as well, so there aren't many use cases for that.



> Besides allowing partial contexts and widen-context, when is side-effecting entries beneficial?
> I assume there are no cases where `exp.full_context` is better?

Besides keeping the ability to compare the two ways of function entry (which should not make any real difference for demand-driven solving), we didn't see any benefit to keeping the non-side-effecting entry.
And if going forward we wish to do more flexible things with partial contexts, then `exp.full_context` just becomes more and more limited with no advantage.

So if you still might have use for `exp.full_context` (or its renamed version), then we can keep it (it's not too much code). If you think you don't, then we can remove it, because nobody else has a use for it.

vogler:
> Besides keeping the ability to compare the two ways of function entry (which should not make any real difference for demand-driven solving), we didn't see any benefit to keeping the non-side-effecting entry.

I think there were only minor differences in the benchmarks I ran.

> So if you still might have use for `exp.full_context` (or its renamed version), then we can keep it (it's not too much code). If you think you don't, then we can remove it, because nobody else has a use for it.

I think we can remove it then.

vogler:
One situation I just remembered where it could be useful is some slr3 solvers that collect `sides` in order to narrow them.
There the solver would arrive at the start of a function and not have a value yet, wait for `sides` and solve again, whereas with `exp.full_context` you'd get the value right away.
In case one wants to think about narrowing sides in current solvers, it's something to keep in mind (but could probably be improved without having `exp.full_context`).

PR #325, opened 2021-08-12
Make loops strict w.r.t. entry edges

### LINKED ISSUE #233: dead code does not propagate through previously live loops?
There's a theoretical issue with our constraints for loops that Helmut mentioned which might be worth thinking about.
The problem is that we just join all incoming edges at the loop head which may be a problem if there is some value > bot from the back-edge from inside the loop and bot from the initial edge.
If a loop is first live, but then after narrowing the initial edge to the nested loop becomes dead, it won't propagate bot, but join it with the back-edge.
The rhs for a loop should be strict for the initial edge `a`, i.e., `if is_bot a then a else join a b` where `b` is the value of the back edge.

The question is whether this can happen at all.
I tried to come up with an example for loops but couldn't find one -- maybe because `Base.invariant` always limits the values.

If it can occur, the other question is how to improve it.
`Constraints.FromSpec().tf_loop` would have to consider a loop head with two incoming edges and make the join left-strict.
Problem: currently these generate a constraint for each edge of a node which are only afterwards joined into equations by `Generic.SimpleSysConverter`.
Also, one would have to think about loops that have jumps into them with goto (I guess you'd have to also identify all incoming edges into the loop body as initial).

sim642:
Closes #233.

This PR currently has #323 as its base. Based on the identified SCCs, it's possible for `FromSpec` to know, which predecessors are loop entry edges (i.e. SCC entry edges). If those transfer functions are bottom, then the entire value is immediately made bottom, even if non-entry edges (back edges from the end of the loop) would be live. For extra optimization, the latter transfer functions aren't even evaluated if this is the case, because they don't matter.

#### Joining of constraint right-hand sides
`FromSpec` defines the constraint system as `GlobConstrSys`, which has a _list_ of right-hand sides for each (local) unknown – one for each incoming edge. Then `IneqConstrSysFromGlobConstrSys` is used to convert that into a `IneqConstrSys`, which also has _lists_ of right-hand sides, but uniform unknowns (locals and globals are represented by an either type). Finally `SimpleSysConverter` is used to convert that into a `EqConstrSys`, which has _option_ of right-hand side, so here's where the incoming edges are joined together. But at this point everything about the CFG is so far abstracted away, that it's not known, which of those to make particularly strict (without passing this information down through all the functors).

#### Moving the joining up
Hence to achieve the original goal, the joining of different incoming edge constraints must be moved up into `FromSpec` itself, because that's the last place where the CFG is nicely available. The easy hack is to just change `FromSpec` to do the join and return a singleton list.

#### Cleaning up constraint systems
Once that's done, there's no need for lists of right-hand sides, so `GlobConstrSys` and `IneqConstrSys` can be switched to use option instead. Then a bunch of functors for converting all these constraint systems can be removed.

### TODO
- [x] What was supposed to be the difference between `IneqConstrSys` (using list) and `EqConstrSys` (using option)? Because even if the list is joined into an option, the solvers would still give a post-solution, so it still only satisfies the inequality, not exact equality. Now that I changed everything to option anyway, there seems to be no difference, so I guess it's safe to just keep one of them?


src/framework/constraints.ml
```diff
@@ -755,8 +755,31 @@ struct
...
+        let tf' eu = tf (v,c) eu getl sidel getg sideg in
+
+        match NodeH.find_option CfgTools.node_scc_global v with
+        | Some scc when NodeH.mem scc.prev v ->
+          let stricts = NodeH.find_all scc.prev v in
+          let xs_stricts = List.map tf' stricts in
+          if List.for_all S.D.is_bot xs_stricts then
+            S.D.bot ()
+          else
+            let xs_strict = List.fold_left S.D.join (S.D.bot ()) xs_stricts in
```
michael-schwarz, line 772: Could we not just compute `xs_strict` and check that for bottom? Might be more efficient since we expect it to be non-bottom most of the time (and in this case we only go over the list once)

Also, it simplifies the code a bit

michael-schwarz:
Regarding these `slr[0-4]x` solvers that this removes, does anyone recall what they were for and why one would want to use them over `slr[0-4]` (without `x`)?
 @vogler @vesalvojdani @kalmera ?

src/framework/constraints.ml
michael-schwarz: I guess joining a bunch of DeadCode bottoms should also be fairly cheap, but you're right, probably not really worth benchmarking.

sim642:
> Regarding these `slr[0-4]x` solvers that this removes, does anyone recall what they were for and why one would want to use them over `slr[0-4]` (without `x`)?

If I remember correctly, using `ExtendInt` it constructed a constraint system, where the original constraint
```
[v, c] geq (f_1 [u_1, c]) join (f_2 [u_2, c]) join ... join (f_n [u_n, c])
```
is replaced with the following family of constraints
```
[v, c, -1]  geq [v, c, 0] join [v, c, 1] join ... join [v, c, n-1]
[v, c, 0]   geq f_1 [u_1, c, -1]
[v, c, 1]   geq f_2 [u_2, c, -1]
...
[v, c, n-1] geq f_n [u_n, c, -1]
```

I think the point of this is that if [u_1, c] changes, but none of the other predecessors do, then when you reevaluate the rhs for [v, c], then you only have to reevaluate f_1, but none of the other incoming transfer functions. Since their inputs were unchanged, they have stored outputs which can be reused.

At some point I thought about a similar optimization, which might be beneficial if our transfer functions are the expensive part, but if we really want that sort of thing, it should be a general thing instead of being specific to some SLR variants. Of course it's unclear whether this actually optimizes anything, because it requires storing a lot more values for constraint system unknowns. So there's a tradeoff that maybe was explored with SLR at one point.

The downside of this PR, which moves the joining into `FromSpec` instead of returning a list, is that a generic functor cannot be used to transform the constraint system in this way. Rather it'd have to be built somewhat into `FromSpec` itself.

michael-schwarz (approved):

sim642:
Before merging this I'd like to clear up the `IneqConstrSys` vs `EqConstrSys` TODO in the issue.

Also I think there's one subtle thing that moving the join up changes. Namely, previously with a list of right-hand sides, verification checked each one separately and reported the diff for each one separately (if there was one). Now they're all joined, so verification only sees a single joined right-hand side. This might make it harder to decipher verification errors since they don't correspond to a single transfer function at join points any more.

And a related thought: if we give up that `join` gives the _least_ upper bound, then it might be possible that the unjoined right-hand sides wouldn't pass verification, but the single joined one (being overly imprecise) does?

vogler:
> * What was supposed to be the difference between `IneqConstrSys` (using list) and `EqConstrSys` (using option)? Because even if the list is joined into an option, the solvers would still give a post-solution, so it still only satisfies the inequality, not exact equality.

Why not? For both a post-solution is a fixpoint of the resp. constraint system.
I think this is based on the PO lecture slides where first `IneqConstrSys` is introduced and then `EqConstrSys` to allow for narrowing.
Is `IneqConstrSys` used by any solvers?

vogler:
> Namely, previously with a list of right-hand sides, verification checked each one separately and reported the diff for each one separately (if there was one). Now they're all joined, so verification only sees a single joined right-hand side. This might make it harder to decipher verification errors since they don't correspond to a single transfer function at join points any more.

The verification output is already not that easy to decipher...
But depends on how it actually changes it.
If it fails in some loop then the diff is probably coming from the rhs of the back-edge only anyway.

sim642:
> I think this is based on the PO lecture slides where first `IneqConstrSys` is introduced and then `EqConstrSys` to allow for narrowing.
> Is `IneqConstrSys` used by any solvers?

We just discussed this point with @vesalvojdani for another reason. Indeed almost all of the existing solvers use `Eq` and have the right-hand sides joined by a functor already.

I think the only one that doesn't is `effectWCon`. In this PR I just made it convert `option -> list` to typecheck, but arguably it defeats the point of this solver, because there's also `effectWConEq`. I suppose that should make `effectWCon` redundant if it just operates on singleton lists.

sim642:
Actually `effectWCon` doesn't even use `IneqConstrSys`, but `GlobConstrSys` which separately exposes local and global unknowns. But that one also used to give a list of right-hand sides, so effectively it's still like `IneqConstrSys`.

But curiously I now discovered that `TD2` in `topDown_deprecated.ml` was the only one to directly use `IneqConstrSys`. Besides being in the deprecated file, it's not even registered into the solver selector. According to git history, it predates @vogler so looks like it's not relevant for anyone any more.

Issue #32, opened 2016-01-11
configuration constraints


vogler:
These options exclude other options or have dependencies:
- [ ] warn if `ana.int.interval` is enabled but `solver` doesn't do widening:
    - [ ] `EffectWCon` and `Worklist` do not support widening. 
    - [ ] Check other solvers whether they work with loops that modify ints  
- [x] warn if `ana.osek.*` is set but `osek` analysis isn't active
- [x] Fail if one of the privatizations that require exact locksets (or must **and** may locksets) is active, but `mutex` is not path-sensitive

vogler:
Partially done here::
https://github.com/goblint/analyzer/blob/d4b087f478b1a2e0cb9171b27710d3cf8561d851/src/maingoblint.ml#L349-L359

PR #330, opened 2021-08-20
Enforce that Mutex is enabled & path sensitive for privatizations that require it.

michael-schwarz:
All privatizations require the mutex analsysis to be active to be useful. Some require it to be path-sensitive for soundness.

This adds appropriate checks and modifies the settings for the regression tests.

Related #32


src/analyses/apron/octApronPriv.apron.ml
```diff
@@ -674,12 +674,16 @@ end
 
 let priv_module: (module S) Lazy.t =
   lazy (
+    let check_mutex_enabled () =
+      let octapron_active = List.exists (fun x -> Json.string x="octApron") (GobConfig.get_list "ana.activated") in
```
sim642, line 678: Maybe it makes sense to do this check in each `init` instead. That should only get called for activated analyses and would avoid this extra check when registering `octApron`.

PR #331, opened 2021-08-20
Add thread IDs with history

sim642:
Implements the abstract thread IDs proposed by the relational traces paper. The resulting thread IDs know their uniqueness on their own, without needing the thread analysis to construct this via global information.

Also adds a `Thread` variant to base's value domain for a set of thread IDs (for `pthread_t` variables). Since the new abstract thread IDs are no longer just `varinfo`s, one cannot abuse the address domain for this.


src/analyses/base.ml
```diff
@@ -375,6 +375,7 @@ struct
     | `List e -> reachable_from_value ask gs st (`Address (ValueDomain.Lists.entry_rand e)) t description
     | `Struct s -> ValueDomain.Structs.fold (fun k v acc -> AD.join (reachable_from_value ask gs st v t description) acc) s empty
     | `Int _ -> empty
+    | `Thread _ -> empty (* TODO: is this right? *)
```
michael-schwarz, line 378: I think this is the best we can do. Developers should consider it opaque anyway. If people on MacOS access some members of this struct, the code is extremely fragile and will work on MacOS only.

src/analyses/base.ml
```diff
@@ -908,6 +910,12 @@ struct
         | `Bot -> Queries.Result.bot q (* TODO: remove *)
         | _ -> Queries.Result.top q
       end
+    | Q.EvalThread e -> begin
+      match eval_rv (Analyses.ask_of_ctx ctx) ctx.global ctx.local e with
+        | `Thread a -> a
+        | `Bot -> Queries.Result.bot q (* TODO: remove *)
```
michael-schwarz, line 916: When would this happen? For globals?

src/cdomains/threadIdDomain.ml
```diff
@@ -0,0 +1,146 @@
...
+
+  let threadenter _ = threadenter
+  let threadspawn () _ _ = ()
+end
+
+module History (Base: Stateless): Stateful =
+struct
+  module P =
+  struct
+    include Printable.Liszt (Base)
```
michael-schwarz, line 78: Maybe already add a comment here that the prefix is in reverse order?

michael-schwarz (approved):

src/analyses/base.ml
sim642: I'm not sure, hopefully never. I just based it on all the other cases which historically had such case and in #221 I didn't want to remove them (let `Bot` return top query result instead) just in case it somehow matters.

If a thread ID is global, then it should get global initialized to a non-bottom anyway, so that shouldn't be it.

sim642:
Since the `ThreadIdDomain.Thread` module is statically used all over the place, it's not so easy to make it dynamically changeable via an option, since that cannot be initialized at module top level, but must only be done with `AfterConfig` and others cannot statically refer to it.

An alternative would be to do what the array domains do (flag configured): have a module which delegates all functions to the chosen implementation separately.

Issue #332, opened 2021-08-24
Fixpoint not reached when enabling refinement, `def_exc`, and `enums` 


michael-schwarz:
For some programs, the fixpoint is not reached when enabling `def_exc`, `enums` and `refinement` at the same time:

~~~C
//PARAM: --sets ana.int.refinement fixpoint --enable ana.int.enums
int x;

void foo() {
  x--;
}

int main() {
  int y;
  x = y;

  while (x > 0) {
    _Bool c;
    if(c) { foo(); }
    else { foo(); }
  }
}
~~~

~~~
./goblint --enable dbg.debug --enable dbg.showtemps --enable dbg.regression --sets warnstyle "legacy" --html --sets ana.int.refinement once --enable ana.int.enums  tests/regression/38-int-refinements/02-problem.c
Fixpoint not reached at node 8 "x > 0" on tests/regression/38-int-refinements/02-problem.c:12:3
 Solver computed:
 {[Reversed (top or Set (Normal Lvals * booleans)):{},
   top or Set (variables):{}, booleans:False, MT mode:Singlethreaded,
   Thread:main,
   value domain * array partitioning deps * P:(mapping {
                                                 Global {
                                                   x ->
                                                     (Unknown int([-31,31]),not {} ([-31,31]))
                                                 }
                                                 Local {
                                                   y ->
                                                     (Unknown int([-31,31]),not {} ([-31,31]))
                                                   c ->
                                                     (Unknown int([0,1]),not {} ([0,1]))
                                                 }
                                               }, mapping {
                                                      }, {}),
   lifted location:Unknown line, Unit:()]}
 Right-Hand-Side:
 {[Reversed (top or Set (Normal Lvals * booleans)):{},
   top or Set (variables):{}, booleans:False, MT mode:Singlethreaded,
   Thread:main,
   value domain * array partitioning deps * P:(mapping {
                                                 Global {
                                                   x ->
                                                     (Unknown int([-31,31]),not {} ([-31,31]))
                                                 }
                                                 Local {
                                                   y ->
                                                     (Unknown int([-31,31]),not {} ([-31,31]))
                                                   c ->   (0,{0})
                                                 }
                                               }, mapping {
                                                      }, {}),
   lifted location:Unknown line, Unit:()]}
 Difference: [Reversed (top or Set (Normal Lvals * booleans)):{},
              top or Set (variables):{}, booleans:False, MT mode:Singlethreaded,
              Thread:main,
[...]

PR #336, opened 2021-08-25
Fix enums leq

jerhard:
This fixes issue #332. The implementation of `Enums.leq` used to be based on  the `join`, but this is inherently buggy for `Enums`, as the `join` does not always deliver the least upper bound of two elements. 
The new implementation has to perform a case distinction on all four possible cases. 

When checking whether an exclusion set is contained in a inclusion set or another exclusion set, we possibly have to iterate over elements in the range of the abstract value that is supposedly smaller. To avoid having to perform this check as much as possible, we first check whether the cardinality of the concretization of the first abstract value is smaller than the concretization of the second abstract value, similar to what @michael-schwarz  suggested in our discussion today.


michael-schwarz:
Is there a specific reason this is still a draft? If not, I'd suggest merging it!

src/cdomains/intDomain.ml
```diff
@@ -1314,9 +1307,9 @@ struct
     (* A definite value is leq all exclusion sets that don't contain it *)
     | `Definite x, `Excluded (s,r) -> in_range r x && not (S.mem x s)
     (* No finite exclusion set can be leq than a definite value *)
-    | `Excluded _, `Definite _ -> false
+    | `Excluded _, `Definite _ -> false (* wrong *)
     (* Excluding X <= Excluding Y whenever Y <= X *)
-    | `Excluded (x,xw), `Excluded (y,yw) -> S.subset y x && R.leq xw yw
+    | `Excluded (x,xw), `Excluded (y,yw) -> S.subset y x && R.leq xw yw (* wrong *)
```
sim642, line 1312: I'm guessing it's a draft because of these "wrongs"?

michael-schwarz: Yes, good catch!

jerhard:
I now moved the handling of comparing two Exclusion sets with `leq` into a new module `Exclusion`, as well as the handling of `leq exclusion_set inclusion_set`.

michael-schwarz (approved):
LGTM!

PR #337, opened 2021-08-31
Add semgrep for custom scanning of Goblint code

sim642:
I heard about this [semgrep](https://semgrep.dev/) tool and tried it out on Goblint's own code since it supports OCaml. It allows writing custom code scanning rules (e.g. specific to our tracing like I did here) fairly easily and running them on the CI (with SARIF output to GitHub Security tab!). Semgrep itself is open-source and runs independently of whatever (possibly paid) services they offer on their own, so there shouldn't be excessive vendor lock-in.

This might be useful if we can encode common student code issues using semgrep rules for future automatic checking. If anyone has any other rule ideas right away, feel free to suggest them.

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,51 @@
+name: Semgrep
+
+on:
+  # Scan changed files in PRs, block on new issues only (existing issues ignored)
+  pull_request: {}
+
+  # Scan all files on branches, block on any issues
+  # push:
+  #   branches: ["master", "main"]
+
+jobs:
+  semgrep:
+    name: Scan
+    runs-on: ubuntu-latest
+    # Skip any PR created by dependabot to avoid permission issues
+    if: (github.actor != 'dependabot[bot]')
+    steps:
+      # Fetch project source
+      - uses: actions/checkout@v2
+
+      - uses: returntocorp/semgrep-action@v1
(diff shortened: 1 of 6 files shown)


michael-schwarz:
One issue that I see quite often is students misusing physical (dis)equality where they should use structural (dis)equality. But it seems hard to capture in a rule, as there are legitimate uses of physical equality too.

sim642:
> One issue that I see quite often is students misusing physical (dis)equality where they should use structural (dis)equality. But it seems hard to capture in a rule, as there are legitimate uses of physical equality too.

We could do something similar to what Jane Street's Core/Base library does: https://ocaml.janestreet.com/ocaml-core/v0.12/doc/base/Base/Poly/index.html. That is, make a separate module with those functions and every intentional physical compare must call those, e.g. `Phys.equal a b` instead of `a == b`. Then one can detect all uses of `==` and exclude the wrapper in `Phys`.

One would have to go over all the existing physical equality uses and confirm that they are intentional though. They're easy to find using semgrep (locally) though :smile:
```
semgrep -l ocaml -e '... == ...' .
```

Something similar might also be reasonable for polymorphic (structural) compare, but that's more annoying, because you'd want to use `=` on `int`s I guess. Unless we go all in and use `Int.equal` like it's just another abstract type.

PR #339, opened 2021-09-01
Add `--sets` behavior to `--set`

sim642:
Adds plain string support to `--set` instead of requiring quotes. This should allow replacing all `--sets` usage with just `--set` (works for everything in this repository at least).

Still keeps `--sets` around but prints a deprecation warning in case some external scripts still use it.

The implementation isn't based on looking up the JSON type for the given option, but rather just adds a lexing/parsing failure fallback.


docs/artifact-descriptions/sas21.md
```diff
@@ -91,7 +91,7 @@ presented in the paper quickly.
         or
 
     * Run some of the regression tests in `tests/regression/13-privatized` by calling `./regtest.sh 13 xx` where `xx` is the number of the test. Especially `xx > 16` are interesting, these were added with the paper and highlight
-      differences between different approaches. Use the `--sets exp.privatization chosenname` option to choose which thread-modular analysis to use.
+      differences between different approaches. Use the `--set exp.privatization chosenname` option to choose which thread-modular analysis to use.
```
michael-schwarz, line 94: Do we want to update this document? Or should it be static such that it works for the artifact linked above?

sim642: That's a good question. I think we should update the things in our repository such that they remain usable as long as possible. If one wants to use the artifact specifically, then it includes the instructions at the time. But the instructions now should reflect how to use the same stuff with the current version of Goblint (especially if we remove `--sets` completely).

This would be similar to how we handle `svcomp21.json`: if you want the exact setup from SVCOMP21, then use that verifier archive and the conf in there, if you want more-or-less equivalent configuration from the current version of Goblint, then you can use the conf from the current master.

It's not reasonable to keep these dated things locked in completely because as we fix things within Goblint, then behavior might still change. The only way to get guaranteed reproducibility is to use respective archived versions (or check out tags from git).

sim642: I'll add a comment to this artifact description that if you're using the artifact, it includes the exact instructions there to make it clear that our documentation version needn't be identical any more.

tests/regression/00-sanity/13-sans_context.c
```diff
@@ -1,4 +1,4 @@
-// PARAM: --set ana.activated "['escape', 'base','threadid','threadflag','mallocWrapper']" --set ana.ctx_insens[+] "'base'" --sets exp.privatization none
+// PARAM: --set ana.activated "['escape', 'base','threadid','threadflag','mallocWrapper']" --set ana.ctx_insens[+] "'base'" --set exp.privatization none
```
vogler, line 1: `--set ana.ctx_insens[+] base` since other `[+]` have no quotes as well?

tests/regression/23-partitioned_arrays_last/13-advantage_for_last.c
```diff
@@ -1,4 +1,4 @@
-// PARAM: --sets solver td3 --enable ana.int.interval  --sets exp.partition-arrays.keep-expr "last" --enable exp.partition-arrays.enabled --set ana.activated "['base','threadid','threadflag','escape','expRelation','octagon','mallocWrapper']" --sets exp.privatization none
+// PARAM: --set solver td3 --enable ana.int.interval  --set exp.partition-arrays.keep-expr "last" --enable exp.partition-arrays.enabled --set ana.activated "['base','threadid','threadflag','escape','expRelation','octagon','mallocWrapper']" --set exp.privatization none
```
vogler, line 1: `last` instead of `"last"`?

tests/sv-comp/observer/fake_true-unreach-call.c
```diff
@@ -1,4 +1,4 @@
-// --sets ana.activated[+] "observer" --sets ana.path_sens[+] "observer"
+// --set ana.activated[+] "observer" --set ana.path_sens[+] "observer"
```
vogler, line 1: `--set ana.path_sens[+] observer` since other `[+]` have no quotes as well?

PR #342, opened 2021-09-02
Replace our own JSON library with Yojson

### LINKED ISSUE #340: Replace our own JSON library, lexer and parser with Yojson
We already integrate Yojson, so we might as well use it for everything. Currently the config system uses our own JSON structures and we have our own JSON lexer and parser.

There's also `JsonSchema` that checks this custom JSON structure, but if we want to keep that, I guess it could still be adapted to just look at a Yojson structure instead.

sim642:
Closes #340.

### TODO
- [x] Rewrite config for Yojson.
- [x] Rewrite JSON schema validation for Yojson.
- [x] Rewrite contain analysis for Yojson.
- [x] Remove our own JSON type and library.


sim642:
Besides the mutability difference mentioned here https://github.com/goblint/analyzer/issues/340#issuecomment-911447547, there's another difference between ours and Yojson. Ours used a `Map.Make` for a JSON object while Yojson uses an association list. In theory the lookups are slower with Yojson's representation, but I doubt it's an issue because the objects in our configuration aren't large.

Also, since there's memoization in front of option getting functions, then it won't even matter for repeated lookups because only the first one traverses the JSON structure, the following lookups of the same option hit the cache.

src/analyses/commonPriv.ml
```diff
@@ -126,16 +126,16 @@ struct
   module RequireMutexActivatedInit =
   struct
     let init () =
-      let analyses = GobConfig.get_list "ana.activated" in
-      let mutex_active = List.exists (fun x -> Json.string x="mutex") analyses || List.for_all (fun x -> Json.string x<>"base") analyses in
+      let analyses = GobConfig.get_string_list "ana.activated" in
+      let mutex_active = List.mem "mutex" analyses || not (List.mem "base" analyses) in
```
michael-schwarz, line 130: Nice catch, my code was a bit over complicated here :roll_eyes:

michael-schwarz (approved):

src/analyses/commonPriv.ml
sim642: Surprisingly it was complicated like this everywhere. I guess when `get_string_list` was added for convenience, the existing uses of `get_list` weren't reviewed.

PR #343, opened 2021-09-02
Rewrite instructions in README

sim642:
The current instructions in README are outdated and quite a mess. For example, to run `make setup` you already need to have installed opam, but that's only mentioned under the alternative outdated `opam install goblint`.

The web frontend instructions are also more explicit in the actual documentation now, so no need to duplicate those in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ Documentation can be browsed on [Read the Docs](https://goblint.readthedocs.io/e
 ## Installing
 Both for using an up-to-date version of Goblint or developing it, the best way is to install from source by cloning this repository.
 
+_The [goblint package on opam](https://opam.ocaml.org/packages/goblint/) is very outdated and should currently not be used._
+
 ### Linux
 1. Install [opam](https://opam.ocaml.org/doc/Install.html).
 2. Run `make setup` to install OCaml and dependencies via opam.
--- a/README.md
+++ b/README.md
@@ -1,44 +1,37 @@
-# goblint
+# Goblint
 [![locked workflow status](https://github.com/goblint/analyzer/actions/workflows/locked.yml/badge.svg)](https://github.com/goblint/analyzer/actions/workflows/locked.yml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,44 +1,39 @@
...
+1. Install [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10).
+2. Continue using Linux instructions in WSL.
 
-### Web frontend
-The analysis results are printed to stdout by default.
-Adding `--html` saves the results as XML, which is then transformed to be viewable in a web browser.
-Use `make jar` to build the needed Java program.
+### Other
+* **[Vagrant](https://www.vagrantup.com/).** Run `vagrant up && vagrant ssh`.
+* **[Docker](https://hub.docker.com/r/voglerr/goblint)** (outdated). Run `docker run -it voglerr/goblint bash`.
```
sim642, line 29: The last build on Docker Hub is a month old by now, probably because the free Docker Cloud tier was shut down, so no new ones are automatically built and published.

Maybe we should just have here the instructions for building the Docker image from the cloned repository? I don't know the right way to do that.

michael-schwarz: @vogler ?

vogler: > Maybe we should just have here the instructions for building the Docker image from the cloned repository? I don't know the right way to do that.

`make docker`?

vogler: > The last build on Docker Hub is a month old by now, probably because the free Docker Cloud tier was shut down, so no new ones are automatically built and published.

I applied for their Open Source Program but they are very slow to respond. Got this yesterday despite having set the license in their form (I see no license field on the website):

 "Thanks again for applying to Docker's Open Source Program. We noticed that you didn't specify the open source license under which your image namespace is under. As a reminder, to qualify for the program, images need to be under a license approved by the OSI."

sim642: I added `make docker` as a non-outdated variant and left Docker Hub as an outdated one for now. This should be clear enough that it's possible to get a Docker build with the right environment, but not to suggest the image from Docker Hub first.

Issue #186, opened 2021-04-09
Uniqueness analysis for mallocs


vesalvojdani:
Would be good to have this sooner rather than later. Goblint can't analyze Juliet suite race detection stuff because they use some wrappers that end up creating pointers to structs containing the mutex. One could obviously just take their lock operations as native, but these are much simpler programs than all the idioms we handle with symbolic locks. It does not seem unreasonable to think that we should be able to analyzer the following.

```c
struct { pthread_mutex_t mutex; } *mylock;
int myglobal;

void *t_fun(void *arg) {
  pthread_mutex_lock(&mylock->mutex);
  myglobal=myglobal+1;
  pthread_mutex_unlock(&mylock->mutex);
  return NULL;
}

int main(void) {
  pthread_t id;
  mylock = malloc(sizeof(*mylock));
  pthread_mutex_init(&mylock->mutex, NULL);

  pthread_create(&id, NULL, t_fun, NULL);
  pthread_mutex_lock(&mylock->mutex);
  myglobal=myglobal+1;
  pthread_mutex_unlock(&mylock->mutex);
  pthread_join (id, NULL);
  return 0;
}
```

michael-schwarz:
This amounts to checking that the `malloc(...)` is reached exactly once as far as I can see.

If it suffices to find this information for `mallocs` that happen in single-threaded mode with `earlyglobs` off, this should be fairly simple:

It would suffice to equip the blob with a boolean `is_unique`. Upon the call to malloc in single-thread mode:
- If there already is a blob associated with this allocation site in the local state: Set its `is_unique` to `false`
- If there is not, create it with `is_unique`  `true`.

When in multi-threaded mode: Create it immediately with `is_unique` false.
Upon going multithreaded, one should simply be able to publish it without any further complications.

You have an undergrad student on these Juliet things, right? Maybe they can also implement this?

sim642:
It seems like a slight variation of `thread` analysis could work for the multithreaded case as well. And it would avoid having to hammer yet another feature into `base` analysis to manage the uniqueness flag in blobs.

michael-schwarz:
A student of ours is implementing this as part of their practical course (special 1-participant instance of the course).

michael-schwarz:
We should investigate how much of this is resolved by #722.

sim642:
[...]

PR #345, opened 2021-09-08
Make malloc-ed variables non-unique

### LINKED ISSUE #316: Unsoundness for comparing pointers to malloced values
While fiddling around with #310, I discovered this issue:

~~~C
#include<stdlib.h>
#include<stdint.h>

void* malloc_2(size_t s) {
    return malloc(s);
}

int main() {
    int* ptr1 = malloc_2(sizeof(int));
    int* ptr2 = malloc_2(sizeof(int));

    // This is technically UB (comparing unrelated pointers)
    assert(ptr1==ptr2);

    // CIL turn this into the following, that is not UB and will fail in the concrete
    // Goblint claims that this holds
    assert((unsigned long) ptr1 == (unsigned long) ptr2); // UNKNOWN!
    
    // Here, we do not claim it holds, as we cast our abstract values to the type for ints on assignment
    int i1 =  (int)ptr1;
    int i2 =  (int)ptr2;
    assert(i1 == i2);

    return 0;
}
~~~

Not sure how high priority fixing this is, but I think it is definitely worth having an issue for it. This could be fixed quite easily once we have uniqueness information.

sim642:
Closes #316.

This implements the solution proposed in https://github.com/goblint/analyzer/pull/310#discussion_r700056687. By answering `IsMultiple` query for malloc-ed variables with false, the address set equality check introduced in #310 automatically considers this non-uniqueness information.

This should be sound and good enough until we design a more precise malloc uniqueness analysis (#186).

### DIFF
--- a/tests/regression/02-base/60-malloc-nonrefl.c
+++ b/tests/regression/02-base/60-malloc-nonrefl.c
@@ -9,11 +9,11 @@ int main() {
     int* ptr1 = malloc_2(sizeof(int));
     int* ptr2 = malloc_2(sizeof(int));
 
-    // This is technically UB (comparing unrelated pointers)
-    assert(ptr1==ptr2); // UNKNOWN!
+    // will fail in the concrete
+    assert(ptr1==ptr2); // UNKNOWN
 
-    // CIL turn this into the following, that is not UB and will fail in the concrete
-    assert((unsigned long) ptr1 == (unsigned long) ptr2); // UNKNOWN!
+    // CIL turns this into the following
+    assert((unsigned long) ptr1 == (unsigned long) ptr2); // UNKNOWN
 
     // Here, we do not claim it holds, as we cast our abstract values to the type for ints on assignment
     int i1 =  (int)ptr1;
--- a/tests/regression/02-base/60-malloc-nonrefl.c
+++ b/tests/regression/02-base/60-malloc-nonrefl.c
@@ -1,0 +1,25 @@
+#include<stdlib.h>
+#include<stdint.h>
+
+void* malloc_2(size_t s) {
+    return malloc(s);
+}
+
+int main() {
+    int* ptr1 = malloc_2(sizeof(int));
+    int* ptr2 = malloc_2(sizeof(int));
+
+    // This is technically UB (comparing unrelated pointers)
+    assert(ptr1==ptr2);
+
+    // CIL turn this into the following, that is not UB and will fail in the concrete
+    // Goblint claims that this holds
+    assert((unsigned long) ptr1 == (unsigned long) ptr2); // UNKNOWN!
+
+    // Here, we do not claim it holds, as we cast our abstract values to the type for ints on assignment
+    int i1 =  (int)ptr1;
+    int i2 =  (int)ptr2;
+    assert(i1 == i2);
+
+    return 0;
+}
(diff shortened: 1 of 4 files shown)


sim642:
Initially this PR failed marshaling regression tests for macOS in the unlocked job.

This unusual behavior seemed to be due to the `heap_vars` hashtable used by mallocWrapper to record which `varinfo`s correspond to mallocs. The problem is that such global (within the module) hashtables are not marshaled and unmarshaled.
Therefore after unmarshaling that hashtable is initialized to be empty and everything is considered a program variable.

The possible reason why it still worked on Ubuntu is that depending on the order of keys in the solution hashtables, verification reevaluates the transfer functions in non-deterministic order. If it happens to evaluate the `malloc` edge before whereever the non-uniqueness of this variable is used, then that would have repopulated the mallocWrapper hashtable and appear to still work.
Something about the macOS environment and the hashtable implementation caused it to differ from Ubuntu. Clearly this non-determinism is bad.

The solution for now was to use the mallocWrapper global invariant, which also has `varinfo` keys with a boolean lattice, in place of this hashtable. Then it's properly marshaled and unmarshaled as part of the solution.
This required undoing part of #214 to be able to do the corresponding `sideg` during a `query` which is used to ask for a malloc variable. The original `part_access` stuff still works around it, so the grouping that was fixed there still works, but now the dangerous ungrouped `sideg` (only used here) is possible again.

The other `heap_hash` table in mallocWrapper is safe for extremely fragile reasons: `Goblintutil.create_var` is deterministic w.r.t the variable name (here including the location), so it gets identically reconstructed. Moreover, this cannot be moved into the analysis global invariant because it uses `location` keys.

This exposes a much deeper marshaling problem that's not specific to here. For example, even the `RETURN` varinfo created by base analysis is similarly imperatively created and globally stored in the base module. It is also not marshaled and the name-determinism of `create_var` just hides this issue by allocating a new `varinfo`, which happens to behave well with the unmarshaled solution containing `RETURN` since they have the same `vid`.

So _any_ kind of state in the analysis modules is problematic for marshaling and should maybe be avoided or replaced with a mechanism that allows it to also be marshaled from all analyses.
This also affects the amazing `RichVarinfo`, which currently doesn't expose the internal hashtable at all, but would have to for marshaling (assuming we want to marshal closures), which largely defeats the point of `RichVarinfo`.

tests/regression/02-base/60-malloc-nonrefl.c
```diff
@@ -0,0 +1,25 @@
...
+
+int main() {
+    int* ptr1 = malloc_2(sizeof(int));
+    int* ptr2 = malloc_2(sizeof(int));
+
+    // This is technically UB (comparing unrelated pointers)
+    assert(ptr1==ptr2); // UNKNOWN!
+
+    // CIL turn this into the following, that is not UB and will fail in the concrete
+    // Goblint claims that this holds
```
michael-schwarz, line 16: Maybe update this comment?

michael-schwarz (approved):

tests/regression/02-base/60-malloc-nonrefl.c
sim642: Are the other comments about UB still right? We discussed this in Munich but I don't remember what the conclusion was and this test probably wasn't updated.

michael-schwarz: > Are the other comments about UB still right?

No, they are wrong. In fact, `==` and `!=` can be used to compare any pointers of compatible types. It is only other comparisons (`<`,`<=`, ...) that are undefined.

PR #347, opened 2021-09-16
Add CITATION file

sim642:
GitHub seems to have this new feature, where citation information can be added to a repository: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files.

I decided to try it out and make the citation to be the same ASE 2016 paper as our website suggests: http://goblint.in.tum.de/downloads. **Do we still want this or maybe a different one or maybe have no `preferred-citation` at all and have citation of the code repository specifically?**

You can see it in action on the right here: https://github.com/goblint/analyzer/tree/cff.
It's still quite limited as the CFF → BibTex conversion doesn't add all the data, so the output is just this:
```bibtex
@inproceedings{Vojdani_Static_race_detection,
author = {Vojdani, Vesal and Apinis, Kalmer and Rõtov, Vootele and Seidl, Helmut and Vene, Varmo and Vogler, Ralf},
doi = {10.1145/2970276.2970337},
pages = {391--402},
title = {{Static race detection for device drivers: the Goblint approach}}
}
```
Slightly weird since it doesn't even have the conference name and year, but at least the DOI is there.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,25 +1,32 @@
 cff-version: 1.2.0
 
-preferred-citation:
-  type: conference-paper # this corresponds to inproceedings in BibTex: https://github.com/citation-file-format/ruby-cff/blob/d5b8dac597946d01e52967cf160325bf6218f9f6/lib/cff/formatter/bibtex_formatter.rb#L86-L87
-  authors:
-    - given-names: Vesal
-      family-names: Vojdani
-    - given-names: Kalmer
-      family-names: Apinis
-    - given-names: Vootele
-      family-names: Rõtov
-    - given-names: Helmut
-      family-names: Seidl
-    - given-names: Varmo
-      family-names: Vene
-    - given-names: Ralf
-      family-names: Vogler
-  title: "Static race detection for device drivers: the Goblint approach"
-  collection-title: "Proceedings of the 31st IEEE/ACM International Conference on Automated Software Engineering, ASE 2016" # not used by GitHub
-  start: 391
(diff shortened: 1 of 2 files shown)


vesalvojdani:
That reference should be used for Goblint as a race detection tool. It would be good to get a zenodo reference for our official releases and use those to "cite the repository".

sim642:
We can do that once we actually make a release for sure. But until then should we:

1. have no CITATION file and "Cite this repository"
 
   or

2. remove the `preferred-citation` and add something generic (like name, list of authors) for the time being, to easily extend once we make a release and can add version number/doi?

sim642:
We also made one for SV-COMP 2021, but both of those aren't really artifacts of the repository itself, so it doesn't make sense to have the repository's DOI link to either. New releases of Goblint wouldn't be published as new versions of either Zenodo artifact, which has one DOI for the artifact (regardless of version) and one DOI per version.

sim642:
In preparation for a release (both on opam and Zenodo) I've now rewritten the CITATION file for Goblint in general. This should be automatically used by Zenodo once I set it up to publish on GitHub releases and do the opam release.

michael-schwarz (approved):
I suggest we add `Sarah Tilscher` to the set of authors. Apart from that, LGTM!

michael-schwarz:
>  So I was thinking we maybe don't include the newest authors for this release, but add them after this release for Goblint "lean" 2.0.0. 

Ok, makes sense!

PR #349, opened 2021-09-20
Optimize TD3 to not unnecessarily store bottom for just eval-ed variables

sim642:
This was revealed by #345.
Simply `get`-ing/`eval`-ing global constraint variables added bottom to `rho`.
This unnecessarily increases rho with default values.

This PR changes TD3 such that it doesn't store those bottoms eagerly when a constraint variable is just read and never receives a real value. Thus most `HM.find rho` are wrapped with catching `Not_found` to return bottom. Is there maybe some reason I'm not aware of why TD3 doesn't use implicit bottom values already?

There were a few places where `HM.mem rho` was used, for which it matters whether the bottom value has been added or not. From what I can tell, these are just for the _space_ variant and actually correspond to `HM.mem wpoint`, which is how those checks are presented in the TD3 paper.

I added two static variables to tune the bottom insertion behavior of `init` such that the old behavior can easily be toggled as well.

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -158,16 +158,16 @@ module WP =
         | Some f -> f get set
       and simple_solve l x y =
         if tracing then trace "sol2" "simple_solve %a (rhs: %b)\n" S.Var.pretty_trace y (S.system y <> None);
-        if S.system y = None then (init y; HM.find rho y) else
-        if HM.mem rho y || not space then (solve y Widen; HM.find rho y) else
-        if HM.mem called y then (init y; HM.remove l y; HM.find rho y) else
-        (* if HM.mem called y then (init y; let y' = HM.find_default l y (S.Dom.bot ()) in HM.replace rho y y'; HM.remove l y; y') else *)
+        if S.system y = None then (init y; add_infl y x; HM.find rho y) else
+        if HM.mem rho y || not space then (solve y Widen; add_infl y x; HM.find rho y) else
+        if HM.mem called y then (init y; HM.remove l y; add_infl y x; HM.find rho y) else
+        (* if HM.mem called y then (init y; let y' = HM.find_default l y (S.Dom.bot ()) in HM.replace rho y y'; HM.remove l y; add_infl y x; y') else *)
         if cache && HM.mem l y then HM.find l y
         else (
           HM.replace called y ();
           let tmp = eq y (eval l x) (side x) in
           HM.remove called y;
-          if HM.mem rho y then (HM.remove l y; solve y Widen; HM.find rho y)
+          if HM.mem rho y then (HM.remove l y; solve y Widen; add_infl y x; HM.find rho y)
           else (if cache then HM.replace l y tmp; tmp)
         )
       and eval l x y =
@@ -175,7 +175,6 @@ module WP =
         get_var_event y;
         if HM.mem called y then HM.replace wpoint y ();
         let tmp = simple_solve l x y in
-        if HM.mem rho y then add_infl y x;
         tmp
       and side x y d = (* side from x to y; only to variables y w/o rhs; x only used for trace *)
         if tracing then trace "sol2" "side to %a (wpx: %b) from %a ## value: %a\n" S.Var.pretty_trace y (HM.mem wpoint y) S.Var.pretty_trace x S.Dom.pretty d;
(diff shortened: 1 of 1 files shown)


src/solvers/td3.ml
```diff
@@ -156,27 +163,23 @@ module WP =
...
+        get_var_event y;
+        if HM.mem called y then HM.replace wpoint y ();
         if tracing then trace "sol2" "simple_solve %a (rhs: %b)\n" S.Var.pretty_trace y (S.system y <> None);
-        if S.system y = None then (init y; HM.find rho y) else
-        if HM.mem rho y || not space then (solve y Widen; HM.find rho y) else
-        if HM.mem called y then (init y; HM.remove l y; HM.find rho y) else
-        (* if HM.mem called y then (init y; let y' = HM.find_default l y (S.Dom.bot ()) in HM.replace rho y y'; HM.remove l y; y') else *)
+        if S.system y = None then (init y; add_infl y x; try HM.find rho y with Not_found -> S.Dom.bot ()) else
+        if not space || HM.mem wpoint y then (solve y Widen; add_infl y x; try HM.find rho y with Not_found -> S.Dom.bot ()) else
+        if HM.mem called y then (init y; HM.remove l y; add_infl y x; try HM.find rho y with Not_found -> S.Dom.bot ()) else
```
sim642, line 173: This `HM.mem called y` case doesn't exist in the TD3 paper. Is it some optimization? If so, then maybe there should be an explanatory comment.

src/solvers/td3.ml
```diff
@@ -227,7 +230,9 @@ module WP =
         if tracing then trace "sol2" "init %a\n" S.Var.pretty_trace x;
         if not (HM.mem rho x) then (
           new_var_event x;
-          HM.replace rho x (S.Dom.bot ())
+          if init_bot_all || (init_bot_rhs && S.system x <> None) then
+            HM.replace rho x (S.Dom.bot ())
+          (* TODO: fix new_var_event repeatedly triggering when init-ing multiple times without value *)
```
sim642, line 235: it looks like the solver variable counting happens very primitively by incrementing a counter:
https://github.com/goblint/analyzer/blob/5a434cdd4faab2d3a711d246d71f9b27f043f51d/src/solvers/generic.ml#L207-L209
Seems to me like this could just be `HM.length rho`, which would fix the new issue.

michael-schwarz:
To recap from Gobcon yesterday: The bottoms that are inserted are meaningful in that the hashtable for a partial (post-solution) should be closed, i.e. not depend on any implicit bottoms for any unknown.
If this is the case, the domain of the hashtable actually contains the information that these are the only unknowns that were queried by the solver during the computation of the fixpoint.

From a purely practical standpoint, I guess the question is whether one wants to introduce an option that still makes this optimization, in case one is not interested in the domain. I think this something we should then only consider if it leads to improved performance of the solver, quicker marshaling, or some other measurable improvement.

sim642:
> To recap from Gobcon yesterday: The bottoms that are inserted are meaningful in that the hashtable for a partial (post-)solution should be closed, i.e. not depend on any implicit bottoms for any unknown.
> If this is the case, the domain of the hashtable actually contains the information that these are the only unknowns that were queried by the solver during the computation of the fixpoint.

In the _space_ variant, the solution hashtable isn't closed anyway and missing keys might even correspond to non-bottom values. And the paper goes through all the trouble to define the closure of such solution to make it meaningful. Surely the reasoning for closing only bottoms is much simpler since it requires no transfer function evaluations (and resolving between _wpoint_).
So I don't think it makes things any more sophisticated than it already is.

Moreover, we already make this assumption of implicit bottoms in the partial solution all over the place: `Verify2`, `Reachability`, `Control`, etc. They all wrap solution lookups in such a way, so this behavior has clearly been part of Goblint's design.

michael-schwarz:
Do we want to do something with this or close it? I would vote for only merging this if we can show it to have significant performance impact.

michael-schwarz:
@stilscher @jerhard @sim642 Opinions on whether we should mege (with a special option) or close this?

sim642:
I think it's not worth complicating things by having another option for this feature, which probably made no performance difference. But I'll have a closer look before closing.

sim642:
I cherry-picked 1f4bd6e9db07526c028aab14e4c3c6e0b59ba3b8 onto master because it's orthogonal cleanup, but let's leave the rest for now.

Issue #346, opened 2021-09-10
Fix overflow and dead branch warnings in Juliet tests


sim642:
We had a bachelor's student look at Goblint's behavior on Juliet suite and he also created some simplified tests, which I committed here in https://github.com/goblint/analyzer/commit/91d35137279af2ca8c5394348edb5af856e3bfcf.

None of them pass for various reasons, but since #338 it would be possible to add CWE-specific warnings:
- [x] Interval analysis only warns about signed overflow, since that's what SV-COMP wanted, but now we should also warn (with different tags) about unsigned overflow.
- [x] We don't distinguish overflow from underflow, but CWE does. For precise warnings we should as well.
- [x] Even if temporarily removing the `isSigned` check, we don't warn about `char` overflows. Probably because arithmetic with `char` is performed as `int` and then cast back. So overflow in cast should warn as well.
- [x] `dbg.print_dead_code` prints some warnings about always true/false conditions, but they have the wrong location: they are printed at the first dead line in the branch body, not the `if` itself. I guess technically for Juliet suite it doesn't matter (?), but it would indeed make more sense to print it at the `if` since the warning contains the expression.
- [x] Conditional expression evaluation sometimes is less precise than `invariant`. Dead branch warnings are printed based on the former, but in 50/08 nothing is printed despite the body being a dead node. Apparently it only becomes dead after refining using `invariant`. _This is semi-intentional: `invariant` is more expensive than `eval_rv` (on base), so all of its logic maybe shouldn't be fallback for `eval_rv`._
- [x] Always true condition warnings aren't printed when the `if` has no else branch or it is empty. Dead branch warnings are somehow tracked separately but also checked using dead lines (?). Since there's no dead node/line in the else branch, this is not detected, despite `branch` raising `Deadcode`.

sim642:
All tasks completed by #353 and #438 together.

PR #353, opened 2021-09-21
Extract and fix dead branch detection

sim642:
This PR fixes some points in issue #346:
1. Extract dead branch detection from base to `Spec` functor.

   This accounts for `Deadcode` raised by other analyses than base,
   but also cases where base's `eval_rv` is less precise than `invariant`.

2. Fix locations for dead branches.

   Use if's location instead of first branch node's.
   This also makes always true conditions work when false branch is missing.


src/framework/constraints.ml
```diff
@@ -991,6 +991,39 @@ struct
...
+
+  let locmap_modify_def d k f h =
+    if Locmap.mem h k then
+      Locmap.replace h k (f (Locmap.find h k))
+    else
+      Locmap.add h k d
+
+  let branch ctx exp tv =
+    if !GU.in_verifying_stage && get_bool "dbg.print_dead_code" then (
+      Locmap.replace Deadcode.dead_branches_cond !Tracing.current_loc exp;
```
michael-schwarz, line 1012: I know this PR didn't change it, but why is this called `dead_branches_cond`? Does it not somehow collect the conditions of all branches?

src/framework/constraints.ml
```diff
@@ -991,6 +991,39 @@ struct
...
+      try
+        let r = branch ctx exp tv in
+        (* branch is live *)
+        Locmap.replace (dead_branches tv) !Tracing.current_loc false;
```
michael-schwarz, line 1016: Why do we add live branches to this data structure?

src/framework/constraints.ml
```diff
@@ -991,6 +991,39 @@ struct
...
+        r
+      with Deadcode ->
+        (* branch is dead *)
+        locmap_modify_def true !Tracing.current_loc (fun x -> x) (dead_branches tv);
```
michael-schwarz, line 1020: what is the deal with `fun x -> x` here? This is the only place where `locmap_modify_def` is called, why does it need such a weird argument?

src/framework/constraints.ml
michael-schwarz: I guess this is to only report branches as dead that are dead in all contexts (and also paths)?
If yes, we should comment this here.

src/framework/control.ml
```diff
@@ -21,6 +21,7 @@ let get_spec () : (module Spec) =
             |> lift (get_bool "ana.sv-comp.enabled") (module HashconsLifter)
             |> lift (get_bool "ana.sv-comp.enabled") (module WitnessConstraints.PathSensitive3)
             |> lift (not (get_bool "ana.sv-comp.enabled")) (module PathSensitive2)
+            |> lift true (module DeadBranchLifter)
```
sim642, line 24: TODO: make this conditional on the dead code printing option?

src/framework/constraints.ml
sim642: Yes, because the dead branch printing in `Control` is done purely by location so it wouldn't know what conditions to print otherwise.

src/framework/constraints.ml
sim642: This functor is now applied outside of `PathSensitive2` (previously it was inside in base), but path-wise it makes no difference because `PathSensitive2` aggregates the `Deadcode` exceptions correctly for us.
So it's just for multiple contexts in deed.

src/framework/constraints.ml
sim642: This is related to the multiple contexts thing above. When a branch is found to be live, then it's deadness is always overwritten to `false`. When a branch is found to be dead here, then it's set to dead (`true`) if it didn't exist in the map before, otherwise the identity function keeps it as it was. If it has always been dead (`true`) then it stays that way, if in some case it was made live (`false`), we don't want to overwrite it, but preserve the liveness.

michael-schwarz: Yes, I agree this is what we want to do. However, this `fun x-> x` somehow obfuscates that more than it helps.

sim642: I added comments explaining this and replaced `locmap_modify_def` with the equivalent `Locmap.modify_def`.

PR #354, opened 2021-09-21
Fix analysis auxiliary data marshaling

### LINKED ISSUE #352: Allow auxiliary analysis-specific data to be marshaled
As discovered in https://github.com/goblint/analyzer/pull/345#issuecomment-916779889, marshaling is extremely fragile w.r.t. data created or stored outside of the solution.

All auxiliary varinfos (including `RETURN` and `RichVarinfo`) should be explicitly marshaled and unmarshaled in analyses that use them.
The reasonable approach would be to extend `Spec`'s `init` and `finalize`.

I already started something in https://github.com/goblint/analyzer/tree/spec-marshal.

sim642:
Closes #352.

### TODO
In analyses:
- [ ] Fix all `Hashtbl`s?
- [ ] Fix all `GoblintUtil.create_var`s?
- [ ] Fix all `RichVarinfo`s?

Since the rest of analyses (besides mallocWrapper) don't seem to have broken under marshaling due to a similar reason, I'm not sure if it's worth the time to go through all analyses to fix such `Hashtbl`s and `create_var`s. Some OSEK/ARINC stuff uses them extensively.

My proposal would be not to for the time being. If an issue appears, it can be fixed because the marshaling interface is now there. If we at some point remove unmaintained analyses, then it's not worth trying to fix and risk introducing new bugs, when they'll get removed anyway.


src/framework/control.ml
```diff
@@ -278,7 +278,11 @@ struct
 
     GU.global_initialization := true;
     GU.earlyglobs := get_bool "exp.earlyglobs";
-    Spec.init ();
+    if get_string "load_run" <> "" then (
+      Spec.init (Some (Serialize.unmarshal (Filename.concat (get_string "load_run") "spec_marshal")))
+    )
+    else
+      Spec.init None;
```
sim642, line 285: `Spec.init` is much earlier than the rest of unmarshaling, not sure if this matters.

src/framework/control.ml
```diff
@@ -577,7 +581,10 @@ struct
     if get_bool "exp.cfgdot" then
       CfgTools.dead_code_cfg file (module Cfg : CfgBidir) liveness;
 
-    Spec.finalize ();
+    let marshal = Spec.finalize () in
+    if get_string "save_run" <> "" then (
+      Serialize.marshal marshal (Filename.concat (get_string "save_run") "spec_marshal")
+    );
```
sim642, line 587: `Spec.finalize` is much later than the rest of marshaling, not sure if this matters.

src/framework/control.ml
michael-schwarz: In principle it should not, just be aware that if this marshaled data contains e.g. references they will not point to the same things after marshaling in as references e.g. in the marshaled in hash table, even if they did when it was marshaled out (which has also related to the issue with hashconsing), but hopefully no analysis has such things in its hash table.

src/framework/control.ml
michael-schwarz: When marshaling happens here is irrelevant.

src/framework/constraints.ml
```diff
@@ -21,6 +21,7 @@ struct
 
   let name () = S.name () ^" hashconsed"
 
+  type marshal = S.marshal (* TODO: should hashcons table be in here to avoid relift altogether? *)
```
michael-schwarz, line 24: Not sure if this would help, this table would be completely disconnected from the marshaled solution, right?

src/analyses/extract_osek.ml
```diff
@@ -321,7 +321,7 @@ struct
   let threadspawn ctx lval f args fctx = ctx.local
   let exitstate  v = D.bot ()
 
-  let init () = (* registers which functions to extract and writes out their definitions *)
+  let init marshal = (* registers which functions to extract and writes out their definitions *)
```
michael-schwarz, line 324: In principle, there is the same oddity as with `extract_arinc`: The `init ()` defined before is also not called here.

michael-schwarz:
> My proposal would be not to for the time being. If an issue appears, it can be fixed because the marshaling interface is now there. If we at some point remove unmaintained analyses, then it's not worth trying to fix and risk introducing new bugs, when they'll get removed anyway.

+1

michael-schwarz (approved):

src/framework/control.ml
sim642: Well, the marshaling for analysis-specific data is very much related to what's in the solution. So this is a very relevant point.

For mallocWrapper hashtable it doesn't matter that the physical objects are not the same, just that both unmarshals separately get objects which have matching identity (by `equal`/`hash`).

So following this logic, all marshaled data should be put in a single file using a wrapping record, such that their inter-references are preserved properly. And that no blocks need to be duplicated on disk or after unmarshaling.

src/framework/control.ml
sim642: If marshaling is to be done as a single object, like I'm now thinking above, this cannot be separate like this, but everything is unmarshaled/marshaled together. This just means all the solver data etc needs to be kept around for saving this late and all the solver etc data needs to be loaded earlier for `init`.

src/framework/constraints.ml
sim642: If spec marshaled data goes together with solution, then it's no longer disconnected. I didn't try this but just thought that maybe if all hashconsing tables are marshaled (not sure if they can though, they use weak data structures internally) and restored, all the solver data is already consistent with hashconsing and nothing needs to be relifted.

jerhard: >if all hashconsing tables are marshaled (not sure if they can though, they use weak data structures internally)

It does indeed not look like this is possible, as the implementation of `BatHashcons.Table` uses `Weak`: 
https://github.com/ocaml-batteries-team/batteries-included/blob/99dcf7d66a6476dcd04f800acfb43b3af09a1d0e/src/batHashcons.ml#L64-L68
and those cannot be marshaled (see here: https://ocaml.org/api/Weak.html, the documentation is referring to the `Marshal` module by the standard library, but you also get an error `Fatal error: exception Invalid_argument("output_value: abstract value (Abstract)")` if your try to marshal the hashcons-tables with [`Batteries.Marshal.output`](https://ocaml-batteries-team.github.io/batteries-included/hdoc2/Batteries.Marshal.html).

sim642: It should still be possible to temporary convert anything still present in the weak hashtable into a normal one for marshaling purposes. If everything gets added back in the correct order during unmarshaling, it should be equivalent.

That would still involve some extra work on the hashtable, but it might not require calling `relift` on everything in the solution and needing `relift` to recurse into inner values through multiple layers. And if it works, then it would also allow removing `relift` from the interface, which is good for cleanup.

Issue #351, opened 2021-09-20
Merge incremental analysis improvements


sim642:
This is a task for @stilscher, but she cannot be assigned since she's not member of the organization yet.

- [x] General improvements (#356)
- [x] Reluctant destabilization (#369)
- [ ] CFG comparison

PR #356, opened 2021-09-22
Incremental/compatibility-updates

stilscher:
This PR contains some general updates for the incremental analysis. Its part of #351, the two improvements build upon this updated version of the incremental analysis. This PR contains:

- small adaptations to the incremental config file using
- making the mallocWrapperAnalysis location independent for compatibility with the incremental analysis
- adapting the incremental script for more precise runtime measurement and adding a script for taking average measurements on multiple incremental runs


conf/incremental.json
```diff
@@ -21,7 +23,8 @@
     },
     "incremental": {
       "mode": "incremental"
-    }
+    },
+    "privatization": "none"
```
michael-schwarz, line 27: Is there a specific reason to turn off default privatization for incremental? It was on before.

src/framework/cfgTools.ml
```diff
@@ -246,7 +246,7 @@ let createCFG (file: file) =
           let newst = mkStmt (Return (None, fd_loc)) in
           let start_id = 10_000_000_000 in (* TODO get max_sid? *)
           let sid = Hashtbl.hash fd_loc in (* Need pure sid instead of Cil.new_sid for incremental, similar to vid in Goblintutil.create_var. We only add one return stmt per loop, so the location hash should be unique. *)
-          newst.sid <- if sid < start_id then sid + start_id else sid;
+          newst.sid <- sid + start_id;
```
michael-schwarz, line 249: Is this change required?

I'm not sure if it still guarantees that there are no `sid` smalller than `10_000_000_000` generated here? What if the hash already is large and then overflows when adding `10_000_000_000` and ends up in the territory where CILs normal ids are?

src/util/defaults.ml
```diff
@@ -207,7 +207,7 @@ let _ = ()
       ; reg Experimental "exp.incremental.mode"   "'off'" "Use incremental analysis in the TD3 solver. Values: off (default), incremental (analyze based on data from a previous commit or fresh if there is none), complete (discard loaded data and start fresh)."
       ; reg Experimental "exp.incremental.stable" "true"  "Reuse the stable set and selectively destabilize it."
       ; reg Experimental "exp.incremental.wpoint" "false" "Reuse the wpoint set."
-      ; reg Experimental "exp.gcc_path"           "'/usr/bin/gcc-6'" "Location of gcc-6. Used to combine source files with cilly."
+      ; reg Experimental "exp.gcc_path"           "'/usr/bin/gcc'" "Location of gcc. Used to combine source files with cilly."
```
michael-schwarz, line 210: Does CIL work correctly with the output of newer gcc versions?

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -54,40 +54,49 @@ struct
...
+    let incremental_mode = get_string "exp.incremental.mode" in
+    if incremental_mode <> "off" then (
+      match Serialize.load_heap_vars () with
+      | Some h -> heap_hash := h
+      | None -> ()
+    )
+
+  let finalize () =
+    let incremental_mode = get_string "exp.incremental.mode" in
+    if incremental_mode <> "off" then Serialize.save_heap_vars !heap_hash
```
sim642, line 99: PR #354 adds general support for this sort of thing (instead of spreading incremental logic all over the place) and also fixes the marshaling issue here, so that one should probably be merged first.

src/framework/myCFG.ml
```diff
@@ -52,3 +52,27 @@ let dummy_func = emptyFunction "__goblint_dummy_init" (* TODO get rid of this? *
...
+  (* Identity *)
+  let compare x y = Node.compare x y
+  let equal x y = Node.compare x y = 0
+  let hash x = Node.hash x
+
+  (* Output *)
+  let pretty () x = Node.pretty_trace () x
+  let show x = Pretty.sprint ~width:max_int (pretty () x)
+  let printXml f x = BatPrintf.fprintf f "<value>\n<data>\n%s\n</data>\n</value>\n" (XmlUtil.escape (show x))
+  let to_yojson x = `String (show x)
```
sim642, line 76: These functions should probably be added to the `Node` module directly to make it `Printable.S`.

conf/incremental.json
stilscher: The incremental mode has not been tested with the mutex analysis yet (also not before) which would be required for privatization. I think there might be a similar location dependence as in the mallocWrapperAnalysis which would need to be removed to make it incrementally runable. Testing the incremental mode with more analyses should definitely be done. I turned off the privatization for now, because it can not be turned on anymore without the mutex analysis activated.

src/solvers/td3.ml
```diff
@@ -287,7 +287,7 @@ module WP =
           let add_stmts (f: fundec) =
             List.iter (fun s -> Hashtbl.replace nodes (Node.show_id (Statement s)) ()) (f.sallstmts)
           in
-          List.iter (fun f -> Hashtbl.replace nodes (Node.show_id (FunctionEntry f)) (); Hashtbl.replace nodes (Node.show_id (Function f)) (); add_stmts f) functions;
+          List.iter (fun f -> Hashtbl.replace nodes (Node.show_id (FunctionEntry f)) (); Hashtbl.replace nodes (Node.show_id (Function f)) (); add_stmts f; Hashtbl.replace nodes (string_of_int (f.svar.vid + 10_000_000_000)) ()) functions;
```
sim642, line 290: What is this magic number? If it's supposed to match some other places, it should probably be a constant variable, so it's not possible to accidentally change it in one place but forget about the others.

sim642 (changes requested):
If malloc `varinfo` names need to be changed from locations to IDs, then shouldn't the same be done to thread ID `varinfo` names? And potentially some other uses as well?

I'm wondering though, how we could make it as user-friendly as before, because now it's very difficult to inspect the analysis results and know where a particular allocation is from. You'd have to click through all function CFGs to find the one which contains the ID in the `varinfo`.

src/framework/cfgTools.ml
stilscher: Good point, I changed it back with efbe9e5ef0acc359cc5a03c9e5a0cb615970007a.

src/solvers/td3.ml
stilscher: This change is for also removing the outdated unknowns of a pseudo return because a pseudo return node from a function has the same id across different program versions. I tried to improve it by extracting the calculation of the pseudo return ids in efbe9e5ef0acc359cc5a03c9e5a0cb615970007a.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -8,7 +8,7 @@ module Spec : Analyses.MCPSpec =
 struct
   include Analyses.DefaultSpec
 
-  module PL = Lattice.Flat (MyCFG.PrintableNode) (struct
+  module PL = Lattice.Flat (Node) (struct
     let top_name = "Unknown line"
     let bot_name = "Unreachable line"
```
sim642, line 13: These names are now also wrong.

src/framework/node.ml
```diff
@@ -1,6 +1,8 @@
 open Cil
 open Pretty
 
+include Printable.Std
```
sim642, line 4: This brings in `let name () = "std"`, which should be overridden here to avoid confusion when using this module inside something else. The previous `PrintableNode` did this correctly.

Other things in `Printable.Std` should be fine by default.

src/framework/node.ml
```diff
@@ -35,6 +37,12 @@ let pretty_trace () = function
   | Function      fd -> dprintf "call of %s" fd.svar.vname
   | FunctionEntry fd -> dprintf "entry state of %s" fd.svar.vname
 
+(** Output functions for Printable interface *)
+let pretty () x = pretty_trace () x
```
sim642, line 41: It's a bit debatable here, which of the existing `pretty_*` variants should be the default. But since we are using a mix of them all over the place for different information and there's no true preferred way, then I guess this is good enough for now. Since there was no `pretty` here, nothing else depended or will get changed by this new default.

sim642 (changes requested):

src/util/defaults.ml
stilscher: For the incremental test runs on the figlet repository it worked with gcc 8, 9 and 10. I don't think there are currently other use cases for combining the files with cilly other than the incremental mode. I found that getting gcc 6 to run on a newer Ubuntu system is rather cumbersome.

michael-schwarz: Alright, let's use `gcc` then :+1: Maybe add a comment that this needs to be changed to `gcc-9` (or some other version) on OS X, where `gcc` is `clang` by default which will lead to `cilly` failing.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -55,34 +55,39 @@ struct
...
   let exitstate  v = D.top ()
 
+
   (* refs to reassign unmarshaled in init *)
   let heap_hash = ref (Hashtbl.create 113)
   let heap_vars = ref (Hashtbl.create 113)
 
   type marshal = {
-    heap_hash: (location, varinfo) Hashtbl.t;
+    heap_hash: (string, varinfo) Hashtbl.t;
```
sim642, line 64: A string-based hashtable is not the best. It'd be better if the keys were `Node.t` directly and no `nodeId` string conversion would be necessary.

To make that work properly, we shouldn't use the polymorphic `Hashtbl` (which would use polymorphic comparisons on nodes), but rather something like
```ocaml
module NH = Hashtbl.Make (Node)
(* ... *)
heap_hash: varinfo NH.t
```

I've thought about refactoring it here before, but there wasn't a real need before, although now would be a good time.
Then the other hashtable could also be properly made to use `CilType.Varinfo` keys directly instead of `int`.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -57,28 +57,29 @@ struct
...
-      | Function f -> "vid:" ^ (string_of_int f.svar.vid)
-      | _ -> raise (Failure "A function entry node can never be the node after a malloc") in
-    try Hashtbl.find !heap_hash nodeId
+    try NH.find !heap_hash node
     with Not_found ->
-      let name = "(alloc@" ^ nodeId ^ ")" in
+      let name = match node with
+        | Node.Statement s -> "(alloc@sid:" ^ (string_of_int s.sid) ^ ")"
+        | Function f -> "(alloc@vid:" ^ (string_of_int f.svar.vid) ^ ")"
+        | _ -> raise (Failure "A function entry node can never be the node after a malloc") in
```
sim642, line 79: Can a function return node actually `malloc` something?

stilscher: The node that is used when creating a heap variable, is the one following the call to `malloc`. The return edges will probably never contain a `malloc`(-Wrapper) call since CIL seems to introduce a temporary variable for function calls in the return statement. In that case, one would actually not need to consider return nodes when creating heap variables.

sim642: Unless we know of an example where the `Function` case is necessary here, I think we should remove it and make it also raise an exception (you can use `failwith` directly). That would simplify the names a bit to always be statement IDs.

src/incremental/serialize.ml
```diff
@@ -7,7 +7,7 @@ let cil_file_name = "ast.data"
 let solver_data_file_name = "solver.data"
 let results_dir = "results"
 let results_tmp_dir = "results_tmp"
-
+let heap_file_name = "heap.data"
```
sim642, line 10: Now that mallocWrapper hashtables are marshaled via `init`/`finalize`, I guess this isn't used anywhere anymore?

sim642 (approved):

src/incremental/serialize.ml
stilscher: Yes, thanks for the hint. Must have missed this when merging.

PR #360, opened 2021-09-28
Optimize recursive struct type comparison in `CompareAST`

### LINKED ISSUE #359: Incremental CIL file comparison is extremely slow with kernel
Even when incrementally analyzing a tiny kernel benchmark from our regression suite without any changes, the incremental run is hundreds of times slower than from-scratch analysis. For example:
```console
./regtest.sh 04 33 --set result none -v --enable incremental.save
./regtest.sh 04 33 --set result none -v --enable incremental.load
```

By using `Stats.time` it appears that a lot of time is spent in `compareCilFiles`:
```
Timings:
TOTAL                          104.756 s
  parse                           0.079 s
  convert to CIL                  0.093 s
  compareCilFiles                104.194 s
  analysis                        0.390 s
    global_inits                    0.078 s
    solving                         0.044 s
    verify                          0.021 s
```

sim642:
To be merged after #358.

Closes #359.

### Fixes
First, there was a catch-all exception handler which was also catching `Stack_overflow` exceptions coming from recursive type comparison and hiding the problem. Those stack overflows just took time to happen.

Second, function argument type comparison was ignoring the `acc` structure that's supposed to avoid infinite recursion. A mix of recursive struct types with function pointer types to those structs caused the infinite recursion.

Thirdly, a global `acc` is added to reuse the recursive type equivalences between multiple globals, instead of having to go through all the trouble again.

Fourthly, the `acc` structure is only used for structs now, because that's what causes recursive types. By doing this the `acc` list is made significantly smaller, so checking it with `List.exists` is also much faster. There's no point in putting trivial equivalences like `(int, int)` into the structure because they're trivially computed from scratch.

### Performance improvement
Before (in #359):
```
Timings:
TOTAL                          104.756 s
  parse                           0.079 s
  convert to CIL                  0.093 s
  compareCilFiles                104.194 s
  analysis                        0.390 s
    global_inits                    0.078 s
    solving                         0.044 s
    verify                          0.021 s
```

After:
```
Timings:
TOTAL                           0.691 s
  parse                           0.074 s
  convert to CIL                  0.086 s
  compareCilFiles                 0.073 s
  analysis                        0.458 s
    global_inits                    0.074 s
    solving                         0.066 s
      S.Dom.equal                     0.001 s
    verify                          0.021 s
```

This makes `compareCilFiles` on the regression test **1400 times faster**!

And as you see, all these fixes were necessary to bring the comparison time of two unchanged ASTs down to be more like its parsing and conversion time, which is how it should be. Previously comparison of ASTs could be much bigger bottleneck than parsing itself!

### DIFF
--- a/src/incremental/compareAST.ml
+++ b/src/incremental/compareAST.ml
@@ -18,12 +18,14 @@ type global_type = Fun | Decl | Var | Other
 
 and global_identifier = {name: string ; global_t: global_type} [@@deriving ord]
 
+exception NoGlobalIdentifier of global
+
 let identifier_of_global glob =
   match glob with
   | GFun (fundec, l) -> {name = fundec.svar.vname; global_t = Fun}
   | GVar (var, init, l) -> {name = var.vname; global_t = Var}
   | GVarDecl (var, l) -> {name = var.vname; global_t = Decl}
-  | _ -> raise (Failure "No variable or function")
+  | _ -> raise (NoGlobalIdentifier glob)
 
 module GlobalMap = Map.Make(struct
     type t = global_identifier [@@deriving ord]
@@ -257,7 +259,7 @@ let compareCilFiles (oldAST: Cil.file) (newAST: Cil.file) =
          then changes.unchanged <- global :: changes.unchanged
          else changes.changed <- {current = global; old = old_global} :: changes.changed
        with Not_found -> ())
-    with e -> () (* Global was no variable or function, it does not belong into the map *)
+    with NoGlobalIdentifier _ -> () (* Global was no variable or function, it does not belong into the map *)
   in
   let checkExists map global =
     let name = identifier_of_global global in
--- a/src/incremental/compareAST.ml
+++ b/src/incremental/compareAST.ml
@@ -78,8 +78,11 @@ and eq_lhost (a: lhost) (b: lhost) = match a, b with
   | _, _ -> false
 
 and eq_typ_acc (a: typ) (b: typ) (acc: (typ * typ) list) =
+  (* if Messages.tracing then Messages.tracei "compare" "eq_typ_acc %a vs %a\n" d_type a d_type b; *)
+  if Messages.tracing then Messages.tracei "compare" "eq_typ_acc %a vs %a (%d)\n" d_type a d_type b (List.length acc); (* TODO: remove because always calls List.length *)
+  let r =
   if( List.exists (fun x-> match x with (x,y)-> a==x && b == y) acc)
-  then true
+  then (if Messages.tracing then Messages.trace "compare" "in acc\n"; true)
   else (let acc = List.cons (a,b) acc in
         match a, b with
         | TPtr (typ1, attr1), TPtr (typ2, attr2) -> eq_typ_acc typ1 typ2 acc && eq_list eq_attribute attr1 attr2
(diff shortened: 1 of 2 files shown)


sim642:
If possible, the existing incremental benchmarks might be worth rerunning with this improvement. Both to see if this bottleneck could also appear there before or to notice possible issues introduced by this supposed fix.

Although it might be a bit difficult to redo that benchmark since this PR is on top of #358, which removes the Git-based incremental usage for the figlet benchmark.

src/incremental/compareAST.ml
```diff
@@ -157,7 +182,10 @@ and eq_compinfo (a: compinfo) (b: compinfo) (acc: (typ * typ) list) =
   a.cdefined = b.cdefined (* Ignore ckey, and ignore creferenced *)
 
 and eq_fieldinfo (a: fieldinfo) (b: fieldinfo) (acc: (typ * typ) list)=
-  a.fname = b.fname && eq_typ_acc a.ftype b.ftype acc && a.fbitfield = b.fbitfield &&  eq_list eq_attribute a.fattr b.fattr
+  if Messages.tracing then Messages.tracei "compare" "fieldinfo %s vs %s\n" a.fname b.fname;
+  let r = a.fname = b.fname && eq_typ_acc a.ftype b.ftype acc && a.fbitfield = b.fbitfield &&  eq_list eq_attribute a.fattr b.fattr in
+  if Messages.tracing then Messages.traceu "compare" "fieldinfo %s vs %s\n" a.fname b.fname;
+  r
```
jerhard, line 188: Should we keep this debugging output?

src/incremental/compareAST.ml
```diff
@@ -110,9 +134,10 @@ and eq_enuminfo (a: enuminfo) (b: enuminfo) =
   eq_list eq_eitems a.eitems b.eitems
 (* Ignore ereferenced *)
 
-and eq_args (a: string * typ * attributes) (b: string * typ * attributes) = match a, b with
-    (name1, typ1, attr1), (name2, typ2, attr2) -> name1 = name2 && eq_typ typ1 typ2 && eq_list eq_attribute attr1 attr2
+and eq_args (acc: (typ * typ) list) (a: string * typ * attributes) (b: string * typ * attributes) = match a, b with
+    (name1, typ1, attr1), (name2, typ2, attr2) -> name1 = name2 && eq_typ_acc typ1 typ2 acc && eq_list eq_attribute attr1 attr2
 
+(* TODO: why unused? *)
 and eq_typsig (a: typsig) (b: typsig) =
```
jerhard, line 141: `eq_typsig` can be removed. It does not suffice to compare `typsig`s (instead of types) in incremental runs, as the `typsig`s e.g. do not contain the fields of unions and structs.

src/incremental/compareAST.ml
```diff
@@ -75,16 +77,19 @@ and eq_lhost (a: lhost) (b: lhost) = match a, b with
...
+
+and mem_typ_acc (a: typ) (b: typ) acc = List.exists (fun p -> match p with (x, y) -> a == x && b == y) acc (* TODO: seems slightly more efficient to not use "fun (x, y) ->" directly to avoid caml_tuplify2 *)
+
 and eq_typ_acc (a: typ) (b: typ) (acc: (typ * typ) list) =
-  if( List.exists (fun x-> match x with (x,y)-> a==x && b == y) acc)
-  then true
-  else (let acc = List.cons (a,b) acc in
-        match a, b with
+  if Messages.tracing then Messages.tracei "compare" "eq_typ_acc %a vs %a\n" d_type a d_type b;
+  (* if Messages.tracing then Messages.tracei "compare" "eq_typ_acc %a vs %a (%d)\n" d_type a d_type b (List.length acc); (* TODO: remove because always calls List.length *) *)
```
jerhard, line 86: Should this commented-out line be removed?

src/incremental/compareAST.ml
sim642: I think it's worth keeping because `eq_type_acc` and `eq_fieldinfo` are essentially the two things which can be used to get an understanding of how this works and what it skips, in case something about it ever needs to be fixed again.

I'll just rename the tracing key from `compare` to `compareast` to be a bit more specific.

jerhard:
I let this run on the figlet repo and compared it to `master`. The run time of Goblint for the commits where things changed is pretty much the same (`master`: 118 seconds, this branch: `115` seconds.).

jerhard (approved):

sim642:
> I let this run on the figlet repo and compared it to `master`. The run time of Goblint for the commits where things changed is pretty much the same (`master`: 118 seconds (in total), this branch: 115 seconds; the time resolution was only on the levels of secons, however), so for the `figlet` repo it does not make so much of a difference.

Thanks for benchmarking it a bit! I think this is expected since figlet probably doesn't have these crazy recursive struct and function pointer types, which caused the biggest issue on Linux headers.

PR #362, opened 2021-09-29
Change "make debug" command so that it uses dune.

jerhard:
The previous debug build with ocamlbuild did not work anymore for me, at least without having Z3 installed.

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -65,7 +65,7 @@ rule() {
     ;; byte)
       eval $(opam config env)
       dune build goblint.byte &&
-      cp _build/default/$TARGET.bc goblint.byte
+      cp _build/default/goblint.byte goblint.byte
     ;; profile)
       # gprof (run only generates gmon.out). use: gprof goblint
       ocb -tag profile $TARGET.p.native &&
--- a/make.sh
+++ b/make.sh
@@ -63,8 +63,9 @@ rule() {
       ocb -no-plugin $TARGET.native &&
       cp _build/$TARGET.native goblint
     ;; debug)
-      ocb -tag debug $TARGET.d.byte &&
-      cp _build/$TARGET.d.byte goblint.byte
+      eval $(opam config env)
+      dune build $TARGET.bc &&
+      cp _build/default/$TARGET.bc goblint.byte
     ;; profile)
       # gprof (run only generates gmon.out). use: gprof goblint
       ocb -tag profile $TARGET.p.native &&
--- a/make.sh
+++ b/make.sh
@@ -62,9 +62,9 @@ rule() {
     ;; ocbnat*)
       ocb -no-plugin $TARGET.native &&
       cp _build/$TARGET.native goblint
-    ;; debug)
+    ;; byte)
       eval $(opam config env)
-      dune build $TARGET.bc &&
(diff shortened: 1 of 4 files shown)


make.sh
```diff
@@ -63,8 +63,9 @@ rule() {
       ocb -no-plugin $TARGET.native &&
       cp _build/$TARGET.native goblint
     ;; debug)
-      ocb -tag debug $TARGET.d.byte &&
-      cp _build/$TARGET.d.byte goblint.byte
+      eval $(opam config env)
+      dune build $TARGET.bc &&
+      cp _build/default/$TARGET.bc goblint.byte
```
sim642, line 68: Couldn't the dune target be named `goblint.byte` as well to be clear that it's the same thing?

make.sh
```diff
@@ -63,8 +63,9 @@ rule() {
       ocb -no-plugin $TARGET.native &&
       cp _build/$TARGET.native goblint
     ;; debug)
```
sim642, line 65: In what sense is this a `debug` build? It's just the bytecode build. I guess that's necessary for earlybird?

Might make sense to call the make target `byte` as well, just like how `native` is consistently called that.

sim642 (commented):
Step towards replacing old `ocamlbuild` stuff with equivalent `dune` is good. At one point we might be able to drop the `ocamlbuild` dependency for good.

make.sh
jerhard: `make debug` might be a bit more intuitive for the most common usage of the byte-code file. Anyways I changed the name to `byte` and the documentation accordingly.

make.sh
jerhard: The target in `./src/dune` is `goblint.bc` because `.bc` is just the extension dune expects for byte-code files. Instead of the target in `./src/dune` it now uses the one in `./dune`.

sim642: Ahh, ok, I wasn't aware that dune required the `.bc` extension itself. Then it will have to stay.

Although I think something is still off with the copying:
1. `dune` copies `_build/default/src/goblint.bc` → `_build/default/goblint.byte`
2. `make` now copies `_build/default/goblint.bc` → `goblint.byte`.

The intermediate doesn't match up it seems.

jerhard: >`make` now copies `_build/default/goblint.bc` → `goblint.byte`

Actually, it copied `_build/default/src/goblint.bc` -> `goblint.byte`, as `$TARGET` expands to `src/goblint`. 

Should I just get rid of the intermediate step in `./dune` and call the target `$TARGET$.bc` in `make.sh` (as it was at first)?

sim642: It's a bit redundant then, but we also do it for normal compilation in that case. I guess what it ensures is that in `_build/default/` you also have a copy as if it's the top level of the repository/build directory.

It might even be necessary to keep it if one wants to `opam install` or `dune @install` since those would only have the build directory somewhere.

jerhard: Now it does the following:

1. `dune` copies `_build/default/src/goblint.bc` → `_build/default/goblint.byte`
2. `make` now copies `_build/default/goblint.byte` → `goblint.byte`.

sim642 (approved):

PR #363, opened 2021-09-29
Incremental TD3: restart destabilized side-effected variables

sim642:
This is an attempt to restart destabilized side-effected variables (at least globals, but could also be function entry nodes in presence of partial contexts) to bottom when loading previous results. This allows globals (and especially their protecting locksets) to improve after fixing the program.

At loading time, obsolete nodes are destabilized with a special `destabilize_with_side` function instead. This requires knowing `infl` for side effects (in `side_infl`) but also dependencies for side effects (in `side_dep`) to replay the unchanged ones. Both additional mappings are constructed during `reachability` at saving time.

### TODO
- [ ] Decide whether to just restart globals or also function entry nodes. The former requires ugly `FromSpec`-specific hack inside TD3.
- [ ] Benchmark on larger programs.


vesalvojdani:
I think unmarshalling version.data is stuck... can you try doing this in the bench repo:
```
../analyzer/goblint --enable kernel linux/drivers/char/applicom.c --enable incremental.save --enable printstats -v
../analyzer/goblint --enable kernel linux/drivers/char/applicom.c --enable incremental.load --enable printstats -v
```
See what happens?

sim642:
We decided to start merging some of these together into the branch `interactive` for better experimentation.

PR #366, opened 2021-09-30
(Incremental) TD3: restart detected wpoints

sim642:
This PR is currently built on top of #363 although technically it's independent of the incremental restarting logic from there.
I'm not sure if this would do anything useful for non-incremental solving though.

This is an attempt to further improve the incremental case, where a global gets restarted, but some local variables still contain the old value (due to global reads). It particularly concerns wpoints, where we don't overwrite and might join/widen the old value with the new one, thus republishing the old value again and having gained nothing from restarting the global.

This is achieved by doing the restart when `eval` detects a `wpoint` to ensure that the loop body (which may cause side effects) isn't ever re-evaluated with the old value.


sim642:
We decided to start merging some of these together into the branch `interactive` for better experimentation.

PR #369, opened 2021-09-30
Incremental/reluctant-destabilization

stilscher:
This pull request contains an improvement for the incremental analysis in the td3 solver (#351). With this, the changed functions are not destabilized right away anymore before solving once completely. Instead it is split as follows:

- The changed functions are solved separately for the contexts from the previous run. For each, the found abstract value for the return node is compared to the one from the previous run. The functions return node needs to be destabilized only further if the found abstract value is not `leq` then the old one.
- After every changed function has been solved separately `solve` is called on the queried unknown. This ensures, that possibly new side-effects affecting the queried unknown are taken into account.

Note that for this to work, it is necessary to adopt the previous ids for formal parameters of changed functions when possible. This way, the contexts of different versions for a function entry node of a changed function can coincide. Then the result from the separate solving phase for changed functions can actually be used and is not unnecessarily re-analyzed in the final `solve`.

A config parameter has been introduced to deactivate the reluctant destabilization when needed.

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -269,12 +269,11 @@ module WP =
         let obsolete_funs = filter_map (fun c -> match c.old with GFun (f,l) -> Some f | _ -> None) S.increment.changes.changed in
         let removed_funs = filter_map (fun g -> match g with GFun (f,l) -> Some f | _ -> None) S.increment.changes.removed in
         (* TODO: don't use string-based nodes, make obsolete of type Node.t BatSet.t *)
-        let obsolete = Set.union (Set.of_list (List.map (fun a -> Node.show_id (Function a))  obsolete_funs))
-                                 (Set.of_list (List.map (fun a -> Node.show_id (FunctionEntry a))  obsolete_funs)) in
+        let obsolete = Set.of_list (List.map (fun a -> Node.show_id (Function a))  obsolete_funs) in
 
         List.iter (fun a -> print_endline ("Obsolete function: " ^ a.svar.vname)) obsolete_funs;
 
-        (* Actually destabilize all nodes contained in changed functions. TODO only destabilize fun_... nodes *)
+        (* Actually destabilize all nodes contained in changed functions. *)
         HM.iter (fun k v -> if Set.mem (S.Var.var_id k) obsolete then destabilize k) stable; (* TODO: don't use string-based nodes *)
 
         (* We remove all unknowns for program points in changed or removed functions from rho, stable, infl and wpoint *)
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -284,8 +284,6 @@ module WP =
             let old_rho = HM.find rho k in
             let old_infl = HM.find_default infl k VS.empty in
             Hashtbl.replace old_ret k (old_rho, old_infl))) rho;
-          (* Do not destabilize changed functions immediately, instead remove ret nodes from stable and wait for result of function-only solve *)
-          Hashtbl.iter (fun k v -> HM.remove stable k) old_ret
         ) else (
           HM.iter (fun k _ -> if Set.mem (S.Var.var_id k) obsolete_entry then destabilize k) stable
         );
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -328,7 +328,8 @@ module WP =
               destabilize x; HM.replace stable x ())
           ) old_ret;
 
-          if !numDest = 0 then print_endline "no actual destabilization needed!"
+          if !numDest = 0 then print_endline "no actual destabilization needed!";
+          print_endline "final solve"
           (* ignore (Pretty.printf "vars = %d    evals = %d  \n" !Goblintutil.vars !Goblintutil.evals) *)
         )
       );
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -269,25 +269,35 @@ module WP =
         let obsolete_funs = filter_map (fun c -> match c.old with GFun (f,l) -> Some f | _ -> None) S.increment.changes.changed in
         let removed_funs = filter_map (fun g -> match g with GFun (f,l) -> Some f | _ -> None) S.increment.changes.removed in
         (* TODO: don't use string-based nodes, make obsolete of type Node.t BatSet.t *)
-        let obsolete = Set.of_list (List.map (fun a -> Node.show_id (Function a))  obsolete_funs) in
+        let obsolete_ret = Set.of_list (List.map (fun a -> Node.show_id (Function a))  obsolete_funs) in
 
         List.iter (fun a -> print_endline ("Obsolete function: " ^ a.svar.vname)) obsolete_funs;
 
-        (* Actually destabilize all nodes contained in changed functions. *)
-        HM.iter (fun k v -> if Set.mem (S.Var.var_id k) obsolete then destabilize k) stable; (* TODO: don't use string-based nodes *)
+
+        (* save entries of changed functions in rho for the comparison whether the result has changed after a function specific solve *)
+        let old_ret = Hashtbl.create 103 in
+        HM.iter (fun k v -> if Set.mem (S.Var.var_id k) obsolete_ret then ( (* TODO: don't use string-based nodes *)
+          print_endline (S.Var.var_id k);
+          ignore @@ Pretty.printf "%a\n" S.Var.pretty_trace k;
+          let old_rho = HM.find rho k in
+          let old_infl = HM.find_default infl k VS.empty in
+          Hashtbl.replace old_ret k (old_rho, old_infl))) rho;
+
+        (* Do not destabilize changed functions immediately, instead remove ret nodes from stable and wait for result of function-only solve *)
+        Hashtbl.iter (fun k v -> HM.remove stable k) old_ret;
 
         (* We remove all unknowns for program points in changed or removed functions from rho, stable, infl and wpoint *)
         (* TODO: don't use string-based nodes, make marked_for_deletion of type unit (Hashtbl.Make (Node)).t *)
-        let add_nodes_of_fun (functions: fundec list) (nodes)=
+        let add_nodes_of_fun (functions: fundec list) (nodes) withEntry =
           let add_stmts (f: fundec) =
             List.iter (fun s -> Hashtbl.replace nodes (Node.show_id (Statement s)) ()) (f.sallstmts)
           in
-          List.iter (fun f -> Hashtbl.replace nodes (Node.show_id (FunctionEntry f)) (); Hashtbl.replace nodes (Node.show_id (Function f)) (); add_stmts f; Hashtbl.replace nodes (string_of_int (CfgTools.get_pseudo_return_id f)) ()) functions;
+          List.iter (fun f -> if withEntry then Hashtbl.replace nodes (Node.show_id (FunctionEntry f)) (); Hashtbl.replace nodes (Node.show_id (Function f)) (); add_stmts f; Hashtbl.replace nodes (string_of_int (CfgTools.get_pseudo_return_id f)) ()) functions;
         in
 
         let marked_for_deletion = Hashtbl.create 103 in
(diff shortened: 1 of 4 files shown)


src/solvers/td3.ml
```diff
@@ -246,6 +246,8 @@ module WP =
          *   if exp.earlyglobs: the contexts will be the same since they don't contain the global, but the start state will be different!
          *)
         print_endline "Destabilizing start functions if their start state changed...";
+        (* record whether any function's start state changed. In that case do not use reluctant destabilization *)
+        let any_changed_start_state = ref false in
```
sim642, line 250: Why not?

src/solvers/td3.ml
```diff
@@ -297,10 +308,26 @@ module WP =
...
         delete_marked stable;
 
-        print_data data "Data after clean-up"
+
+        print_data data "Data after clean-up";
+
+        List.iter set_start st;
+
+        if not !any_changed_start_state && GobConfig.get_bool "incremental.reluctant" then (
+          (* solve on the return node of changed functions. Only destabilize influenced nodes outside the function if the analysis result changed *)
```
sim642, line 317: This is not necessarily true though, the following `solve` might call `destabilize` something outside this function itself already and still unnecessarily destabilize large parts of the program. In order for this to actually be true, the `solve` from here should (in its all recursive usages) use a different `destabilize`, which comes back and does it reluctantly again or somehow prevent it from going wildly beyond this function.

As it turned out in #365, it's not clear how to even do that properly.

src/solvers/td3.ml
```diff
@@ -297,10 +308,26 @@ module WP =
...
+          print_endline "Separately solving changed functions...";
+          Hashtbl.iter (fun x (old_rho, old_infl) -> ignore @@ Pretty.printf "test for %a\n" Node.pretty_trace (S.Var.node x);
+            solve x Widen;
+            if not (S.Dom.leq (HM.find rho x) old_rho) then (
```
sim642, line 321: The downside here is also that if the change makes the state more precise, the new precision gets discarded in favor of the old less precise solution. This is a problem if we want to allow the user to fix a problem and have the related warnings disappear without having to reanalyze from scratch.

src/solvers/td3.ml
```diff
@@ -297,10 +308,26 @@ module WP =
...
+              HM.replace infl x old_infl;
+              destabilize x; HM.replace stable x ())
+          ) old_ret;
+
+          print_endline "Final solve..."
+        )
       );
 
-      List.iter set_start st;
+      if not (GobConfig.get_bool "incremental.load") then List.iter set_start st;
```
sim642, line 330: Maybe this could be in the `else` branch of the massive `if` before. It's a bit confusing to read it like this because it makes it look like `set_start` is somehow unnecessary for incremental, while it's actually still done in the incremental case, but simply earlier in the `then` branch.

src/util/defaults.ml
```diff
@@ -152,6 +152,7 @@ let _ = ()
       ; reg Incremental "incremental.save"   "false" "Store incremental analysis results."
       ; reg Incremental "incremental.stable" "true"  "Reuse the stable set and selectively destabilize it (recommended)."
       ; reg Incremental "incremental.wpoint" "false" "Reuse the wpoint set (not recommended). Reusing the wpoint will combine existing results at previous widening points."
+      ; reg Incremental "incremental.reluctant" "true" "Destabilize nodes in changed functions reluctantly"
```
sim642, line 155: I'm wondering if this should be enabled by default, given how its behavior can be quite unpredictable as is: it depends on the order of the functions destabilized reluctantly and may bypass reluctant destabilization as mentioned above.

src/solvers/td3.ml
stilscher: Maybe I should create a configuration option to specify whether to use `leq` or `eq` at this point? I think both can have their benefits. `eq` is useful if one does want more precision, whereas `leq` might be the better choice when one focuses on reducing the number of evaluations and possibly the runtime of the analysis.

src/solvers/td3.ml
stilscher: You are right, this incremental reluctant destabilization can lead to destabilizing nodes outside of a changed function during the phase of solving it separately. This should be correct even if it does not reduce the number of evaluations as effectively as one could hope. It also ensures that new side-effects are evaluated. I changed the comment because it was not formulated precisely. I think one should look further into how to make the destabilization reluctantly also inside the `solve`. Maybe we should address this in another issue?

vesalvojdani: Yes, this is definitely a separate issue, and at some point we should try to understand what exactly goes wrong when trying to interleave the solve and destabilization. @sim642 When your regressions failed: was it only lack of precision or was it even unsound? This isn't urgent, but would be good to understand if want some intuition about whether it is at all possible to limit destabilization with side-effects.

src/util/defaults.ml
stilscher: I turned it on as default because for my test runs on the _figlet_ repository the activation of reluctant destabilization did reduce the number of evaluations for quite a few commits but was never disadvantageous in this respect. But of course testing it on other repositories as well would be required for a better justified statement.

src/solvers/td3.ml
sim642: Just imprecise: https://github.com/goblint/analyzer/pull/365/checks?check_run_id=3744468024#step:6:674.

src/solvers/td3.ml
stilscher: If `earlyglobs` is turned on and a global initializer changes, the start state of a start function changes but its context stays the same. To still trigger the reanalysis for the start function, a check is implemented at the beginning of the incremental preprocessing in the td3 solver to detect this and explicitly destabilize the start function. If this happens and the start function is already fully destabilized there is no need to apply reluctant destabilization on the changed functions afterwards.

sim642: > If `earlyglobs` is turned on and a global initializer changes, the start state of a start function changes but its context stays the same.

If `earlyglobs` is on, the globals are never in the local states even, so changing a global initializer doesn't change the start state of the start function (nor its context). The change to the global initializer should just be destabilizing the global and thus everywhere its read.
Or do you mean when `earlyglobs` is _off_, because then globals are initially in the start state?

> If this happens and the start function is already fully destabilized there is no need to apply reluctant destabilization on the changed functions afterwards.

The start state of a function changing is a function-specific property though. If one function's start state changes, but another's doesn't, then reluctant destabilization could still be applied to the latter. Changing one shouldn't prevent reluctant destabilization of others, but just the changed one.

Also, destabilizing a function entry that was already destabilized shouldn't really be doing anything: the first `destabilize` clears its `infl`, so the second one doesn't do anything anyway.

src/solvers/td3.ml
sim642: Having an option might be good indeed.

vesalvojdani:
Some very preliminary benchmarking on this is available here: 
https://kodu.ut.ee/~vesal/results/211004-reluctant/

It seems that this reluctant stuff should be on (although we should compare with how the cut-off stuff plays out and if these things play well together), but there is also some strange behavior on this one patch where I move the locking to cover also some initialization code. This removes a false alarm. The incremental is then is even with reluctance much slower than the from scratch computation.

The brilliant script does not display the patches, so here's the problematic change:
https://github.com/goblint/bench/blob/incremental-bench/linux/drivers/char/apm-emulation01.patch

src/solvers/td3.ml
stilscher: > If earlyglobs is on, the globals are never in the local states even, so changing a global initializer doesn't change the start state of the start function (nor its context).

There seems to be some inconsistent behavior now. If you run the following minimalistic example with the incremental config (i.e. privatization off) the function `main` has a changed start state (because of the changed global initializer) and is therefor destabilized during the preparation for the incremental run. The global `g` is not a member of the list `st`.
```
int g = 0; // change to g = 2;

int main() {
  return 0;
}
```
This is the behavior of the incremental analysis as I know it and as it is described in the comments of the changed start state handling. I managed however to reconstruct what you described by turning privatization and the (probably not incremental compatible) mutex analysis on. Do you know if there was any recent change to the privatization that could affect this? Why are globals part of `st` with privatization turned on?

> The start state of a function changing is a function-specific property though. If one function's start state changes, but another's doesn't, then reluctant destabilization could still be applied to the latter.

With the behavior I described above, a changed global initializer affects the start state of every start function so all of them are destabilized anyway. If it weren't like that, it would be nicer to make a distinction between which function's start state changed. The problem I see is, that one can't relate an arbitrary changed function to the start function it is (indirectly) being called from. So it would be difficult (at least efficiently) to tell whether the changed function needs to be solved separately as part of reluctant destabilization.

> Also, destabilizing a function entry that was already destabilized shouldn't really be doing anything: the first `destabilize` clears its `infl`, so the second one doesn't do anything anyway.

The point is not to avoid the destabilization, but to avoid the function specific solve on a possibly outdated context. If the start function has been destabilized completely anyway, solving it with the top-down solver directly avoids unnecessary evaluations.

sim642: > If you run the following minimalistic example with the incremental config (i.e. privatization off) the function `main` has a changed start state (because of the changed global initializer) and is therefor destabilized during the preparation for the incremental run. The global `g` is not a member of the list `st`.

I've never been a fan of `earlyglobs`, so I'm not 100% sure what its expected behavior should be. But my understanding has been that it should from the beginning (global initialization) make global program variables into global constraint system variables, so `g` should be in `st`. Not being there and instead remaining in the entry state of the function sounds like an `earlyglobs` bug to me.

>  the (probably not incremental compatible) mutex analysis

I'm not sure why the mutex analysis should be incompatible with incremental because incremental analysis on the level of constraint systems shouldn't care whatsoever about the content of the constraint system.
The only problem currently (before #363) is that protecting locksets in global constraint variables cannot improve, but that should be all.

> Do you know if there was any recent change to the privatization that could affect this?

No, but it could be a bug remaining from the time all the traces article privatization implementations were added. `earlyglobs` caused a bunch of annoying issues there, but those got fixed to the extent that the regression tests pass. Although there might still be some bug that wasn't caught by the tests then.

> Why are globals part of `st` with privatization turned on?

Without `earlyglobs`, globals are kept in the local state just like all other variables. This allows them to be analyzed flow-sensitively. As long as the program is in single-threaded mode, they will be there.
What privatization does is that when multi-threaded mode is entered, they are removed from the local state and instead tracked flow-insensitively in the global invariant.

sim642: > With the behavior I described above, a changed global initializer affects the start state of every start function so all of them are destabilized anyway.

With `earlyglobs` off, the values for globals should be in the entry states of the start functions indeed, so then there would be no way around it I suppose. At least until maybe #178 when changes to global initializers would also mean changes to the constraint system itself.

jerhard: >But my understanding has been that it should from the beginning (global initialization) make global program variables into global constraint system variables, so g should be in st.

A potentiel issue here is that `set_start st` is called before solving. 
 https://github.com/goblint/analyzer/blob/1c21b2f82b1db1070c3dffc21598f75c2b325040/src/solvers/td3.ml#L230-L236
For global variables, this would set the value of the corresponding unknown to the initializer value. For soundness it then is required that all side effects to the global are triggered again.
I guess this currently actually happens, as writes to globals also contain a read; but this should not be relied upon.

sim642: Is there a reason to have such a `set_start`? Couldn't it just call `side` to set those, which wouldn't overwrite but would accumulate soundly?

The start values are a weird thing altogether: in #370 I realized I have to use them to construct transformed right-hand sides, which also join those desired start values. Because all they really are, is right-hand sides with a constant value.

stilscher: > I'm not sure why the mutex analysis should be incompatible with incremental because incremental analysis on the level of constraint systems shouldn't care whatsoever about the content of the constraint system.

Sorry, I mixed this up. I was thinking about the ids of threads that seem to be based on the location of the function to be executed. So this would be a problem for the consistency of ids across different versions in an incremental run.

PR #370, opened 2021-10-04
Combined generic postsolving mechanism

### LINKED ISSUE #367: Incomplete verify: queried variables are not checked
During #366 we had a bug, where the added restarting behavior made some regression tests fail in an unexpected manner.
Some asserts were never reached, so there was unsoundly dead code, but our verify didn't complain at all, but it should definitely have, since there were live constraints that weren't actually satisfied.

Turns out that verify is complete and doesn't check anything about the variables that were queries by `solve`. Instead it just checks that everything in the solution hashtable holds:
https://github.com/goblint/analyzer/blob/addf09589243a6d42a08e60fb899e7b1da210973/src/framework/constraints.ml#L1177

But there is no check that the queried variables are there and that their constraints are (recursively) satisfied all the way up.
With the current check, a solver could simply return an empty hashtable as the solution and since there's nothing to iterate over, verify would be happy with it, despite obviously missing our queried variables.

In a more subtle case (which is what we had), the solution might be a closed subset of the true solution, but be missing some actually reachable parts.

Hence, we need a proper verify must follow the inductive definition of a partial solution _w.r.t. the queried variables_. This might be a primitive top-down verify, which doesn't need to solve, but just check in demand-driven manner from the queried variables.

sim642:
Closes #367, among other things.

### Changes
1. Replace the different reachability, verify, warn, etc post-solving phases with a single top-down (i.e. demand-driven) postsolving phase. This takes the top-down reachability phase, but allows the other features to hook into the same pass over the constraints. This means that all post-solving features use the same right-hand side evaluation instead of repeating it multiple times. This mechanism is flexible enough to allow the side infl collecting of #363 to also be integrated into the single postsolving pass.
2. Fix queried and start variable verification (closes #367).
3. Create separate signatures for incremental and non-incremental solvers. Non-incremental solvers are lifted into the incremental signature with no incremental data. This conceptually means that they are also "incremental" but actually solve from scratch.
4. Remove mostly unused `Hash` module.


src/solvers/td3.ml
```diff
@@ -426,25 +416,29 @@ module WP =
          *   the old values, whereas with hashcons, we would get a context with a different tag, could not find the old value for that var, and have to recompute all vars in the function (without access to old values).
          *)
         if loaded && GobConfig.get_bool "ana.opt.hashcons" then (
+          let rho' = HM.create (HM.length data.rho) in
           HM.iter (fun k v ->
-            HM.remove data.rho k; (* remove old values *)
             (* call hashcons on contexts and abstract values; results in new tags *)
             let k' = S.Var.relift k in
             let v' = S.Dom.join (S.Dom.bot ()) v in
```
jerhard, line 423: see #371

PR #371, opened 2021-10-05
Fix relifting for incremental analysis with hashconsing

### LINKED ISSUE #361: Incremental Analysis: Detecting changed start states does not work when Hashconsing is enabled.
TD3 does not detect that a start state changed for a function when hashconsing is activated. 

Consider the following example code. After the first analysis, the initial value of the global is changed in the source code. If the example is incrementally analyzed with hashconsing, it is not the detected that the start state of `main` changed, so no reanalysis is triggered.
```
int g = 1; // Change the initial value to e.g. 5
int main(){
	int x = g;
	return x;
}
```
This has to do with the fact that with hashconsing activated only tags are compared that are erroneously duplicated.

jerhard:
Closes #361. Objects of type `S.Dom.t` are now relifted instead of joined with bottom. `Printable.Lift.relift` and `Printable.Lift2.relift` now properly delegate to the `relift` of the passed base-modules.


sim642 (commented):
What's so special about `Lift` and `Lift2` that their join with bottom doesn't ensure relifting while previously it was thought that joining with bottom is sufficient for all lattices?
I think I've wondered before why after adding `relift` for contexts we didn't apply it to domain values as well.

Also, there are plenty of other functors in `Printable`/`Lattice` that don't override the `Std.relift`? For example, `LiftBot` and `LiftTop`. Shouldn't all of those functors implement `relift` properly then?

src/domains/printable.ml
```diff
@@ -120,7 +120,6 @@ struct
   let name () = "HConsed "^Base.name ()
   let hash x = x.BatHashcons.hcode
   let tag x = x.BatHashcons.tag
-  let equal x y = x.BatHashcons.tag = y.BatHashcons.tag
```
michael-schwarz, line 123: why did we remove the equal here?

jerhard: There is a another copy of this defintion of `equal` lower down (line 142/143). So this first definition is just redundant.

vesalvojdani:
So, is there something outstanding on this one? No need to rush, but the number of branches and pull requests is monotonically increasing... we should aim for convergence soon. :)

jerhard:
>What's so special about Lift and Lift2 that their join with bottom doesn't ensure relifting while previously it was thought that joining with bottom is sufficient for all lattices?

They have an explicit ``Bot` element. A join with ``Bot` will return the other operand unchanged.

sim642:
> There is this `MapDomain.HashCached` functor, that just has the identity function as `relif`. Is this ok?

If there was hashconsing inside it, then I guess it should also `lift_f'` it. But we don't do that anywhere AFAIK and it also doesn't make much sense since hashconsing already caches hash values (among other things), so it wouldn't be very useful anyway.

PR #373, opened 2021-10-05
Fix TD3 narrowing non-wpoint

sim642:
While debugging increased evals in incremental analysis compared to from-scratch analysis (https://github.com/goblint/analyzer/pull/369#issuecomment-933254273), I found the following problem.
The TD3 implementation switches to narrowing phase even for non-wpoint variables, which causes extra evaluations. In the TD3 paper, the extra condition `x \in point` is used there.

Another question is how this check should be performed: `HM.mem wpoint x` or `wp` (the previous value cached from _before_ evaluating right-hand side). The choice of whether to widen is done based on the latter, so at widening points, the first is always an overwrite (no join or widen). The difference is in the case when `x` becomes a wpoint during the evaluation of its right-hand side.

I don't know yet whether this fixes all the excessive evals in the linked cases though, but at least on tiny examples from #363 it does.

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -130,21 +130,26 @@ module WP =
               else
                 box x old tmp
           in
+          if tracing then trace "sol" "Old value:%a\n" S.Dom.pretty old;
+          if tracing then trace "sol" "New Value:%a\n" S.Dom.pretty tmp;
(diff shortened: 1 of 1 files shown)


vogler:
Yes, this is definitely a mistake and should be merged.

PR #376, opened 2021-10-07
SARIF output format

AlexanderEichler:
Integrates the Sarif Format as a possible output for Goblint.
It is called with the option --sarif. 
An output file can be declared as normal with -o out.sarif (since Sarif is a Json file, the ending .json works too).
For testing the output there is a small script: scripts/createSarifTestOutput.sh.
One small error still exists in e.g. tests/regression/50-juliet/02, because a " is used in the warning text in Messages.ml Piece.t text. 

The Github Action can be used in C repositories, by adding it to the .github/workflows folder and configuring the path to the main file.


sim642 (approved):
I now did some fixing, cleanup and refactoring myself to get it suitable for merging.

sim642:
By the way, for anyone not on Slack, since I made a proper GitHub Action for Goblint, I tried it in action together with the SARIF support here: https://github.com/sim642/goblint-action-test. I think the Security tab is only visible to me there, so here's two screenshots:

### Alerts list
image
### First alert details
image

Since it doesn't require any location hacks anymore, it also shows the code snippet. Currently it's a bit silly because `endLine = startLine + 4` is hardcoded since Goblint cannot detect it.
This also shows where which texts are used. Clearly the `shortDescription` for the race condition rule is way too long.

Anyway, I'm a bit skeptical about declaring rules for Goblint anyway because that requires maintaining a consistent and complete rule ID scheme, which isn't very doable without major effort (categorizing all the unknown warnings).

PR #379, opened 2021-10-08
Relational Mutex-Meet-Privatization making use of Thread IDs

michael-schwarz:
This is the progress towards realizing the ideas from https://versioncontrolseidl.in.tum.de/schwarz/more-traces.

The implementation deviates from the description in the manuscript in the following way:
- We do not know all initial mutexes and protecting sets a priori, thus we cannot init the local state to set all variables protected by them to their initial values:
   - We always read from an unknown that contains the initial values, unless we know that the corresponding mutex must have been already published to. (amounting to a local write, making that reading superfluous). (`LMust` set)
- The join of values into local `L` at thread-join only happens if the thread is unique and must joined. This is a small optimization that preserves soundness but is potentially more precise. 
- The handling of single-threaded mode:
  - No publishing of single-threaded values of main thread until we go multithreaded (improving precision)
  
This subsumes #385, not sure how we want to proceed there?


michael-schwarz:
For, test 52/03, it succeeds if I add an upper bound on len, e.g. `len<6000000`. This seems very odd, as `len` should have an automatic upper bound (that of `int`) and is never computed with :thinking: .

https://github.com/goblint/analyzer/blob/b16a8aed8a1e10d904ff2e218bcf777b922a7857/tests/regression/52-apron-mukherjee/03-mukherjee_sssc12.c#L61

 Might it have something to do with our Apron overflow detection, @sim642 ?

sim642:
> Might it have something to do with our Apron overflow detection, @sim642 ?

As far as I see in the tracing output, there's a `side widen` on `next` which loses its upper bound and thus `next = next + 10` might overflow according to that check.

michael-schwarz:
Hmm, interesting!

I thought that `next + 10 <= len` ought to be enough to reestablish that `next <= len` but probably it is not because of overflows :(

michael-schwarz:
Actually `--set sem.int.signed_overflow assume_none` helps!

sim642:
> I thought that `next + 10 <= len` ought to be enough to reestablish that `next <= len` but probably it is not because of overflows :(

It actually can establish `next <= len` and `len <= 2147483647` as far as I can see, but that's not enough to prove `next + 10 <= 2147483647` which would exclude overflow.

I'm not sure what's happening, it somehow seems that Apron doesn't properly restrict to `next + 10 <= len`, which is the condition we have. I tried with polyhedra (for better readability) and it derives something weird: `10len-9next-10>=0`.

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -491,9 +492,10 @@ struct
   let name () = "W"
 end
 
-module type ClusterArg =
+module type ClusterArg = functor (AD: ApronDomain.S2) ->
 sig
   module LAD: Lattice.S
+  (* module AD: ApronDomain.S2 *)
```
jerhard, line 498: Commented out code?

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -503,10 +505,10 @@ sig
 end
 
 (** No clustering. *)
-module NoCluster: ClusterArg =
+module NoCluster:ClusterArg = functor (AD: ApronDomain.S2) ->
 struct
-  open CommonPerMutex
-
+  module AD = AD
```
sim642, line 509: These `module AD = AD` lines are unnecessary, no?

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -717,13 +714,15 @@ struct
...
 struct
-  open CommonPerMutex
+  open CommonPerMutex(AD)
   open ExplicitMutexGlobals
   include ConfCheck.RequireThreadFlagPathSensInit
 
-  module LAD = Cluster.LAD
+  module NC = Cluster(AD)
+  module Cluster = NC
+  module LAD = NC.LAD
```
sim642, line 724: What's `NC`? Just `module Cluster = Cluster (AD)` might be enough.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -834,12 +834,14 @@ struct
...
+    if Environment.equal x_env y_env  then
+      if GobConfig.get_bool "ana.apron.threshold_widening" && Oct.manager_is_oct Man.mgr then
+        let octmgr = Oct.manager_to_oct Man.mgr in
         let ts = Lazy.force widening_thresholds_apron in
-        let r = Oct.widening_thresholds Man.mgr (Abstract1.abstract0 x) (Abstract1.abstract0 y) ts in
-        {x with abstract0 = r}
+        let x_oct = Oct.Abstract1.to_oct x in
+        let y_oct = Oct.Abstract1.to_oct y in
+        let r = Oct.widening_thresholds octmgr (Abstract1.abstract0 x_oct) (Abstract1.abstract0 y_oct) ts in
+        Oct.Abstract1.of_oct {x_oct with abstract0 = r}
```
sim642: Instead of doing these (literally magic) conversions, it might be neater to just add a `widen_threshold` function to our own `Manager` signature, such that octagons can specially implement it in a safe way, while the others can just do normal widening for the time being.

sim642:
Also, what's with the CI failures in 52/15 and 52/16 now? Do those really take so long to analyze that they're borderline for the timeout?

src/analyses/apron/apronPriv.apron.ml
jerhard: Yes, my bad. They were necessary in some previous state when the `AD` needed to be re-exported. 
I can directly change that on `master`, if that's ok?

sim642: If they're there too, then sure. I thought it was somehow introduced in the merge conflicts, like the commented out code.

michael-schwarz:
> Do those really take so long to analyze that they're borderline for the timeout?

They do. Not sure why though.

sim642:
> They do. Not sure why though.

Just had a quick look. They have global variables which are completely unused (`OVERFLOW`, `arr1`, `arr2`). Just removing those makes them more than twice as fast to analyze. Not surprising, since you then have just 6 clusters instead of 21.

michael-schwarz:
I'd propose we merge this into `mater` now. Any cleanup that I should do before we do?

.github/workflows/locked.yml
```diff
@@ -43,6 +43,10 @@ jobs:
       - name: Test apron regression # skipped by default but CI has apron, so explicitly test group (which ignores skipping -- it's now a feature!)
         run: ruby scripts/update_suite.rb group apron -s
 
+      - name: Test apron regression (Mukherjee et. al  SAS '17 paper') # skipped by default but CI has apron, so explicitly test group (which ignores skipping -- it's now a feature!)
+        if: ${{ matrix.apron }}
```
sim642, line 47: This conditional is unnecessary in the locked workflow, where apron is always enabled via the lockfile. Given that this variable doesn't exist, does this even run?

conf/traces-rel.json
```diff
@@ -0,0 +1,52 @@
...
+        "ldv_malloc",
+
+        "kzalloc_node",
+        "ldv_zalloc",
+        "kmalloc_array",
+        "kcalloc",
+
+        "ldv_xmalloc",
+        "ldv_xzalloc",
+        "ldv_calloc"
```
sim642, line 43: This list is now longer than the svcomp confs have. Where did these come from?

src/analyses/threadJoins.ml
```diff
@@ -27,8 +27,12 @@ struct
   let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
     match LibraryFunctions.classify f.vname arglist with
     | `ThreadJoin (id, ret_var) ->
-      (
-        let threads = TIDs.elements (ctx.ask (Queries.EvalThread id)) in
+      let threads = ctx.ask (Queries.EvalThread id) in
+      if TIDs.is_top threads then
+        ctx.local (* TODO: why needed? *)
+      else (
+        (* elements throws if the thread set is top *)
```
sim642, line 34: Does this still happen after #388?

src/cdomains/threadIdDomain.ml
```diff
@@ -39,6 +39,9 @@ sig
 
   val threadenter: t * D.t -> Node.t -> varinfo -> t
   val threadspawn: D.t -> Node.t -> varinfo -> D.t
+
+  (** If it is possible to get a list of unique thread create thus far, get it *)
+  val created: t -> D.t -> (t list) option
```
sim642, line 44: Does this need to be `t list option` or could it just be `t list`, which returns `[]` when it's not possible to get a list?

src/util/defaults.ml
```diff
@@ -179,7 +180,9 @@ let _ = ()
       ; reg Experimental "exp.privatization"     "'protection-read'" "Which privatization to use? none/protection-old/mutex-oplus/mutex-meet/protection/protection-read/protection-vesal/mine/mine-nothread/mine-W/mine-W-noinit/lock/write/write+lock"
       ; reg Experimental "exp.priv-prec-dump"    "''"    "File to dump privatization precision data to."
       ; reg Experimental "exp.priv-distr-init"   "false"  "Distribute global initializations to all global invariants for more consistent widening dynamics."
-      ; reg Experimental "exp.apron.privatization" "'mutex-meet'" "Which Apron privatization to use? dummy/protection/protection-path/mutex-meet"
+      ; reg Experimental "exp.apron.privatization" "'mutex-meet-tid-cluster12'" "Which apron privatization to use? dummy/protection/protection-path/mutex-meet"
```
sim642, line 183: This description has an outdated list of privatizations now.

Also, do we want to make `tid-cluster12` the default then. The default base privatization is one without thread IDs though.

src/util/defaults.ml
```diff
@@ -200,7 +203,7 @@ let _ = ()
       ; reg Experimental "exp.no-narrow"         "false" "Overwrite narrow a b = a"
       ; reg Experimental "exp.basic-blocks"      "false" "Only keep values for basic blocks instead of for every node. Should take longer but need less space."
       ; reg Experimental "exp.solver.td3.term"   "true"  "Should the td3 solver use the phased/terminating strategy?"
-      ; reg Experimental "exp.solver.td3.side_widen" "'sides'" "When to widen in side. never: never widen, always: always widen, sides: widen if there are multiple side-effects from the same var resulting in a new value, cycle: widen if a called or a start var get destabilized, unstable_called: widen if any called var gets destabilized, unstable_self: widen if side-effected var gets destabilized."
+      ; reg Experimental "exp.solver.td3.side_widen" "'sides-pp'" "When to widen in side. never: never widen, always: always widen, sides: widen if there are multiple side-effects from the same var resulting in a new value, cycle: widen if a called or a start var get destabilized, unstable_called: widen if any called var gets destabilized, unstable_self: widen if side-effected var gets destabilized."
```
sim642, line 206: I guess this includes the workaround from #390, which wasn't discussed in any other PR? Is this a hack we should make the default or should it not (and just be part of the `traces-rel` conf)?

tests/regression/52-apron-mukherjee/01-mukherjee_reorder_2.c
```diff
@@ -0,0 +1,70 @@
+// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --set ana.activated[+] threadJoins
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <pthread.h>
+#include "assert.h"
```
sim642, line 6: It would be more canonical to include `<assert.h>` instead. This also applies to all the following Muhkerjee benchmarks it seems.

sim642:
There are two older comments, which have possibly been buried in the history here:
1. https://github.com/goblint/analyzer/pull/379/files/691be385371d514564d7e3a4c6b56960bef51354..128b1d85e0fdb1d089a73f3f6501e506e0a7339b#diff-58830344d2e5f5b8023333061fe1170c8a94a5e1ba32b98ccf6739eaf20d9123
2. https://github.com/goblint/analyzer/pull/379/files/691be385371d514564d7e3a4c6b56960bef51354..128b1d85e0fdb1d089a73f3f6501e506e0a7339b#diff-f389b2fa9a7f2374023c34533d99d4ed44f8da04d6f46bec9a089ab9ee67b2bd

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -466,112 +481,558 @@ struct
...
+    List.cartesian_product gs gs
+    |> List.filter (fun (g1, g2) -> CilType.Varinfo.compare g1 g2 <= 0) (* filter flipped ordering, keep equals for next step *)
+    |> List.map (fun (g1, g2) -> [g1; g2]) (* if g1 = g2, then we get a singleton cluster *)
+
+  let name () = "cluster12"
+end
+
+(** All clusters of size 2. *)
+(* Caution: It is questionable if this works 100%, see comment https://github.com/goblint/analyzer/pull/417/files#diff-58830344d2e5f5b8023333061fe1170c8a94a5e1ba32b98ccf6739eaf20d9123 *)
+module Clustering2: ClusteringArg =
```
jerhard, line 548: Is this resolved?

sim642: I think #424 resolved it.

src/util/defaults.ml
michael-schwarz: I guess we should just make regular `mutex-meet` the default then.

src/util/defaults.ml
```diff
@@ -203,7 +203,7 @@ let _ = ()
...
       ; reg Experimental "exp.basic-blocks"      "false" "Only keep values for basic blocks instead of for every node. Should take longer but need less space."
       ; reg Experimental "exp.solver.td3.term"   "true"  "Should the td3 solver use the phased/terminating strategy?"
-      ; reg Experimental "exp.solver.td3.side_widen" "'sides-pp'" "When to widen in side. never: never widen, always: always widen, sides: widen if there are multiple side-effects from the same var resulting in a new value, cycle: widen if a called or a start var get destabilized, unstable_called: widen if any called var gets destabilized, unstable_self: widen if side-effected var gets destabilized."
+      ; reg Experimental "exp.solver.td3.side_widen" "'sides'" "When to widen in side. never: never widen, always: always widen, sides: widen if there are multiple side-effects from the same var resulting in a new value, cycle: widen if a called or a start var get destabilized, unstable_called: widen if any called var gets destabilized, unstable_self: widen if side-effected var gets destabilized, sides_pp: widen if there are multiple side-effects from the same program point resulting in a new value."
```
sim642, line 206: `sides_pp` is a typo here.

src/cdomains/threadIdDomain.ml
michael-schwarz: It needs to be `t list option` to distinguish between whether no threads are created yet, or whether such information is not available in the current domain.

https://github.com/goblint/analyzer/blob/e5d63c856d8284cd9fb4a85a33408333ba28a6eb/src/analyses/threadId.ml#L89-L97

conf/traces-rel.json
michael-schwarz: They are also in the `traces.json` configuration, so I copied them from there

src/analyses/threadJoins.ml
michael-schwarz: Yes, e.g. in `04/58` it does when I enabled this analysis. Seen that `ctx.local` is always sound here, I'd suggest to keep it such that this analysis doesn't cause exceptions in strange setting such as `04/58` or some other weirdness that one might encounter for thread ids.

michael-schwarz:
The failing indentation seems to be unrelated, as this file was not touched in the PR.
If the changes sufficiently addresses your comments @sim642, I suggest we merge this now.

conf/traces-rel.json
sim642: Interesting, I'll look into whether they should also be part of the svcomp conf (for next year).

sim642 (approved):

sim642:
> The failing indentation seems to be unrelated, as this file was not touched in the PR.

I fixed it directly on master, because a change there caused to be already misindented code to appear in unrelated diffs now.

PR #396, opened 2021-10-14
`OPAMCLI=2.0` for `make deps` to allow `--unlock-base` for opam 2.1.0

vogler:
opam 2.1.0 otherwise fails:

--unlock-base was removed in version 2.1 of the opam CLI, but version 2.1 has been requested. Use --update-invariant instead or set OPAMCLI environment variable to 2.0.

Was there a need to have it?

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -88,7 +88,7 @@ rule() {
 
     # setup, dependencies
     ;; deps)
-      opam update; OPAMCLI=2.0 opam install -y . --deps-only --locked --unlock-base
+      opam update; OPAMCLI=2.0 opam install -y . --deps-only --locked --unlock-base; opam upgrade $(opam list --pinned -s)
     ;; setup)
       echo "Make sure you have the following installed: opam >= 2.0.0, git, patch, m4, autoconf, libgmp-dev, libmpfr-dev"
       echo "For the --html output you also need: javac, ant, dot (graphviz)"


sim642:
This is needed for upgrading the OCaml version. Otherwise it will refuse to upgrade anything since the OCaml version in the installed switch is locked, but our dependencies (especially our lock file) asks for a different version.

sim642:
> I meant if we need it to be part of this `make deps`. If a user has a switch where this is the case, opam will complain and they can add `--unlock-base` or `--update-invariant` depending on which opam version they use.

Kind of defeats the point if having a target for it if you have to manually copy the command out and know what to add to it.

> Alternative is that we keep `--unlock-base` and prefix `OPAMCLI=2.0`.

If it works, then it seems like a reasonable solution for now.

vogler:
> We could add `opam upgrade goblint-cil ppx_deriving zarith apron`, but that adds another place to keep in sync for pins.

Added `opam upgrade $(opam list --pinned -s)`.

This adds a section "Synchronising pinned packages".
"Synchronising development packages" from before was not enough to make it upgrade?

```console
$ make deps

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><>  🐫
[default] no changes from https://opam.ocaml.org

<><> Synchronising development packages <><><><><><><><><><><><><><><><><><>  🐫
[apron.v0.9.13] synchronised (no changes)
[goblint-cil.1.8.1] synchronised (no changes)
[ppx_deriving.5.2.1] synchronised (no changes)
[zarith.1.12-gob0] synchronised (no changes)
Nothing to do.

<><> Synchronising pinned packages ><><><><><><><><><><><><><><><><><><><><>  🐫
[ppx_deriving.5.2.1] synchronised (no changes)
[apron.v0.9.13] synchronised (no changes)
[goblint-cil.1.8.1] synchronised (no changes)
[zarith.1.12-gob0] synchronised (no changes)

Already up-to-date.
Nothing to do.
```

sim642:
> This has always confused me as well because I think "Synchronising development packages" will show now versions of pins, but then claim there's nothing to do, as if those versions are now installed, while they yet aren't.

Interestingly, I just had to update to opam 2.1 for dune-release and with opam 2.1 it seems that just installing the new pins does indeed immediately upgrade them as well, which makes a lot of sense. The explicit upgrade command can stay though because not everyone has opam 2.1 set up.

PR #400, opened 2021-10-19
Comparing incremental and original runs

vesalvojdani:
I'm getting some unexpected results on the comparison of these incremental runs on the benchmark. But I'm not sure if `compare_runs` does the right thing here. If the expectation is that incremental should be sound w.r.t. original solution, then this is not what compare runs [outputs](https://kodu.ut.ee/~vesal/results/211018-interactive/aget_comb01.patch.Reluctant.txt.compare.txt).

I don't trust the benchmarking script to get the paths right, but I think the simple script in this branch is fairly simple, and already `./scripts/test-incremental.sh 00-local` seems to show incremental being more precise. Could someone look over this pull request to check if the way I apply compare_runs should result in the incremental being greater-equal? Is there some way this save_runs still outputs unreachable parts of the solution or something?


vesalvojdani:
But probably you're right in that even the new program will probably have different ids for program points. Simmo suggested that maybe one could do the incremental construction of the CFG, but then solve it all from scratch. Hmm...

Issue #398, opened 2021-10-14
Install Goblint to opam switch


sim642:
It would be nice (especially for IDE integration) that one can install Goblint (from the source in this repo) locally into an opam switch, to make it available as `goblint`, not `./goblint`. This partially works, but auxiliary file handling is broken: such Goblint always complains about not finding its own includes directory.

It's possible to configure what extra files to install via dune. Additionally at runtime it must be possible to get the path/prefix where dune/opam install has installed the auxiliary files such that it works both in the development directory directly and as installed.

sim642:
> It's not a problem for c-files since they're passed to cpp first?

For C files there's no problem.

> What about configs, runs etc.?

If we want to distribute our standard `conf/` files with the installed Goblint such that they work globally, then they'd have to be installed in some way. But that's not the biggest concern even, since we don't have to do that and simply require users to have their own conf files whereever they choose (e.g. in their project, like GobPie will require).

Saved runs, incremental marshaling, etc should go to working directory anyway, so shouldn't be a problem. If anything we might at some point make all their paths consistent to have stuff be in `.goblint`.

> Could we just prefix `Sys.getcwd ()`?

It's not that simple. Developing in our repository, you have `./includes` relative to the binary and the working directory. If installed into opam, it will be installed into some prefix, where the binary will be `<prefix/bin/goblint` and the auxiliary files will not be in `<prefix>/share/goblint/includes` or something similar. The relative paths between everything changes, so we should do things properly and rely on the intended dune mechanisms.

sim642:
> Didn't get the last part. "the auxiliary files will not be in `<prefix>/share/goblint/includes` or something similar" - not be?

Hmm yes, I think I typo-ed something because yes, they will be in `<prefix>/share/goblint/includes`, but not in `<prefix>/bin/includes`.

> My idea was for files that we ship to first check `Sys.getcwd ()` and then the dune install path if a relative path is given.

Indeed, I have implemented something along those lines in the `install` branch, where these things mostly work, but there are things to still clean up.

PR #407, opened 2021-10-20
Import Gobview

keremc:
* Import the source code of Gobview into the `gobview` subdirectory.
* Not using a submodule since Gobview is tightly coupled with the main analyzer code anyway.
* See [this link](https://github.com/keremc/analyzer/blob/gobview/docs/user-guide/inspecting.md) for the documentation and build/installation steps.


keremc:
@sim642 Thanks for typing up most of what I wanted to say. Gobview doesn't make sense as an independent project because of how tightly coupled it is with Goblint. I guess a submodule might still make sense for organizational purposes. (e.g., to keep the project histories separate) It is also why I decided against using subtrees and filter-branch/filter-repo. I wanted to avoid cluttering up the Goblint history with ~500 commits.

The imported version of Gobview has a few differences, but most are configuration-related. dune now automatically installs the npm dependencies and executes `webpack build`, whereas I previously performed these steps manually. Also, I removed `goblint` from the list of pinned dependencies of Gobview.

sim642:
>  It is also why I decided against using subtrees and filter-branch/filter-repo. I wanted to avoid cluttering up the Goblint history with ~500 commits.

Commits are cheap in git, so that's not really a problem IMO.

> Gobview doesn't make sense as an independent project because of how tightly coupled it is with Goblint. I guess a submodule might still make sense for organizational purposes. (e.g., to keep the project histories separate)

A submodule would still keep it quite independent though, because changing Goblint without having the submodule checked out (or not building it all the time) could still break Gobview and require fixes. Although it's probably easier to develop Gobview that way since there's no need to do any opam pinning locally for all the experimentation.

Gobview has a dependency on the `goblint` package, which in its repository currently has to be provided through an opam pin. If gobview were a submodule (so a subdirectory) here, would dune just automatically find the goblint package from the sibling directory and use it directly? Or does it require the goblint package to actually be installed into the local switch (#398)? Or does it complain about something else?

Because the submodule idea might not be as bad as I first believed. Currently g2html (and some webapp) both are submodules, which is convenient. The coupling is two-way anyway since Goblint has gobiew, etc specific options anyway.

keremc:
OK, I've now pushed a version that uses submodules. The submodule tracks the `goblint` branch of [Gobview](https://github.com/goblint/gobview/tree/goblint).

michael-schwarz:
From my perspective, we can merge then. @sim642?

sim642:
I'm now having some kind of problems building it.

At one point the `index.html` etc symlinks in the run directory were broken and pointing to `_build/default/gobview/dist/index.html`, which didn't exist for some reason after `dune build gobview`.

Now I've tried cleaning things up and doing `dune build gobview` and it immediately fails with
```
         npx gobview/dist/main.js (exit 1)
(cd _build/default/gobview && /usr/bin/npx webpack build)
CLI for webpack must be installed.
  webpack-cli (https://github.com/webpack/webpack-cli)

We will use "npm" to install the CLI via "npm install -D webpack-cli".
Do you want to install 'webpack-cli' (yes/no): npm ERR! code 1
npm ERR! path /home/simmo/dev/goblint/sv-comp/goblint/_build/default/gobview
npm ERR! command failed
npm ERR! command sh -c webpack "build"

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/simmo/.npm/_logs/2021-11-02T08_12_28_650Z-debug.log
```

I don't want to have to completely nuke my local Goblint repository just to get this working and make any changes. There must be a reason behind this.

Looking at the dune files, I don't really understand the dependency structure inside Gobview. Why is there a rule for the `npm.log` target? And where's the rule for `dist/index.html` etc targets? There's `dist/dune` with just the `gobview` alias, but that doesn't seem to declare everything it depends on and it creates, so dune doesn't rebuild everything properly if I don't change those couple of files or something.

sim642:
Definitely something weird happening: if I do `dune build gobview` a second time after the first clean build is successful, then in `_build/default/gobview` `node_modules` somehow gets deleted and so does `dist/index.html`. But the weird `npm.log` rule, which calls `npm ci` to install things (using the incorrect npm function even), doesn't do anything because its only dependency `package-lock.json` did not change, so there's no dependencies around, not even `webpack`.

keremc:
It should build correctly now.

> But the weird npm.log rule, which calls npm ci to install things (using the incorrect npm function even), doesn't do anything because its only dependency package-lock.json did not change, so there's no dependencies around, not even webpack.

I remember having some issues with directory targets when I first started maintaining Gobview, but apparently, they don't apply in this case. So, I removed the `npm.log` rule, which was just a workaround acting as a timestamp for the last npm run.

I'm using `npm ci`, because it respects the lock file. (which is the default behavior for newer package managers like Yarn AFAIK)

sim642:
I guess that might be solved now, or at least multiple attempts to build it didn't have that problem anymore. But now that master was merged into this, there seems to be a problem with the dune-site integration:
```
File "gobview/src/.App.eobjs/dune_site_data.ml-gen", line 1:
Error: Could not find the .cmi file for interface
       gobview/src/.App.eobjs/dune_site_data.ml-gen.
```
I'm not sure if there's anything to be done about it or if it's just a bug in dune/js_of_ocaml.

sim642:
I overengineered a workaround to the dune-site problem via virtual libraries. I also made a bunch of other fixes to both this PR and Gobview, so now it seems to work.

The only thing which doesn't is clicking on the source file name in Gobview: it's still trying to access some original path that was passed as argument for Goblint but which isn't really relative to the `save_run` directory.

.gitmodules
```diff
@@ -4,3 +4,7 @@
 [submodule "g2html"]
 	path = g2html
 	url = https://github.com/goblint/g2html.git
+[submodule "gobview"]
+	path = gobview
+	url = https://github.com/goblint/gobview.git
+	branch = goblint
```
sim642, line 10: I think this can be merged now, but maybe we should move the contents of `goblint` branch into Gobview's `master` branch and omit this alternative branch name here before we merge?

PR #410, opened 2021-10-26
Add semgrep rules for List.length and List.combine

sim642:
Based on old and new observations in #409 I decided to write some semgrep rules to find more cases and we have a bunch.
Currently this set of rules finds 53 problems.

I will fix them once #409 is merged to avoid merge conflicts.


src/analyses/spec.ml
```diff
@@ -62,7 +62,7 @@ struct
         else if may then D.may_goto key loc state m else D.goto key loc state m
 
     (* match spec_exp, cil_exp *)
```
michael-schwarz, line 64: I guess the comment is superfluous now?

michael-schwarz (approved):

PR #417, opened 2021-10-27
Relational mutex-meet clustering

sim642:
Instead of duplicating the entire mutex-meet-tid analysis, which would be annoying to merge if #379 gets some refactorings, I managed to surgically generalize one domain it uses to make it work for the current no-clustering case and also the clustering case.

The trick is replacing the values in the `L` domain from relations (`AD`) with maps cluster→relation. Exactly the same is done in the global domain for mutexes (under thread IDs).
Since the no-clustering logic joins the compatible thread relations, now that join happens over the clustering maps pointwise. This naturally performs the inner join at clustered `lock`. Since the global and local structures use the same abstraction, the joining of those also happens naturally pointwise.
The only real logic for the clustering case is then to meet over the values of the resulting map.

For unlocking, the clustering just explodes the side-effected relation into the map. The clustering has clusters of size 1 or 2, although it'd be very easy to also support full powerset for experimental reasons.

It passes all the previous tests and also the clustering examples from the more traces article. Moreover, if disabling 1-clusters (and only keeping 2-clusters) the second example fails as expected.

### DIFF
--- a/src/analyses/apron/apronPriv.apron.ml
+++ b/src/analyses/apron/apronPriv.apron.ml
@@ -512,8 +512,56 @@ struct
     oct_side
 end
 
+module type ClusteringArg =
+sig
+  val generate: varinfo list -> varinfo list list
+end
+
 (** All clusters of size 1 and 2. *)
-module Cluster12: ClusterArg =
+module Clustering12: ClusteringArg =
+struct
+  let generate gs =
+    List.cartesian_product gs gs
+    |> List.filter (fun (g1, g2) -> CilType.Varinfo.compare g1 g2 <= 0) (* filter flipped ordering, keep equals for next step *)
+    |> List.map (fun (g1, g2) -> [g1; g2]) (* if g1 = g2, then we get a singleton cluster *)
+end
+
+(** All clusters of size 2. *)
+module Clustering2: ClusteringArg =
+struct
+  let generate gs =
+    List.cartesian_product gs gs
+    |> List.filter (fun (g1, g2) -> CilType.Varinfo.compare g1 g2 < 0) (* filter flipped ordering, forbid equals for just clusters of size 2 *)
+    |> List.map (fun (g1, g2) -> [g1; g2])
+end
+
+(** All subset clusters. *)
+module ClusteringPower: ClusteringArg =
+struct
+  let generate gs =
+    gs
+    |> List.map (fun _ -> [true; false])
+    |> List.n_cartesian_product (* TODO: exponential! *)
+    |> List.map (fun bs ->
+      (* list globals where omega is true *)
+      List.fold_left2 (fun acc g b ->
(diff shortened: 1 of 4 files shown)


sim642:
>    * After [67abb3d](https://github.com/goblint/analyzer/commit/67abb3d305b195c3d61241c0bb551e5fc1c5a012) only `mutex-meet-tid-cluster2` is broken.

In https://github.com/goblint/analyzer/pull/417/commits/df90c19d32d3374623d5afcdfa52a6922ec67913 I mostly fixed this by having clusters of size 1 when only one variable is even being protected (to allow all `m_g` to have values). Somehow it still fails 36/22. Not sure if we care enough about that when `cluster12` does work properly and that's what we really wanted anyway.

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -474,30 +474,183 @@ struct
...
+        else
+          None
+      ) octs
+
+  let lock local_m get_m =
+    let joined = LAD.join local_m get_m in
+    if M.tracing then M.traceli "apronpriv" "cluster12 lock:\n  local=%a\n  get=%a\n  joined=%a\n" LAD.pretty local_m LAD.pretty get_m LAD.pretty joined;
+    let r = LAD.fold (fun _ -> AD.meet) joined (AD.bot ()) in (* bot is top with empty env *)
+    if M.tracing then M.traceu "apronpriv" "-> %a\n" AD.pretty r;
+    r
```
michael-schwarz, line 615: How does this take care of clusters that are too large initially, where the smaller subclusters are not tracked?
Consider e.g. at some mutex m, we initially considered it protecting for `{g,h,i}`, but later discovered that it only protects `{g,h}`:

In the case of Cluster2, we have clusters `{{g,h},{h,i}.{g,i}}`. Later, we will only publish to the cluster `{g,h}`. We will however still read things from the clusters `{h,i}` (restricted to `{h}`) and `{g,i}` (restricted to `{g}`). These clusters never get any new contributions after we have found out that `i` is not protected.

At the `lock(m)`, however, we join cluster-wise over all different threads (including those ones that are restricted to `{g}` and `{h}`), and then **meet** the result of the different clusters. Would this meet then not include the value for the clusters `{g}` and `{h}`, that does not soundly describe all values of `g` and `h`? Or am I missing something here?

sim642: This probably still is why `cluster2` doesn't work completely. It's exactly the decreasing locksets problem that I've been having issues with. Not sure what the solution is here because at some point you do need to take the values from the larger lockset and have it propagated into the smaller one.

So far I tried to somehow do it at read time, like we've been doing for everything else, but maybe it's not fully possible here and there needs to be an explicit step for it when protecting locksets change.

PR #418, opened 2021-10-28
Make Apron domain configurable 

### LINKED ISSUE #415: Add support for Apron Box domain to relational analysis
Add support for the Apron Box domain to our relational analysis.

jerhard:
This PR makes the Apron domain configurable using the option `"ana.apron.domain"`. The supported options right now are 'octagon', 'interval' and 'polyhedra'.
Making the Apron-Domain configurable required making several modules a functor that directly or indirectly depend on the `Manager` that is needed for interfacing with the corresponding Apron domain.

Closes #415

### DIFF
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -6,17 +6,17 @@ open ApronDomain
 
 module M = Messages
 
-module SpecFunctor (Priv: ApronPriv.S) : Analyses.MCPSpec =
+module SpecFunctor (AD: ApronDomain.S2) (Priv: ApronPriv.S) : Analyses.MCPSpec =
 struct
   include Analyses.DefaultSpec
 
   let name () = "apron"
 
-  module D = ApronComponents (Priv.AD) (Priv.D)
+  module Priv = Priv(AD)
+  module D = ApronComponents (AD) (Priv.D)
   module G = Priv.G
   module C = D
 
-  module AD = Priv.AD
   open AD
 
   let should_join = Priv.should_join
@@ -390,8 +390,10 @@ end
 
 let spec_module: (module MCPSpec) Lazy.t =
   lazy (
+    let module Man = (val ApronDomain.get_manager ()) in
+    let module AD = ApronDomain.D2 (Man) in
     let module Priv = (val ApronPriv.get_priv ()) in
-    let module Spec = SpecFunctor (Priv) in
+    let module Spec = SpecFunctor (AD) (Priv) in
     (module Spec)
   )
 
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -27,8 +27,8 @@ struct
     else
       D.bot () (* just like startstate, heterogeneous AD.bot () means top over empty set of variables *)
 
-  let exitstate  _ = { oct = AD.bot (); priv = Priv.startstate () }
-  let startstate _ = { oct = AD.bot (); priv = Priv.startstate () }
+  let exitstate  _ = { apr = AD.bot (); priv = Priv.startstate () }
+  let startstate _ = { apr = AD.bot (); priv = Priv.startstate () }
 
   (* Functions for manipulating globals as temporary locals. *)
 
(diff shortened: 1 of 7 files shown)


src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -12,11 +12,12 @@ struct
 
   let name () = "apron"
 
-  module D = ApronComponents (Priv.D)
+  module D = ApronComponents (Priv.AD) (Priv.D)
```
sim642, line 15: Here privatization determines the local domain to be used. This seems backwards. Instead `AD` should be given as argument to `Priv`.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -768,61 +815,175 @@ struct
...
+  include DWithOps (Man) (DHetero (Man))
+  module Man = Man
+  (* Copy-paste from BaseDomain... *)
+  type 'a aproncomponents_t = {
+    oct: t; (* TODO: rename not to be just "oct" *)
+    priv: 'a;
+  } [@@deriving eq, ord, to_yojson]
+end
 
+module type S2 =
```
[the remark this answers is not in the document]
sim642: I guess this is supposed to be the signature for the module returned by `D2`. Might be better to define it before and use it for `D2`.

Regarding the signature, maybe `module type of` can help: https://ocaml.org/manual/moduletypeof.html.

src/analyses/apron/apronAnalysis.apron.ml
jerhard: At the instantiation site of the Privatization, the `AD` is required as an argument: 
https://github.com/goblint/analyzer/blob/3bd9fa731065542c99ba4302ef6aebab56bfcd82/src/analyses/apron/apronPriv.apron.ml#L682-L698
The passed module is then stored as module `AD` within the privatizations. This module is then retrieved here again. 

If one were to change the quoted location to return a functor from `S2` to `ApronPriv.S`, one would also have to apply the `TracingPriv` functor in the Analysis, because one would not have a `ApronPriv.S` to pass to it in `ApronPriv`. 
Would this be preferable overall?

sim642: Couldn't one just change `ApronPriv.S` to be `functor (AD: S2) -> sig (* the current signature *) end`? Should still be able to make `TracingPriv` do such lifting.

jerhard: One could alternatively also expose the `ApronComponents` modules of the privatizations, so one would simply use `Priv.ApronComponents` here. But for most of the analysis, one does not really need the whole `ApronComponents` module, but only its type `t`.

src/cdomains/apron/apronDomain.apron.ml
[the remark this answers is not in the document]
jerhard: Now `D2` is annotated with `S2` and the signature reuses more signatures of modules.

src/analyses/apron/apronAnalysis.apron.ml
jerhard: >Couldn't one just change ApronPriv.S to be functor (AD: S2) -> sig (* the current signature *) end? Should still be able to make TracingPriv do such lifting.

Yes, I'll try it.

jerhard: Done.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -768,71 +815,102 @@ struct
...
+sig
+  module Man: Manager
+  module Tracked : Tracked
+  module Bounds : module type of Bounds (Man)
+  include module type of AOps (Tracked) (Man)
+  include Tracked
+  include SLattice with type t = Man.mt A.t
+
+  type 'a aproncomponents_t = { apr : t; priv : 'a; }
```
sim642, line 827: It's a bit weird that this is defined here, because nothing needs it in this module. Instead it could be defined later outside of modules using two type parameters if necessary for `ApronComponents`.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -768,71 +815,102 @@ struct
...
+  val eval_int : t -> exp -> IntDomain.IntDomTuple.t
+
+  val equal_aproncomponents_t :
+    ('a -> 'a -> bool) ->
+    'a aproncomponents_t -> 'a aproncomponents_t -> bool
+  val compare_aproncomponents_t :
+    ('a -> 'a -> int) ->
+    'a aproncomponents_t -> 'a aproncomponents_t -> int
+  val aproncomponents_t_to_yojson :
+    ('a -> Printable.json) -> 'a aproncomponents_t -> Printable.json
```
sim642, line 843: It would be good to get `aproncomponents_t` out of here, because it's an awkward place for these. If it's not possible, then I think it's also possible to use `[@@deriving ...]` inside the signature as well.

src/cdomains/apron/apronDomain.apron.ml
jerhard: It's extracted now. You might have a look at it again.

src/cdomains/apron/apronDomain.apron.ml
jerhard: Now the deriving is done for the type outside the signature.

sim642 (approved):

Issue #101, opened 2020-09-24
Hoare powerset domain widening doesn't guarantee ascending chain condition and more


sim642:
### Links
In my attempt to understand the Hoare powerset domain (`SetDomain.Hoare`) used for path-sensitivity I happened upon the following:
* [some slides from Miné](https://www.di.ens.fr/~rival/semverif-2017/sem-13-ai.pdf) (slide 20),
* and referred from those: [Bagnara et al. Widening Operators for Powerset Domains](https://www.cs.unipr.it/~bagnara/Papers/PDF/STTT06.pdf).

### Issue
Our current widening is implemented as follows:
https://github.com/goblint/analyzer/blob/08316c5560b1c1298d8b0b29af4b076f12de0eb5/src/domains/setDomain.ml#L623-L628
This corresponds exactly to Definition 7 from the latter paper. However, it is only called an "extrapolation heuristics" and not "widening" because it doesn't guarantee the ascending chain condition. So we don't do true widening there.

### Possible fix
The same paper also presents three different generic proper widening operators for such a domain using the inner domain's widening. Implementing any of them might not be straightforward though because they all require some kind of additional operator.

### Merging in Hoare domains
The paper also discusses merging of elements in a Hoare set. Therefore such notion seems closely related but Goblint has two different Hoare powerset domains with quite different characteristics:
1. Path-sensitivity lifter (`PathSensitive2`) domain is defined through `SetDomain.Hoare` with additional joining of elements wrapped around in `reduce_join` through `should_join`s from inner `Spec`s.
2. `AddressSet` domain is defined through `SetDomain.HoarePO` but the partitioning of elements into joined sets is deeply encoded into `HoarePO` through questionable means. It assumes the inner domain's operators (e.g. `join`) only act on elements which should be joined and raise `Lattice.Incomparable` on anything else. From what I understand, this just encodes a `should_join`-like query through exceptions while making no sense as a lattice (two elements can always be joined).

If the `should_join`-like operator is passed to a hoare powerset domain functor and the merging of elements by equivalence is moved to a single place, a single implementation should do. It would remove the need for exception-based control flow hacks while allowing an optimized bucket-based Hoare domain implementation to still be used.

vogler:
> From what I understand, this just encodes a should_join-like query through exceptions while making no sense as a lattice (two elements can always be joined).

[...]

PR #419, opened 2021-10-29
Add set struct domains

jakoberzar:
The existing struct abstract domain in Goblint (`Simple`) is a mapping from struct fields to their abstract values. It does not keep any information about relations between the fields.
We have tried to improve the struct domains in Goblint by adding two new domains (`Sets` and `KeyedSets`) that keep this information by using a set of mappings from fields to abstract values.
We have also added a third domain that can be picked within `FlagConfiguredStructDomain` that selects the struct domain dynamically depending on the struct field types.

I have squashed all of my commits into one.
Feel free to suggest improvements to the code. Currently, some code is kind of duplicated between Sets & KeyedSets (and some with HoareSet), I don't know how to solve that best.


michael-schwarz (approved):
I suggest we merge this and then do any further required changes by need.

src/cdomains/structDomain.ml
```diff
@@ -90,3 +105,496 @@ struct
...
+  let pretty_diff () (x,y) =
+    Pretty.dprintf "{@[%a@] ...}" HS.pretty_diff (x,y)
+  let printXml f xs = HS.printXml f xs
+
+  let widen_with_fct f =
+    let product_widen op a b = (* assumes b to be bigger than a *) (* from HS.product_widen *)
+      let xs,ys = HS.elements a, HS.elements b in
+      List.map (fun x -> List.map (fun y -> op x y) ys) xs |> List.flatten |> fun x -> HS.of_list (List.append x ys)
+    in
+    product_widen (fun x y -> if SS.leq x y then (SS.widen_with_fct f) x y else SS.bot ())
```
sim642, line 207: Why is `widen` not redefined via this?

src/cdomains/structDomain.ml
```diff
@@ -90,3 +105,496 @@ struct
...
+
+  let leq_with_fct f a b =
+    let mem x s f = HS.exists ((SS.leq_with_fct f) x) s in (* from HS.mem *)
+    HS.for_all (fun x -> mem x b f) a
```
sim642, line 211: Why is `leq` not redefined via this?

src/cdomains/structDomain.ml
```diff
@@ -90,3 +105,496 @@ struct
...
+              let res = join_f acc el in
+              if Messages.tracing then Messages.tracel "simplesets-fct" "Join-fct joining others!\nacc: %a\nel: %a\nres: %a\n" SS.pretty acc SS.pretty el SS.pretty res;
+              res
+            ) h overlapping in
+          aux (joined::rem_uniq) t
+      in aux [] xs
+    in
+    let res = reduce_list_with_fct (SS.join_with_fct f) appended x in
+    if Messages.tracing then Messages.tracel "simplesets-fct" "Join-fct result!\nx: %a\ny: %a\nconverted: %a\nres: %a\n" pretty x pretty y pretty (HS.of_list appended) pretty res;
+    res
```
sim642, line 232: Why is `join` not redefined via this?

src/cdomains/structDomain.ml
```diff
@@ -90,3 +105,496 @@ struct
...
+              |> List.filter (fun (ssx, ssy) -> maps_overlap ssx ssy)
+              |> List.map (fun (ssx, ssy) -> f ssx ssy)
+              |> HS.of_list
+      in reduce_key (s, take_some_key kx ky s)
+    end
+
+  let meet x y = meet_narrow_common x y SS.meet
+
+  let join (x, kx) (y, ky) = let s = HS.join x y in reduce_key (s, take_some_key kx ky s)
+  let leq (x, _) (y, _) = HS.leq x y
```
sim642, line 411: This is suspicious. How can the second component not matter for `leq` if it does for `join`? This violates a lattice property.

src/analyses/base.ml
```diff
@@ -1597,13 +1598,27 @@ struct
...
+         | Var var, o ->
+           (* For variables, this is done at to the level of entire variables to benefit e.g. from disjunctive struct domains *)
+           let oldv = get_var a gs st var in
+           let offs = convert_offset a gs st o in
+           let newv = VD.update_offset a oldv offs c' (Some exp) x (var.vtype) in
+           let v = VD.meet oldv newv in
+           if is_some_bot v then raise Deadcode
+           else (
+             if M.tracing then M.tracel "inv" "improve variable %a from %a to %a (c = %a, c' = %a)\n" d_varinfo var VD.pretty oldv VD.pretty v ID.pretty c VD.pretty c';
+             set' (Var var,NoOffset) v st
```
sim642, line 1611: Should this also use `o` instead of `NoOffset`? Or should it above just match `NoOffset` to only handle variables directly?

sim642 (commented):
I'm kind of bothered about the amount of copy-paste code from `HoareDomain` as well, because that adds two more Hoare domain implementations to the list in https://github.com/goblint/analyzer/issues/101#issuecomment-829334352.

src/analyses/base.ml
michael-schwarz: No. The idea here is to get the entire abstract value stored at this variable `get_var`.  Then this entire value is updated at offset o via update_offset and the result met with the old value. This is then set for the entire variable, thus `NoOffset`  should be the right thing here.

michael-schwarz:
All `x` could indeed be redefined via `let x = x_with_fct x` but I'm not sure if that makes the code nicer to understand overall.

michael-schwarz:
The copy-paste from the HoareDomain in some sense seems inevitable to me, unless we are willing to equip it with those `x_with_fct` functions (which are needed for structs that contain arrays), but feel like they are leaking abstraction.

Even if we are willing to do that, there will still be some code to handle the particular weirdness of these partitioned arrays, that we will not be able to abandon.

src/cdomains/structDomain.ml
michael-schwarz: It is deterministically the same for the same `compinfo`. So there should never be any joins with different keys.

michael-schwarz: The reason all of this crap is needed at all because all structs (regardless of their type) are represented as a map, where e.g. bot is `{}`

sim642 (approved):
I guess the partial duplication of Hoare domain can stay, otherwise this will never get merged.

For the cases where I pointed out `op` being different from `op_with_fct`, maybe just add comments to acknowledge this mismatch.

michael-schwarz:
I redefined all operations `x` in terms of `x_with_fct` now (where appropriate).

PR #421, opened 2021-10-31
Apron threshold widening

### LINKED ISSUE #420: Investigate widenings with thresholds for Apron
We should investigate if we can use `widening_threshold` (from http://apron.cri.ensmp.fr/library/0.9.10/mlapronidl/Abstract1.html), or `widening_thresholds` (from http://apron.cri.ensmp.fr/library/0.9.10/mlapronidl/Oct.html) where we use Apron Domains, to maybe be more precise on examples such as tests/regression/36-apron/90-mine14-5b.c, that currently work only if we replace join with widening.

Candidates might e.g. be all constant appearing the program.

This might also be needed to succeed on Mukherjee's examples that @ivanazuzic is extracting.

michael-schwarz:
This adds a simple threshold widening to our Apron analysis:

- Thresholds are all syntactic constants that appear in the program (and their value `*2` so that the octagon can keep constraints for `x<=C` which it represents as `x+x <= 2C`).
- It is controlled by means of an option `ana.apron.threshold_widening` which defaults to false now.
- Some programs from our related work require this option, so I added  it.

The following things remain TODO should be investigated (but probably not for the current deadline):

- My suspicion is that for larger programs using all constants from a program is too expensive. We want to be smarter here! (cue @ivanazuzic Machine Learning things maybe, or just using constants appearing in comparisons or some such thing)
- There is no reason that the base analysis should not also be able to make use of such thresholds: Any precision that is lost for globals once they are published can never be restored by narrowing.

Closes #420. :wink:

### DIFF
--- a/src/cdomains/apron/apronDomain.apron.ml
+++ b/src/cdomains/apron/apronDomain.apron.ml
@@ -19,7 +19,7 @@ let widening_thresholds_apron = lazy (
   (* Adding double value of all constants so that we can establish for single variables that they are <= const*)
   let t = List.append (!Cilfacade.widening_thresholds) (List.map (fun x-> 2*x) (!Cilfacade.widening_thresholds)) in
   let ts = List.sort_uniq compare t in
-  Array.of_list (List.map (fun x -> Printf.printf "one is: %i" x; Apron.Scalar.of_int x) ts)
+  Array.of_list (List.map (Apron.Scalar.of_int) ts)
 )
 
 module Var =
--- a/src/cdomains/apron/apronDomain.apron.ml
+++ b/src/cdomains/apron/apronDomain.apron.ml
@@ -16,10 +16,9 @@ module M = Messages
     - heterogeneous environments: https://link.springer.com/chapter/10.1007%2F978-3-030-17184-1_26 (Section 4.1) *)
 
 let widening_thresholds_apron = lazy (
-  (* Adding double value of all constants so that we can establish for single variables that they are <= const*)
-  let t = List.append (!Cilfacade.widening_thresholds) (List.map (fun x-> 2*x) (!Cilfacade.widening_thresholds)) in
-  let ts = List.sort_uniq compare t in
-  Array.of_list (List.map (Apron.Scalar.of_int) ts)
+  let t = WideningThresholds.thresholds () in
+  let r = List.map (fun x -> Apron.Scalar.of_mpqf @@ Mpqf.of_string @@ Z.to_string x) t in
+  Array.of_list r
 )
 
 module Var =
--- a/src/cdomains/apron/apronDomain.apron.ml
+++ b/src/cdomains/apron/apronDomain.apron.ml
@@ -765,7 +765,7 @@ struct
     let y_env = A.env y in
     if Environment.equal x_env y_env then
       (* widen if env didn't increase *)
-      if GobConfig.get_bool "ana.apron.threshhold_widening" then
+      if GobConfig.get_bool "ana.apron.threshold_widening" then
         let ts = Lazy.force widening_thresholds_apron in
         let r = Oct.widening_thresholds Man.mgr (Abstract1.abstract0 x) (Abstract1.abstract0 y) ts in
         {x with abstract0 = r}
--- a/src/cdomains/apron/apronDomain.apron.ml
+++ b/src/cdomains/apron/apronDomain.apron.ml
@@ -15,6 +15,12 @@ module M = Messages
     - C API docs PDF (alternative mathematical descriptions): https://antoinemine.github.io/Apron/doc/api/c/apron.pdf
     - heterogeneous environments: https://link.springer.com/chapter/10.1007%2F978-3-030-17184-1_26 (Section 4.1) *)
 
+let widening_thresholds_apron = lazy (
+  (* Adding double value of all constants so that we can establish for single variables that they are <= const*)
+  let t = List.append (!Cilfacade.widening_thresholds) (List.map (fun x-> 2*x) (!Cilfacade.widening_thresholds)) in
+  let ts = List.sort_uniq compare t in
+  Array.of_list (List.map (fun x -> Printf.printf "one is: %i" x; Apron.Scalar.of_int x) ts)
+)
 
 module Var =
 struct
@@ -758,7 +764,13 @@ struct
     let x_env = A.env x in
     let y_env = A.env y in
     if Environment.equal x_env y_env then
-      A.widening Man.mgr x y (* widen if env didn't increase *)
+      (* widen if env didn't increase *)
+      if GobConfig.get_bool "ana.apron.threshhold_widening" then
+        let ts = Lazy.force widening_thresholds_apron in
+        let r = Oct.widening_thresholds Man.mgr (Abstract1.abstract0 x) (Abstract1.abstract0 y) ts in
+        {x with abstract0 = r}
+      else
+        A.widening Man.mgr x y
     else
       y (* env increased, just use joined value in y, assuming env doesn't increase infinitely *)
 
--- a/src/util/cilfacade.ml
+++ b/src/util/cilfacade.ml
@@ -43,7 +43,6 @@ let init () =
 let current_statement = ref dummyStmt
 let current_file = ref dummyFile
 let showtemps = ref false
-let widening_thresholds = ref []
 
 let parse fileName =
   Frontc.parse fileName ()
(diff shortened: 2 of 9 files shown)


michael-schwarz:
As expected, it is very far from practical to just enable it, one needs to be more selective about which constants ought to become thresholds.

threshold

src/util/cilfacade.ml
```diff
@@ -102,6 +122,7 @@ let createCFG (fileAST: file) =
   (* Since we want the output of justcil to compile, we do not run allBB visitor if justcil is enable, regardless of  *)
   (* exp.basic-blocks. This does not matter, as we will not run any analysis anyway, when justcil is enabled.         *)
   if not (get_bool "exp.basic-blocks") && not (get_bool "justcil") then end_basic_blocks fileAST;
+  if get_bool "ana.apron.threshold_widening" then compute_widening_thresholds fileAST;
```
sim642, line 125: All of this logic could be performed inside the Apron lazy value itself. Then it doesn't have to be spread into seemingly unrelated modules.

src/util/cilfacade.ml
```diff
@@ -77,10 +78,29 @@ class allBBVisitor = object (* puts every instruction into its own basic block *
   method! vtype _ = SkipChildren
 end
 
+class extractConstantsVisitor(list) = object
+  inherit nopCilVisitor
+
+  method! vexpr e =
+    match e with
+    | Const (CInt64(i,ik,_)) ->
+      (if (Int64.compare i (Int64.of_int min_int) >= 0) && (Int64.compare i (Int64.of_int max_int)) <= 0 then
```
sim642, line 87: What do these comparisons really guarantee? If inside `CInt64` the integer component already overflowed, it might have wrapped around into this supposedly valid range still and you couldn't detect it after the fact.

It wouldn't be any more difficult to just handle the string component case as well (and use bigints) and ensure no such problem can occur.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -15,6 +15,12 @@ module M = Messages
     - C API docs PDF (alternative mathematical descriptions): https://antoinemine.github.io/Apron/doc/api/c/apron.pdf
     - heterogeneous environments: https://link.springer.com/chapter/10.1007%2F978-3-030-17184-1_26 (Section 4.1) *)
 
+let widening_thresholds_apron = lazy (
+  (* Adding double value of all constants so that we can establish for single variables that they are <= const*)
+  let t = List.append (!Cilfacade.widening_thresholds) (List.map (fun x-> 2*x) (!Cilfacade.widening_thresholds)) in
```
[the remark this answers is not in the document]
michael-schwarz: > but that's not what Apron seems to do: it has interval constraints as x <= c still.

Interesting: It was needed for this example: tests/regression/36-apron/89-mine14.c. Without the doubling, it required adding an additional constant variable of value `100` to work, the guess being that it would maintain `x <= 100` via this constant. After adding the doubling, it now works. :thinking:

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -758,7 +764,13 @@ struct
     let x_env = A.env x in
     let y_env = A.env y in
     if Environment.equal x_env y_env then
-      A.widening Man.mgr x y (* widen if env didn't increase *)
+      (* widen if env didn't increase *)
+      if GobConfig.get_bool "ana.apron.threshold_widening" then
+        let ts = Lazy.force widening_thresholds_apron in
+        let r = Oct.widening_thresholds Man.mgr (Abstract1.abstract0 x) (Abstract1.abstract0 y) ts in
```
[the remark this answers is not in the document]
michael-schwarz: Yes, I guess there one would have to use 

~~~OCaml
val widening_threshold : 'a Manager.t -> 'a t -> 'a t -> Lincons1.earray -> 'a t
~~~

from `Abstract1` (https://antoinemine.github.io/Apron/doc/api/ocaml/Abstract1.html), but I don't really know how to specify these constants as a `Lincons1.earray`...

src/util/cilfacade.ml
michael-schwarz: The reasoning might be that threshold widening could (should?) be a general concept, that we may later wish to apply also to our intervals...

src/util/cilfacade.ml
michael-schwarz: That's true, they don't guarantee much here. The reason I did not go with bigints here was that Apron in the end will only accept `int` anyway, but I guess you're right and one should use BigInts here.

src/util/cilfacade.ml
sim642: Sure, but then base intervals could also force the same lazy thing if they ever want it. It'd just have to be moved our of apron analysis for that to make more sense.

It's just cleaner if `Cilfacade` doesn't have to be cluttered with special-purpose things that all the analyses require and have to conditionally perform them in the right place during `createCFG`, which doesn't fit that function's purpose.

src/util/cilfacade.ml
sim642: > The reason I did not go with bigints here was that Apron in the end will only accept `int` anyway

It accepts bigints through its own GMP interface just fine, just not the Zarith ones:
https://github.com/goblint/analyzer/blob/791b561ea789f06b084cbb752b3349e38f5cd3a3/src/cdomains/apron/apronDomain.apron.ml#L105-L110

If this mechanism is to be general, the general list (or really a set), should be of our Zarith bigints anyway and apron will have to do the conversion to use them.

src/cdomains/apron/apronDomain.apron.ml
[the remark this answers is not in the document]
sim642: [The C API docs are more descriptive](https://antoinemine.github.io/Apron/doc/api/c/Abstract-values-and-operations-of-level-1.html#index-ap_005fabstract1_005fwidening_005fthreshold):

> Intersect the result of the standard widening with all the constraints in array that are satisfied by both a1 and a2.

Sounds like one would have to convert the numeric thresholds into constraints like `x <= 100`. But then for all variables and other left-hand sides? That sounds fairly expensive. And it essentially is what `strengthening` would do (although for joins).

Not sure if `Oct.widening_thresholds` somehow delegates to this and therefore is so slow, or they're completely separate mechanisms and the `Oct` version should be more optimal.

src/cdomains/apron/apronDomain.apron.ml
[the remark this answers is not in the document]
sim642: Could also be that the apron constraint output just simplifies `x + x <= 2c` to `x <= c` for output, while internally still having the former and also doing the widening with that doubled constant. There's no mention of that anywhere though.

michael-schwarz:
> An optimization that needn't be done here could be that the thresholds are collected and used per function.

Yes. the other thing I was thinking if one maybe only uses such constants that appear as an operator to a comparison. Of course then one would lose information for special constants etc, but that might make it practical.

So in the run to generate these numbers I had some code in that wrote out the number of thresholds: Sometimes it is about 200 thresholds, so I guess one shouldn't be surprised that that is horribly slow.
[index.zip](https://github.com/goblint/analyzer/files/7451832/index.zip)

sim642:
I looked at the implementation of `Oct.widening_thresholds`: https://github.com/antoinemine/apron/blob/28e9c36fe6f64c2365b2535ca28f55673d7858e9/octagons/oct_nary.c#L229-L248.
A lot of low-level detail but it looks like it will check all given thresholds on each variable (or each octagon left-hand side pair?), so it's linear in the number of given thresholds. No surprise it's expensive then.

Also this looks like the general widening threshold mechanism then, but without having to construct the intermediate lincons objects unnecessarily. 
If those `r->dim` things are actually all the octagon left-hand side pairs, then indeed the conjecture about doubling the thresholds makes sense, because there doesn't seem to be special logic for the interval case. But then why add the undoubled constraints at all?

michael-schwarz:
>  But then why add the undoubled constraints at all?

One might want to prove properties such as `x+y <= 100` and then you would want to undoubled ones.

---
One another note: There seems to be a ton of research on ideas how this might be done. I think it would make a great Bachelor's topic: Investigate what sort of things are proposed in literature and then implement them in Goblint, and see how they do. As a jumping off point one could for example give this paper:  https://link.springer.com/chapter/10.1007%2F978-3-319-47958-3_2 (a machine learning approach to this (CC @ivanazuzic :wink:) which also cites a lot of heuristics based ones).

michael-schwarz:
> A lot of low-level detail but it looks like it will check all given thresholds on each variable (or each octagon left-hand side pair?), so it's linear in the number of given thresholds. No surprise it's expensive then.

I don't think it even is the operation itself that is expensive here, but more the insane number of iterations needed to converge if I have 200 thresholds in between, which may even lead to additional contexts:

e.g. without thresholds `vars = 1533    evals = 5845`, and with them `vars = 10060    evals = 54624 `

src/util/cilfacade.ml
michael-schwarz: I moved it into a separate module now and made it more generic

PR #424, opened 2021-11-02
Fix apron mutex-meet-tid-cluster decreasing protecting locksets

sim642:
This combines the two different approaches I tried in https://github.com/goblint/analyzer/pull/417#issuecomment-954708437.

The logic is to first just consider clusters from still fully protected locksets. If that would give bottom, then use another component which contains all clusters that have nonempty intersection with the protecting lockset. This allows importing contributions from larger lockset clusters to smaller lockset clusters. At a following unlock they would be published to the smaller lockset clusters, so next time this backup shouldn't be necessary.

Somehow it's also possible that both give bottom, in that case `raise Deadcode` seems to work. Probably something else needs to side effect before to these clusters and destabilize this point again for actual recomputation.

This seems to make all the existing clusterings work: none give exceptions any more.

It also might make things somewhat more expensive since often the strong and weak mapping contain the same things, but the domain cannot be lazy.

### DIFF
--- a/src/analyses/apron/apronPriv.apron.ml
+++ b/src/analyses/apron/apronPriv.apron.ml
@@ -488,7 +488,7 @@ sig
   val keep_only_protected_globals: Q.ask -> LockDomain.Addr.t -> LAD.t -> LAD.t
   val keep_global: varinfo -> LAD.t -> LAD.t
 
-  val lock: LAD.t -> LAD.t -> AD.t
+  val lock: AD.t -> LAD.t -> LAD.t -> AD.t
   val unlock: W.t -> AD.t -> LAD.t
 end
 
@@ -505,8 +505,8 @@ struct
     let g_var = V.global g in
     AD.keep_vars oct [g_var]
 
-  let lock local_m get_m =
-    AD.join local_m get_m
+  let lock oct local_m get_m =
+    AD.meet oct (AD.join local_m get_m)
 
   let unlock w oct_side =
     oct_side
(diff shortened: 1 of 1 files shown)


sim642:
I extracted a separate functor for clusterings which are downward-closed. In that case it's possible to be more efficient by not tracing the secondary data structure.

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -573,7 +581,70 @@ struct
...
+
+  let keep_only_protected_globals ask m octs =
+    (* normal (strong) mapping: contains only still fully protected *)
+    let octs =
+      (* must filter by protection to avoid later meeting with non-protecting *)
+      LAD.filter (fun gs _ ->
+          VS.for_all (is_protected_by ask m) gs (* TODO: is this subset check right? *)
+        ) octs
+    in
+    LAD.map (keep_only_protected_globals ask m) octs (* TODO: is this even necessary if keys are filtered above? *)
```
michael-schwarz, line 594: It should not be necessary here. Are there any tests that fail if we remove this?

michael-schwarz (approved):

src/analyses/apron/apronPriv.apron.ml
sim642: No, I removed it now. It was left around from all my experiments with also trying just non-empty intersections (which need to be filtered), but that happens separately anyway.

PR #429, opened 2021-11-03
Mukherjee benchmarks

ivanazuzic:
These benchmarks have been added according to Table 2 from paper [A Thread-Local Semantics and Efficient Static
Analyses for Race Free Programs](https://arxiv.org/pdf/2009.02775.pdf). They are converted to C from [this RATCOP tests](https://bitbucket.org/suvam/ratcop/src/master/src/svcomp15/).

### DIFF
--- a/tests/regression/36-apron/79-mukherjee_DoubleLock_P3.c
+++ b/tests/regression/36-apron/79-mukherjee_DoubleLock_P3.c
@@ -1,0 +1,37 @@
+#include <pthread.h>
+#include "assert.h"
+
+unsigned int count;
+
+pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
+
+void* T1_DLP3(void* arg){
+    pthread_mutex_lock(&lock);
+    count++;
+    count--;
+    pthread_mutex_unlock(&lock);
+
+    pthread_mutex_lock(&lock);
+    count--;
+    count++;
+    pthread_mutex_unlock(&lock);
+    return NULL;
+}
+
+void* T2_DLP3(void* arg){
+    pthread_mutex_lock(&lock);
+    assert(count >= -1);  
+    pthread_mutex_unlock(&lock);
+    return NULL;
+}
+
+int main(){
+    count = 0;
+    
+    pthread_t t1;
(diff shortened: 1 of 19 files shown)


sim642:
Wouldn't our bench repo be a better place for these?

ivanazuzic:
> Wouldn't our bench repo be a better place for these?

MAybe, but I'll leave them here because Michael sad he'll cherry-pick them to some other place.

michael-schwarz:
Bnechmarks is not really a good word for it. They're more regression tests than anything else. None of them are longer than 100 LoC.

tests/regression/36-apron/79-mukherjee_DoubleLock_P3.c
```diff
@@ -0,0 +1,40 @@
+// SKIP
```
michael-schwarz, line 1: Te assert fails when running with GCC.
Are you sure it was `unsigned int count` @ivanazuzic?

sim642:
> Bnechmarks is not really a good word for it. They're more regression tests than anything else. None of them are longer than 100 LoC.

Maybe a separate group for them at least to keep this set separate from our own stuff?

michael-schwarz:
Separate group on top of `use_tids` now.
@ivanazuzic for the problems with the benchmarks that you identified, can you make a PR on top of `use_tids`?

michael-schwarz:
`52/05` must be wrong, the assert there is never reachable?

michael-schwarz:
`52/12` asserts that something does hold and simultaneously does not hold?

michael-schwarz:
`52/06`, `52/09`, `52/10` Mukherjee can also not do according to the table in his own paper.

michael-schwarz:
After annotating according to the table of what Mukherjee can do, we have three tests left where we are worse:

- `"52/03 mukherjee_sssc12", "52/05 mukherjee_simpleLoop"` we report the asserts are unreachable, maybe something is wrong with the benchmark?
- `"52/17 mukherjee_sync01"` has Undefined Behavior

michael-schwarz:
Added on top of `use_tids`.

PR #431, opened 2021-11-04
Fix symb_locks combine to use returned lockset

### LINKED ISSUE #430: Symbolic locks analysis doesn't handle lock/unlock wrapper functions
Hi. Thanks for the great analyzer! I'm working on a research project translating C code with `pthread_mutex` to Rust code with `Mutex`, which couples a mutex and data to protect. For this purpose, I need to find which data are protected by each mutex, and this analyzer seems to fit this goal.

While experimenting with the analyzer, I found a situation in which it loses its precision. Some projects define their own lock/unlock functions calling `pthread_mutex_lock`/`pthread_mutex_unlock`. For example, the following code defines `lock` and `unlock`.

```c
typedef struct { int myint; pthread_mutex_t mymutex; } mystruct;

void lock(mystruct *s) { pthread_mutex_lock(&s->mymutex); }
void unlock(mystruct *s) { pthread_mutex_unlock(&s->mymutex); }

void *foo(void *arg) {
  mystruct *s = (mystruct *) arg;

  lock(s);
  s->myint=s->myint+1;
  unlock(s);

  return NULL;
}

int main() {
  mystruct *s = (mystruct *) malloc(sizeof(*s));
  pthread_mutex_init(&s->mymutex, NULL);

  pthread_t id1, id2;
  pthread_create(&id1, NULL, foo, s);
  pthread_create(&id2, NULL, foo, s);
  pthread_join(id1, NULL);
  pthread_join(id2, NULL);
}
```

When analyzing this code, Goblint reports a data race:

```
[Warning][Race] Memory location (alloc@sid:20).myint@:-1:-1 (race with conf. 110):
  read with {thread:([foo@a.c:23:3, main], {})} (conf. 110)  (exp: & s->myint) (a.c:12:3)
  read with {thread:([foo@a.c:24:3, main], {})} (conf. 110)  (exp: & s->myint) (a.c:12:3)
  write with {thread:([foo@a.c:23:3, main], {})} (conf. 110)  (exp: & s->myint) (a.c:12:3)
  write with {thread:([foo@a.c:24:3, main], {})} (conf. 110)  (exp: & s->myint) (a.c:12:3)
```

However, when `foo` calls `pthread_mutex_lock`/`pthread_mutex_unlock` instead, every access is considered safe as expected.

```c
void *foo(void *arg) {
  mystruct *s = (mystruct *) arg;

  pthread_mutex_lock(&s->mymutex);
  s->myint=s->myint+1;
  pthread_mutex_unlock(&s->mymutex);

  return NULL;
}
```

```
[Success][Race] Memory location (alloc@sid:20).myint@:-1:-1 (safe):
  read with {p-lock:*.mymutex, thread:([foo@a.c:23:3, main], {})} (conf. 110)  (exp: & s->myint) (a.c:12:3)
  read with {p-lock:*.mymutex, thread:([foo@a.c:24:3, main], {})} (conf. 110)  (exp: & s->myint) (a.c:12:3)
  write with {p-lock:*.mymutex, thread:([foo@a.c:23:3, main], {})} (conf. 110)  (exp: & s->myint) (a.c:12:3)
  write with {p-lock:*.mymutex, thread:([foo@a.c:24:3, main], {})} (conf. 110)  (exp: & s->myint) (a.c:12:3)
```

Is there an option to prevent this imprecision? Currently, I'm using the following configuration:

```json
{
  "ana": {
    "int": {
      "def_exc": true,
      "interval": false,
      "enums": true
    },
    "activated": [
      "expRelation",
      "base",
      "threadid",
      "threadflag",
      "threadreturn",
      "escape",
      "mutex",
      "mallocWrapper",
      "var_eq",
      "symb_locks",
      "region",
      "thread"
    ],
    "context": {
      "widen": true
    }
  },
  "exp": {
    "privatization": "protection"
  }
}
```
I expect I'll ask other questions in the near future. Please let me know if this issue tracker isn't a proper place to ask questions like this.

Thank you!

sim642:
Closes #430.

For some reason `combine` used the caller's local state instead of the callee's returned one. I don't know the internals of the symbolic locks analysis, so I don't know if this possibly causes some issues.

### DIFF
--- a/src/analyses/symbLocks.ml
+++ b/src/analyses/symbLocks.ml
@@ -44,7 +44,7 @@ struct
     List.fold_right D.remove_var (fundec.sformals@fundec.slocals) ctx.local
 
   let enter ctx lval f args = [(ctx.local,ctx.local)]
-  let combine ctx lval fexp f args fc st2 = ctx.local
+  let combine ctx lval fexp f args fc st2 = st2
 
   let get_locks e st =
     let add_perel x xs =
--- a/src/cdomains/lockDomain.ml
+++ b/src/cdomains/lockDomain.ml
@@ -140,7 +140,10 @@ struct
     let no_casts = S.map Expcompare.stripCastsDeepForPtrArith (eq_set ask e) in
     let addrs = S.filter (function AddrOf _ -> true | _ -> false) no_casts in
     S.union addrs st
-  let remove ask e st = S.diff st (eq_set ask e)
+  let remove ask e st = 
+    let no_casts = S.map Expcompare.stripCastsDeepForPtrArith (eq_set ask e) in
+    let addrs = S.filter (function AddrOf _ -> true | _ -> false) no_casts in
+    S.diff st addrs
   let remove_var v st = S.filter (fun x -> not (Exp.contains_var v x)) st
 
   let kill_lval (host,offset) st =
--- a/src/cdomains/lockDomain.ml
+++ b/src/cdomains/lockDomain.ml
@@ -141,6 +141,7 @@ struct
     let addrs = S.filter (function AddrOf _ -> true | _ -> false) no_casts in
     S.union addrs st
   let remove ask e st = 
+    (* TODO: Removing based on must-equality sets is not sound! *)
     let no_casts = S.map Expcompare.stripCastsDeepForPtrArith (eq_set ask e) in
     let addrs = S.filter (function AddrOf _ -> true | _ -> false) no_casts in
     S.diff st addrs
--- a/tests/regression/06-symbeq/26-symb_lockfuns.c
+++ b/tests/regression/06-symbeq/26-symb_lockfuns.c
@@ -1,0 +1,28 @@
+// PARAM: --disable ana.mutex.disjoint_types --set ana.activated[+] "'var_eq'"  --set ana.activated[+] "'symb_locks'"
+#include <pthread.h>
+
+typedef struct { int myint; pthread_mutex_t mymutex; } mystruct;
+
+void lock(mystruct *s) { pthread_mutex_lock(&s->mymutex); }
+void unlock(mystruct *s) { pthread_mutex_unlock(&s->mymutex); }
+
+void *foo(void *arg) {
+  mystruct *s = (mystruct *) arg;
+
+  lock(s);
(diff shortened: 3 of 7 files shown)


vesalvojdani:
Symbolic locksets should hold the set of addresses that are definitely equal to definitely locked addresses. They should follow the same semantics as address must-equality analysis. The concern would be with recursive calls. It's possible that returning the second component would result in flattening of all recursive calls and having some invalid equalities. We can discuss later today.

vesalvojdani:
I looked into concerns with recursion, but there are more serious issues with the symbolic lockset. I'd say this change should be made. Simmo pointed out that otherwise unlocks within functions will be ignored. I wanted to create a regression test that would expose that flaw in master, but it turns out we have more serious unsoundness issues with unlocks. I'll create a separate ticket for unsound unlocks. 

I'm not 100% certain that the current thing does the right thing with recursion, but I can't think of any way to fool it. Since it is a must set, any attempted change invalidates it. Here's an example of the opposite effect where master is more precise:

```c
// PARAM: --disable ana.mutex.disjoint_types --set ana.activated[+] "'var_eq'"  --set ana.activated[+] "'symb_locks'"
#include <stdlib.h>
#include <pthread.h>

typedef struct { int myint; pthread_mutex_t mymutex; } mystruct;

void *foo(void *arg) {
  if (__VERIFIER_nondet_int()) return NULL;

  mystruct *s = (mystruct *) malloc(sizeof(*s));
  pthread_mutex_init(&s->mymutex, NULL);

  pthread_mutex_lock(&s->mymutex);
  foo(NULL);
  s->myint = 0; // NORACE
  pthread_mutex_unlock(&s->mymutex);

  return NULL;
}

int main() {
  pthread_t id1, id2;
  pthread_create(&id1, NULL, foo, NULL);
  pthread_create(&id2, NULL, foo, NULL);
  pthread_join(id1, NULL);
  pthread_join(id2, NULL);
}
```

I didn't want to check this example in, but I will check in the unsound unlock cases even if we would not fix them here.

vesalvojdani (approved):
I think we can merge this now. I created ticket #434 for the deeper issues. Tests pass on my machine. I'm not sure why the github actions are failing...

PR #433, opened 2021-11-04
Add option to store local results of Apron-Analysis for comparison

### LINKED ISSUE #423: Allow serialization and deserialization of Apron Analysis results for precision comparison
The serialization of results of the Apron analysis currently fails with execption: `Fatal error: exception Invalid_argument("output_value: abstract value (Custom)")`. 

This needs to be fixed ~~for the incremental analysis and~~ for comparing different runs with the Apron analysis. 

Edit: For now we only need to focus on allowing a comparison.

jerhard:
This PR adds an option to store apron analysis results in a file, such that different analysis runs can be compared using the `apronPrivPrecCompare` program that is introduced and based on `privPrecCompare`. 

Storing the results of the Apron analysis is activated by setting the string option for `"exp.apron-prec-dump"` to a non-empty string, which will be used as the file name.

Apron does not offer serialization for its Intervals ("Boxes") and Polyhedra ("Polka"). 
We thus convert the Apron domain used in the analysis to the octagon domain before serialization.  This is done via a conjunction of linear constraints(see [here](https://antoinemine.github.io/Apron/doc/api/ocaml/Lincons0.html) and [here](https://antoinemine.github.io/Apron/doc/api/ocaml/Abstract1.html), i.e. with `to_lincons_array` and `of_lincons_array`).

### What is stored / compared
The results for the analysis are stored per program node. Results for different contexts but the same node are joined together. Results of global unknowns are not stored.  

### What still needs to be done
Storing the results (done in `finalize`) should only be done in the `postsolving` stage. 
However, when adding such a check, the code is not called. Maybe one needs to change some configuration?

### How to test
The comparison program ~~`apronPrivPrecCompare`~~ `apronPrecCompare` is built with:
```
make apronPrecCompare 
```
The comparison then can be tested on a simple example as follows:
```
./regtest.sh 36 71 --set ana.apron.domain "octagon" --set exp.apron.prec-dump apron_oct.save
./regtest.sh 36 71 --set ana.apron.domain "interval" --set exp.apron.prec-dump apron_int.save

./apronPrecCompare apron_oct.save apron_int.save
```

Closes #423


sim642:
> Storing the results (done in `finalize`) should only be done in the `postsolving` stage.
> However, when adding such a check, the code is not called. Maybe one needs to change some configuration?

`finalize` is called at the very end of `Control` anyway and nowhere else (AFAIK), so there doesn't need to be any condition in there. "Postsolving" only refers to the single final evaluation of the entire constraint system and the boolean is actually disabled after that.

sim642:
Currently the CI fails for all non-apron builds because some util module depends on the Apron stuff. That probably should be separated.

src/apronPrivPrecCompare.ml
```diff
@@ -0,0 +1 @@
+module B = PrePrivPrecCompare.ComparePrec (PrivPrecCompareUtil.ApronUtil)
```
sim642, line 1: It's kind of weird that just applying this functor will already do something imperative. I think there should be an actual function to be explicitly called from these, e.g. `B.main ()`.

jerhard:
>Currently the CI fails for all non-apron builds because some util module depends on the Apron stuff. That probably should be separated.

The utility module for the apron precision comparison is now moved to a separate module using an optional apron dependency, so that should be fixed.

.gitignore
```diff
@@ -58,6 +58,7 @@ goblint.bc.js
 sv-comp/goblint.zip
 
 privPrecCompare*
+apronPrivPrecCompare
```
sim642, line 61:
```suggestion
apronPrecCompare
```

src/util/defaults.ml
```diff
@@ -178,6 +178,7 @@ let _ = ()
       (* TODO: priv subobject *)
       ; reg Experimental "exp.privatization"     "'protection-read'" "Which privatization to use? none/protection-old/mutex-oplus/mutex-meet/protection/protection-read/protection-vesal/mine/mine-nothread/mine-W/mine-W-noinit/lock/write/write+lock"
       ; reg Experimental "exp.priv-prec-dump"    "''"    "File to dump privatization precision data to."
+      ; reg Experimental "exp.apron-prec-dump"    "''"    "File to dump apron precision data to."
```
sim642, line 181: Maybe group it under `exp.apron` as `exp.apron.prec-dump`?

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -19,6 +20,10 @@ struct
 
   open AD
 
+  open ApronPrecCompareUtil.Util
```
sim642, line 23: This `Util.Util` is awkward. Why not just include its contents directly at the top level of the module file?

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -378,12 +383,45 @@ struct
       st
 
   let sync ctx reason =
+    (* After the solver is finished, store the results (for later comparison) *)
+    if !GU.postsolving then begin
+      let old_value = match RH.find_option results ctx.node with
+        | Some v -> v
+        | None -> AD.bot ()
+      in
```
sim642, line 391: This is a bit verbose for `find_default`.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -75,9 +75,15 @@ let priv_manager =
        "polyhedra", (module PolyhedraManager: Manager)]
     in
     let domain = (GobConfig.get_string "ana.apron.domain") in
-    match List.assoc_opt domain options with
-    | Some man -> man
-    | None -> failwith @@ "Apron domain " ^ domain ^ " is not supported. Please check the ana.apron.domain setting."
+    let man = (match List.assoc_opt domain options with
+        | Some man -> man
+        | None -> failwith @@ "Apron domain " ^ domain ^ " is not supported. Please check the ana.apron.domain setting.")
```
sim642, line 80: The parenthesis in these right-hand sides are unnecessary.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -64,7 +64,7 @@ struct
   type mt = Box.t
   type t = mt Manager.t
   let mgr = Box.manager_alloc ()
-  let name = "Interval Manager"
+  let name () = "Interval"
 end
 
 let priv_manager =
```
sim642, line 70: I guess this is a leftover from #418, but it shouldn't be called "priv" anymore.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -75,9 +75,15 @@ let priv_manager =
...
-    | Some man -> man
-    | None -> failwith @@ "Apron domain " ^ domain ^ " is not supported. Please check the ana.apron.domain setting."
+    let man = (match List.assoc_opt domain options with
+        | Some man -> man
+        | None -> failwith @@ "Apron domain " ^ domain ^ " is not supported. Please check the ana.apron.domain setting.")
+    in
+    (* Set the manager for deserialization (this step is required when apron results are loaded) *)
+    (* As of now, serialization/deserialization only works for octagons *)
+    let module Man = (val man: Manager) in
+    Manager.set_deserialize Man.mgr;
```
sim642, line 85: This looks almost like the manager is calling itself to set itself for deserialization, but the second `Manager` actually refers to `Apron.Manager` I guess?

Also, wouldn't it make more sense for `ApronAnalysis` to set this since it unwraps the module anyway, no? Maybe not the best style to have a `lazy` computation perform side effects like this that just getting the manager will quietly change some Apron global state.

src/privPrecCompare.ml
```diff
@@ -1,45 +1,6 @@
...
-  List.iter (fun {lvh; _} ->
-      LVH.iter (fun (l, x) _ ->
-          LH.replace locations l ();
-          LVH.replace location_vars (l, x) ()
-        ) lvh
-    ) dumps;
-  (LH.length locations, LVH.length location_vars)
+(* Utility program to compare precision of results of the base analysis with different privatizations *)
+(* The result files that can be compared here may be created by passing an output file name in "exp.priv-prec-dump" to Goblint *)
+module A = PrePrivPrecCompare.ComparePrec (PrivPrecCompareUtil.Util)
```
sim642, line 3: `Prec` and `Compare` are named twice in `PrePrivPrecCompare.ComparePrec`. Maybe just move and rename it to `PrecCompare.Make` to avoid the unnecessary duplication and also remove `Priv` since the Apron version isn't priv-specific anymore.

src/util/PrecCompareUtil.ml
```diff
@@ -0,0 +1,70 @@
...
+
+  (* Leave loaded value untouched by default *)
+  let unmarshal x = x
+end
+
+
+module Node =
+struct
+  include Node
+  let name () = "nodes"
```
sim642, line 68: Doesn't `node.ml` already define the name?

src/cdomains/apron/apronDomain.apron.ml
jerhard: Setting the deserialization manager is actually no longer needed, as no deserialization of Apron data structures is done in Goblint itself, but only in the `apronPrecCompare`. There the deserialization manager is already initialized.

jerhard: So I could remove the code in question entirely.

src/util/PrecCompareUtil.ml
jerhard: Yes, but it is overridden here to be the plural, so it fits better into the output of `prePrivPrecCompare`, defined here: https://github.com/goblint/analyzer/blob/f79dbc877b3738cf328d34671964079514d83fb2/src/prePrivPrecCompare.ml#L46

src/privPrecCompare.ml
jerhard: @sim642 There is already a module `PrecCompare.Make`. Should one just move that into  `PrecCompare.Comparison.Make`?

sim642: I think not, it doesn't completely make sense there because the `Comparison` module is just for that type.

Maybe `MakeDump` or something then, because the content of `ComparePrec` has to do with dumping and comparing the dumps.

jerhard: Ok, changed it.

sim642 (approved):

src/analyses/apron/apronAnalysis.apron.ml
jerhard: Changed it, so the `.Util` is no longer needed.

PR #438, opened 2021-11-08
Improve interval overflow detection

sim642:
This fixes some more points in #346:

1. Simplify `cast_to` to just `norm`, which already does all overflow handling. This means we also warn about overflows in casts. Not sure why it was so involved before.
2. Distinguish overflow warnings from underflow warnings. This is just to better match the CWEs, internally we call them all still overflows.


jerhard (approved):
Looks good!

PR #442, opened 2021-11-10
Add stubs for `qsort` and `bsearch`

### LINKED ISSUE #426: Add qsort implementation to headers
We identified an unsoundness for an SV-Comp benchmark involving a call to qsort, where the reachability of a program point within the passed comparator function was asked for, see https://github.com/goblint/analyzer/issues/382#issuecomment-949382913.

One idea to fix this we discussed today is providing an implementation of `qsort` from a standard library that could be used within the analysis of user code.

sim642:
Closes #426.

### DIFF
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -243,6 +243,8 @@ let preprocess_files () =
   if get_bool "custom_libc" then
     cFileNames := find_custom_include "lib.c" :: !cFileNames;
 
+  cFileNames := find_custom_include "stdlib.c" :: !cFileNames;
+
   if get_bool "ana.sv-comp.functions" then
     cFileNames := find_custom_include "sv-comp.c" :: !cFileNames;
 
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -239,10 +239,6 @@ let preprocess_files () =
     );
   );
 
-  (* possibly add our lib.c to the files *)
-  if get_bool "custom_libc" then
-    cFileNames := find_custom_include "lib.c" :: !cFileNames;
-
   cFileNames := find_custom_include "stdlib.c" :: !cFileNames;
 
   if get_bool "ana.sv-comp.functions" then
--- a/includes/lib.c
+++ b/includes/lib.c
@@ -1,4 +1,0 @@
-int ftw(const char *dir, int (*fn)(const char *file, const struct stat *sb, int flag), int nopenfd) {
-  fn("", 0, 0);
-  return 0;
-}
--- a/includes/stdlib.c
+++ b/includes/stdlib.c
@@ -1,0 +1,29 @@
+#include <stddef.h>
+
+void qsort(void *ptr, size_t count, size_t size, int (*comp)(const void*, const void*)) {
+  // call all possible compares first, before invalidating array elements
+  for (size_t i = 0; i < count; i++) {
+    for (size_t j = 0; j < count; j++) {
+      comp(ptr + i * size, ptr + j * size);
+    }
+  }
+
+  // randomly swap all possible, invalidates array elements
+  for (size_t i = 0; i < count; i++) {
+    for (size_t j = 0; j < count; j++) {
+      int r; // rand
+      if (r) {
+        // swap elements byte-by-byte, no other way to do it, because we cannot allocate and copy/swap abstract elements
+        for (size_t k = 0; k < size; k++) {
+          char *a = ptr + i * size + k;
+          char *b = ptr + j * size + k;
+          char c = *a;
+          *a = *b;
+          *b = c;
+        }
+      }
+    }
+  }
+
+  // array isn't actually sorted! just pretent calls for Goblint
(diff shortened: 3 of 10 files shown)


michael-schwarz (approved):
LGTM!

src/maingoblint.ml
```diff
@@ -243,6 +243,8 @@ let preprocess_files () =
   if get_bool "custom_libc" then
     cFileNames := find_custom_include "lib.c" :: !cFileNames;
 
+  cFileNames := find_custom_include "stdlib.c" :: !cFileNames;
```
sim642, line 246: This makes me wonder what the point of `custom_libc` really is. The only place were AFAIK it is used is `pfscan` in the bench repository to have an overly simple stub for `ftw` (which isn't even the C standard library, but some Unix one).

I didn't want to add `qsort` and `bsearch` in `lib.c` because `custom_libc` is disabled by default, but we should still be sound for `qsort` and `bsearch` even without enabling some custom option.

These now additions make the meaning of `custom_libc` confusing though, because this is another custom libc thing which actually doesn't obey the option.

sim642: `ftw` is another of those functions which uses a callback though, so maybe it should be just in `stdlib.c` always anyway. Or possibly another file `ftw.c` which is added by default to be more clear that it isn't C standard library.

sim642: Since it was just specific to pfscan, we decided to move its stub to the bench repository: https://github.com/goblint/bench/commit/c31ef8287e2b77cdbe0c2ba577e29b0bfa7cb4d0. Now `custom_libc` option is removed here and the stdlib stubs are always used.

michael-schwarz:
This now leads to other unsoundness problems, e.g. for the following program from sv-comp:

~~~C

int bsearch(int i, int j)
{
  if (i>=j) return i;
  int mid = (i+j)/2;
  if (__VERIFIER_nondet_int())
    return bsearch(i,mid);
  return bsearch(mid+1,j);
}


int main() {
  int x = __VERIFIER_nondet_int();
  int y = __VERIFIER_nondet_int();
  bsearch(x, y);
}
~~~

Here, we do not warn about the potential overflow of `i+j`.

```
michael@michael-ThinkPad-X1-Carbon-6th:~/Documents/goblint-cil/analyzer$ ./goblint 1.c
[Info][Imprecise] Invalidating expressions: AddrOf(Var(tmp, NoOffset)) (1.c:13:3-13:17)
[Info][Imprecise] Invalidating expressions: AddrOf(Var(tmp___0, NoOffset)) (1.c:14:3-14:17)
[Error][Imprecise][Unsound] Function definition missing for bsearch (1.c:15:3-15:16)
[Info][Imprecise] INVALIDATING ALL GLOBALS! (1.c:15:3-15:16)
[Info][Imprecise] Invalidating expressions: Lval(Var(x, NoOffset)), Lval(Var(y, NoOffset)) (1.c:15:3-15:16)
1.c:2: Warning: Incompatible declaration for bsearch (from /home/michael/Documents/goblint-cil/analyzer/goblint_temp_51932e/1.c(1)).
 Previous was at /home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c:33 (from /home/michael/Documents/goblint-cil/analyzer/goblint_temp_51932e/stdlib.c (0)) (different type constructors: void * vs. int ) 
1.c:2: Warning: def'n of func bsearch at 1.c:2 (sum 49803) conflicts with the one at /home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c:34 (sum 1086600); keeping the one at /home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c:34.
```

michael-schwarz:
Maybe we should only include our function body if the appropriate header is included?

sim642:
> Maybe we should only include our function body if the appropriate header is included?

It's not so obvious how to achieve this because in CIL we don't know about includes any more. So we would have to grep the input files directly for such includes? And what about transitive includes?
Moreover, it's not strictly necessary to include a standard library header to call those functions: if you declare them (with the correct signature) as `extern`, then it will also link correctly without ever knowing where the `extern` declaration came from (it was just included by the preprocessor).

I think this should somehow be solved in CIL's merging process, which currently warns and then forgets _both_ definitions. The same issue is being solved by GCC when it links the program: it also has two definitions of the function, one in this file and one in its standard library.
It shouldn't be too hard to make it keep the first definition it sees and ignore the second one (and then we make sure to add our stubs to the merger last).

sim642:
Actually, I doubt that program is even valid, because:

> The names of all library types, macros, variables and functions that come from the ISO C standard are reserved unconditionally; your program **may not** redefine these names.

(http://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html)

sim642:
I just opened a MR to sv-benchmarks regarding this: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1358. We'll see how that goes.

PR #443, opened 2021-11-10
TD3: optimize switch to narrow by reusing computed right-hand side

sim642:
When `widen` leaves the value unchanged and stable, and we switch to `narrow` phase, then the first `eq` call unnecessarily recomputes the same unchanged right-hand side. This optimization passes the last `eq` value (prior to `widen`) to the new `solve` to reuse it immediately for the corresponding `narrow`.

### TODO
- [x] Benchmark to see if makes any measurable difference.

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -112,7 +112,7 @@ module WP =
             HM.remove stable y;
             HM.mem called y || destabilize_vs y || b || was_stable && List.mem y vs
           ) w false
-      and solve x phase =
+      and solve ?reuse_eq x phase =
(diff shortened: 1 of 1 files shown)


sim642:
I ran some before (x-axis) vs after (y-axis) benchmarks on a subset of sv-benchmarks (SoftwareSystems, ConcurrencySafety) and here's the runtime comparison:
image

Slightly under the identity line, so there's a bit of improvement on the large end: some benchmarks improved by 1-2s. It's more than I expected because it's just one extra reevaluation (with stable dependencies?) on local wpoints.

I guess we can go for this because it's a relatively simple code change.

PR #445, opened 2021-11-11
Add option to join with interval information only in Apron-privatizations.

### LINKED ISSUE #435: Benchmarking: Add Apron-Analysis option for relational local states and non-relational globals
## Motivation
We want to benchmark the impact of having relational information for global variables. 

## Background
With #433, we can compare runs of the Apron analysis with different Apron domains, e.g. with octagons and intervals ("Boxes"). 
However, an analysis with octagons will in many cases already be more precise than an analysis with intervals, no matter whether relational information for global variables is found out or not in the analysis using octagons. This is simply because the octagons will yield more precise results for local computations. 

## Goal
There should be a mode where only interval information is kept about globals, while local states contain relational information. 
This could then be compared to a setting where both globals and local states have relational information, to check whether the relational information we obtain about globals actually yields any benefit in precision.

## Remark
This feature is primarily intended for benchmarking, not actual usage.

jerhard:
This PR adds the option `exp.apron.priv.only-interval`. When it is enabled, joins on the Apron-domain in the Apron-privatizations are performed by projecting the operands to intervals, joining those, and converting them back to the Apron-domain used in the analysis. This should sufficiently destroy relational information that is not already implied by the interval information:
All side-effects are joined into the global invariant, so every information that arrives there must have gone through the join.

Closes #435


src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -728,6 +775,27 @@ struct
...
+      let y_env = A.env y in
+      let c_env = gce x_env y_env in
+      let x_c = A.change_environment IntervalManager.mgr x c_env false in
+      let y_c = A.change_environment IntervalManager.mgr y c_env false in
+      let join_c = A.join IntervalManager.mgr x_c y_c in
+      let j_env = Environment.lce x_env y_env in
+      let r = A.change_environment IntervalManager.mgr join_c j_env false in
+      r
+    ) in
+    of_interval r
```
sim642, line 797: Why does it repeat all the `join_interval` logic from `DLift` instead of just delegating to that `join_interval` in the middle (like `join` does on common environment)?

sim642 (commented):
Since the intervalization is only done on `join`, couldn't this still leave relational information in globals if that global is never joined into again?

src/cdomains/apron/apronDomain.apron.ml
jerhard: Changed it, you may look at it again.

jerhard:
>Since the intervalization is only done on join, couldn't this still leave relational information in globals if that global is never joined into again?

At least in td3, the first side effect to an unknown is joined with bottom. In this case, the `join_interval` defined here will map the non-bottom operand to the interval representation and back. So all relational information that survives this transformation must already be implied by the interval representation of the operand, no?

Probably I still have to fix the handling of the `l` set in the `PerMutexPrivTID` analysis, so it is also ensured that no relational (that is not already implied by intervals) is kept there.

sim642:
> At least in `td3` and the `topDown` solver, the first side effect to an unknown is joined with bottom. In this case, the `join_interval` defined here will map the non-bottom operand to the interval representation and back. So all relational information that survives this transformation must already be implied by the interval representation of the operand, no?
> 
> Probably I still have to fix the handling of the `l` set in the `PerMutexPrivTID` analysis, so it is also ensured that no relational (that is not already implied by intervals) is kept there.

Feels quite _ad hoc_ though. A cleaner approach might be to use a different domain (with interval manager) inside a priv module, instead of using the outside provided one. And then just apply the conversion functions whereever relations need to be passed between the global and local states.

But given that this is just for some benchmarking comparison purposes, I don't know if it's worth the effort to do any cleaner.

jerhard:
>Feels quite ad hoc though. A cleaner approach might be to use a different domain (with interval manager) inside a priv module, instead of using the outside provided one. And then just apply the conversion functions whereever relations need to be passed between the global and local states.

Yes, the approach with a separate domain for the privatization would certainly be cleaner. 
Doing it as it is done now seemed easier to do, and I think we could (should) revert these changes at some point.

jerhard:
>Probably I still have to fix the handling of the l set in the PerMutexPrivTID analysis, so it is also ensured that no relational (that is not already implied by intervals) is kept there.

Now abstract values that are added to the `l`-map are first projected to intervals and back to the abstract domain used in the analysis.

sim642:
> Doing it as it is done now seemed easier to do, and I think we could (should) revert these changes at some point.

Probably, yes.

Also, all the apron CI jobs are failing now with some exception.

jerhard:
@sim642 I tried implementing using different domains for the global and local state in #447 and it seems to be easier than I initially thought.

sim642:
Closing for now since this wasn't used for the evaluation and there's no need to keep this around permanently.

PR #446, opened 2021-11-11
Fix signed conversion overflow unsoundness with assume_none

sim642:
After #438 I ran sv-benchmarks locally and noticed a handful of unsound cases where we haven't been unsound before. The more precise cast handling of intervals by just `norm` is problematic with `sem.int.signed_overflow` being `assume_none` (which is what we do for SV-COMP).

Namely, the conversion from `unsigned long` to `long` (which might not be entirely contained) is not signed overflow per se, but rather a conversion, where this is not undefined behavior, but implementation-defined behavior. So `assume_none` cannot soundly exclude that possibility. This is the case for some sv-benchmarks where GCC's implementation-defined behavior (wraparound) is still valid.

Also fixes a bug from #225, where all return values got cast to `void*`, not just the thread return ones.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1826,12 +1826,14 @@ struct
           | TVoid _ -> M.warn "Returning a value from a void function"; assert false
           | ret -> ret
         in
-        (* Evaluate exp and cast the resulting value to the void-pointer-type.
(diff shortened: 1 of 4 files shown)


jerhard (approved):
Looks good!

PR #447, opened 2021-11-12
Allow to use apron-interval domain for globals independent of apron domain used for locals

### LINKED ISSUE #435: Benchmarking: Add Apron-Analysis option for relational local states and non-relational globals

jerhard:
This PR adds the option to actually use the apron interval domain for the privatizations while the local state is abstracted with a possibly other domain.
This is done by adding a `Converter`-Functor for the apron-privatizations. It takes a `ApronDomain` which should be used for the global state and a privatization as arguments. the Privatization is then instantiated using this global apron domain. 
The converter does the conversion between local and global domains necessary for satisfying the interface of an Apron privatization used with the local domain. The conversion is done via linear_constraints, i.e. `to_lincons_array` and `of_lincons_array` of [Abstract1](https://antoinemine.github.io/Apron/doc/api/ocaml/Abstract1.html). 

This is hopefully a less hacky alternative to #445
Fixes #435


src/analyses/apron/apronPriv.apron.ml
```diff
@@ -43,6 +42,74 @@ module type S =
...
+
+  let should_join a b = Priv.should_join (to_global_components a) (to_global_components b)
+
+  let read_global ask getg st g x =
+    let r = Priv.read_global ask getg (to_global_components st) g x in
+    to_local r
+
+  let write_global ?(invariant=false) ask getg sideg st g x =
+    let r = Priv.write_global ask getg sideg (to_global_components st) g x in
+    to_local_components r
```
sim642, line 79: If the local state `st` contains some non-interval constraints, then converting it to interval manager will lose those and converting from intervals back to octagons won't restore them. I think the same applies to all the other transfer functions too. Or am I misunderstanding this converter?

I think one way to make it right is to give each `Priv` module two `AD` arguments for the local (inside components) and globals (inside global invariants). And then have each of them do the conversions at appropriate places when they're transferring information from one to the other.

A simpler approach might be to lift `getg` and `sideg` with this converter and have them perform the conversion on the global parts, not the local parts. That is, `sideg` will take the octagon and convert it to a box and `getg` will do vice versa (create an octagon which just has box constraints).

sim642:
Closing for now since this wasn't used for the evaluation and there's no need to keep this around permanently.

PR #448, opened 2021-11-12
Add Compilation Database support

### LINKED ISSUE #406: Compilation Database support
From an old Slack message:

> Regarding this build systems stuff, I remembered a feature CLion added at one point: compilation database. It's some generic JSON format for C/C++ compilations to list the compiled files etc from a number of different build systems. So maybe this is what we should focus on by just assuming the user has created a compilation database ~~and let magpie extract the list of files from that~~. Or maybe even **make Goblint directly parse compilation databases**.
> * CLion documentation about it: https://www.jetbrains.com/help/clion/compilation-database.html
>     * CMake has the ability to output that
>     * Makefiles can also somehow be converted to it with compiledb-generator directly or the cilly-like interception tool Bear: https://github.com/rizsotto/Bear
> * Clang documentation about it: https://clang.llvm.org/docs/JSONCompilationDatabase.html
> * Some other overview of it (lists other analysis tools that can use it): https://sarcasm.github.io/notes/dev/compilation-database.html

sim642:
Closes #406.

So far I've just tested it with this tiny example CMake project I set up: https://github.com/sim642/goblint-action-test/tree/compilation-database/cmake-project. Nevertheless, it has a library component and also has custom `-D` preprocessing arguments for the build.

### Changes
1. Parsing of compilation databases.
2. Preprocessing of compilation database command objects, either from `command` (from CMake) or `arguments` (from [Bear](https://github.com/rizsotto/Bear), [compiledb](https://github.com/nickdiego/compiledb)).
3. Output (compilation database) preprocessed files into new `.goblint/preprocessed/` subdirectory. Over time we should consolidate all of Goblint's generated data (e.g. `incremental_data`) there for ease of management instead of scattering our files in top-level directories.
4. Support for directories to be analyzed, e.g. `goblint .` for current directory. Looks up compilation database or `Makefile` in the directory and uses that; doesn't recurse into individual C files because that's extremely unlikely to be correct on any real project.
4. Change `cppflags` option from `string` to `string list`. This integrates better with command objects with `arguments` (otherwise we'd have to parse shell commands _a la_ Python's `shlex`).
5. `autogen` support by @stilscher. It can be there although I believe Goblint shouldn't be in charge of configuring the project that it's analyzing (and sometimes you might want to use an alternative configuration instead of the default we currently call). Instead the user should have configured the project as desired (including with CMake) and we analyze just that (e.g. the compilation database or `Makefile`).

### TODO
- [ ] Try on real projects to see if it works.


sim642:
I also added support for analyzing directories, e.g. `goblint .`. This will be useful for GobPie, so it can just call Goblint on the enabled directory instead of duplicating the logic to figure out what (compilation database or `Makefile`) to use.

michael-schwarz (approved):

PR #449, opened 2021-11-16
Add location ranges to output

### LINKED ISSUE #428: Location ranges for output
Currently we only have start locations (file, line, column) for statements from CIL, but it's increasingly useful to also have end locations, i.e. the ability to describe location ranges.

For example, GobPie needs them to specify where a warning's squiggly underline should end. Right now it uses a heuristic to search for the next `;` from the start location.

Similarly SARIF (#376) needs end locations. Right now it very stupidly just uses `endLine = startLine + 4`.

Since there are at least two downstream uses for location ranges, it might be more sensible for Goblint to be able to provide them itself. The options are:
1. Move the `;` finding heuristic into Goblint. This can internally be replaced with something better later without altering any of its uses.
2. Change CIL to provide end locations for at least statements if nothing else. This is probably a bigger undertaking and requires hacking its lexer/parser to return more information, which it technically does know.

sim642:
Closes #428.

This is based on https://github.com/goblint/cil/pull/51.

### TODO
- [ ] ~~Add extra locations to functions for just the name/declaration.~~ Not currently needed.
- [x] Fix opam pin after https://github.com/goblint/cil/pull/51 is merged.

### DIFF
--- a/goblint.opam
+++ b/goblint.opam
@@ -55,7 +55,7 @@ dev-repo: "git+https://github.com/goblint/analyzer.git"
 # on `dune build` goblint.opam will be generated from goblint.opam.template and dune-project
 # also remember to generate/adjust goblint.opam.locked!
 pin-depends: [
-  [ "goblint-cil.1.8.1" "git+https://github.com/goblint/cil.git#loc-end" ]
+  [ "goblint-cil.1.8.2" "git+https://github.com/goblint/cil.git#a3c91aa6e8f946fec9a9a13361b051a73b12a65c" ]
   [ "ppx_deriving.5.2.1" "git+https://github.com/ocaml-ppx/ppx_deriving.git#0a89b619f94cbbfc3b0fb3255ab4fe5bc77d32d6" ]
   [ "ppx_deriving_yojson.3.6.1" "git+https://github.com/ocaml-ppx/ppx_deriving_yojson.git#e030f13a3450e9cf7d2c43fa04e709ef608486cd" ]
   [ "zarith.1.12-gob0" "git+https://github.com/goblint/Zarith.git#goblint-release-1.12" ]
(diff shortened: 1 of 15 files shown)


sim642:
Currently for testing purposes I made `CilType.Location.show` fail if the end location is missing in order to easily find all constructs where it was still missing. I will undo that now, but it did reveal two interesting cases.

### 12/01 containment/analyse
Crashes with
```
Option warning: Without thread escape analysis, every local variable whose address is taken is considered escaped, i.e., global!
Unimplemented: joinLoc LLVM INTERNAL:1:1--1:-1 regression.h:5:1--1:-1
```
Apparently that old C++ -> C converter produced some crazy location info where a function beginning is in one filename and the end is in another. There's no sensible way to construct such location range.

So far I haven't seen any other case where such insanity happens, so maybe we can just ignore it and it'll be fixed after we remove the contain analysis from "lean" Goblint.

### 24/08 octagon/previously_problematic_f
Crashes with
```
[Warning][Integer > Overflow][CWE-190] Signed integer overflow ([Error][Unknown] About to crash! (Fatal error: exception Failure("no end at initfun:0:0")
```
The output is wacky because the exception happens when printing the location of the previous message, but in essence the non-apron octagon analysis uses `IntDomain` internally and in its computations performs operations which would be an overflow (despite the program itself not having one). And this doesn't happen during some transfer functions, but lattice operations like `join`, which is the reason for the weird `initfun:0:0` location.

sim642:
I now adapted this to https://github.com/goblint/cil/pull/51#issuecomment-973006021 as well. The just expression locations for `if`s are now used for the `Test` edges in the CFG, instead of the location spanning the entire `if`. This way warnings about the expression are nice and local.

Issue #403, opened 2021-10-19
Make `MapBot` total again


sim642:
The following are excerpts from a Slack conversation. The goal is to fix `MapBot` such that it normalizes itself to not contain bottom values (and have all its operations behave right w.r.t that). This requires fixing some usages of `MapBot` (like base's `CPA`) which currently abuse the map's keyset (e.g. using `mem`, `fold`, `iter`, etc).

---

**Here's a fun little puzzle in fundamental lattice theory:**
Consider our `MapBot` domain, which return default `bot` values of the inner lattice for missing keys. And consider `a = {k -> bot}` and `b = {}` (empty map) as elements of `MapBot`.
For all keys `k'` we have `find k' a = find k' b` due to the default `bot` behavior. So they're observably the same map (functional extensionality).
Should `a leq b` be true or false?

Goblint says false because of the following code: https://github.com/goblint/analyzer/blob/cbd7987327def50883abbe07fde03f08bc8776be/src/domains/mapDomain.ml#L363-L372

Since `k` is not in `b` and the default-bottoming `find` is not yet defined (the standard `Map.find` is used), `Not_found` is raised and the result is false.
On the other hand, they're observably the same map, so reflexivity should hold.

---

**An even more basic puzzle:** should `a equal b` be true or false?
Again, Goblint says false because it uses the standard `Map.equal` which says false due to differing key sets.

The consequence of it is that if you put such maps into the context, then they're considered different, despite functionally being the same. This doubles the number of contexts per each key where this may happen.
Same goes for keeping such maps inside sets.

---

**Yet another puzzle about these maps:** for all maps `m` and keys `k`, is `add k (find k m) m` equal to `m`?
If you do this for `b` and `k` from above, you get a since it makes the previously implicit default value explicit. But should it?

This isn't hypothetical, but actually happens inside Goblint.

---

I think it boils down to what mathematical structure `MapBot` is supposed to be, because there are two incompatible concepts:
Partial maps, which carry their domain set (which is what OCaml's `Map` does) and have their equality/leq also depend on the compatibility of each of their domains. But if this were so, `find` should never return default bottoms, because they're outside the mathematical domain of the partial map. So pretending to have gotten a value from a non-existent key is nonsense.
[...]

PR #450, opened 2021-11-17
Remove outdated and broken ocamlbuild

sim642:
AFAIK nobody is using these ocamlbuild rules for anything anymore. I tried to and they're utterly broken as well because our dune setup has evolved considerably to all sorts of advanced mechanisms: alternative dependencies for Apron and Z3; virtual libraries for dune-site (and Gobview). It's probably not worth the effort trying to make the inferior ocamlbuild work, so we can just drop the unnecessary dependency.

On a related note, we still have the `travis-ci.sh` script, which we don't use for CI anymore, but `Vagrantfile` does. It contains the following line:
https://github.com/goblint/analyzer/blob/0da0464dfd65c3d84566a71b55f9e5f19b218843/scripts/travis-ci.sh#L21
Now that the ocamlfind dependency is also dropped, do we not need `m4` anymore? I don't know what it would be needed for, but currently we still claim to require it in our README etc.

### DIFF
--- a/_tags
+++ b/_tags
@@ -1,5 +1,0 @@
-<src> or <src/**>: include
-true: package(findlib), package(goblint-cil.all-features), package(batteries.unthreaded), package(ppx_deriving.std), package(ppx_deriving_yojson), package(ppx_distr_guards), package(qcheck-core.runner)
-<unittest/**>: include, package(oUnit)
-<src/extract/*.*>: package(ocaml-monadic)
-true: bin_annot, debug, optimize(3), thread
--- a/dune-project
+++ b/dune-project
@@ -37,9 +37,6 @@
     dune-site
     sha
     benchmark ; TODO: make this optional somehow, (optional) on bench executable doesn't work
-    ; TODO still need the following after switch to dune?
-    ocamlbuild
-    ocamlfind
   )
   (depopts
     apron
--- a/goblint.opam
+++ b/goblint.opam
@@ -34,8 +34,6 @@ depends: [
   "dune-site"
   "sha"
   "benchmark"
-  "ocamlbuild"
-  "ocamlfind"
 ]
 depopts: ["apron" "z3"]
 build: [
--- a/make.sh
+++ b/make.sh
@@ -19,14 +19,6 @@ opam_setup() {
   # opam install camlp4 mongo # camlp4 needed for mongo
 }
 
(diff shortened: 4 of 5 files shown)


michael-schwarz:
This also removes our way of building unittests and the `mainarinc` executable. Is this intended?
If yes, we should discuss whether those things should be removed.

If no, we should add the possibility to build those things again.

sim642:
I tried to build the unit tests recently and that doesn't work right now anyway. They haven't been executed for so long that they just assume nonexistant modules and interfaces. Maybe we shouldn't delete them entirely yet though, because for #403 I was thinking of maybe writing some tests eventually. Given that they don't work with ocamlbuild right now anyway, this doesn't make their building any worse.

I have no idea about `mainarinc` though. Probably hasn't been built in forever either.

sim642:
Actually, I just remembered: `mainarinc` is listed under `modules` here:
https://github.com/goblint/analyzer/blob/fd4aecfa33db6f74de6ec0b510cf03aa570cc053/src/dune#L49-L53

But I have no idea what that's supposed to mean because it's not given an executable name and the existing ones don't even match up with the listed modules.

michael-schwarz:
`mainarinc` still compiles though. It seems to take a marshalled ARINC result and then print some Promela model and dot graph.
Maybe @vogler knows the details here, and whether this executable is worth keeping?

I think this model extraction for ARINC/pthreads is something that we should keep also in the "slim" Goblint.

sim642:
I just tried as well that both `mainarinc` and `mainspec` compile with dune, so I added them. Interesting how `mainspec` was already listed under `modules`, but there wasn't anything in make for that.

Also, apparently `names` and `modules` don't need to zip up, but `names` themselves are names of modules and `modules` is just there because we reuse the directory also for the library.

michael-schwarz (approved):

vogler:
> `mainarinc` still compiles though. It seems to take a marshalled ARINC result and then print some Promela model and dot graph. Maybe @vogler knows the details here, and whether this executable is worth keeping?
> 
> I think this model extraction for ARINC/pthreads is something that we should keep also in the "slim" Goblint.

Yes, it's the Promela codegen part which also has some options for which you don't want to re-run the analysis.
Those other main* could be integrated into `maingoblint.ml`, but IMO that's too messy (and pulls in a lot of deps for their build).

Issue #213, opened 2021-05-04
Migrate from CIL's `Pretty` to OCaml's `Format`


sim642:
In #210 it came up multiple times that OCaml's standard `Format` module could replace `Pretty` from CIL with additional benefits:
1. It's maintained in OCaml's standard library while `Pretty` isn't properly maintained (maybe we should do this migration even in our CIL fork eventually).
2. It has much more features for producing pretty output.
3. It's supposed to be faster by not constructing `Pretty.doc` structures but writing to a formatter directly (a la how `Printable.printXml` currently works).
4. It's supposed to be faster by using GADTs for internal implementation instead of `Obj` (https://github.com/goblint/analyzer/issues/210#issuecomment-831395278).
5. `ppx_deriving.std` has `deriving show` that also derives a pretty-printer in addition to a simple `show` (https://github.com/goblint/analyzer/issues/210#issuecomment-831395278).

### TODO
- [ ] Replace `Printable.pretty` with `val pp: Format.formatter -> t -> unit`. This is the de facto standard naming for these pretty printer and `deriving show` also derives this signature.
- [ ] Rewrite every implementation of `pretty` as `pp`.
- [ ] Rewrite every `Pretty.printf` with `Format.printf` by using `pp` instead of `pretty`.
- [ ] Rewrite tracing system to use `Format` instead of `Pretty`.
- [ ] Rewrite every `trace` by using `pp` instead of `pretty`.
- [ ] Implement `pp` for CIL types, e.g. `pp_exp` to replace `d_exp`. CIL has a lot of `Pretty`-based code to pretty-print everything (including entire programs). We don't need all of it though unless we also decide to migrate CIL itself.

michael-schwarz:
> maybe we should do this migration even in our CIL fork eventually

I think we should probably do this at the same time, because otherwise we will be relying on `Pretty.sprint` for the output of CIL types to strings which makes for strange code:

~~~OCaml
Pretty.sprint ~width:80 (Cil.d_exp () exp)
~~~

sim642:
[...]

PR #455, opened 2021-11-23
Remove generation of goblint.ml with all opens

sim642:
This nonsense has driven me mad forever, especially because no other OCaml project as far as I have seen needs to do this.
It has been especially annoying if you delete, rename or move modules and want to switch between branches, because the `(source_tree .)` cannot possibly know about deletions, so it will not regenerate goblint.ml and compilation fails due to the missing module. The solution has been to `make clean` to force it to be regenerated, but that wastes so much time unnecessarily.

Apparently this has been completely unnecessary under dune all along because the `-linkall` option already guarantees that all modules are linked. The `at_exit` is still necessary to ensure that Goblint's entry point runs after all the toplevel code in all the linked modules.

To be honest, I'm also tempted to get rid of `scripts/set_version.sh` and especially the `trace_on`/`off` counterparts. The version script also does some MacOS hackery at configure time, but finding the correct `cpp` to use could easily be done at runtime, which would also be more portable.


vesalvojdani:
It's great to simplify these things, but what would you do about the tracing? Do you think we could get rid of the if-statements if one had a properly lazy approach, like [this](https://erratique.ch/software/logs/doc/Logs/index.html#basics), or do we still need the conditional crap? I think our current framework also does not create the docs, but it still invokes the printf pseudo-printer, even when the tracing subsystem is disabled.

sim642:
> It's great to simplify these things, but what would you do about the tracing?

This is particularly confusing, even from purely build system point of view. I thought it would be good to use dune's promotion such that these generated files show up in the `src` directory instead of having to inspect them inside `_build/default/src` if necessary. Removing `make gen` seems to make it possible at first sight (removes a cyclic dependency), but it doesn't actually work: these `trace_on`/`off` scripts change the ones in `src` directly, but promotion always regenerates them inside `_build` and copies them back (by overwriting).

One idea I had would be to use dune profiles such that one has to create (but not commit) a `dune-workspace` file to change the profile. And based on the profile the boolean could be switched via some dune rules hopefully. But it's not also very clear if this is an intended use case of dune profiles or not.

> Do you think we could get rid of the if-statements if one had a properly lazy approach, like [this](https://erratique.ch/software/logs/doc/Logs/index.html#basics), or do we still need the conditional crap?

I wonder how that approach actually compares to ours, it's not so obvious. That one still always calls some external library function, not sure how much that gets optimized. Ours on the other hand has an explicit `if` using a compile-time constant boolean, which I guess is supposed to be optimizable. I haven't tried to ever check though whether it actually is.

> I think our current framework also does not create the docs, but it still invokes the printf pseudo-printer, even when the tracing subsystem is disabled.

That's probably true, but it might not have notable overhead since by ignoring everything it doesn't really do much other than iterating through the format string once. Generalizing to `Format` might be more optimal due to all the GADT optimizations with format strings, but that's a whole project on its own anyway (#213).

---

Another thought I've had at some point is maybe using a ppx rewriter to have a special extension point for our tracing, which would avoid having to write the `if`s. But this makes the code even less readable and reusable because it depends on our own ppx rewriters, so I'm not sure I'd actually want to ever do that.

sim642:
I went all in now and completely replaced these generation things with dune rules, including the entirety of `set_version.sh`. I wonder if we should be this extreme.

michael-schwarz (approved):
LGTM! Nice to see things becoming more clean!

vesalvojdani:
We should update the docs as well. I think they explicitly refer to these trace scripts.

vesalvojdani:
Also, I get build errors when trying this. Interesting that the CI has no problems. Would be nice to have something like `make upgrade` that basically works like `rm -rf _opam; make setup; make dev`. Typically the `opam install --deps-only .` and upgrading cil would work. Not sure if that's quite the same as `make deps`?

michael-schwarz:
>  I think they explicitly refer to these trace scripts.

The scripts are still there and still work though, right? Or did I misunderstand?

vesalvojdani:
Yes. They work, but it's not ideal to recommend stuff that immediately print out "hey, we're deprecated" :)

sim642:
> The scripts are still there and still work though, right? Or did I misunderstand?

Yeah, I decided to adapt them for now, so the previous workflow for switching to trace mode still works. Although you could just create the `dune-workspace` file yourself and simply go edit that to change, no need to invoke the scripts.

PR #456, opened 2021-11-24
Fix: max vid computation in incremental analysis

### LINKED ISSUE #427: Fix computation of `max_vid` in VersionLookup
For the computation of the `vid` one needs to iterate over the local variables within functions. 

`UpdateCil` also has some code to compute the max_ids, could it be reused?

See https://github.com/goblint/analyzer/pull/409#discussion_r739295930_

stilscher:
Closes #427 

This PR introduces the following changes:
- the computation of the maximum `vid` (when running in incremental mode but without loading any previous results) of a function is fixed such that it not only considers the functions `vid` but also includes the `vids` of local variables and formal parameters. 
- The code is refactored to remove the then duplicate computation of the maximum ids in `UpdateCil.ml`.
- Additionally, the incremental.sh script for running the incremental analysis on a repository is changed such that the removing of the possibly not existing output and incremental data directories will not cause it to fail.

### DIFF
--- a/src/incremental/versionLookup.ml
+++ b/src/incremental/versionLookup.ml
@@ -21,7 +21,8 @@ let create_map (new_file: Cil.file) =
   let add_to_hashtbl tbl (global: Cil.global) =
     match global with
     | GFun (fund, _) as f -> (match fund.smaxstmtid with
-        | Some i -> update_vid fund.svar.vid; update_sid i; Hashtbl.replace tbl (identifier_of_global f) f
+        | Some i -> update_vid fund.svar.vid; List.iter (fun v -> update_vid v.vid) fund.slocals;
+          List.iter (fun v -> update_vid v.vid) fund.sformals; update_sid i; Hashtbl.replace tbl (identifier_of_global f) f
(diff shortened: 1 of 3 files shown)


src/incremental/versionLookup.ml
```diff
@@ -16,19 +32,14 @@ let updateMap (oldFile: Cil.file) (newFile: Cil.file) (ht: (global_identifier, C
...
-        | Some i -> update_vid fund.svar.vid; update_sid i; Hashtbl.replace tbl (identifier_of_global f) f
-        | None -> failwith "smaxstmtid should have been computed when calling Cil.computeCFGInfo")
-    | GVar (var, _, _) as v -> update_vid var.vid; Hashtbl.replace tbl (identifier_of_global v) v
-    | GVarDecl (var, _) as v -> update_vid var.vid; Hashtbl.replace tbl (identifier_of_global v) v
-    | other -> ()
-  in
+    | GFun _
+    | GVar _
+    | GVarDecl _ -> Hashtbl.replace tbl (identifier_of_global global) global
+    | other -> () in
```
jerhard, line 40: Maybe `other` should be changed to `_`?

jerhard (approved):

PR #457, opened 2021-11-25
Add shorthand for `--set`

sim642:
This allows `--option value` as shorthand for `--set option value`, which saves a bit of typing on the command line. It doesn't mean that we should completely get rid of `--set` everywhere, but this is just a small convenience.

It's a tiny change code-wise, but I wanted to have a PR for visibility instead of secretly slipping it straight into master without anyone else being aware of this.

I thought about even allowing just `--option` (no argument) for boolean options, but that only goes one way: it could make something true, which is false by default, but not vice versa (without being super confusing). I think there's no need to go that far, because `--enable`/`--disable` are much clearer and not too different. Also one can still do `--option true`/`--option false` this way.

### DIFF
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -84,6 +84,11 @@ let option_spec_list =
     set_bool "dbg.print_dead_code" true;
     set_string "result" "sarif"
   in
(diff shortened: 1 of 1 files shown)


michael-schwarz (approved):

michael-schwarz:
Good idea!

PR #463, opened 2021-11-26
Use ID.starting 0 as default length for an array

michael-schwarz:
Follow-up to #462.
Assume that arrays for which the index expression is not precisely known at least have a positive size.


src/cdomains/valueDomain.ml
```diff
@@ -1054,12 +1054,14 @@ struct
...
         let newl = match e with
-          | None -> ID.top_of (Cilfacade.ptrdiff_ikind ()) (* TODO: must be non-negative, top is overly cautious *)
+          | None -> ID.starting (Cilfacade.ptrdiff_ikind ()) Z.zero
           | Some e ->
             begin
               match eval_exp e with
               | `Int x -> ID.cast_to (Cilfacade.ptrdiff_ikind ())  x
-              | _ -> ID.top_of (Cilfacade.ptrdiff_ikind ()) (* TODO:Warn *)
+              | _ ->
+                M.warn "Expression for size of VLA did not evaluate to Int at declaration";
```
sim642, line 1063: Given that we continue regardless, I think it should be `M.debug ~category:Analyzer` like some other expression evaluation messages in base now.

sim642 (approved):

PR #466, opened 2021-11-29
Invalidate variables that are assigned return values of `pthread_create(...)` and `pthread_join(...)`

### LINKED ISSUE #387: Missing invalidate of assigned variable on `pthread_join`
The variable that the return value of a `pthread_join` operation is assigned to needs to be invalidated, as it is assigned either `0` or an error code, depending on whether `pthread_join` succeeded. It currently is not.

Minimal example:
https://github.com/goblint/analyzer/blob/79c30aae5e75c15661f3eac4b4db8db39351b856/tests/regression/01-cpa/54-tid-invalidate.c#L1-L16

michael-schwarz:
Previously the invalidation in these two cases was missing.

For most library functions the invalidation is done by default as `special` falls through to the default case.
https://github.com/goblint/analyzer/blob/2e8aa66c46643d6774f3c60e70f988004aa121a9/src/analyses/base.ml#L2250

For these two, however, the call needs to be made explicit.

Closes #387

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2057,7 +2057,12 @@ struct
     invalidate ~ctx (Analyses.ask_of_ctx ctx) gs st addrs
 
   let special ctx (lv:lval option) (f: varinfo) (args: exp list) =
-    (*    let heap_var = heap_var !Tracing.current_loc in*)
+    let invalidate_ret_lv st = match lv with
+      | Some lv ->
+        if M.tracing then M.tracel "invalidate" "Invalidating lhs %a for function call %s\n" d_plainlval lv f.vname;
+        invalidate ~ctx (Analyses.ask_of_ctx ctx) ctx.global st [Cil.mkAddrOrStartOf lv]
+      | None -> st
+    in
     let forks = forkfun ctx lv f args in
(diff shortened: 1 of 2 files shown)


src/analyses/base.ml
```diff
@@ -2170,21 +2175,24 @@ struct
...
+          | `Address ret_a ->
+            begin match eval_rv (Analyses.ask_of_ctx ctx) gs st id with
+              | `Thread a ->
+                let a = List.fold AD.join (AD.bot ()) (List.map (fun x -> AD.from_var (ThreadIdDomain.Thread.to_varinfo x)) (ValueDomain.Threads.elements a)) in
+                (* TODO: is this type right? *)
+                set ~ctx:(Some ctx) (Analyses.ask_of_ctx ctx) gs st ret_a (Cilfacade.typeOf ret_var) (get (Analyses.ask_of_ctx ctx) gs st a None)
+              | _      -> invalidate ~ctx (Analyses.ask_of_ctx ctx) gs st [ret_var]
+            end
+          | _      -> invalidate ~ctx (Analyses.ask_of_ctx ctx) gs st [ret_var]
+        end
```
sim642, line 2193: You can get rid of the outer `begin...end` and one level of indentation now because it's inside `let`.

sim642 (approved):

PR #471, opened 2021-11-30
Flag Helper for Binary Flag Configured `Printable`s, `Groupable`s, and `Lattice.PO`s

michael-schwarz:
This enables some more code reuse.
There is also `StructDomain.FlagConfiguredStructDomain`, but in this case there is three options and it is not immediately clear whether there can be meaningful reuse of code between this one and the other one which has two options.

### DIFF
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -689,19 +689,12 @@ struct
   type idx = Idx.t
   type value = Val.t
 
-  include FlagHelper(P)(T)(struct
+  include LatticeFlagHelper(P)(T)(struct
       let msg = "FlagConfiguredArrayDomain received a value where not exactly one component is set"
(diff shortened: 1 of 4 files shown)


sim642:
> There is also `StructDomain.FlagConfiguredStructDomain`, but in this case there is three options and it is not immediately clear whether there can be meaningful reuse of code between this one and the other one which has two options.

Well, you could just define `Flag3 (A) (B) (C) = Flag2 (A) (Flag2 (B) (C))`, no?

michael-schwarz:
> Well, you could just define Flag3 (A) (B) (C) = Flag2 (A) (Flag2 (B) (C)), no?

I tried that, but I am not sure if this is nicer now?

michael-schwarz:
I now added another helper for things that are (almost) lattices, I think this is the most reuse we can obtain reasonably.

sim642 (approved):

Issue #460, opened 2021-11-25
Remove unmaintained features


sim642:
After Goblint 1.1.0 aka "fat" the plan is to release Goblint 2.0.0 aka "lean", where we have the opportunity to remove numerous unmaintained and deprecated features which might hinder refactoring and development. This issue is to collect ideas for such removals and discuss them.

### Ideas
- [x] **Containment analysis.** Requires LLVM C++ → C conversion, which has been removed upstream. `Goblintutil` has a bunch of name demangling logic specific to that conversion.
- [x] **Deadlocks by races analysis.** Seems suspiciously incomplete: uses must and may locks, but in the predicates completely ignores both. Probably an attempt to implement this paper: https://www.sciencedirect.com/science/article/pii/S2352220814000492.
- [x] **Shapes analysis.** Highly specific to cyclic doubly-linked lists. Builds on region analysis, making the latter more difficult to fix (#107).
- [x] **Non-Apron octagon analysis.**
- [x] **nestra tests.** Currently unused. Should be converted into regression tests (with desired checks added) or just removed.
- [x] **osek**, **osektransactionality**, **flag**, **flagmodes**?
- [x] **`presub` and `postsub`.** They are part of every analysis spec `ctx`, but only used by some old analyses and mostly replaced with queries. During profiling I have noticed that `filter_presubs` can take 10% of runtime manipulating the assoc lists used for `MCP2`. (Might also be a good idea to rewrite that using `Map`?)
    - [x] `postsub` removal is part of #737
    - [x] `presub` is part of #812
- [x] **ListDomain in base analysis.** Very unused, untested and broken. https://github.com/goblint/analyzer/pull/737#discussion_r874431687
- [ ] ~~**condvars analysis.** Unused, not subsumed by var_eq?~~ Now part of #743.
- [x] **Phases.** Currently all phases must be manually specified ahead of time and there's no pure communication between the phases. We probably want something more dynamic _a la_ refinement anyway. The `phases` field in the JSON configuration complicates things due to its recursive schema (#476). #820
- [x] **arinc**? @vogler (#812)
- [x] **extraction** (arinc ~~, osek~~)? #812

vesalvojdani:
[...]

PR #473, opened 2021-12-01
Generalize global constraint variables to `module V: Printable.S` in each analysis

### LINKED ISSUE #469: Generalize global constraint variables from `varinfo`
Each activated analysis should be able to define its own global constraint variables (and their type). All these should correspond to different constraint variables in the constraint system such that their dependencies are more fine-grained and independent.

Keeping them separate could also improve precision because widening happens separately per-analysis.

sim642:
Closes #469.

### Changes
1. Adds `module V: Printable.S` to each analysis spec, which will be the module used for global constraint variables for that analysis. Since it's abstract `V.t` can be anything, not necessarily `varinfo`.
2. Adds `'v` type parameter to `ctx`. This is necessary to make `ctx.global` and `ctx.sideg` correctly typed for that `V.t`.
3. Rewrite `MCP`'s globals handling to use "variants" instead of lists for both global variables and their values. Each variant is just a pair of the analysis ID (`int`) and the corresponding variable/value as `Obj.t`.
4. Refactor some analyses to use non-`varinfo` globals: thread, threadJoins, arinc and extract_arinc.
5. Require analyses to explicitly define `V` if they want to use any globals (including `varinfo`). `DefaultSpec` just uses an empty (variant) type, which means no globals can be used at all.
6. Fix g2html to handle the changed XML where globals have just one analysis value, not all.

### TODO
- [x] Comparison runs on sv-benchmarks to see if the abstract variant for globals has any overhead compared to straight `varinfo`. Also to see if immediately any performance or precision improvement arises from the separation of globals for different analyses.
- [x] Refactor region analysis as well, after removing the dependent shapes analysis (#474, #460).
- [x] Fix Gobview build.
- [ ] Refactor access analysis in #391 to avoid grouping all struct-based accesses together.


michael-schwarz:
Looks good thus far! I have to say I am thoroughly impressed that this did not require a major overhaul of everything!

michael-schwarz:
What would be interesting is if the number of globals that are widening points changed, the hypothesis being that less widening might be applied. But then this is super hard to measure because we now have a lot more globals than we did before.

Also, it would be interesting to see what happens with runtime and the number of variables and evaluations if one runs one of the larger multi-threaded programs (e.g. the traces or more traces bechmarks), or some of the other larger projects from our test repo with `earlyglobs`.

sim642:
> Also, it would be interesting to see what happens with runtime and the number of variables and evaluations if one runs one of the larger multi-threaded programs (e.g. the traces or more traces bechmarks), or some of the other larger projects from our test repo with `earlyglobs`.

I ran the traces benchmarks before and after, didn't notice much of a difference there either (and the bigger programs were from sv-benchmarks anyway). I don't know about the single-threaded benchmarks in bench though.

It's also the case that the base analysis and privatizations still use `varinfo` variables for everything. I don't know yet if there are any real opportunities to use the new generalization there, because most of the additional global variable components (background lockset, write lockset, thread ID) still need to be universally quantified over, so the (nested) maps need to remain to make that possible.

sim642:
> It's also the case that the base analysis and privatizations still use `varinfo` variables for everything. I don't know yet if there are any real opportunities to use the new generalization there, because most of the additional global variable components (background lockset, write lockset, thread ID) still need to be universally quantified over, so the (nested) maps need to remain to make that possible.

I have now also refactored all the privatizations at the top level at least. So no more `MUTEX_GLOBAL_g` and `MUTEX_INITS` varinfos any more. Similarly for base protection-based I separated the unprotected and protected globals. Moreover, I replaced the global product domains for those things with `Lift2`s.

sim642:
I ran sv-benchmarks again after my futher refactorings and there is now a 3% improvement in the regression line compared to `master`:
image
(Full comparisons here: https://goblint.cs.ut.ee/results/50-all-fast-gen-glob-after-refactor/results/).

I'm not sure where this improvement exactly came from though because the changes were to privatizations, which only matter for concurrent benchmarks and even then we don't use any of the more expensive privatizations.

src/analyses/basePriv.ml
```diff
@@ -293,19 +294,10 @@ struct
...
-    | `Return -> (* required for thread return *)
-      begin match ThreadId.get_current ask with
-        | `Lifted x when CPA.mem (ThreadIdDomain.Thread.to_varinfo x) st.cpa ->
-          let x = ThreadIdDomain.Thread.to_varinfo x in
-          let v = CPA.find x st.cpa in
-          sideg (mutex_global x) (CPA.singleton x v);
-          {st with cpa = CPA.remove x st.cpa}
-        | _ ->
-          st
-      end
```
michael-schwarz, line 308: Why can we now remove this?

michael-schwarz (approved):

PR #477, opened 2021-12-01
Do not evaluate global integer variables when writing them.

### LINKED ISSUE #472: Should not `eval` unread globals of integral type
Writes to globals as handled by the Base analysis should not trigger evaluation of those globals as this introduces unnecessary dependences, see #254. As noted in the discussion of #254, it would be a good starting point to try this for global variables of integral type.

jerhard:
This PR adds a check that avoids to evaluate a global variable, when analyzing a write to it in Base, in case that global variable is a global integer variable.

For this, in `set` and `assign` it is checked whether the lval of the global that is assigned to is an integer variable. 

Closes #472.


src/analyses/base.ml
```diff
@@ -1145,7 +1155,12 @@ struct
          * side-effects here, but the code still distinguishes these cases. *)
       if (!GU.earlyglobs || ThreadFlag.is_multi a) && is_global a x then begin
         if M.tracing then M.tracel "setosek" ~var:x.vname "update_one_addr: update a global var '%s' ...\n" x.vname;
-        let new_value = update_offset (Priv.read_global a gs st x) in
+        let old_value = if Option.map_default exp_is_integral_var false lval_raw then
```
jerhard, line 1158: It seems like it should be possible to apply this optimzation in the more general case that `Cil.isIntegralType lval_type` holds, but this somehow produces unreached fixed-points in multiple test cases, e.g. `02/24` that operates on malloced integers.

sim642: The side effect is `(3)`, which I think is just an integer, but the solver has `((3), (4), true)`, which is the same integer, but inside a blob. There probably needs to be some special handling of blobs (or at least ignoring the optimization for them).

sim642: Actually the zero init/origin boolean is confusingly flipped:
https://github.com/goblint/analyzer/blob/3d0cf6e1025814fa22a7cf1e57d578d5c9b3c4e5/src/cdomains/valueDomain.ml#L55-L61
If it's with `calloc`, then it's zero initialized, but then the boolean is `false`.

jerhard: >The side effect is (3), which I think is just an integer, but the solver has ((3), (4), true), which is the same integer, but inside a blob of size 4 (without (!) zero init, see next comment). There probably needs to be some special handling of blobs (or at least ignoring the optimization for them).

I now went with checking whether the lval is an allocated block and then not doing the optimization.

michael-schwarz:
I'll start an `sv-comp` run to see if this has any effect on the benchmark set.

src/analyses/base.ml
```diff
@@ -1732,8 +1740,13 @@ struct
       | _ -> ()
       );
       match lval with (* this section ensure global variables contain bottom values of the proper type before setting them  *)
-      | (Var v, _) when AD.is_definite lval_val && v.vglob ->
-        let current_val = eval_rv_keep_bot (Analyses.ask_of_ctx ctx) ctx.global ctx.local (Lval (Var v, NoOffset)) in
+      | (Var v, offs) when AD.is_definite lval_val && v.vglob ->
```
michael-schwarz, line 1743: Maybe a comment why it is not necessary to handle the case of blobs here?

tests/regression/02-base/66-no-eval-on-write-multi.c
```diff
@@ -0,0 +1,36 @@
...
+int main() {
+  pthread_t t;
+
+  pthread_create(&t, NULL, t_fun, NULL);
+  pthread_mutex_lock(&mutex);
+  bar();
+  pthread_mutex_unlock(&mutex);
+  pthread_join(t, NULL);
+  assert(glob >= 1);
+  assert(glob <= 2); //UNKNOWN
```
sim642, line 33: Should this be `UNKNOWN!`?

sim642:
> I'll start an `sv-comp` run to see if this has any effect on the benchmark set.

How about that?

sim642:
There's also a small conflict to be fixed with `priv_getg`.

tests/regression/02-base/66-no-eval-on-write-multi.c
jerhard: In the concrete semantics, the value of `glob` should either be `2`, from the write in `bar`, or `1`, from the last write in `glob`. AFAIK the `UNKNOWN!` is intended for something that is also unknown in the concrete, so that would not be the case. The abstraction used does not know about the `thread_join`, of course.
But it's probably less confusing if I remove the assert.

sim642: Ahh, right, I didn't really look at the test.

michael-schwarz:
> How about that?

No meaningful difference in either runtime or results.

michael-schwarz (approved):

src/analyses/base.ml
```diff
@@ -1167,7 +1166,16 @@ struct
       if (!GU.earlyglobs || ThreadFlag.is_multi a) && is_global a x then begin
         if M.tracing then M.tracel "setosek" ~var:x.vname "update_one_addr: update a global var '%s' ...\n" x.vname;
         let priv_getg = priv_getg gs in
-        let new_value = update_offset (Priv.read_global a priv_getg st x) in
+        (* Optimization to avoid evaluating integer values when setting them.
+           The case when invariant = true requires the old_value to be sound for the meet.
+           Allocated blocks are representend by Blobs with additional information, so they need to be looked-up. *)
+           let old_value = if not invariant && Cil.isIntegralType x.vtype && not (a.f (IsHeapVar x)) then begin
```
sim642, line 1172: Indentation has gotten a bit off when adding the comment,
```suggestion
        let old_value = if not invariant && Cil.isIntegralType x.vtype && not (a.f (IsHeapVar x)) then begin
```

sim642 (approved):

PR #479, opened 2021-12-03
Adapt to new CIL with CInt instead of CInt64

michael-schwarz:
- [x] Check if there are any instances where now using `cilint` causes probems
- [x] Wait for https://github.com/goblint/cil/pull/53 to be merged
- [x] Update pins

### DIFF
--- a/src/analyses/tutorials/signs.ml
+++ b/src/analyses/tutorials/signs.ml
@@ -2,6 +2,7 @@
 
 open Prelude.Ana
 open Analyses
+open Cilint
 
 module Signs =
 struct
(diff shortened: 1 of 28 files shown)


src/analyses/tutorials/signs.ml
```diff
@@ -60,7 +60,7 @@ struct
 
   (* This should now evaluate expressions. *)
   let eval (d: D.t) (exp: exp): SL.t = match exp with
-    | Const (CInt64 (i, _, _)) -> SL.top () (* TODO: Fix me! *)
+    | Const (CInt (i, _, _)) -> SL.top () (* TODO: Fix me! *)
```
sim642, line 63: The solution to this tutorial is supposed to use `SL.of_int: int64 -> SL.t` here, but that won't work now with `cilint`. The above `Signs.of_int` should be adapted to do comparisons using `cilint` instead such that this is still as straightforward as before.

sim642 (approved):

PR #481, opened 2021-12-06
Add option `incremental.force-reanalyze.funs`

michael-schwarz:
This adds an option such that a user can specify a list of function names that are to be considered as changed regardless of the result of the AST/CFG comparison.
This is need such that certain functions can be marked as "changed" when different configuration options have been enabled for them.

### DIFF
--- a/src/incremental/compareCIL.ml
+++ b/src/incremental/compareCIL.ml
@@ -35,19 +35,22 @@ let eqF (a: Cil.fundec) (b: Cil.fundec) (cfgs : (cfg * cfg) option) =
       List.for_all2 eq_varinfo a.sformals b.sformals
     with Invalid_argument _ -> false in
   let identical, diffOpt =
-    try
-      let sameDef = unchangedHeader && List.for_all2 eq_varinfo a.slocals b.slocals in
-      match cfgs with
-      | None -> sameDef && eq_block (a.sbody, a) (b.sbody, b), None
-      | Some (cfgOld, cfgNew) ->
-        let module CfgOld : MyCFG.CfgForward = struct let next = cfgOld end in
-        let module CfgNew : MyCFG.CfgForward = struct let next = cfgNew end in
-        let matches, diffNodes1, diffNodes2 = compareFun (module CfgOld) (module CfgNew) a b in
-        if not sameDef then (false, None)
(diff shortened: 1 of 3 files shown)


sim642:
I can guess how it's supposed to be used: Gobview changes some precision information, then adds the function to this option, reanalyzes and removes the function from this option. it feels a bit of a roundabout way to do things though.

Wouldn't it just be more general to allow passing extra `incremental_data` to the analysis (via the protocol) to force such function to be reanalyzed?

michael-schwarz:
> Wouldn't it just be more general to allow passing extra incremental_data to the analysis (via the protocol) to force such function to be reanalyzed?

That somehow feels a bit heavy-handed and also restricts applicability: The way it currently is, it would allow to specify functions that should be reanalyzed (effectively from scratch) even when Goblint is used via the commandline.

sim642:
True, although there is no use case outside of incremental analysis for doing that. And it's an option meant for just that internal use, which is easy to misuse from the user perspective, especially since it has this unusual usage pattern of turning on and off.

Maybe it would make more sense as an option if it simply were called `incremental.force-reanalyze-funs` or something. Then it would be about an external thing, not an internal thing. The fact that the forced reanalysis is implemented by adding those functions to the changed list would simply be implementation detail.

jerhard:
>Maybe it would make more sense as an option if it simply were called `incremental.force-reanalyze-funs` or something

Yes, I think that name would convey the meaning quite well.

vesalvojdani:
Would it be possible to use this also for forcing restart of selected globals? At least, maybe the outward interface should look the same: `incremental.force-reanalyze.funs` and `incremental.force-reanalyze.globals`. It would be nice to be able to force the the restart of specific globals, but treating that as a change to global initializers is maybe not the best way to do it.

sim642:
> Would it be possible to use this also for forcing restart of selected globals? At least, maybe the outward interface should look the same: `incremental.force-reanalyze.funs` and `incremental.force-reanalyze.globals`. It would be nice to be able to force the the restart of specific globals, but treating that as a change to global initializers is maybe not the best way to do it.

Indeed it might make sense to have `force-reanalyze` as a subobject, where to place other things besides functions.

sim642 (approved):

PR #483, opened 2021-12-06
Output either release version or commit hash

vesalvojdani:
This should be fairly non-controversial, but it needs review because I'm not 100% this would actually do what I want for a release version... Also, I started to doubt the idea of either showing the proper release version or the git hash (never both): is there a need somewhere (e.g. sv-comp) where displaying both would be good to precisely identify the competing version?

### DIFF
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -10,8 +10,7 @@ let writeconffile = ref ""
 
 (** Print version and bail. *)
 let print_version ch =
-  let open Version in
-  printf "Goblint version: %%VERSION_NUM%% (%s)\n" goblint;
+  printf "Goblint version: %%VERSION_NUM%% (%s)\n" ConfigVersion.version;
   printf "Cil version:     %s\n" Cil.cilVersion;
   printf "Profile:         %s\n" ConfigProfile.profile;
   exit 0
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -12,7 +12,7 @@ let writeconffile = ref ""
 let print_version ch =
   let release_version = "%%VERSION_NUM%%" in
   let version =
-    if String.starts_with release_version "%" then ConfigVersion.version
+    if String.starts_with release_version "%" then Version.goblint
     else release_version in
   printf "Goblint version: %s\n" version;
   printf "Cil version:     %s\n" Cil.cilVersion;
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -10,11 +10,7 @@ let writeconffile = ref ""
 
 (** Print version and bail. *)
 let print_version ch =
-  let release_version = "%%VERSION_NUM%%" in
-  let version =
-    if String.starts_with release_version "%" then Version.goblint
-    else release_version in
-  printf "Goblint version: %s\n" version;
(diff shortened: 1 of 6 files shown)


src/maingoblint.ml
```diff
@@ -10,8 +10,11 @@ let writeconffile = ref ""
 
 (** Print version and bail. *)
 let print_version ch =
-  let open Version in
-  printf "Goblint version: %%VERSION_NUM%% (%s)\n" goblint;
+  let release_version = "%%VERSION_NUM%%" in
+  let version =
+    if String.starts_with release_version "%" then ConfigVersion.version
+    else release_version in
```
sim642, line 16: This is the exact kind of logic that should go into `version.ml` such that our version display is consistent across all four places.

sim642 (changes requested):

sim642:
> Also, I started to doubt the idea of either showing the proper release version or the git hash (never both): is there a need somewhere (e.g. sv-comp) where displaying both would be good to precisely identify the competing version?

Having both for any kind of release is very much useful to avoid mixups. For example, all the SV-COMP submissions are locally tagged with `svcomp22` such that it would be contained in the version output, but if there's need for an updated SV-COMP submission, then that tag is moved. Without the commit hash anywhere in the version output, it would be much harder to verify which binary was compiled from which sources, since both just report as `svcomp22`.
The same goes for opam releases where you might locally tag it multiple times only to find out problems in the release process that need to be fixed and tagged again.

vesalvojdani:
Okay, I agree with svcomp having a precise version id printed is useful, but having the hashes in our opam releases is like admitting we don't trust our release process.

sim642:
I suspect installing directly from opam without out repository won't even have the commit hash because it doesn't use the git repository for getting the sources, but a source archive, and using our `git describe` script stuff to fill that in. I tried at some point in a fresh switch, but I don't remember anymore, I think it said something like `1.1.1 (unknown)`.

This could be avoided if we didn't use our script (or even a shell command in a dune rule), but also another [substitution](https://dune.readthedocs.io/en/stable/usage.html?highlight=subst#dune-subst) like `%%VCS_COMMIT_ID%%`. But that of course does nothing during development strictly in the repo and not installing (https://github.com/ocaml/dune/issues/1539).

There's also another difference. Our script uses `git describe --all --long --dirty` while there's a substitution `VERSION`, but that uses `git describe --always --dirty`. The difference is that our custom thing includes branch names, while the normal describe doesn't. The ugly downside of ours is that it prefixes everything with `heads/` or `tags/`.

I've found a third way of doing this: https://dune.readthedocs.io/en/stable/dune-libs.html#build-info. According to that description, it works for both releases (just containing the version number, no commit hash) and development (`git describe --always --dirty`). But again, that differs from our script.
There's a bunch of open dune issues related to having more control over dune-build-info: https://github.com/ocaml/dune/issues/4453, https://github.com/ocaml/dune/issues/3856 and https://github.com/ocaml/dune/pull/3523.
Also, it doesn't work for Gobview until dune 3.0 is released: https://github.com/ocaml/dune/issues/4444.

In conclusion, there's no good way to do it exactly as we might want right now. I'll have to rethink this for 2.0 by maybe combining some of the above to get everything to work in all the possible build and install scenarios.

sim642 (approved):

Issue #485, opened 2021-12-09
Cleanup of the array domains


michael-schwarz:
Currently, the partitioned array domain is

~~~OCaml
  module Expp = ExpDomain
  module Base = Lattice.Prod3 (Val) (Val) (Val)
  include Lattice.ProdSimple(Expp) (Base)
~~~

with the additional understanding that the first component is never bottom and that if it is top, all three parts of the second component have the same value.

However, this is error prone. See, e.g.,  #482.

Therefore, this domain should be cleaned up such that illegal values can no longer be expressed in it (Probably by switching from `Lattice.ProdSimple` to a custom variant type).

michael-schwarz:
While we do this, we should also cleanup `init_value` and `top_value`:

@michael-schwarz This is starting look like a leaky abstraction. Couldn't the array domains implement it in their corresponding modules? Currently the array domain selection is repeated 5 times instead of 1.

_Originally posted by @sim642 in https://github.com/goblint/analyzer/pull/577#discussion_r794263673_

michael-schwarz:
While doing this, one should also look into improving the unrolled domain to profit from information coming, e.g., from defexc.

PR #486, opened 2021-12-09
Fix meet of arrays with different partitioning expressions

### LINKED ISSUE #482: `Fatal error: exception Failure("ArrayDomain: Unallowed state (one of the partitioning expressions is bot)")` for 120 programs
At some point, we introduced a regression that leads to 120 SVComp tests failing with this new error message.

michael-schwarz:
In the course of #419, I modified the `invariant` function to work on the entire variable instead of only on an offset. The offset is updated and after this the meet between the old and the new value happens.
This is beneficial e.g. with our new struct domains.

However, it uncovered an issue with the `meet` in the partitioned array domain: If two partitioned values are met and they are partitioned according to different expressions, the resulting component-wise `meet` is partitioned according to a bottom expression, which is an invalid value that we never create. (same goes for `narrow`).

This fixes the immediate problem. In the long run, I will later clean up this mess in #485.

TODO:
- [x] Check if this fixes all instances of #482 on SV-Comp

Closes #482.

### DIFF
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -567,7 +567,14 @@ struct
   let smart_widen = smart_widen_with_length None
   let smart_leq = smart_leq_with_length None
 
-  let meet a b = normalize @@ meet a b
+  let meet (e1,v1) (e2,v2) = normalize @@
+    match e1,e2 with
+    | `Lifted e1e, `Lifted e2e when not (Basetype.CilExp.equal e1e e2e) ->
+      (* partitioned according to two different expressions -> meet can not be element-wise *)
+      (* TODO: do smart things if the relationship between e1e and e2e is known *)
+      (e1,v1)
+    | _ -> meet (e1,v1) (e2,v2)
+
   let narrow a b = normalize @@ narrow a b
 
   let update_length _ x = x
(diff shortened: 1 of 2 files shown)


src/cdomains/arrayDomain.ml
```diff
@@ -567,8 +567,21 @@ struct
...
   let smart_leq = smart_leq_with_length None
 
-  let meet a b = normalize @@ meet a b
-  let narrow a b = normalize @@ narrow a b
+  let meet (e1,v1) (e2,v2) = normalize @@
+    match e1,e2 with
+    | `Lifted e1e, `Lifted e2e when not (Basetype.CilExp.equal e1e e2e) ->
+      (* partitioned according to two different expressions -> meet can not be element-wise *)
+      (* TODO: do smart things if the relationship between e1e and e2e is known *)
+      (e1,v1)
```
sim642, line 575: I guess it's safe to return the first one because if it's partitioned by multiple expressions, their meet is partitioned by both, but we can only keep one, so arbitrarily prefer the first?

Seems weird that it's not commutative though, but I have no idea what else could be done here. Maybe worth a code comment pointing this out?

michael-schwarz: I added a comment to this effect. In `join` there is some elaborate logic on what expression to keep. In #485, that logic should be applied everywhere (making a `smart_meet` and `smart_narrow` necessary).

michael-schwarz:
I ran sv-comp again and this fix indeed solves all 120 instances of this problem.

PR #489, opened 2021-12-09
Incremental/automated tests

stilscher:
This pull request includes changes to also automatically run incremental test cases in the github workflow. This should help to detect changes that might break the incremental analysis. The following changes were made:

- make incremental test cases executable with `update_suite.rb` script (run with option `-i`)
    - create sub classes for marshaling and incremental tests and
    - encapsulate running, parsing and evaluation of test sets in classes to avoid code duplication
    - extended html output for incremental test cases (complete only when using the record option `-r`, just as before)
- include the execution of incremental test cases in the github workflow
- introduce sub folders to be able to group incremental test cases in future

The incremental test cases can be found in `tests/incremental/`. The expected result for incremental test cases is encoded just as for the usual regression tests with the comments after an `assert`, for the original run directly in the `.c` file and for the following incremental run the desired changes are put in the `.patch` file. Each test case additionally has a separate configuration file to specify the parameters.


vesalvojdani:
Excellent! But... As I've also been messing with the benchmarking scripts, we've been discussing with Simmo that we really ought to replace all this mess with a well-designed thing in Python that would include the functionality of both this and the benchmarking scripts. Anyway, we definitely need these improvements now, but we should also create a ticket to brainstorm what a grand unified testing framework would look like!

tests/incremental/01-sideeffects.c
```diff
@@ -1,15 +0,0 @@
-// The patch for this test case is empty on purpose. The problem described does occur in the incremental run even without any changes.
```
sim642, line 1: This comment has disappeared in the renamed copy.

tests/incremental/02-changed_start_state.c
```diff
@@ -1,12 +0,0 @@
-#include <assert.h>
-
-int g = 1;
-
-int main() {
-    // After the presolve phase, g is in the start state but neither in the context nor in the start variables.
-    // If the start state of main is not overwriten when main is destabilized because of a changed start state,
-    // the first assert will wrongly succeed and the second fail in the incremental run
-    assert(g == 1); // success before, unknown after
-    assert(g == 2); // fail before, unknown after
```
sim642, line 10: Is there a reason this assertion is flipped in the modified copy?

tests/incremental/00-basic/02-changed_start_state1.patch
```diff
@@ -0,0 +1,19 @@
...
+
+ int main() {
+     // After the presolve phase, g is in the start state but neither in the context nor in the start variables.
+     // If the change of the start state of main would not be propagated by the call to side on all start variables, the
+     // asserts in the incremental run would wrongly fail. Side however only joins with the previous value instead of
+     // overwriting, therefore the current imprecision.
+-    assert(g == 1);
+-    assert(g != 2);
++    assert(g != 1); // TODO (restarting)
++    assert(g == 2); // TODO
```
sim642, line 17: What exactly is happening here? The annotations for before and after incremental run are no longer in the actual code file, but they're also in the patch? And this patch is also changing the assertions themselves, which the test originally did not do.

tests/incremental/00-basic/03-changed_start_state2.patch
```diff
@@ -0,0 +1,19 @@
...
+
+ int main() {
+     // After the presolve phase, g is in the start state but neither in the context nor in the start variables.
+     // If the change of the start state of main would not be propagated by the call to side on all start variables, the
+     // asserts in the incremental run would wrongly fail. Side however only joins with the previous value instead of
+     // overwriting, therefore the current imprecision.
+-    assert(g == 1);
+-    assert(g != 2);
++    assert(g != 1); // TODO (restarting)
++    assert(g == 2); // TODO
```
sim642, line 17: Why are these TODO now? In the previous copy of this test they were not specified as unknown, neither before nor after.

.github/workflows/unlocked.yml
```diff
@@ -75,6 +75,9 @@ jobs:
       - name: Test marshal regression
         run: ruby scripts/update_suite.rb -m
 
+      - name: Test incremental regression
+        run: ruby scripts/update_suite.rb -i
+
```
sim642, line 80: It takes a second to run, so we could easily also have these in the locked workflow.

tests/incremental/00-basic/03-changed_start_state2.patch
stilscher: With commit 2d99ffeff615e49663a7c6724398bac47cd51ea1 we changed the handling of start functions whose start state changed without the context being different. We now propagate the changes in the start state by calling side on all start variables. This joins with the previous value instead of overwriting. Due to this, the assertions in the test cases `changed_start_state1` and `changed_start_state2` could not be verified anymore because the abstraction is more imprecise after the change mentioned above. I think the asserts should be provable when one can restart globals, therefore I have used the `TODO`. But I will gladly change this if it is unfitting and you can suggest something else. I also thought about whether these test cases should be removed instead, because with the `UNKNOWN` assert in the incremental run they don't seem very meaningful.

tests/incremental/01-sideeffects.c
stilscher: I missed that, good that you noticed it. I will reinsert it.

tests/incremental/00-basic/02-changed_start_state1.patch
stilscher: - By putting the annotations for the incremental run in the patch file, the code for the parsing of the annotations can easily be reused for the incremental test case and is the same as for the usual regression tests. Also I think it is nice to separate them from the original non-incremental test case.
- The previous annotations for the incremental run of this test case were no longer verifiable (see comment below)
- Maybe I should revert the flipping of the asserts, I thought it might be clearer this way
- The `SUCCESS` annotation is superfluous, isn't it?

.github/workflows/unlocked.yml
stilscher: Sure, I can do that. I guess we can also always remove it again if it takes too long when we have more incremental test cases.

vesalvojdani:
Now that this script uses proper classes, ruby doesn't look that awful... Clearly, we need similar ideas into the bench repo as well, so I wonder still if there should be a mother of all benchmarking scripts to do it all or we just reimplement these ideas again?

sim642:
> Now that this script uses proper classes, ruby doesn't look that awful...

I don't know, three massive classes with large amounts of similar logic doesn't make it any prettier IMO. To refactor all the logic and clean up all the historic hacks, the script would practically have to be rewritten from scratch anyway. And at that point it would be easier for all of us to do that in Python than Ruby, which all of us have less experience with.


> Clearly, we need similar ideas into the bench repo as well, so I wonder still if there should be a mother of all benchmarking scripts to do it all or we just reimplement these ideas again?

I've added this to the Gobcon agenda for discussion, but I still believe it's worth rethinking all of it instead of duplicating logic back and forth.

sim642 (approved):

PR #490, opened 2021-12-09
TD3 cleanup

vogler:
Some cleanup, renaming, comments.


src/solvers/td3.ml
```diff
@@ -113,49 +104,50 @@ module WP =
...
-            if not wp then tmp
+          let wpd = (* d after widen/narrow (if wp) *)
+            if not wp then eqd
             else
               if term then
-                match phase with Widen -> S.Dom.widen old (S.Dom.join old tmp) | Narrow -> S.Dom.narrow old tmp
+                match phase with
+                | Widen -> S.Dom.widen old (S.Dom.join old eqd)
+                | Narrow ->
+                  assert S.Dom.(leq eqd old || not (leq old eqd)); (* eqd > old would be unsound, but can be incomparable *)
```
sim642, line 129: When can it be incomparable?

Also, is this assertion with two extra `leq` checks something we want to keep here? Was there some issue with this at some point?

sim642 (commented):
Now that the journal version of the TD3 paper is out, is there anything more to be done here?
If not, then I can take care of the merge conflicts and merge it.

src/solvers/td3.ml
sim642: On sv-benchmarks this assertion fails on 7 benchmarks:
image
There's also a ~1% slowdown from this extra checking.

I will try looking into these failures to see what's going on there.

sim642: It's hard to trace the solving of so large programs, but at least at the point of this assert failing there appear to be some odd values. In the narrowing phase, somehow intdomains, address sets, partitioned arrays and var_eq equivalences can suddenly increase. This happens even with the witness lifting disabled.

I think we can keep the assertion because it is still pointing out some potential issues. Such increases in the narrowing phase might also be the reason for their non-termination and are thus worth looking into separately.

This might somehow be related to the fact that `equal` isn't always equivalent to two-way `leq`. We use `equal` to detect the end of the widening phase and switch to narrowing, after which the `leq` in this assertion confusingly no longer holds.

michael-schwarz: We decided in Gobcon to remove this assert.

sim642: I will remove the assertion for merge. It was discussed at GobCon that in non-monotonic systems this isn't necessarily an issue, but just something which theoretically can happen and hasn't been previously observed.

src/solvers/td3.ml
```diff
@@ -301,33 +294,35 @@ module Base =
...
-            update_var_event x old tmp;
-            HM.replace rho x tmp;
+            update_var_event x old wpd;
+            HM.replace rho x wpd;
             destabilize x;
             (solve[@tailcall]) x phase
           ) else (
             (* TODO: why non-equal and non-stable checks in switched order compared to TD3 paper? *)
-            if not (HM.mem stable x) then (
+            if not (HM.mem stable x) then ( (* value unchanged, but not stable, i.e. destabilized itself during rhs? *)
```
jerhard, line 325: Why is there a question mark in the comment?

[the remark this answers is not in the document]
sim642: That was in this PR from the beginning, I removed it now because it should be accurate.

Issue #492, opened 2021-12-10
Change Goblint attribute syntax such that programs will still compile


michael-schwarz:
None of the programs in category `42` compile, as GCC reject them all for misplaced attributes. We should be able to find an attribute placement that only leads to a warning because of an unknown attribute instead of a compile error.

```
michael@michael:~/Documents/goblint-cil/analyzer$ gcc tests/regression/42-annotated-precision/01-def_exc.c 
tests/regression/42-annotated-precision/01-def_exc.c:4:1: error: attributes should be specified before the declarator in a function definition
 int f(int in) __attribute__ ((goblint_precision("def_exc", "interval"))) {
 ^~~
tests/regression/42-annotated-precision/01-def_exc.c:9:1: error: attributes should be specified before the declarator in a function definition
 int main() __attribute__ ((goblint_precision("def_exc"))) {
 ^~~
 ```

sim642:
They need to be placed onto a separate declaration of the same function. See the context attribute tests.

Issue #476, opened 2021-12-01
Replace JSON schema validation and add full options schema


sim642:
Our JSON schema validation for options is something custom, not the standard http://json-schema.org/. Moreover, our default schema is massively incomplete and fails to detect typos when merging entire JSON files in with `--conf`.

We should describe our entire options structure with the official JSON schema and use the only existing OCaml library to validate it:  https://github.com/ocamlpro/ocplib-json-typed (supports Yojson conversion).

The complete schema could entirely replace our `Defaults` module, by also having the default values and descriptions inside the schema. This would also allow describing finite string choice options in the schema and detect invalid values already there.

sim642:
Apparently the library doesn't support arbitrary schema validation out-of-the-box, since it's more geared towards constructing static type-safe encodings: https://github.com/OCamlPro/ocplib-json-typed/issues/38. Also, it's successor is https://gitlab.com/nomadic-labs/json-data-encoding.

I think I'll still be able to abuse the library a bit to do what we need without writing everything from scratch.

vogler:
This is just about runtime validation, in OCaml the options are still referenced by strings, right?
Could define options by mutable records and ppx-generate the json (un)parsing.
Gives you static checks for reading, autocomplete, and presumably better performance - the idea was to be able to bake a fixed configuration by making fields immutable and let the compiler hopefully optimize.
I have a prototype somewhere.

sim642:
Yes, this doesn't change the internal structure that stores the runtime configuration or modifies it. Changing that would be a separate issue.

What nested record types don't give you is the ability to attach descriptions like the current `Defaults` has and like JSON schemas also do. Sure, you can put descriptions into ocamldoc comments, but that's completely static and not exposed in any way at runtime (e.g. `--print_all_options` or to Gobview, etc to show in some interface.

The json-data-encoding library actually is mostly designed around doing something statically typed: you define the whole structure as a GADT with JSON-schema-like constraint possibilities. It even allows exporting a JSON schema itself. But the massive downside is that it has no reasonable way to have those types be nice records, but `objN` and `merge_objs` constructors that it provides just construct a mess of nested tuples, which won't really work for the size of objects we would like to define.

vogler:
> What nested record types don't give you is the ability to attach descriptions like the current `Defaults` has and like JSON schemas also do.

Yes, that was the reason why I didn't continue with it. Would require some custom ppx or registering each option's default value together with a description string.
[...]

PR #493, opened 2021-12-10
Annotations via options

### LINKED ISSUE #491: Configuration options for controlling enabled int domains or context options per function
There should be the ability to control which int domains are used for a function or in what context it is analyzed not just via annotations but also via configuration options.

michael-schwarz:
This adds the option to configure per-function options not just via annotations in the source code but also via dedicated options.

On a side note, it also fixes the placement of these attributes such that the programs inf folder `42` are not rejected by GCC.
Closes #491, #492

### DIFF
--- a/src/util/defaults.ml
+++ b/src/util/defaults.ml
@@ -12,7 +12,7 @@ type category = Std             (** Parsing input, includes, standard stuff, etc
               | Incremental     (** Incremental features                          *)
               | Semantics       (** Semantics                                     *)
               | Transformations (** Transformations                               *)
-              | Annotation      (** Features for annotations                       *)
+              | Annotation      (** Features for annotations                      *)
               | Experimental    (** Experimental features of analyses             *)
               | Debugging       (** Debugging, tracing, etc.                      *)
               | Warnings        (** Filtering warnings                            *)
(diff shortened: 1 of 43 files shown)


src/util/defaults.ml
```diff
@@ -181,6 +181,24 @@ let _ = ()
...
+      ; reg Annotation "annotation.goblint_context.no-widen" "[]" ""
+      ; reg Annotation "annotation.goblint_context.widen" "[]" ""
+      ; reg Annotation "annotation.goblint_precision.no-def_exc" "[]" ""
+      ; reg Annotation "annotation.goblint_precision.def_exc" "[]" ""
+      ; reg Annotation "annotation.goblint_precision.no-interval" "[]" ""
+      ; reg Annotation "annotation.goblint_precision.interval" "[]" ""
+      ; reg Annotation "annotation.goblint_precision.no-enums" "[]" ""
+      ; reg Annotation "annotation.goblint_precision.enums" "[]" ""
+      ; reg Annotation "annotation.goblint_precision.no-congruence" "[]" ""
+      ; reg Annotation "annotation.goblint_precision.congruence" "[]" ""
```
michael-schwarz, line 201: This is all very verbose, any suggestions on how to fix this?

sim642: Maybe use JSON objects for values instead?
```json
{
    "fun1": ["base.int", "widen"],
    "fun2": ["no-widen"]
}
```

michael-schwarz: Yes, I also thought about that, it does indeed appear cleaner.

The downside seems to be that we either

- lose the ability to conveniently set this from the command line and
- lose the ability to reject malformed options such as `non_ptr` instead of `non-ptr`

or need custom code to support these things in such a setting as well.

Maybe for now we go with this more verbose form and rethink it as part of #476 ?

sim642: > * lose the ability to conveniently set this from the command line and

If the default value is `{}`, then it's just another JSON object in our entire options object, so `--set annotation.goblint_context.fun1[+] base.int` should work just as well I think.



> * lose the ability to reject malformed options such as `non_ptr` instead of `non-ptr`

We already lose that when using annotations in the code: we don't fail for unknown annotations and that's how it should be anyway. The ability to do it for options would be just extra safety. With #476 it might become possible to actually specify that as well.

michael-schwarz: > If the default value is {}, then it's just another JSON object in our entire options object, so --set annotation.goblint_context.fun1[+] base.int should work just as well I think.

No, it complains about `Unknown path`. One would have to modify it to just create any path that does not exist yet. I'm skeptical if we want that.

sim642: Right. I guess I'll just have to refactor all this in #476 then, because there it would be completely safe to allow new subobjects to be created at any time. The schema would validate whether a certain object allows arbitrary keys or not. In fact the schema for my suggested way would be notably simpler and less repetitive than it is for this.

sim642 (approved):

PR #494, opened 2021-12-13
Bump goblint-cil to add C11 generic support

michael-schwarz:
Bump goblint-cil so changes from https://github.com/goblint/cil/pull/48 become available. Currently still fails for the following tests:

`12 test(s) failed: ["00/14 startstate", "04/33 kernel_rc", "04/34 kernel_nr", "04/39 rw_lock_nr", "04/40 rw_lock_rc", "04/48 assign_spawn", "04/53 kernel-spinlock", "09/07 kernel_list_rc", "09/08 kernel_list_nr", "09/14 kernel_foreach_rc", "09/15 kernel_foreach_nr", "41/02 bsearch"]`

### DIFF
--- a/goblint.opam
+++ b/goblint.opam
@@ -58,7 +58,7 @@ dev-repo: "git+https://github.com/goblint/analyzer.git"
 # on `dune build` goblint.opam will be generated from goblint.opam.template and dune-project
 # also remember to generate/adjust goblint.opam.locked!
 pin-depends: [
-  [ "goblint-cil.1.8.2" "git+https://github.com/goblint/cil.git#568dc23716a2a5f308fc8e332c7a0ef1302823c4" ]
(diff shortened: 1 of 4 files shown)


michael-schwarz:
The issue is in CIL, not Goblint.

sim642:
I fixed the `bsearch` test: our stub signature was missing `const` from an argument. The corrected `combineTypes` for C11 generics noticed the difference, while the old one did not.

The kernel ones might be related to the `CInt` change somehow, because that error comes from some integer manipulation.

sim642:
> The kernel ones might be related to the `CInt` change somehow, because that error comes from some integer manipulation.

Actually no, it was also related to `const` (or more specifically `pconst`) in `constFold` of casts. I've fixed that in CIL and repinned.

michael-schwarz:
:+1: 
I'll start another sv-comp run to see if we now fail on some program where it worked before and then we can merge.

michael-schwarz:
Looks like this was not enough, the newly added testcase over on the CIL repo still fails.

PR #497, opened 2021-12-15
Expression splitting analysis prototype

sim642:
This is a prototype implementation of an analysis which can split the state by arbitrary (integer) expressions and be path-sensitive w.r.t. those split states.

For the purposes of this prototype, I added two special functions `__goblint_split_begin(exp)` and `__goblint_split_end(exp)` to start and end the splitting, which correspondingly add and remove the splitting expression in the expsplit analysis domain. See the added tests for usage.
I don't know if we want to keep this for annotation (currently the analyzed programs cannot be compiled and especially linked, because there's no runtime no-op stub for them), but these were surprisingly easy to add for fine-grained testing of this feature (to see in HTML where the expressions get added and removed). Also it's neat because the splitting expressions therefore actually correctly refer to the correct `varinfo`s in the scope and CIL itself ensures that the expression only refers to variables in the scope. (I'm saying this because function attributes would not have this property: they would be strings that need to be parsed and the variables associated with those in the scope, which is a lot more trouble).

It was surprisingly easy to implement: there wasn't even need to use `ctx.split`, but just emit a special event, which allows the analysis to evaluate expressions on the post-state and update its own domain accordingly. Moreover, since there's no splitting by a branching condition and condition values are just integers, this whole thing works for splitting by arbitrary integer expressions for free!

I also had to fix an issue, where the path-sensitive functor did not run `should_join` after each non-joining transfer function. Otherwise the split didn't actually disappear at all. It might be nicer to implement `Dom.add` which immediately accounts for `should_join` when adding instead of doing it in batch.

The added tests show, that this works for the sort of case that @michael-schwarz was worried about, where the expressions must necessarily be evaluated on the states after transfer functions but before the join of incoming edges.

### TODO
- [x] Add `UpdateExpSplit` emitting to all transfer functions (and add tests).
- [x] Test if it works when splitting by multiple expressions simultaneously (and add tests).
- [ ] ~~Check for performance overhead of all these extra events, expression evaluations and `join_reduce`s~~. Expsplit analysis is not activated by default, so makes no difference normally. Whether the events and evaluations are overly expensive when using it, we will find out when someone actually starts using this.


michael-schwarz:
I think it might make sense to merge master back into this and then aim to merge to PR soon! This is a useful feature which we are likely to want, and we should not let this get so stale it becomes hard to merge.

sim642:
Sure, I can bring this up to date. This implementation is very self-contained, so it won't cause maintenance issues even if we don't end up using this anytime soon.

src/analyses/expsplit.ml
```diff
@@ -0,0 +1,86 @@
...
+
+  let body ctx (f:fundec) =
+    emit_splits_ctx ctx
+
+  let return ctx (exp:exp option) (f:fundec) =
+    emit_splits_ctx ctx
+
+  let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc au =
+    let d = D.join ctx.local au in
+    emit_splits ctx d
```
sim642, line 54: I'm not sure how the interprocedurality of this should be. Right now the splitting expressions are passed to functions and are in contexts. They're also joined here on `combine`, in case the function added more splits. But this is practically quite limited because it only works on globals, because locals go out of scope and thus evaluate the same either way.

It might be more interesting with var_eq information to modify the splitting accordingly when going over function boundaries, but that can be a future experiment.

michael-schwarz (approved):
Nice job! I am really impressed by how clean this ended up being!

michael-schwarz:
Interestingly, there seems to be a 1:1 correspondence between this and the value partitioning feature of Frama-C's EVA:

Section 6.5.2, p 86ff
https://frama-c.com/download/frama-c-value-analysis.pdf

PR #498, opened 2021-12-15
Store and load analysis specific data in incremental run 

### LINKED ISSUE #496: Varinfos for allocated variables are not loaded in incremental analysis run
The `MallocWrapperAnalysis` uses functionality within `RichVarinfo` to keep track of identifiers (`varinfo`s) of allocated blocks. Within `RichVarinfo` a  hashtable is used to keep track of these `varinfo`s. This hashtable is erroneously currently not populated in an incremental run. 

This can lead to problems where the analysis expects to know whether a `varinfo` refers to an allocated block or not.

jerhard:
The different analyses offer `finalize` methods that might provide data for marshaling, and `init` methods where this data can be passed from a previous run for initialization. Previously, this data was only loaded / stored when the option `load_run` / `save_run` options were activated, which are intended for comparison of results between runs, not incremental analysis. 

With this PR, analysis data is also stored when `incremental.save` is activated and loaded when `incremental.load` is active.
Closes #496.

### DIFF
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -298,11 +298,15 @@ struct
 
     GU.global_initialization := true;
     GU.earlyglobs := get_bool "exp.earlyglobs";
-    if get_string "load_run" <> "" then (
-      Spec.init (Some (Serialize.unmarshal (Filename.concat (get_string "load_run") "spec_marshal")))
-    )
-    else
-      Spec.init None;
+    let marshal =
+      if get_string "load_run" <> "" then
+        Some (Serialize.unmarshal (Filename.concat (get_string "load_run") "spec_marshal"))
+      else if Serialize.results_exist () && get_bool "incremental.load" then
+        Some (Serialize.load_data Serialize.AnalysisData)
+      else
+        None
+    in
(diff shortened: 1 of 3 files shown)


sim642 (approved):
Did you check that mallocWrapper locations now nicely show up in the HTML view after `incremental.load`?
If so, do they also refer to the updated locations or still the old ones? If I remember correctly, we had two different functions for looking up locations with some differences in case of incremental analysis. Or is that differentiation now unnecessary that `malloc` variables and thread IDs refer to creation points by node, not by their original location?

jerhard:
>Did you check that mallocWrapper locations now nicely show up in the HTML view after incremental.load?
If so, do they also refer to the updated locations or still the old ones?

Yes, they do! The line that is shown in the HTML view shows the new line of the allocation, if the allocation site was shifted. That works because the `describe_varinfo` function looks up the location of the associated node in some map: 
https://github.com/goblint/analyzer/blob/84eda518eecbdb7073ea9e80684d3f8f823d281c/src/analyses/mallocWrapperAnalysis.ml#L18-L21

PR #500, opened 2021-12-16
Remove custom octagon analysis, make Apron Analysis answer MustBeEqual, MayBeEqual, and MayBeLess queries

jerhard:
As we have working Apron bindings, the custom implemented octagon analysis is now superfluous. Also, the OctagonAnalysis did not take different `ikind`s of integers into account, and was coupled with some parts of `IntDomain`, making changes there more costly to do.

Part of #460.


jerhard:
The test `23/13-advantage_for_last.c` for  was annotated to use octagons but didn't need them, so I simply removed the option.

There were two other tests for the octagon analysis together with partitioned arrays (`"22/10 array_octagon", "23/10 array_octagon"` ) and copies of these for the precision adaption (`"42/14 22_10-array_octagon", "42/20 23_10-array_octagon"`). I changed the test cases to use the Apron analysis instead of the Octagon analysis, but these test cases fail now due to imprecision.

jerhard:
Interestingly, the assert in line 56 fails, while the asserts in lines 54, 55 succeed, and the partitioned array knows that it is partitioned by `j`, and that the left part of the array has value `1`, and the middle part has value `2`. Maybe the apron analysis is somehow not queried by the partitioned arrays/the apron analysis does not answer?
https://github.com/goblint/analyzer/blob/c88666e84379ea2ab438209e5685f2829bca4d89/tests/regression/22-partitioned_arrays/10-array_octagon.c#L49-L56

jerhard:
I would suggest to annotate the tests in question with `SKIP` and create a separate issue about these. Any objections?

sim642:
I think the octagon analysis answers queries like `MustBeEqual` which partitioned arrays use, but the apron analysis doesn't implement them. It shouldn't be too hard to add support for them though.

jerhard:
@sim642  I extended the apron analysis to also answer the `MustBeEqual`, `MayBeEqual` and `MayBeLess` queries.
Most asserts in the both test cases now work, except for a few. And some asserts now can be shown that previously couldn't.

sim642:
Maybe @michael-schwarz knows what aspects the octagon analysis had that the apron analysis doesn't that it's not precise enough for that. Did the octagon analysis have something very specific for the partitioned arrays or maybe it was just more precise because it was unsound?

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -353,15 +353,34 @@ struct
...
+    | Queries.MustBeEqual (exp1,exp2) ->
+      let exp = (BinOp (Cil.Eq, exp1, exp2, TInt (IInt, []))) in
+      let must_eq = eval_int exp in
+      Option.default false (ID.to_bool must_eq)
+    | Queries.MayBeEqual (exp1,exp2) ->
+      (* We transform the may be equal into a must query to apron using this equivalence:
+         exp1 MayBeEqual exp2  <=> not (exp1 MustBeUnequal exp2 *)
+      let exp = (BinOp (Cil.Ne, exp1, exp2, TInt (IInt, []))) in
+      let must_neq = eval_int exp in
+      Option.map_default not true (ID.to_bool must_neq)
```
sim642, line 377: Would it not be possible to do without using `!=` in the expression (which isn't the best with Apron)?
Could do the same `==` expression, but look for the `Some false` (definitely not equal) result.

michael-schwarz:
> Did the octagon analysis have something very specific for the partitioned arrays or maybe it was just more precise because it was unsound?

No, it did not have anything very specific. IO think it completely ignored overflows, so maybe that is why?

tests/regression/22-partitioned_arrays/10-array_octagon.c
```diff
@@ -96,8 +96,10 @@ void example1(void) {
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 2); //UNKNOWN
   assert(z >= 0);
-  assert(z <= j);
-  assert(a[z] == 0); //FAIL
+  assert(z <= j); //UNKNOWN
+
+  // Used to work with the custom octagon analysis, does not work with apron octagons
+  assert(a[z] != 0); //UNKNOWN
```
michael-schwarz, line 102: The thing not working on arrays here is a direct result of the octagon analysis not knowing that `z <=j ` holds, which is a bit strange because that should be known.

tests/regression/22-partitioned_arrays/10-array_octagon.c
```diff
@@ -139,7 +141,9 @@ void example2(void) {
   // Values that may be read are 1 or 0
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 0); //UNKNOWN
-  assert(a[z] == 2); //FAIL
+
+  // Used to work with the custom octagon analysis, does not work with apron octagons
+  assert(a[z] != 2); //UNKNOWN
```
michael-schwarz, line 146: Same as above, for some reason apron does not know that `z >= j` here...

tests/regression/22-partitioned_arrays/10-array_octagon.c
sim642: With `--set sem.int.signed_overflow assume_none` it passes, to overflows are somehow the reason. I think it doesn't have good bounds for `j`, so `j - 1` might underflow and would by default be set to top.
I suppose we can just add that parameter to the tests where octagon analysis was previously more precise and see if that makes them all pass. I would think it should be able to.

tests/regression/19-spec/03-inter.c
```diff
@@ -0,0 +1,18 @@
...
+int readG(int x ){
+  return myglobal;
+}
+
+int main(void) {
+  int a = 43;
+  myglobal = 3;
+  int b = readG(a);
+  return 0;
+}
```
sim642, line 18: What is this spec test doing here? Doesn't seem related.

tests/regression/22-partitioned_arrays/10-array_octagon.c
```diff
@@ -1,5 +1,6 @@
-// PARAM: --set solver td3 --enable ana.int.interval --enable exp.partition-arrays.enabled  --set ana.activated "['base','threadid','threadflag','escape','expRelation','octagon','mallocWrapper']" --set exp.privatization none
+// PARAM: --set solver td3 --enable ana.int.interval --enable exp.partition-arrays.enabled  --set ana.activated "['base','threadid','threadflag','escape','expRelation','apron','mallocWrapper']" --set exp.privatization none
```
sim642, line 1: This and all the other tests that now use apron should be set up like the apron groups are: `SKIP`, but then in CI workflows explicitly test those groups.

tests/regression/24-octagon/03-previously_problematic_a.c
```diff
@@ -1,27 +0,0 @@
-// PARAM: --set solver td3 --enable ana.int.interval --enable exp.partition-arrays.enabled  --set ana.activated "['base','threadid','threadflag','expRelation','octagon','mallocWrapper']" --set exp.privatization none
-// These examples were cases were we saw issues of not reaching a fixpoint during development of the octagon domain. Since those issues might
-// resurface, these tests without asserts are included
```
sim642, line 3: Could these also not be just converted to apron analysis tests? Keeping them around wouldn't hurt in these cases where the relational analysis is invoked in a bunch of different ways.

src/analyses/apron/apronAnalysis.apron.ml
jerhard: Yes, that works, and also works for `MayBeLess`. I first tried to do something directly on the Apron API, and this seemed to only yield `true` in must cases, and false otherwise. With `eval_int`, this is not necessary; so I changed it accordingly.

tests/regression/22-partitioned_arrays/10-array_octagon.c
jerhard: I moved them to the `regression/36-apron` folder.

tests/regression/22-partitioned_arrays/10-array_octagon.c
jerhard: > I suppose we can just add that parameter to the tests where octagon analysis was previously more precise and see if that makes them all pass. I would think it should be able to.

I changed the tests accordingly.

tests/regression/24-octagon/03-previously_problematic_a.c
jerhard: Yes, I re-added the tests, changed them to use apron and modified the workflow to run them.

tests/regression/36-apron/100-array_octagon_keep_last_prec.c
```diff
@@ -68,7 +68,9 @@ void example1(void) {
   // Values that may be read are 1 or 2
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 2); //UNKNOWN
-  assert(a[z] == 0); //FAIL
+
+  // Relies on option sem.int.signed_overflow assume_none
+  assert(a[z] != 0); //UNKNOWN
```
michael-schwarz, line 73: Would this not be `FAIL` then, given that we now have ` --set sem.int.signed_overflow assume_none` ?

tests/regression/36-apron/100-array_octagon_keep_last_prec.c
```diff
@@ -110,7 +112,9 @@ void example2(void) {
   // Values that may be read are 1 or 0
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 0); //UNKNOWN
-  assert(a[z] == 2); //FAIL
+
+  // Relies on option sem.int.signed_overflow assume_none
+  assert(a[z] != 2); //UNKNOWN
```
michael-schwarz, line 117: Would this not be `FAIL` then, given that we now have ` --set sem.int.signed_overflow assume_none` ?

tests/regression/36-apron/97-array_octagon.c
```diff
@@ -53,7 +95,11 @@ void example1(void) {
   // Values that may be read are 1 or 2
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 2); //UNKNOWN
-  assert(a[z] == 0); //FAIL
+  assert(z >= 0);
+
+  // Relies on option sem.int.signed_overflow assume_none
+  assert(z <= j); //UNKNOWN
+  assert(a[z] != 0); //UNKNOWN
```
michael-schwarz, line 102: Would this not be `FAIL` then, given that we now have ` --set sem.int.signed_overflow assume_none` ?

tests/regression/36-apron/97-array_octagon.c
```diff
@@ -95,7 +141,9 @@ void example2(void) {
   // Values that may be read are 1 or 0
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 0); //UNKNOWN
-  assert(a[z] == 2); //FAIL
+
+  // Relies on option sem.int.signed_overflow assume_none
+  assert(a[z] != 2); //UNKNOWN
```
michael-schwarz, line 146: Would this not be `FAIL` then, given that we now have ` --set sem.int.signed_overflow assume_none` ?

tests/regression/36-apron/98-array_octagon_keep_last.c
```diff
@@ -53,7 +53,9 @@ void example1(void) {
   // Values that may be read are 1 or 2
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 2); //UNKNOWN
-  assert(a[z] == 0); //FAIL
+
+  // Relies on option sem.int.signed_overflow assume_none
+  assert(a[z] != 0); //UNKNOWN
```
michael-schwarz, line 58: Would this not be `FAIL` then, given that we now have ` --set sem.int.signed_overflow assume_none` ?

tests/regression/36-apron/98-array_octagon_keep_last.c
```diff
@@ -95,7 +97,9 @@ void example2(void) {
   // Values that may be read are 1 or 0
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 0); //UNKNOWN
-  assert(a[z] == 2); //FAIL
+
+  // Relies on option sem.int.signed_overflow assume_none
+  assert(a[z] != 2); //UNKNOWN
```
michael-schwarz, line 102: Would this not be `FAIL` then, given that we now have ` --set sem.int.signed_overflow assume_none` ?

tests/regression/36-apron/99-array_octagon_prec.c
```diff
@@ -69,7 +69,9 @@ void example1(void) {
   // Values that may be read are 1 or 2
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 2); //UNKNOWN
-  assert(a[z] == 0); //FAIL
+
+  // Relies on option sem.int.signed_overflow assume_none
+  assert(a[z] != 0); //UNKNOWN
```
michael-schwarz, line 74: Would this not be `FAIL` then, given that we now have ` --set sem.int.signed_overflow assume_none` ?

tests/regression/36-apron/99-array_octagon_prec.c
```diff
@@ -111,7 +113,9 @@ void example2(void) {
   // Values that may be read are 1 or 0
   assert(a[z] == 1); //UNKNOWN
   assert(a[z] == 0); //UNKNOWN
-  assert(a[z] == 2); //FAIL
+
+  // Relies on option sem.int.signed_overflow assume_none
+  assert(a[z] != 2); //UNKNOWN
```
michael-schwarz, line 118: Would this not be `FAIL` then, given that we now have ` --set sem.int.signed_overflow assume_none` ?

tests/regression/36-apron/100-array_octagon_keep_last_prec.c
jerhard: Not fail, but known, since I switched from `==` to `!=`.

sim642 (approved):

michael-schwarz (approved):

Issue #508, opened 2021-12-21
Make precision refinement compatible with reluctant destabilization


stilscher:
Reluctant destabilization takes the abstract value of the entry point of a changed function from the previous run to re-evaluate the changed function. With the precision refinement the new (increased) precision is projected at the entry edges. With reluctant destabilization this entry edge to a changed function will however never be reevaluated if there are no actual code changes in the changed function that affect the abstract value at its return point. This results in the fixpoint not being reached at the entry state of the changed function.

As a test case for this issue one can refer to `incremental/01-force-reanalyze/00-int` with additionally activated reluctant destabilization.

jerhard:
>With reluctant destabilization this entry edge to a changed function will however never be reevaluated if there are no actual code changes in the changed function that affect the abstract value at its return point. This results in the fixpoint not being reached at the entry state of the changed function.

To elaborate, the verifier complains about an unreached fixed-point, when the verifier itself is not incremental.  Then the verifier *does re-evaluate* the `enter` and thus obtains a state with a different set of active domains.

jerhard:
The idea to fix would be to also re-evaluate the call edges for calls to changed functions in the reluctant analysis.
One would utilize the `side_dep` set and now additionally remove the unknowns from `stable` that depend on any unknown relating to the return node of the changed functions. The `solve` would then be called on these unknowns. 

This will lead to some more evaluations in the reluctant setting, especially if there are many calls to the changed function. This leads to the question whether re-doing the `enter`/`combine` should be an option, or whether we can get around re-doing the `enter`/`combine` for all changed functions -- it really should only be needed for functions for which the activated int-domains or the context-information that is tracked has changed.

sim642:
Why not just exclude `force-reanalyze` functions from reluctant destabilization?

jerhard:
That actually seems like a good idea: When a function is re-analyzed with different int-domains, the resulting end state should never be `leq` compared to the old state. (Unless there are no ints in the abstract states of the function, in which case activating different int-domains would be meaningless.)

jerhard:
Closing this as #600 is merged into `incremental`.

PR #509, opened 2021-12-21
Better interactive tests

stilscher:
This PR extends the regression test script for incremental tests to also handle patch files that contain changes in multiple files. This can be used to encode changes in the configuration for the incremental run. An instance for which this is needed is the implemented test case that targets the forceful re-evaluation of functions with an increased precision.

Due to the mentioned test case we noticed that the precision refinement is not compatible with the reluctant destabilization yet. Therefore I deactivated this option in the test case and created a new issue for this (#508).

### DIFF
--- a/scripts/test-incremental.sh
+++ b/scripts/test-incremental.sh
@@ -15,11 +15,11 @@ args="--enable dbg.debug --enable printstats -v"
 
 ./goblint --conf $conf $args --enable incremental.save $source &> $base/$test.before.log
 
-patch -b $source $patch
+patch -p0 -b <$patch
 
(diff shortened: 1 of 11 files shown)


michael-schwarz (approved):

sim642:
Maybe we should document somewhere, how these new patches have to be created to have the exact multiple-file format.

stilscher:
That is probably a good idea. I could write a paragraph about the incremental tests in the _Developer Guide -> Testing_ chapter of Goblint's documentation?

sim642:
> I could write a paragraph about the incremental tests in the _Developer Guide -> Testing_ chapter of Goblint's documentation?

That would be a good place indeed.

PR #513, opened 2022-01-01
Cleanup `nestra` tests

michael-schwarz:
This moves the `nestra` tests that were unused and in a separate folder in the `tests` directory into a new category of regression tests.

I added assertions where appropriate such that the tests actually test something as well.

Does anyone recall why they are called `nestra`? Is it named after Härmel Nestra, who as Helmut told me worked on the Trier analyzer for a while? If yes, should we think of a less confusing name? E.g. `misc`?

Part of #460


vesalvojdani:
Well, "misc" is not that much more informative... :P  We could call them `trier-analyzer-regression` to denote that these were tests that the Trier analyzer could handle. More consistent with our non-existence naming conventions would be just `44-trier_analyzer`.

tests/regression/44-trier_analyzer/02-abc.c
```diff
@@ -1,5 +1,8 @@
 #include <pthread.h>
-#include "/home/psi/nestra/kodu/mat/inf/p/teor/anal/DAEDALUS/tmp/errors.h"
+
+void err_abort(int i, char* ptr) {
+  exit(0);
+}
```
sim642, line 5: Looks like this `err_abort` from `errors.h` was likely this macro: https://github.com/thlorenz/threads/blob/cbe6d187b82cf0a6be05245799ea02ef0a501d40/src/prog-with-posix-threads/errors.h#L38-L42. This substitute function is good though.

tests/regression/44-trier_analyzer/20-Pchain.c
```diff
@@ -2,10 +2,10 @@
 
 extern void *malloc(size_t);
 extern int putchar(int);
-extern int printf();
```
sim642, line 5: What's the reason for removing this declaration here, but not the others or elsewhere?

I'm now thinking maybe it'd be a good idea to replace all these extern standard library definitions with actual includes to avoid any possible conflicts with types or other standard library internal trickery.

tests/regression/44-trier_analyzer/27-rec4.c
```diff
@@ -12,7 +12,9 @@ void rec (int **u, int i) {
   else {
     i++;
     u = (int **)malloc(sizeof(int *));
+    int** v = u;
     rec(u, i);
+    assert(v == u); //UNKNOWN
```
sim642, line 17: Should this also be `UNKNOWN!`?

tests/regression/44-trier_analyzer/34-S0.c
```diff
@@ -20,6 +20,7 @@ main () {
   init(&c, 1, 4);
   init(&d, 2, 2);
   c = d;
-  d.im = -3;  
+  d.im = -3;
+  assert(d.re == 1); // TODO (float)
```
sim642, line 24: I guess it's fine as a TODO, although I'm actually unsure if this floating-point equality is even guaranteed in the concrete.

tests/regression/44-trier_analyzer/41-strcpy.c
```diff
@@ -8,5 +8,9 @@ main () {
   strcpy (s, "Hello, strings!\n");
   strcpy (t, "Bye-bye!\n");
   y = strcpy (s, t);
+
+  assert(*y == 'B'); //TODO
+  char c = *y; //TODO
```
sim642, line 13: What's the second TODO about? There's no assertion.

tests/regression/44-trier_analyzer/34-S0.c
michael-schwarz: I would assume this equality also to hold in the concrete, as there is no computation going on at all, and however `1.0` is represented should be the same regardless of whether the constant comes from the assignment and in the comparison.
(`c == c` should hold for constants `c` that are not `NaN`)

tests/regression/44-trier_analyzer/41-strcpy.c
michael-schwarz: Probably left over from some intermediate state or something.

tests/regression/44-trier_analyzer/27-rec4.c
michael-schwarz: Why?
`u` is a local variable right, so the formal `u` of the callee is assigned the pointer the caller allocated. 
In the recursive call, `u` is assigned a new value (a fresh pointer in this case), but that is a simple assignment to a local variable that does not affect the value in the caller. 
The case where it is unknown would be if `*u` was assigned. Or am I missing something?

tests/regression/44-trier_analyzer/02-abc.c
michael-schwarz: Yeah, that seems to be it. I switched it back.

tests/regression/44-trier_analyzer/27-rec4.c
sim642: I didn't actually investigate the program, but am just wondering about the clarity of our annotations going forward. It should either be `UNKNOWN!` if it's unknown in concrete, `TODO` if it's something we might want to handle in the future or `UNKNOWN (explanation)` if it's intended to be less precise than ideal. A simple `UNKNOWN` is ambiguous about the intent.

michael-schwarz: We still have `//UNKNOWN` all over our tests, but I guess I can change it to `//TODO` here.

sim642 (approved):

PR #516, opened 2022-01-03
Categorize `exp` options

michael-schwarz:
As a part of #28, we decided we want to somewhat cleanup our options:

- [x] Move witness options from `exp.witness.*`  to ` witness.*`
- [x] Move solver options from `exp.solver.*` to `solvers.*`

What other option should be moved out of `exp`?

I could, e.g., see:
- [x] moving `exp.malloc.fail` and `exp.failing_locks` to `sem`
- [x] moving `exp.malloc.wrappers` to `ana.malloc.wrappers`
- [x] moving `exp.partition-arrays.*` to `ana.base.partition-arrays`
- [x] moving `exp.structs.*`to `ana.base.structs.*`
- [x] moving `exp.privatization` to `ana.base.privatization`
- [x] moving  `exp.apron.privatization` to `ana.apron.privatization`
- [x] moving `exp.apron.priv.*` to `ana.apron.priv.*`

WDYT?


sim642:
Regarding all the others you've listed, I think they all sound good! I think if we do all of them it's already a big enough batch of option renames for this PR.

src/util/options.schema.json
```diff
@@ -804,6 +938,34 @@
...
+              "description":
+                "Consider the case where malloc or calloc fails.",
+              "type": "boolean",
+              "default": false
+            }
+          },
+          "additionalProperties": false
+        },
+        "locks": {
+          "title": "sem.locks",
```
sim642, line 957: Maybe `sem.lock` would be a bit better name.

sim642 (approved):

PR #517, opened 2022-01-04
Remove SolverInteractiveWGlob

michael-schwarz:
This removes the `SolverInteractiveWGlob` following our discussion at GobCon.

References #460


sim642:
What's the g2html submodule change about? That repository doesn't have the new commit upstream, did you forget to push that?

michael-schwarz:
It passes commands for the interactive mode, so i removed that as well. Pushed it now

sim642:
Wow, g2html even contains some kind of web server that takes input from some HTTP requests and passes the respective signals to the interactive solver: https://github.com/goblint/g2html/blob/f4c35689508b841b4ed518864e1e4d19df934c50/src/g2html/GoblintServer.java#L117-L134. Now I'm a bit tempted to even try to run it that way to see what it looks like (if it works at all, because the interactive solver stuff could likely be broken since nobody has been using it).

sim642:
Left a comment on the g2html change: https://github.com/goblint/g2html/commit/b8cf8b6ef625685b349cd8a9e58befd4ad5b821a#r63011361. Other than that looks good!

michael-schwarz:
I now removed the unused import from `g2html`.

PR #518, opened 2022-01-05
Use MHP to exclude races & Refactoring MHP

michael-schwarz:
This now uses the MHP information we compute to exclude more races.
To this end, at each access, we record a tuple of MHP information along with the access.

It also refactors some of our more-traces code by moving MHP to a separate module.

The code around race warnings should probably be looked over carefully, the PR contains a first step towards moving back to pair-wise race warnings.


src/domains/access.ml
```diff
@@ -263,20 +262,21 @@ let get_val_type e (vo: var_o) (oo: off_o) : acc_typ =
...
   if is_ignorable lv then () else begin
+    Printf.printf "adding %s\n" (ThreadIdDomain.ThreadLifted.show mhp.tid);
     some_accesses := true;
     let tyh = TypeHash.find_def accs  ty (lazy (LvalOptHash.create 10)) in
     let lvh = LvalOptHash.find_def tyh lv (lazy (PartOptHash.create 10)) in
     let loc = !Tracing.current_loc in
     let add_part ls =
       PartOptHash.modify_def lvh (Some(ls)) (lazy (Set.empty,lp)) (fun (s,o_lp) ->
-          (Set.add (conf, w,loc,e,lp) s, LSSet.inter lp o_lp)
+          (Set.add (conf,mhp, w,loc,e,lp) s, LSSet.inter lp o_lp)
```
sim642, line 274: Now this polymorphic `Set` contains some non-primitive data structures that make up `MHP.t`, including thread IDs in non-string forms. It would probably be safer to replace this `Set` with one constructed using `Set.Make`, which actually delegates to (currently non-existent) `MHP.compare`, etc. Using `[@@deriving ord]` hopefully makes this easy enough.

src/domains/access.ml
```diff
@@ -455,23 +455,30 @@ let check_accs (prev_r,prev_lp,prev_w) (conf,w,loc,e,lp) =
...
+  if (not w) && (not w2) then
+    None (* two read/read accesses do not conflict *)
+  else if not (LSSet.is_empty @@ LSSet.inter lp lp2) then
+    None (* the labelled string set excludes that these conflict *)
+  else if not (MHP.may_happen_in_parallel mhp mhp2) then
+    None (* They may not be in parallel *)
+  else
+    Some (max conf conf2)
+
+let check_safe ls (accs,_) prev_safe =
```
sim642, line 469: I now remember that this code is notably refactored in #397, which is currently only part of the interactive branch for interactive race warnings. I'll have to check exactly how big the conflicts would be, but in that form the whole `MHP.t` stuff also needs to be a `Lattice.S` to be usable in the global invariant.

src/domains/access.ml
```diff
@@ -455,23 +455,30 @@ let check_accs (prev_r,prev_lp,prev_w) (conf,w,loc,e,lp) =
     (new_r, new_lp, new_w)
   | _ -> (prev_r,prev_lp,prev_w)
 
-let check_safe ls (accs,lp) prev_safe =
+(* Check if two accesses race and if yes with which confidence *)
+let conflict2 (conf,mhp,w,loc,e,lp) (conf2,mhp2,w2,loc2,e2,lp2) =
+  if (not w) && (not w2) then
+    None (* two read/read accesses do not conflict *)
+  else if not (LSSet.is_empty @@ LSSet.inter lp lp2) then
+    None (* the labelled string set excludes that these conflict *)
```
sim642, line 463: Is this not duplicating some of the logic in `check_accs`?

src/domains/access.ml
```diff
@@ -455,23 +455,30 @@ let check_accs (prev_r,prev_lp,prev_w) (conf,w,loc,e,lp) =
...
   if ls = None then
     prev_safe
   else
-    let ord_enum = Set.backwards accs in (* hope that it is not nil *)
-    let lp_start = (fun (_,_,_,_,lp) -> lp) (BatOption.get (BatEnum.peek ord_enum)) in
-    (* ignore(printf "starting with lockset %a\n" LSSet.pretty lp_start); *)
-    match BatEnum.fold check_accs (None, lp_start, false) (Set.backwards accs), prev_safe with
-    | (None, _,_), _ ->
-      (* ignore(printf "this batch is safe\n"); *)
+    let cart = BatSet.cartesian_product accs accs in
```
sim642, line 473: I think this cartesian product currently includes unnecessarily many pairs, even though by symmetry only half should be needed. Also pairs with itself should be unnecessary, at least they were never checked in the old linear fold code.

src/domains/access.ml
```diff
@@ -455,23 +455,30 @@ let check_accs (prev_r,prev_lp,prev_w) (conf,w,loc,e,lp) =
...
+    let conf = BatSet.filter_map (fun (x,y) -> conflict2 x y) cart in
+    if BatSet.cardinal conf = 0 then
```
sim642, line 475: So just `is_empty`?

src/domains/access.ml
sim642: Also, I'm a bit confused why this cartesian product is here, because below `check_safe` is still folded linearly over some set of accesses. Isn't this now doing O(n^2) work _for each access_, totalling O(n^3) overall if accounting to that outer fold?

sim642 (commented):
Here's something I've wondered before: could there be any way to entirely get rid of the (labeled) string stuff from the accesses and replace them with something more type-safe (allowing for additional operations in pairwise race checking)?
Because this now adds a history-thread-IDs-specific component to the entire access structure in addition to the strings. It's not very flexible if this has to be done for every other analysis-specific race check we might have.

src/domains/access.ml
sim642: I think we're not yet at the point to merge #391 entirely, but I suppose I could merge #397 also to master now to allow those conflicts to be solved already here instead of leaving more conflicts to be merged for #391.

michael-schwarz: Yes, I think that makes sense!

src/domains/access.ml
michael-schwarz: Yup.

src/domains/access.ml
michael-schwarz: The way I understood it, the fold is over the different regions (or partitions) and the pair-wise race check than happens inside such a region. So if we have `m` regions of `n` accesses each, we'd have `O(m*n^2)`, right?

michael-schwarz: > Also pairs with itself should be unnecessary

I don't think so. Consider e.g. a global that is accessed only from one(!) thread with a non-unique thread id.

> at least they were never checked in the old linear fold code.

src/cdomains/mHP.ml
```diff
@@ -0,0 +1,37 @@
...
+    if ConcDomain.ThreadSet.is_top created then
+      false
+    else
+      not @@ ConcDomain.ThreadSet.exists (ident_or_may_be_created) created
+
+(** Must the thread with thread id other be already joined  *)
+let must_be_joined other joined =
+  try
+    List.mem other (ConcDomain.ThreadSet.elements joined)
+  with _ -> false
```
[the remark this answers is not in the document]
michael-schwarz: :+1:

src/domains/access.ml
michael-schwarz: Done!

src/domains/access.ml
sim642: > This is true, but I am not so sure how to elegantly do such a half-product (for lack of a better word).

If the elements have a `compare` then one could just filter the pairs with `fun (x, y) -> E.compare x y <= 0` or something. I think there are other places where I've done it like that.
Of course that's not optimal either because it involves first constructing all the pairs, but at least it avoids duplicating the possibly more expensive checks and warning generations.

src/domains/access.ml
sim642: Actually, it probably would just have to be `Printable.S`, not `Lattice.S`, since it's either only used in the variables or some set elements.

src/domains/access.ml
sim642: After working on #519 I remembered that that already seems to have a proper `SetDomain` for these anyway.

src/domains/access.ml
sim642: > Consider e.g. a global that is accessed only from one(!) thread with a non-unique thread id.

You're probably right about that. Since we're going to make it more pairwise anyway, I cannot be bothered to figure out exactly how that worked out before. Probably just because such accesses would have no locks nor thread ID, so some set is empty and hence trivially racing.

michael-schwarz:
> Here's something I've wondered before: could there be any way to entirely get rid of the (labeled) string stuff from the accesses and replace them with something more type-safe (allowing for additional operations in pairwise race checking)?

I have also given this some thought, but I did not find any immediately obvious solution: What one would want is some `RaceInformation` module.

~~~OCaml
module type RaceInformation = sig
  type t = MHP of MHP.t | Lockset of unit | ...
  val may_race: t -> t -> bool
end
~~~

One would then collect sets of such things and the check for two such sets if there exists `a` and `b` such that `may_race a b` is false.

However, we do not really win much there as this `RaceInformation` thing would still have to explicitly have all possible values that could be the cause for such an exclusion....

So that doesn't seem to be it either.

src/domains/access.ml
```diff
@@ -356,29 +356,29 @@ and distribute_access_exp f w r c = function
...
 struct
   include Printable.Std
-  type t = int * bool * CilType.Location.t * CilType.Exp.t * LSSet.t [@@deriving eq, ord]
+  type t = int * MHP.t * bool * CilType.Location.t * CilType.Exp.t * LSSet.t [@@deriving eq, ord]
 
-  let hash (conf, w, loc, e, lp) = 0 (* TODO: never hashed? *)
+  let hash (conf, mhp, w, loc, e, lp) = 0 (* TODO: never hashed? *)
 
-  let pretty () (conf, w, loc, e, lp) =
+  let pretty () (conf, mhp, w, loc, e, lp) = (* TODO:Print MHP? *)
```
sim642, line 381: I think this `pretty` isn't really used for user output, but I am wondering of the MHP information should somehow be in the `print_accesses` below, because it could be confusing to debug if certain races are excluded by information which isn't shown.

src/domains/access.ml
```diff
@@ -421,35 +421,31 @@ end
...
-      (* ignore(printf "race with %d and %d \n" n m); *)
-      Some (max n m)
-    | (Some n,_,_), None ->
-      (* ignore(printf "race with %d\n" n); *)
-      Some n
+    else
+      let maxconf = List.max conf in
+      match prev_safe with
+      | None -> Some maxconf
+      | Some prev -> Some (max maxconf prev)
```
sim642, line 448: Looks fine to me, although maybe @vesalvojdani should also have a look before we change this critical race detection code.

sim642 (approved):

src/domains/access.ml
michael-schwarz: I added printing here now.

vesalvojdani (approved):
These code changes look good. We have quite many thread analyses now.... `thread`, `threadflag` and `threadJoins`.

PR #519, opened 2022-01-06
Refactor race warnings (accesses) to use global invariant on master

michael-schwarz:
This is merging #397 into master, such that conflicts with #518 can be resolved there instead of having to be resolved together with all the changes originating from the interactive analysis.

>I think we're not yet at the point to merge #391 entirely, but I suppose I could merge #397 also to master now to allow those conflicts to be solved already here instead of leaving more conflicts to be merged for #391.

_Originally posted by @sim642 in https://github.com/goblint/analyzer/pull/518#discussion_r779482123_


sim642:
@michael-schwarz Since you opened this PR, it seems like GitHub doesn't let me request a review from you, but I think I managed to get this ready for master. I also cherry-picked a bunch of other commits and changes from interactive surrounding this, hopefully didn't miss anything important, because this also now contains the change of using non-`varinfo` global constraint variables for the access analysis. That wasn't part of #397 originally.

sim642:
Hmm, 00-basic/06-threadid from the interactive tests does fail now with the incremental run detecting a race:
```
[Warning][Race] Memory location myglobal@06-threadid.c:4:5-4:13 (race with conf. 110):
  write with {thread:([t_fun@./tests/incremental/00-basic/06-threadid.c:21:3-21:40, main], {})} (conf. 110)  (exp: & myglobal) (./tests/incremental/00-basic/06-threadid.c:7:3-7:14)
  write with {thread:([t_fun@./tests/incremental/00-basic/06-threadid.c:24:3-24:40, main], {})} (conf. 110)  (exp: & myglobal) (./tests/incremental/00-basic/06-threadid.c:7:3-7:14)
```
I think the problem with that is that without restarting globals (which is done on interactive), the set of accesses that's now in the global invariant doesn't get cleared and thus the access with the old thread ID remains in the set and appears to race with the new one.

Not sure if anything can be done about it before #391 itself is merged, so maybe this just needs to be skipped for the time being?

src/analyses/accessAnalysis.ml
```diff
@@ -0,0 +1,231 @@
...
+      let d =
+        let open Access in
+        PM.singleton ls_opt (
+          AS.singleton (conf, w, loc, e, lp)
+        )
+      in
+      ctx.sideg (lv_opt, ty) d
+    )
+    else
+      ctx.sideg (lv_opt, ty) (G.bot ()) (* HACK: just to pass validation with MCP DomVariantLattice *)
```
michael-schwarz, line 49: Why is this needed? would these not the default value `bot` anyway?

michael-schwarz:
Yes, let's just mark the failing test as `//SKIP` and then I'd say we're good to merge (after addressing my two questions above).

src/analyses/accessAnalysis.ml
sim642: The problem is that they're different bottoms. If nothing is side-effected to these variables during the actual solve, then they have the bottom from the `Lift2` of local and global domains, not the bottom of `DomVariantLattice` or this `G`.

This forces all the lifted layers to be created such that during verify the hacky `G.leq` from above is actually called. Otherwise we'd just be calling the `leq` for the outermost `Lift2`, which we don't want to allow such verification bypass in and everything would fail verification.

This is the unfortunate side effect of only wanting to collect accesses into the global invariant during postsolving, but still pass the verification, which happens during the same postsolving.

sim642:
I didn't investigate it further, but I suspect this failure might be just due to not having incremental warnings here, so even without any changes, postsolving reevaluates everything once and side-effects something spurious (which isn't a verification error because of the `G.leq` hack of the access analysis) that the incremental warnings do not.

PR #520, opened 2022-01-08
Analysis of `pthread_cond`

michael-schwarz:
Pthreads uses condition variables or synchronization (e.g. `signal`,`broadcast`,`wait`, ...). So far we do not really do anything meaningful for them, however there are some possible uses:

- **Detecting more Dead Code**
  - The most trivial case is that we can determine that a condition variable `cv` is never signaled, making everything past `pthread_cond_wait(cv)` dead.
  - The threads that can signal `cv` are all:
     - (transitively) started by the current thread after its `pthread_cond_wait(cv)`.
     - (transitively) joined before this `pthread_cond_wait(cv)` is reached, thus meaning  they can not be the signaler
     
- **Improving Race Detection**
   - Combined with our MHP information, one can exclude races for threads `t1` and `t2` where `t2` has observed signal  (Must-Received) `cv` and all the threads signalling `cv` are not started at this location in `t1` yet. If one additionally collects May-Sent-Signal sets, one can also exclude cases where `t1` is the one that will signal but has not reached that program point yet.
   
- **Improving Privatization**
  -  Investigate to what extent privatization can benefit from this, I have not yet found a convincing use case there.   
   
Things that are TODO here:

- [x] ~Raising `Deadcode` during `special` for `pthread_cond_wait` does not work, probably related to the `split` and `event` business in `mutexAnalysis.ml`. Any ideas there @sim642?~
- [x] Cleanup MHP code when #518 lands
- [x] ~Implement the Race Detection part of this PR~ 
- [x] ~Dream up improved Privatization schemes~


sim642:
> * Raising `Deadcode` during `special` for `pthread_cond_wait` does not work, probably related to the `split` and `event` business in `mutexAnalysis.ml`. Any ideas there @sim642?

You would just conditionally have to not make the `ctx.split` calls. Then no splits are created and `raise Deadcode` makes it truly dead.

michael-schwarz:
> Improving Race Detection

Actually, I am starting to wonder if improving the Race Detection with this information is actually something that we want to do. The standard allows for spurious wakeups, so if a race is only excluded due to these conditions on condition variables, the race is still there in fact (!).

So maybe this information should be computed, but the race still reported with something such as (condition variable `v` does not prevent races due to spurious wakeup).

And for the reporting of code as dead, probably a warning should be generated that says only alive due to spurious wakeup or something... At least there should be an option to toggle how we handle spurious wakeups.

michael-schwarz:
I would propose to leave the TODOs as is for now, as we have thus far not come up with a convincing use case yet.

We can already detect an interesting class of bugs with this analysis, namely code that is only alive due to spurious wakeup, which already is an interesting concurrency bug.

src/analyses/libraryFunctions.ml
```diff
@@ -30,6 +30,10 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
 let pthread_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("pthread_create", special [__ "thread" [w]; drop "attr" [r]; __ "start_routine" [s]; __ "arg" []] @@ fun thread start_routine arg -> ThreadCreate { thread; start_routine; arg }); (* For precision purposes arg is not considered accessed here. Instead all accesses (if any) come from actually analyzing start_routine. *)
     ("pthread_exit", special [__ "retval" []] @@ fun retval -> ThreadExit { ret_val = retval }); (* Doesn't dereference the void* itself, but just passes to pthread_join. *)
+    ("pthread_cond_signal", special [__ "cond" []] @@ fun cond -> Signal cond);
+    ("pthread_cond_broadcast", special [__ "cond" []] @@ fun cond -> Broadcast cond);
+    ("pthread_cond_wait", special [__ "cond" []; __ "mutex" []] @@ fun cond mutex -> Wait {cond; mutex});
+    ("pthread_cond_timedwait", special [__ "cond" []; __ "mutex" []; __ "abstime" []] @@ fun cond mutex abstime -> TimedWait {cond; mutex; abstime});
```
michael-schwarz, line 36: @sim642 I would appreciate your feedback on the switching to the new library mechanism here.

sim642: The last line should have `"abstime" [r]` because that function reads the time struct through the pointer, other than that these look right!

This reminds me, sometime ago I started doing some cleanup, including these cond waits: https://github.com/goblint/analyzer/commit/6fb277fc75dd00a3e51df045f5a9e2594838d708#diff-f11dd379df0d8dcedf47eec4569e585bdf20a8671f9194a6cdfd69e65ce0c4e5R33-R34. I should probably revive that.

src/analyses/pthreadSignals.ml
```diff
@@ -0,0 +1,109 @@
...
+        let module TID = ThreadIdDomain.FlagConfiguredTID in
+        let not_self_signal = (not (TID.is_unique tid)) || (not (TID.equal tid other)) in
+        let may_be_running () =
+          not @@ MHP.definitely_not_started (tid, ctx.ask Queries.CreatedThreads) other
+        in
+        let may_not_be_joined () =
+          not @@ MHP.must_be_joined other (ctx.ask Queries.MustJoinedThreads)
+        in
+        not_self_signal && may_be_running () && may_not_be_joined ()
+      in
```
sim642, line 80: Is this not practically the same as this check in `ApronPriv`/`CommonPriv`? https://github.com/goblint/analyzer/pull/789/files#diff-8110758eb49bb6b89ebc1da162d30a7e5e46ad3fb34639830982a68ca91951c4R257-R268
If so, then I guess we should merge #789 first and adapt this to avoid duplication.

src/analyses/pthreadSignals.ml
```diff
@@ -0,0 +1,109 @@
...
+      (match ctx.ask CurrentThreadId with
+       | `Lifted tid ->
+         (match possible_vinfos (Analyses.ask_of_ctx ctx) cond with
+          | [a] ->
+            (* Only report if there is one definite condition variable for which wait is called *)
```
sim642, line 85: Doesn't it naturally generalize to larger points-to sets? If `for_all` there's no signaller, then warn?
I suppose the only singleton thing is the `Signals.add` below.

src/analyses/pthreadSignals.ml
```diff
@@ -0,0 +1,109 @@
...
+              (M.warn ~category:Deadcode "The condition variable %s is never signalled concurrently, succeeding code is live due to spurious wakeups only!" a.vname; ctx.local)
+          | _ -> ctx.local)
+       | _ -> ctx.local)
+    | TimedWait _ ->
+      (* Time could simply have elapsed *)
+      ctx.local
+    | _ -> ctx.local
+
+  let startstate v = Signals.empty ()
+  let threadenter ctx lval f args = [ctx.local]
```
sim642, line 103: Should the created thread have the same must-received signals set as the parent thread, not empty?
Also, nothing seems to even be looking up the must-received signals from the local state. The warning stuff is based on the thread globals.

src/analyses/pthreadSignals.ml
michael-schwarz: That's true!

src/analyses/pthreadSignals.ml
michael-schwarz: It is correct like this: The meaning is that somewhere in the local thread a `signal` for that mutex must have been called (because it has been received).
Currently, we are not using this information for anything, but it is very cheap to track and I think might be of interest to users.

src/analyses/pthreadSignals.ml
michael-schwarz: Yes, it generalizes naturally, I'll add it here.

michael-schwarz:
I switched the analysis to use the full power of MHP. This is both more powerful, and allows removing a lot of duplicated code.

I made the assumption that only few signals typically happen in programs for a given condition variable and that correspondingly the sets of MHP information at globals do not get too large.
If this turns out to be problematic, one could go for HoareSet of MHP information.

src/analyses/pthreadSignals.ml
```diff
@@ -0,0 +1,113 @@
...
+      in
+      let open Signalled in
+      let add_if_singleton conds = match conds with | [a] -> Signals.add (ValueDomain.Addr.from_var a) ctx.local | _ -> ctx.local in
+      let conds = possible_vinfos (Analyses.ask_of_ctx ctx) cond in
+      (match List.fold_left (fun acc cond -> can_be_signalled cond ||| acc) Never conds with
+       | PossiblySignalled -> add_if_singleton conds
+       | NotConcurrently ->
+         (M.warn ~category:Deadcode "The condition variable(s) pointed to by %s are never signalled concurrently, succeeding code is live due to spurious wakeups only!" (Basetype.CilExp.show cond); ctx.local)
+       | Never ->
+         (M.warn ~category:Deadcode "The condition variable(s) pointed to by %s are never signalled, succeeding code is live due to spurious wakeups only!" (Basetype.CilExp.show cond); ctx.local)
```
sim642, line 98:
```suggestion
         (M.warn ~category:Deadcode "The condition variable(s) pointed to by %a are never signalled concurrently, succeeding code is live due to spurious wakeups only!" Basetype.CilExp.pretty cond; ctx.local)
       | Never ->
         (M.warn ~category:Deadcode "The condition variable(s) pointed to by %a are never signalled, succeeding code is live due to spurious wakeups only!" Basetype.CilExp.pretty cond; ctx.local)
```

sim642 (approved):

PR #522, opened 2022-01-10
Server mode

keremc:
Implement a server mode. It uses the [jsonrpc](https://opam.ocaml.org/packages/jsonrpc/) library for processing messages. [rpclib](https://opam.ocaml.org/packages/rpclib/) was another viable alternative, but `jsonrpc` has no dependencies and is `yojson`-compatible.

Example usage:
```
$ ./goblint --enable server.enabled example.c
```

Example input:
```json
{"jsonrpc":"2.0","id":0,"method":"analyze","params":{}}
```

Example output:
```json
{"id":0,"jsonrpc":"2.0","result":null}
```

### DIFF
--- a/src/incremental/serialize.ml
+++ b/src/incremental/serialize.ml
@@ -44,13 +44,15 @@ let type_to_file_name = function
 
 (** Used by the server mode to avoid serializing the solver state to the filesystem *)
 let server_solver_data : Obj.t option ref = ref None
+let server_analysis_data : Obj.t option ref = ref None
 
 (** Loads data for incremental runs from the appropriate file *)
 let load_data (data_type: incremental_data_kind) =
   if server () then
     match data_type with
     | SolverData -> !server_solver_data |> Option.get |> Obj.obj
-    | _ -> raise (Invalid_argument "Can only load solver data")
+    | AnalysisData -> !server_analysis_data |> Option.get |> Obj.obj
+    | _ -> failwith "Can only load solver and analysis data"
   else
     let p = Filename.concat (gob_results_dir ()) (type_to_file_name data_type) in
     unmarshal p
@@ -60,6 +62,7 @@ let store_data (data : 'a) (data_type : incremental_data_kind) =
   if server () then
     match data_type with
     | SolverData -> server_solver_data := Some (Obj.repr data)
+    | AnalysisData -> server_analysis_data := Some (Obj.repr data)
     | _ -> ()
   else (
     ignore @@ Goblintutil.create_dir (gob_directory ());
--- a/src/incremental/serialize.ml
+++ b/src/incremental/serialize.ml
@@ -16,6 +16,8 @@ let gob_results_dir () =
 let gob_results_tmp_dir () =
   Filename.concat (gob_directory ()) results_tmp_dir
 
+let server () = GobConfig.get_bool "server"
+
 let marshal obj fileName  =
   let chan = open_out_bin fileName in
   Marshal.output chan obj;
@@ -38,22 +40,35 @@ let type_to_file_name = function
   | CilFile -> cil_file_name
   | VersionData -> version_map_filename
 
+let solver_data : Obj.t option ref = ref None
+
 (** Loads data for incremental runs from the appropriate file *)
 let load_data (data_type: incremental_data_kind) =
-  let p = Filename.concat (gob_results_dir ()) (type_to_file_name data_type) in
-  unmarshal p
+  if server () then
+    match data_type with
+    | SolverData -> !solver_data |> Option.get |> Obj.obj
+    | _ -> raise (Invalid_argument "Can only load solver data")
+  else
+    let p = Filename.concat (gob_results_dir ()) (type_to_file_name data_type) in
+    unmarshal p
 
 (** Stores data for future incremental runs at the appropriate file, given the data and what kind of data it is. *)
-let store_data (data: 'a) (data_type: incremental_data_kind) =
-  ignore @@ Goblintutil.create_dir (gob_directory ());
-  let d = gob_results_tmp_dir () in
-  ignore @@ Goblintutil.create_dir d;
-  let p = Filename.concat (d) (type_to_file_name data_type) in
-  marshal data p
+let store_data (data : 'a) (data_type : incremental_data_kind) =
+  if server () then
+    match data_type with
+    | SolverData -> solver_data := Some (Obj.repr data)
+    | _ -> ()
+  else (
+    ignore @@ Goblintutil.create_dir (gob_directory ());
+    let d = gob_results_tmp_dir () in
+    ignore @@ Goblintutil.create_dir d;
+    let p = Filename.concat d (type_to_file_name data_type) in
+    marshal data p)
 
 (** Deletes previous analysis results and moves the freshly created results there.*)
 let move_tmp_results_to_results () =
-  if Sys.file_exists (gob_results_dir ()) then begin
-    Goblintutil.rm_rf (gob_results_dir ());
-  end;
-  Sys.rename (gob_results_tmp_dir ()) (gob_results_dir ())
+  if not (server ()) then (
+    if Sys.file_exists (gob_results_dir ()) then begin
+      Goblintutil.rm_rf (gob_results_dir ());
+    end;
+    Sys.rename (gob_results_tmp_dir ()) (gob_results_dir ()))
--- a/src/incremental/serialize.ml
+++ b/src/incremental/serialize.ml
@@ -16,7 +16,7 @@ let gob_results_dir () =
 let gob_results_tmp_dir () =
   Filename.concat (gob_directory ()) results_tmp_dir
 
-let server () = GobConfig.get_string "server" <> ""
+let server () = GobConfig.get_bool "server.enabled"
 
 let marshal obj fileName  =
   let chan = open_out_bin fileName in
--- a/src/incremental/serialize.ml
+++ b/src/incremental/serialize.ml
@@ -16,7 +16,7 @@ let gob_results_dir () =
 let gob_results_tmp_dir () =
   Filename.concat (gob_directory ()) results_tmp_dir
 
-let server () = GobConfig.get_bool "server"
+let server () = GobConfig.get_string "server" <> ""
 
 let marshal obj fileName  =
   let chan = open_out_bin fileName in
--- a/src/incremental/serialize.ml
+++ b/src/incremental/serialize.ml
@@ -40,13 +40,14 @@ let type_to_file_name = function
   | CilFile -> cil_file_name
   | VersionData -> version_map_filename
 
-let solver_data : Obj.t option ref = ref None
+(** Used by the server mode to avoid serializing the solver state to the filesystem *)
+let server_solver_data : Obj.t option ref = ref None
 
 (** Loads data for incremental runs from the appropriate file *)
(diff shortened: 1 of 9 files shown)


src/incremental/serialize.ml
```diff
@@ -38,22 +40,35 @@ let type_to_file_name = function
...
+    let p = Filename.concat d (type_to_file_name data_type) in
+    marshal data p)
 
 (** Deletes previous analysis results and moves the freshly created results there.*)
 let move_tmp_results_to_results () =
-  if Sys.file_exists (gob_results_dir ()) then begin
-    Goblintutil.rm_rf (gob_results_dir ());
-  end;
-  Sys.rename (gob_results_tmp_dir ()) (gob_results_dir ())
+  if not (server ()) then (
```
michael-schwarz, line 70: Would we not still need to do this?

michael-schwarz (approved):

src/util/server.ml
```diff
@@ -0,0 +1,155 @@
...
+    type result =
+      ((string * CilType.Location.t * string * int) * bool option) list [@@deriving to_yojson]
+    let process query serv =
+      GobConfig.set_auto "trans.activated[+]" "'expeval'";
+      ExpressionEvaluation.gv_query := Some query;
+      analyze serv;
+      GobConfig.set_auto "trans.activated[-]" "'expeval'";
+      !ExpressionEvaluation.gv_results
+  end);
+
```
sim642, line 149: Do you have a full list of the possible commands we discussed at the server mode meeting a while ago? Even if they're not all implemented yet, maybe it would be a good idea to leave TODOs here with the ones that are still missing. Would make it easier to know right from the code instead of having to find some external document or list.

src/util/server.ml
```diff
@@ -0,0 +1,155 @@
...
+    match args with
+    | Ok args -> Ok args
+    | Error err ->
+      (* This is a hack to handle cases where C.args is primitive type like int or string. *)
+      match params with
+      | Some `List [param] -> (
+          match C.args_of_yojson param with
+          | Ok arg -> Ok arg
+          | _ -> Error err)
+      | _ -> Error err
```
sim642, line 48: What exactly is a problem with those?

Also, I think it might be better to not allow them anyway, but have them be in records, where the field name (key) is at least descriptive (like the reset field for the analyze command below), rather than just having values like `5` with no apparent meaning.

src/util/server.ml
jerhard: @keremc I posted a list of these on the #interactive slack channel.

src/incremental/serialize.ml
keremc: Since I hijack all calls to `store_data`, we never actually create the `gob_results_tmp_dir ()` directory in the server mode, so the `Sys.rename` throws.

src/incremental/serialize.ml
```diff
@@ -38,22 +40,35 @@ let type_to_file_name = function
   | CilFile -> cil_file_name
   | VersionData -> version_map_filename
 
+let solver_data : Obj.t option ref = ref None
```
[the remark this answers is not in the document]
keremc: OK, I did both.

src/util/server.ml
```diff
@@ -0,0 +1,155 @@
...
+  while not (Stream.is_empty stream) do
+    let line = Stream.next stream in
+    match line with
+    | `Json json -> (
+        try
+          let message = Message.either_of_yojson json in
+          match message.id with
+          | Some id -> handle_request serv message id
+          | _ -> () (* We just ignore notifications for now. *)
+        with exn -> prerr_endline (Printexc.to_string exn))
```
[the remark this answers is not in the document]
keremc: OK, I'm a bit pickier about the types of exceptions I catch now. `config` and `merge_config` still have catch-all blocks, but I feel those are justified.

src/util/server.ml
keremc: jsonrpc expects `params` to be an `` `Assoc `` or `` `List ``, and this fails for simple types like `string` and `int`. So, as a workaround, if the initial parsing fails, I check if the `params` is a list that contains a single object, and I try to parse this object.

I guess I could remove it. Originally, I used it for `analyze`, but I also thought a single boolean argument looked ugly and replaced it with a record type.

src/incremental/serialize.ml
```diff
@@ -38,22 +40,35 @@ let type_to_file_name = function
...
-  let d = gob_results_tmp_dir () in
-  ignore @@ Goblintutil.create_dir d;
-  let p = Filename.concat (d) (type_to_file_name data_type) in
-  marshal data p
+let store_data (data : 'a) (data_type : incremental_data_kind) =
+  if server () then
+    match data_type with
+    | SolverData -> solver_data := Some (Obj.repr data)
+    | _ -> ()
+  else (
```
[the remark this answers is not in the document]
keremc: > Is init (i.e. Spec.init) called for the analyses before an incremental run in server mode? If init is called again, but without passing the data from the previous run, the data from the previous run might be overwritten and lost.

Yeah, ATM, `Spec.init None` is called in [control.ml](https://github.com/goblint/analyzer/blob/0aa80fd0ef3e6c7c9a198633809ce7100bffb498/src/framework/control.ml#L305) as usual since I don't set `load_run`. I'll fix it.

src/util/server.ml
sim642: Thanks for the explanation, I wasn't too familiar with what what JSON-RPC allows there.
I guess we can leave this fallback as well, might be a bit more convenient for experimenting, although all commands that we intend to keep using should use records.

src/incremental/serialize.ml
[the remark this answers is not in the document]
jerhard: Yes, #498 is not on master yet, I think you might just want to merge it into your branch.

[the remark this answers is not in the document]
sim642: I think #498 could easily be merged into master or this PR. I think I didn't notice at the time that it was against interactive not master.

src/util/server.ml
```diff
@@ -0,0 +1,155 @@
...
+        | Ok args -> (
+            try
+              C.process args serv
+              |> C.result_to_yojson
+              |> Response.ok id
+            with exn -> handle_exn id exn)
+        | Error s -> Response.Error.(make Code.InvalidParams s () |> Response.error id))
+    | _ -> Response.Error.(make Code.MethodNotFound message.method_ () |> Response.error id)
+  in
+  Response.yojson_of_t response |> Yojson.Safe.to_string |> print_endline
```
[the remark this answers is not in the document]
keremc: I now create a UNIX socket for the JSON stuff.

src/util/server.ml
```diff
@@ -0,0 +1,175 @@
...
+            | Some id -> handle_request serv message id
+            | _ -> () (* We just ignore notifications for now. *)
+          with Json.Of_json (s, _) -> prerr_endline s)
+      | `Exn exn -> prerr_endline (Printexc.to_string exn))
+
+let make ?(input=stdin) ?(output=stdout) file do_analyze : t = { file; do_analyze; input; output }
+
+let bind () =
+  let server = GobConfig.get_string "server" in
+  if server = "-" then None, None else (
```
sim642, line 89: I was about to ask why this is `"-"` not the default `""`, but I get it now:
* `""` means no server mode.
* `"-"` means server mode with stdio.
* `"otherwise"` means server mode with "otherwise" as UNIX socket.

I think this might be overloading the `server` option a bit too much already. It would be clearer and more future-proof to make `server` a category and have numerous options under it like:
* `server.enabled` - boolean.
* `server.mode` - `stdio`/`unix`
* `server.unix-socket` - name of UNIX socket to use in `unix` mode.

src/util/server.ml
```diff
@@ -0,0 +1,175 @@
...
+    if Sys.file_exists server then
+      Sys.remove server;
+    let socket = Unix.socket PF_UNIX SOCK_STREAM 0 in
+    Unix.bind socket (ADDR_UNIX server);
+    Unix.listen socket 1;
+    let conn, _ = Unix.accept socket in
+    Unix.close socket;
+    Sys.remove server;
+    Some (Unix.input_of_descr conn), Some (Unix.output_of_descr conn))
```
sim642, line 98: The way I understand it, this currently opens the UNIX socket once to accept one client and none any further. It might be nice if the UNIX socket remained and accepted another connection after the first one closes. So no parallel handling of multiple connections, but just sequentially. This could be useful if you want to reconnect a client to the server without having to kill the server.

src/util/server.ml
keremc: Good idea. Your comment also reminded me to update the option description since "Server mode" is a bit unexplanatory now.

src/util/server.ml
keremc: Yeah, that is what it does. My main goal was to simplify the process clean-up in case the parent HTTP server dies. But I guess it isn't complicated to make the parent process terminate its children before crashing.

sim642: Depending how the parent process dies (e.g. if it's killed or crashes), then it might not have the opportunity to terminate its children anymore.
I think for the time being we can leave it as-is until there's actual need to reconnect to the server.

src/incremental/serialize.ml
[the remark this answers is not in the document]
michael-schwarz: I made a PR where I cherry-picked the relevant part of #498: https://github.com/goblint/analyzer/pull/537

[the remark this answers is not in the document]
jerhard: @keremc It's now merged into `master`, so you might want to merge `master` into this branch again.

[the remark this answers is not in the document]
keremc: OK, I merged `master`.

src/util/server.ml
keremc: Thanks, I added some TODOs in the code.

sim642 (approved):

src/util/server.ml
```diff
@@ -0,0 +1,184 @@
...
+            with Failure (code, msg) -> Response.Error.(make code msg () |> Response.error id))
+        | Error s -> Response.Error.(make Code.InvalidParams s () |> Response.error id))
+    | _ -> Response.Error.(make Code.MethodNotFound message.method_ () |> Response.error id)
+  in
+  Response.yojson_of_t response |> Yojson.Safe.to_string |> IO.write_line serv.output;
+  IO.flush serv.output
+
+let serve serv =
+  serv.input
+  |> IO.to_input_channel
```
jerhard, line 72: @keremc  The documentation on `BatIO.to_input_channel` says that the operation is "extremely costly and is provided essentially for debugging purposes", see [here](https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatIO.html). As `Yojson.Safe.linestream_from_channel` requires an `Stdlib.in_channel` maybe one just change `serv.input` to be of type `in_channel`?

sim642: If anyone's wondering how costly is _extremely_, then here's the answer: https://github.com/ocaml-batteries-team/batteries-included/blob/a7276a8d41b31035423235b2395aa82ab5d6f392/src/batIO.ml#L687-L716 — it might go through a temporary file...

keremc: Oops. Thanks for pointing this out. I plan to make a server mode-related PR soon anyway, so I'll include the fix in it.

PR #525, opened 2022-01-11
Use global invariant for dead branch detection

sim642:
While debugging other incremental issues, I realized that some dead branch warnings go missing. This was because dead branch detection used global hashtables in the `Deadcode` module to accomplish its job.

This PR replaces that with the general global invariant mechanism. While doing so, it switches from using locations to nodes, making it even more robust for the incremental case.
A useful side effect of the latter is that now dead branch warnings are also given for subexpressions of short-circuting conditions (e.g. `a && b`). Previously this was broken since both branching nodes had the same location, which quietly broke the location-based hashtable.

This PR is on top of #391. On master the incremental problem doesn't appear because postsolving isn't done incrementally, so all the dead branch warnings get recreated anyway.

### TODO
- [x] Fix global view in HTML.

### DIFF
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -1127,22 +1127,16 @@ struct
   let threadenter ctx = S.threadenter (conv ctx)
   let threadspawn ctx lv f args fctx = S.threadspawn (conv ctx) lv f args (conv fctx)
 
-  module Locmap = Deadcode.Locmap
-
-  let dead_branches = function true -> Deadcode.dead_branches_then | false -> Deadcode.dead_branches_else
 
   let branch ctx exp tv =
     if !GU.postsolving then (
-      Locmap.replace Deadcode.dead_branches_cond !Tracing.current_loc exp;
       try
         let r = branch ctx exp tv in
         (* branch is live *)
-        Locmap.replace (dead_branches tv) !Tracing.current_loc false; (* set to live (false) *)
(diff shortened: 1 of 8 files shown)


src/framework/constraints.ml
```diff
@@ -1056,27 +1056,102 @@ module DeadBranchLifter (S: Spec): Spec =
...
+      | `Bot -> EM.bot ()
+      | `Lifted2 x -> x
+      | _ -> failwith "DeadBranchLifter.node"
+    let create_s s = `Lifted1 s
+    let create_node node = `Lifted2 node
+
+    let printXml f = function
+      | `Lifted1 x -> S.G.printXml f x
+      | `Lifted2 x -> BatPrintf.fprintf f "<analysis name=\"dead-branch\">%a</analysis>" EM.printXml x
+      | x -> BatPrintf.fprintf f "<analysis name=\"dead-branch-lifter\">%a</analysis>" printXml x
```
michael-schwarz, line 1092: It is on purpose that the analysis names are different here?

michael-schwarz (approved):

src/framework/constraints.ml
sim642: This shouldn't be needed in proper cases anyway, but I didn't want the top and bot output going under dead-branch, because that would mix them up in the output with the top and bot of the lifted boolean lattice, making it incredibly hard to debug, if anything ever goes wrong with it.

PR #528, opened 2022-01-12
Add more `__builtin`s  and `strsignal` to library functions

michael-schwarz:
> Returns x with the order of the bytes reversed; for example, 0xaabb becomes 0xbbaa. Byte here always means exactly 8 bits.

from https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

Adds these functions as `readsAll` to our library functions. This leads to less needless invalidation of all globals in some benchmarks.

EDIT:

Also adds `__builtin` variant of some other library functions that we encountered in benchmarks (including their `_chk` versions that are used when fortification is on in GCC).


sim642:
Do you remember, which benchmarks these cover? Some of the existing ones in our bench repo or some new project?

tests/regression/04-mutex/61-allfuns-globs.c
```diff
@@ -0,0 +1,23 @@
+// PARAM: --set allfuns true
```
sim642, line 1: How is this `allfuns` test related?

tests/regression/04-mutex/61-allfuns-globs.c
```diff
@@ -0,0 +1,23 @@
...
+
+int myglobal;
+int three = 3;
+pthread_mutex_t A_mutex = PTHREAD_MUTEX_INITIALIZER;
+pthread_mutex_t B_mutex = PTHREAD_MUTEX_INITIALIZER;
+
+void t1() {
+  pthread_mutex_lock(&A_mutex);
+  myglobal++; //RACE!
+  pthread_mutex_lock(&A_mutex);
```
sim642, line 12: Unlock?

tests/regression/04-mutex/61-allfuns-globs.c
```diff
@@ -0,0 +1,23 @@
...
+
+  assert(three == 3);
+}
+
+void t2() {
+  pthread_mutex_lock(&B_mutex);
+  myglobal++; //RACE!
+  pthread_mutex_lock(&B_mutex);
```
sim642, line 20: Unlock?

tests/regression/04-mutex/61-allfuns-globs.c
```diff
@@ -0,0 +1,23 @@
...
+int three = 3;
+pthread_mutex_t A_mutex = PTHREAD_MUTEX_INITIALIZER;
+pthread_mutex_t B_mutex = PTHREAD_MUTEX_INITIALIZER;
+
+void t1() {
+  pthread_mutex_lock(&A_mutex);
+  myglobal++; //RACE!
+  pthread_mutex_lock(&A_mutex);
+
+  assert(three == 3);
```
sim642, line 14: `assert.h` isn't even included.

sim642 (approved):
The added library functions look fine, but I'm confused by the test.

michael-schwarz:
> Do you remember, which benchmarks these cover? Some of the existing ones in our bench repo or some new project?

Not 100% certain anymore, but I think those come from OpenSSL.

tests/regression/04-mutex/61-allfuns-globs.c
michael-schwarz: Because function definitions included from headers such as `"/usr/include/x86_64-linux-gnu/bits/byteswap.h" call these functions. I noticed the problem with test `04/32` which is an allfuns test that this one is modelled after.

sim642: But that test doesn't include the byteswap header: it's not anywhere in the preprocessed output.

michael-schwarz: On my machine it does:

```
./regtest.sh 04 32 --enable justcil > out.cil.c
```

~~~C
#line 108 "/usr/include/x86_64-linux-gnu/bits/byteswap.h"
__inline static __uint64_t __bswap_64(__uint64_t __bsx ) 
{ 
  long tmp ;

  {
#line 111
  tmp = __builtin_bswap64((long )__bsx);
#line 111
  return ((__uint64_t )tmp);
}
}
~~~

sim642: Hold on, those builtins also declared by CIL: https://github.com/goblint/cil/blob/1e53aa83af4c7d51353092eb1d7ff52979fb20ad/src/cil.ml#L2879-L2891. So all the missing ones (e.g. `bswap128`) should also be added there.

PR #529, opened 2022-01-12
Generalize pairwise race detection

### LINKED ISSUE #526: Replace labeled string based access partitioning with a richer type-safe one
As discussed in GobCon, it would be very good to replace the labeled string based access partitioning with something that is richer (allowing more sophisticated pairwise conflict checking like #518) and more type-safe.
Also we figured that the partitioning component (`LSSSet`) wouldn't even be necessary, but only the "lockset" component (`LSSet`), since partitioning is only used by region analysis and the same should be achievable by storing regions in the latter and doing conflict checking using the region information).

This is best done after #518 to reduce conflicts.

sim642:
Closes #526.
This PR is currently opened on top of #518.

After hours of thinking I managed to find a way to find a way to do this in a satisfactory way. The trick was to use the combined access (`A`) module directly from the activated analyses, instead of accessing it through the outermost `Spec`. Achieving this required splitting parts of `MCP` into other modules (`MCPRegistry` and `MCPAccess`) to avoid circular dependencies.
One last circular dependency is broken by having the access analysis query `PartAccess` (which is now only answered by `MCP`, which uses the `access` functions of activated analyses) return the result as `Obj.t`. The only reason for that is to avoid mentioning `MCPAccess` in `Queries` (which is required by `MCP` again).

### TODO
- [x] Comment some parts.
- [x] Clean up code.
- [x] Make race output readable (by omitting useless units from analyses that don't supply access information).


src/analyses/threadId.ml
```diff
@@ -96,15 +87,35 @@ struct
...
+    let should_print _ = true
+  end
+  let access ctx e vo w =
+    let unique =
+      if is_unique ctx then
+        let tid = fst ctx.local in
+        Some tid
+      else
+        None
+    in
```
michael-schwarz, line 115: Is this actually needed here? Would MHP not also detect this?

src/analyses/mCPAccess.ml
```diff
@@ -0,0 +1,43 @@
...
+      f a n (assoc_dom n) d1 d2
+    in
+    try if length x <> length y
+      then raise (DomListBroken "binop_fold : differing lengths")
+      else fold_left (fun a (n,d) -> f a n d @@ assoc n y) a x
+    with Not_found -> raise (DomListBroken "binop_fold : assoc failure")
+
+  let may_race x y = not @@ binop_fold (fun a n (module S: Analyses.MCPA) x y ->
+      a || not (S.may_race (obj x) (obj y))
+    ) false x y
```
michael-schwarz, line 31: Is there some deeper meaning behind the double negation?

```suggestion
  let may_race x y = binop_fold (fun a n (module S: Analyses.MCPA) x y ->
      a && (S.may_race (obj x) (obj y))
    ) true x y
```

src/analyses/threadId.ml
sim642: I wasn't sure, so I kept all the information that we previously had. The logic for this is that a unique thread doesn't race with itself. If MHP implies that, then I can remove this component indeed.

src/analyses/mCPAccess.ml
sim642: At some point I had to insert the negations to get things to work, but I'm not sure anymore. It might've been because `conflict` was a confusing name, which implied nothing about may vs must. I'll check again if this can be simplified.

src/analyses/threadId.ml
michael-schwarz: https://github.com/goblint/analyzer/blob/f9b403da7ff28458401095f7153bf4391b39b683/src/cdomains/mHP.ml#L29-L35

This should cover that case, right?

sim642: There is a subtle difference actually. MHP uses just `is_unique` of the thread ID domain, but the one above uses `MustBeUniqueThread` query, which is answered by more analyses. In particular, if simple thread IDs are used and the old thread analysis is activated, it would also work, whereas through the simple thread IDs, MHP cannot determine uniqueness.

sim642:
@michael-schwarz Do you think it's ready to merge now?

michael-schwarz (approved):

PR #530, opened 2022-01-12
Ignore `.s` files (assembly) in compilation database

michael-schwarz:
`cpp` does not generate `.i` files from them, but Goblint later goes looking for them, failing because it can't find them.
Needed for https://github.com/goblint/bench/issues/7

### DIFF
--- a/src/util/compilationDatabase.ml
+++ b/src/util/compilationDatabase.ml
@@ -49,42 +49,46 @@ let load_and_preprocess ~all_cppflags filename =
   let preprocessed_dir = Goblintutil.create_dir (Filename.concat ".goblint" "preprocessed") in
   let preprocess obj =
     let file = obj.file in
-    let preprocessed_file = Filename.concat preprocessed_dir (Filename.chop_extension (GobFilename.chop_common_prefix database_dir file) ^ ".i") in
-    GobSys.mkdir_parents preprocessed_file;
-    let preprocess_command = match obj.command, obj.arguments with
-      | Some command, None ->
-        (* TODO: extract o_file *)
-        let command = reroot command in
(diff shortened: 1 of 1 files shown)


sim642:
Do you have a complete command object from a compilation database for such assembly file?

I understand that nothing meaningful for us would come out of assembly code, but I'm just surprised that whatever binary (`gcc` or something else?) is used to compile them accepts all the preprocessing and output arguments but doesn't actually output anything. Would've expected it to just fail at that point.

michael-schwarz:
```
 {
  "directory": "/home/michael/Documents/goblint-cil/bench-ideas/openssl",
  "arguments": [
   "gcc",
   "-fPIC",
   "-pthread",
   "-m64",
   "-Wa,--noexecstack",
   "-Wall",
   "-O3",
   "-DOPENSSL_USE_NODELETE",
   "-DL_ENDIAN",
   "-DOPENSSL_PIC",
   "-DOPENSSLDIR=\"/usr/local/ssl\"",
   "-DENGINESDIR=\"/usr/local/lib64/engines-3\"",
   "-DMODULESDIR=\"/usr/local/lib64/ossl-modules\"",
   "-DOPENSSL_BUILDING_OPENSSL",
   "-DNDEBUG",
   "-c",
   "-o",
   "crypto/ec/libcrypto-shlib-ecp_nistz256-x86_64.o",
   "crypto/ec/ecp_nistz256-x86_64.s"
  ],
  "file": "crypto/ec/ecp_nistz256-x86_64.s"
 }
 ```

sim642:
Apparently GCC does produce preprocessing output for capitalized `.S`, which is supposed to be preprocessed. Although being assembly we wouldn't be able to parse it anyway.

Oddly for `.s` with `-E` everything is just ignored:
> Input files that don’t require compilation are ignored. 

(https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#index-E)

sim642 (approved):

PR #532, opened 2022-01-13
Experiment with intermediate CIL merges

sim642:
Implements an experimental version of my idea from https://github.com/goblint/bench/issues/7#issuecomment-1011933335.


michael-schwarz:
That  doesn't seem to help unfortunately. It does not significantly reduce the memory usage (still about 20 GB) for https://github.com/goblint/bench/issues/7 but is drastically slower (>18min until CIL is done merging).

sim642:
Closing as not really useful.

PR #533, opened 2022-01-14
Use nodes for messages instead of locations

### LINKED ISSUE #524: Incrementally reused warnings have old locations
@karoliineh discovered that warnings during incremental loading can remain at their old locations. For example, when simply adding/removing empty lines. That causes no AST changes, so no nodes change and their warnings are reused, but locations do change.

The problem is that warnings (messages) have their location (or multiple if grouped!), not a node.

sim642:
Closes #524.
This is currently opened on top of #525 since that already adapts dead branch warnings to nodes instead of locations.

By using nodes for locations, warnings in syntactically unchanged but moved functions should move accordingly.

### TODO
- [x] Use incremental locations for messages.
- [x] Check that warning locations move if empty lines inserted/removed, etc.


karoliineh (approved):
Works for me.

Warning locations do move when empty lines are inserted or removed (with "load" set to true in "incremental" in the conf).

PR #534, opened 2022-01-14
Fix Address Domain calls to `widen x y` where `x \leq y` did not hold & Ensure `widen x x = x` in `def_exc` domain

michael-schwarz:
With @jerhard  we discovered that the issue with `def_exc` not terminating seems to be that `widen` is somehow called from `narrow` in the addressDomain without ensuring that the second argument is greater than the first one.

Independently of that, the `widen` of `def_exc` should be fixed such that `widen x x = x`.

### DIFF
--- a/src/cdomains/addressDomain.ml
+++ b/src/cdomains/addressDomain.ml
@@ -31,6 +31,7 @@ struct
   let widen x y =
     if M.tracing then M.traceli "ad" "widen %a %a\n" pretty x pretty y;
     let r = widen x y in
+    M.tracel "ad" "%s\n" (Printexc.raw_backtrace_to_string (Printexc.get_callstack 50));
     if M.tracing then M.traceu "ad" "-> %a\n" pretty r;
     r
 
@@ -133,7 +134,7 @@ struct
     | false, false -> cop x y
 
   let meet x y   = merge join meet x y
-  let narrow x y = merge widen narrow x y
+  let narrow x y = merge (fun x y -> (join x y) |> widen x) narrow x y
 
   let invariant c x =
     let c_exp = Cil.(Lval (BatOption.get c.Invariant.lval)) in
--- a/src/cdomains/addressDomain.ml
+++ b/src/cdomains/addressDomain.ml
@@ -31,7 +31,6 @@ struct
   let widen x y =
     if M.tracing then M.traceli "ad" "widen %a %a\n" pretty x pretty y;
     let r = widen x y in
-    M.tracel "ad" "%s\n" (Printexc.raw_backtrace_to_string (Printexc.get_callstack 50));
     if M.tracing then M.traceu "ad" "-> %a\n" pretty r;
     r
 
--- a/src/cdomains/addressDomain.ml
+++ b/src/cdomains/addressDomain.ml
@@ -134,7 +134,7 @@ struct
     | false, false -> cop x y
 
   let meet x y   = merge join meet x y
-  let narrow x y = merge (fun x y -> (join x y) |> widen x) narrow x y
+  let narrow x y = merge (fun x y -> widen x (join x y)) narrow x y
 
   let invariant c x =
     let c_exp = Cil.(Lval (BatOption.get c.Invariant.lval)) in
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -1439,8 +1439,9 @@ struct
   let widen ik x y =
     if get_bool "ana.int.def_exc_widen_by_join" then
       join' ik x y
+    else if equal x y then
+      x
     else
-      (* if equal x y then x else *)
       join' ~range:(size ik) ik x y
 
 
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -1436,11 +1436,13 @@ struct
 
   let join ik = join' ik
 
-  let widen ik =
+  let widen ik x y =
     if get_bool "ana.int.def_exc_widen_by_join" then
-      join' ik
+      join' ik x y
     else
-      join' ~range:(size ik) ik
+      (* if equal x y then x else *)
+      join' ~range:(size ik) ik x y
+
 
   let meet ik x y =
     match (x,y) with
--- a/src/util/options.schema.json
+++ b/src/util/options.schema.json
@@ -394,9 +394,9 @@
             "def_exc_widen_by_join": {
               "title": "ana.int.def_exc_widen_by_join",
               "description":
-                "Perform def_exc widening by joins. Gives threshold-widening like behavior, with thresholds given by the ranges of different integer types. TODO: Value false has termination problems.",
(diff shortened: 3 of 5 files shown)


src/cdomains/intDomain.ml
```diff
@@ -1436,11 +1436,13 @@ struct
...
   let join ik = join' ik
 
-  let widen ik =
+  let widen ik x y =
     if get_bool "ana.int.def_exc_widen_by_join" then
-      join' ik
+      join' ik x y
     else
-      join' ~range:(size ik) ik
+      (* if equal x y then x else *)
```
michael-schwarz, line 1443: Once we're satisfied with the other fix, we should add this back in

src/cdomains/addressDomain.ml
```diff
@@ -133,7 +134,7 @@ struct
     | false, false -> cop x y
 
   let meet x y   = merge join meet x y
-  let narrow x y = merge widen narrow x y
+  let narrow x y = merge (fun x y -> (join x y) |> widen x) narrow x y
```
sim642, line 137: You beat me to it opening the PR, but this is indeed where the invalid call is. I would format it like how it's done elsewhere (e.g. solvers):
```suggestion
  let narrow x y = merge (fun x y -> widen x (join x y)) narrow x y
```

src/cdomains/intDomain.ml
sim642: Why is this required though? Just a bit of precision?

vesalvojdani:
Could you also change `//PARAM: --disable ana.int.def_exc_widen_by_join` for knot test? (03/23).

src/cdomains/addressDomain.ml
```diff
@@ -31,6 +31,7 @@ struct
   let widen x y =
     if M.tracing then M.traceli "ad" "widen %a %a\n" pretty x pretty y;
     let r = widen x y in
+    M.tracel "ad" "%s\n" (Printexc.raw_backtrace_to_string (Printexc.get_callstack 50));
```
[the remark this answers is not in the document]
vesalvojdani: The complaint is about the missing "if"... it should be there, no?

michael-schwarz:
> Could you also change //PARAM: --disable ana.int.def_exc_widen_by_join for knot test? (03/23).

Done!

src/cdomains/addressDomain.ml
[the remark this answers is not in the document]
vesalvojdani: Maybe I can't read the diffs right... :)  The semgrep is complaining about the missing `if M.tracing then ...`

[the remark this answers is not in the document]
michael-schwarz: Ah, ok, now I understood!
This is actually super verbose (prints the entire callstack on widen calls), so the plan was to remove it anyway before merging the PR.

vesalvojdani:
I think someone commented, but assuming this works, it would make sense to make it the default again? Or maybe even remove the option. Might be good to have a full SV-COMP run again before doing that. I will have a look now at the bench repo to merge Julian's thing and run it again.

sim642:
Indeed, we should test the cases where we observed termination problems with this originally to confirm this covers all of them.
Then we can change the default back. I'd still keep the option just in case for the time being.

michael-schwarz:
I think switching the default makes sense.

I'd want to hold on to the option though because it's s very helpful for globals where any precision we give up during widening can never be recovered as there is no narrowing. This is the equivalent to threshold widening which I added for intervals.

src/cdomains/addressDomain.ml
```diff
@@ -133,7 +133,7 @@ struct
     | false, false -> cop x y
 
   let meet x y   = merge join meet x y
-  let narrow x y = merge widen narrow x y
+  let narrow x y = merge (fun x y -> widen x (join x y)) narrow x y
```
vogler, line 136: Should this be fixed here where it's just applied or further up?
`widen` seems to be included from `HoareDomain.HoarePO`.

src/cdomains/intDomain.ml
```diff
@@ -1436,11 +1436,14 @@ struct
 
   let join ik = join' ik
 
-  let widen ik =
+  let widen ik x y =
     if get_bool "ana.int.def_exc_widen_by_join" then
-      join' ik
+      join' ik x y
+    else if equal x y then
```
vogler, line 1442: Good point. Don't want to increase the range if nothing changes.

src/cdomains/addressDomain.ml
jerhard: The question is whether the `widen` operations in general can expect that the first value is less or equal to the second. If yes, then performing a join on the second operand in the `widen` seems redundant.

vesalvojdani (approved):
Seems to work on the interactive benches. It also makes smtprc almost 2x faster compared to the version that widens via joins:
- [no_widen_by_join](https://goblint.cs.ut.ee/vesal-results/2022-01-14-performance-no_def_exc_widen_by_join/) (with this fix)
- [widen_by_join](https://goblint.cs.ut.ee/vesal-results/2022-01-14-performance-def_exc_widen_by_join/)

src/cdomains/addressDomain.ml
vogler: We join in the solver, i.e. at the very top. So either it really is required here due to some complicated merge/join/widen interaction or a join is not done somewhere else.

vogler: > The question is whether the widen operations in general can expect that the first value is less or equal to the second.

Yes, I thought that's what we're going for.

michael-schwarz: > We join in the solver, i.e. at the very top. So either it really is required here due to some complicated merge/join/widen interaction or a join is not done somewhere else.

https://github.com/goblint/analyzer/blob/e00c961b38cb4bd2e7a417cab576de9f9d9acf46/src/cdomains/addressDomain.ml#L124-L136

We are inside a `narrow` here where we want to have special handling if both pointers contain the top pointer (namely not throw away definite addresses). So it's not a place where the solver calls `widen` it is just us manually deciding we want to do that in some cases instead of a `narrow`. So I think we need to fix it here.

michael-schwarz:
>  It also makes smtprc almost 2x faster compared to the version that widens via joins:

Wow, that is quite a difference! Also, the number of variables is almost cut in half as well!
Would be interesting if there is a difference in the amount of dead code `--enable dbg.print_dead_code` between the different configurations....

vogler (approved):

sim642 (approved):

sim642:
> Wow, that is quite a difference! Also, the number of variables is almost cut in half as well!
> Would be interesting if there is a difference in the amount of dead code `--enable dbg.print_dead_code` between the different configurations....

I guess this is the classic pitfall of there being too many unnecessary thresholds (for def_exc ranges) that have to be analyzed through before ending up with the same result. With intervals active, it's a complete waste anyway because the interval can be both more precise and faster (widens to inf, narrows to loop bound, no intermediate thresholds).

And if I were to guess, the number of variables might have just dropped because there aren't unnecessary contexts with those intermediate def_exc ranges.

PR #535, opened 2022-01-14
Add files list to json-messages output

sim642:
Adds a list of analyzed (and dependency) files to the json-messages output used by GobPie.
It would be very useful for the latter when using compilation databases to know the list of files that Goblint read from the compilation database and analyzed. Otherwise GobPie would have to unnecessarily duplicate compilation database logic.

Moreover, what this list should also contain is all the dependency files, i.e. included headers (transitively!), because any change to those would also require reanalysis. This however is nontrivial and I currently see two possible ways to go about it:
1. Extend CIL's `#line` etc pragma parser to not only change the current location inside its lexer, but record each filename encountered and expose that to Goblint. Just iterating over all locations inside the AST might be insufficient if the header file contains no definitions!
2. Use the preprocessor with `-M` to ask it for all dependencies in Makefile format and parse that in Goblint.

@michael-schwarz Do you have a preference or any other idea?

### TODO
- [x] Add included header files.

### DIFF
--- a/src/framework/analyses.ml
+++ b/src/framework/analyses.ml
@@ -274,13 +274,13 @@ struct
         (* TODO: change CIL to extract all include file names from line pragmas? *)
         (* TODO: or do extra preprocessing with -M? *)
         iterGlobals file (function
-            | GFun (fd, loc) -> SH.replace files loc.file ()
+            | GFun (fd, loc) -> SH.replace files loc.file (Hashtbl.find_option Preprocessor.dependencies loc.file)
             | _ -> () (* TODO: add locs from everything else, including all AST nodes *)
           );
-        files |> SH.keys |> List.of_enum
+        files |> SH.to_list
       in
       let json = `Assoc [
-        ("files", `List (List.map (fun s -> `String s) files));
+        ("files", `Assoc (List.map (Tuple2.map2 [%to_yojson: string list option]) files));
         ("messages", Messages.Table.to_yojson ());
       ]
       in
--- a/src/framework/analyses.ml
+++ b/src/framework/analyses.ml
@@ -271,11 +271,11 @@ struct
       let files =
         let module SH = BatHashtbl.Make (Basetype.RawStrings) in
         let files = SH.create 100 in
-        (* TODO: change CIL to extract all include file names from line pragmas? *)
-        (* TODO: or do extra preprocessing with -M? *)
         iterGlobals file (function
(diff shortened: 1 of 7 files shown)


sim642:
I implemented the dependencies Makefile approach. With `-MMD` as an additional argument to the preprocessor, it creates a `.d` file on the side in Makefile format, listing all the included (non-system) dependencies of the preprocessed file in the same go. This is then parsed in Goblint and added to the json-messages output in the following way:
```json
  "files": {
    "/home/simmo/Desktop/goblint-action-test/cmake-project/main.c": [
      "/home/simmo/dev/goblint/sv-comp/goblint/_opam/share/goblint/includes/assert.h",
      "/home/simmo/Desktop/goblint-action-test/cmake-project/lib/lib.h"
    ],
    "/home/simmo/dev/goblint/sv-comp/goblint/_opam/share/goblint/includes/stdlib.c":
      [],
    "/home/simmo/Desktop/goblint-action-test/cmake-project/lib/lib.c": [
      "/home/simmo/Desktop/goblint-action-test/cmake-project/lib/lib.h"
    ]
  },
```

Doing it in CIL might be a bit more general if we want to analyze already preprocessed programs that contain those line pragmas, because this wouldn't know about those. Although I don't think we should bother because we cannot reasonably handle such programs with GobPie etc anyway: the warning locations would be in possibly nonexistent physical files that were included, not in the physical file that was actually analyzed, so all the output would be a mess anyway.

michael-schwarz:
Looks good!

Would this information also be sufficient to make the pre-processing (and potentially parsing into `CIL` files) incremental, in that if no input files have changed, no new run of `cpp` or no new computation of a CIl `file` is neccessary?

In order to guarantee this, transitive dependencies would probably have to be considered as well...

sim642:
> Would this information also be sufficient to make the pre-processing (and potentially parsing into `CIL` files) incremental, in that if no input files have changed, no new run of `cpp` or no new computation of a CIl `file` is neccessary?

I think it should be. The use case in GobPie is somewhat related.

> In order to guarantee this, transitive dependencies would probably have to be considered as well...

This should cover all includes transitively, because they're inlined transitively during preprocessing, so the corresponding indirect dependency should be direct in this output.

sim642:
Due to some MagpieBridge oddities this information cannot be used to its true extent in GobPie for the time being. But I'm merging it anyway since it might become useful.

PR #536, opened 2022-01-16
Reuse more simple printing code

sim642:


src/cdomains/lval.ml
```diff
@@ -336,35 +333,9 @@ struct
   let meet = merge `Meet
   let narrow = merge `Narrow
 
-  let printXml f x = BatPrintf.fprintf f "<value>\n<data>\n%s\n</data>\n</value>\n" (XmlUtil.escape (show x))
-  let to_yojson x = `String (show x)
-
   let pretty_diff () (x,y) = dprintf "%s: %a not leq %a" (name ()) pretty x pretty y
 end
 
-module Stateless (Idx: Printable.S) =
```
sim642, line 345: This was unused and its `show` implementation didn't even match its type `t`.

src/cdomains/lval.ml
```diff
@@ -9,13 +9,6 @@ type ('a, 'b) offs = [
   | `Index of 'b * ('a,'b) offs
 ] [@@deriving eq, ord]
 
-type ('a,'b) offs_uk = [
-  | `NoOffset
-  | `UnknownOffset
-  | `Field of 'a * ('a,'b) offs
-  | `Index of 'b * ('a,'b) offs
-]
```
sim642, line 17: This was only used by the unused (and now removed) `Stateless` below.
Moreover, this contained a copy-paste error: it used `offs` instead of recursively using `offs_uk` like it should've. That's the reason why `Stateless` wasn't even correctly typed.

michael-schwarz (approved):

Issue #541, opened 2022-01-18
IntDomTuple only compares activated domains


sim642:
Here's the implementation of its `equal`:
https://github.com/goblint/analyzer/blob/b9995c8d8ab5d1677b72168b44be3b6b29b5a82d/src/cdomains/intDomain.ml#L2730-L2732
Here's `map2p`:
https://github.com/goblint/analyzer/blob/b9995c8d8ab5d1677b72168b44be3b6b29b5a82d/src/cdomains/intDomain.ml#L2717-L2721
And here's `opt_map2`, which ignores the case of `Some x, None`:
https://github.com/goblint/analyzer/blob/b9995c8d8ab5d1677b72168b44be3b6b29b5a82d/src/cdomains/intDomain.ml#L2636-L2637
And here's how `for_all` is implemented:
https://github.com/goblint/analyzer/blob/b9995c8d8ab5d1677b72168b44be3b6b29b5a82d/src/cdomains/intDomain.ml#L2639-L2641
As the comment about `to_list` says, it filters out only the activated domains.

Therefore `(Some defexc, None, None, None)` and `(Some defexc, Some interval, None, None)` are considered equal.
This in turn breaks incremental precision improvement annotations: even if the new more precise analysis projects the previously only def_exc context into a def_exc+interval context, then in the solver's hashtable, that yields a constraint variable, which is `equal` to the old one. Hence the more precise context is never actually created and the result doesn't get more precise!

The same applies to `compare` and probably most of the functions using `to_list` or `opt_map2`. This wasn't a problem before #395 because IntDomTuples with different activations never occurred during a single run of Goblint (since the activation was effectively fixed by the options). So `opt_map2` ignoring `Some x, None` never happened and `to_list` didn't filter anything like that out.
But this is now possible and IntDomTuple should be fixed to account for this. At minimum, the identity operations (`equal`, `compare`, `hash`) must handle these cases not to swallow more precise values like this. Other integer domain operations don't have to handle these cases (the projections in base analysis should ensure that), but they should throw exceptions if that somehow ever happens to not quietly hide such issues.

sim642:
Actually I now remembered that the type of `IntDomTuple` isn't even anything exotic (unlike `MCP`):
https://github.com/goblint/analyzer/blob/b9995c8d8ab5d1677b72168b44be3b6b29b5a82d/src/cdomains/intDomain.ml#L2605-L2606
[...]

PR #542, opened 2022-01-18
Add sim642/ppx_deriving_hash

sim642:
Part of #31.

A while back I toyed around with ppx derivers and, as mentioned in the above issue, implemented https://github.com/sim642/ppx_deriving_hash to do `[@@deriving hash]`. This is simpler and without dependencies, unlike ppx_hash, which require's Jane Street's Base to work.

This revealed #541, which has nothing to do with hashing. It's just that previously by sheer luck, the tuples with such different precisions had different hashes, thus they never fell into the same hashtable buckets to need `equal` comparison.

Another open question is, what are the best hash functions to derive for primitives, products (tuples, records) and sums (variants). Our manual implementations include both `Hashtbl.hash` based things and arbitrary calculations with magic constants.

### TODO
- [x] Benchmark with derived hash implementations.

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -276,7 +276,7 @@ module IntDomLifter (I : S) =
 struct
   open Cil
   type int_t = I.int_t
-  type t = { v : I.t; ikind : ikind }
+  type t = { v : I.t; ikind : (ikind [@equal (=)] [@compare Stdlib.compare] [@hash fun x -> Hashtbl.hash x]) } [@@deriving eq, ord, hash]
 
   (* Helper functions *)
   let check_ikinds x y = if x.ikind <> y.ikind then raise (IncompatibleIKinds ("ikinds " ^ Prelude.Ana.sprint Cil.d_ikind x.ikind ^ " and " ^ Prelude.Ana.sprint Cil.d_ikind y.ikind ^ " are incompatible. Values: " ^ Prelude.Ana.sprint I.pretty x.v ^ " and " ^ Prelude.Ana.sprint I.pretty y.v)) else ()
@@ -300,30 +300,7 @@ struct
   let meet = lift2 I.meet
   let widen = lift2 I.widen
   let narrow = lift2 I.narrow
-  let equal x y = if x.ikind <> y.ikind then false else I.equal x.v y.v
 
-  let hash x =
-    let ikind_to_int (ikind: ikind) = match ikind with (* TODO replace with `int_of_string % Batteries.dump` or derive *)
-    | IChar 	-> 0
-    | ISChar 	-> 1
-    | IUChar 	-> 2
-    | IBool 	-> 3
-    | IInt 	  -> 4
-    | IUInt 	-> 5
-    | IShort 	-> 6
-    | IUShort -> 7
-    | ILong 	-> 8
-    | IULong 	-> 9
-    | ILongLong -> 10
-    | IULongLong -> 11
-    | IInt128 -> 12
-    | IUInt128 -> 13
-    in
-    3 * (I.hash x.v) + 5 * (ikind_to_int x.ikind)
-  let compare x y = let ik_c = compare x.ikind y.ikind in
-    if ik_c <> 0
-      then ik_c
-      else I.compare x.v y.v
   let show x = I.show x.v  (* TODO add ikind to output *)
   let pretty () x = I.pretty () x.v (* TODO add ikind to output *)
   let pretty_diff () (x, y) = I.pretty_diff () (x.v, y.v) (* TODO check ikinds, add them to output *)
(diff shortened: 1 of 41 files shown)


sim642:
I ran it before vs after on sv-benchmarks ConcurrencySafety and SoftwareSystems and to my surprise, it even got ~5% faster:
image
[Full results table](https://goblint.cs.ut.ee/results/57-all-fast-ppx_deriving_hash-after/results/results.2022-01-21_15-44-02.table.html#/scatter?filter=0(0*status*(status(notIn(ERROR%20(verify),OUT%20OF%20MEMORY,TIMEOUT,TIMEOUT%20(true))))),1(0*status*(status(notIn(ERROR%20(verify),OUT%20OF%20MEMORY,TIMEOUT,TIMEOUT%20(true)))))&toolY=1&columnY=1&scaling=Linear&line=1.2&regression=Linear).

I haven't put any thought into optimizing ppx_deriving_hash, so I didn't expect this. If I were to guess, this improvement is just due to replacing some `Hashtbl.hash` with a proper implementation. `IntDomain` had a many of those accidentally overridden via `include Std`, so that might make an impact with integers being in all contexts of all nodes.

src/cdomains/intDomain.ml
```diff
@@ -959,8 +935,7 @@ struct
   let show (x: Ints_t.t) = if (Ints_t.to_int64 x) = GU.inthack then "*" else Ints_t.to_string x
 
   include Std (struct type nonrec t = t let name = name let top_of = top_of let bot_of = bot_of let show = show let equal = equal end)
-  (* FIXME: poly compare *)
```
michael-schwarz, line 962: Is this not still a usage of polymorphic compare?

michael-schwarz (approved):
LGTM! My only question is if we want to fork the ppx_deriver under our organization, such that all developers can potentially work on it?

src/cdomains/intDomain.ml
sim642: It doesn't appear so, because `Std` doesn't implement `compare`. I guess it must've been removed at some point, but the comment remained. And here the polymorphic `hash` is also removed from `Std`.

sim642:
> My only question is if we want to fork the ppx_deriver under our organization, such that all developers can potentially work on it>

I thought about that, but all the other ppx forks have been under personal accounts as well:
* https://github.com/vogler/ppx_distr_guards
* https://github.com/sim642/ppx_deriving
* https://github.com/sim642/ppx_deriving_yojson
Given the obscurity of ppx deriver internals, I'm doubtful that anyone else would currently work on it anyway and deal with its publishing (which I'm planning on doing soon to get rid of the pin). If there's ever real need, then it's still possible to transfer or fork it under the organization.

sim642:
The package has been accepted to opam: https://github.com/ocaml/opam-repository/pull/20555. Once it becomes available through the website, I will remove the pin, fix conflicts and do the merge.

sim642:
There's some very weird crash in incremental tests now:
```
Fatal error: exception Invalid_argument("List.iter2")
Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
Called from UpdateCil.update_ids.reset_fun in file "src/incremental/updateCil.ml", line 48, characters 4-74
Called from Stdlib__List.iter in file "list.ml", line 110, characters 12-15
Called from UpdateCil.update_ids in file "src/incremental/updateCil.ml", line 126, characters 2-43
Called from Maingoblint.diff_and_rename in file "src/maingoblint.ml", line 454, characters 22-92
Called from Maingoblint.main in file "src/maingoblint.ml", line 506, characters 110-130
Called from Stdlib.at_exit.new_exit in file "stdlib.ml", line 560, characters 59-63
Called from Stdlib.do_at_exit in file "stdlib.ml" (inlined), line 566, characters 20-61
Called from Std_exit in file "std_exit.ml", line 18, characters 8-20
```

It's somehow related to `Node.hash`. The derived version causes the crash, but the slightly modified version does not:
```ocaml
let hash = function
  (* crashes *)
  (* | Statement x0 -> (31 * 0) + (CilType.Stmt.hash x0) *)
  (* does not crash *)
  | Statement x0 -> 1 + (CilType.Stmt.hash x0)
  | FunctionEntry x0 -> (31 * 1) + (CilType.Fundec.hash x0)
  | Function x0 -> (31 * 2) + (CilType.Fundec.hash x0)
```
Is `CompareAST`, `CompareCFG` or `UpdateCil` somehow assuming that the hash of a statement node does not equal the hash of the statement itself? @jerhard @stilscher I'm extremely confused how this could cause `List.iter2` to go wrong.

sim642:
It's also _somehow_ related to `includes/pthread.c`. If that is excluded, or even if just `init_routine()` is removed, the crash disappears. But what's confusing is that the crash happens during compare/update of `main`, not `pthread_once`.

sim642:
The opam package is now published: https://opam.ocaml.org/packages/ppx_deriving_hash/. I've removed the opam pin and am merging this.

PR #547, opened 2022-01-19
Add implementation for `pthread_once`

michael-schwarz:
https://github.com/goblint/bench/issues/7 identfied that we are unsound in the presence of calls to `pthread_once`.
This takes the same approach as we did for our support for `bsearch` and `qsort` by adding an implementation for it.

### DIFF
--- a/includes/stdlib.c
+++ b/includes/stdlib.c
@@ -1,5 +1,5 @@
 #include <stddef.h>
-
+#include <pthread.h>
 void qsort(void *ptr, size_t count, size_t size, int (*comp)(const void*, const void*)) __attribute__((goblint_stub));
 void qsort(void *ptr, size_t count, size_t size, int (*comp)(const void*, const void*)) {
   // call all possible compares first, before invalidating array elements
@@ -41,4 +41,11 @@ void* bsearch(const void *key, const void *ptr, size_t count, size_t size, int (
   }
 
   return NULL;
+}
+
+int pthread_once(pthread_once_t *once_control,void (*init_routine)(void)) {
+  // Not actually once, just call it
+  int top;
+  init_routine();
+  return top;
 }
--- a/includes/stdlib.c
+++ b/includes/stdlib.c
@@ -1,4 +1,5 @@
 #include <stddef.h>
+
 void qsort(void *ptr, size_t count, size_t size, int (*comp)(const void*, const void*)) __attribute__((goblint_stub));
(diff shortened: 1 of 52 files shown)


includes/stdlib.c
```diff
@@ -42,3 +42,10 @@ void* bsearch(const void *key, const void *ptr, size_t count, size_t size, int (
 
   return NULL;
 }
+
+int pthread_once(pthread_once_t *once_control,void (*init_routine)(void)) {
+  // Not actually once, just call it
+  int top;
+  init_routine();
+  return top;
+}
```
sim642, line 51: I think we should keep these stub implementations organized according to their headers, so this should maybe go into a new file `pthread.c`.

tests/regression/02-base/10-init_allfuns.c
```diff
@@ -1,4 +1,4 @@
-// PARAM: --enable allfuns --set ana.activated "['base','threadid','threadflag','escape','mutex','access','mallocWrapper']"
+// PARAM: --enable allfuns --set ana.activated "['base','threadid','threadflag','escape','mutex','access','mallocWrapper']" --disable sem.unknown_function.invalidate.globals
```
sim642, line 1: What's this being added for? To avoid spawning the `pthread_once` stub with `allfuns`? Just add the `goblint_stub` attribute to the stub implementation to achieve that without having to change any configuration.

tests/regression/02-base/71-pthread-once.c
```diff
@@ -0,0 +1,16 @@
+//PARAM: --disable sem.unknown_function.spawn
+#include <pthread.h>
+#include <assert.h>
+
+int g;
+pthread_once_t once = PTHREAD_ONCE_INIT;
+
+void *t_fun(void *arg) {
```
sim642, line 8: I think the correct `init_routine` signature would be:
```suggestion
void t_fun() {
```

src/maingoblint.ml
```diff
@@ -280,7 +280,7 @@ let preprocess_files () =
 
   let extra_arg_files = ref [] in
 
-  extra_arg_files := find_custom_include "stdlib.c" :: !extra_arg_files;
+  extra_arg_files := find_custom_include "stdlib.c" :: find_custom_include "pthread.c" :: !extra_arg_files;
```
sim642, line 283: `pthread.c` also needs to be added to `includes/dune` and `sv-comp/archive.sh`.

sim642 (approved):

Issue #545, opened 2022-01-19
Extract incremental logic from TD3


sim642:
- [x] Unmarshaling etc should be done outside (by `Control`?)
- [ ] Finding nodes to destabilize due to change should be done outside (no node type hacks in TD3)
- [ ] TD3 should be non-incremental, TD3-incr should wrap it with pre- and post-processing.
- [ ] Experimental TD3 improvements should go to a new copy (TD-inter?)

PR #549, opened 2022-01-20
Collect constraint unknown sets in global invariant, use them for TD3 incremental load

### LINKED ISSUE #544: Collect sets of constraint unknowns corresponding to program variables, functions and nodes
Collecting such map of sets should help to clean up the incremental loading (can remove obsolete unknowns corresponding to functions without total iteration and string-based comparison) and make it easier (and more performant) to restart unknowns related to certain syntactic program features.
For example, a single program global variable `g` has many different unknowns related to it since #473:
* From base privatization (depending on chosen privatization).
* Protecting lockets from mutex analysis.
* Accesses from access analysis.

### TODO
- [x] Program variables.
- [x] ~~Program functions.~~
- [x] Program nodes.

sim642:
Closes #544.
This is currently on top of #391 since there are more incremental tests, although it might be possible to rebase it against `master` instead.

As a proof of concept that this indeed works out, I refactored the incremental loading (destabilization and deletion) in TD3 to query the constraint unknowns corresponding to CFG nodes. This removes all the string-based hacks! So it's also a step towards #545.

In addition to querying the constraint unknowns of CFG nodes, it's also possible to query the constraint unknowns of a global program variable. This properly includes the base privatization, protecting locksets and accesses.

### TODO
- [x] Check if HTML view still works.
- [x] Document.


src/framework/varQuery.ml
```diff
@@ -0,0 +1,6 @@
+type t =
+  | Global of CilType.Varinfo.t
+  | Node of {node: Node.t; fundec: CilType.Fundec.t option} (** Optional [fundec] override to allow querying old state in incremental. *)
+[@@deriving ord]
```
jerhard, line 4: What exactly is the meaning or use case of the `fundec` here? I.e. how does the `fundec` help in querying an old states of a node?

sim642: Without it, the way it's implemented in `FromSpec.iter_vars` is to take the given node, find it's `fundec` using `Node.find_fundec` and then look up the contexts for that function.

The problem in the incremental case is that `Node.find_fundec` doesn't do the lookup in the old AST, but the new one (by default in `Cilfacade`), so it would fail to find any old contexts for the old changed function.

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -342,6 +344,8 @@ struct
   let threadenter ask getg (st: apron_components_t): apron_components_t =
     {apr = getg (); priv = startstate ()}
 
+  let iter_sys_vars getg vq vf = () (* TODO: or report singleton global for any Global query? *)
```
jerhard, line 347: > TODO: or report singleton global for any Global query? 

One would indeed expect that `vf` is applied to all constraint system unknowns that somehow relate to `vq`, i.e. here to the single global unknown. No?

sim642: Well, it's not so clear cut what should and should not be considered to be related to a global `g`, i.e. what does "somehow relate" mean.

In this case, since all globals are related under a single constraint unknown, restarting one (based on the result of this query) would mean restarting them all! Same goes for the region analysis, which currently also doesn't report it.

Not only that, suppose `g` is part of a context (within any analysis), should `Global g` additionally be able to return all unknowns of all functions of all contexts that contain it? That requires a whole lot of extra mappings to collect. Not to mention, it would require some additional mechanism to allow checking if an abstract context (depending on activated analysis and their states) contains `g`.

Therefore to avoid overly complicating things for now, I thought it'd be enough to relate unknowns which very directly relate to a particular global or node.

src/analyses/basePriv.ml
```diff
@@ -473,6 +484,11 @@ struct
     CPA.fold side_var st.cpa st
 
   let threadenter = old_threadenter
+
+  let iter_sys_vars getg vq vf =
+    match vq with
+    | VarQuery.Global g -> vf g
+    | _ -> ()
```
jerhard, line 491: This code is easy, but it's copied three times. Maybe one can extract it out?

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -39,6 +39,7 @@ module type S =
 
     val thread_join: Q.ask -> (V.t -> G.t) -> Cil.exp -> apron_components_t -> apron_components_t
     val thread_return: Q.ask -> (V.t -> G.t) -> (V.t -> G.t -> unit) -> ThreadIdDomain.Thread.t -> apron_components_t -> apron_components_t
+    val iter_sys_vars: (V.t -> G.t) -> VarQuery.t -> V.t VarQuery.f -> unit (** [Queries.IterSysVars] for base. *)
```
jerhard, line 42: > (** [Queries.IterSysVars] for base. *)

Maybe that was copy-pasted and forgotten to change?

src/analyses/apron/apronPriv.apron.ml
jerhard: >In this case, since all globals are related under a single constraint unknown, restarting one (based on the result of this query) would mean restarting them all! Same goes for the region analysis, which currently also doesn't report it.

That's true. In a hypothetical setting where we have incremental restarting for the  apron analysis, I would just assume that a user might be confused, when he triggers a re-analysis of a certain global and nothing happens. But maybe the decision here only really matters when #558 is fixed. 

>Not only that, suppose g is part of a context (within any analysis), should Global g additionally be able to return all unknowns of all functions of all contexts that contain it? That requires a whole lot of extra mappings to collect. Not to mention, it would require some additional mechanism to allow checking if an abstract context (depending on activated analysis and their states) contains g.

Yes, that would certainly lead to far. Although in a `earlyglobs` setting this is not such issue, but with privatization it again would be.

sim642: > I would just assume that a user might be confused, when he triggers a re-analysis of a certain global and nothing happens.

This is just the `protection` Apron privatization though. Others, including our default `mutex-meet` have separate unknowns per global and those are returned by the varquery, so it shouldn't be too bad.

I suppose if we at some point want better control over how much things related to a global should be restarted, more arguments could be added to that varquery to choose how specific or general things may be returned. Kind of a precision knob.

src/domains/queries.ml
```diff
@@ -282,6 +286,7 @@ struct
       | Any (IsMultiple v1), Any (IsMultiple v2) -> CilType.Varinfo.compare v1 v2
       | Any (EvalThread e1), Any (EvalThread e2) -> CilType.Exp.compare e1 e2
       | Any (WarnGlobal vi1), Any (WarnGlobal vi2) -> compare (Hashtbl.hash vi1) (Hashtbl.hash vi2)
+      | Any (IterSysVars (vq1, vf1)), Any (IterSysVars (vq2, vf2)) -> VarQuery.compare vq1 vq2 (* not comparing fs *)
```
jerhard, line 289: Where do we use the `Any.compare` function? Was it for preventing circular query cycles or something?

src/analyses/accessAnalysis.ml
```diff
@@ -34,14 +57,28 @@ struct
...
     unsafe := 0
 
+  let side_vars ctx lv_opt ty =
+    match lv_opt with
+    | Some (v, _) ->
+      let d =
+        if !GU.should_warn then
+          G.create_vars (V0Set.singleton (lv_opt, ty))
+        else
+          G.bot () (* HACK: just to pass validation with MCP DomVariantLattice *)
```
jerhard, line 67: Could this hack hide unreached fixed-point problems?

src/solvers/td3.ml
```diff
@@ -905,6 +960,49 @@ module WP =
 
       print_data data "Data after postsolve";
 
+      (* TODO: example debug output for queries, remove *)
+      Cil.iterGlobals !Cilfacade.current_file (function
+          | GVar (g, _, _) ->
+            ignore (Pretty.printf "%a:\n" CilType.Varinfo.pretty g);
```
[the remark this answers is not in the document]
sim642: Yup, I still left it in in case anyone reviewing wants to see it in action. I'll make sure to remove it before the merge.

src/analyses/accessAnalysis.ml
sim642: Not really. The existing global invariant based access collecting already does the same below. This bottom side effect has absolutely no bad effect. The only problematic place would be the overridden `leq` implementation during `postsolving`. But that doesn't affect the fixpoint errors for any other analyses.

jerhard (approved):

PR #550, opened 2022-01-20
Filter non-racing accesses of racing variable

sim642:
Previously, if we determined a variable to have racing accesses, then we printed all of its accesses. This is excessive since there might be accesses that don't race with any other.

This PR filters the accesses of a racing variable to only contain those that actually race with another access to it. This is now possible due to the pairwise may-race checking and avoids some spurious warnings.

### DIFF
--- a/src/domains/access.ml
+++ b/src/domains/access.ml
@@ -333,8 +333,16 @@ struct
       let pretty = pretty
     end
     )
+
+  let conf (conf, _, _, _, _) = conf
 end
-module AS = SetDomain.Make (A)
+module AS =
+struct
+  include SetDomain.Make (A)
+
+  let max_conf accs =
+    accs |> elements |> List.map A.conf |> List.max
+end
 module T =
 struct
   include Printable.Std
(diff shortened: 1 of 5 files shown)


sim642:
Umm... what?! Why did pushing https://github.com/goblint/analyzer/commit/a9e03d83cf4c72b7b04cad17797831f68c71db0b to `filter-accesses` merge this PR?
Also the branch didn't actually get deleted.

Whatever, if there's a problem with this, we can still discuss it here and fix it directly on master.

michael-schwarz:
I was also surprised why you'd request a review and then merge it immediately :wink: 
But LGTM!

The only question is whether we may want to explicitly mention somewhere in the output that we only print accesses potentially involved in a race?

sim642:
I think it's fine not to, especially since the old behavior was more confusing: the race warning listed accesses that didn't actually race with anything and there was no mention of that.

Also, with `allglobs` everything still gets printed.

src/domains/access.ml
```diff
@@ -333,8 +333,16 @@ struct
...
     )
+
+  let conf (conf, _, _, _, _) = conf
+end
+module AS =
+struct
+  include SetDomain.Make (A)
+
+  let max_conf accs =
+    accs |> elements |> List.map A.conf |> List.max
```
michael-schwarz, line 344: I get a compile error here on my machine. Going to `(List.max ~cmp:Int.compare)` fixes it.

sim642: That's weird, there was a `List.max` in this file before this PR as well. Do you have an outdated version of batteries or something? Our lower bound is 3.4.0 and that's also locked and tested by the CI, where it works.

michael-schwarz: No, I have the most recent batteries. The same fix was also needed in the intDomain.ml at some point. Maybe it has to do with using it through `|>`?

sim642: That fix was exactly related to batteries 3.4.0.

sim642: Ahh, the typechecking of `|>` was specialized in OCaml 4.13: https://github.com/ocaml/ocaml/pull/10081.

PR #559, opened 2022-01-21
Fix address domain widening creating duplicates

### LINKED ISSUE #554: Fixpoint not reached with address sets
I extracted this small sample from https://github.com/goblint/bench/issues/7 via `creduce`.

~~~C
int a;

int main() { c(&a); }
int c(int *f) {
  unsigned long d;
  int *e = &a;
  d = 0;
  while (1) {
    e = f++;
    d++;
  }
}
~~~

The issue is `HoarePO: {a[def_exc:Unknown int([-63,63])], a, NULL, ?} not leq {a[def_exc:Unknown int([-63,63])], a[def_exc:Unknown int([-63,63])], NULL, ?}. `

Full message:
```
michael@michael-ThinkPad-X1-Carbon-6th:~/Documents/goblint-cil/analyzer$ ./goblint 1.c
Fixpoint not reached at node 84 "tmp = f;" on 1.c:10:5-10:12
 Solver computed:
 {([Reversed (top or Set (Normal Lvals * booleans)):{}, Unit:(),
    top or Set (variables):{}, booleans:False, MT mode:Singlethreaded,
    Thread * lifted created and Unit:(([main], {}), bot),
    value domain * array partitioning deps * Vars with Weak Update * P:(mapping {
                                                                          Global {
                                                                            a ->
                                                                              (0)
                                                                          }
                                                                          Local {
                                                                            d ->
                                                                              (Unknown int([0,64]))
                                                                            e ->
                                                                              {&a[def_exc:Unknown int([-63,63])], &a[def_exc:Unknown int([-63,63])], &a}
                                                                          }
                                                                          Parameter {
                                                                            f ->
                                                                              {&a[def_exc:Unknown int([-63,63])], &a}
                                                                          }
                                                                          Temp {
                                                                            tmp ->
                                                                              {&a[def_exc:Unknown int([-63,63])], &a[def_exc:Unknown int([-63,63])], NULL, ?}
                                                                          }
                                                                        }, mapping {
                                                                               }, {}, {}),
    lifted node:Unknown node, Unit:()], mapping {
                                          })}
 Right-Hand-Side:
 {([Reversed (top or Set (Normal Lvals * booleans)):{}, Unit:(),
    top or Set (variables):{}, booleans:False, MT mode:Singlethreaded,
    Thread * lifted created and Unit:(([main], {}), bot),
    value domain * array partitioning deps * Vars with Weak Update * P:(mapping {
                                                                          Global {
                                                                            a ->
                                                                              (0)
                                                                          }
                                                                          Local {
                                                                            d ->
                                                                              (Unknown int([0,8]))
                                                                            e ->
                                                                              {&a[def_exc:Unknown int([-63,63])], &a}
                                                                          }
                                                                          Parameter {
                                                                            f ->
                                                                              {&a[def_exc:Unknown int([-63,63])], &a}
                                                                          }
                                                                          Temp {
                                                                            tmp ->
                                                                              {&a[def_exc:Unknown int([-63,63])], &a, NULL, ?}
                                                                          }
                                                                        }, mapping {
                                                                               }, {}, {}),
    lifted node:Unknown node, Unit:()], mapping {
                                          })}
 Difference: ([Reversed (top or Set (Normal Lvals * booleans)):{}, Unit:(),
               top or Set (variables):{}, booleans:False,
               MT mode:Singlethreaded,
               Thread * lifted created and Unit:(([main], {}), bot),
               value domain * array partitioning deps * Vars with Weak Update * P:(mapping {
                                                                                     Global {
                                                                                       a ->
                                                                                         (0)
                                                                                     }
                                                                                     Local {
                                                                                       d ->
                                                                                         (Unknown int([0,8]))
                                                                                       e ->
                                                                                         {&a[def_exc:Unknown int([-63,63])], &a}
                                                                                     }
                                                                                     Parameter {
                                                                                       f ->
                                                                                         {&a[def_exc:Unknown int([-63,63])], &a}
                                                                                     }
                                                                                     Temp {
                                                                                       tmp ->
                                                                                         {&a[def_exc:Unknown int([-63,63])], &a, NULL, ?}
                                                                                     }
                                                                                   }, mapping {
                                                                                          }, {}, {}),
               lifted node:Unknown node, Unit:()], mapping {
                                                     }):
 not leq ([Reversed (top or Set (Normal Lvals * booleans)):{}, Unit:(),
           top or Set (variables):{}, booleans:False, MT mode:Singlethreaded,
           Thread * lifted created and Unit:(([main], {}), bot),
           value domain * array partitioning deps * Vars with Weak Update * P:(mapping {
                                                                                 Global {
                                                                                   a ->
                                                                                     (0)
                                                                                 }
                                                                                 Local {
                                                                                   d ->
                                                                                     (Unknown int([0,64]))
                                                                                   e ->
                                                                                     {&a[def_exc:Unknown int([-63,63])], &a[def_exc:Unknown int([-63,63])], &a}
                                                                                 }
                                                                                 Parameter {
                                                                                   f ->
                                                                                     {&a[def_exc:Unknown int([-63,63])], &a}
                                                                                 }
                                                                                 Temp {
                                                                                   tmp ->
                                                                                     {&a[def_exc:Unknown int([-63,63])], &a[def_exc:Unknown int([-63,63])], NULL, ?}
                                                                                 }
                                                                               }, mapping {
                                                                                      }, {}, {}),
           lifted node:Unknown node, Unit:()], mapping {
                                                 }) because Map: tmp =
 HoarePO: {a[def_exc:Unknown int([-63,63])], a, NULL, ?} not leq {a[def_exc:Unknown int([-63,63])], a[def_exc:Unknown int([-63,63])], NULL, ?}. 
```

It is a slightly different case than https://github.com/goblint/bench/issues/7 where both address sets are (at least textually) the same, so there might be several issues here or one issues is manifesting itself differently.

sim642:
Closes #554.

In the problematic `widen {a[def_exc:1], a} {a[def_exc:Not {0}([0,7])]}`, previously it iterated over the two elements of the old set and widened those with the single element of the new set. Those inner widenings gave identical results and both remained.

Now it iterates over the single element of the new set and widening that with the two elements just yields one element, avoiding the duplicate.

None of `HoarePO` contains explicit logic for avoiding duplicates though, it just uses lists for the buckets. So I cannot be fully sure that there aren't some other weird cases possible, where it can still break.


michael-schwarz:
Is the issue not even earlier, namely that `widen {a[def_exc:1], a} {a[def_exc:Not {0}([0,7])]}` violates the rule that the first argument of a widen needs to be less or equal to its second argument?

michael-schwarz:
If this property is violated here, this would result in loosing all elements in the left argument that are incomparable with all elements in the right set, right?

michael-schwarz:
But it does fix the issue we had for openSSL (and not just the reduced example).

sim642:
> Apparently `{a[def_exc:1], a} leq {a[def_exc:Not {0}([0,7])]}` does hold. Maybe it is just me, but it is this not a bit questionable? Why is a set that contains `a` without offset leq a set where `a[0]` is explicitly excluded?

That's a good point. I didn't really think about it and just thought that's due to some special address domain thing. So there might be more wrong here indeed.

Also, I suspect the `join` might be wrong too then because the right argument of `widen` already has the left one joined in, so `a` should've been joined into something, but apparently not?


> If this property is violated here, this would result in loosing all elements in the left argument that are incomparable with all elements in the right set, right?

Hmm, I think due to the right argument being a join, all the addresses from the left should be contained and thus comparable with something on the right, so nothing should disappear.

sim642:
I added some unit tests for the lval domain used for the elements of address domain and there's an even more fundamental issue I guess:
* `a` is equal to `a[0]`,
* `a[0] join a[1]` is equal to `a[top]`,
* `a join a[1]` is _not_ equal to `a[top]`, because the `join` raises `Uncomparable` instead.

I'm not completely sure anymore, what the intended semantics is supposed to be. Are `a` and `a[0]` supposed to be different representations of the same thing? Having multiple representations is always a pain to deal with.

sim642:
> Apparently `{a[def_exc:1], a} leq {a[def_exc:Not {0}([0,7])]}` does hold. Maybe it is just me, but it is this not a bit questionable? Why is a set that contains `a` without offset leq a set where `a[0]` is explicitly excluded?

I looked into this and the problem was that `a` is `leq` than `a[not 0]`, which it should not be. Turns out this was due to a bug in `DefExc.equal_to` with an exclusion set, which caused wrong results from some of the offset lattice operations dealing with offsets of differing length. This should be fixed according to the tests I added.

src/domains/hoareDomain.ml
```diff
@@ -27,10 +27,10 @@ struct
       | [] -> [e]
       | x::xs -> try op e x :: xs with Lattice.Uncomparable -> x :: join op e xs
 
-    (* widen element e with bucket using op *)
+    (* widen new(!) element e with old(!) bucket using op *)
     let rec widen op e = function
       | [] -> []
-      | x::xs -> try if E.leq e x then [op e x] else widen op e xs with Lattice.Uncomparable -> widen op e xs (* only widen if valid *)
+      | x::xs -> try if E.leq x e then [op x e] else widen op e xs with Lattice.Uncomparable -> widen op e xs (* only widen if valid *)
```
jerhard, line 33: The exception should be thrown by the `E.leq`, right? Maybe one can make that more obvious by doing  a match on the result of `leq` and to match both `false` and `exception Lattice.Uncomparable` to `widen op e xs`.

sim642: `E.leq` (which is `Lval.NormalLval.leq` for the address domain) doesn't throw anything, it just returns `false` for the uncomparable values. It's `op` that throws.
Also they aren't duplicating each other because not raising `Uncomparable` doesn't yet mean that the underlying widening is correctly called (arguments are not switched or it's another unrelated value that happens to be in the same `hash` bucket.

michael-schwarz (approved):
Does not solve all the strange things going on here, but at least the ones where we are clear on how to solve them. I'd advocate merging.

PR #560, opened 2022-01-21
Reenable and fix some OCaml warnings

sim642:
This reenables two OCaml warnings and fixes the code such that they aren't emitted:
1. `labels-omitted`. Using labels for calling functions with labelled arguments makes code more readable, especially when the labelled arguments are there for good reason, e.g. to distinguish two arguments of the same type (some string functions).
2. `unused-value-declaration`. I have no idea why this would be disabled because it immediately found loads of unused code and _numerous bugs_ where an `include` was accidentally overridding a definition.

### DIFF
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -43,8 +43,8 @@ end
 
 module Trivial (Val: Lattice.S) (Idx: Lattice.S): S with type value = Val.t and type idx = Idx.t =
 struct
-  let name () = "trivial arrays"
   include Val
+  let name () = "trivial arrays"
   type idx = Idx.t
   type value = Val.t
 
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -63,7 +63,6 @@ struct
   let fold_left2 f a x y = f a x y
 
   let set_inplace = set
-  let copy a = a
   let printXml f x = BatPrintf.fprintf f "<value>\n<map>\n<key>Any</key>\n%a\n</map>\n</value>\n" Val.printXml x
   let smart_join _ _ = join
   let smart_widen _ _ = widen
@@ -443,7 +442,6 @@ struct
   let length _ = None
 
   let set_inplace = set
-  let copy a = a
 
   let smart_op (op: Val.t -> Val.t -> Val.t) length ((e1, (xl1,xm1,xr1)) as x1) ((e2, (xl2,xm2,xr2)) as x2) x1_eval_int x2_eval_int =
     normalize @@
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -62,7 +62,6 @@ struct
   let fold_left f a x = f a x
   let fold_left2 f a x y = f a x y
 
-  let set_inplace = set
   let printXml f x = BatPrintf.fprintf f "<value>\n<map>\n<key>Any</key>\n%a\n</map>\n</value>\n" Val.printXml x
   let smart_join _ _ = join
   let smart_widen _ _ = widen
@@ -440,8 +439,6 @@ struct
       (Expp.top(), (v, v, v))
 
   let length _ = None
-
-  let set_inplace = set
 
   let smart_op (op: Val.t -> Val.t -> Val.t) length ((e1, (xl1,xm1,xr1)) as x1) ((e2, (xl2,xm2,xr2)) as x2) x1_eval_int x2_eval_int =
     normalize @@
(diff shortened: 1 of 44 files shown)


src/cdomains/arrayDomain.ml
```diff
@@ -43,8 +43,8 @@ end
 
 module Trivial (Val: Lattice.S) (Idx: Lattice.S): S with type value = Val.t and type idx = Idx.t =
 struct
-  let name () = "trivial arrays"
   include Val
+  let name () = "trivial arrays"
```
michael-schwarz, line 47: Nice catch!

src/util/oilParser.mly
```diff
@@ -176,7 +176,3 @@ attribute_value:
   | OIL_STRING   {String $1}
   | AUTO         {Auto}
 ;
-
-%%
-
-let parse_error s = print_endline s
```
michael-schwarz, line 182: Are you sure this is not a false positive? It seems to be the most recent definition of `parse_error`?

src/util/options.schema.json
```diff
@@ -738,6 +738,12 @@
               "type": "boolean",
               "default": true
             },
+            "strengthening": {
+              "title": "ana.apron.strengthening",
+              "description": "Apply strengthening in join for extra precision with heterogeneous environments. Expensive!",
+              "type": "boolean",
+              "default": false
+            },
```
michael-schwarz, line 746: Is it always sound to apply this? I am not 100% convinced if we do not rely on not having strengthening for some heterogeneous environments now?

michael-schwarz (approved):

src/util/oilParser.mly
sim642: It's not. I also wondered about this and indeed `parse_error` is a special thing you can define for ocamlyacc, but that has to be done in the header, not the trailer. I looked at the ocamlyacc output for this and it puts the default `parse_error` into its data structure before this definition, so it was completely useless.

I thought about moving this to the header, but since it never was there, nothing could possibly rely on it, so removing gives equivalent behavior. Not sure if actually overriding it wouldn't do something worse.

src/util/options.schema.json
sim642: The algorithm of strengthening contains a soundness check within. It just tries reasserting all constraints from both arguments and checks if the result is still sound.
None of the privatization stuff if affected by this anyway because there the environments are homogeneous.

michael-schwarz: Ok! Since we now have that option, should we maybe add a test for it then?

sim642: Good idea! I don't remember what the original examples were, where this was needed (possibly some initial write-centered Apron thing), but I found an example where `protection` needs it to keep precision.

sim642:
@jerhard This deletes some nontrivial functions from `IntDomain`: `range_min_max`, `lift_comp` and `bitcomp`. Is it correct that these have been replaced with more specific implementations, leaving the generic ones unused? Or maybe they are accidentally unused?

src/cdomains/intDomain.ml
```diff
@@ -1296,15 +1277,15 @@ struct
     | `Excluded (s,l) when S.is_empty s -> "Unknown int" ^ short_size l
     (* Prepend the exclusion sets with something: *)
     | `Excluded (s,l) -> "Not " ^ S.show s ^ short_size l
+
+  include Std (struct type nonrec t = t let name = name let top_of = top_of let bot_of = bot_of let show = show let equal = equal end)
+  (* FIXME: poly compare? *)
```
jerhard, line 1282: `Std` does not contain a `compare` function anymore, so one can remove the comment.

sim642: Felt like I had already done it, but apparently that's in #542. Did it here as well.

jerhard:
>This deletes some nontrivial functions from IntDomain: `range_min_max`, `lift_comp` and `bitcomp`

>`bitcomp` in `Congruences`

This was supposed to implement some logic for bitwise operations on congruences presented in https://www.dsi.unive.it/~avp/domains.pdf  `bitor`, `bitand`, `bitxor`. It was not implemented correctly so we went with an simpler approach. 

>`lift_comp` in `DefExc`

This was a replica of `lift2`, which is used for `lt`, `le`, `gt` and `ge`, while `eq` and `ne` are customly coded.

>` range_min_max` in DefExc

Was replaced by functions `max_of_range`, `min_of_range`.

jerhard:
One can also remove `range` from `Size` if one also removes it from the `.mli`. 
Now there's a couple of functions in `Size` that end with `big_int` that do not have a non-big-int equvialent, so maybe we should rename them?

sim642:
> One can also remove `range` from `Size` if one also removes it from the `.mli`.
> Now there's a couple of functions in `Size` that end with `big_int` that do not have a non-big-int equvialent, so maybe we should rename them?

I went ahead and did both here.

jerhard (approved):

PR #561, opened 2022-01-21
Direct conversion between Zarith and Apron via `Z_mlgmpidl`

sim642:
Both Zarith and Apron (through Mlgmpidl) use GMP to work with big ints, but they aren't directly compatible. Thus so far we converted between them via strings, which is a bit silly given that the underlying GMP data is the same.

I discovered that the Zarith repo contains a module called `Z_mlgmpidl`, which actually provides these conversions. Unfortunately the Zarith opam package does not include this module (probably because that would require Zarith depending on Mlgmpidl mostly unnecessarily).

To still use these conversions, I have just vendored that tiny module into our codebase such that we can use those direct conversions. The two C functions `ml_z_mlgmpidl_of_mpz` and `ml_z_mlgmpidl_set_mpz`, which it's implemented with are still provided by Zarith using GMP's C interface directly.

I doubt this makes any real performance difference, but at least it's more satisfying than string-based conversion.


michael-schwarz:
Nice, the conversion through strings was very ugly.

Did you check whether the changes we make for gobview to Zarith (see https://github.com/goblint/Zarith/commit/db75b80a67173826aa02fdd5b514c819791f603c) also enforce that values that are too large are not represented as OCaml values if they are created here?

(Technically not relevant yet as Gobview can not work with the Apron analysis anyway for now, but I guess it would still be good to do this).

sim642:
> Did you check whether the changes we make for gobview to Zarith (see [goblint/Zarith@db75b80](https://github.com/goblint/Zarith/commit/db75b80a67173826aa02fdd5b514c819791f603c)) also enforce that values that are too large are not represented as OCaml values if they are created here?

Looking at the two special functions I mentioned above, the conversion calls `ml_z_reduce` eventually, which reduces the number to direct ints based on the bounds in macro definitions. So that should account for the reduced bounds in our fork.

michael-schwarz (approved):

PR #563, opened 2022-01-21
Loop Unrolling for the first `exp.unrolling-factor` Iterations

mikcp:
**Implementation of the loop unrolling functionality.** 

- `exp.unrolling-factor` is 0 by default, case in which the unrolling does not take place.
- if `exp.unrolling-factor`>0, all loops will be unrolled _factor_ times.

This is the first part of a precision optimization. Unrolling the loop's body _k_ times outside the loop, will allow us to see those _k_ iterations more precisely.


michael-schwarz:
Could you maybe also add tests for this feature?

One should be able to observe a precision increase for some programs even without the modified array domain.
E.g.  programs where Goblint is newly able to show precise asserts about a variable that is modified in the loop other than the ones used in the loop head, provided the loop goes through less iterations than `exp.unrolling-factor`?

src/util/cilfacade.ml
```diff
@@ -79,6 +79,81 @@ let end_basic_blocks f =
   let thisVisitor = new allBBVisitor in
   visitCilFileSameGlobals thisVisitor f
 
+class loopUnrollingVisitor = object
+  inherit nopCilVisitor
+  method! vstmt s =
+    (* All the duplicated stmts need to be re-created or the sid will fail*)
+    let newsid st = { st with sid = st.sid } in
```
sim642, line 86: This looks fishy. I guess the point is to create a copy of `st`, which is physically unequal to it? Is there some guarantee that the OCaml compiler won't just optimize this to identity?

Also, if this is used to copy the body of the loop however many times necessary, the copies need to have distinct `sid`s because CFG nodes are compared based on the `sid` of the statement. If they have the same `sid`s then the CFG won't contain any copies. Or am I wrong?

src/util/cilfacade.ml
```diff
@@ -79,6 +79,81 @@ let end_basic_blocks f =
...
+      | If(e,b1,b2,l1,l2) -> mkStmt(If(e,(mkb b1.bstmts),(mkb b2.bstmts),l1,l2))
+      | Loop(bl,l1,l2,s1,s2) -> 
+        let create_stmt = mkStmt(Loop((mkb bl.bstmts),l1,l2,s1,s2)) in
+        create_stmt.labels <- st.labels;
+        create_stmt
+      | Block(bl) -> mkStmt(Block(mkb bl.bstmts))
+      | Switch(e,bl,stl,l1,l2) -> mkStmt(Switch(e,(mkb bl.bstmts),(List.map newsid_stmt stl),l1,l2))
+      | _ -> newsid st in
+    let mkBlock' b = mkBlock (List.map newsid_stmt b) in
+    let mk_stmt_from_stmt_list bl = mkStmt(Block(mkBlock' bl)) in
```
sim642, line 100: This looks like part of `copyFunctionVisitor` in CIL, which seems to handle `Goto`s for the copies as well. Maybe something could be extracted from there.

src/util/cilfacade.ml
```diff
@@ -79,6 +79,81 @@ let end_basic_blocks f =
...
+    match s.skind with
+    | Loop(b, loc, _, _, _) ->
+      let get_unrolling_factor = GobConfig.get_int "exp.unrolling-factor" in
+      (* All unrollings will leave the original loop at the end. The label makes sure it's not unrolled twice.*)
+      let add_label_to_remainder_loop st_loop loc_loop = 
+        st_loop.labels <- [Label(Cil.freshLabel "remainder_loop", loc_loop, false)];
```
sim642, line 106: This removes all the existing labels on `st_loop`!

src/util/cilfacade.ml
```diff
@@ -79,6 +79,81 @@ let end_basic_blocks f =
...
+        match loop_stmt_labels with
+        | [] -> false
+        | Label(lab,_,_)::tl ->
+          begin
+            try
+              match String.sub lab 0 14 with
+              | "remainder_loop" -> true
+              | _ -> false
+            with _-> false
+          end
```
sim642, line 118:
```suggestion
          BatString.starts_with lab "remainder_loop"
```

sim642: Also, this doesn't recurse into `tl`, so it just checks the first label. I guess this could be `List.exists` instead.

src/util/cilfacade.ml
```diff
@@ -79,6 +79,81 @@ let end_basic_blocks f =
...
+        |0-> sl
+        |_ ->
+          let x = body @ sl in
+          unroll x (factor-1) in
+      let unroll_helper st = 
+        let x = mk_stmt_from_stmt_list (unroll [(add_label_to_remainder_loop st loc)] get_unrolling_factor) in
+        mkStmt (Block (mkBlock [x;break_stmt])) in
+      let is_loop_unrollable s = 
+        if is_remainder_loop s.labels then false
+        else if get_unrolling_factor=0 then false
```
sim642, line 145: `loop_unrolling` isn't even called in this case, so the check is redundant.

src/util/cilfacade.ml
```diff
@@ -79,6 +79,81 @@ let end_basic_blocks f =
...
+      let break_stmt = mkStmt (Instr []) in
+      break_stmt.labels <- [Label(Cil.freshLabel "unroll_while_break",loc,false)] ;
+      let rec rm_breaks_st st = 
+        let rm_breaks_st_list stl = mkBlock (List.map rm_breaks_st stl) in
+        match st.skind with
+        | Break(l) -> mkStmt (Goto (ref break_stmt, loc))
+        | If(e,tb,fb,l1,l2) -> mkStmt (If (e,(rm_breaks_st_list tb.bstmts), (rm_breaks_st_list fb.bstmts),l1,l2))
+        | Block(bl) -> mkStmt (Block (rm_breaks_st_list bl.bstmts))
+        |Switch(e,bl,stl,l1,l2) -> mkStmt (Switch (e, (rm_breaks_st_list bl.bstmts), (List.map rm_breaks_st stl),l1, l2))
+        | _ -> st in
```
sim642, line 131: Why not just run this visitor _after_ `preparedCFG` and not reinvent the wheel?
Also this doesn't handle `continue`, which would have to jump to the next unrolled loop iteration.

sim642 (changes requested):

src/util/cilfacade.ml
michael-schwarz: >  I guess the point is to create a copy of st, which is physically unequal to it?

Yes. It is an idiom that is used in several places in CIL. I didn't find a more explicit way to make a copy of a record with mutable members, but according not the manual this indeed creates a (shallow) copy, so I think the compiler may not optimize this out.

michael-schwarz: > If they have the same sids then the CFG won't contain any copies. Or am I wrong?

According to the documentation http://goblint.in.tum.de/assets/goblint-cil/api/Cil.html#TYPEstmt this should only be computed later, right?

> A number (>= 0) that is unique in a function. Filled in only after the CFG is computed.

sim642: You're right, the OCaml manual says this creates a shallow copy, so it should be fine. But I think there should be a comment to clarify what it's achieving, because otherwise someone might be tempted at some point to "optimize" it away.

Indeed, the statements are really numbered by `computeCFGInfo`. So considering my remark below about doing this after `prepareCFG` to normalize breaks, continues, etc. This visitor should be invoked between the two.

src/util/cilfacade.ml
```diff
@@ -107,12 +163,21 @@ let createCFG (fileAST: file) =
...
+  (* The unrolling needs to take place between prepareCFG and computeCFGInfo, since prepareCFG handles the breaks     *)
+  (* and continues, which allow us to replicate the loop's body outside of the loop. computeCFG info creates the CFG, *)
+  (* so it needs to happen before that.                                                                               *)
+  if (get_int "exp.unrolling-factor")>0 then loop_unrolling fileAST;
+  iterGlobals fileAST (fun glob ->
+    match glob with
+    | GFun(fd,_) ->
+      computeCFGInfo fd true
+    | _ -> ()
+  );
```
sim642, line 180: I think this could still be done within a single `iterGlobals`, but the loop unrolling visitor should be applied with `visitCilFunction` instead of `visitCilFileSameGlobals`.

src/util/cilfacade.ml
mikcp: I took a look at it and I don't see a way to use the `copyFunctionVisitor` directly. The only option I see is replicating what happens there, which is what I did. Do you maybe have a suggestion on how to approach this?

sim642: It probably cannot be used directly, because it's meant for copying entire functions and that's how it's also exposed from CIL. But I was thinking something along the lines of copying its code from CIL here (as `copyStmtVisitor` or something) and simplifying it a bit (removing the function-specific parts and maybe removing `Switch` support, which are handled before by `prepareCFG` anyway).

The nice thing about such visitor is that it doesn't need to handle all `stmtkind`s separately and recurse into them, but the visitor logic itself already takes care of all of that. And the `patches` component of it could also be kept to fix `Goto`s for the copies. I think that will inevitably be needed because the prepared switches and loops contain them. I haven't tried, but I feel like currently things might go wrong if the unrolled loop contains a switch or another loop.

src/util/cilfacade.ml
mikcp: When I changed the loop unrolling call so it was called after `prepareCFG` it broke the nested loops use case. I was not able to fix it so I changed the old version so it supports `continue` instructions.

michael-schwarz:
I just started another sv-comp run with an unrolling of 5. If there are no abnormalities there, I think we can merge this.

sim642:
> If there are no abnormalities there, I think we can merge this.

I would still be hesitant about that, regarding the unresolved points above. Not handling gotos like `copyFunctionVisitor` in unrolled loops could be very problematic, it might not just show as unsoundness on sv-benchmarks.
Also duplicating the functionality of `prepareCFG` and this not working right when relying on `prepareCFG` instead is again quite suspicious. It could also be related to the gotos that `prepareCFG` introduces.

sim642:
> I just started another sv-comp run with an unrolling of 5.

Also, this is a bit laughable considering Frama-C-SV used loop unrolling of 1024 according to their tool paper.

michael-schwarz:
A run with an unrolling of 5 reveals ~48 cases where Goblint now succeeds (including these `product-lines/elevator_spec13_product23.cil.yml` like tasks), and ~33 cases where we newly fail.

An interesting class of programs for which we newly fail contain loops of the flavor:

~~~C
int main()
{
	int current_execution_context;
	unsigned short random;

	do
	{
		random = __VERIFIER_nondet_ushort();

		switch (random)
		{
		case 1:
			current_execution_context = 1;

			break;

		default:
			break;
		}
	} while (random);
}
~~~

Without unrolling, everything works. With it Goblint fails with:

```
Fatal error: exception Failure("MyCFG.createCFG.find_real_stmt: >1 succ")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from CfgTools.createCFG.find_real_stmt in file "src/framework/cfgTools.ml", line 230, characters 14-44
Called from Stdlib__list.map in file "list.ml", line 92, characters 20-23
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from CfgTools.createCFG.(fun).handle in file "src/framework/cfgTools.ml", line 309, characters 48-61
Called from Stdlib__list.iter in file "list.ml", line 110, characters 12-15
Called from CfgTools.createCFG.(fun) in file "src/framework/cfgTools.ml", line 383, characters 8-37
Called from Stdlib__list.iter in file "list.ml", line 110, characters 12-15
Called from Cil.iterGlobals in file "src/cil.ml", line 5635, characters 2-29
Called from CfgTools.createCFG in file "src/framework/cfgTools.ml", line 239, characters 2-1023
Called from CfgTools.getCFG in file "src/framework/cfgTools.ml", line 593, characters 19-33
Called from Control.compute_cfg in file "src/framework/control.ml", line 671, characters 19-39
Called from Control.analyze in file "src/framework/control.ml", line 682, characters 21-37
Called from Maingoblint.do_analyze.do_all_phases.do_one_phase in file "src/maingoblint.ml", line 357, characters 12-48
Re-raised at Maingoblint.do_analyze.do_all_phases.do_one_phase in file "src/maingoblint.ml", line 366, characters 10-51
Called from BatEnum.iter.loop in file "src/batEnum.ml", line 315, characters 4-16
Called from BatEnum.iter in file "src/batEnum.ml", line 319, characters 4-10
Called from Maingoblint.do_analyze.do_all_phases in file "src/maingoblint.ml", line 378, characters 16-68
Called from Stats.repeattime.repeatf in file "src/ocamlutil/stats.ml", line 106, characters 10-15
Re-raised at Stats.repeattime.repeatf in file "src/ocamlutil/stats.ml", line 110, characters 1-8
Called from Maingoblint.main in file "src/maingoblint.ml", line 510, characters 6-34
Called from Stdlib.at_exit.new_exit in file "stdlib.ml", line 554, characters 59-63
Called from Stdlib.do_at_exit in file "stdlib.ml" (inlined), line 560, characters 20-61
Called from Std_exit in file "std_exit.ml", line 18, characters 8-20
```

Any idea what's happening there @mikcp?

michael-schwarz:
The corresponding tables if anyone is interested:
[default-vs-unroll5-2min-all.zip](https://github.com/goblint/analyzer/files/8079853/default-vs-unroll5-2min-all.zip)

michael-schwarz:
More natural example of the same issue

~~~C
int main(void) {
	for (int i = 0; i < 2; ++i) {
		switch (i) {
		case 0:
			break;
		case 1:
		}
	}

	return 0;
}
~~~

sim642:
I wouldn't be the least surprised if the crash was due to the lacking goto handling, because the invariants here come from CIL itself:
https://github.com/goblint/analyzer/blob/a0bb6404d724ef42b7ce3a957a8ba2f1ecb635b5/src/framework/cfgTools.ml#L195-L209
If the AST is additionally screwed with without maintaining those, things are bound to be wrong.

michael-schwarz:
I have now re-implemented it. They key idea now is: We still do the transformation before `prepareCfg` to avoid complicated reasoning about which `goto`s were originally `goto`s and which gotos correspond to loop control structures and hence need to be modified in some way.

This comes at the price of putting each replicated body into a fake `while(1)` loops, inserting an appropriate `break` at the end of the replicated bodies and patching all top-level `continue`s to be breaks. Apart from that we remove all labels from the unrolled loop, getting rid of issues with duplicate labels.

The resulting code is a bit ugly, but the unnecessary loops will hopefully not lead to widening as there is no back edge, and the resulting code is cleaner than the approach originally taken for the thesis.

src/util/cilfacade.ml
```diff
@@ -80,71 +80,45 @@ let end_basic_blocks f =
   let thisVisitor = new allBBVisitor in
   visitCilFileSameGlobals thisVisitor f
 
+class removeLabelsVisitor = object
+  inherit nopCilVisitor
+
+  val nestingDepth = ref 0
```
sim642, line 86: I think the more canonical way would be `mutable val nestingDepth = 0`, which needs no dereferencing and can be updated with `<-` just like mutable record fields.

src/util/cilfacade.ml
```diff
@@ -80,71 +80,45 @@ let end_basic_blocks f =
...
+
+  method! vstmt s =
+    let new_labels = List.filter_map (function Label(str,loc,b) -> None | x -> Some x) s.labels in
+    match s.skind with
+    | Loop _ -> nestingDepth := !nestingDepth+1; ChangeDoChildrenPost({s with labels = new_labels}, fun x -> nestingDepth := !nestingDepth-1; x)
+    | Continue loc -> if !nestingDepth = 0 then
+        ChangeTo({s with labels = new_labels; skind = Break loc})
```
sim642, line 93: What's this about? It's not just removing labels.

src/util/cilfacade.ml
```diff
@@ -80,71 +80,45 @@ let end_basic_blocks f =
...
-      | If(e,b1,b2,l1,l2) -> mkStmt(If(e,(mkb b1.bstmts),(mkb b2.bstmts),l1,l2))
-      | Loop(bl,l1,l2,s1,s2) ->
-        let create_stmt = mkStmt(Loop((mkb bl.bstmts),l1,l2,s1,s2)) in
-        create_stmt.labels <- st.labels;
-        create_stmt
-      | Block(bl) -> mkStmt(Block(mkb bl.bstmts))
-      | Switch(e,bl,stl,l1,l2) -> mkStmt(Switch(e,(mkb bl.bstmts),(List.map newsid_stmt stl),l1,l2))
-      | _ -> newsid st in
-    let create_shallow_copies bl = List.map newsid_stmt bl in
+    let labelsRemover = new removeLabelsVisitor in
```
sim642, line 108: This could instead be a `val` in the `loopUnrollingVisitor` class. Currently it allocates a new object for each visited unconditionally.

src/util/cilfacade.ml
```diff
@@ -80,6 +80,51 @@ let end_basic_blocks f =
...
+  inherit nopCilVisitor
+    
+  method! vstmt s =
+    let labelsRemover = new removeLabelsVisitor in
+    match s.skind with
+    | Loop (b,loc, loc2, break , continue) ->
+      let duplicate_and_rem_labels s =
+        match s.skind with
+        | Loop (b,loc, loc2, break , continue) ->
+          let s = { s with sid = s.sid } in
```
sim642, line 114: The comments for these are now missing.

src/util/cilfacade.ml
```diff
@@ -80,6 +80,51 @@ let end_basic_blocks f =
...
+          let factor = GobConfig.get_int "exp.unrolling-factor" in
+          let bstmts = b.bstmts @ [mkStmt @@ Break loc] in
+          let copies = List.init (factor) (fun _ -> visitCilStmt labelsRemover (mkStmt (Loop (mkBlock bstmts,loc,loc2,None,None)))) in
```
sim642, line 117: This calls `labelsRemover` `factor` times, instead of doing it just once. Could just do with `BatList.make`.

src/util/cilfacade.ml
michael-schwarz: All unrollings are now in a `while(true)` loop. Top-level `Continue`s inside this loop now should continue to the next (unrolled) iteration, i.e. break out of the infinite loop.

src/util/cilfacade.ml
michael-schwarz: what do you mean by this?

src/util/cilfacade.ml
michael-schwarz: Yes, and this is by design! We need fresh copies of the loop body, otherwise we run into issues with sharing again.

src/util/cilfacade.ml
sim642: Before your rewrite there were comments which made this trick explicit: https://github.com/goblint/analyzer/pull/563#discussion_r789786142.

src/util/cilfacade.ml
michael-schwarz: Yes, I always forget what is syntactic sugar for what mutable record field for references or the other way around.

src/util/cilfacade.ml
sim642: Won't the top level breaks then be wrong? They would have to jump over all the loops.

src/util/cilfacade.ml
sim642: Oh, so `removeLabelsVisitor` implicitly does a lot more than just remove labels. It also copies the entire statement recursively. It's not just that `{s with labels = new_labels}` in the default case there is used to remove the labels, but in all the cases where there are no labels, this still guarantees a fresh physical instance of the record.

I think all this physical instance trickery should be explicitly commented because it's key to all this working but very easy to accidentally want to optimize away.

src/util/cilfacade.ml
michael-schwarz: Yes, you are right of course. I had thought about needing it and the somehow forgot. I guess this is part of the reason, why it still so simple - simply wrong :wink:

src/util/cilfacade.ml
```diff
@@ -80,6 +80,109 @@ let end_basic_blocks f =
...
+        x
+    in
+    let rename_labels sn =
+      let new_labels = List.map (function Label(str,loc,b) -> Label (Cil.freshLabel str,loc,b) | x -> x) sn.labels in
+      (* this makes new physical copy*)
+      let new_s = {sn with labels = new_labels} in
+      if new_s.labels <> [] then
+        (* Use original s, ns might be temporay e.g. if the type of statement changed *)
+        (* record that goto s; appearing in the current fragment should later be patched to goto new_s *)
+        Hashtbl.add gotos (ref s) (ref new_s);
```
sim642, line 129: The keys of the hashtable are _mutable_ references? That's a big no, because if such reference actually changes, then the (polymorphic) hash for it could change and break the hashtable lookups.

michael-schwarz: We have the guarantee that these do not change during this phase though (as it would break invariants in a lot of other places). I don't think there is anything else one can use to uniquely identify statements.
In principle one could make an association list based on `==` of `stmt`, but that would be linear in the lookup, so I'm not sure that this is much better.

michael-schwarz: I might try going via the names then maybe, somehow feels a bit less elegant. Any other suggestions?

sim642: If they need to be identified by physical equality, then a `Hashtbl.Make` with the following in the argument module would make that more explicit:
https://github.com/goblint/analyzer/blob/2669f675b7f64f92e1f5db876c429b68487ec67e/src/framework/cfgTools.ml#L59-L61
Of course that still assumes for `hash` that the fields aren't mutated, but at least there isn't an additional wrapper reference to mutate.

michael-schwarz: :+1:

sim642 (approved):

PR #568, opened 2022-01-25
Add ikind argument to IntDomain arbitraries, fix some bugs

sim642:
### TODO
Enabling `ILong` ikind for IntDomain property tests still fails one test for intervals:
```
test `integerset -> intervals (long): valid rem` failed on ≥ 1 cases:
([-1L], [-1L]) (after 124 shrink steps)
```

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -1358,7 +1358,7 @@ struct
     let norm ik v =
       match v with
       | `Excluded (s, r) ->
-        let possibly_overflowed = not (R.leq r (size ik)) in
+        let possibly_overflowed = not (R.leq r (size ik)) || not (S.for_all (in_range (size ik)) s) in
         (* If no overflow occurred, just return x *)
         if not possibly_overflowed then (
           v
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -123,6 +123,7 @@ sig
   val of_bool: bool -> t
   val of_interval: Cil.ikind -> int_t * int_t -> t
   val of_congruence: Cil.ikind -> int_t * int_t -> t
+  val arbitrary: unit -> t QCheck.arbitrary
 end
 (** Interface of IntDomain implementations that do not take ikinds for arithmetic operations yet.
    TODO: Should be ported to S in the future. *)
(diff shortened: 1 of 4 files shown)


sim642:
Apparently `rem` with definite intervals `-1L` fails due to this change by @vogler: https://github.com/goblint/analyzer/commit/a58611994c731f72ebfc5b53ca9dbf59298246dc.
Reverting to the old implementation (adapted to `ikind` arguments and bigints) works:
```ocaml
let y' = sub ik y (of_int ik (Ints_t.one)) in
meet ik (bit (fun _ik -> Ints_t.rem) ik x y) (join ik y' (neg ik y'))
```

I don't completely understand what the new version is trying to do, but I think it manages to construct `range` to be `[0, -1]`, which is invalid and messes something up. This could either be because it never checks if `xu` is negative (only `xl`) or because of the off-by-one technique used, which in this special case gives `b = 0`.

src/cdomains/intDomain.ml
```diff
@@ -2231,7 +2231,7 @@ struct
     let congruence_series a c m =
       let rec next a1 c1 a2 c2 =
         if a2 |: a1 then (a2, c2)
-        else next a2 c2 (a1 %: a2) ((c1 -: c2) *: (a1 /: a2))
+        else next a2 c2 (a1 %: a2) (c1 -: (c2 *: (a1 /: a2)))
```
michael-schwarz, line 2234: What is the intuition behind this change?

sim642: There were test failures with `meet`, like it not being even commutative! So I dug up the referenced paper and saw that this was just implemented incorrectly.

michael-schwarz: Ok, I guess it is like that in [Granger89], so I guess it was simply a typo in the implementation.

Since we have also discovered typos in the Granger paper, I'd be even more comfortable if we manage to construct an example where the old result leads to unsoundness.

Should be fairly easy given the output of the unit tests?

sim642: If I change it back now, the failures are:
```
test `congruences (int): meet assoc` failed on ≥ 1 cases:
(3ℤ, 4ℤ, 1+7ℤ) (after 235 shrink steps)

test `congruences (int): meet comm` failed on ≥ 1 cases:
(2ℤ, 1+29ℤ) (after 145 shrink steps)

test `congruences (int): meet leq` failed on ≥ 1 cases:
(8ℤ, 1+3ℤ) (after 158 shrink steps)
```
It could just be that the difference was in precision, not soundness, because all the abstraction validity properties still pass.

sim642:
I wonder what we should do about this. Should we try to fix the failures here or should we merge this and list the still disabled int domain failures in a new issue? Because besides `ILong` the tests also disable all unsigned ikinds as for some reason most domains fail there.

michael-schwarz:
I'd propose merging and opening a new issue for the other ones.

PR #569, opened 2022-01-26
Remove `SafePtr`, replace singleton lists with option in `Lval`

sim642:


src/util/gobOption.ml
```diff
@@ -0,0 +1,9 @@
+open Prelude
+
+let exists p = function
+  | Some x -> p x
+  | None -> false
+
+let for_all p = function
+  | Some x -> p x
+  | None -> true
```
michael-schwarz, line 9: These are very confusing, in what sense is this an `exists` or a `for_all`?

sim642: In the sense that `option` is a singleton list or an empty list. This is a fairly standard generalization of `exists` and `for_all` to other data structures. For example Scala and Haskell have such operations for their optional types.

src/analyses/base.ml
```diff
@@ -371,9 +364,9 @@ struct
       in
       let f x =
         match Addr.to_var_offset x with
-        | [x] -> f_addr x                    (* normal reference *)
-        | _ when x = Addr.NullPtr -> VD.bot () (* null pointer *)
-        | _ -> `Int (ID.top_of IChar)       (* string pointer *)
+        | Some x -> f_addr x                    (* normal reference *)
+        | None when x = Addr.NullPtr -> VD.bot () (* null pointer *)
+        | None -> `Int (ID.top_of IChar)       (* string pointer *)
```
michael-schwarz, line 369: Why can we be sure it is a StringPointer here? Could it not be a UnknownPointer?

michael-schwarz (approved):

src/analyses/base.ml
sim642: Good point! There is a comment below:

> Finally we join over all the addresses in the set. If any of the addresses is a topped value, joining will fail.

But I don't think it's even true, because I don't see why `AD.fold` should throw if one of the elements is `Addr.UnknownPtr`. According to git blame this code and comments originate from the beginning of time. I suspect `AD` might have had a lifted top way back (so couldn't have known addresses beside unknown) and then the comments would've been true. But maybe since making unknown pointer another element in the set, this became wrong but was never noticed.

I suppose for unknown pointer this could (and should) just return `VD.top ()`, but I wonder if there's any actual code involving an unknown pointer that does something wrong because of this oversight.

sim642: I opened #588 about this to investigate how big of an impact this could have.

PR #570, opened 2022-01-26
Optimize short-circuitable folds in `MCP`

sim642:
This PR is an attempt to optimize the following `MCP` domain/context/access functions:
* `equal`
* `compare`
* `leq`
* `is_top`
* `is_bot`
* `may_race`

The previous implementations used folds, which inevitably had to iterate over the entire list each time. Even though the folding functions themselves used the short-circuiting binary operators, this didn't short-circuit much. Moreover, all the inefficient `List.assoc` calls were still happening just to pass the corresponding arguments to the folding function (which in the short-circuiting case didn't even use them).
These `assoc`s are a pain because as the fold goes down the list, each corresponding `assoc` needs to go through more of the other list. Overall, this means O(n²) complexity.

This PR implements alternative short-circuitable folds for `MCP`, which abort the list iteration via local exception to short-circuit. It not only avoids the unnecessary list tail iteration, but also all the `assoc` calls that would unnecessarily happen there.

This is quite relevant because while profiling slow race warning performance that @vesalvojdani noticed, it turns out that 30% of race warning time is just spent on `MCPAccess.A.compare` while manipulating sets of accesses.
I'm speculating that this might even have general performance improvement since path-sensitivity also uses sets of `MCP.D` values.

### TODO
- [x] Benchmark.

### DIFF
--- a/src/analyses/mCPRegistry.ml
+++ b/src/analyses/mCPRegistry.ml
@@ -119,9 +119,6 @@ struct
       (name, S.to_yojson (obj x)) :: a
     in `Assoc (unop_fold f [] xs)
 
-  let binop_fold f a (x:t) (y:t) =
-    GobList.fold_left3 (fun a (n,d) (n',d') (n'',s) -> assert (n = n' && n = n''); f a n s d d') a x y (domain_list ())
-
   let binop_for_all f (x:t) (y:t) =
     GobList.for_all3 (fun (n,d) (n',d') (n'',s) -> assert (n = n' && n = n''); f n s d d') x y (domain_list ())
 
@@ -260,9 +257,6 @@ struct
 
   let binop_for_all f (x:t) (y:t) =
     GobList.for_all3 (fun (n,d) (n',d') (n'',s) -> assert (n = n' && n = n''); f n s d d') x y (domain_list ())
-
-  let unop_fold f a (x:t) =
-    fold_left2 (fun a (n,d) (n',s) -> assert (n = n'); f a n s d) a x (domain_list ())
 
   let unop_for_all f (x:t) =
     List.for_all2 (fun (n,d) (n',s) -> assert (n = n'); f n s d) x (domain_list ())
--- a/src/analyses/mCPRegistry.ml
+++ b/src/analyses/mCPRegistry.ml
@@ -122,7 +122,11 @@ struct
   let binop_fold f a (x:t) (y:t) =
     GobList.fold_left3 (fun a (n,d) (n',d') (n'',s) -> assert (n = n' && n = n''); f a n s d d') a x y (domain_list ())
 
-  let equal   x y = binop_fold (fun a n (module S : Printable.S) x y -> a && S.equal (obj x) (obj y)) true x y
+  let binop_for_all f (x:t) (y:t) =
+    GobList.for_all3 (fun (n,d) (n',d') (n'',s) -> assert (n = n' && n = n''); f n s d d') x y (domain_list ())
+
+  let equal   x y = binop_for_all (fun n (module S : Printable.S) x y -> S.equal (obj x) (obj y)) x y
+  (* TODO: something for compare? *)
   let compare x y = binop_fold (fun a n (module S : Printable.S) x y -> if a <> 0 then a else S.compare (obj x) (obj y)) 0 x y
 
   let hashmul x y = if x=0 then y else if y=0 then x else x*y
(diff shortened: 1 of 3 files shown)


src/analyses/mCPRegistry.ml
```diff
@@ -251,24 +273,41 @@ struct
...
+    try if length x <> length y
+      then raise (DomListBroken "binop_short_fold_bool : differing lengths")
+      else (
+        iter (fun (n,d) -> f n d @@ assoc n y) x;
+        a
+      )
+    with
+      | Not_found -> raise (DomListBroken "binop_short_fold_bool : assoc failure")
+      | ShortCircuitBool b -> b
+
```
michael-schwarz, line 289: Do we really need to give `binop_short_fold_bool` both in `DomListPrintable` and in `DomListLattice`?

sim642: It's for the same reason there were `binop_fold` in both before: one uses first class modules of type `(module Printable.S)` and the other of `(module Lattice.S)`.

michael-schwarz:
If it is really the calls to `assoc` that are biting us here, one could also think about at least reducing the complexity of `assoc_dom` calls by introducing some dedicated data structure for it that makes accesses `O(1)`.

As the list of all available analyses is typically small, does not change during analysis, and numbered starting at zero, one could think about generating a `Module option array` where the index corresponds to the number of the analysis?

sim642:
> If it is really the calls to `assoc` that are biting us here, one could also think about at least reducing the complexity of `assoc_dom` calls by introducing some dedicated data structure for it that makes accesses `O(1)`.
> 
> As the list of all available analyses is typically small, does not change during analysis, and numbered starting at zero, one could think about generating a `(module Bla.S) option array` where the index corresponds to the number of the analysis?

I've been thinking about this as well, but also the fact that the whole assoc list representation in MCP might need overhauling, for example to use `Map` instead. Another possibility would be to ensure that analyses in those lists have well-defined and consistent order (e.g. sorted), so they could simply be zipped. I'll open a new issue about this.

sim642:
On sv-benchmarks there's not much of a difference, maybe ~2% improvement:
image
[Full results table](https://goblint.cs.ut.ee/results/59-all-fast-mcp-short-fold-after/results/results.2022-01-27_13-51-13.table.html#/scatter?toolY=1&columnY=1&scaling=Linear&results=All&regression=Linear&filter=0(0*status*(status(in(unknown,true,EXCEPTION%20(Failure),EXCEPTION,EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError))))),1(0*status*(status(notIn(ERROR%20(verify),EXCEPTION%20(Stack%20overflow),OUT%20OF%20MEMORY,TIMEOUT,TIMEOUT%20(true)))))&line=1.1)

But there also isn't big multithreaded benchmarks there, so the differences aren't as pronounced as in #571.

sim642:
I would merge this after #578 because there are major conflicts between the two.

sim642:
I now rebased this on top of #578 and also simplified the implementations: instead of using exceptions to escape the fold, I'm now just using `for_all*` functions on lists, which also stop iterating on first `false`.

I also benchmarked this new version and it's much more impressive than the previous version.

### This compared to #578
There's an **additional ~10% speedup:**
image

### #578 and this compared to master
Together the two give **~27% speedup:**
image

[Full results table](https://goblint.cs.ut.ee/results/62-all-fast-mcp-no-assoc-after-short-circuit/results/results.2022-01-31_14-04-22.table.html#/scatter?toolY=2&columnY=1&toolX=0&columnX=1&results=All&filter=0(0*status*(status(in(unknown,true,EXCEPTION,EXCEPTION%20(Failure),EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError))))),1(0*status*(status(in(unknown,true,EXCEPTION,EXCEPTION%20(Failure),EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError))))),2(0*status*(status(in(unknown,true,EXCEPTION%20(Failure),EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError)))))&regression=Linear&scaling=Linear&line=2)

michael-schwarz:
Nice! I'm always amazed at how much time we seem to waste in non-obvious places.

michael-schwarz (approved):

PR #573, opened 2022-01-27
Add option for manual restarting of globals

### LINKED ISSUE #543: Minimal incremental restarting of sided variables
It was discussed during GobCon that a minimal sided variable restarting could also be possible, as opposed to doing it transitively during `destabilize_with_side`. Since you already discussed this on the Munich side and probably have a better idea how it should work/help, I'd leave it up to you.

jerhard:
Add option to restart unknowns belonging to a list of global variables selected by the user.
The list can be passed with `incremental.restart_globs.globs` (which should be renamed before merging this).
 
The option is marked to conflict with `incremental.restart.enabled` which performs an automatic restarting.
 
Closes #543. Depends on #549.


src/util/server.ml
```diff
@@ -113,7 +113,8 @@ let analyze ?(reset=false) ({ file; do_analyze; _ }: t)=
     | Some solver_data ->
       let changes = CompareCIL.compareCilFiles file file in
       let old_data = Some { Analyses.cil_file = file; solver_data } in
-      { Analyses.changes; old_data; new_file = file }, false
+      (* TODO: get globals for restarting from config *)
```
sim642, line 116: Shouldn't this logic be deduplicated?

src/solvers/td3.ml
```diff
@@ -719,6 +719,44 @@ module WP =
...
+          destab_side_dep x;
+          destabilize x;
+          restart_leaf x
+        in
+        let globals_to_restart = S.increment.restarting in
+        let get x = try HM.find rho x with Not_found -> S.Dom.bot () in
+        List.iter
+          (fun g ->
+            (* All gs arriving here should yield Some varinfo. *)
+            let g = Option.get @@ Cilfacade.varinfo_from_global g in
```
sim642, line 747: How about just putting the `varinfo`s into the restarting list in the first place? This is kind of repeating the lookup.

src/solvers/td3.ml
```diff
@@ -719,6 +719,44 @@ module WP =
...
+            | Some w ->
+              HM.remove side_dep x;
+              VS.iter (fun y ->
+                  HM.remove stable y;
+                  destabilize y;
+                ) w;
+            | None ->
+              ignore @@ Pretty.printf
+                "Could not find globals in side_dep, so side_dep %a is not restarted.\n"
+                S.Var.pretty_trace x
```
sim642, line 736: It is restarted though: `restart_leaf` is called below regardless of this case. But I guess then the side effects are not retriggered and the analysis becomes unsound?

src/solvers/td3.ml
```diff
@@ -719,6 +719,44 @@ module WP =
...
+          destabilize x;
+          restart_leaf x
+        in
+        let globals_to_restart = S.increment.restarting in
+        let get x = try HM.find rho x with Not_found -> S.Dom.bot () in
+        List.iter
+          (fun g ->
+            (* All gs arriving here should yield Some varinfo. *)
+            let g = Option.get @@ Cilfacade.varinfo_from_global g in
+            ignore @@ Pretty.printf "Restarting global %s.\n" g.vname;
```
sim642, line 748: `restart_leaf` already prints this.

src/solvers/td3.ml
```diff
@@ -719,6 +719,44 @@ module WP =
...
+        (* Must be called on a leaf! *)
+        let destabilize_leaf (x : S.v) =
+          let destab_side_dep (x : S.v) =
+            match HM.find_option side_dep x with
+            | Some w ->
+              HM.remove side_dep x;
+              VS.iter (fun y ->
+                  HM.remove stable y;
+                  destabilize y;
+                ) w;
```
sim642, line 732: This is missing a lot of the `side_dep` logic that `destabilize_with_side` has. In particular the superstable and abort front stuff. I think that part of `destabilize_with_side` could just be copied here and removing the recursive call.

src/maingoblint.ml
```diff
@@ -433,8 +433,8 @@ let check_arguments () =
...
   (* order matters: non-ptr=false, int=true -> int=false cascades to interval=false with warning *)
   if get_bool "ana.base.context.interval" && not (get_bool "ana.base.context.int") then (set_bool "ana.base.context.interval" false; warn "ana.base.context.interval implicitly disabled by ana.base.context.int");
-  if get_bool "incremental.only-rename" then (set_bool "incremental.load" true; warn "incremental.only-rename implicitly activates incremental.load. Previous AST is loaded for diff and rename, but analyis results are not reused.")
-
+  if get_bool "incremental.only-rename" then (set_bool "incremental.load" true; warn "incremental.only-rename implicitly activates incremental.load. Previous AST is loaded for diff and rename, but analyis results are not reused.");
+  if get_bool "incremental.restart.sided.enabled" && get_string_list "incremental.restart_globs.globs" <> [] then (warn "Passing a non-empty list to incremental.restart_globs.globs (manual restarting) while incremental.restart.sided.enabled (automatic restarting) is activated. Aborting."; exit 5 )
```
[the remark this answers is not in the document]
jerhard: >Is it actually incompatible though? You could still manually restart globals that the change-based sided did not.

`destabilize_leaf` calls `destabilize`, which is changed to `destabilize_with_side`, in case that `restart_sided` is activated.  This changes the behavior of the manual restarting so that it restarts more. Of course this could be changed, but it seems to me that one would usually not want to have both the automatic and the manual restarting. And adding code for that case that resets the `destabilize` back and forth would make the code a bit more convoluted.

I guess instead of aborting, one could just give a warning here.

tests/incremental/04-manual-restart/02-read_write_global.c
```diff
@@ -0,0 +1,16 @@
+int g = 0;
+
+void foo(){
+    g = 1;
+}
+
+void bar(){
+    int x = g;
+    assert(x % 2 == 0); // UNKNOWN
```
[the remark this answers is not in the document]
jerhard: ><span class="x x-first x-last">!</span>

Is that HTML code something annotation I am not aware of, or was it just  accidental?

Also, in the concrete this is not unknown, but actually will always fail. `earlyglobs` is the reason why it is unknown in the analysis.

src/solvers/td3.ml
jerhard: The warning is indeed unnecessary, because there may be globals that are never written by a function, but have an initializer; so I removed the warning.

tests/incremental/04-manual-restart/02-read_write_global.c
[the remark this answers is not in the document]
sim642: Then it should probably say `UNKNOWN (earlyglobs)` to be specific.

src/util/cilfacade.ml
```diff
@@ -21,6 +21,41 @@ let rec get_labelsLoc = function
...
+module String = struct
+  include String
+  let hash = Hashtbl.hash
+end
+
+module SM = Hashtbl.Make(String)
+let varinfo_map: Cil.global SM.t option ref = ref None
+
+let varinfo_from_global (g : Cil.global) : Cil.varinfo option = match g with
+  | GFun (f, _) -> Some f.svar
```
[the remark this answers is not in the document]
jerhard: I added a check in `destabilize_leaf` (taken from `destabilize_with_side`) that only performs restarting on global variables.

src/solvers/td3.ml
```diff
@@ -719,6 +719,40 @@ module WP =
...
+              HM.remove side_dep x;
+              (* add side_dep to front to prevent them from being aborted *)
+              destabilize_front ~front:true x w;
+              (* destabilize side dep to redo side effects *)
+              VS.iter (fun y ->
+                  if tracing then trace "sol2" "destabilize_leaf %a side_dep %a\n" S.Var.pretty_trace x S.Var.pretty_trace y;
+                  if tracing then trace "sol2" "stable remove %a\n" S.Var.pretty_trace y;
+                  HM.remove stable y;
+                  HM.remove superstable y;
+                  destabilize y
```
sim642, line 737: Should this be `destabilize` or `destabilize_normal` like below?

src/solvers/td3.ml
```diff
@@ -719,6 +719,40 @@ module WP =
...
+          destabilize_normal x
+
+        in
+        let globals_to_restart = S.increment.restarting in
+        let get x = try HM.find rho x with Not_found -> S.Dom.bot () in
+
+        List.iter
+          (fun g ->
+            S.iter_vars get g
+               (fun v -> if HM.mem stable v then destabilize_leaf v)
```
sim642, line 752: Maybe this should also check if `v` is a leaf (`S.system v = None`) because it might be that at some point there's a varquery which returns something local as well.

src/util/cilUtil.ml
```diff
@@ -0,0 +1,43 @@
+(** Functionality to create [VarQuery.t] lists from a list of names of global variable and function definitions *)
+(* Separate file from cilfacade.ml, for functions that depend both on [Node] and [VarQuery]. Avoids dependency cycle *)
```
sim642, line 2: Couldn't all of this just go into `VarQuery` itself?
`CilUtil` is quite confusing considering there's also `Cilfacade` in `src/util/`.

src/util/cilUtil.ml
```diff
@@ -0,0 +1,43 @@
...
+              | true -> vq::acc
+              | _ -> acc
+            end
+          | None, None -> acc
+          | _, _ -> assert false
+        ) [] in
+    globals
+end
+
+include M
```
sim642, line 43: Why this submodule? The restricting signature should just go into a `.mli` file.

tests/incremental/04-manual-restart/03-partial_contexts.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git tests/incremental/04-manual-restart/03-partial_contexts.json tests/incremental/04-manual-restart/03-partial_contexts.json
+index 99334fccb..3da6121a1 100644
+--- tests/incremental/04-manual-restart/03-partial_contexts.json
++++ tests/incremental/04-manual-restart/03-partial_contexts.json
+@@ -13,7 +13,7 @@
+             }
+         },
+         "restart_globs": {
+-            "globs": []
++            "globs": ["g"]
```
sim642, line 23: There's no global `g` in this program, so I guess this is testing nothing?
Also might be useful to then have an error when no globals have the names given here.

src/solvers/td3.ml
jerhard: Good catch!

src/solvers/td3.ml
jerhard: I am now checking this and print some warning when the unknown is not  a leaf.

tests/incremental/04-manual-restart/03-partial_contexts.patch
jerhard: It should restart the start node of `foo`, so I changed the patch accordingly. When it is `g`, the assert is `Unknown`. 

For names that do not have a matching global in the CIL-file, it now also emits a warning.

src/framework/varQuery.mli
```diff
@@ -0,0 +1,6 @@
+type t =
+    Global of Cil.varinfo
+  | Node of { node : Node.t; fundec : Cil.fundec option; }
+val compare : t -> t -> int
+type 'v f = 'v -> unit
+val varqueries_from_names : Cil.file -> string list -> t list * string list
```
sim642, line 6: I think now that there's a `.mli` file for this module, the documentation comments `(** ... *)` should be moved from the implementation to the interface, otherwise they aren't displayed as documentation anywhere.

sim642 (approved):

src/util/server.ml
jerhard: @sim642  You marked that as resolved --- I am not sure whether it is. Or is the idea that this should be deduplicated by #591?

sim642: I think the TODO might be fine for the time being. The deduplication doesn't strictly need to happen in #591, but it's something we'll have to do soon anyway to a common reusable module for both. Ideally there would be no real logic in `Maingoblint` left, except for command-line arguments handling.

PR #577, opened 2022-01-27
Array unrolling abstract domain

mikcp:
**Implementation of the unrolling array domain**

Third abstract array domain in use. `FlagConfiguredArrayDomain` will now choose between one of three array domains (`PartitionedWithLength`, `TrivialWithLength` and `Unroll`) depending on the value of `exp.arrays-domain`.

When `unroll` is chosen, `exp.array-unrolling-factor` need to be set as well. The array representation will then consist of a tuple _(xl, xr)_. 

- _xl_ = list containing the first k values of the array.
- _xr_ = single abstract value representing the rest of the array. 

A new module `ProdList` was add in `lattice.ml` to help with the `xl` part of the unroll domain. 

All uses and references to `ana.base.partition-arrays.enabled` have been substituted by `exp.arrays-domain` = "partitioned".

Tests for this domain have also been added under _tests/regression/54-unroll_arrays_

This is the second part of a precision optimization. Displaying the first `k` iterations of the array more precisely, will help increase the precision in some cases. Specially if used together with loop unrolling.

### DIFF
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -74,7 +74,7 @@ end
 module Unroll (Val: Lattice.S) (Idx:IntDomain.Z): S with type value = Val.t and type idx = Idx.t =
 struct
   module Factor = struct let x () = (get_int "exp.array-unrolling-factor") end
-  module Base = Lattice.ProdArray (Val) (Factor)
+  module Base = Lattice.ProdList (Val) (Factor)
   include Lattice.ProdSimple(Base) (Val)
 
   let name () = "unrolled arrays"
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -93,7 +93,7 @@ struct
   let pretty_diff () (x,y) = dprintf "%s: %a not leq %a" (name ()) pretty x pretty y
   let extract x = match x with
       | Some c -> c
-      | None -> Z.zero
+      | None -> failwith "arrarDomain: that sould not happen"
   let get (ask: Q.ask) (xl, xr) (_,i) =
     let search_unrolled_values min_i max_i =
       let mi = Z.to_int min_i in
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -865,7 +865,7 @@ struct
   let smart_leq f g = binop' (P.smart_leq f g) (T.smart_leq f g) (U.smart_leq f g)
   let update_length newl x = unop_to_t' (P.update_length newl) (T.update_length newl) (U.update_length newl) x
 
-  (* Functions that make us of the configuration flag *)
+  (* Functions that make use of the configuration flag *)
   let chosen_domain () = get_string "ana.base.arrays.domain"
 
   let name () = "FlagConfiguredArrayDomain: " ^ match chosen_domain () with
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -139,7 +139,7 @@ struct
     else ((update_unrolled_values min_i (Z.of_int ((factor ())-1))), (Val.join xr v))
   let make _ v =
     let xl = BatList.make (factor ()) v in
-    (xl,v)
+    (xl,Val.bot ())
   let length _ = None
   let move_if_affected ?(replace_with_const=false) _ x _ _ = x
   let get_vars_in_e _ = []
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -777,11 +777,49 @@ struct
   let to_yojson (x, y) = `Assoc [ (Base.name (), Base.to_yojson x); ("length", Idx.to_yojson y) ]
 end
 
+module UnrollWithLength (Val: Lattice.S) (Idx: IntDomain.Z): S with type value = Val.t and type idx = Idx.t =
+struct
+  module Base = Unroll (Val) (Idx)
+  include Lattice.Prod (Base) (Idx)
+  type idx = Idx.t
+  type value = Val.t
+
+  let get (ask : Q.ask) (x, (l : idx)) ((e: ExpDomain.t), v) =
+    (array_oob_check (module Idx) (x, l) (e, v));
+    Base.get ask x (e, v)
+  let set (ask: Q.ask) (x,l) i v = Base.set ask x i v, l
+  let make l x = Base.make l x, l
+  let length (_,l) = Some l
+
+  let move_if_affected ?(replace_with_const=false) _ x _ _ = x
+  let map f (x, l):t = (Base.map f x, l)
+  let fold_left f a (x, l) = Base.fold_left f a x
+  let fold_left2 f a (x, l) (y, l) = Base.fold_left2 f a x y
+  let get_vars_in_e _ = []
+
+  let smart_join _ _ = join
+  let smart_widen _ _ = widen
+  let smart_leq _ _ = leq
+
+  (* It is not necessary to do a least-upper bound between the old and the new length here.   *)
+  (* Any array can only be declared in one location. The value for newl that we get there is  *)
+  (* the one obtained by abstractly evaluating the size expression at this location for the   *)
+  (* current state. If newl leq l this means that we somehow know more about the expression   *)
+  (* determining the size now (e.g. because of narrowing), but this holds for all the times   *)
+  (* the declaration is visited. *)
+  let update_length newl (x, l) = (x, newl)
+
+  let printXml f (x,y) =
+    BatPrintf.fprintf f "<value>\n<map>\n<key>\n%s\n</key>\n%a<key>\n%s\n</key>\n%a</map>\n</value>\n" (XmlUtil.escape (Base.name ())) Base.printXml x "length" Idx.printXml y
+
+  let to_yojson (x, y) = `Assoc [ (Base.name (), Base.to_yojson x); ("length", Idx.to_yojson y) ]
+end
+
 module FlagConfiguredArrayDomain(Val: LatticeWithSmartOps) (Idx:IntDomain.Z):S with type value = Val.t and type idx = Idx.t =
 struct
   module P = PartitionedWithLength(Val)(Idx)
   module T = TrivialWithLength(Val)(Idx)
-  module U = Unroll(Val)(Idx)
+  module U = UnrollWithLength(Val)(Idx)
 
   type idx = Idx.t
   type value = Val.t
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -137,8 +137,8 @@ struct
     if Z.geq min_i f then (xl, (Val.join xr v))
     else if Z.lt max_i f then ((update_unrolled_values min_i max_i), xr)
     else ((update_unrolled_values min_i (Z.of_int ((factor ())-1))), (Val.join xr v))
-  let make i v =
-    let xl = Array.to_list (Array.make (factor ()) v) in
+  let make _ v =
+    let xl = BatList.make (factor ()) v in
     (xl,v)
   let length _ = None
   let move_if_affected ?(replace_with_const=false) _ x _ _ = x
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -808,7 +808,7 @@ struct
   (* Simply call appropriate function for component that is not None *)
   let get a x (e,i) = unop' (fun x ->
     if e = `Top then
-      let e' = BatOption.map_default (fun x -> `Lifted (Cil.kinteger64 IInt x)) (`Top) (Option.map BI.to_int64 @@ Idx.to_int i) in
+      let e' = BatOption.map_default (fun x -> `Lifted (Cil.kintegerCilint (Cilfacade.ptrdiff_ikind ()) x)) (`Top) (Idx.to_int i) in
       P.get a x (e', i)
     else
       P.get a x (e, i)
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -146,8 +146,6 @@ struct
   let map f (xl, xr) = ((List.map f xl), f xr)
   let fold_left f a x = f a (join_of_all_parts x)
   let fold_left2 f a x y = f a (join_of_all_parts x) (join_of_all_parts y)
-  let set_inplace = set
-  let copy a = a
   let printXml f (xl,xr) = BatPrintf.fprintf f "<value>\n<map>\n
   <key>unrolled array</key>\n
   <key>xl</key>\n%a\n\n
@@ -792,11 +790,6 @@ struct
     let msg = "FlagConfiguredArrayDomain received a value where not exactly one component is set"
     let name = "FlagConfiguredArrayDomain"
   end
-
-  let of_t = function
-  | (Some p, None) -> (Some p, None, None)
-  | (None, Some (t,u)) -> (None, t, u)
-  | _ -> failwith "FlagConfiguredArrayDomain received a value where not exactly one component is set"
 
   let to_t = function
   | (Some p, None, None) -> (Some p, None)
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -71,6 +71,93 @@ struct
   let update_length _ x = x
 end
 
+module Unroll (Val: Lattice.S) (Idx:IntDomain.Z): S with type value = Val.t and type idx = Idx.t =
+struct
+  module Factor = struct let x () = (get_int "exp.array-unrolling-factor") end
+  module Base = Lattice.ProdArray (Val) (Factor)
+  include Lattice.ProdSimple(Base) (Val)
+
+  let name () = "unrolled arrays"
+  type idx = Idx.t
+  type value = Val.t
+  let factor () = get_int "exp.array-unrolling-factor"
+  let join_of_all_parts (xl, xr) = List.fold_left Val.join xr xl
+  let show (xl, xr) =
+    let rec show_list xlist = match xlist with
+      | [] -> " --- "
+      | hd::tl -> (Val.show hd ^ " - " ^ (show_list tl)) in
+    "Array (unrolled to " ^ (Stdlib.string_of_int (factor ())) ^ "): " ^
+    (show_list xl) ^ Val.show xr ^ ")"
+  let pretty () x = text "Array: " ++ text (show x)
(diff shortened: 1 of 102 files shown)


src/util/options.schema.json
```diff
@@ -1407,6 +1400,20 @@
...
+            "The domain that should be used for arrays. trivial/partitioned/unroll. When employing the partition array domain, make sure to enable the expRelation analysis as well. When employing the unrolling array domain, make sure to set the exp.array-unrolling-factor >0.",
+          "type": "string",
+          "default": "trivial"
+        },
+        "array-unrolling-factor": {
+          "title": "exp.array-unrolling-factor",
+          "description":
+            "Indicates how many values will the unrolled part of the unrolled array domain contain.",
+          "type": "integer",
+          "default": 0
```
sim642, line 1416: These should be moved back under `ana.base`, for example:
* `ana.base.arrays.domain` (to match `ana.base.structs.domain`),
* `ana.base.arrays.unrolling-factor`.

src/util/options.schema.json
```diff
@@ -1407,6 +1400,20 @@
             "Path to C preprocessor (cpp) to use. If empty, then automatically searched.",
           "type": "string",
           "default": ""
+        },
+        "arrays-domain": {
+          "title": "exp.arrays-domain",
+          "description":
+            "The domain that should be used for arrays. trivial/partitioned/unroll. When employing the partition array domain, make sure to enable the expRelation analysis as well. When employing the unrolling array domain, make sure to set the exp.array-unrolling-factor >0.",
+          "type": "string",
```
sim642, line 1408: Since there are only fixed string values acceptable for this option, they should be specified by the schema:
```suggestion
          "type": "string",
          "enum": ["trivial", "partitioned", "unroll"],
```

tests/regression/54-unroll_arrays/01-simple_array.c
```diff
@@ -0,0 +1,40 @@
+// PARAM: --set solver td3 --enable ana.int.interval --set exp.arrays-domain unroll --set exp.array-unrolling-factor 5
+int global;
```
sim642, line 2: Here and in the other added test:
```suggestion
#include <assert.h>
int global;
```

sv-comp/benchexec/my-bench/goblint.xml
```diff
@@ -11,7 +11,7 @@
   <option name="--enable">ana.int.interval</option>
   <!-- <option name="-sets solver td3"/> -->
   <option name="--enable">ana.context.widen</option>
-  <option name="--enable">ana.base.partition-arrays.enabled</option>
+  <option name="-sets exp.arrays-domain partitioned"/>
```
sim642, line 14: This doesn't actually matter. This file isn't used for anything any more and BenchExec doesn't support setting three-argument options from XML, so this choice would be done by the svcomp conf files anyway.

src/domains/lattice.ml
```diff
@@ -581,6 +581,33 @@ struct
     Pretty.dprintf "%a not leq %a" pretty x pretty y
 end
 
+module type Num = sig val x : unit -> int end
+module ProdList (Base: S) (N: Num) =
+struct
+  include Printable.Liszt (Base)
+
+  let bot () = Array.to_list (Array.make (N.x ()) (Base.bot ()))
```
sim642, line 589: Pointless intermediate array, use `BatList.make` instead.

src/domains/lattice.ml
```diff
@@ -581,6 +581,33 @@ struct
...
+  let is_bot =
+    let f acc x = Base.is_bot x && acc in
+    List.fold_left f true
```
sim642, line 592: This is just `List.for_all`.

src/cdomains/valueDomain.ml
```diff
@@ -148,10 +148,10 @@ struct
     | TComp ({cstruct=true; _} as ci,_) -> `Struct (Structs.create (fun fd -> init_value fd.ftype) ci)
     | TComp ({cstruct=false; _},_) -> `Union (Unions.top ())
     | TArray (ai, None, _) ->
-      `Array (CArrays.make (IndexDomain.bot ())  (if get_bool "ana.base.partition-arrays.enabled" then (init_value ai) else (bot_value ai)))
+      `Array (CArrays.make (IndexDomain.bot ())  (if (get_string "exp.arrays-domain"="partitioned" || get_string "exp.arrays-domain"="unroll") then (init_value ai) else (bot_value ai)))
```
sim642, line 151: @michael-schwarz This is starting look like a leaky abstraction. Couldn't the array domains implement it in their corresponding modules? Currently the array domain selection is repeated 5 times instead of 1.

src/cdomains/arrayDomain.ml
```diff
@@ -693,55 +783,84 @@ module FlagConfiguredArrayDomain(Val: LatticeWithSmartOps) (Idx:IntDomain.Z):S w
...
-  let smart_leq f g = binop (P.smart_leq f g) (T.smart_leq f g)
-  let update_length newl x = unop_to_t (P.update_length newl) (T.update_length newl) x
-
-  (* Functions that make use of the configuration flag *)
-  let name () = "FlagConfiguredArrayDomain: " ^ if get_bool "ana.base.partition-arrays.enabled" then P.name () else T.name ()
-
-  let partition_enabled () = get_bool "ana.base.partition-arrays.enabled"
+  let get a x (e,i) = unop' (fun x ->
+    if e = `Top then
+      let e' = BatOption.map_default (fun x -> `Lifted (Cil.kinteger64 IInt x)) (`Top) (Option.map BI.to_int64 @@ Idx.to_int i) in
```
sim642, line 818: This is reverting #510 and #514, but shouldn't.

src/cdomains/arrayDomain.ml
```diff
@@ -69,6 +69,96 @@ struct
...
+  module Factor = struct let x () = (get_int "exp.array-unrolling-factor") end
+  module Base = Lattice.ProdList (Val) (Factor)
+  include Lattice.ProdSimple(Base) (Val)
+
+  let name () = "unrolled arrays"
+  type idx = Idx.t
+  type value = Val.t
+  let factor () = 
+    match get_int "exp.array-unrolling-factor" with
+    | 0 -> failwith "ArrayDomain: exp.array-unrolling-factor needs to be set when using the unroll domain"
```
sim642, line 83: Why is 0 not valid? Shouldn't it just give equivalent behavior to the trivial array domain with all elements joined?

src/cdomains/arrayDomain.ml
```diff
@@ -69,6 +69,96 @@ struct
...
+      if mi=ma then full_update xl 0
+      else weak_update xl 0 in
+    let f = Z.of_int (factor ()) in
+    let min_i = extract(Idx.minimal i) in
+    let max_i = extract(Idx.maximal i) in
+    if Z.geq min_i f then (xl, (Val.join xr v))
+    else if Z.lt max_i f then ((update_unrolled_values min_i max_i), xr)
+    else ((update_unrolled_values min_i (Z.of_int ((factor ())-1))), (Val.join xr v))
+  let make i v =
+    let xl = Array.to_list (Array.make (factor ()) v) in
```
sim642, line 141: `BatList.make` here as well.

Also, isn't the `i` argument the array length? If so, then that's currently ignored, but feels like maybe it shouldn't. For example, if the unrolling factor is 1000, but the program contains an array of length 10, then it's quite wasteful to create a massive 1000 element list. Not only does it consume memory, but all the lattice operations act elementwise on all the 1000 elements each time.

src/cdomains/arrayDomain.ml
```diff
@@ -69,6 +69,96 @@ struct
...
+    (xl,v)
+  let length _ = None
+  let move_if_affected ?(replace_with_const=false) _ x _ _ = x
+  let get_vars_in_e _ = []
+  let map f (xl, xr) = ((List.map f xl), f xr)
+  let fold_left f a x = f a (join_of_all_parts x)
+  let fold_left2 f a x y = f a (join_of_all_parts x) (join_of_all_parts y)
+  let set_inplace = set
+  let copy a = a
```
sim642, line 150: These are unused (#560):
```suggestion
```

sim642 (changes requested):
The CI shows two tests failing:
```
2 test(s) failed: ["22/17 large_arrays", "54/01 simple_array"]
```

src/util/options.schema.json
mikcp: Fixed on commit ec075afd3fe3b71d6d58ea278861e2a5063b2d5c (array-domain and array-unrolling factor flags moved under ana.base. All references fixed)

src/util/options.schema.json
mikcp: Fixed on d7fed6da2acc18c9211cce3cd5db38a47d37e5f5

tests/regression/54-unroll_arrays/01-simple_array.c
mikcp: Fixed on 1ad669514c033c4d1791608a3d8a64d06d1ba6d0

sv-comp/benchexec/my-bench/goblint.xml
mikcp: Fixed on d0f99ffeec2abff3b5e230d32268be769a5c3696

src/domains/lattice.ml
mikcp: Fixed on 68a4ea357119629e5288312b6cd654afda60ac43

src/domains/lattice.ml
mikcp: Fixed on cddbe1a915f6704d7226f549abac76bdadf73bba

src/cdomains/arrayDomain.ml
mikcp: Fixed on 7f5529c3a5996a4ac3019aadeadbb9428165d3ff

src/cdomains/arrayDomain.ml
mikcp: Fixed on 90173324d5943b9fb12d4d98f95a57e29be45815

src/cdomains/arrayDomain.ml
mikcp: This is one of the updates that we talked could be implemented on the future; having a general unrolling factor, and then an heuristic that could change that value for specific arrays.
Right now we are assuming the unrolling factor assigned would not be too high, since otherwise using Lists doesn't make much sense either.

mikcp: `BatList.make` fix added on commit bebf563d0ab7b45c1b58874a29351b84f736e2c9

src/cdomains/valueDomain.ml
michael-schwarz: I agree that it is a bit messy. To avoid this, we would have to make the array domains not take a `Lattice.S` but a new module type that also exposes `init_value` and `top_value` (or just add it to the `LatticeWithSmartOps` and use that one everywhere), and then in turn include `init_value` and `top_value` again in the signature of the `ArrayDomain`.

I added a note in https://github.com/goblint/analyzer/issues/485 that we want to do this together with the general cleanup of the partitioned array domain.

src/cdomains/arrayDomain.ml
```diff
@@ -69,6 +69,94 @@ struct
...
+    let rec show_list xlist = match xlist with
+      | [] -> " --- "
+      | hd::tl -> (Val.show hd ^ " - " ^ (show_list tl)) in
+    "Array (unrolled to " ^ (Stdlib.string_of_int (factor ())) ^ "): " ^
+    (show_list xl) ^ Val.show xr ^ ")"
+  let pretty () x = text "Array: " ++ text (show x)
+  let pretty_diff () (x,y) = dprintf "%s: %a not leq %a" (name ()) pretty x pretty y
+  let extract x = match x with
+      | Some c -> c
+      | None -> Z.zero
```
michael-schwarz, line 96: When is this ever `None`? And if yes, why is `Z.zero` the right choice in that case both for maximum and minimum?

src/cdomains/arrayDomain.ml
```diff
@@ -69,6 +69,94 @@ struct
...
+  let get (ask: Q.ask) (xl, xr) (_,i) =
+    let search_unrolled_values min_i max_i =
```
michael-schwarz, line 98: Would it not make sense to also check if the meet of the constant we are currently considering during this search and the value passed for the index is non-bot here?
Consider e.g. the case where I have DefExc enabled on top of intervals and know the value is in the range from `1` to `10` but definitely not `8` and have the unrolling set to `10`.

src/cdomains/arrayDomain.ml
```diff
@@ -69,6 +69,94 @@ struct
...
+      let mi = Z.to_int min_i in
+      let ma = Z.to_int max_i in
+      let rec subjoin l i = match l with
+        | [] -> Val.bot ()
+        | hd::tl ->
+          begin
+            match i>ma,i<mi with
+            | false,true -> subjoin tl (i+1)
+            | false,false -> Val.join hd (subjoin tl (i+1))
+            | _,_ -> Val.bot ()
```
michael-schwarz, line 108: When exactly does this happen? If it shouldn't happen, we should maybe `raise` here?

src/cdomains/arrayDomain.ml
```diff
@@ -69,6 +69,94 @@ struct
...
+  let set (ask: Q.ask) (xl,xr) (_,i) v =
+    let update_unrolled_values min_i max_i =
+      let mi = Z.to_int min_i in
+      let ma = Z.to_int max_i in
+      let rec weak_update l i = match l with
+        | [] -> []
+        | hd::tl ->
+          if i<mi then hd::(weak_update tl (i+1))
+          else if i>ma then (hd::tl)
+          else (Val.join hd v)::(weak_update tl (i+1)) in
```
michael-schwarz, line 126: Same here, one could once again check if the meet is non-bot to avoid updating certain elements

src/cdomains/arrayDomain.ml
```diff
@@ -693,55 +781,79 @@ module FlagConfiguredArrayDomain(Val: LatticeWithSmartOps) (Idx:IntDomain.Z):S w
...
+  let to_t = function
+  | (Some p, None, None) -> (Some p, None)
+  | (None, Some t, None) -> (None, Some (Some t, None))
+  | (None, None, Some u) -> (None, Some (None, Some u))
+  | _ -> failwith "FlagConfiguredArrayDomain received a value where not exactly one component is set"
+
+  module I = struct include LatticeFlagHelper (T) (U) (K) let name () = "" end
+  include LatticeFlagHelper (P) (I) (K)
+
+  let binop' ops ophs opks = binop ops (I.binop ophs opks)
```
michael-schwarz, line 803: Why `ops`, `ophs`, and `opks`? These made sense for the struct domain with

~~~OCaml
  module S = Simple(Val)
  module HS = Sets(Val)
  module KS = KeyedSets(Val)
 ~~~
 
 Here, they should probably be named `opp` `opt` and `opu`.

src/cdomains/arrayDomain.ml
```diff
@@ -693,55 +781,79 @@ module FlagConfiguredArrayDomain(Val: LatticeWithSmartOps) (Idx:IntDomain.Z):S w
 struct
   module P = PartitionedWithLength(Val)(Idx)
   module T = TrivialWithLength(Val)(Idx)
+  module U = Unroll(Val)(Idx)
```
michael-schwarz, line 784: Is there a particular reason why we do not have a `WithLength` variant of `Unroll`? As it currently stands, if I enable `Unroll`, I will get warnings about out-of-bounds accesses everywhere, which is not really desirable.

src/cdomains/arrayDomain.ml
```diff
@@ -693,55 +781,79 @@ module FlagConfiguredArrayDomain(Val: LatticeWithSmartOps) (Idx:IntDomain.Z):S w
...
+
+  let move_if_affected ?(replace_with_const=false) (ask:Q.ask) x v f = unop_to_t' (fun x -> P.move_if_affected ~replace_with_const:replace_with_const ask x v f) (fun x -> T.move_if_affected ~replace_with_const:replace_with_const ask x v f)
+  (fun x -> U.move_if_affected ~replace_with_const:replace_with_const ask x v f) x
+  let get_vars_in_e = unop' P.get_vars_in_e T.get_vars_in_e U.get_vars_in_e
+  let smart_join f g = binop_to_t' (P.smart_join f g) (T.smart_join f g) (U.smart_join f g)
+  let smart_widen f g = binop_to_t' (P.smart_widen f g) (T.smart_widen f g) (U.smart_widen f g)
+  let smart_leq f g = binop' (P.smart_leq f g) (T.smart_leq f g) (U.smart_leq f g)
+  let update_length newl x = unop_to_t' (P.update_length newl) (T.update_length newl) (U.update_length newl) x
+
+  (* Functions that make us of the configuration flag *)
```
michael-schwarz, line 830:
```suggestion
  (* Functions that make use of the configuration flag *)
```

src/cdomains/arrayDomain.ml
mikcp: Fixed in 0522ddbdd4938551fddb99dc8d6107a323aa272e

src/cdomains/arrayDomain.ml
mikcp: Fixed on 3c946da793bb674eff15d0316c9686ebf05e9301

src/cdomains/arrayDomain.ml
mikcp: Fixed on 3ff05ea751b96b5b94d9f92a2ca6c26f129ae82b

src/cdomains/arrayDomain.ml
mikcp: Fixed on 750e88cd54796a384942a5e5ccd11c6bd44d5dc3

src/cdomains/arrayDomain.ml
```diff
@@ -69,6 +69,94 @@ struct
...
+      else weak_update xl 0 in
+    let f = Z.of_int (factor ()) in
+    let min_i = extract(Idx.minimal i) in
+    let max_i = extract(Idx.maximal i) in
+    if Z.geq min_i f then (xl, (Val.join xr v))
+    else if Z.lt max_i f then ((update_unrolled_values min_i max_i), xr)
+    else ((update_unrolled_values min_i (Z.of_int ((factor ())-1))), (Val.join xr v))
+  let make _ v =
+    let xl = BatList.make (factor ()) v in
+    (xl,v)
```
michael-schwarz, line 142: Actually, one should probably initialize the right part to `Val.bot ()` here. Otherwise, the `init_value` provides you with `top` and then you can never find out anything about the summarized part of the array, which makes even tracking it sort of pointless.

That would be a fix for now, we can then take a more principled approach to this in #485.

mikcp: Fixed on d8d16b694c8423f3d6caf0940aa0203a23f2cc68

PR #578, opened 2022-01-28
Optimize `MCP` by removing `assoc` usage

sim642:
### Changes
1. Change `presub` and `postsub` from eagerly constructed assoc lists to lazy functions. This avoids the expensive (~5% of total runtime) `filter_presubs`, which constructed those lists for each transfer function, but they aren't even used by anything other than ARINC and OSEK. This slowdown was my original motivation for getting rid of these (#460), but now that they aren't slowing down the usual analyses, they're not so much of a problem.
2. Reimplement `unop_fold` and `binop_fold` of `MCP`'s `Printable` and `Lattice` using `fold_left2` and `fold_left3` respectively, by folding the domain elements simultaneously with the list of activated analysis `Spec`s. This avoids any `assoc` in those operations. To avoid nasty segfaults, there are assertions to check that the component IDs still match up.
3. Reimplement `MCP`'s `combine` and `threadspawn` by folding over both domain elements (and optional contexts) simultaneously.
4. Replace `MCPRegistry` global assoc lists with `Hashtbl`s for faster lookup where still needed.

### TODO
- [x] Benchmark.
- [x] Figure out what Gobview needs the old `MCP.analyses_table` for, which was just a list of (analysis ID, analysis name) pairs.

### DIFF
--- a/src/analyses/mCP.ml
+++ b/src/analyses/mCP.ml
@@ -84,8 +84,7 @@ struct
     activated_ctx_sens := List.filter (fun (n, _) -> not (List.mem n !cont_inse)) !activated;
     match marshal with
     | Some marshal ->
-      combine !activated marshal
-      |> iter (fun ((_,{spec=(module S:MCPSpec); _}), marshal) -> S.init (Some (Obj.obj marshal)))
+      iter2 (fun (_,{spec=(module S:MCPSpec); _}) marshal -> S.init (Some (Obj.obj marshal))) !activated marshal
     | None ->
       iter (fun (_,{spec=(module S:MCPSpec); _}) -> S.init None) !activated
 
--- a/src/analyses/mCP.ml
+++ b/src/analyses/mCP.ml
@@ -161,7 +161,7 @@ struct
         f ((k,v::a')::a) b
     in f [] xs
 
-  let assoc_sub n xs name =
+  let assoc_sub xs name =
     let n' = List.assoc_inv name !analyses_table in
     assoc n' xs
 
@@ -237,8 +237,8 @@ struct
             ; edge   = ctx.edge
             ; ask    = (fun (type a) (q: a Queries.t) -> query ctx q)
             ; emit   = (fun e -> emits := e :: !emits)
-            ; presub = assoc_sub n ctx.local
-            ; postsub= assoc_sub n post_all
+            ; presub = assoc_sub ctx.local
+            ; postsub= assoc_sub post_all
             ; global = (fun v      -> ctx.global (v_of n v) |> g_to n |> obj)
             ; spawn  = (fun l v a  -> spawns := (v,(n,l,a)) :: !spawns)
             ; split  = (fun d es   -> splits := (n,(repr d,es)) :: !splits)
@@ -255,8 +255,8 @@ struct
             ; edge   = octx.edge
             ; ask    = (fun (type a) (q: a Queries.t) -> query octx q)
             ; emit   = (fun e -> emits := e :: !emits)
-            ; presub = assoc_sub n octx.local
-            ; postsub= assoc_sub n post_all
+            ; presub = assoc_sub octx.local
+            ; postsub= assoc_sub post_all
             ; global = (fun v      -> octx.global (v_of n v) |> g_to n |> obj)
             ; spawn  = (fun l v a  -> spawns := (v,(n,l,a)) :: !spawns)
             ; split  = (fun d es   -> splits := (n,(repr d,es)) :: !splits)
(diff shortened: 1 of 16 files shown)


michael-schwarz:
>  Figure out what Gobview needs the old `MCP.analyses_table` for, which was just a list of (analysis ID, analysis name) pairs.

Can you shed some light on that @keremc?

src/util/gobList.ml
```diff
@@ -0,0 +1,4 @@
+let rec fold_left3 f acc l1 l2 l3 = match l1, l2, l3 with
+  | [], [], [] -> acc
+  | x1 :: l1, x2 :: l2, x3 :: l3 -> fold_left3 f (f acc x1 x2 x3) l1 l2 l3
+  | _, _, _ -> invalid_arg "GobList.fold_left3"
```
michael-schwarz, line 4: `zip` from `Goblinutil` should probably also be moved here then

src/analyses/mCP.ml
```diff
@@ -741,7 +731,17 @@ struct
...
+      | [], _, [] -> acc
+      | (n, x) :: l1, Some ((n', y) :: l2), (n'', z) :: l3 when n = n' -> (* context-sensitive *)
+        assert (n = n'');
+        spec_list3_rev_acc ((n, spec n, (x, Some y, z)) :: acc) l1 (Some l2) l3
+      | (n, x) :: l1, l2_opt, (n'', z) :: l3 -> (* context-insensitive *)
+        assert (n = n'');
+        spec_list3_rev_acc ((n, spec n, (x, None, z)) :: acc) l1 l2_opt l3
+      | _, _, _ -> invalid_arg "MCP.spec_list3_rev_acc"
+    in
+    let f post_all (n,(module S:MCPSpec),(d,fc,fd)) =
```
michael-schwarz, line 744: This is (at least to me) very hard to understand. Could you elaborate on what exactly happens here?

sim642: It's like `spec_list2`, but for three lists, so the recursion here is basically what `map3_rev` would be. But here things are too nasty to have that because the context is optional (due to context insensitivity), so:
1. `MCP.combine` gets `fc: (int * Obj.t) list option`, so this handles the case when we don't have the second list at all. Then `None` is given to all inner analyses.
2. If that list is present, then there's still a second problem. If the analysis with ID `n` is in `ctx_insens`, then it has local domain elements from both the caller and the callee, but the corresponding element in the context list might have a different ID because it corresponds to already some following analysis. In that case also `None` is given to the inner analysis.
    Basically, the context list can be shorter than the local domain list and this matches them up by ID instead of doing something wrong.

sim642:
I'm still running sv-benchmarks, but here's ypbind_comb.c analysis times.

#### Before
```
TOTAL                           8.226 s
  parse                           0.055 s
  convert to CIL                  0.028 s
  analysis                        8.144 s
    global_inits                    0.002 s
    solving                         7.453 s
      S.Dom.equal                     0.027 s
      postsolver                      1.815 s
        access compare                  0.055 s
    warn_global                     0.669 s
      access                          0.668 s
        group_may_race                  0.613 s
          bfs                             0.613 s
            access compare                  0.330 s
            may_race                        0.109 s
        may_race                        0.001 s
        access compare                  0.004 s
```

#### After
```
TOTAL                           7.699 s
  parse                           0.054 s
  convert to CIL                  0.026 s
  analysis                        7.620 s
    global_inits                    0.002 s
    solving                         7.074 s
      S.Dom.equal                     0.026 s
      postsolver                      1.774 s
        access compare                  0.056 s
    warn_global                     0.521 s
      access                          0.520 s
        group_may_race                  0.459 s
          bfs                             0.458 s
            access compare                  0.234 s
            may_race                        0.078 s
        may_race                        0.001 s
        access compare                  0.004 s
```

sim642:
Here's the before vs after runtime on sv-benchmarks:
image
[Full results table](https://goblint.cs.ut.ee/results/61-all-fast-mcp-no-assoc-after/results/results.2022-01-28_15-22-21.table.html#/scatter?toolY=1&columnY=1&regression=Linear&scaling=Linear&line=2&results=All&filter=0(0*status*(status(notIn(ERROR%20(verify),EXCEPTION%20(Stack%20overflow),OUT%20OF%20MEMORY,TIMEOUT,TIMEOUT%20(true))))),1(0*status*(status(notIn(ERROR%20(verify),EXCEPTION%20(Stack%20overflow),OUT%20OF%20MEMORY,TIMEOUT,TIMEOUT%20(true))))))

**That's a whopping ~19% improvement on average!**

src/analyses/mCP.ml
michael-schwarz: Thanks for explaining!

sim642:
I think I managed to make the Gobview adaptation. Function entry/return states and global states at least display correctly, statement nodes for some reason don't, but it's already that way on master: https://github.com/goblint/gobview/issues/6.

keremc:
@michael-schwarz I don't know if this still holds, but I think the analysis ID used to be just an incremental ID and depended on the module initialization order. As a result, the ID was not stable across native and JS platforms, so I used `MCP.analyses_table` to reorder the analysis modules and reassign the IDs. At some point, the IDs ended up becoming stable across platforms somehow (maybe because of the `-linkall` flag), but I didn't remove the code for reordering.

michael-schwarz:
Does https://github.com/goblint/gobview/compare/3acd89c9ec9b0714433ae3b9aa82d596a5c34aac...d0d5b065be6a763f78987e89172abc04afdc7dd5#diff-22c5afd13651a2ae673bd24ed0234d0b09f51d2ede9fd99236875dbc8715fd73 sufficienlty address this issue then? At least at first glance, it seems like it. Or would we need more code for that?

sim642:
It's possible that #455 made that more consistent between native and JS builds. Anyway, my Gobview patch seemed to work as much as it could. And the reordering code is much shorter because there are fewer data structures to reorder.

PR #581, opened 2022-01-31
Fix no offset assert

stilscher:
When trying to run Goblint on [chrony](https://chrony.tuxfamily.org/) I discovered a case where a pointer points to something and has an integral type and has an offset. This causes the assert that checks that there is no offset in the optimization that avoids the evaluation of integer values when setting them to fail.

I added a minimal example as regression test in this PR. The proposed fix that I discussed with @michael-schwarz and @jerhard is to not apply the optimization if the integral type has an offset.


michael-schwarz (approved):

michael-schwarz:
This causes some false negatives where one could avoid reading the global first, but that should be in very rare cases. We discussed this in Munich and decided that having clean code is more important than avoiding reading from the global in such cases as well.

jerhard (approved):

PR #584, opened 2022-02-01
Preprocessor directives to disable Goblint includes

michael-schwarz:
Add preprocessor directives to exclude `GOBLINT_NO_xxxx` to exclude certain definitions from `pthread.c` or `stdlib.c` if (merged) programs contain these themselves.

Currently CIL just chooses on of the differing implementations producing warnings, which is not really nice. In the future I would like to re-enabled treating serious CIL warnings as errors, at which point this becomes needed to work with such programs.

Needed e.g. for https://github.com/goblint/bench/issues/16 and might also help with https://github.com/goblint/bench/issues/15


tests/regression/54-no-stdlib/01-noqsort.c
```diff
@@ -0,0 +1,10 @@
+// PARAM: --set cppflags[+] -DGOBLINT_NO_QSORT
```
sim642, line 1: There's no need for a new group just for this. Could just go into 41-stdlib.

tests/regression/41-stdlib/03-noqsort.c
```diff
@@ -0,0 +1,10 @@
+// PARAM: --set cppflags[+] -DGOBLINT_NO_QSORT
+#include<stddef.h>
+
+// There should be no CIL warning about multiple definitions here
+void qsort(void *ptr, size_t count, size_t size, int (*comp)(const void*, const void*), int more) {
```
sim642, line 5: Could we have `// NOWARN` here? Or does the testing script not detect the location as printed by CIL directly?
because currently this doesn't get automatically checked at all.

michael-schwarz: It's not a Goblint warning that gets generated here, but something internal to CIL that we do not check for in the regression tests

```
tests/regression/41-stdlib/03-noqsort.c:5: Warning: Incompatible declaration for qsort (from /home/michael/Documents/goblint-cil/analyzer/goblint_temp_003bde/03-noqsort.c(2)).
 Previous was at /home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c:6 (from /home/michael/Documents/goblint-cil/analyzer/goblint_temp_003bde/stdlib.c (0)) (different number of arguments) 
 
tests/regression/41-stdlib/03-noqsort.c:5: Warning: def'n of func qsort at tests/regression/41-stdlib/03-noqsort.c:5 (sum 204) conflicts with the one at /home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c:7 (sum 84380618838); keeping the one at /home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c:7.
```

sim642: Sure, but the script doesn't do anything smart, it just greps for locations and I've had debug prints confuse it before, so I thought it should be also to pick up the locations printed by these warnings. It doesn't matter that it's CIL printing them, they all appear in the same Goblint process's output.

sim642 (approved):

PR #585, opened 2022-02-02
Adaptations to preprocessing

stilscher:
This PR includes
- some fixes for the preprocessing with `cilly make` 
- some changes in the preprocessing using a compilation database that were required for [goblint/bench@#17](https://github.com/goblint/bench/issues/17). In the compilation database of this project, sometimes the `-o` option was not included in the arguments explicitly and so the preprocessing in Goblint failed. Since the generated compilation database generally looked fine to me and I couldn't find anything about the option `-o` being a required option in the arguments, I decided to extend the preprocessing in Goblint to handle this case.


src/util/compilationDatabase.ml
```diff
@@ -68,17 +68,20 @@ let load_and_preprocess ~all_cppflags filename =
...
-                  failwith "CompilationDatabase.preprocess: no -o argument value found for " ^ file
-              end
-            | exception Not_found ->
-              failwith "CompilationDatabase.preprocess: no -o argument found for " ^ file
+          let preprocess_arguments =
+            let suf =
+              begin match List.findi (fun i e -> e = "-o") arguments with
+                | (o_i, _) ->
+                  begin match List.drop o_i arguments with
+                    | (_ :: o_file :: arguments_tl) -> preprocessed_file :: arguments_tl
```
sim642, line 76: I think this should still be `split_at` because otherwise the arguments before `-o`, i.e. `arguments_init` are lost.

src/util/compilationDatabase.ml
```diff
@@ -68,17 +68,20 @@ let load_and_preprocess ~all_cppflags filename =
...
+                    | _ -> failwith ("CompilationDatabase.preprocess: no argument found for -o option for " ^ file)
+                  end
+                | exception Not_found -> [preprocessed_file]
+              end in
+            all_cppflags @ "-E" :: "-MMD" :: "-MT" :: file :: arguments @ "-o" :: suf in
```
sim642, line 81: Or umm, this `arguments` duplicates some of `-o` exists, then `arguments_tl` are included twice (once in `arguments` and once in `suf`). Moreover, `arguments` also contains the old `-o` argument in addition to the new one. Maybe this happens to work because one comes after the other, but this is fragile.

The old approach should be preferred that actually removes the old `-o` argument and doesn't duplicate or miss anything.

src/util/compilationDatabase.ml
```diff
@@ -68,17 +68,20 @@ let load_and_preprocess ~all_cppflags filename =
             preprocess_command
```
sim642, line 68: The same missing `-o` logic also needs to be in the `command` case above.

sim642 (changes requested):

src/util/compilationDatabase.ml
stilscher: Thanks for the note, I must have missed that when refactoring.

src/util/compilationDatabase.ml
stilscher: I now added the -o logic also for the command case and tested it with `compiledb --command-style make` for https://github.com/goblint/bench/issues/17.

src/util/compilationDatabase.ml
```diff
@@ -61,24 +61,30 @@ let load_and_preprocess ~all_cppflags filename =
...
-                  failwith "CompilationDatabase.preprocess: no -o argument value found for " ^ file
-              end
-            | exception Not_found ->
-              failwith "CompilationDatabase.preprocess: no -o argument found for " ^ file
+          let preprocess_arguments =
+            let suf =
+              begin match List.findi (fun i e -> e = "-o") arguments with
+                | (o_i, _) ->
+                  begin match List.split_at o_i arguments with
+                    | (arguments_init, o_file :: arguments_tl) -> arguments_init @ preprocessed_file :: arguments_tl
```
sim642, line 79: Previously the pattern contained `_ :: o_file`, where the wildcard was used to skip the `-o` argument. Is this somehow not necessary now?

stilscher: It still is. The string concatenation was mixed up, but I hope I fixed it now. Can you take another look at it?

sim642:
I fixed the merge conflict myself.

PR #586, opened 2022-02-03
Optimize `Queries.ID`

sim642:
`Queries.ID`, which is the domain used for the result of queries like `EvalInt` needs `top` and `bot`, but `IntDomTuple` itself does not have those. The previous hack was to abuse the top and bottom of `IInt` for that purpose, but this was far from ideal:
1. Implementations of `join` and `meet` were overridden to provide special behavior for those to avoid crashing due to mismatching ikinds. The implementations of the checks were like `let is_top x = equal (top ()) x`, which was straightforward, but awfully inefficient because every `meet` of the query result called that two times and each call of `IntDomTuple.top ()` looked up four options to know which int domains to activate. These accounted for most of the option lookups during execution, despite their caching.
2. Reimplementing `is_top` to delegate to `IntDomTuple.is_top_of` improved it a bit, but not much. This was due to the horribly inefficient `exists` and `for_all` implementations in `IntDomTuple`, which did the check through multiple levels of intermediate conversions (`BatEnum` and `List`). According to the added `BenchTuple4`, straight pattern matching is _hundreds of times faster_.
3. Reimplementing `is_bot` like that wasn't possible since there's no `is_bot_of`.
4. Abusing `IInt` ikind for a special purpose creates a conflict, where some actual `IInt` results could be treated differently from other ikinds.

Therefore this PR changes `Queries.ID` to lift `IntDomTuple` to fix all the above issues: there are explicit `Top` and `Bot` variants which are much cheaper to construct, check, join and meet, while being distinct from anything inside `IntDomTuple`.

### TODO
- [x] Add benchmark results.
- [x] Try to explain high variance in long-running benchmarks.

### DIFF
--- a/bench/basic/benchTuple4.ml
+++ b/bench/basic/benchTuple4.ml
@@ -70,5 +70,72 @@ let () =
       ]
   )
 
+
+let () =
+  let for_all0 =
+    let open Batteries in
+    let to_list x = Tuple4.enum x |> List.of_enum |> List.filter_map identity in
+    let f g = g identity % to_list in
+    List.(f for_all)
+  in
+
+  let for_all1 = function
+    | (Some false, _, _, _)
+    | (_, Some false, _, _)
+    | (_, _, Some false, _)
+    | (_, _, _, Some false) ->
+      false
+    | _ ->
+      true
+  in
+
+  let for_all2 (a, b, c, d) = not (a = Some false || b = Some false || c = Some false || d = Some false) in
+
+
+  register (
+    "for_all" @>>> [
+        "all None" @> lazy (
+            let args = (None, None, None, None) in
+            throughputN 1 [
+              ("0", for_all0, args);
+              ("1", for_all1, args);
+              ("2", for_all2, args);
+            ]
+          );
+        "all Some true" @> lazy (
+            let args = (Some true, Some true, Some true, Some true) in
+            throughputN 1 [
+              ("0", for_all0, args);
+              ("1", for_all1, args);
+              ("2", for_all2, args);
+            ]
+          );
+        "all Some false" @> lazy (
+            let args = (Some false, Some false, Some false, Some false) in
+            throughputN 1 [
+              ("0", for_all0, args);
+              ("1", for_all1, args);
+              ("2", for_all2, args);
+            ]
+          );
+        "all None except last Some false" @> lazy (
+            let args = (None, None, None, Some false) in
+            throughputN 1 [
+              ("0", for_all0, args);
+              ("1", for_all1, args);
+              ("2", for_all2, args);
+            ]
+          );
+        "all Some true except last Some false" @> lazy (
+            let args = (Some true, Some true, Some true, Some false) in
+            throughputN 1 [
+              ("0", for_all0, args);
+              ("1", for_all1, args);
+              ("2", for_all2, args);
+            ]
+          );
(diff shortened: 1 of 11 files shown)


sim642:
## sv-benchmarks
[Full results table](https://goblint.cs.ut.ee/results/65-all-fast-mcp-no-assoc-after-short-circuit-queries-id-3/results/results.2022-02-02_12-53-31.table.html#/scatter?filter=0(0*status*(status(in(unknown,true,EXCEPTION%20(Failure),EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError))))),1(0*status*(status(in(unknown,true,EXCEPTION%20(Failure),EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError))))),2(0*status*(status(in(unknown,true,EXCEPTION%20(Failure),EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError))))),3(0*status*(status(in(unknown,true,EXCEPTION%20(Failure),ASSERTION,EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError)))))&toolY=3&columnY=1&scaling=Linear&results=All&regression=Linear&toolX=0&columnX=1&line=1.5)

### Linear plot
At first sight, the before vs after analysis time results on sv-benchmarks look disappointing. There's no clear trend of better or worse performance, but rather in the bigger benchmarks there's a lot of variance.
image

### Log plot
Looking at the log plot, though, reveals something interesting. Most of the points are below the blue identity line, which is indicative of better performance. Although no linear regression is possible in log scale to measure how much.
image

### Linear plot <10s
Based on the finding from the log plot, if we limit the plot to just <10s runtime, then the linear plot also shows a clear ~10% optimization.
image

### Hypothetical explanation
I've found these results very surprising, because based on profiling, it seemed like an obvious improvement to avoid an excessive amount of int domain option lookups on each query, but throughout the multiple steps that I went through (listed in the original description), there wasn't anything clear in the results.

So I've spent the last two days benchmarking and profiling these to understand what exactly is causing the sudden slowdown in some bigger benchmarks. In one case I profiled, I noticed an increased amount of `HoarePO` function calls (`equal`, `leq`, `for_all`, `mem`, `exists`, `elements` — all of those are terribly implemented to convert the whole map of buckets into a list each time, but that's an optimization for another time). A change to the bottom and top of `Queries.ID` should make no difference to how the address domain behaves. Even though address index offsets contain int domain elements, they are `IntDomainTuple`, not `Queries.ID`.

I haven't managed to nail this down for certain, but the only remaining theory I have is that previously abusing the `IInt` ikind for the top and bottom of `EvalInt` queries themselves caused some unnecessary mixups with actual results of that type. I suppose those mixups might have caused us to be unnecessarily imprecise in some crazy cases occurring in the big benchmarks. With the cleanup of that hack, it might simply be that we're a tiny amount more precise, but that itself could be creating noticeably more contexts to analyze or something.

michael-schwarz (approved):

sim642:
I'm not too happy about not knowing for certain where the slowdown comes from, but at this point I'm out of ideas how to find that out. So let's just hope that the hack bottom and top were somehow behind this.

PR #587, opened 2022-02-03
Compute dependencies of file via CIL instead of from `.d` files generated during pre-processing

michael-schwarz:
This is an alternative to https://github.com/goblint/analyzer/pull/535 which it reverts.

Instead of generating `.d` files during pre-processing and then parsing those in, we use the fact that the desired information is newly exposed by CIL as of https://github.com/goblint/cil/pull/73 

The hope is that this simplifies the generation of these dependencies and also makes it less fragile (as observed for https://github.com/goblint/analyzer/pull/557 and https://github.com/goblint/bench/issues/17).

CC: @karoliineh since this affects GobPie.


src/framework/analyses.ml
```diff
@@ -274,13 +274,13 @@ struct
         iterGlobals file (function
             | GFun (_, loc)
             | GVar (_, _, loc) ->
-              SH.replace files loc.file (Hashtbl.find_option Preprocessor.dependencies loc.file)
+              SH.replace files loc.file (Hashtbl.find_default Preprocessor.dependencies loc.file [])
```
sim642, line 277: Why this change if the data structure is the same?

src/maingoblint.ml
```diff
@@ -291,7 +288,12 @@ let preprocess_files () =
 let merge_preprocessed cpp_file_names =
   (* get the AST *)
   if get_bool "dbg.verbose" then print_endline "Parsing files.";
-  let files_AST = List.map Cilfacade.getAST cpp_file_names in
+  let get_ast_and_record_deps (orig, f) =
+    let file = Cilfacade.getAST f in
+    Hashtbl.add Preprocessor.dependencies orig file.files;
```
sim642, line 293: What's with this `orig`? CIL already provides `file.fileName`, which was fine before.

src/maingoblint.ml
```diff
@@ -256,7 +253,7 @@ let preprocess_files () =
   let rec preprocess_arg_file = function
     | filename when Filename.basename filename = "Makefile" ->
       let comb_file = MakefileUtil.generate_and_combine filename ~all_cppflags in
-      [basic_preprocess ~all_cppflags comb_file]
+      [(comb_file, basic_preprocess ~all_cppflags comb_file)]
```
sim642, line 256: I don't think this really works for the old Makefile support because having cilly combine the C files doesn't allow us to know what the actual files were. The `comb_file` isn't actually modified by the user, so we'd be oblivious to the changes and would have to rerun cilly anyway.

src/util/preprocessor.ml
```diff
@@ -34,15 +34,6 @@ let cpp =
...
 let dependencies: (string, string list) Hashtbl.t = Hashtbl.create 3
 
-let parse_makefile_deps makefile =
-  BatFile.with_file_in makefile (fun input ->
-      let lexbuf = Lexing.from_channel input in
-      let (rule, deps) = MakefileParser.deps MakefileLexer.token lexbuf in
-      let deps = List.remove deps rule in
-      Hashtbl.replace dependencies rule deps
-    )
+let get_cpp () = Lazy.force cpp
```
sim642, line 39: Unnecessary reordering.

src/framework/analyses.ml
michael-schwarz: I personally think having `null` instead of the empty list in the JSON is ugly, but I can switch it back if we want to do that.

src/maingoblint.ml
michael-schwarz: Yes, it is still somewhat broken, but arguably less so then before, where the dependency information as simply empty.

src/framework/analyses.ml
sim642: Previously the possibility of `None`/`null` was necessary for the cases where we failed to extract the dependencies via `-MD`. I guess now they should always be there, so it should be safe to just use `Hashtbl.find` without the default here. That way if there's ever any issue, it's immediately obvious via the exception rather than being silently replaced by pretending to have no dependencies, which could even be unsound.

src/maingoblint.ml
michael-schwarz: For example:

```
orig is /home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c and fileName is /home/michael/Documents/goblint-cil/analyzer/goblint_temp_78a89f/stdlib.c
orig is /home/michael/Documents/goblint-cil/analyzer/includes/pthread.c and fileName is /home/michael/Documents/goblint-cil/analyzer/goblint_temp_78a89f/pthread.c
orig is 1.c and fileName is /home/michael/Documents/goblint-cil/analyzer/goblint_temp_78a89f/1.c
```

The thing that is recorded in the merged file is also the original file name, so not using `orig` leads to an empty list of dependencies:


~~~JSON
  "files": {
    "1.c": [],
    "/usr/include/x86_64-linux-gnu/bits/byteswap.h": [],
    "/home/michael/Documents/goblint-cil/analyzer/includes/pthread.c": [],
    "/usr/include/x86_64-linux-gnu/bits/uintn-identity.h": [],
    "/home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c": []
  }
~~~

vs

~~~JSON
"files": {
    "1.c": [
      "/usr/include/x86_64-linux-gnu/bits/types/struct_timeval.h",
      "/usr/include/x86_64-linux-gnu/bits/types/__sigset_t.h", "1.c",
      "/usr/include/x86_64-linux-gnu/bits/types/timer_t.h",
      "/usr/include/x86_64-linux-gnu/gnu/stubs.h",
      "/usr/include/x86_64-linux-gnu/sys/sysmacros.h",
      "/usr/include/x86_64-linux-gnu/bits/floatn-common.h",
      "/usr/include/x86_64-linux-gnu/bits/byteswap.h",
      "/usr/include/x86_64-linux-gnu/bits/typesizes.h",
      "/usr/include/endian.h", "/usr/include/x86_64-linux-gnu/sys/select.h",
      "/usr/include/x86_64-linux-gnu/bits/waitflags.h",
      "/usr/include/x86_64-linux-gnu/bits/wordsize.h",
      "/usr/include/x86_64-linux-gnu/sys/cdefs.h",
      "/usr/include/stdc-predef.h",
      "/usr/include/x86_64-linux-gnu/bits/pthreadtypes.h", "<command-line>",
      "/usr/include/x86_64-linux-gnu/bits/long-double.h",
      "/usr/include/x86_64-linux-gnu/bits/types/clockid_t.h",
      "/usr/include/x86_64-linux-gnu/bits/sysmacros.h",
      "/usr/lib/gcc/x86_64-linux-gnu/7/include/stddef.h",
      "/usr/include/x86_64-linux-gnu/bits/byteswap-16.h",
      "/usr/include/x86_64-linux-gnu/bits/uintn-identity.h",
      "/usr/include/x86_64-linux-gnu/bits/pthreadtypes-arch.h",
      "/usr/include/x86_64-linux-gnu/gnu/stubs-64.h",
      "/usr/include/x86_64-linux-gnu/bits/types/clock_t.h",
      "/usr/include/x86_64-linux-gnu/bits/waitstatus.h",
      "/usr/include/x86_64-linux-gnu/bits/endian.h",
      "/home/michael/Documents/goblint-cil/analyzer/includes/assert.h",
      "<built-in>", "/usr/include/x86_64-linux-gnu/bits/select.h",
      "/usr/include/x86_64-linux-gnu/bits/thread-shared-types.h",
      "/usr/include/x86_64-linux-gnu/sys/types.h", "/usr/include/stdlib.h",
      "/usr/include/x86_64-linux-gnu/bits/types.h",
      "/usr/include/x86_64-linux-gnu/bits/floatn.h",
      "/usr/include/x86_64-linux-gnu/bits/libc-header-start.h",
      "/usr/include/x86_64-linux-gnu/bits/types/time_t.h",
      "/usr/include/x86_64-linux-gnu/bits/types/struct_timespec.h",
      "/usr/include/x86_64-linux-gnu/bits/stdint-intn.h",
      "/usr/include/alloca.h", "/usr/include/features.h",
      "/usr/include/x86_64-linux-gnu/bits/types/sigset_t.h",
      "/usr/include/x86_64-linux-gnu/bits/stdlib-float.h"
    ],
    "/usr/include/x86_64-linux-gnu/bits/byteswap.h": [],
    "/home/michael/Documents/goblint-cil/analyzer/includes/pthread.c": [
      "/usr/include/x86_64-linux-gnu/bits/types/timer_t.h",
      "/usr/include/x86_64-linux-gnu/gnu/stubs.h", "/usr/include/sched.h",
      "/usr/include/x86_64-linux-gnu/bits/byteswap.h",
      "/usr/include/endian.h",
      "/usr/include/x86_64-linux-gnu/bits/typesizes.h",
      "/usr/include/x86_64-linux-gnu/bits/types/locale_t.h",
      "/usr/include/pthread.h",
      "/usr/include/x86_64-linux-gnu/bits/setjmp.h",
      "/usr/include/x86_64-linux-gnu/bits/wordsize.h",
      "/usr/include/x86_64-linux-gnu/sys/cdefs.h",
      "/usr/include/stdc-predef.h",
      "/usr/include/x86_64-linux-gnu/bits/cpu-set.h",
      "/usr/include/x86_64-linux-gnu/bits/pthreadtypes.h", "<command-line>",
      "/usr/include/x86_64-linux-gnu/bits/long-double.h",
      "/usr/include/x86_64-linux-gnu/bits/types/clockid_t.h",
      "/usr/include/x86_64-linux-gnu/bits/byteswap-16.h",
      "/usr/include/x86_64-linux-gnu/bits/uintn-identity.h",
      "/usr/lib/gcc/x86_64-linux-gnu/7/include/stddef.h",
      "/usr/include/x86_64-linux-gnu/bits/pthreadtypes-arch.h",
      "/usr/include/x86_64-linux-gnu/gnu/stubs-64.h",
      "/usr/include/x86_64-linux-gnu/bits/types/clock_t.h",
      "/usr/include/x86_64-linux-gnu/bits/endian.h",
      "/usr/include/x86_64-linux-gnu/bits/types/__locale_t.h", "<built-in>",
      "/usr/include/time.h",
      "/usr/include/x86_64-linux-gnu/bits/thread-shared-types.h",
      "/usr/include/x86_64-linux-gnu/bits/sched.h",
      "/home/michael/Documents/goblint-cil/analyzer/includes/pthread.c",
      "/usr/include/x86_64-linux-gnu/bits/types/struct_itimerspec.h",
      "/usr/include/x86_64-linux-gnu/bits/types.h",
      "/usr/include/x86_64-linux-gnu/bits/time.h",
      "/usr/include/x86_64-linux-gnu/bits/types/time_t.h",
      "/usr/include/x86_64-linux-gnu/bits/types/struct_timespec.h",
      "/usr/include/x86_64-linux-gnu/bits/types/struct_tm.h",
      "/usr/include/features.h"
    ],
    "/usr/include/x86_64-linux-gnu/bits/uintn-identity.h": [],
    "/home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c": [
      "<command-line>", "/usr/lib/gcc/x86_64-linux-gnu/7/include/stddef.h",
      "<built-in>",
      "/home/michael/Documents/goblint-cil/analyzer/includes/stdlib.c",
      "/usr/include/stdc-predef.h"
    ]
  },
~~~

src/maingoblint.ml
sim642: Might be worth adding a comment that this doesn't give the original dependency structure, unlike all the other mechanisms.

karoliineh:
In #535 implementation the json output files object was something like:

```
files": {
    "/home/ubuntu/DemoProject/src/subproblems/10-nullpointer-dereference-simple.c": [
      "/home/ubuntu/DemoProject/src/subproblems/10-nullpointer-dereference-simple.h"
    ],
    "/home/ubuntu/DemoProject/src/subproblems/01-assert.c": [
      "/home/ubuntu/goblint/analyzer/_opam/share/goblint/includes/assert.h",
      "/home/ubuntu/DemoProject/src/subproblems/01-assert.h"
    ],
    "/home/ubuntu/goblint/analyzer/_opam/share/goblint/includes/stdlib.c": [],
...
```

now it has a ton of different stuff in it:

```
files": {
    "/home/ubuntu/DemoProject/src/subproblems/10-nullpointer-dereference-simple.c": [
      "/usr/include/x86_64-linux-gnu/gnu/stubs.h",
      "/usr/include/x86_64-linux-gnu/bits/types/__FILE.h",
      "/usr/include/x86_64-linux-gnu/bits/typesizes.h",
      "/usr/include/x86_64-linux-gnu/bits/types/__fpos64_t.h",
      "/usr/include/x86_64-linux-gnu/bits/types/FILE.h",
      "/home/ubuntu/DemoProject/src/subproblems/10-nullpointer-dereference-simple.c",
      "/usr/include/x86_64-linux-gnu/bits/wordsize.h",
      "/usr/include/x86_64-linux-gnu/sys/cdefs.h",
      "/usr/include/x86_64-linux-gnu/bits/types/struct_FILE.h",
      "/usr/include/stdc-predef.h", "<command-line>",
      "/usr/include/x86_64-linux-gnu/bits/long-double.h",
 ...
```

for GopPie, we only need to get back the list of the project files that have been sent for Goblint to be analysed (both .c and .h), so that we could track if any new files have been passed to be analysed since the last analysis. Therefore from the first implementation the `../goblint/analyzer/_opam/share/goblint/includes/stdlib.c` files can be removed as well.

src/framework/analyses.ml
michael-schwarz: No, this dependencies map can still be empty for header files that are not seen by the cpp on its own. 
In my example this is the case for  `/usr/include/x86_64-linux-gnu/bits/byteswap.h`.

It may have dependencies, and they are transitively included in any file passed to the CPP, but not recorded for the file itself.

So maybe you're right and we should go back to having `None/null` here.

sim642:
Overall, having now seen the actual outputs, I see a bunch of problems:
- [ ] Includes `<built-in>`.
- [ ] Includes `<command-line>`.
- [ ] Each file includes itself as dependency!
- [ ] System headers cannot be distinguished.

So actually, I think the reasonable thing to do would be the following:
1. Keys of the map are preprocessed files (so no `orig`). This would be consistent with old Makefile support, because there the preprocessed (i.e. preprocessed _and_ combined) file would be the only key.
2. That automatically solves the self-dependency problem, because then the preprocessed file has the original as its dependency, which is completely true. And again consistent with the old Makefile support, where all the combined original files would be listed as dependencies anyway.
3. CIL should exclude `<built-in>` and `<command-line>`.
4. CIL should return `(string * bool) list` or `(string, bool) Hashtbl.t`, where the value boolean indicates whether the included file is a system one or not. It can easily know this (just like `cpp -MD` does) by just looking at the output line markers. The number indicates which one it is: https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html#Preprocessor-Output. It would be best to rely on `cpp` doing it for us, rather than using some heuristics in Goblint for this purpose (e.g. guessing based on directories).

src/framework/analyses.ml
sim642: That `/usr/include/x86_64-linux-gnu/bits/byteswap.h` shouldn't appear in the dependency keys in the first place. If it's included by something, it's already listed as a dependency of a C file, so a header on its own makes zero sense. I'm not exactly sure where it now suddenly appears from, but my suggestion here https://github.com/goblint/analyzer/pull/587#issuecomment-1029102095 would actually work around this problem altogether.

michael-schwarz: It is not in the keys, but is looked up because some `GVar`s have it as their location.

michael-schwarz:
I'll give that a shot tomorrow then.

src/framework/analyses.ml
sim642: As mentioned by the TODO there, it's probably incomplete as an approach anyway. The complete approach would be to just dump the entire dependencies structure, that must necessarily contain all filenames that could possibly occur everywhere. And it would be much simpler than having a CIL visitor go through the whole AST just to find all the filenames, which the dependencies structure already has.

src/maingoblint.ml
```diff
@@ -256,6 +253,7 @@ let preprocess_files () =
   let rec preprocess_arg_file = function
     | filename when Filename.basename filename = "Makefile" ->
       let comb_file = MakefileUtil.generate_and_combine filename ~all_cppflags in
+      (* The resulting dependency information will not be correctly tracked, because the merge happens in cilly *)
```
sim642, line 256: I suppose this comment isn't necessary anymore since this would just make all the combined files dependencies of `comb_file`, which is right.

src/framework/analyses.ml
```diff
@@ -268,19 +268,9 @@ struct
...
-            | GVar (_, _, loc) ->
-              SH.replace files loc.file (Hashtbl.find_option Preprocessor.dependencies loc.file)
-            | _ -> () (* TODO: add locs from everything else? would also include system headers *)
-          );
-        files |> SH.to_list
-      in
+      let files = Hashtbl.to_list Preprocessor.dependencies in
       let json = `Assoc [
-          ("files", `Assoc (List.map (Tuple2.map2 [%to_yojson: string list option]) files));
+          ("files", `Assoc (List.map (Tuple2.map2 [%to_yojson: (string*bool) list]) files));
```
sim642, line 273: I didn't look at the output, but if I remember correctly, these pairs show up as JSON arrays like `["file.c", false]`, which aren't very good style. I'm thinking maybe we should just filter out the system ones and leave the non-system filenames (without the booleans).
Of course it's true that the preprocessing depended on those system headers, but I don't think there's value in watching them for changes, etc, because we already assume in many other ways that GCC doesn't randomly change (e.g. CIL's `Machdep` would have to be forcefully rebuilt then anyway).

sim642 (approved):

PR #590, opened 2022-02-04
Refactor MCP transfer functions and add query cache

sim642:
### Changes
1. Removes all `ctx` duplication from `MCP` transfer functions. This significantly shortens its code and makes future changes much easier as opposed to having to change a dozen `ctx`s.
2. The extracted `ctx` creation allows to easily add a cache for queries for the duration of a single transfer function. Hopefully this gives a performance benefit.
3. Replace `ctx.assign` with `Assign` event.

### TODO
- [x] Benchmark.

### DIFF
--- a/src/analyses/arinc.ml
+++ b/src/analyses/arinc.ml
@@ -337,7 +337,7 @@ struct
       let assign_id exp id =
         match exp with
         (* call assign for all analyses (we only need base)! *)
-        | AddrOf lval -> ctx.assign ~name:"base" lval (mkAddrOf @@ var id)
+        | AddrOf lval -> ctx.emit (Assign {lval; exp = mkAddrOf @@ var id})
         (* TODO not needed for the given code, but we could use Queries.MayPointTo exp in this case *)
         | _ -> failwith @@ "Could not assign id. Expected &id. Found "^sprint d_exp exp
       in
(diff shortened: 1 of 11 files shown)


sim642:
sv-benchmarks before vs after shows ~6% improvement:
image
[Full results table](https://goblint.cs.ut.ee/results/67-all-fast-mcp-ctx-before/results/results.2022-02-05_10-34-34.table.html#/scatter?toolY=1&columnY=1&results=All&scaling=Linear&filter=0(0*status*(status(in(unknown,true,EXCEPTION%20(Failure),EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError))))),1(0*status*(status(in(unknown,true,EXCEPTION%20(Failure),EXCEPTION%20(IntDomain.IncompatibleIKinds),EXCEPTION%20(Cilfacade.TypeOfError)))))&regression=Linear&line=1.5)

I was hoping it would make a bigger difference, because the amount of repeated `EvalInt` queries alone is insane if you trace `evalint`.

sim642:
As a little bit of cleanup, I now also removed `ctx.assign` and replaced it with an event. It's only used by arinc/extract_arinc/extract_osek, none of which actually test `ctx.assign`. Manually looking at the analysis results in HTML the ARINC task IDs do get assigned to locals in base, so it should work as before.

michael-schwarz (approved):

PR #591, opened 2022-02-08
Reparse files before each analysis run in server mode

keremc:
This PR adds a new option `server.reparse`. When it is on, each `analyze` request causes the input files to be reparsed. Also fixes [this issue](https://github.com/goblint/analyzer/pull/522#discussion_r794594560) that @jerhard pointed out.

Example invocation:
`$ ./goblint --enable server.enabled --enable server.reparse .`

This might be interesting for @karoliineh.


src/util/server.ml
```diff
@@ -98,26 +110,57 @@ let bind () =
...
+  | Some solver_data when reparsed ->
+    let _, changes = VersionLookup.updateMap s.file file s.version_map in
+    let old_data = Some { Analyses.cil_file = s.file; solver_data } in
+    s.max_ids <- UpdateCil.update_ids s.file s.max_ids file s.version_map changes;
+    { Analyses.changes; old_data; new_file = file }, false
+  | Some solver_data ->
+    let changes = virtual_changes file in
+    let old_data = Some { Analyses.cil_file = file; solver_data } in
+    { Analyses.changes; old_data; new_file = file }, false
+  | _ -> Analyses.empty_increment_data file, true
```
sim642, line 148: Isn't this duplicating some logic from elsewhere in simplified form? Because these things look very specific to incremental analysis. It would be more error-prone if this logic isn't duplicated in different places because they could easily go out of sync: it's easy to forget to fix things here in server mode if modifying the general incremental logic.

src/util/server.ml
```diff
@@ -4,7 +4,10 @@ open Jsonrpc
 exception Failure of Response.Error.Code.t * string
 
 type t = {
-  file: Cil.file;
+  mutable file: Cil.file;
+  mutable version_map: (CompareCIL.global_identifier, Cil.global) Hashtbl.t;
+  mutable max_ids: VersionLookup.max_ids;
+  preprocess_and_merge: unit -> Cil.file;
   do_analyze: Analyses.increment_data -> Cil.file -> unit;
```
sim642, line 11: I might've forgotten to bring this up in #522, but what's the reason for collecting these things into a record for server mode?

keremc: The functions are there because I cannot access anything defined in `maingoblint.ml` without causing a cyclic dependency. The CIL stuff is there because `do_analyze` needs it.

I can also create a new module and move `do_analyze` & `preprocess_and_merge` to this new module to break the cyclic dependency if that is preferable.

sim642: Moving these out of `Maingoblint` would be preferable indeed.

karoliineh (commented):
If I start the server with command:

`goblint --conf goblint.json --enable server.enabled --enable server.reparse --set result json-messages -o analysisResults.json main.c`

with program

```
#include <stdio.h>
#include <assert.h>

int main() {

  int n = 5;
  assert(n == 1);

  return 0;
}
```

then the first time I send a request `{"jsonrpc":"2.0","id":0,"method":"analyze","params":{}}` i get the warnings:

`[Error][Assert] Assertion "n == 1" will fail. (./src/main.c:26:3-26:17)`

which is correct, but the second time I send the same request without changing anything in the `main.c` file I get 

```
[Warning][Unknown] Calculated state for undefined function: unexpected node Statement return (0); (initfun:0:0)
[Warning][Unknown] Calculated state for undefined function: unexpected node Statement assert(n == 1); (initfun:0:0)
[Warning][Unknown] Calculated state for undefined function: unexpected node Statement n = 5; (initfun:0:0)
```

src/maingoblint.ml
```diff
@@ -318,6 +303,8 @@ let merge_preprocessed cpp_file_names =
   Cilfacade.current_file := merged_AST;
   merged_AST
 
+let preprocess_and_merge () = preprocess_files () |> merge_preprocessed
```
sim642, line 306: Another issue that Karoliine stumbled upon is related to `merge_preprocessed` opening and setting `Goblintutil.out`, but that is never closed. This causes the `json-messages` output, which prints to that `out_channel` to append to the initially opened file, instead of overwriting its contents, yielding invalid JSON.

In the server mode, `Goblintutil.out` should at some point be closed as well (in non-server mode we just leave it up to the OS as the program exits). Then the subsequent opening would actually reopen the file and overwrite, instead of further appending to the already-opened file.

sim642:
I fixed the "Calculated state for undefined function" issues at least, hopefully the initial "Function definition missing" ones as well, @karoliineh.

In `Cilfacade` we have a few global lookup tables that are constructed lazily from `current_file`. Having parsed and loaded a new file, those lookup tables are now invalid and need to be reset. Since it's not possible to reset OCaml's builtin lazies, I implemented a tiny `ResettableLazy` module, which allows that.

src/maingoblint.ml
sim642: Actually no, I got it backwards: reopening the same file each time in `merge_preprocessed` is not problematic, because the reopened descriptor truncates the file still. Although never closing it a small file descriptor leak.

The problem actually happens with `--disable server.reparse` because then `merge_preprocessed` is not called for reanalysis, so the same `Goblintutil.out` channel is opened and appended to. Opening of those channels should probably be pulled out of `merge_preprocessed` somewhere that gets re-run regardless of reparsing.

michael-schwarz:
@keremc What is the state of these improvements, would be great to merge this soon!

src/maingoblint.ml
keremc: Just wondering, but is an incremental run guaranteed/supposed to produce the same set of warnings as a full run? I moved the line that sets `Goblintutil.out` to `do_analyze` and now clear the message table before each run because otherwise it ends up containing stale warnings.

src/util/server.ml
keremc: Instead of moving these out of `Maingoblint`, I moved `main` into `Goblint`. Hope that's fine.

keremc:
@michael-schwarz There are still some soundness issues remaining. A trivial one I found is this:
```c
int main(int argc, char **argv) { return 0; }
```
If I start the server with `./goblint --enable server.enabled --enable server.reparse main.c` and request `{"jsonrpc":"2.0","id":0,"method":"analyze","params":{}}` twice, the second run prints this warning: `[Info][Unsound] Unknown value type given as function argument (initfun:0:0)`

src/maingoblint.ml
sim642: Yes, the incremental analysis makes sure to output all the warnings again, so they go into the table. On `master`, this happens trivially through the postsolving/verify phase in the solver; on `interactive`, there's some special incremental warning logic that uses separate solver data structures to reuse some warnings. But either way they should all go back into the table, so clearing it is the way to go. Especially since in non-server incremental runs the messages table would be empty as well.

src/util/server.ml
sim642: I think that's kind of backwards now, `Maingoblint` should be the one to contain the main function for the goblint binary and the rest could go into `Goblint`. It should be equally easy to have them swapped because nothing else could depend on `Maingoblint` before anyway.

sim642:
> @michael-schwarz There are still some soundness issues remaining. A trivial one I found is this:
> 
> ```c
> int main(int argc, char **argv) { return 0; }
> ```
> 
> If I start the server with `./goblint --enable server.enabled --enable server.reparse main.c` and request `{"jsonrpc":"2.0","id":0,"method":"analyze","params":{}}` twice, the second run prints this warning: `[Info][Unsound] Unknown value type given as function argument (initfun:0:0)`

I pushed a fix to this, it was just that `should_warn` remained true from the previous run, so a spurious warning from the initialization code was printed.

src/util/server.ml
keremc: OK, I swapped the two modules. Unfortunately, dune doesn't seem to like it when the executable name doesn't match the module name. So, I think I have to rename `src/goblint.exe` to `src/maingoblint.exe`.

sim642: Ugh, ok. I thought `Maingoblint` was the module for `goblint.exe` since we got rid of the stupid `Goblint` module generation in #455, but I forgot that it still left the tiny module to do the `at_exit` hack.

I'd rather not have the binary itself renamed, so I guess we should just revert the module swap and live with the confusing naming for the time being (until `Maingoblint` gets split up even more to more descriptive modules). Sorry for the confusion!

michael-schwarz (commented):
I think when the module swap has been undone, we should be good to merge this.

sim642 (approved):

PR #592, opened 2022-02-08
Affine Equalities

MartinWehking:
This draft PR introduces an analysis for affine equalities (ref: [Michael Karr , 1976](https://scholar.google.com/scholar_url?url=https://idp.springer.com/authorize/casa%3Fredirect_uri%3Dhttps://link.springer.com/content/pdf/10.1007/BF00268497.pdf%26casa_token%3DjHkKNvRPTY4AAAAA:yxzkcPFZP1RWxLZY-TWAg7cO8EfC0Txc78-CTZiDMFnyvkXp54p3s6RHEnxG0Auh7NRZELlImjxkZTat&hl=en&sa=T&oi=gsb-gga&ct=res&cd=0&d=14162577617566575899&ei=444CYu9y-pXL1g-ynpHACg&scisig=AAGBfm24cgr8mbazZ4npHDMjTnSTHqGDUA)) that can be run independently from the already existing apron analyses.
Abstract states in the newly added domain are represented by structs containing a matrix and an apron environment.
Matrices are modeled as proposed by Karr: Each variable is assigned to a column and each row represents a linear affine relationship that must hold at the corresponding program point. 
The apron environment is hereby used to organize the order of columns and variables.
Furthermore, the apron analysis and its respective privatization modules have been modified to fit the needs for apron-independent relational analyses.
A short list giving a quick (and perhaps not full) overview of what's been added/modified:
- **affineEqualityDomain** - Contains the majority of code for the new domain.
-  **sharedFunctions** - Contains functions and modules that are shared among the original _apronDomain_ and the new  _affineEqualityDomain_
- **relationDomain** - Provides interfaces/implementations that generalize the _apronDomain_ and _affineEqualityDomain_.
- **relationAnalysis** - Contains most of the implementation of the original _apronDomain_, but now solely operates with functions provided by _relationDomain_.
- **relationPriv**- Has been modified to work with any domain that uses the functions provided _relationDomain_
- An extended overflow handling inside _relationAnalysis_ for expression assignments when overflows are assumed to occur. Since affine equalities can only keep track of integer bounds of expressions evaluating to definite constants, we now query the integer bounds information for expressions from other analysis.
If an analysis returns bounds that are unequal to min and max of _ikind_ , we can exclude the possibility that an overflow occurs and the abstract effect of the expression assignment can be used, i.e. we do not have to set the variable's value to top.
Additionally, we now also refine after positive guards  when overflows might occur and there is only one variable inside the expression and the expression is an equality constraint check (==). We check after the refinement if the new value of the variable is outside its integer bounds and if that is the case, either revert to the old state or set it to bottom.

The new analysis can be activated by the following parameter:
`--set ana.activated[+] affeq`
In order to compare affeq with the apron affine equalities implementation, the affeq analysis inside apron can be selected by the following parameter (when apron is activated):
`--set ana.apron.domain "affeq"`
- One can switch between the array and list matrices for the affine equality domain by the parameter:
   `--set ana.affeq.matrix list/array` 

**A more detailed descriptions of all files/modules and interfaces:**

-**affineEqualityDomain** (contains mostly new code)**:
*Modules*
-**VarManagement**: functor parameterized by `AbstractMatrix` and `AbstractVector`. It defines the type t of the affine equality domain (a struct that contains an optional matrix and an apron environment) and provides the functions needed for handling variables (which are defined by `RelationDomain.D2`) such as `add_vars` `remove_vars`. Furthermore, it provides the function `get_coeff_vec` that parses an _apron expression_ into a vector of coefficients if the _apron expression_ has an affine form.
-**ExpressionBounds**: functor that takes the same arguments as **VarManagement**. It's interface is `SharedFunctions.ConvBounds` and calculates the bounds of an `apron expression`. As it is specifically used for the new affine equality domain, it can only provide bounds if the expression contains known constants only and in that case, min and max are the same.
- **D2**: functor that takes the same arguments as **VarManagement**. Provides the rest of the functions required by `RelationDomain.D2` (including the lattice operations such as `meet`, `join`, etc.) and the main code for the new affine equality domain.

**vectorMatrix**
*Interfaces/module types*
- **RatOps**: Abstracts the functions of the `Mpqf` module for rationals from Apron that implements multi-precision rationals.
               One could later exchange "Mpqf" with a different module that provides the functions specified by this interface.
- **Vector**: High-level abstraction of a vector. 
- **Matrix**: High-level abstraction of a matrix.
- **AbstractVector**: functor that includes the signature of `Vector` and takes `A:RatOps` as argument .It substitutes the abstract type _num_ (short for "numerical")  of `Vector` with type t of module `A`. Some functions inside have the suffix __pt_with_,
which means that the function could have potential side effects
- **AbstractMatrix**; A functor including sig of Matrix with arguments `A:RatOps` and `V:AbstractVector`. Similarly to AbstractVector, AbstractMatrix substitutes  `num` of `Matrix` with type t of `A:RatOps` but also the abstract type _vec_ with type t of a module `V:AbstractVector`. Some functions inside have the suffix __pt_with_,
which means that the function could have potential side effects

*Modules*
- **ConvenienceOps**: A functor parameterized by `RatOps` . It provides more readable infix operators for the functions of `RatOps`. It is designed to be included by modules that make use of `RatOps`'s functions
- **ListVector**: `AbstractVector` with type t `A.t List.t`. Reuses most of the already existing `List` functions such as `map2`, `rev`, etc. and adds some for vector manipulations (e.g. `insert_val` `remove_val`...). None of its functions has side-effects
- **ListMatrix**: `AbstractMatrix` with type `V.t list`. In contrast to `ListVector`, it does not include the functions of `List` directly but uses them to provide unique ones. It also includes various `normalize` functions that transform a matrix into reduced row echelon form. These are in general not efficient as they perform a full Gauss-Jordan elimination that is slow for lists.
- **ArrayVector**: `AbstractVector` with type t `A.t array` (from the `Batteries` library). Provides more efficient side-effecting functions than `ListVector`. Provides a copy function since the array data structure it contains is mutable.
- **ArrayMatrix**: `AbstractMatrix` with type `A.t array array` (from the `Batteries` library). Provides more efficient side-effecting functions than `ListMatrix`. Similar to `ListMatrix`, it provides a normalization function to reduce a matrix into reduced row echelon form. However, its normalization functions are faster not only because they use faster side-effect functions but also because they exploit that the input matrix/matrices are in reduced row echelon form already

**ApronDomain** (Contains most of its old code. Some parts have been moved. Not much new has been added)
*Changes:*
- **AffeqManager** has been added which makes it possible to run the apron affine equality implementation as well
- **Var** has been moved to sharedFunctions as it is also used by affineEqualityDomain now.
- **VarMetadataTbl** and **VM** were moved to relationDomain.
- **Convert* has been moved to sharedFunctions as it converts CIL expressions to apron expressions which is also used by `affineEqualityDomain`
- A few code elements for environment changes from functions as **remove_vars** etc. have been moved to sharedFunctions as they are needed in a similar way inside `affineEqualityDomain`
- `ov` parameter has been added to functions where functions of the Convert module are used. This is to also to make used of the improved overflow handling.
- `assert_inv`, `eval_interval_expr` and `eval_int`  have been moved into a new `AssertionModule` inside `SharedFunctions` as they are also used by `affineEqualityDomain`
- **ApronComponents** was renamed to `RelComponent` and moved to RelationDomain 
- **D2Complete** and **D2** have been added (containing functions of `RelationDomain.D2`). 
- **AD2Complete** was added which includes the functions of `SharedFunctions.AssertionModule` and **D2Complete**. It contains all functions specified by `RelationDomain.RelD2` . **AD2Complete** is required by `ApronPrecCompareUtil`

**RelationDomain** (Contains some of ApronDomain's code and new interfaces)

*New module types/ Interfaces*

- **RelVar** which abstracts the extended apron `Var` that was inside `apronDomain` (Moved to `sharedFunctions`).
- **D2** which defines most of the functions that are needed by `RelationAnalysis`. It includes for instance lattice functions such as `join`, `meet` ...
Some of the functions end with the suffix __pt_with_ which indicates that the function could have potential side effects.
In that case, the functions are supposed to return their modified input.
- **RelD2** which includes **D2** but extends it by a few functions. This signature is used by the `AssertionModule` in `sharedFunctions'
- **RD** which serves as a parameter for `RelationAnalysis.SpecFunctor`. The 3 modules that it contains must be implemented for each analysis and bundled into a module of this type. It is intended to group the implementation for variables and the domain.

*Moved modules + module types/ interfaces*
- **RelVM** (**VM** before) plus its interface **VarMetadata**, **VarMetadataTbl** and **V** were moved from `ApronDomain` into `RelationDomain`. Functor arguments for **RelVar** have been added.
- **RelComponent** (before:**ApronComponents**) was also moved here from `ApronDomain` and accepts **RelD2** now as functor argument instead of **ApronDomain.S2**

**SharedFunctions (Contains mostly code from `apronDomain` which has been adapted to fit to affineEqualityDomain`**)
*Moved modules + module types/interfaces*
- **Tracked** from `apronDomain`
- **Convert** from `apronDomain`. It takes a module of type **ConvBounds** as functor argument now. Used to convert CIL expressions to apron expressions.

*New module types/interfaces*
- **ConvBounds** an interface for **Bounds** which calculates bounds for expressions and is used inside the - **Convert** module
- **AssertionRelD2** : a more specific module type for `RelationDomain.RelD2` with **ConvBounds** integrated and various apron elements. It is designed to be the interface for the **D2** modules in `affineEqualityDomain` and `apronDomain` and serves as a functor argument for **AssertionModule**

*New modules*
- **EnvOps**: A module that includes various methods used by variable handling operations such as `add_vars`, `remove_vars` etc. in `apronDomain` and `affineEqualityDomain` 
- **AssertionModule**: Contains functions that are required by `RelationAnalysis` and defined by `RelationDomain.RelD2` and are exactly the same for `apronDomain` and `affineEqualityDomain` , i.e. they don`t have to be defined twice inside both domains. It uses the functions of its functor argument **AssertionRelD2** 

**RelationAnalysis (Contains the `SpecFunctor` of `apronDomain`)**
- **SpecFunctor** accepts `RelationDomain.RD` now instead of `ApronDomain.S2` and uses the domain and variable provided by it.
It also accepts `RelationPrecCompareUtil.Util` now as functor argument as `ApronPrecCompareUtil` usages inside have been removed.
Some of the definitely side-effecting functions ending with __with_ are now expected to only potentially have side effects and
their return values are therefore utilized.

**ApronAnalysis**
- **ExtendedSpecFunctor**: Uses the **SpecFunctor** defined inside `RelationAnalysis` and extends it by data storing methods etc. It also passes `ApronPrecCompareUtil` to `RelationAnalysis.SpecFunctor`. 
- Initialization methods (such as `spec_module`, `get-spec`) for apron analyses are also included and taken from the old `apronAnalysis`. They were slightly adapted to the new module structure (e.g. module **RD** with apron domain etc. are initialized inside)

**AffineEqualityAnalysis**
- Similar to `ApronAnalysis`, initialization methods for the new affine equality analysis were added. However, it does not add specific data storing methods to `RelationAnalysis.SpecFunctor` and passes `RelationPrecCompareUtil.DummyUtil` instead of an apron-specific implementation to the `SpecFunctor`

**RelationPrecCompareUtil and ApronPrecCompareUtil**
- **RelationPrecCompareUtil.Util** abstracts **ApronPrecCompareUtil** now and also provides a **DummyUtil** for the new affine equality analysis


sim642:
> can be run independently from the already existing apron analyses.

> Abstract states in the newly added domain are represented by structs containing a matrix and an apron environment.

I am immediately confused by this. If the new affeq domain/analysis uses Apron environments, variables and expressions, then it's not really independent from Apron. If Apron isn't installed, then the affeq domain cannot even be compiled.

To me, this is a particularly weird choice of abstraction, because it would be more logical to use non-Apron variables, expressions and environments (like the base analysis does) for the domain, which intends to be separate from Apron's. I would've expected the common interface between Apron domains and this affeq domain to use `Cil.varinfo`, `Cil.exp`, etc.

By still requiring the presence of Apron, the point of having this domain is defeated, because if I already have Apron, I might as well use the equivalent equalities domain provided by it. That's very likely more efficient and more reliable.

MartinWehking:
> > can be run independently from the already existing apron analyses.
> 
> > Abstract states in the newly added domain are represented by structs containing a matrix and an apron environment.
> 
> I am immediately confused by this. If the new affeq domain/analysis uses Apron environments, variables and expressions, then it's not really independent from Apron. If Apron isn't installed, then the affeq domain cannot even be compiled.
> 
> To me, this is a particularly weird choice of abstraction, because it would be more logical to use non-Apron variables, expressions and environments (like the base analysis does) for the domain, which intends to be separate from Apron's. I would've expected the common interface between Apron domains and this affeq domain to use `Cil.varinfo`, `Cil.exp`, etc.
> 
> By still requiring the presence of Apron, the point of having this domain is defeated, because if I already have Apron, I might as well use the equivalent equalities domain provided by it. That's very likely more efficient and more reliable.

@sim642 
The goal is not to get completely rid of apron. We rather tried to have an analysis that can be run in parallel to apron/other analyses. Currently, it's for example not possible to run the polyhedra and affeq analysis simultaneously. That's what we wanted to change. Plus, we"re also trying more or less to compare this implementation to the already existing affeq domain. The apron-independent interfaces that are provided can be seen as a nice-to-have feature so to speak. In future, one could add more relational analyses that use these interfaces without necessarily relying on apron, but this is beyond the scope of this specific project. It would also be possible to remove the few apron components that are used inside affeq domain/analysis, but again, this would be future work.

sim642:
> Currently, it's for example not possible to run the polyhedra and affeq analysis simultaneously. That's what we wanted to change.

Currently (on `master`) there's no affeq to begin with, so I guess you meant polyhedra (`Polka.manager_alloc_loose`) and equalities (`Polka.manager_alloc_equalities`)? That doesn't make sense either, because `loose` automatically can represent all equalities as well, so you wouldn't need to use them simultaneously.

A combination like octagon and equalities wouldn't fall under that, so isn't currently directly supported. But achieving that doesn't require implementing one of the two from scratch! Thanks to the genericity of `ApronAnalyis.SpecFunctor`, it could simply be instantiated twice with different Apron domains. The two registered analyses could automagically run simultaneously already.

Alternatively, Apron has (reduced) products of its domains, which could achieve the same.

MartinWehking:
> > Currently, it's for example not possible to run the polyhedra and affeq analysis simultaneously. That's what we wanted to change.
> 
> Currently (on `master`) there's no affeq to begin with, so I guess you meant polyhedra (`Polka.manager_alloc_loose`) and equalities (`Polka.manager_alloc_equalities`)? That doesn't make sense either, because `loose` automatically can represent all equalities as well, so you wouldn't need to use them simultaneously.
> 
> A combination like octagon and equalities wouldn't fall under that, so isn't currently directly supported. But achieving that doesn't require implementing one of the two from scratch! Thanks to the genericity of `ApronAnalyis.SpecFunctor`, it could simply be instantiated twice with different Apron domains. The two registered analyses could automagically run simultaneously already.
> 
> Alternatively, Apron has (reduced) products of its domains, which could achieve the same.

True, there is indeed no affeq on master at the moment but adding it required just a few more lines of code.
Yes, running affine equalities and polyhedra simultaneously doesn't make that much sense result-wise since polyhedra covers affeq as you said. For checking runtimes or checking other properties it could be interesting though. 

Having a completely apron-free implementation also didn't make sense for us since we wanted to keep the apron abstract syntax tree plus its conversion functions. For comparing our affeq implementation with the apron one this was somewhat crucial as we didn't want ours to suffer from wrong/incomplete  CIL expression handling. Thus, keeping it also gave us a better common basis for the comparison.
In future one can decide if they want to reuse the AST + conversion functions from apron as well or leave it out entirely for implementations of other relational domains.
Hope this gave a better justification for the existence of the new affeq analysis/domain

Do you perhaps have some remarks on the code as well? I would be really glad to get as much feedback as possible since the implementation is an integral part of my thesis.

.github/workflows/indentation.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           key: opam-ocp-indent-${{ runner.os }}-${{ matrix.ocaml-compiler }}
 
       - name: Set up OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v1 # intentionally use v1 instead of v2 because it's faster with manual caching: https://github.com/goblint/analyzer/pull/308#issuecomment-887805857
+        uses: ocaml/setup-ocaml@v2 # intentionally use v1 instead of v2 because it's faster with manual caching: https://github.com/goblint/analyzer/pull/308#issuecomment-887805857
```
michael-schwarz, line 38: Any specific reason for updating this?

michael-schwarz:
Why does this PR update `g2html`? Were there any necessary changes, or is this by mistake?

src/dune
```diff
@@ -24,9 +24,17 @@
       (apron -> apronAnalysis.apron.ml)
       (-> apronAnalysis.no-apron.ml)
     )
-    (select apronPriv.ml from
-      (apron -> apronPriv.apron.ml)
-      (-> apronPriv.no-apron.ml)
```
michael-schwarz, line 29: Does the `ApronPriv` (renamed `RelationPriv`) really not require apron anymore? (We do currently not check this in the CI, so one has to verify this locally in a switch where apron is not installed).

src/analyses/relationAnalysis.ml
```diff
@@ -0,0 +1,453 @@
...
+    {st with apr = res}
+
+
+  (* Function call transfer functions. *)
+
+  let enter ctx r f args =
+    let st = ctx.local in
+    if M.tracing then M.tracel "combine" "apron enter f: %a\n" d_varinfo f.svar;
+    if M.tracing then M.tracel "combine" "apron enter formals: %a\n" (d_list "," d_varinfo) f.sformals;
+    if M.tracing then M.tracel "combine" "apron enter local: %a\n" D.pretty ctx.local;
```
michael-schwarz, line 198: I guess all these "apron XXX" tracing outputs should be replaced with `relation` or something similar.

src/analyses/relationAnalysis.ml
```diff
@@ -0,0 +1,453 @@
...
+
+  let sync ctx reason =
+    (* After the solver is finished, store the results (for later comparison) *)
+    if !GU.postsolving then begin
+      let old_value = PCU.RH.find_default results ctx.node (AD.bot ()) in
+      let new_value = AD.join old_value ctx.local.apr in
+      PCU.RH.replace results ctx.node new_value;
+    end;
+    Priv.sync (Analyses.ask_of_ctx ctx) ctx.global ctx.sideg ctx.local (reason :> [`Normal | `Join | `Return | `Init | `Thread])
+end
```
michael-schwarz, line 453: Usually, we like to have a newline to end each file

src/cdomains/apron/affineEqualityDomain.apron.ml
```diff
@@ -0,0 +1,747 @@
...
+        in Some (List.sort sort c)
+      | exception NoSolution -> None
+
+    let find_opt = List.find_opt
+
+    let map2 f m v = List.map2 f m (V.to_list v)
+
+    let of_list l = l
+
+    let to_list m = m
```
michael-schwarz, line 284: Might be worth pulling all of this out into a separate module, making the domain less huge here.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -62,6 +56,15 @@ struct
   let name () = "Polyhedra"
 end
 
+module AffEqManager =
```
michael-schwarz, line 59: Maybe follow the format here that is also used for the other Managers, i.w. give a link to the documentation in a comment at the top.

src/cdomains/apron/sharedDomain.apron.ml
```diff
@@ -0,0 +1,291 @@
+open Cil
```
michael-schwarz, line 1: In what sense is `sharedDomain` a `domain` here? Maybe rename?

michael-schwarz:
This is a huge diff (by necessity) and that makes it very hard to review. Could you maybe provide an overview over the different modules (and module types) that are now used (there was a lot of them before already), and I think we now have more that are even more involved?

This would be similar in spirit to the overview over the different modules you give in the description of the PR, but at a higher level of detail. Also, it would be very helpful to indicate which files contain significant new code, and which contain code that has been moved from somewhere else. I think that would make it easier (for all of us) to review the PR in detail in detail and also offer more specific suggestions for some of the parts.

src/dune
sim642: The unlocked workflow, which checks without apron installed as well, can manually be run from the Actions tab. Although since the PR is coming from a fork, @MartinWehking needs to do that himself on his fork.

.github/workflows/indentation.yml
sim642: No, dependabot probably complained on their fork, but we have this suggestion ignored.

MartinWehking: yes, dependabot opened a PR on my fork. I have reverted it to v1 again.

src/cdomains/apron/sharedDomain.apron.ml
MartinWehking: I've renamed it to `sharedFunctions`. Although, I'm actually not sure if that's much better.

src/dune
MartinWehking: unlocked throws no errors in runs without apron. What's still quite odd though is that the 'build'  process fails in every run using Ocaml version 4.10.x with the message `Error: Unbound value equal` in `affineEqualityDomain` (line 108) but that's a different issue.

MartinWehking:
> This is a huge diff (by necessity) and that makes it very hard to review. Could you maybe provide an overview over the different modules (and module types) that are now used (there was a lot of them before already), and I think we now have more that are even more involved?
> 
> This would be similar in spirit to the overview over the different modules you give in the description of the PR, but at a higher level of detail. Also, it would be very helpful to indicate which files contain significant new code, and which contain code that has been moved from somewhere else. I think that would make it easier (for all of us) to review the PR in detail in detail and also offer more specific suggestions for some of the parts.

I've added a more detailed description to my initial comment.

sim642:
> Having a completely apron-free implementation also didn't make sense for us since we wanted to keep the apron abstract syntax tree plus its conversion functions. For comparing our affeq implementation with the apron one this was somewhat crucial as we didn't want ours to suffer from wrong/incomplete CIL expression handling. Thus, keeping it also gave us a better common basis for the comparison.

But the Apron `texpr1` that we get from CIL's `exp` is extremely close: https://github.com/goblint/analyzer/blob/4c7576af93584aae20ce7a7ba6f3edc34a378fa7/src/cdomains/apron/apronDomain.apron.ml#L158-L200
All it would've taken to not depend on Apron was in the affeq implementation to pattern match on CIL's `BinOp (PlusA, ...)` instead of Apron's `Binop (Add, ...)`.

> In future one can decide if they want to reuse the AST + conversion functions from apron as well or leave it out entirely for implementations of other relational domains.

I guess there's not much to be done about it at this point, but it's a weird early design decision that now is deeply integrated in this. It would require major refactoring now to change.

tests/regression/54-affeq/08-unsigned-overflow.c
```diff
@@ -0,0 +1,14 @@
+//SKIP PARAM: --set ana.activated[+] affeq  --enable ana.int.interval
+
+int main() {
+    //Overflow
+    int c = 2147483647;
```
sim642, line 5: The file is called _unsigned_ overflow, but this is signed.

src/cdomains/relationDomain.ml
```diff
@@ -0,0 +1,188 @@
...
+  val assign_var_parallel' : t -> var list -> var list -> t
+  val substitute_exp : t -> var -> exp -> bool -> t
+  val unify: t -> t -> t
+  val tag: t -> int
+  val relift: t -> t
+
+  val marshal: t -> marshal
+
+  val unmarshal: marshal -> t
+end
```
sim642, line 60: Instead of duplicating other interfaces this should include the standard ones (`Printable`, `Lattice`).

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -0,0 +1,300 @@
...
+  let keep_filter_with env f =
+    (* Instead of removing undesired vars,
+       make a new env with just the desired vars. *)
+    let vs' =
+      vars env
+      |> List.enum
+      |> Enum.filter f
+      |> Array.of_enum
+    in
+    Environment.make vs' [||]
```
sim642, line 221: Why are these `*_with` functions no longer in-place like they used to be? They instead copy the whole environment every time like the non-with ones used to.

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -0,0 +1,300 @@
...
+
+  let rec contains_unsigned_var expr =
+    match expr with
+    | Lval (Var v, NoOffset) when Tracked.varinfo_tracked v -> not @@ Cil.isSigned @@ Cilfacade.get_ikind_exp expr
+    | Const _ -> false
+    | exp -> begin match exp with
+                  | UnOp (_, e, _) -> contains_unsigned_var e
+                  | BinOp (_, e1, e2, _) -> contains_unsigned_var e1 || contains_unsigned_var e2
+                  | CastE (_, e) -> contains_unsigned_var e
+                  | _ -> false end
```
sim642, line 108: Why does this require a separate recursion over the expression instead of just having the signedness check be part of the actual conversion recursion?

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -0,0 +1,300 @@
...
+  include AD2
+
+  module Convert = Convert (Bounds)
+
+  let type_tracked typ =
+    isIntegralType typ
+
+  let varinfo_tracked vi =
+    (* no vglob check here, because globals are allowed in apron, but just have to be handled separately *)
+    type_tracked vi.vtype && not vi.vaddrof
```
sim642, line 251: These are now duplicated for no reason.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -487,7 +362,7 @@ struct
     let texpr1 = Texpr1.of_expr (A.env nd) (Var v') in
     A.substitute_texpr_with Man.mgr nd v texpr1 None
 
-  let meet_tcons d tcons1 =
+  let meet_with_tcons d tcons1 =
     let earray = Tcons1.array_make (A.env d) 1 in
     Tcons1.array_set earray 0 tcons1;
     A.meet_tcons_array Man.mgr d earray
```
sim642, line 368: This isn't an in-place operation, so the `with` in the name is misleading.

src/analyses/relationPriv.ml
```diff
@@ -45,340 +42,344 @@ module type S =
...
+    (** Locally may-written protected globals that have been continuously protected since writing. *)
+    (* TODO: is this right? *)
+    module W =
+    struct
+      include MayVars
+      let name () = "W"
+    end
+
+    module D = Lattice.Prod (P) (W)
+    module G = RD.D2
```
sim642, line 104: The fact that instead of `AD` one has to now use `RD.D2` everywhere is very annoying. The domain module contains a domain module in itself?

src/analyses/relationAnalysis.ml
```diff
@@ -0,0 +1,458 @@
...
+        let max = eval_int ctx (BinOp (Lt,  exp, kinteger64 ik int_max, intType)) in
+        match min, max with
+        | Some (x), _ when x = (IntOps.BigIntOps.of_int 1) -> true
+        | _, Some (x) when x = (IntOps.BigIntOps.of_int 1) -> true
+        | _, _ -> false
+
+  let no_overflow ctx exp =
+    let res = no_overflow ctx exp in
+    if M.tracing then M.tracel "no_ov" "no_ov %b exp: %s\n" res
+        (Pretty.sprint ~width:1 (Cil.printExp Cil.defaultCilPrinter () exp)); res
```
sim642, line 148: Why do this here? The expression conversion already does overflow handling.

src/cdomains/apron/affineEqualityDomain.apron.ml
```diff
@@ -0,0 +1,725 @@
...
+  val add : t -> t -> t
+  val sub : t -> t -> t
+  val mul : t -> t -> t
+  val div : t -> t -> t
+  val neg : t -> t
+  val abs : t -> t
+  val cmp : t -> t -> int
+  val to_string:  t -> string
+  val of_int: int -> t
+end
```
sim642, line 20: Why this? There's already `IntOps` for this exact reason to abstract over integer types.

src/cdomains/apron/affineEqualityDomain.apron.ml
```diff
@@ -0,0 +1,725 @@
...
+    change_d t env' false
+
+  let vars t = vars t.env
+
+  let type_tracked typ =
+    isIntegralType typ
+
+  let varinfo_tracked vi =
+    (* no vglob check here, because globals are allowed in apron, but just have to be handled separately *)
+    type_tracked vi.vtype && not vi.vaddrof
```
sim642, line 343: This again duplicates something from shared functions.

src/cdomains/apron/affineEqualityDomain.apron.ml
```diff
@@ -0,0 +1,725 @@
...
+            Matrix.remove_row (multiply_by_t' a sub_col a_r max) r, Matrix.remove_row (multiply_by_t' b sub_col b_r max) r, (max - 1)
+        in
+        let col_a, col_b = Matrix.get_col a s, Matrix.get_col b s in
+        match Int.compare (Matrix.num_rows a) r, Int.compare (Matrix.num_rows b) r with
+        | 1 , 1 ->  let a_rs, b_rs = Mpqf.to_float (Vector.nth col_a r), Mpqf.to_float (Vector.nth col_b r) in
+          begin match  a_rs, b_rs with
+            | 1., 1. -> lin_disjunc (r + 1) (s + 1) a b
+            | 1., 0. -> lin_disjunc r (s + 1) (case_two a r col_b) b
+            | 0., 1. -> lin_disjunc r (s + 1) a (case_two b r col_a)
+            | 0., 0. ->  let new_a, new_b, new_r = case_three a b col_a col_b r in
```
sim642, line 545: This match involves floating-point equality of _non-constant_ numbers, which is very fragile.

sim642 (changes requested):

sim642:
> * Another major change is that functions with side effects (usually ending with "_with") defined inside `apronDomain` were replaced by functional equivalents

This point is easily missed when looking at the diffs because a rename is also involved, but this is a no-go. The in-place functions of Apron are a feature and we very intentionally use them to achieve better performance. Because the functional equivalents always involve a memory copy and then do that in-place operation on the new memory.

There's a reason Apron provides these. By reverting that optimization you're (maybe unintentionally) shooting the Apron domains in the foot and crippling their performance, which invalidates the core goal of these changes to compare the two.

sim642:
> * implementations for vectors and matrices (realized through lists and lists of lists. Could be replaced in future

This is also a major point when it comes to performance and its comparison. Functional lists are infamously inefficient for the kinds of operations vectors and matrices are used for here (indexing). If I were to guess, these list manipulations make up a major bottleneck.

src/cdomains/apron/affineEqualityDomain.apron.ml
MartinWehking: Ok but do you perhaps have suggestions where I could move it? I could maybe create a new file `Matrix.ml` or something like that and place it inside _src/cdomains_

src/cdomains/relationDomain.ml
MartinWehking: `D2` includes `Lattice.S` now

src/analyses/relationPriv.ml
MartinWehking: Makes sense. `D2` is now directly included inside `RD`

src/cdomains/apron/affineEqualityDomain.apron.ml
MartinWehking: `IntOps` does unfortunately not include abstractions for rationals. The 'smallest' interface inside `IntOps`,` IntOpsBase`, would require shifting operations etc. which don't work for rationals.

src/cdomains/apron/sharedFunctions.apron.ml
MartinWehking: That was a naming issue, the `_with` ones also made use of these functions. The reason why those functions were put into `sharedFunctions` was that they were also used by `affineEqualityDomain` at a certain point. However, in the current state it just uses about 2 of them anymore. In the latest commit, I therefore moved some functions back again into `apronDomain` and put them in place again.

src/cdomains/apron/affineEqualityDomain.apron.ml
sim642: Confusingly, nothing in this interface implies nor requires the use of rationals to implement it though. There's just `of_int`, which suggests that integers would suffice and indeed every `IntOps` implementation also satisfies this interface (other than `compare` having the non-standard name `cmp`). What would break down if one were to try doing that?

MartinWehking: Ok that was true until my latest commit 6654acd where I added the `get_den` and `get_num` functions to the interface. Now it should be clearly distinct from `IntOps`

src/analyses/relationAnalysis.ml
MartinWehking: We came up with an improved overflow handling that queries whether the other analyses might evaluate to top (eval_int and then bounds check) for expression assignments and if not, an overflow can definitely be excluded. This yields better precision for `affeq` because unlike polyhedra and octagon it can not determine any inequalities itself

src/cdomains/apron/sharedFunctions.apron.ml
MartinWehking: This recursion function has been updated. It's now used to determine integer bounds for equality constraint checks when only one variable occurs inside the overall expression. We use that information for positive guard exploitation. I've updated my intial comment a bit to explain how this is useful if overflows are assumed to occur.
The reason why this is separate from the expression conversion is that it is only used in a very special case, while the conversion is used much more frequently. For all other analyses this check would also not be very useful

src/cdomains/apron/affineEqualityDomain.apron.ml
```diff
@@ -17,6 +17,16 @@ sig
...
+  val get_den: t -> IntOps.BigIntOps.t
+  val get_num: t -> IntOps.BigIntOps.t
+end
+
+module Mpqf = struct
+  include Mpqf
+
+  let get_den x = IntOps.BigIntOps.of_string @@ Mpzf.to_string @@ Mpqf.get_den x
+
+  let get_num x = IntOps.BigIntOps.of_string @@ Mpzf.to_string @@ Mpqf.get_num x
```
sim642, line 29: Just wondering, is your branch so far behind `master` that it doesn't have the `Z_mlgmpidl` module? Because the latter provides a way to convert between these without having to go through strings.

src/cdomains/apron/affineEqualityDomain.apron.ml
sim642: I'm wondering of something like `RatOps` would be clearer that it's about rationals, not just any arithmetic.

MartinWehking: Yes, that's indeed clearer. Renamed in a50ac6

MartinWehking:
> > * Another major change is that functions with side effects (usually ending with "_with") defined inside `apronDomain` were replaced by functional equivalents
> 
> This point is easily missed when looking at the diffs because a rename is also involved, but this is a no-go. The in-place functions of Apron are a feature and we very intentionally use them to achieve better performance. Because the functional equivalents always involve a memory copy and then do that in-place operation on the new memory.
> 
> There's a reason Apron provides these. By reverting that optimization you're (maybe unintentionally) shooting the Apron domains in the foot and crippling their performance, which invalidates the core goal of these changes to compare the two.

That makes sense. We've added the in-place functions to relationAnalysis again and indicated by a suffix "_pt_with" that the functions could have potential side-effects (which is the case for all "_pt_with" labeled functions in the apron domain but not for the ones inside affeq)

MartinWehking:
> > * implementations for vectors and matrices (realized through lists and lists of lists. Could be replaced in future
> 
> This is also a major point when it comes to performance and its comparison. Functional lists are infamously inefficient for the kinds of operations vectors and matrices are used for here (indexing). If I were to guess, these list manipulations make up a major bottleneck.

Okay a quick update on this. We have added a matrix data structure that relies on the `array` implementation provided by the `Batteries` library that outperforms the list version by far. The latter was indeed a major bottleneck inside our implementation, but we were also expecting that of course. What also harmed our runtime was the extremely slow `normalization` function which was called inside `leq` for instance. A function that is used quite often... There was also this idea of using dynamic matrices from `Batteries`. However, we eventually decided against it since most functions actually need an array copying one way or another. For instance removing a variable and resizing the input array could potentially break something in the `relationAnalysis` as it expects the input to remain unchanged (side effects...). One of the few functions where it could make sense is the `join` operation as it involves multiple recursion steps that reshape remove rows quite often. We have inspected time stats, however and saw that usually the `join` operation does not harm the performance that much.

src/cdomains/apron/apronDomain.apron.ml
MartinWehking: 21a4e189ddb8aa968383badb06a5ae4b473a3dbb

MartinWehking:
> Might make sense to also add a regression test for such things, such that we'll notice if we inadvertently break it again in the future!

56b4125b384ef9a89f249f65e643ba036153a236

src/cdomains/apron/affineEqualityDomain.apron.ml
MartinWehking: The module is currently called `VectorMatrix.ml ` in src/cdomains. Might not be the ideal location and could of course be moved

src/cdomains/apron/apronDomain.apron.ml
MartinWehking: I added a description. Since the underlying library is also Polka, the description might seem a bit duplicated, but that should be fine I think.

src/dune
MartinWehking: unlocked works without errors anymore.  9470349ad42bc98999fe7c16c88ce80edba022b2 and b287752600c0f9949d97a000da12fcbd26377511

MartinWehking:
> Why does this PR update `g2html`? Were there any necessary changes, or is this by mistake?

I've reverted the updates to all submodules. These were indeed by mistake

src/analyses/apron/affineEqualityAnalysis.apron.ml
```diff
@@ -0,0 +1,42 @@
...
+
+
+let () =
+  Printexc.register_printer
+    (function
+      | Apron.Manager.Error e ->
+        let () = Apron.Manager.print_exclog Format.str_formatter e in
+        Some(Printf.sprintf "Apron.Manager.Error\n %s" (Format.flush_str_formatter ()))
+      | _ -> None (* for other exceptions *)
+    )
```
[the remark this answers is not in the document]
MartinWehking: Fixed in eca76e91e06487d8f3acc2bb2ed06cf048f8fdd6

src/analyses/relationAnalysis.ml
MartinWehking: Should be resolved for all files

michael-schwarz:
Could you merge master into this one last time? Then, I think I'll start another run on sv-comp and then we should be good to go!

michael-schwarz:
The run of cotagons before and after finished now, here are the results.

[results.zip](https://github.com/goblint/analyzer/files/9724308/results.zip)

michael-schwarz:
After the fixes, most issues are fixed. [comparison.zip](https://github.com/goblint/analyzer/files/9846279/comparison.zip)

I think we should probably merge this to avoid to merge again and again, and also to make sure we don't do the merge shortly before the sv-comp deadline which may be risky. What do you think @sim642?

sim642:
Indeed, we shouldn't make such significant changes right before SV-COMP. Based on a quick look at the related thesis, it seems like in SV-COMP this doesn't give us an advantage over octagons both in time and score, right?

Since this involves quite big changes (that are hard to follow on diffs alone), I'd want to take a good look at this again, because it has been a while. Instead of rushing this now, it seems slightly more reasonable to me to have this merged _right after_ SV-COMP. I'm then willing to take care of any merge conflicts myself.

src/cdomains/apron/affineEqualityDomain.apron.ml
```diff
@@ -0,0 +1,630 @@
...
+    Environment.equal t1.env t2.env &&
+    match t1.d, t2.d with
+    | None, None -> true
+    | Some x1, Some x2 -> Matrix.equal x1 x2
+    | _ -> false
+
+  let hash t =
+    Hashtbl.hash t
+
+  let compare t1 t2 = Stdlib.compare t1 t2
```
sim642, line 270: Polymorphic hash and compare.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -427,9 +127,9 @@ sig
   val keep_vars : t -> Var.t list -> t
   val keep_filter : t -> (Var.t -> bool) -> t
   val forget_vars : t -> Var.t list -> t
-  val assign_exp : t -> Var.t -> exp -> t
+  val assign_exp : t -> Var.t -> exp -> bool Lazy.t -> t
```
sim642, line 130: What do these lazy bools mean?

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -515,25 +237,36 @@ sig
   val mem_var : t -> Var.t -> bool
   val assign_var_parallel' :
     t -> Var.t list -> Var.t list -> t
-  val meet_tcons : t -> Tcons1.t -> t
+  val meet_tcons : t -> Tcons1.t -> exp -> t
```
sim642, line 240: What does this `exp` mean?

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -732,23 +453,46 @@ struct
...
 
-  let meet_tcons d tcons1 =
+  let meet_tcons d tcons1 e =
     let earray = Tcons1.array_make (A.env d) 1 in
     Tcons1.array_set earray 0 tcons1;
-    A.meet_tcons_array Man.mgr d earray
+    let res = A.meet_tcons_array Man.mgr d earray in
+    match Man.name () with
+    | "ApronAffEq" ->
+      let overflow_res res = if IntDomain.should_ignore_overflow (Cilfacade.get_ikind_exp e) then res else d in
```
sim642, line 462: Why is this handled specially?

src/analyses/relationAnalysis.ml
```diff
@@ -0,0 +1,651 @@
...
+    include Priv.V
+    include StdV
+  end
+
+  module RV = RD.V
+
+  module PCU = PCU(RD)
+
+  (* Result map used for comparison of results *)
+  let results = PCU.RH.create 103 (*ToDO Remove this*)
```
sim642, line 26: Remove.

sim642 (changes requested):
I started adding some comments, but right now many of them are just for me. I will start doing fixes and dealing with the merge myself on this branch.

src/util/options.schema.json
```diff
@@ -1,15 +1,15 @@
...
+    "type": "object",
+    "properties": {
+      "files": {
+        "title": "files",
+        "description": "Files to analyze.",
+        "type": "array",
+        "items": {
+          "type": "string"
+        },
+        "default": []
```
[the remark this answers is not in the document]
michael-schwarz: Yes, it sometimes is not so clear what the correct thing should be during development and things keep changing. This happens in my PRs all the time as well.
I agree that it is a good idea to squash the changes here to end up with a more linear history. 

------
I think we should try to make the effort to phrase our comments such that they are guaranteed to be perceived as constructive. This is especially true for first-time contributors/students, who do not know many of us in person :blush:

src/cdomains/vectorMatrix.ml
```diff
@@ -0,0 +1,857 @@
...
+
+  val set_val: t -> int -> num ->  t
+
+  val set_val_pt_with: t -> int -> num -> t
+
+  val insert_val: int -> num ->  t ->  t
+
+  val apply_with_c: (num -> num -> num) -> num ->  t ->  t
+
+  val apply_with_c_pt_with: (num -> num -> num) -> num -> t -> t
```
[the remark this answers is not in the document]
sim642: I'm still not sure what that means in this context. Apron uses `_with` suffix for imperative/in-place modification and that's what our Apron analysis currently does on master as well, but here many (if not most) also have that `pt`.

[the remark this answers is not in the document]
michael-schwarz: The issue here was that whether these functions modify the value or not depends on the implementation. Thus, for these `pt` variants it may make a modification in place (e.g. if invoked for a array-based module), or may return a new copy (e.g. if invoked for a list-based module).

Thus, a caller may not rely on either behavior being the one that happens and must a) provide defensive copies if they want the original value to be unmodified  (as it is potentially `with`) and b) use the returned value instead of relying on the modification happening in-place (as it may return a new value).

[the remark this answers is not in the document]
sim642: I see. It would be possible to provide unambiguous interfaces by lifting immutable lists using refs, but we can skip that for now.

The bigger question is https://github.com/goblint/analyzer/pull/592#discussion_r1027993185, i.e. whether we even want to keep the inferior list versions around like that. If not, then that would also avoid this issue I guess.

src/analyses/relationAnalysis.ml
sim642: Actually we have this on master without such TODO for `exp.apron.prec-dump` for dumping octagons for some traces-rel comparison. @michael-schwarz, will this actually still be needed or can it be removed after all?

src/analyses/relationAnalysis.ml
```diff
@@ -0,0 +1,692 @@
...
+        let eval_int exp = Queries.ID.to_bool (ask.f (EvalInt exp)) in
+        (* TODO: duplicates assert_type_bounds *)
+        let type_min, type_max = IntDomain.Size.range ik in
+        (* TODO: why not Le? *)
+        let min = eval_int (BinOp (Lt, kintegerCilint ik type_min, exp, intType)) in
+        let max = eval_int (BinOp (Lt, exp, kintegerCilint ik type_max, intType)) in
+        match min, max with (* TODO: don't need both? *)
+        | Some true, _
+        | _, Some true -> true
+        | _, _ -> false
```
sim642, line 187: Am I the only one to whom this no overflow check doesn't make sense? It checks both bounds with strict inequality, although equaling a type bound is also overflow-free, and only one of the two bounds needs to fit the type. With wraparound, it doesn't even bother doing the bounds checks although they might actually still fit.

src/cdomains/vectorMatrix.ml
```diff
@@ -0,0 +1,857 @@
...
+      ["list", (module ListMatrix: AbstractMatrix);
+       "array", (module ArrayMatrix: AbstractMatrix);]
+    in
+    let matrix = (GobConfig.get_string "ana.affeq.matrix") in
+    match List.assoc_opt matrix options with
+    | Some man -> man
+    | None -> failwith @@ "Matrix " ^ matrix ^ " is not supported. Please check the ana.affeq.matrix setting."
+  )
+
+let get_matrix () = Lazy.force matrix
```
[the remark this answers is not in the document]
michael-schwarz: I think while the list implementation was interesting for benchmarking, we can remove it here in the interest of maintainability.

src/analyses/relationAnalysis.ml
michael-schwarz: I would vote to leave these facilities (`exp.apron.prec-dump`) there, at least until the corresponding paper is accepted. After that we can remove it, if it really hinders maintenance. Otherwise it is nice to have in case we want to do a follow-up paper.

tests/regression/57-affeq/15-mutex-meet.c
```diff
@@ -0,0 +1,25 @@
+//SKIP PARAM:  --set ana.activated[+] "affeq" --set ana.apron.domain "affeq" --set ana.affeq.matrix "array" --set sem.int.signed_overflow assume_none --set ana.relation.privatization mutex-meet
```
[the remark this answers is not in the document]
michael-schwarz: Yeah, these seem spurious if `apron` is not activated.

tests/regression/57-affeq/14-norm_inv.c
```diff
@@ -0,0 +1,33 @@
+//SKIP PARAM: --set ana.activated[+] affeq  --set ana.affeq.matrix "list" --sem.int.signed_overflow "assume_none"  --enable ana.int.interval
+// Snippet from https://github.com/sosy-lab/sv-benchmarks/blob/master/c/nla-digbench/hard2.c
+// Normalization should be triggered when an invertible expression is assigned.
```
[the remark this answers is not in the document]
michael-schwarz: It may have been a fixpoint issue or crash at some point. Can we add asserts that make sense here?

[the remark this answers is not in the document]
sim642: That would be great. I added some to 57-affeq/01-rel_simple from Karr's paper, but I'm not familiar enough with the details to know what should be asserted regarding this description.

src/cdomains/apron/apronDomain.apron.ml
michael-schwarz: >   (** Lazy bool ov parameter has been added to functions where functions of the Convert module are used.
>      This is to also to make used of the improved overflow handling. *)

src/cdomains/apron/apronDomain.apron.ml
michael-schwarz: I think this is the original expression to then be able to do the overflow checks for it.

src/cdomains/apron/apronDomain.apron.ml
michael-schwarz: I think is so apron `affeq` and ours behave similarly, such that one can make a fair comparison. The equivalent is done in `meet_tcons` here:

https://github.com/MartinWehking/analyzer/blob/0ab9127c45235a9cf3d32f8af9f8431cc60d9e58/src/cdomains/apron/affineEqualityDomain.apron.ml#L527

src/analyses/relationAnalysis.ml
michael-schwarz: For both exceptions above, `false` is the safe default. Otherwise this is a bit of a hack that we did because we were too lazy for a proper solution.

In cases where we are ignoring overflows, we just return true.

The values I get from `eval_int` have already been handling for overflow. The thing we abuse here is that if an overflow occurred and we did go to top, the upper AND lower bound will be top, if either of these is a non-top value, now overflow can have happened.

This come at the price that we sometimes spuriously assume overflows where it would not be necessary as the value simply is top without any overflows happening.

This logic does however not apply for unsigned types as they can have overflowed without the resulting value being top.

This is all a bit hacky, and maybe one should install proper `MayOverflow` queries later, but I think one can leave this for a follow-up PR.

src/cdomains/vectorMatrix.ml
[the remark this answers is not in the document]
sim642: Since I now removed the list-based implementations, I managed to also get rid of this ambiguous `_pt_with` things by replacing them with `_with` versions. So now these interfaces follow Apron's and our previous conventions.
Moreover, it minimizes the diffs as well (e.g. `ApronAnalysis` on master vs `RelationAnalysis` in this PR) – it's just renames, which makes such large changes easily reviewable.

tests/regression/57-affeq/14-norm_inv.c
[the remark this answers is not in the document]
michael-schwarz: I was not able to dig exactly into what it was (the commit that added this was improperly rebased and has some unresolved conflicts now), but suspect a fixpoint issue.

I would propose we just add a comment that says this was likely a fixpoint issue and still leave it here: if it caused a subtle bug we minimized from sv-comp tasks once, a similar bug may resurface later so I think it would be unwise to remove it, even if we don't know what exactly was the issue.

src/analyses/relationAnalysis.ml
michael-schwarz: Yes, it should!

src/cdomains/apron/apronDomain.apron.ml
sim642: I guess in the thesis this is described as "positive guard refinement for equality constraints in the context of
overflows", but here is no check for the constraint being equality.

Besides that, I don't really understand what this is achieving even, because 57-affeq/09-bounds_guards_no_ov and 57-affeq/10-bounds_guards.ov pass just as well without it.
Moreover, it's computing bounds of the single variable occurring in the constraint, but in affeq the lower and upper bound must be equal to begin with.
And then refinement is done regardless how the single variable occurs in the constraint. For example `x - 1 == 0` is refined in the same way as `x - 42 == 0`, which seems odd.

The mathematical description

$$
[x]^\sharp D_{refined} \in ikind(x)\text{ with }ikind : Vars \to \mathbb{N} \times \mathbb{N}
$$

in the thesis doesn't make sense to me either. $ikind(x)$ is a pair of naturals (or should it be integers?), but pairs cannot contain ( $\in$ ) some abstract value.

michael-schwarz: > Besides that, I don't really understand what this is achieving even, because 57-affeq/09-bounds_guards_no_ov and 57-affeq/10-bounds_guards.ov pass just as well without it.

Both of those are tests for our `affeq` and not the affine equalities that come with apron though.

sim642: I mean disabling this special logic in `AffineEqualityDomain` will still pass those.

michael-schwarz: I think the notation is a bit sloppy, it should actually be $\subseteq$ and `ikind` should be from `Vars to V^\sharp` or something, but I think the intent is still clear.

michael-schwarz: > but here is no check for the constraint being equality. 

I think this refers to the constraint inside the domain being equality between  a variable and a constant?

michael-schwarz: > Additionally, we now also refine after positive guards when overflows might occur and there is only one variable inside the expression and the expression is an equality constraint check (==).  We check after the refinement if the new value of the variable is outside its integer bounds and if that is the case, either revert to the old state or set it to bottom.

I think this comment explains the intention here.

michael-schwarz: Also, if I disable these extra refinements, for me test 57/02 crashes.

sim642: > I think this refers to the constraint inside the domain being equality between a variable and a constant?

It doesn't seem like that though: the affeq version of this explicitly checks for the refining constraint being equality/non-equality.

> > Additionally, we now also refine after positive guards when overflows might occur and there is only one variable inside the expression and the expression is an equality constraint check (==).  We check after the refinement if the new value of the variable is outside its integer bounds and if that is the case, either revert to the old state or set it to bottom.
> 
> I think this comment explains the intention here.

It doesn't make it any clearer to me. That says "after the refinement", but this logic _is_ the refinement.
Also if it's just about variables having out of ikind bounds integer values, what does it have to do with refining _equality_ constraints and why just one variable. That seems like a general reduction step for all variables in the environment.

> Also, if I disable these extra refinements, for me test 57/02 crashes.

That test crashing doesn't explain this refinement though. If this is removed, it should just be less precise and fail some assertion, not crash altogether. If it's just to avoid that crash, this is just a workaround for a crash (due to some particular form of constraint for some reason), not a refinement.

---

Is there an example that directly shows the benefit of this refinement? As in, without this the assertion is unknown, but with this it succeeds.

michael-schwarz:
~~~C
int main() {
    int x;
    int itv;
    int top;

    itv = 0;

    if(top) {
      itv = 10;
    }

    if(x == itv) {
      __goblint_check(x == itv);
    }
}
~~~

This is the sort of thing one would hope that this can handle, Unfortunately, it does not seem like it actually helps here.

michael-schwarz: I have identified cases where it matters though. Consider, e.g., the following program


~~~C
//SKIP PARAM: --set ana.activated[+] affeq --set sem.int.signed_overflow assume_top --enable ana.int.interval

int main() {
    int top;

    if(top == 8) {
      __goblint_check(top == 8);
    }
}
~~~

With this refinement, by inspecting the HTMl output, we can see that affeq knows that `top-8=0`, without it, it only has `T`. Though I have not succeeded making an example where this shows in terms of asserts.

michael-schwarz:
~~~C
//SKIP PARAM: --set ana.activated[+] affeq --set sem.int.signed_overflow assume_top --enable ana.int.interval

int main() {
    int top;
    int blarg = top;

    if(top == 8) {
      __goblint_check(top == 8);
      __goblint_check(top == blarg);
    }
}
~~~

Took some further constructing, but this example requires the refinement to work.

sim642: Thanks for constructing the example! I'll add that as a test here.
I originally wanted to extract this common code, but ended up with lesser refactoring since that would've required another functor.

I'm still a bit puzzled by this special logic, but I'll just leave it.
There's no operation that could even overflow here, so refining with `top == 8` should succeed regardless of signedness or overflow settings.

michael-schwarz: The example where things overflow is checks such as `if(x-2 == MAX_INT)` and such, where it would be wrong to add `x = MAX_INT + 2` as s constraint, as that involves a signed wraparound.

sim642: I'm not sure checking the single variable occurring in the condition is sufficient though.
For example, consider the affeq state `x + y == MAX_INT`  refined by the condition `y == -1`. By Gaussian elimination, this will also give `x == MAX_INT + 1`, but just checking the bounds of the single variable `y` in the condition doesn't catch that.

sim642:
@michael-schwarz I think I've now done all the tinkering on this that I intended, so you can also have a final look.

Just a reminder here at the end again: this is to be **squash merged**.

src/util/options.schema.json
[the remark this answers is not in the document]
MartinWehking: Noted. Next time I'll take more care of this. I remember having lots of troubles with indentation...

michael-schwarz (approved):
LGTM! Thanks @MartinWehking and thanks @sim642 for the improvements!

sim642 (approved):

michael-schwarz:
I guess we'll also need to make changes in https://github.com/goblint/bench now that this is merged.

PR #595, opened 2022-02-15
Extract MHP analysis, simplify printing in race output

### LINKED ISSUE #552: User-friendly output of thread IDs and MHP
Since #331 and #518 the output of races has become much harder to read.
1. The thread IDs with history can be long and even when they're simple and unique, there's unnecessary internal notation, e.g. `([main], {})`.
2. The MHP component also contains the thread ID (which in the default configuration is the same duplicated) and two sets of such verbose thread IDs (created and must joined).

We should think of some neater way to output these such that the race output isn't such incomprehensible wall of text.

### LINKED ISSUE #571: Configuration of access collecting precision and performance
The following table gives an overview of producing race warnings for ypbind_comb from bench:

| Optimization | Race warning time |
| --- | --- |
| Baseline (#551, d2a50c589e231199b8075df2811663984472ccb4) | 1.822s |
| Access (def70a21ab0adf7d594e1c3c60068bf2fa02d309) | 0.917s |
| MCP (#570) | 0.728s |
| Exclude MHP | 0.420s |
| Exclude node | 0.258s |
| Exclude MHP and node | 0.104s |

This confirms that the MHP component for collected accesses can have a significant performance impact on race warnings. The reason is that for each access location it can create many more accesses based on the thread ID and especially the created threads set. So in large programs with many variables and accesses, this has the measurable impact from above.

Therefore it might be useful to be able to dynamically (de)activate what information is collected for accesses to trade off between precision and performance. If on these programs MHP doesn't really rule out races, then for interactive benchmarking it would be beneficial to disable it and save time.

sim642:
Closes #571. Closes #552.

### Changes
1. Extracts the MHP component of accesses to be provided by a separately activated mhp analysis, as suggested by https://github.com/goblint/analyzer/issues/571#issuecomment-1023243418. That analysis does nothing on its own, but just provides the more precise MHP component to accesses using data queried from other analyses.
2. The mhp analysis now needs to be explicitly enabled, it's not activated by default, because it's not clear whether its worth doing on most benchmarks where there are other bigger sources of imprecision. It's now easy to experiment either way and if useful for a benchmark, add a conf file. The few tests for MHP races now explicitly activate it.
3. History thread ID prefixes are now "unreversed", so previous `[t1, main]` is shown as `[main, t1]`. The reversed order is an internal optimization which could be misleading compared to the relational traces paper.
4. History thread ID sets are omitted when empty, so previous `([t1, main], {})` is shown as `[main, t1]`.
5. If `created` or `must_joined` components of `MHP` are empty, they are omitted for clarity.


michael-schwarz:
It makes sense to have it configurable... What strikes me as odd though is that in the default setting we now compute the expensive thread ids, only to then not really use them.

This seems strange, so I'd suggest we enable this by default.

sim642:
That's true... I guess my point is even a bit broader and orthogonal then because history thread IDs aren't used by default analyses and privatizations either. So maybe all the bigger benchmarking should start with the cheaper old thread IDs.

I suppose I could just add mhp to default analyses then still.

src/analyses/threadId.ml
```diff
@@ -96,25 +96,20 @@ struct
...
-    include Printable.Prod (Printable.Option (ThreadLifted) (struct let name = "nonunique" end)) (MHP)
-    let name () = "thread * mhp"
+    include Printable.Option (ThreadLifted) (struct let name = "nonunique" end)
+    let name () = "thread"
     let may_race (t1: t) (t2: t) = match t1, t2 with
-      | (Some t1, _), (Some t2, _) when ThreadLifted.equal t1 t2 -> false
-      | (_, mhp1), (_, mhp2) when not (MHP.may_happen_in_parallel mhp1 mhp2) -> false
-      | (_, _), (_, _) -> true
-    let should_print _ = true
+      | Some t1, Some t2 when ThreadLifted.equal t1 t2 -> false
```
michael-schwarz, line 102: I think this warrants a comment that this information is only recorded for unique threads in the first place, hence `equal` is correct here.

michael-schwarz (approved):
I think the changes where we add the `mhp` analysis to regression tests can be reverted, as it is now part of the defaults, right?

sim642:
> I think the changes where we add the `mhp` analysis to regression tests can be reverted, as it is now part of the defaults, right?

Could be, but the test suite is more robust this way when the feature under test is made explicit. Otherwise changes in the defaults could cause tests to fail, just because there's an implicit assumption about what the defaults are, not that disabling mhp is an unsoundness problem or something.

PR #596, opened 2022-02-15
Extract function and node specific logic from TD3

sim642:
Part of issue #545.

This is a quick idea I had to move all the incrementally changed functions stuff out of TD3. The point is to expose separate constraint system incremental data to TD3, which just lists the unknowns to obsolete, delete and reluctantly resolve. These lists are populated by `FromSpec` itself, which is the last thing that still should know about the specifics of the constraints.
With this change it's actually unnecessary for the solver to even have access to `iter_vars` from #549 because `FromSpec` uses it directly to do the conversion.

~~This is currently on top of #391 although something similar could be done on master as well. Although the annoyance is that it would have to be essentially redone completely there since the changes code in TD3 is changed significantly by #391, particularly by #549.~~

### TODO
- [x] Remove `iter_vars` access from solver.
- [x] Fix server non-relifting in TD3.
- [ ] Benchmark performance change due to intermediate lists?


src/framework/constraints.ml
```diff
@@ -760,6 +790,133 @@ struct
...
+    (* Unknowns from partially changed functions need only to be collected for eager destabilization when reluctant is off *)
+    (* We utilize that force-reanalyzed functions are always considered as completely changed (and not partially changed) *)
+    if not reluctant then (
+      List.iter (fun (f, pn, _) ->
+          List.iter (fun n ->
+              mark_node obsolete_prim f n
+            ) pn;
+          mark_node obsolete_ret f (Function f);
+        ) part_changed_funs;
+    );
```
jerhard, line 864: When merging this, care should be taken that the changes to `td3.ml` #841 are moved here.

Issue #589, opened 2022-02-03
Parallel preprocessing and parsing


sim642:
In the context of interactive analysis, minimizing the wall time of preprocessing and parsing of large projects with many files would be beneficial. But this would also be useful non-incrementally.

## Parallel preprocessing
Currently we use `Unix.system` to run the preprocessor, but this function waits until that subprocess terminates to start preprocessing the next file. Such sequentialization is completely unnecessary and could be performed in parallel instead.

Doing so doesn't require running any OCaml code in parallel! Instead we could use functions from here: https://ocaml.org/api/Unix.html#1_Highlevelprocessandredirectionmanagement. We could easily start all the preprocessor subprocesses first and not wait for any termination in between. And then wait for all of them to terminate. This is completely OS-level subprocess management, with no parallelism or even concurrency inside Goblint.

## Parallel parsing
CIL's parsing is implemented in OCaml, so it's not as simple to parallelize. But again, there might not be a need to wait for Multicore OCaml. Instead one of these libraries might be sufficient:
* https://opam.ocaml.org/packages/parmap/
* https://opam.ocaml.org/packages/parany/

What they do instead of Multicore OCaml is to fork the Goblint process, open a socket pair between the parent and child process and execute different (parsing) code in the forked child process. Then the resulting OCaml data structure is sent back to the parent Goblint process through the socket. These libraries completely automate that process and make it transparent.

This should work because parsing the individual files is an independent process. No data needs to be continuously shared on the OCaml heap, so the concurrent GC of Multicore OCaml isn't needed either.

PR #597, opened 2022-02-16
Parallel preprocessing

sim642:
Implements the first half of #589.

The separate calling of `Unix.open_process` and `Unix.close_process` was extremely simple, but trying it out on OpenSSL it revealed a problem: it crashed with the Unix error `EMFILE` (Too many open files by the process). Apparently starting 1009 child processes at the same time isn't a good idea.

Of course `ulimit -n` could've worked around the problem, but I thought I'd be nice and implement a `ProcessPool` module, which runs at most `jobs` subprocesses in parallel.

By default (`-j 0`), the number of parallel jobs is derived from the number of CPU cores. Otherwise the `-j` command line argument (which is really the `jobs` option) controls the limit.

### OpenSSL
Trying this on OpenSSL (just preprocessing, no parsing or merging), with `-j 1` it takes:
```
real	0m37,266s
user	0m28,811s
sys	0m8,312s
```

Whereas with `-j 16` it takes:
```
real	0m5,856s
user	1m4,144s
sys	0m16,108s
```

So here the preprocessing takes **6.4 times less wall time**!

### DIFF
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -155,18 +155,13 @@ let basic_preprocess ~all_cppflags fname =
   (* The actual filename of the preprocessed sourcefile *)
   let nname =  Filename.concat !Goblintutil.tempDirName (Filename.basename fname) in
   if Sys.file_exists (get_string "tempDir") then
-    nname
+    (nname, None)
   else
     (* Preprocess using cpp. *)
     (* ?? what is __BLOCKS__? is it ok to just undef? this? http://en.wikipedia.org/wiki/Blocks_(C_language_extension) *)
     let command = (Preprocessor.get_cpp ()) ^ " --undef __BLOCKS__ " ^ String.join " " (List.map Filename.quote all_cppflags) ^ " \"" ^ fname ^ "\" -o \"" ^ nname ^ "\"" in
     if get_bool "dbg.verbose" then print_endline command;
-
-    try match Unix.system command with
-      | Unix.WEXITED 0 -> nname
-      | _ -> eprintf "Goblint: Preprocessing failed."; raise Exit
-    with Unix.Unix_error (e, f, a) ->
-      eprintf "%s at syscall %s with argument \"%s\".\n" (Unix.error_message e) f a; raise Exit
+    (nname, Some {ProcessPool.command; cwd = None})
 
 (** Preprocess all files. Return list of preprocessed files and the temp directory name. *)
 let preprocess_files () =
@@ -282,7 +277,14 @@ let preprocess_files () =
   if get_bool "ana.sv-comp.functions" then
     extra_arg_files := find_custom_include "sv-comp.c" :: !extra_arg_files;
 
-  List.concat_map preprocess_arg_file (!extra_arg_files @ !arg_files)
+  let preprocessed = List.concat_map preprocess_arg_file (!extra_arg_files @ !arg_files) in
+  let preprocess_tasks = List.filter_map snd preprocessed in
+  let terminated task = function
+    | Unix.WEXITED 0 -> ()
+    | process_status -> failwith (GobUnix.string_of_process_status process_status)
+  in
+  ProcessPool.run ~terminated preprocess_tasks;
+  List.map fst preprocessed
 
 (** Possibly merge all postprocessed files *)
 let merge_preprocessed cpp_file_names =
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -284,7 +284,7 @@ let preprocess_files () =
     | Unix.WEXITED 0 -> ()
     | process_status -> failwith (GobUnix.string_of_process_status process_status)
   in
-  ProcessPool.run ~terminated preprocess_tasks;
+  ProcessPool.run ~jobs:(Goblintutil.jobs ()) ~terminated preprocess_tasks;
   List.map fst preprocessed
 
 (** Possibly merge all postprocessed files *)
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -90,6 +90,7 @@ let option_spec_list =
   let tmp_arg = ref "" in
   [ "-o"                   , Arg.String (set_string "outfile"), ""
   ; "-v"                   , Arg.Unit (fun () -> set_bool "dbg.verbose" true; set_bool "printstats" true), ""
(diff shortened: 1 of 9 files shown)


src/maingoblint.ml
```diff
@@ -155,18 +156,13 @@ let basic_preprocess ~all_cppflags fname =
   (* The actual filename of the preprocessed sourcefile *)
   let nname =  Filename.concat !Goblintutil.tempDirName (Filename.basename fname) in
   if Sys.file_exists (get_string "tempDir") then
-    nname
+    (nname, None)
   else
     (* Preprocess using cpp. *)
     (* ?? what is __BLOCKS__? is it ok to just undef? this? http://en.wikipedia.org/wiki/Blocks_(C_language_extension) *)
```
michael-schwarz, line 162: Not really part of this PR, but we should take the opportunity to update this comment. It is indeed the thing that is linked and seems to be unavailable in GCC (except for GCC variants shipped by Apple). 

https://clang.llvm.org/docs/BlockLanguageSpec.html

src/util/goblintutil.ml
```diff
@@ -183,3 +183,8 @@ let rec for_all_in_range (a, b) f =
   else f a && (for_all_in_range (BI.add a (BI.one), b) f)
 
 let dummy_obj = Obj.repr ()
+
+let jobs () =
+  match get_int "jobs" with
+  | 0 -> Cpu.numcores ()
+  | n -> n
```
michael-schwarz, line 190: Would this not be fit better into `Gobconfig` as it is a configuration option?

src/util/options.schema.json
```diff
@@ -217,6 +217,12 @@
       "type": "boolean",
       "default": false
     },
+    "jobs": {
+      "title": "jobs",
+      "description": "Maximum number of parallel jobs. If 0, then number of cores is used.",
```
michael-schwarz, line 222: Maybe we should clarify here that (for now?) this is only about preprocessing.

src/util/compilationDatabase.ml
```diff
@@ -87,8 +76,8 @@ let load_and_preprocess ~all_cppflags filename =
       let cwd = reroot obj.directory in
       if GobConfig.get_bool "dbg.verbose" then
         Printf.printf "Preprocessing %s\n  to %s\n  using %s\n  in %s\n" file preprocessed_file preprocess_command cwd;
-      system ~cwd preprocess_command; (* command/arguments might have paths relative to directory *)
-      Some preprocessed_file
+      let preprocess_task = {ProcessPool.command = preprocess_command; cwd = Some cwd} in (* command/arguments might have paths relative to directory *)
```
michael-schwarz, line 79: Is prefixing `command` with `ProcessPool` necessary here for the correct type to be inferred?

src/util/processPool.ml
```diff
@@ -0,0 +1,44 @@
...
+        | None ->
+          Unix.open_process task.command
+      in
+      let pid = Unix.process_pid proc in
+      Hashtbl.replace procs pid (task, proc);
+      run tasks
+    | [] when Hashtbl.length procs = 0 ->
+      ()
+    | _ ->
+      let (pid, status) = Unix.wait () in
```
michael-schwarz, line 30: Maybe a quick comment about what `Unix.wait ()` does (wait for a child process to terminate) would be helpful to understand what's going on here without familiarity with the `Unix` module. The rest of the used functions is quite straight-forward.

src/maingoblint.ml
```diff
@@ -155,18 +156,13 @@ let basic_preprocess ~all_cppflags fname =
   (* The actual filename of the preprocessed sourcefile *)
   let nname =  Filename.concat !Goblintutil.tempDirName (Filename.basename fname) in
   if Sys.file_exists (get_string "tempDir") then
-    nname
+    (nname, None)
```
michael-schwarz, line 159: This case seems a bit un-intuitive, when does that happen?

michael-schwarz (approved):

src/util/compilationDatabase.ml
sim642: Yes, because the `command_object` type in this module also has a `command` field.

src/maingoblint.ml
sim642: Apparently the `tempDir` option is for "Reuse temporary directory for preprocessed files". So it assumes that the preprocessed file already exists in the temporary directory.

I've been thinking about generalizing the concept to also apply to compilation database preprocesing, but I'd leave that for another PR.

src/util/goblintutil.ml
sim642: Not sure, because `GobConfig` only implements the JSON-structured storage and lookup in general, but nothing specific to particular options.

src/maingoblint.ml
sim642: I've always wondered what exactly is this undefine achieving, what problem is it solving and whether it's still necessary. Because we only undefine it for basic preprocessing, but not compilation database or cilly.

michael-schwarz: Someone who uses OS X would have to investigate that. @vogler @vesalvojdani ?

sim642:
I'm now thinking that actually it might be safer to have `-j 1` as the default, especially if we want to use multiple jobs for more exotic tricks like #598 as well. I don't think other programs like `make` implicitly use all cores, but require the user to explicitly choose that.

It would also be nicer for regression testing, because now that script itself will start Goblints on call cores, each of which also wants to run multiple preprocessors, at which point the CPU is overloaded anyway.

src/maingoblint.ml
sim642: It was introduced in [fd29d6d6873e5833d58c64cd1a32a02271142de6](https://github.com/goblint/analyzer/commit/fd29d6d6873e5833d58c64cd1a32a02271142de6#diff-c4a7423227c515488a65063285c91d2defd87ab7962f4ed66c2f76bd4f728f98R149) without any explanation when the standards for Goblint weren't as high.

sim642: There's also this weird page for some CIL-based tool that suggests a bunch of OSX workarounds: http://tigress.cs.arizona.edu/machine.html.

sim642: Moved to #605.

PR #600, opened 2022-02-17
Incremental analysis: Eagerly analyze functions in `force-reanalyze`, even when `reluctant` is on

### LINKED ISSUE #508: Make precision refinement compatible with reluctant destabilization

jerhard:
#### Problem
For functions for which the precision setting was changed for the incremental run, the analysis cannot be done reluctantly, as the `enter` needs to be redone for the precision adjustment to take effect, which does not happen in an reluctant incremental analysis. See issue #508.
Functions for which the precision was adjusted are listed in `incremental.force-reanalyze.funs`. Thus, a reluctant analysis should not be performed for functions contained in this set.

Note that functions for which the precision was changed *through an in-code annotation* there is no problem with the reluctant analysis: This is because the call-sites of such functions are detected as changed; thus all functions calling such a function with a new in-code precision annotation are reanalyzed. So no further changes are required for the in-code annotation case.

#### Background on reluctant incremental analysis
The reluctant analysis stores the old abstract values at the return nodes of changed functions; it omits the eager destabilization of these changed functions. Instead it calls solve on the return nodes of changed functions and checks whether the abstract state changed.

#### Description of changes
This PR adapts the reluctant analysis such that the entry nodes of functions in force-reanalyze are eagerly destabilized. This is done by collecting the function entry nodes of such functions, and performing the destabilization of  such function entry unknowns (in `obsolete_entry`) even when `reluctant` is on.  
The call to `solve` on the return node of these functions is therefore omitted. This is achieved by not collecting unknowns belonging to return nodes of functions in `incremental.force-reanalyze.funs` in `obsolete_ret`.

Closes #508.

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -16,8 +16,6 @@ open Messages
 open CompareCIL
 open Cil
 
-module VarinfoSet = Set.Make(CilType.Varinfo)
-
 module WP =
   functor (Arg: IncrSolverArg) ->
   functor (S:EqConstrSys) ->
@@ -630,11 +628,10 @@ module WP =
         in
 
         let reluctant = GobConfig.get_bool "incremental.reluctant.on" in
-        let force_reanalyze = VarinfoSet.of_list (List.map (fun f -> f.svar) S.increment.changes.force_reanalyze) in
         let reanalyze_entry f =
           (* destabilize the entry points of a changed function when reluctant is off,
              or the function is to be force-reanalyzed  *)
-          (not reluctant) || VarinfoSet.mem f.svar force_reanalyze
+          (not reluctant) || CompareCIL.VarinfoSet.mem f.svar S.increment.changes.force_reanalyze
         in
         let obsolete_ret = HM.create 103 in
         let obsolete_entry = HM.create 103 in
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -642,7 +642,7 @@ module WP =
         let obsolete_prim = HM.create 103 in
 
         (* When reluctant is on:
-          Only add function entry nodes to obsolete_entry if they are in force-reanalyze *)
+           Only add function entry nodes to obsolete_entry if they are in force-reanalyze *)
         List.iter (fun f ->
             if reanalyze_entry f then
               (* collect function entry for eager destabilization *)
@@ -671,7 +671,7 @@ module WP =
         );
 
         if not (HM.is_empty obsolete_entry) || not (HM.is_empty obsolete_prim) then
-           print_endline "Destabilizing changed functions and primary old nodes ...";
+          print_endline "Destabilizing changed functions and primary old nodes ...";
         HM.iter (fun k _ ->
             if HM.mem stable k then
               destabilize k
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -629,12 +629,27 @@ module WP =
             )
         in
 
+        let force_reanalyze = StringSet.of_list @@ GobConfig.get_string_list "incremental.force-reanalyze.funs" in
+        let eager = (not (GobConfig.get_bool "incremental.reluctant.on")) in
+        let reanalyze_entry f =
+          (* destabilize the entry points of a changed function when reluctant is off,
+             or the function is to be force-reanalyzed  *)
+          eager || StringSet.mem f.svar.vname force_reanalyze
+        in
+
         let obsolete_ret = HM.create 103 in
         let obsolete_entry = HM.create 103 in
         let obsolete_prim = HM.create 103 in
+
+        (* When reluctant is on:
+          Only add function entry nodes to obsolete_entry if they are in force-reanalyze *)
         List.iter (fun f ->
-            mark_node obsolete_entry f (FunctionEntry f);
-            mark_node obsolete_ret f (Function f);
+            if reanalyze_entry f then
+              (* collect function entry for eager destabilization *)
+              mark_node obsolete_entry f (FunctionEntry f)
+            else
+              (* collect function return for reluctant analysis *)
+              mark_node obsolete_ret f (Function f)
           ) changed_funs;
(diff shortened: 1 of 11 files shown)


jerhard:
An alternative approach to fix #508 would be to mark functions that call functions in `incremental.force-reanalyze.funs` as changed in `CompareCIL`. This way, no changes to the solver would be required, but only to the incremental pre-processing. This approach would result in the same behavior as if the annotation were done in the code (instead of through an configuration option and with `force-reanalyze`). However, this also means that fewer old results are reused.

src/solvers/td3.ml
```diff
@@ -651,44 +668,43 @@ module WP =
...
+        if not (HM.is_empty obsolete_entry) || not (HM.is_empty obsolete_prim) then
+          print_endline "Destabilizing changed functions and primary old nodes ...";
+        HM.iter (fun k _ ->
+            if HM.mem stable k then
+              destabilize k
+          ) obsolete_entry;
+        HM.iter (fun k _ ->
+            if HM.mem stable k then
+              destabilize k
+          ) obsolete_prim;
```
jerhard, line 682: This will probably cause problems when `reluctant` and CFG-comparison are activated.

jerhard: Should be resolved with only collecting `obsolete_prim` when `reluctant` is off.

src/solvers/td3.ml
```diff
@@ -627,22 +629,40 @@ module WP =
...
+           Only add function entry nodes to obsolete_entry if they are in force-reanalyze *)
         List.iter (fun f ->
-            mark_node obsolete_entry f (FunctionEntry f);
-            mark_node obsolete_ret f (Function f);
+            if reanalyze_entry f then
+              (* collect function entry for eager destabilization *)
+              mark_node obsolete_entry f (FunctionEntry f)
+            else
+              (* collect function return for reluctant analysis *)
+              mark_node obsolete_ret f (Function f)
```
sim642, line 652: Wasn't there some reason before, why `Function` was marked obsolete even though obsoleting/destabilizing `FunctionEntry` should do the same?

src/solvers/td3.ml
```diff
@@ -627,22 +629,40 @@ module WP =
...
           ) changed_funs;
-        List.iter (fun (f, pn, _) ->
-            List.iter (fun n ->
-                mark_node obsolete_prim f n
-              ) pn;
-            mark_node obsolete_ret f (Function f);
-          ) part_changed_funs;
+        (* Unknowns from partially changed functions need only to be collected for eager destabilization when reluctant is off *)
```
sim642, line 654: But also when the partially changed function is force reanalyzed, no?

src/solvers/td3.ml
```diff
@@ -627,12 +629,27 @@ module WP =
             )
         in
 
+        let force_reanalyze = StringSet.of_list @@ GobConfig.get_string_list "incremental.force-reanalyze.funs" in
+        let eager = (not (GobConfig.get_bool "incremental.reluctant.on")) in
+        let reanalyze_entry f =
+          (* destabilize the entry points of a changed function when reluctant is off,
+             or the function is to be force-reanalyzed  *)
+          eager || StringSet.mem f.svar.vname force_reanalyze
+        in
```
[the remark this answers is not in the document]
sim642: I guess that could be covered by #596, but I'm a bit hesitant on doing that since it will make the master vs interactive branch TD3s significantly more different and thereby complicating merges (either way) between the branches.

But for now, a possibility might be an extra field in `change_info` about force reanalyze functions and the current logic inside the solver could look that up instead of dealing with the option. The incremental preprocessing would then be in charge of populating the list appropriately.

src/solvers/td3.ml
jerhard: Force-reanalyzed functions are not (and should not) be treated as partially changed: 
https://github.com/goblint/analyzer/blob/304cfd62b978352c62fbc74dabc975971ca696fb/src/incremental/compareCIL.ml#L37-L40

src/solvers/td3.ml
jerhard: The unknowns collected in `obsolete_ret` are really only used in the `reluctant` mode, to collect `old_ret`, a mapping of unknowns to their old value in `rho` and their `infl` set. This `old_ret` is then used iterate over and call `solve` on; if the new result is different than the old one for an unknown, it is destabilized.
https://github.com/goblint/analyzer/blob/b6b9b44909e9709086af38f88ad799afe0c06db7/src/solvers/td3.ml#L645-L653
So in the regular incremental mode, only `obsolete_entry`  and `obsolete_prim` are used for eager destabilization.

sim642: Ah, right, yes. I missed that they get collected into different sets and just remembered some old discussion.

src/solvers/td3.ml
sim642: Right. I guess what makes it confusing is how `incremental.force-reanalyze.funs` is handled in two totally different places now. The suggestion from https://github.com/goblint/analyzer/pull/600#discussion_r809787070 might help.

jerhard: Now the list of `varinfo`s of functions in force-reanalyze is also collected in a `VarinfoSet` in `CompareCIL`, in the `change_info`. So the solver does not have to query that option here again.
These force-reanalyze functions are also still collected in the `changed` field of the `change_info`, because otherwise one would have to treat them separately everywhere where the `changed` set is used (e.g. in `UpdateCIL` or the solver).

sim642 (approved):

PR #603, opened 2022-02-18
Speedup connectedness by doing it on a smaller hashtable

michael-schwarz:
For FFmpeg, the CFG computation phase did previously not terminate in a reasonable amount of time.
With these changes (use a separate smaller hashtable to check for connectedness, this seems to work). Also this drastically reduces RAM usage, which makes me slightly suspicious

TODO:
- [ ] Check this is actually correct
- [ ] Cleanup

### DIFF
--- a/src/framework/cfgTools.ml
+++ b/src/framework/cfgTools.ml
@@ -182,11 +182,11 @@ let createCFG (file: file) =
   let addEdge_fromLoc fromNode edge toNode = addEdge fromNode (Node.location fromNode, edge) toNode in
 
   (* Find real (i.e. non-empty) successor of statement.
-      CIL CFG contains some unnecessary intermediate statements.
-      If stmt is succ of parent, then optional argument parent must be passed
-      to also detect cycle ending with parent itself.
-      If not_found is true, then a stmt without succs will raise Not_found
-      instead of returning that stmt. *)
+     CIL CFG contains some unnecessary intermediate statements.
+     If stmt is succ of parent, then optional argument parent must be passed
+     to also detect cycle ending with parent itself.
+     If not_found is true, then a stmt without succs will raise Not_found
+     instead of returning that stmt. *)
   let find_real_stmt ?parent ?(not_found=false) stmt =
     if Messages.tracing then Messages.tracei "cfg" "find_real_stmt not_found=%B stmt=%d\n" not_found stmt.sid;
     let rec find visited_sids stmt =
@@ -282,7 +282,7 @@ let createCFG (file: file) =
           | Instr [] ->
             (* CIL sometimes inserts empty Instr, which is like a goto without label. *)
             (* Without this special case, CFG would contain edges without label or transfer function,
-                which is unwanted because such flow is undetectable by the analysis (especially for witness generation). *)
+               which is unwanted because such flow is undetectable by the analysis (especially for witness generation). *)
             (* Generally these are unnecessary and unwanted because find_real_stmt skips over these. *)
             (* CIL uses empty Instr self-loop for empty Loop, so a Skip self-loop must be added to not lose the loop. *)
             begin match real_succs () with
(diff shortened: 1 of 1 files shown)


michael-schwarz:
Most of things marked as different is due to wrapping things into calls to `Stats.time`. I'll comment on the relevant changes.

src/framework/cfgTools.ml
```diff
@@ -156,316 +156,326 @@ let () = Printexc.register_printer (function
...
-          failwith "MyCFG.createCFG: unsupported stmt"
+  let createCFG (file: file) =
+    let cfgF = H.create 113 in
+    let cfgB = H.create 113 in
+    if Messages.tracing then Messages.trace "cfg" "Starting to build the cfg.\n\n";
+
+    let fd_nodes = NH.create 113 in
+
+    let fun_cfgF = H.create 113 in
+    let fun_cfgB = H.create 113 in
```
michael-schwarz, line 167: Hashtables that are reset after every function, used for connectedness check.

src/framework/cfgTools.ml
```diff
@@ -156,316 +156,326 @@ let () = Printexc.register_printer (function
...
+    let addEdges fromNode edges toNode =
+      if Messages.tracing then
+        Messages.trace "cfg" "Adding edges [%a] from\n\t%a\nto\n\t%a ... "
+          pretty_edges edges
+          Node.pretty_trace fromNode
+          Node.pretty_trace toNode;
+      NH.replace fd_nodes fromNode ();
+      NH.replace fd_nodes toNode ();
+      H.add cfgB toNode (edges,fromNode); H.add fun_cfgB toNode (edges,fromNode);
+      H.add cfgF fromNode (edges,toNode); H.add fun_cfgF fromNode (edges,toNode);
```
michael-schwarz, line 178: Also add edges to per-function table

src/framework/cfgTools.ml
```diff
@@ -156,316 +156,326 @@ let () = Printexc.register_printer (function
...
+        match glob with
+        | GFun (fd, fd_loc) ->
+          Printf.printf "Looking at the function %s.\n" fd.svar.vname;
+
+          if get_bool "dbg.cilcfgdot" then
+            Cfg.printCfgFilename ("cilcfg." ^ fd.svar.vname ^ ".dot") fd;
+
+          NH.clear fd_nodes;
+          H.clear fun_cfgB;
+          H.clear fun_cfgF;
```
michael-schwarz, line 253: Reset per function tables

sim642:
But the connectedness check is only done using the current function's nodes using `fd_nodes` anyway, which follows the exact same clearing logic:
https://github.com/goblint/analyzer/blob/4c5135629f6f5456dcd2b57a4458c1dea985a2b0/src/framework/cfgTools.ml#L397

michael-schwarz:
For FFmpeg, timing goes from aborting with:

```
    makeCFG                         1.047 s
    connect                        171.697 s
```

to finishing this phase with

```
    makeCFG                         7.298 s
    connect                        55.658 s
```

sim642:
> Yes, but it still uses the large CFG hash table of all functions. We could be seeing effects where the per-function hashtable fits into a cache, and the huge one for all functions does not...

But that's far from specific to the connectedness check, Goblint has been using a single hashtable for all CFGs of all functions since the very beginning of time. And that also affects every single transfer function evaluation as well. If that's really the entirety of the bottleneck, then we have far greater problems and might have to reengineer our whole CFG representation throughout Goblint and all the downstream tools.

sim642:
I would really love if there was a profilable reproduction of the issue (~10s runtime) that already reveals the bottleneck, because that would reveal whether the true problem is cache locality (which no part of Goblint ever considers), unsuitable hashing or something else.

michael-schwarz:
Yes, that seems to make sense, but still doesn't quite explain the memory usage...

sim642:
I've spent extensive amount of time profiling the counterintuitive memory usage reduction and have finally come to a conclusion.
The reason isn't something about OCaml's memory management, but rather a blatantly simple one: this changes the amount of computations (and allocations) done by `computeSCCs`! If `Cfg.next` is based on `cfgF`, then it sometimes returns more elements than `Cfg.next` based on `fun_cfgF` _when looking up the same node_! How is this possible?

Function entry (and return) nodes are the culprit: their identity (`equal` and `hash`) is based on their underlying variable (`svar`) only. So the difference appears in the form of a single function entry node having multiple successors with different `sid`s although semantically identical content. **These conflicts appear after merging of inlines was disabled in https://github.com/goblint/cil/pull/72.** Turning that on again fixes both the time and memory issue.

sim642:
Thus, I believe that this PR alone isn't what we need. Even if we use this optimization just during the CFG construction, it does not and cannot do anything to avoid the issue in transfer functions during solving. There the same conflict would probably cause a slowdown of a similar factor (the number of unmerged copies of a function _per each unmerged copy of that function_), but to the analysis itself, not to a lightweight graph algorithm.

PR #604, opened 2022-02-19
Bump CIL to support Universal Character Names, `char16_t`, and `char32_t` (part of C99/C11)

michael-schwarz:
This bumps the version of CIL to the one including https://github.com/goblint/cil/pull/80.

This also requires some small changes, as `CStr` and `CWStr` now come with an encoding. We currently don't use the encoding anywhere.


sim642 (approved):
> We currently don't use the encoding anywhere.

Would there even be anything to do with the encoding? Or would it only matter for some string analysis that we won't have any time soon?

michael-schwarz:
I don't really see what one would do with the encoding right now, but I guess it doesn't hurt to still have it in CIL, if we ever add such string analyses or discover that there are some stupid corner cases where soundness somehow depends on it.

PR #608, opened 2022-02-22
Add more invalidate actions for library functions

michael-schwarz:
This adds some more invalidate actions for functions from the standard that are used by https://github.com/goblint/bench/issues/16.

Most warnings `Function definition missing for ...` are resolved by this.

However, it still leaves some 18 warnings `Function definition missing for ...`. In this case, these are library functions that are not part of the standard and not included in the source. Cutting down the number of such warnings produced by things from the standard also means that one can more easily spot those that are problematic

### DIFF
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -180,9 +180,9 @@ let invalidate_actions = [
     "setvbuf", writes[2];
     "fprintf", writes [1];          (*keep [1]*)
     "__fprintf_chk", writes [1];    (*keep [1]*)
-    "fread", writes [1];            (*keep [1]*)
-    "__fread_alias", writes [1];            (*keep [1]*)
-    "__fread_chk", writes [1];            (*keep [1]*)
+    "fread", writes [1;4];
+    "__fread_alias", writes [1;4];
+    "__fread_chk", writes [1;4]; 
     "utimensat", readsAll;
     "free", writesAll; (*unsafe*)
     "fwrite", readsAll;(*safe*)
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -176,7 +176,7 @@ let invalidate_actions = [
     "fclose", readsAll;           (*safe*)
     "fflush", writesAll;          (*unsafe*)
     "fopen", readsAll;            (*safe*)
-    "fdopen", writes [1];            (*safe*)
+    "fdopen", readsAll;            (*safe*)
     "setvbuf", writes[2];
     "fprintf", writes [1];          (*keep [1]*)
     "__fprintf_chk", writes [1];    (*keep [1]*)
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -167,23 +167,34 @@ let invalidate_actions = [
     "GetSpinlock", readsAll;
     "ReleaseSpinlock", readsAll;
     "atoi", readsAll;             (*safe*)
+    "__builtin_ctz", readsAll;
+    "__builtin_ctzl", readsAll;
+    "__builtin_ctzll", readsAll;
+    "__builtin_clz", readsAll;
     "bzero", writes [1]; (*keep 1*)
     "connect", readsAll;          (*safe*)
     "fclose", readsAll;           (*safe*)
     "fflush", writesAll;          (*unsafe*)
     "fopen", readsAll;            (*safe*)
+    "fdopen", writes [1];            (*safe*)
+    "setvbuf", writes[2];
     "fprintf", writes [1];          (*keep [1]*)
     "__fprintf_chk", writes [1];    (*keep [1]*)
     "fread", writes [1];            (*keep [1]*)
+    "__fread_alias", writes [1];            (*keep [1]*)
+    "__fread_chk", writes [1];            (*keep [1]*)
+    "utimensat", readsAll;
     "free", writesAll; (*unsafe*)
     "fwrite", readsAll;(*safe*)
     "getopt", writes [2];(*keep [2]*)
     "localtime", readsAll;(*safe*)
     "memcpy", writes [1];(*keep [1]*)
+    "__builtin_memcpy", writes [1];(*keep [1]*)
     "mempcpy", writes [1];(*keep [1]*)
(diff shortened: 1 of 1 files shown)


src/analyses/libraryFunctions.ml
```diff
@@ -167,23 +167,34 @@ let invalidate_actions = [
...
+    "__builtin_ctz", readsAll;
+    "__builtin_ctzl", readsAll;
+    "__builtin_ctzll", readsAll;
+    "__builtin_clz", readsAll;
     "bzero", writes [1]; (*keep 1*)
     "connect", readsAll;          (*safe*)
     "fclose", readsAll;           (*safe*)
     "fflush", writesAll;          (*unsafe*)
     "fopen", readsAll;            (*safe*)
+    "fdopen", writes [1];            (*safe*)
```
jerhard, line 179: According to the man page, the signature is `FILE *fdopen(int fd, const char *mode)`, so it should only read its arguments.

src/analyses/libraryFunctions.ml
```diff
@@ -167,23 +167,34 @@ let invalidate_actions = [
...
+    "setvbuf", writes[2];
```
jerhard, line 180: `int setvbuf(FILE *stream, char *buf, int mode, size_t size);`
According to [this](https://www.cplusplus.com/reference/cstdio/setvbuf/), this sets `buf` to be the output buffer for operation on the `FILE *stream*`. That would mean that any future write to (or `flush` of) `stream` could cause a write to the buffer.
Basically we would have to treat the `buffer` as top after such a call. Maybe add a TODO?

jerhard: I am not really aware what the type behind `FILE` is, and how it is represented in Goblint, but it may be that `stream` is actually also written?

src/analyses/libraryFunctions.ml
```diff
@@ -167,23 +167,34 @@ let invalidate_actions = [
...
     "fclose", readsAll;           (*safe*)
     "fflush", writesAll;          (*unsafe*)
     "fopen", readsAll;            (*safe*)
+    "fdopen", readsAll;            (*safe*)
+    "setvbuf", writes[2];
     "fprintf", writes [1];          (*keep [1]*)
     "__fprintf_chk", writes [1];    (*keep [1]*)
     "fread", writes [1];            (*keep [1]*)
+    "__fread_alias", writes [1];            (*keep [1]*)
+    "__fread_chk", writes [1];            (*keep [1]*)
```
jerhard, line 185: I assume that they also might modify the `FILE` pointer they are given (e.g. advancing some cursor?)

src/analyses/libraryFunctions.ml
```diff
@@ -265,10 +280,12 @@ let invalidate_actions = [
     "pthread_cond_wait", readsAll; (*safe*)
     "pthread_cond_signal", readsAll;(*safe*)
     "pthread_cond_broadcast", readsAll;(*safe*)
+    "pthread_cond_destroy", readsAll;(*safe*)
```
jerhard, line 283: That function may set the passed condition variable to an invalid value, [source](https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_cond_destroy.html).

src/analyses/libraryFunctions.ml
```diff
@@ -265,10 +280,12 @@ let invalidate_actions = [
...
     "__pthread_cond_init", readsAll; (*safe*)
     "__pthread_cond_wait", readsAll; (*safe*)
     "__pthread_cond_signal", readsAll;(*safe*)
     "__pthread_cond_broadcast", readsAll;(*safe*)
+    "__pthread_cond_destroy", readsAll;(*safe*)
```
jerhard, line 288: Same as for `pthread_cond_destroy`.

src/analyses/libraryFunctions.ml
michael-schwarz: True. I got this from our handling of `pthread_mutex_destroy`

~~~OCaml
    "pthread_mutex_destroy", readsAll;(*safe*)
 ~~~
 
I think the idea is that we use a special abstraction for mutexes anyway (distinct from the way the base analysis represents it), so even if we set this to `writesAll`, one would probably still need a separate use-after-free style analysis for mutexes, condition variables, ...

I'll make a separate issue for this.

src/analyses/libraryFunctions.ml
michael-schwarz: I changed it such that is written to, but my suspicion is that we don't really do anything terribly useful with them anyway.

src/analyses/libraryFunctions.ml
michael-schwarz: I changed it such that this is written to, but my suspicion is that we don't really do anything terribly useful with them anyway.

src/analyses/libraryFunctions.ml
michael-schwarz: I actually enhanced the issue we already have now: https://github.com/goblint/analyzer/issues/200

src/analyses/libraryFunctions.ml
```diff
@@ -378,6 +408,9 @@ let invalidate_actions = [
     "BZ2_bzBuffToBuffDecompress", writes [3;4]; (*keep [3;4]*)
     "uncompress", writes [3;4]; (*keep [3;4]*)
     "stat", writes [1]; (*keep [1]*)
+    "__xstat", writes [1]; (*keep [1]*)
+    "__lxstat", writes [1]; (*keep [1]*)
```
jerhard, line 412: signature of `__xstat`, `__lxstat`, accoding to [this](https://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/baselib-xstat-1.html): 
```
int __xstat(int ver, const char * path, struct stat * stat_buf);
int __lxstat(int ver, const char * path, struct stat * stat_buf);
```
Both functions write their third argument.

michael-schwarz: Good catch! I wrongly copied from `stat` (which was incorrect in the first place as well)

src/analyses/libraryFunctions.ml
jerhard: >my suspicion is that we don't really do anything terribly useful with them anyway.

Probably.

jerhard (approved):

PR #612, opened 2022-02-24
Call `sync` after `assign` transfer function in global initializers to make `None` privatization work properly

michael-schwarz:
Short version: Some privatizations (e.g. `None`) do not immediately side-effect globals but keep them in the local state to be side-effected upon `sync`. 
For global initializers, the call to the `sync` transfer function that gets added by the `FromSpec` functor automatically inside the solver needs to be called manually after `assign`, to ensure the values do get side-effected as expected.

**Longer version**:

The writes always happened to the local state, but the value was correctly read from the globals. Thus, e.g. for structs, the value read from the global was always `ValueDomain.bot ()` (our default for values that are not found), which lead to Goblint creating a `bot` of the appropriate type for the variable, and then updating one of its components. By this mechanism, only the last component to be updated arrived in the information that was passed to the solver, while all other components had the value of their appropriate `bot`.

Thus, in the example that I added (extracted from https://github.com/goblint/bench/issues/19), either `sqlite3Config.blarg` or `sqlite3Config.m.iValdue` would have the value bot of their type (instead of `{NULL}` and `0` respectively)

If it was `sqlite3Config.blarg` that was `{}`, this lead to both branches over the if being dead, and control flow dying out at this if and never reaching the end of `main`.

Final twist: The initializers are stored inside a hashtable that also includes the location of the declarations. Thus, a change in whitespace could change which one of these initializers happens last, and by extension whether control flow actually reaches the end of main or not in the example. (such fun to debug things like this)

### DIFF
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -275,7 +275,11 @@ struct
               (try funs := Cilfacade.find_varinfo_fundec f :: !funs with Not_found -> ())
             | _ -> ()
           );
-          Spec.assign {ctx with local = st} lval exp
+          let res = Spec.assign {ctx with local = st} lval exp in
+          (* Needed for privatizations (e.g. None) that do not side immediately *)
+          let res' = Spec.sync {ctx with local = res} `Normal in
+          if M.tracing then M.trace "global_inits" "\t\t -> state:%a\n" Spec.D.pretty res;
+          res'
         | _                       -> failwith "Unsupported global initializer edge"
       in
       let with_externs = do_extern_inits ctx file in
--- a/tests/regression/00-sanity/30-both_branches.c
+++ b/tests/regression/00-sanity/30-both_branches.c
@@ -16,6 +16,7 @@ int main(int argc , char **argv )
   int rc = 0;
 
   assert(sqlite3Config.blarg == 0);
+  assert(sqlite3Config.m.iValdue == 0);
 
   if ((unsigned long )sqlite3Config.blarg == (unsigned long )((void *(*)(int  ))0)) {
       rc = 5;
--- a/tests/regression/00-sanity/30-both_branches.c
+++ b/tests/regression/00-sanity/30-both_branches.c
@@ -15,6 +15,8 @@ int main(int argc , char **argv )
 {
   int rc = 0;
 
+  assert(sqlite3Config.blarg == 0);
+  
   if ((unsigned long )sqlite3Config.blarg == (unsigned long )((void *(*)(int  ))0)) {
       rc = 5;
   }
--- a/tests/regression/00-sanity/30-both_branches.c
+++ b/tests/regression/00-sanity/30-both_branches.c
@@ -16,7 +16,7 @@ int main(int argc , char **argv )
   int rc = 0;
 
   assert(sqlite3Config.blarg == 0);
-  
+
   if ((unsigned long )sqlite3Config.blarg == (unsigned long )((void *(*)(int  ))0)) {
       rc = 5;
   }
--- a/tests/regression/00-sanity/30-both_branches.c
+++ b/tests/regression/00-sanity/30-both_branches.c
@@ -1,4 +1,4 @@
-// PARAM: --set ana.base.privatization none --enable exp.earlyglobs
+// PARAM: --set ana.base.privatization none --enable exp.earlyglobs --disable exp.fast_global_inits
 union bloirg {
    int iValdue ;
 };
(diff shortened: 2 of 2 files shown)


michael-schwarz:
With this change, analysis of sqlite3 does no longer terminate instantly.

src/framework/control.ml
```diff
@@ -275,7 +275,11 @@ struct
               (try funs := Cilfacade.find_varinfo_fundec f :: !funs with Not_found -> ())
             | _ -> ()
           );
-          Spec.assign {ctx with local = st} lval exp
+          let res = Spec.assign {ctx with local = st} lval exp in
+          (* Needed for privatizations (e.g. None) that do not side immediately *)
+          let res' = Spec.sync {ctx with local = res} `Normal in
+          if M.tracing then M.trace "global_inits" "\t\t -> state:%a\n" Spec.D.pretty res;
+          res'
```
sim642, line 282: Wouldn't it be enough to only `sync` once after all the `assign`s, instead of after each `assign`. With a large number of global initializers, this could be wasteful. Also each `sync` in those privatizations iterates over the entire state again. Doing that once should be more optimal.

sim642 (commented):
I'm surprised that this is necessary, because if I remember correctly, before side effects were possible during global inits, everything went into the local state and there wasn't any extra `sync`. Rather the first `sync` for `enter` of `main` was the one to publish everything. Why doesn't that work any more?

src/framework/control.ml
michael-schwarz: No, that would not suffice. With none privatization, the analysis looks up the value via `getg`, so `sync` has to happen right away, for the correct result to be returned (in case two initializers refer to to the same global, e.g., a struct)

michael-schwarz:
> everything went into the local state and there wasn't any extra sync.

Yes, that used to be waht was done. But since we allowed side effects, we have removed all sorts of special handling and weird hacks for global initialization in the privatization, and we should probably leave it like this.

src/framework/control.ml
michael-schwarz: Also, the state should be very small, so iterating over it will not be too expensive I think. One could think about only doing this for `earlyglobs`, but even if we do this, it sounds a bit cumbersome.

michael-schwarz:
After this change the issue with not enough code being live in ffmpeg and sqlite seem to have resolved, both analyses are now running for several hours.

sim642:
> By this mechanism, only the last component to be updated arrived in the information that was passed to the solver, while all other components had the value of their appropriate `bot`.

Why is that? The simulated `sideg` should be joining those structs: https://github.com/goblint/analyzer/blob/36a5c5d20e75c06c979a46996c202ed7384ec0ef/src/framework/control.ml#L226-L236
Is the join of structs not commutative or how can it depend on any order?

michael-schwarz:
No calls to `sideg` happened at all, because they only happen in `sync` for `None` privatization.

sim642:
Isn't the problem simply then that `NonePriv.read_global` should also be looking at the local state?
https://github.com/goblint/analyzer/blob/36a5c5d20e75c06c979a46996c202ed7384ec0ef/src/analyses/basePriv.ml#L90-L91
Because it seems to me that the inconsistency is simply there: it writes to local state but reads from global state.

That's different from `ProtectionBasedOldPriv`, which was perfectly fine with the only `sync` being the one from `body` of `main`. Before traces, how global structs were initialized worked perfectly fine and `getg` also always returned `bot` (and `sideg` simply raised an exception, so the `sync`s even couldn't be there). Seems like the only reason it wasn't a problem was because it looks up the local state first.

And the fact that `NonePriv` doesn't is likely just my mistake when I refactored everything and nobody ever really tested it in the meanwhile.

I don't necessarily have a problem with having the `sync Normal`s during global inits, but I just really wanted to understand how they were never necessary before.

michael-schwarz:
> Isn't the problem simply then that `NonePriv.read_global` should also be looking at the local state?

Yes, that is an alternative fix that one can make, and it was initially also the one I tried.

Still, this would mean that the `None` privatization works differently during initialization than normal analysis, where the sync call happens after each transfer function, and the values are hence side-effected to the global (as one would expect with `None` privatization and `earlyglobs`).

So I'd argue the approach from this PR is in some sense the more principled and less fragile one, and we call `sync` everywhere during the analysis anyway, so I don't think it will make too much of a difference performance-wise overall.

sim642:
> Still, this would mean that the `None` privatization works differently during initialization than normal analysis, where the sync call happens after each transfer function, and the values are hence side-effected to the global (as one would expect with `None` privatization and `earlyglobs`).

I now started worrying that this might break with `NonePriv` and no `earlyglobs` during single-threaded mode after global inits, but maybe not. There's just annoyingly many combinations to consider with these things.

sim642 (approved):

PR #615, opened 2022-02-28
Do not allow to collect globals with duplicate identifiers in map

stilscher:
The general comparison for the incremental analysis (used for both, the cfg and ast comparison) assumes that globals have a unique identifier. But it is not warned if this is not the case. Instead, the incremental analysis run will often fail later on in `UpdateCil.ml` because `List.iter2` expects lists of the same length. This happened for example in https://github.com/goblint/bench/issues/16.
This PR only inserts a warning when globals with duplicate identifiers are found to make it easier for now to detect this problem. One could of course also think further about how to make the comparison robust against globals with the same name.

### DIFF
--- a/src/incremental/compareCIL.ml
+++ b/src/incremental/compareCIL.ml
@@ -69,7 +69,13 @@ let compareCilFiles ?(eq=eq_glob) (oldAST: file) (newAST: file) =
 
   let addGlobal map global  =
     try
-      GlobalMap.add (identifier_of_global global) global map
+      let gid = identifier_of_global global in
+      let gid_to_string gid = match gid.global_t with
+        | Var -> "Var " ^ gid.name
+        | Decl -> "Decl " ^ gid.name
+        | Fun -> "Fun " ^ gid.name
+        | _ -> raise (NoGlobalIdentifier global) in
+      if GlobalMap.mem gid map then failwith ("Duplicate global identifier: " ^ gid_to_string gid) else GlobalMap.add gid global map
     with
       NoGlobalIdentifier _ -> map
   in


src/incremental/compareCIL.ml
```diff
@@ -69,7 +69,13 @@ let compareCilFiles ?(eq=eq_glob) (oldAST: file) (newAST: file) =
...
   let addGlobal map global  =
     try
-      GlobalMap.add (identifier_of_global global) global map
+      let gid = identifier_of_global global in
+      let gid_to_string gid = match gid.global_t with
+        | Var -> "Var " ^ gid.name
+        | Decl -> "Decl " ^ gid.name
+        | Fun -> "Fun " ^ gid.name
+        | _ -> raise (NoGlobalIdentifier global) in
+      if GlobalMap.mem gid map then failwith ("Duplicate global identifier: " ^ gid_to_string gid) else GlobalMap.add gid global map
```
sim642, line 78: Actually with OCaml hashtables one should almost always use `replace` instead of `add` since the shadowing behavior is surprising, memory leaky and inefficient (for even representing multimaps, see #614). Then it wouldn't possible to store two globals for the same name.
Of course that wouldn't solve anything here, it would just override the old one.

stilscher: The `GlobalMap` is created with OCamls `Map.Make` functor and therefore does not have a replace function as a `Hashtbl` would have. The documentation indicates that `add` has an overwriting behavior if a key already existed. Should we switch to a `Hashtbl`?

sim642: Ahh, nevermind then. I thought it's a hashtable like other mutated data structures around these parts. As long as it's not a bottleneck, it should be fine.

sim642 (approved):

Issue #107, opened 2020-10-02
Cleaning up and documenting region analysis


vesalvojdani:
Simmo fixed some soundness issues (#106) that arise when the region analysis is executed with symbolic equality and field-sensitive refinement for all benchmarks. Previously, these flags were only turned on for specific benchmarks. The pull request fixed all surface issues that emerged, but it did not raise our confidence in the code. 

- [ ] Figure out if we really need general `exp` rather than `lval` to be stored for accesses.
- [ ] The field offset handling is fairly suspect and should be documented. I am not 100% convinced adding field-sensitivity to the region analysis in this way is sound, even for non-overlapping fields.
- [ ] Eliminate the todos for the offset handling. Why are some offsets dropped and some are handled.
- [ ] Build the partition domain on top of the Powerset domain. 

The region analysis is very simple on paper, but we have not worked out these interactions with field-sensitivity. We may need a deeper understanding of these things. We take a pointer `p` into an equivalence class and upon `p -> f` just appending `f` to all members of that class. This is sound under certain specific assumptions, but it's not clear that these assumptions are maintained by all operations of the region analysis.

michael-schwarz:
While doing this, one should probably also look at #66 again.

sim642:
Just to document this here: yesterday we discovered region analysis unsoundness in `ldv-linux-3.14-races/linux-3.14--drivers--usb--misc--iowarrior.ko.cil.i` as well, where we miss a race on the `opened` field of `iowarrior` struct, because it appears to have `no region` (bullet) due to an unknown pointer.

Although the benchmark itself is also questionable, since in one case `dev->opened` is accessed on a `malloc`ed but uninitialized struct, which is UB and shouldn't be in sv-benchmarks to begin with.

michael-schwarz:
As of #641, another example for an unsoundness is `09/38`.

sim642:
When @karoliineh extracted a simplified non-escape analysis from region, a fixpoint error arose from a ZSTD test we have. I managed to reproduce an analogous fixpoint error with the original region analysis here: https://github.com/goblint/analyzer/commit/16077e3b146d464072004eaf0cd1480a2a7b99d1.

[...]

PR #618, opened 2022-03-01
ldv-benchmarks races benchmarking

sim642:
Goblint changes for benchmarking https://github.com/goblint/bench/issues/3.

### Changes
1. Fix `ValueDomain.is_safe_cast` to correctly handle casts between signed and unsigned. This caused unsoundness in some `invariant` (test added).

### Problems
- [ ] Some benchmarks would require #54, because CIL inserts builtin function signatures (based on `Machdep`) which are incompatible with those preprocessed into ldv-benchmarks:
    ```
    old type = TFun(TInt(short, ), : TInt(short, ), )
    new type = TFun(TInt(unsigned short, ), : TInt(unsigned short, ), )
    /home/debian/klever-inst/klever-work/native-scheduler/scheduler/jobs/b8b212dd331653bfe7f8cbf43b3041a7/klever-core-work-dir/job/vtg/drivers/net/wireless/intersil/orinoco/orinoco_usb.ko/linux:concurrency safety/emg/environment_model.c:19: Error: Declaration of __builtin_bswap16 does not match previous declaration from /mnt/goblint-svcomp/sv-comp/goblint/includes/sv-comp.c:35 (different integer types).
    ```
    Our `sv-comp.c` stub doesn't explicitly define or include any `__builtin_bswap16`, so it must be coming from CIL's builtins.
- [x] `ldv-commits-races/tasks/f1a8a3f-1.cil.i` unsound. Uses custom `pthread_create_N`.
- [x] `ldv-commits-races/tasks/4036523-1.i` unsound. Uses custom `pthread_create_N`.
- [x] `linux-4.2.6-races/Unsafes/u__linux-concurrency_safety__drivers---net---irda---ksdazzle-sir.ko.cil.i` unsound. Model appears to be wrong for `usb_endpoint_dir_in` (bitwise op replaced with tautological unsigned comparison).
- [x] `linux-4.18-races/Unsafes/u__linux-concurrency_safety__drivers---net---ethernet---8390---ax88796.ko.cil.i` unsound. Model appears to be wrong for `ax_reset_8390` (bitwise op replaced with tautological unsigned comparison).


sim642:
Besides the region analysis causing unsoundness (#107), which doesn't show in the verdicts, it's also completely unsuitable for interactive benchmarking since the region partitioning is stored under a single global unknown and it's not clear whether this could be any different.

sim642:
The unsoundness fix contained here is actually relevant on sv-benchmarks as well. From the same kernel module there's a benchmark `ldv-challenges/linux-3.14_linux-kernel-locking-spinlock_drivers-net-ethernet-amd-pcnet32.cil`, which without intervals has 60s timeout, but with intervals gives immediate `true` result (which just happens to be correct).

sim642:
Since this is small and contains a possibly important unsoundness fix, I'll just merge it.

PR #632, opened 2022-03-09
Add splitting of Compilation Database entries with multiple .c files

### LINKED ISSUE #624: Compilation Database: Support for entries with multiple `.c` files in one command
@stilscher discovered this issue with older versions of `zstd` (e.g. `5c5c47633826426a3fcc717000a352d52bd0ac22`).
The complication database in this case contains entries where the command specifies several `.c` files at the same time.

~~~JSON
 {
  "directory": "/home/michael/Documents/bench-repos/zstd/programs",
  "arguments": [
   "cc",
   "-I../lib",
   "-I../lib/common",
   "-I../lib/compress",
   "-I../lib/dictBuilder",
   "-DXXH_NAMESPACE=ZSTD_",
   "-I../lib/legacy",
   "-DZSTD_MULTITHREAD",
      // more defines
   "-DZSTD_NO_INTRINSICS",
   "-O3",
   "-Wall",
   "-Wextra",
       // more warnings
   "-pthread",
   "-lz",
   "-llzma",
   "../lib/common/debug.c",
   "../lib/common/entropy_common.c",
       // more files
   "../lib/dictBuilder/zdict.c",
   "zstdcli.o",
   "util.o",
   "fileio.o",
   "bench.o",
   "datagen.o",
   "dibio.o",
   "-o",
   "zstd",
   "-pthread",
   "-lz",
   "-llzma"
  ],
  "file": "../lib/dictBuilder/zdict.c"
 }
~~~

Goblint in this case adds an `-E` flag and `-o` to get preprocessor output. The preprocessor does not like this, and fails with

```
cc: fatal error: cannot specify -o with -c, -S or -E with multiple files
compilation terminated.
```

sim642:
Closes #624.

Since all this string manipulation of compiler commands could be fragile, this feature is off by default and must be enabled via the `pre.compdb.split` option.

I tried it on my tiny project with manually written Makefile, where this is needed, and it works there, but I didn't try it on that zstd commit.


sim642:
I tested the preprocessing on the zstd commit mentioned in #624 and with `--enable pre.compdb.split` it doesn't crash. Also since this is unit tested and disabled by default, I'll just merge it now.

PR #633, opened 2022-03-09
Hack incremental accesses to only restart during postsolving

sim642:
This is on top of #391.

I'm _not_ proud to present this hack to restart incremental accesses only during postsolving. This (hopefully) works for the same reason incremental warnings work, but instead of a new generic interface this is the crudest possible way to achieve the same. **No guarantees!**

### DIFF
--- a/src/solvers/postSolver.ml
+++ b/src/solvers/postSolver.ml
@@ -72,8 +72,11 @@ module Verify: F =
   struct
     include Unit (S) (VH)
 
+    let accs = VH.create 123
+
     let init () =
-      Goblintutil.verified := Some true
+      Goblintutil.verified := Some true;
+      VH.clear accs
 
     let complain_constraint x ~lhs ~rhs =
       Goblintutil.verified := Some false;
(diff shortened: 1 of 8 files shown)


vesalvojdani:
Yeah, but this is the morally right approach to write-only information. We should think of the cleaner interface for accumulating stuff, like warnings and accesses, but this is still a close approximation to it. This is at least the right place where resets should happen for the accesses.

src/solvers/postSolver.ml
```diff
@@ -84,8 +87,18 @@ module Verify: F =
...
+      let is_acc = String.starts_with (S.Var.var_id y) "access:" in
+      let y_lhs =
+        if is_acc && not (VH.mem accs y) then (
+          VH.replace accs y ();
+          S.Dom.bot ()
+        )
+        else
+          try VH.find vh y with Not_found -> S.Dom.bot ()
+      in
+      if not is_acc && not (S.Dom.leq d y_lhs) then
```
michael-schwarz, line 101: Would we not still need some special logic to account for side-effects from things that are superstable?

sim642: Possibly. When I later thought about it, I was confused about that too, but it somehow works on all the incremental tests we have.

Maybe this is completely broken after all...

sim642:
Superseded by #634.

PR #638, opened 2022-03-10
Ignore exceptions from `Cilfacade.typeOf` when computing an equivalent array expression

michael-schwarz:
We have long had some issues with strange offsets appearing in addresses. This one here leads to Goblint crashing on zstd if one adds more malloc wrappers to potentially make race detection more precise.

The place where it fails is when trying to compute an equivalent expression for an array access that would then be used by the partitioned array. Here it is always sound to put the top expression, so I suggest to ignore errors here.

### DIFF
--- a/src/cdomains/valueDomain.ml
+++ b/src/cdomains/valueDomain.ml
@@ -767,24 +767,22 @@ struct
           | Some (v') ->
             begin
               try
-                begin
-                  (* This should mean the entire expression we have here is a pointer into the array *)
-                  if Cil.isArrayType (Cilfacade.typeOfLval v') then
-                    let expr = ptr in
-                    let start_of_array = StartOf v' in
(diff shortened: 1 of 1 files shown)


src/cdomains/valueDomain.ml
```diff
@@ -766,22 +766,26 @@ struct
...
-                if start_type = expr_type then
-                  `Lifted (equiv_expr expr v')
-                else
-                  (* If types do not agree here, this means that we were looking at pointers that *)
-                  (* contain more than one array access. Those are not supported. *)
-                  ExpDomain.top ()
-              else
-                ExpDomain.top ()
+              try
+                begin
```
sim642, line 770: This inner `begin...end` is redundant here.

sim642 (approved):

Issue #639, opened 2022-03-11
Challenging Example for Race Detection from zstd


michael-schwarz:
I extracted the following code from `zstd` that creates a thread pool and manages the threads contained in the queue. It seems to be representative of the race warnings we produce on the entire zstd program.

Most (all?) accesses to the members of `ctx` are protected by `ctx.queueMutex`, but we still flag races despite this (10 with default configuration, 6 with region analysis).

I think it might be good to try to investigate what is the root cause here, to be able to reduce the false warnings for more real-world scenarios.

~~~C
/* SPDX-License-Identifier: BSD-3-Clause */
/*
 * Copyright (c) Facebook, Inc.
 * All rights reserved.
 *
 * This code is a challenging example for race detection extracted from zstd.
 * Copyright (c) The Goblint Contributors
 */

#include<stdlib.h>
#include<pthread.h>
#define ZSTD_pthread_mutex_t            pthread_mutex_t
#define ZSTD_pthread_mutex_init(a, b)   pthread_mutex_init((a), (b))
#define ZSTD_pthread_mutex_destroy(a)   pthread_mutex_destroy((a))
#define ZSTD_pthread_mutex_lock(a)      pthread_mutex_lock((a))
#define ZSTD_pthread_mutex_unlock(a)    pthread_mutex_unlock((a))

#define ZSTD_pthread_cond_t             pthread_cond_t
#define ZSTD_pthread_cond_init(a, b)    pthread_cond_init((a), (b))
#define ZSTD_pthread_cond_destroy(a)    pthread_cond_destroy((a))
#define ZSTD_pthread_cond_wait(a, b)    pthread_cond_wait((a), (b))
#define ZSTD_pthread_cond_signal(a)     pthread_cond_signal((a))
#define ZSTD_pthread_cond_broadcast(a)  pthread_cond_broadcast((a))

#define ZSTD_pthread_t                  pthread_t
#define ZSTD_pthread_create(a, b, c, d) pthread_create((a), (b), (c), (d))
#define ZSTD_pthread_join(a, b)         pthread_join((a),(b))

#define ZSTD_malloc(s) malloc(s)
#define ZSTD_calloc(n,s) calloc((n), (s))
#define ZSTD_free(p) free((p))
#define ZSTD_memset(d,s,n) __builtin_memset((d),(s),(n))

typedef struct POOL_ctx_s POOL_ctx;

typedef void* (*ZSTD_allocFunction) (void* opaque, size_t size);
[...]

PR #641, opened 2022-03-12
Add zstd thread pool example to regression tests

michael-schwarz:
This adds the example we discovered in #639 to the regression tests. It is a challenging and natural example, we might want to keep it on hand.

### DIFF
--- a/tests/regression/09-regions/38-pool.c
+++ b/tests/regression/09-regions/38-pool.c
@@ -1,0 +1,263 @@
+// PARAM: --set ana.activated[+] region --set ana.activated[+] symb_locks
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) Facebook, Inc.
+ * All rights reserved.
+ *
+ * This code is a challenging example for race detection extracted from zstd.
+ * Copyright (c) The Goblint Contributors
+ */
+
+#include<stdlib.h>
+#include<pthread.h>
+#define ZSTD_pthread_mutex_t            pthread_mutex_t
+#define ZSTD_pthread_mutex_init(a, b)   pthread_mutex_init((a), (b))
+#define ZSTD_pthread_mutex_destroy(a)   pthread_mutex_destroy((a))
+#define ZSTD_pthread_mutex_lock(a)      pthread_mutex_lock((a))
+#define ZSTD_pthread_mutex_unlock(a)    pthread_mutex_unlock((a))
+
+#define ZSTD_pthread_cond_t             pthread_cond_t
+#define ZSTD_pthread_cond_init(a, b)    pthread_cond_init((a), (b))
+#define ZSTD_pthread_cond_destroy(a)    pthread_cond_destroy((a))
+#define ZSTD_pthread_cond_wait(a, b)    pthread_cond_wait((a), (b))
+#define ZSTD_pthread_cond_signal(a)     pthread_cond_signal((a))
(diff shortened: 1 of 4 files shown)


tests/regression/09-regions/38-pool.c
```diff
@@ -0,0 +1,263 @@
+// PARAM: --set ana.activated[+] region --set ana.activated[+] symb_locks
```
sim642, line 1: Was it regions or symb_locks that was really required for this? Because currently this example is in the regions directory.

tests/regression/09-regions/38-pool.c
```diff
@@ -0,0 +1,263 @@
...
+#define ZSTD_pthread_cond_broadcast(a)  pthread_cond_broadcast((a))
+
+#define ZSTD_pthread_t                  pthread_t
+#define ZSTD_pthread_create(a, b, c, d) pthread_create((a), (b), (c), (d))
+#define ZSTD_pthread_join(a, b)         pthread_join((a),(b))
+
+#define ZSTD_malloc(s) malloc(s)
+#define ZSTD_calloc(n,s) calloc((n), (s))
+#define ZSTD_free(p) free((p))
+#define ZSTD_memset(d,s,n) __builtin_memset((d),(s),(n))
```
sim642, line 33: Would it hurt to inline these? They're one-to-one anyway.

sim642 (commented):
I agree it's a good idea to have an extracted example from the real world that we can analyze around.

tests/regression/09-regions/38-pool.c
michael-schwarz: both to get to zero warnings

tests/regression/09-regions/38-pool.c
michael-schwarz: I guess not, but I did not want to fiddle around with the actual code too much to stay as close to the source one as possible.

tests/regression/09-regions/38-pool.c
sim642: With symb_locks activated and regions deactivated, there's a race on the `threadLimit` field:
```
[Warning][Race] Memory location (struct POOL_ctx_s).threadLimit (race with conf. 100):
  read with [symblock:{p-lock:*.queueMutex}, mhp:{tid=[main], {POOL_thread@tests/regression/09-regions/38-pool.c:221:13-225:9}}] (conf. 100)  (exp: & ctx->threadLimit) (tests/regression/09-regions/38-pool.c:149:17-150:58)
  read with [symblock:{p-lock:*.queueMutex}, mhp:{tid=[main, POOL_thread@tests/regression/09-regions/38-pool.c:221:13-225:9]}, thread:[main, POOL_thread@tests/regression/09-regions/38-pool.c:221:13-225:9]] (conf. 100)  (exp: & ctx->threadLimit) (tests/regression/09-regions/38-pool.c:149:17-150:58)
  write with [mhp:{tid=[main]; created={[main, POOL_thread@tests/regression/09-regions/38-pool.c:221:13-225:9]}}, thread:[main]] (conf. 100)  (exp: & ctx->threadLimit) (tests/regression/09-regions/38-pool.c:227:9-227:38)
```
And I think that's actually a valid race: the threads read it with the mutex, but the write is completely unlocked, no?

Therefore I think this is another example where our region analysis unsoundly assigns `no region` (bullet) to something which actually can be shared (by passing the pointer to the thread via its create argument).

sim642:
Since region is unsound on this test, I deactivated that (and moved the test to a non-region group).

PR #642, opened 2022-03-16
Refactor path handling using Fpath library

sim642:
String-based path manipulations are quite fragile and the standard `Filename` module doesn't provide all the necessary operations. That is especially true for the more involved manipulations for compilation databases that use some custom `GobFilename` functions. For example, there isn't even a way to normalize paths (i.e. eliminate and cancel out `.` and `..`).

Instead of reinventing the wheel, we should just use a high-quality library like [`Fpath`](https://erratique.ch/software/fpath/doc/Fpath/index.html) that provides all these operations.

Among other things, this should be an alternative fix to #674 which doesn't require OCaml 4.13.

### TODO
- [x] Compilation database
- [x] Preprocessing
- [x] `GoblintDir`
- [x] Makefile support
- [x] `Goblintutil`
- [x] `GobSys`
- [x] `save_run`
- [x] `load_run`
- [x] `Serialize`


src/framework/control.ml
```diff
@@ -470,15 +470,16 @@ struct
...
             Serialize.store_data solver_data Serialize.SolverData;
           if save_run <> "" then (
-            let analyses = Filename.concat save_run "analyses.marshalled" in
-            let config = Filename.concat save_run "config.json" in
-            let meta = Filename.concat save_run "meta.json" in
-            let solver_stats = Filename.concat save_run "solver_stats.csv" in (* see Generic.SolverStats... *)
-            let cil = Filename.concat save_run "cil.marshalled" in
-            let warnings = Filename.concat save_run "warnings.marshalled" in
-            let stats = Filename.concat save_run "stats.marshalled" in
+            let save_run = Fpath.v save_run in
```
michael-schwarz, line 473: Maybe not name it `save_run` here, it is a bit confusing as the string `save_run` is used again later, which one might get confused if one doesn't see this let is inside an if.

michael-schwarz (approved):

sim642:
I tried to check whether this also fixes #674, but now checking out a Goblint version before that still works for me, so I'm confused about how that problem even could be reproduced.

sim642:
> I tried to check whether this also fixes #674, but now checking out a Goblint version before that still works for me, so I'm confused about how that problem even could be reproduced.

AFAIK this should keep those fixes but just replace them with Fpath equivalents, so hopefully no regression from this refactoring. If there's a problem again, feel free to complain.

PR #645, opened 2022-03-16
Replace custom `assert` declaration with macro

sim642:
The C11 standard says:

> The `assert` macro shall be implemented as a macro, not as an actual function. If the macro definition is suppressed in order to access an actual function, the behavior is undefined.

Therefore this PR changes our custom `assert.h` to also define `assert` as a macro to the special undefined function `__goblint_assert` to be standards-compliant.

This fixes a parsing error I saw in zstd, which under some conditions defines `#define assert(expression) ((void)0)`. If our `assert.h` is included _after_ that definition, then the declaration `void assert(int expression);` was turned into `void ((void)0);`, which is syntactically invalid.

Moreover, this immediately revealed over 100 regression tests that used `assert` without including `assert.h`. They relied on implicit function declaration of `assert`, which undefined behavior according to the above quote, since it accesses `assert` as a function, not a macro definition (which is only supplied by `assert.h`, either standard or ours).


sim642:
It's still relevant, but probably got sidelined because `GOBLINT_NO_ASSERT` was used instead for zstd benchmarking to avoid the described problem.
I'll have to bring this branch up to date.

sim642:
I brought it up to date, but somehow one assertion in one Apron test is now unknown due to the name change, which is very weird.

sim642:
> somehow one assertion in one Apron test is now unknown due to the name change, which is very weird.

Apparently there was a hardcoded `assert` function name in `WideningThresholds`, which was at fault. After breaking a very intricate dependency cycle, I ported it to not hardcode the function name, so it should all work now.

michael-schwarz:
Still having this line is confusing now:

https://github.com/goblint/analyzer/blob/2a2c294a8c8b370d20fcf23f00ffc820be8e47ab/src/analyses/base.ml#L2472

The same goes for the `__goblint_assert` in the apron analysis.

michael-schwarz:
And while we're at it, should we maybe not replace `__goblint_assert(...)` with `__goblint_check(...)` in the headers we use for our regression tests?

michael-schwarz:
And maybe not use `dbg.debug` with such an innocuous description as `Debug mode: for testing the analyzer itself` to change the semantic we assign to `assert(...)` statements.
Maybe we ought to have `sem.assert.check` and `sem.assert.assume` or something similar?

sim642:
> And while we're at it, should we maybe not replace `__goblint_assert(...)` with `__goblint_check(...)` in the headers we use for our regression tests?

Why? It's still an assertion.

> And maybe not use `dbg.debug` with such an innocuous description as `Debug mode: for testing the analyzer itself` to change the semantic we assign to `assert(...)` statements.
> Maybe we ought to have `sem.assert.check` and `sem.assert.assume` or something similar?

Once again, I have to point out that the entire idea of #278 was to clear this situation up once and for all, but nobody seemed enthusiastic about it. It's bothered me since the beginning of time how the semantics of `assert` changes under debugging, which makes debugging very confusing if some problem then disappears/appears due to the semantics change.

It definitely doesn't make sense to redo that work now here. If we agree that our semantics of `assert` is the same as defined by the standard, then we can go through the trouble of bringing #278 up to date and also stop writing tests that use `assert` when they don't actually want to refine.

michael-schwarz:
My issue with #278 is that it makes writing regression tests super inconvenient and also means the semantics of all regression tests that we already have changes and none of the regression tests still test what they should test.

If we agree to
-  keep a switch to not have a assert behave as the standard for the regression test and keep using that there or
- replace all asserts with `__goblint_check(...)` 

I'd be down with #278.

Using `assert(...)` with the semantics defined by the standard in regression tests is just painful and cumbersome and thus why I immensely disliked it at the time.

sim642:
> * keep a switch to not have a assert behave as the standard for the regression test and keep using that there

But that's exactly what #278 does: it introduces `sem.assert.refine` and disables it in all the tests that (at the time) implicitly assumed non-refinement for their testing.

> Using `assert(...)` with the semantics defined by the standard in regression tests is just painful and cumbersome and thus why I immensely disliked it at the time.

It only makes things difficult/inconvenient if `__goblint_check` is meant but not used in the first place. It has existed in Goblint before any of us started working on it.

michael-schwarz:
> But that's exactly what #278 does:

If memory serves, it did this in a hap-hazard way only where required instead of everywhere to preserve the intended semantics everywhere. If we move to have assert behave according to the standard, I'd suggest we replace all `assert`s with `__goblint_check` in the current regression tests.

michael-schwarz:
But I don't think it needs to be done in this PR. If we fix the confusing occurence of `__goblint_check` in the two analyses, I think we can merge this.

sim642:
> If memory serves, it did this in a hap-hazard way only where required instead of everywhere to preserve the intended semantics everywhere.

That "hap-hazard" way was a complete review of all the assertions in our regression suite to determine, which use and which don't use the implicit non-standard non-refinement assumption.

> If we move to have assert behave according to the standard, I'd suggest we replace all `assert`s with `__goblint_check` in the current regression tests.

And I would be quite against doing that completely and blindly:
1. It will make it effectively impossible to submit our regression tests, because none of them would use a standard construct. And the only way to submit tasks would be to manually look at each `__goblint_check` every year to determine whether it's also valid as a standard assertion and then manually convert that into an SV-COMP task. Whereas as if we clearly use `assert` where possible and `__goblint_check` where necessary, we clearly distinguish them in our suite and can trivially and automatically convert the tasks. At least I'm not going to manually review hundreds of tasks (and growing) for sv-benchmarks submission.
2. It will also make it impossible to use our regression tests as reusable benchmarks like we did with some Apron privatization tests to run Duet on them (which doesn't know anything about `__goblint_check`) and propose them as a general benchmark set into the literature. Again one would have to manually review them after the fact each time instead of maintaining a proper classification continuously in our repository.

michael-schwarz (approved):

PR #648, opened 2022-03-18
Add messagesCompare for comparing json-messages results

### LINKED ISSUE #607: Allow comparison of warnings produced by different analyzer runs
For experiments, it should be possible to compare warnings produced by different analyzer runs, e.g. between incremental and non-incremental runs. Implementing this comparison on the OCaml side with Goblint-datastructures, similar to `privPrecCompare` is probably more reliable than doing a text-based diff between warnings produced by different runs.

The comparison should in particular yield the number of warnings that where produced in both runs, in run A but not run B and vice-versa.

sim642:
Closes #607.

Given two `json-messages` outputs, the `messagesCompare` executable will compare them as sets (ignoring order, unlike a straight `diff`). and outputs the messages that are only in one or the other using the usual pretty style.

This should make it easier to compare user-visible precision via messages.

---

I initially wanted to just throw together a quick Python script, but Python's `json.load` is so stupid that the loaded data is represented using `dict`s and `list`s, which are unhashable and cannot just be thrown into a set...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ sv-comp/goblint.zip
 
 privPrecCompare*
 apronPrecCompare
+messagesCompare
 tests/regression/*/run
 
 # csmith
--- a/make.sh
+++ b/make.sh
@@ -46,6 +46,11 @@ rule() {
       dune build src/apronPrecCompare.exe &&
       rm -f apronPrecCompare &&
       cp _build/default/src/apronPrecCompare.exe apronPrecCompare
+    ;; messagesCompare)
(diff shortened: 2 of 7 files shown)


sim642:
Also by the way, our Makefile stuff is stupid: if you `make messagesCompare` and change something, then `make messagesCompare` will refuse to do anything, because make's super-heuristic change detection kicks in and thinks nothing changed. Only when you manually remove the `messagesCompare` binary, then it will work again and correctly run `dune` to rebuild...

michael-schwarz (approved):

Issue #650, opened 2022-03-21
Revamp Deadlock-Detection to work incrementally and consider MHP information


michael-schwarz:
Another issue that we discussed at Gobcon some weeks ago, adding it here so we don't lose track.

Things that would need to be done:
- [x] Incrementalize the collection of lock-order constraints (the approach from #634 by @sim642 should work here)
- [x] Replace `Cil.location` with `Node.t`
- [ ] ~~Maybe: Incrementalize the building of the locking graph (?)~~
- [x] Make use of MHP information to exclude some cycles. (Naively, this would require collecting lock/unlock pairs for mutexes to exclude cycles where the cycle in the locking order is excluded by MHP information, e.g.
    -  T1 locks a and then b,
    -  T2 locks b and then a
    - But T2 is not started yet when T1 does the locking)
    
This would then be a second class of concurrency bug that we can find with Goblint.

sim642:
A very quick Google search yielded this article: https://arxiv.org/abs/1607.06927. They point out of the lack of sound static analysis tools for deadlock analysis, so it might be good to have a maintained deadlock analysis indeed. Also they claim to have evaluated theirs on an extensive set of pthread programs, which would be of great interest to our benchmarking. Moreover, they have something called "non-concurrency analysis" which sounds a lot like our MHP, so another thing to look at there.

sim642:
> * Maybe: Incrementalize the building of the locking graph (?)

I'm not sure what you mean by this. In #655 the locking order graph is implicitly constructed at the end, based on the information in the global invariant. This is the same as how we construct the race graph.
Both are constructed from scratch each time, based on the incremental global invariant. I'm not sure there is any other way, because somehow storing more of the graph in another derived global invariant would also require somehow subtracting/restarting parts of it, which makes it more involved than just set removal. Also since the locking order graph involves transitivity, removing one edge might require the removal of many derived edges, which isn't really possible without having the non-transitive graph around.


> * Make use of MHP information to exclude some cycles. (Naively, this would require collecting lock/unlock pairs for mutexes to exclude cycles where the cycle in the locking order is excluded by MHP information [...])

I'm thinking we should somehow be able to reuse most of our existing `access` implementations for `may_race`. Most of them (like MHP, but also mutex (must-lockset)) are just about concurrency and independent of the variable being accessed. Just like we record `MCPAccess.A` for each access, we could record it for each locking event. Then later when traversing the graph, the same information can be checked pairwise like `may_race`.

[...]

Issue #176, opened 2021-04-06
Mutex sanity analysis


sim642:
While debugging some benchmark programs, I noticed they themselves contain silly bugs:
1. Forgetting to unlock a mutex (e.g. at early return), which makes any following call of the function deadlock or undefined behavior: https://github.com/goblint/bench/blob/16ee6e2dfa54881be2161cb5bb6eee8cce5736d7/pthread/pfscan_comb.c#L1234-L1237.
2. Unlocking a mutex, which isn't held, which is undefined behavior, e.g. forgetting to relock at the end of a loop for protected condition check: https://github.com/goblint/bench/blob/16ee6e2dfa54881be2161cb5bb6eee8cce5736d7/pthread/smtprc_comb.c#L2379-L2386.

At least in the case of non-recursive mutexes, these issues and other similar ones might be useful to have some simple analysis for to warn about.

sim642:
A related point that arose during GobCon is that joining a thread multiple times is also undefined behavior.

michael-schwarz:
Other instances Helmut and I identified as easy wins to claim we are fully-fledged analyzer of concurrency bugs:

- Mutexes being used before being initialized (or after being destroyed)
   - Warn when dynamically allocated mutex escapes before being initialized
   - Warn if `pthread_mutex_destroy` is called while other thread that use it may still be running or it is used in the thread that called destroy after it was destroyed.
- Returning from a thread while holding a mutex (more likely to be a bug than just returning from a function while holding a mutex)
- Calling `pthread_cond_wait` with a mutex that is not held

michael-schwarz:
Related #1100

sim642:
There are some behaviors to check for regarding `pthread_join` man page:
- [ ] Same thread ID is not being joined in parallel:
   > If multiple threads simultaneously try to join with the same thread, the results are undefined.
- [ ] Same thread ID has not already been joined before:
   > Joining with a thread that has previously been joined results in undefined behavior.

This is because the moment a thread has been joined, its concrete ID may be reused for a newly-created thread (although I guess this is very unlikely in practice?).
Thus, in the latter case, one may end up accidentally joining any other completely unrelated thread!
`pthread_self` man page also notes:
> A thread ID may be reused after a terminated thread has been joined, or a detached thread has terminated.

PR #655, opened 2022-03-21
Reimplement deadlock analysis and extend with MHP

sim642:
The refactoring of deadlock analysis for #650 turned into an almost complete rewrite. Extending with common lockset and MHP information was very straightforward.

### Changes
1. Use split global constraint unknowns for recording locking orders instead of a global OCaml variable.
2. Replace transitive locking order shortcutting on addition with a locking order graph search (DFS) at the end. This is more flexible and wastes less memory by not actually storing all the edges of the transitive closure.
3. Replace `Cil.location` with `Node` for locking events. This makes it incrementally more stable.
4. Improve readability of deadlock warnings by outputting the entire locking order cycle in a grouped message. Previously the transitive shortcutting made cycles longer than 2 difficult to interpret, because intermediate mutexes and locations were omitted.
5. Add `Deadlock` message category.
6. Exclude deadlocks by `access` information, which is independent of memory location (common lockset, MHP).

This is directly on master, but I will take care of the merge into #391 when this lands. I already know that will require a few changes to adapt this to #634, which is only on interactive).

### Example
Here's how `./regtest.sh 15 03` outputs the three-mutex locking cycle:
```
[Warning][Deadlock] Locking order cycle:
  lock before: mutex2 with [mhp:{tid=[main, t2@tests/regression/15-deadlock/03-triple_deadlock.c:42:5-42:40]}, thread:[main, t2@tests/regression/15-deadlock/03-triple_deadlock.c:42:5-42:40]] (tests/regression/15-deadlock/03-triple_deadlock.c:20:3-20:30)
  lock after: mutex3 with [mhp:{tid=[main, t2@tests/regression/15-deadlock/03-triple_deadlock.c:42:5-42:40]}, lock:{mutex2}, thread:[main, t2@tests/regression/15-deadlock/03-triple_deadlock.c:42:5-42:40]] (tests/regression/15-deadlock/03-triple_deadlock.c:21:3-21:30)
  lock before: mutex3 with mhp:{tid=[main], {t3@tests/regression/15-deadlock/03-triple_deadlock.c:43:5-43:40}} (tests/regression/15-deadlock/03-triple_deadlock.c:29:3-29:30)
  lock after: mutex1 with [mhp:{tid=[main], {t3@tests/regression/15-deadlock/03-triple_deadlock.c:43:5-43:40}}, lock:{mutex3}] (tests/regression/15-deadlock/03-triple_deadlock.c:30:3-30:30)
  lock before: mutex1 with mhp:{tid=[main], {t1@tests/regression/15-deadlock/03-triple_deadlock.c:41:5-41:40}} (tests/regression/15-deadlock/03-triple_deadlock.c:11:3-11:30)
  lock after: mutex2 with [mhp:{tid=[main], {t1@tests/regression/15-deadlock/03-triple_deadlock.c:41:5-41:40}}, lock:{mutex1}] (tests/regression/15-deadlock/03-triple_deadlock.c:12:3-12:30)
```


tests/regression/15-deadlock/03-triple_deadlock.c
```diff
@@ -1,4 +1,4 @@
-// SKIP PARAM: --set ana.activated[+] deadlock
```
sim642, line 1: Curiously, this was skipped before, because the transitive shortcutting meant that the pairwise forbidden order printing couldn't print all three.

src/analyses/deadlock.ml
```diff
@@ -4,109 +4,94 @@ open Prelude.Ana
...
+                ) lock_event_pairs
+            ) (ctx.global lock)
+        end
+      in
+
+      Stats.time "deadlock" (iter_lock LS.empty []) g
+    | _ -> Queries.Result.top q
 end
 
 let _ =
```
michael-schwarz, line 97: Here, we should probably record a dependency on `mutex`.
If it is not enabled, the `Lock` and `Unlock` events won't be emitted, and we wrongly report no deadlocks.

src/analyses/deadlock.ml
```diff
@@ -4,109 +4,94 @@ open Prelude.Ana
...
-        let lockAddrs = eval_exp_addr (Analyses.ask_of_ctx ctx) (List.hd arglist) in
-        if List.compare_length_with lockAddrs 1 = 0 then
-          let inLockAddrs e = List.exists (fun r -> ValueDomain.Addr.equal r e.addr) lockAddrs in
-          D.filter (neg inLockAddrs) ctx.local
-        else ctx.local
-      | _ -> ctx.local
-
+  let event ctx (e: Events.t) octx =
+    match e with
+    | Lock addr ->
```
michael-schwarz, line 42: Do we actually always emit this for may locks? I am a bit worried about this place here:

https://github.com/goblint/analyzer/blob/b899e74ce22a5631bc811bc06f901256511eabe0/src/analyses/mutexAnalysis.ml#L99-L104

Or to phrase it in a more general way: Would this entire analysis not need to be based on may-locks?
(They happen to coincide with must locks pretty often as we are path-sensitive wr.t. to must locksets by default, but there is nothing mandating that we always are)

michael-schwarz:
Also, I think we now get to the point, where we can probably not just ignore double locking anymore. While it may be undefinied behavior in pthreads, I think double locking a mutex that is not configured to support it, also constitutes a deadlock in some philosophical sense, right?

src/analyses/deadlock.ml
sim642: True, this analysis is useless otherwise, so I'll add the dependency.

src/analyses/deadlock.ml
sim642: The deadlock analysis indeed needs to be based on may locksets. Both before and after this PR, the local domain `D` for the deadlock analysis does use a may set of locks (with locations) instead of a reversed must set.

@vesalvojdani had the same concerns as you because indeed we don't have a maintained may lockset analysis. Besides the deadlock analysis keeping its own may locks events, we also have some ancient maylocks analysis, which is very outdated compared to mutex analysis. With path-sensitivity they should almost always coincide.

The case code you linked is just the tip of the ice berg though of a more general issue that we need to refactor our must/may lockset analyses. Ideally the `Lock`/`Unlock` events would be emitted the same regardless and the difference is just in whether the local domain is reversed or not. For example, both mutex and maylocks analyses could be just instantiated from a functor with local domains having locksets with differing lattice orderings. And similarly this deadlock analysis would just instantiate it with a may lock events set.

The code you point out isn't just a problem for the deadlock analysis, but probably also some privatizations, where we require mutex path-sensitivity to avoid the may-must distinction. With ambiguous lockings the code avoids any events altogether (which is just appropriate for must locksets), but then we also lack any path-sensitivity accounting for the actual possibilities, which in turn could make such privatizations unsound.

I think this is something we should address separately.

sim642:
> Also, I think we now get to the point, where we can probably not just ignore double locking anymore. While it may be undefinied behavior in pthreads, I think double locking a mutex that is not configured to support it, also constitutes a deadlock in some philosophical sense, right?

That's a good point indeed, especially since recursive mutexes don't self-deadlock (although for the new privatizations we assume they do AFAIK). That should probably be addressed separately too, to have tracking of mutex types and then we could relatively easily simply even warn about that default undefined behavior. This is related to #176.

michael-schwarz:
> although for the new privatizations we assume they do AFAIK

No, iirc we simply assume all mutexes are recursive in the implementation, and in the paper we don't deal with them at all.

sim642:
Turns out our framework is so great that adding common lockset and MHP information to locking events to exclude some deadlocks was very straightforward, so I added that into this PR. With our existing mutex, mhp and threadJoins analyses, we can deal with the example from here: https://arxiv.org/abs/1607.06927 (also added as regression test).

src/analyses/deadlock.ml
sim642: This will be refactored and fixed by #662.

michael-schwarz:
I am still adding some examples and checking some smaller things. When I'm done with that (hopefully later today), I'd be cool with merging it and then fixing the issues with Must vs May locks in #662 as you suggested.

michael-schwarz:
I think the most problematic issue we still have here is threads that end while holding locks. Any future attempt to acquire that lock may result in a deadlock. (Not sure if this is a deadlock according to some definitions that require each thread to be waiting on a resource, but I still think this something we should attempt to catch, maybe we could do it an a separate PR though).

----

Also, I wasted a lot of time trying to come up with an example where one has a deadlock despite not all locking operations pairwise may-happen-in parallel, which turns out was a waste of time.
In the process, I found a (maybe questionable) enhancement of MHP: P1 and P2 may not happen in parallel, if some thread is must-joined at P1 but definitely not created at P2. I now added it and an example for it to this PR, but we can of course also move that to somewhere else.

src/domains/queries.ml
```diff
@@ -59,7 +59,11 @@ module Unit = Lattice.Unit
 type maybepublic = {global: CilType.Varinfo.t; write: bool} [@@deriving ord, hash]
 type maybepublicwithout = {global: CilType.Varinfo.t; write: bool; without_mutex: PreValueDomain.Addr.t} [@@deriving ord, hash]
 type mustbeprotectedby = {mutex: PreValueDomain.Addr.t; global: CilType.Varinfo.t; write: bool} [@@deriving ord, hash]
-type partaccess = {exp: CilType.Exp.t; var_opt: CilType.Varinfo.t option; write: bool} [@@deriving ord, hash]
+type memory_access = {exp: CilType.Exp.t; var_opt: CilType.Varinfo.t option; write: bool} [@@deriving ord, hash]
+type access =
+  | Memory of memory_access (** Memory location access (race). *)
+  | Point (** Program point and state access (MHP), independent of memory location. *)
+[@@deriving ord, hash] (* TODO: fix ppx_deriving_hash on variant with inline record *)
```
michael-schwarz, line 66: I think you already fixed this in `ppx_deriving_hash` by now, right?

michael-schwarz (approved):
I think how easy it was to enhance this with MHP information is a good sign we went for the right design here, and as always, I'm impressed with the speed with which you implemented this Simmo!

michael-schwarz:
Not sure if we want to solve the issue with threads ending while holding mutexes here or make a new issue and PR for it?

src/domains/queries.ml
sim642: Indeed, but I didn't yet make a new release just for that nor add an opam pin to Goblint to annoy everyone. There's no downside to having a separate named record type here, so I'd just leave it as is for now and change in sometime in the future.

sim642:
> Not sure if we want to solve the issue with threads ending while holding mutexes here or make a new issue and PR for it?

I think we should `SKIP` it here and have a separate issue, because that's a somewhat different form of a deadlock. It's not just about locking order any more.

sim642:
> In the process, I found a (maybe questionable) enhancement of MHP: P1 and P2 may not happen in parallel, if some thread is must-joined at P1 but definitely not created at P2. I now added it and an example for it to this PR, but we can of course also move that to somewhere else.

This is the reason why 15-deadlock-mhp2 doesn't have a deadlock, right? It had no annotations, so I added them, hopefully correctly.

michael-schwarz:
> This is the reason why 15-deadlock-mhp2 doesn't have a deadlock, right?

Yes!

PR #656, opened 2022-03-21
Improve cfg comparison

stilscher:
This PR contains an alternative implementation of the phase 2 from the CFG comparison. I prefer it over the previous one because I think it is easier to understand and read and it also seems to be quite a bit faster at least on this small example:
https://github.com/goblint/bench/blob/master/pthread/knot_comb.c with https://github.com/goblint/bench/blob/master/pthread/knot_comb01.patch:
```
TOTAL                           1.503 s
  parse                           0.027 s
  convert to CIL                  0.021 s
  compareCilFiles                 0.019 s
    compare-phase1                  0.003 s
    compare-phase2                  0.001 s
  analysis                        1.438 s
    global_inits                    0.003 s
    solving                         1.383 s
      S.Dom.equal                     0.009 s
      postsolver                      0.317 s
    warn_global                     0.026 s
      access                          0.026 s
```
previously:
```
TOTAL                           4.824 s
  parse                           0.028 s
  convert to CIL                  0.021 s
  compareCilFiles                 3.492 s
    compare-phase1                  0.003 s
    compare-phase2                  3.469 s
  analysis                        1.285 s
    global_inits                    0.002 s
    solving                         1.268 s
      S.Dom.equal                     0.008 s
      postsolver                      0.278 s
    warn_global                     0.006 s
      access                          0.005 s
```

In phase 1 of the CFG comparison all matching nodes are collected, that are reachable via at least one path of matching nodes from the entry node of the function. This set of nodes is however too big because it might contain a node that is influenced by a change through a new back edge (which was not recorded in the old `infl` map and will therefore not lead to the nodes destabilization before solving).
Phase 2 therefore refines this set of matching nodes found in phase 1. Previously a DFS on the new CFG was used to detect possible back edges and remove influenced nodes from the set of matching nodes. Now I implemented something like a fixpoint algorithm, that removes nodes from the set of matching nodes that have a predecessor which is not in the set of matching nodes, until it stabilizes. Any node that is removed will also be added to the set of differing nodes (an over-approximation of the set of nodes that must be destabilized).
For the set of matching nodes I now maintain two hash maps to find the corresponding node fast no matter whether the first or second node needs to be looked up.

EDIT: I'm still wondering a little about how big the difference in performance is. But so far I couldn't find any problematic behavior or bugs

### DIFF
--- a/src/incremental/compareCFG.ml
+++ b/src/incremental/compareCFG.ml
@@ -94,42 +94,27 @@ let compareCfgs (module CfgOld : CfgForward) (module CfgNew : CfgForward) fun1 f
 (* This is the second phase of the CFG comparison of functions. It removes the nodes from the matching node set 'same'
  * that have an incoming backedge in the new CFG that can be reached from a differing new node. This is important to
  * recognize new dependencies between unknowns that are not contained in the infl from the previous run. *)
-let reexamine f1 f2 (same : biDirectionNodeMap ref) (diffNodes1 : unit NH.t) (module CfgOld : CfgForward) (module CfgNew : CfgForward) =
-  NH.iter (fun n1 _ -> if NH.mem !same.node1to2 n1 then (let n2 = NH.find !same.node1to2 n1 in NH.remove !same.node1to2 n1; NH.remove !same.node2to1 n2)) diffNodes1;
-  NH.replace !same.node1to2 (FunctionEntry f1) (FunctionEntry f2); NH.replace !same.node2to1 (FunctionEntry f2) (FunctionEntry f1);
-  let module NS = Set.Make(Node) in
-  let diffNodes2 = NH.create 103 in
+let reexamine f1 f2 (same : biDirectionNodeMap ref) (diffNodes1 : unit NH.t) (module CfgOld : CfgForward) (module CfgNew : CfgForward) (module CfgNewPred : CfgBackward) =
+  let entry1 = (FunctionEntry f1) in
+  let entry2 = (FunctionEntry f2) in
+  NH.replace !same.node1to2 entry1 entry2; NH.replace !same.node2to1 entry2 entry1;
+  let rec repeat () =
+    let ls = List.of_seq (NH.to_seq_keys !same.node2to1) in
+    let rec check_nodes_in_same ps n = match ps with
+      | [] -> true
+      | p::ps' -> if not (NH.mem !same.node2to1 p) then (
+          let n1 = NH.find !same.node2to1 n in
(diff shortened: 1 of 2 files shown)


src/incremental/compareCFG.ml
```diff
@@ -45,7 +42,7 @@ module NTH = Hashtbl.Make(
  * comparison is done for its successors. The two function entry nodes make up the tuple to start the comparison from. *)
 let compareCfgs (module CfgOld : CfgForward) (module CfgNew : CfgForward) fun1 fun2 =
   let diff = NH.create 113 in
-  let same = NTH.create 113 in
+  let same = ref {node1to2=NH.create 113; node2to1=NH.create 113} in
```
sim642, line 45: Does this need to be a `ref` at all? Both the hashtables themselves are mutable anyway.

src/incremental/compareCFG.ml
```diff
@@ -65,11 +62,10 @@ let compareCfgs (module CfgOld : CfgForward) (module CfgNew : CfgForward) fun1 f
             (* TODO: don't allow pseudo return node to be equal to normal return node, could make function unchanged, but have different sallstmts *)
             if eq_node (toNode1, fun1) (toNode2, fun2) && eq_edge_list edgeList1 edgeList2 then
               begin
-                let notInSame = not (NTH.mem same (toNode1, toNode2)) in
-                let matchedAlready = NTH.fold (fun (toNode1', toNode2') _ acc ->
-                    acc || (Node.equal toNode1 toNode1' && not (Node.equal toNode2 toNode2'))) same false in
+                let notInSame = not (NH.mem !same.node1to2 toNode1 && Node.equal (NH.find !same.node1to2 toNode1) toNode2) in
+                let matchedAlready = NH.mem !same.node1to2 toNode1 && not (Node.equal (NH.find !same.node1to2 toNode1) toNode2) in
```
sim642, line 66: These look awfully similar now and do the same hashtable lookup 4 times. It might be easier to just do a single `match NH.find_opt ... with ...`. Especially since it looks like all four combinations of the two booleans wouldn't even be possible.

sim642 (commented):
The phase 2 logic is currently very involved, written using three nested (not just mutual!) recursive functions. If at all possible, it would be good to replace some of those with standard higher-order functions on the data structures. It would immensely simplify the understanding of this code, which currently tightly mixes crucial logic and basic list recursions.

sim642 (approved):
Assuming it still works as intended after my code style suggestions, it should be good to merge.

Issue #658, opened 2022-03-22
Mutex types support


sim642:
With the revival of deadlock analysis (#655), it becomes more and more important that we handle behavior specific to mutex types.

Via [`pthread_mutexattr_settype`](https://linux.die.net/man/3/pthread_mutexattr_settype) there are the following possibilities:
- [ ] `PTHREAD_MUTEX_DEFAULT` – misuse is undefined behavior
- [ ] `PTHREAD_MUTEX_NORMAL` – relocking deadlocks, other misuse is undefined behavior
- [ ] `PTHREAD_MUTEX_ERRORCHECK` – misuse returns errors
- [ ] `PTHREAD_MUTEX_RECURSIVE` – relocking allowed, other misuse returns errors

Additionally:
- [ ] We don't do anything smart with read-write locks (`pthread_rwlock_rdlock`, `pthread_rwlock_wrlock`).
- [ ] Since some of the types are about differences in return values, we should also consider those, instead of completely ignoring them.

The different mutex types affect multiple features:
* Race detection
* Deadlock detection
* Base and Apron privatization

michael-schwarz:
> We don't do anything smart with read-write locks (`pthread_rwlock_rdlock`, `pthread_rwlock_wrlock`).

Are you sure? I think we have support for those.

sim642:
That support is questionable at best:
https://github.com/goblint/analyzer/blob/179c8b94ae0c801683d00fbf89f14e7e81223091/src/analyses/mutexAnalysis.ml#L165-L171
Whether a memory access is a read or write has nothing to do with whether we're holding a read or write lock!
Instead they relax one form of pairwise exclusion: reader-reader pairs do not mutually exclude each other. That's in no way reflected in the `may_race` check there.

sim642:
That is also reflected by this existing regression test:
https://github.com/goblint/analyzer/blob/179c8b94ae0c801683d00fbf89f14e7e81223091/tests/regression/04-mutex/41-pt_rwlock.c#L9-L24
Writer-reader pairs do mutually exclude each other, so there should be no race, but we report one because our assumption is violated since `data2` is written under a read lock.

I suppose we might still be at least sound here, although we could also be more precise and prove both globals race-free by dropping our silly assumption.

PR #662, opened 2022-03-23
Refactor mutex analyses

### LINKED ISSUE #659: Refactor and fix different lockset analyses
We have at least three lockset analyses:
1. mutex – must locksets,
2. maylocks – may locksets,
3. deadlock – may locking events (locks with location).

These could be combined into a single framework, where `Lock` and `Unlock` events work for all of them.

---

We don't have a maintained may lockset analysis. Besides the deadlock analysis keeping its own may locks events, we also have some ancient maylocks analysis, which is very outdated compared to mutex analysis. With path-sensitivity they should almost always coincide.

The case code you linked is just the tip of the ice berg though of a more general issue that we need to refactor our must/may lockset analyses. Ideally the `Lock`/`Unlock` events would be emitted the same regardless and the difference is just in whether the local domain is reversed or not. For example, both mutex and maylocks analyses could be just instantiated from a functor with local domains having locksets with differing lattice orderings. And similarly this deadlock analysis would just instantiate it with a may lock events set.

The code you point out isn't just a problem for the deadlock analysis, but probably also some privatizations, where we require mutex path-sensitivity to avoid the may-must distinction. With ambiguous lockings the code avoids any events altogether (which is just appropriate for must locksets), but then we also lack any path-sensitivity accounting for the actual possibilities, which in turn could make such privatizations unsound.

_Originally posted by @sim642 in https://github.com/goblint/analyzer/pull/655#discussion_r831902662_

sim642:
Closes #659.

### Changes
1. Add mutexEvents analysis to emit events handled by mutex, mayLocks and deadlock analyses.
2. Replace analysis-specific ambiguous lock pointer handling with path splitting via `ctx.split`. One split is introduced for each possible mutex pointer. Joining of paths automagically takes care of the ambiguity (see table below).
    This also increases the precision of locking an ambiguous lock pointer in the path-sensitive mutex analysis: previously no must locks were added, but now it properly splits into multiple paths with different locksets.
4. Add more complete handling of failing locks (includes try locks) by just having an extra path split via `ctx.split` if failure is allowed. Again, the joining of paths takes care of everything by considering the extra path where no lock was acquired,
5. Add more complete handling of blob locks via `IsMultiple` query (see table below). This should additionally account for weakly updated locals too.
6. Add more complete handling of unknown lock pointers using `Addr.UnknownPtr` (see table below). Previously lockset analyses just dropped these from the `MayPointsTo` query result conversion.
7. Fix deadlock detection w.r.t. ambiguous, blob and unknown lock pointers.
8. Extract very simple must and may lockset analysis functors and instantiate mutex, maylocks and deadlock analyses using these.

### Behavior
The following tables give an overview of all the cases that are now considered.

#### Must lockset
| | Definite pointer | Ambiguous pointer | Blob pointer | Unknown pointer | Failing lock |
| --- | --- | --- | --- | --- | --- |
| Lock | Add pointer | Don't add any | Don't add | Don't add | Don't add |
| Unlock | Remove pointer | Remove all ambiguous pointers | N/A (never added) | Remove all | N/A |

#### May lockset
| | Definite pointer | Ambiguous pointer | Blob pointer | Unknown pointer | Failing lock |
| --- | --- | --- | --- | --- | --- |
| Lock | Add pointer | Add all ambiguous pointers | Add blob pointer | Add unknown pointer | Add |
| Unlock | Remove pointer | Don't remove any | Don't remove | Don't remove any | N/A |

### TODO
- [x] Add unknown mutex pointer handling for both must and may locksets.
- [x] Make failing locks emit all events.
- [x] Fix mayLocks and deadlock unlocking.
- [x] Determine whether `MustLock` and `MustUnlock` events are still necessary.
- [x] Merge #655.

### DIFF
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -412,12 +412,12 @@ struct
     let st = ctx.local in
     match e with
     | Events.Lock (addr, _) when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
-      Priv.lock (Analyses.ask_of_ctx octx) octx.global st addr
+      Priv.lock (Analyses.ask_of_ctx ctx) ctx.global st addr
     | Events.Unlock addr when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
-      Priv.unlock (Analyses.ask_of_ctx octx) octx.global octx.sideg st addr
+      Priv.unlock (Analyses.ask_of_ctx ctx) ctx.global ctx.sideg st addr
     (* No need to handle escape because escaped variables are always referenced but this analysis only considers unreferenced variables. *)
     | Events.EnterMultiThreaded ->
-      Priv.enter_multithreaded (Analyses.ask_of_ctx octx) octx.global octx.sideg st
+      Priv.enter_multithreaded (Analyses.ask_of_ctx ctx) ctx.global ctx.sideg st
     | _ ->
       st
 
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -408,16 +408,16 @@ struct
   let threadspawn ctx lval f args fctx =
     ctx.local
 
-  let event ctx e aprx =
+  let event ctx e octx =
     let st = ctx.local in
     match e with
     | Events.Lock (addr, _) when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
-      Priv.lock (Analyses.ask_of_ctx aprx) aprx.global st addr
+      Priv.lock (Analyses.ask_of_ctx octx) octx.global st addr
     | Events.Unlock addr when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
-      Priv.unlock (Analyses.ask_of_ctx aprx) aprx.global aprx.sideg st addr
+      Priv.unlock (Analyses.ask_of_ctx octx) octx.global octx.sideg st addr
     (* No need to handle escape because escaped variables are always referenced but this analysis only considers unreferenced variables. *)
     | Events.EnterMultiThreaded ->
-      Priv.enter_multithreaded (Analyses.ask_of_ctx aprx) aprx.global aprx.sideg st
+      Priv.enter_multithreaded (Analyses.ask_of_ctx octx) octx.global octx.sideg st
     | _ ->
       st
 
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -411,9 +411,9 @@ struct
   let event ctx e aprx =
     let st = ctx.local in
     match e with
-    | Events.Lock addr when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
+    | Events.MustLock addr when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
       Priv.lock (Analyses.ask_of_ctx aprx) aprx.global st addr
-    | Events.Unlock addr when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
+    | Events.MustUnlock addr when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
       Priv.unlock (Analyses.ask_of_ctx aprx) aprx.global aprx.sideg st addr
     (* No need to handle escape because escaped variables are always referenced but this analysis only considers unreferenced variables. *)
     | Events.EnterMultiThreaded ->
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -414,6 +414,8 @@ struct
     | Events.Lock (addr, _) when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
       Priv.lock (Analyses.ask_of_ctx ctx) ctx.global st addr
     | Events.Unlock addr when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
+      if addr = UnknownPtr then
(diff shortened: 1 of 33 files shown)


sim642:
> * Determine whether `MustLock` and `MustUnlock` events are still necessary.

The only difference that these still provide is that when the general `Lock` is emitted, `MustLock` is emitted conditionally:
https://github.com/goblint/analyzer/blob/d805791602dfbd2881b907b537c7ecef5076949f/src/analyses/mutexAnalysis.ml#L17-L19
Removing that relocking condition causes a failure in just one test: 13-privatized/70-mm-reentrant.

At first sight it seems like a reasonable thing to do: only emit `MustLock` when it's not already locked. But that relocking check is itself based on must locksets. If the mutex isn't already must locked but may locked, then it is treated as if it wasn't locked at all, which doesn't account for the possibility that it might have been already locked. In such scenario we don't know whether it's the first locking or relocking.

Actually emitting `MustLock` events like this isn't even necessarily right, because in the case of ambiguity or failure, the particular split path still sees the `MustLock` event, although the following path join might remove it from the must lockset. So this event cannot even guarantee anything. It's similar to the whole branched thread creation and `EnterMultithreaded` event problem in that a later join would have to account for events that along some path didn't happen.

Therefore I'm unsure whether this is even the right thing to do. Proper handling of recursive locks is a whole other issue #658.

src/analyses/mutexAnalysis.ml
```diff
@@ -23,92 +20,42 @@ end
...
+      D.add l ctx.local
+
+    let remove ctx l =
+      D.remove (l, true) (D.remove (l, false) ctx.local)
+
+    let remove_all ctx =
+      (* TODO: implement in privatizations? *)
+      (* Mutexes.iter (fun m ->
+          ctx.emit (MustUnlock m)
+        ) (D.export_locks ctx.local); *)
```
sim642, line 39: There is the question how privatizations should handle unlocking of unknown pointers, i.e. the must lockset essentially becoming empty. We haven't handled this before, but now there's the possibility.

I wouldn't want to try implementing a `unlock_all` for all of our privatizations right away, so maybe just have an unsoundness message in this case?

michael-schwarz: >  so maybe just have an unsoundness message in this case?

Yes, I think that should suffice for now.

src/analyses/mutexEventsAnalysis.ml
```diff
@@ -0,0 +1,153 @@
...
+    match x with
+    | `NoOffset    -> `NoOffset
+    | `Index (Const (CInt (i,_,s)),o) -> `Index (IntDomain.of_const (i,Cilfacade.ptrdiff_ikind (),s), conv_offset o)
+    | `Index (_,o) -> `Index (ValueDomain.IndexDomain.top (), conv_offset o)
+    | `Field (f,o) -> `Field (f, conv_offset o)
+
+  let rec conv_offset_inv = function
+    | `NoOffset -> `NoOffset
+    | `Field (f, o) -> `Field (f, conv_offset_inv o)
+    (* TODO: better indices handling *)
```
michael-schwarz, line 34: Should we do this here then? It should be trivial to go back to `Const (CInt (i,_,s))` here, right?

src/analyses/mutexEventsAnalysis.ml
```diff
@@ -0,0 +1,153 @@
...
+let verifier_atomic = LockDomain.Addr.from_var (Goblintutil.create_var (makeGlobalVar "[__VERIFIER_atomic]" intType))
+
+module Spec: MCPSpec =
+struct
+  include UnitAnalysis.Spec
+  let name () = "mutexEvents"
+
+
+  (* NB! Currently we care only about concrete indexes. Base (seeing only a int domain
+     element) answers with the string "unknown" on all non-concrete cases. *)
```
michael-schwarz, line 23: Is this comment still up-to-date? Where does Base answer with the string(?!) `unknown`?

src/analyses/mutexAnalysis.ml
```diff
@@ -23,92 +20,42 @@ end
 (** Mutex analyzer without base --- this is the new standard *)
```
michael-schwarz, line 20: I think it's been the new standard for long enough, we can drop that comment.

src/analyses/mutexAnalysis.ml
```diff
@@ -23,92 +20,42 @@ end
...
-  (* NB! Currently we care only about concrete indexes. Base (seeing only a int domain
-     element) answers with the string "unknown" on all non-concrete cases. *)
-  let rec conv_offset x =
-    match x with
-    | `NoOffset    -> `NoOffset
-    | `Index (Const (CInt (i,_,s)),o) -> `Index (IntDomain.of_const (i,Cilfacade.ptrdiff_ikind (),s), conv_offset o)
-    | `Index (_,o) -> `Index (ValueDomain.IndexDomain.top (), conv_offset o)
-    | `Field (f,o) -> `Field (f, conv_offset o)
-
   let rec conv_offset_inv = function
```
michael-schwarz, line 53: This is duplicated now?

michael-schwarz:
> EDIT: I managed to get rid of those events by just adding the recursive locking check into mutex-oplus and mutex-meet specifically to make them sound on that test, others were already correct.

Would the apron mutex meet not also require such special handling then?

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -408,16 +408,16 @@ struct
...
     ctx.local
 
-  let event ctx e aprx =
+  let event ctx e octx =
     let st = ctx.local in
     match e with
-    | Events.Lock addr when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
-      Priv.lock (Analyses.ask_of_ctx aprx) aprx.global st addr
+    | Events.Lock (addr, _) when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
+      Priv.lock (Analyses.ask_of_ctx ctx) ctx.global st addr
```
michael-schwarz, line 415: Would we not need this handling of relocking here?

src/analyses/mutexAnalysis.ml
sim642: Indeed, it's unused in MutexEventsAnalysis, so I'll remove it from there.

src/analyses/mutexEventsAnalysis.ml
sim642: Done in MutexAnalysis, symmetrically to `conv_offset`.

src/analyses/mutexEventsAnalysis.ml
sim642: I don't think we do indeed, unless there's somehow actually `unknown_exp` in an index. I just removed that sentence.

michael-schwarz (approved):

PR #663, opened 2022-03-23
Exclude functions with changed header from reluctant destabilization

stilscher:
We discovered that one common case where reluctant destabilization leads to a greater number of evaluation and a worse runtime as without reluctant destabilization, is when the header (that is for example the type) of a function changes.

When the type of a function changes, the contexts for which it needs to be analyzed also change and therefore it is superfluous to analyze it for the previous but outdated context first. Also for functions whose type changed, the program point where this function is called is also detected to have changed, so the changed function is destabilized completely anyway. 

This optimization substantially reduces the number of commits in the zstd repository for which reluctant destabilization deteriorates the runtime of an incremental analysis.

### DIFF
--- a/src/incremental/compareCIL.ml
+++ b/src/incremental/compareCIL.ml
@@ -23,15 +23,17 @@ type change_info = {
   mutable unchanged: global list;
   mutable removed: global list;
   mutable added: global list;
-  mutable force_reanalyze: VarinfoSet.t;
+  mutable exclude_from_rel_destab: VarinfoSet.t;
   (** Set of functions that are to be force-reanalyzed.
       These functions are additionally included in the [changed] field, among the other changed globals. *)
 }
 
 let empty_change_info () : change_info =
-  {added = []; removed = []; changed = []; unchanged = []; force_reanalyze = VarinfoSet.empty}
+  {added = []; removed = []; changed = []; unchanged = []; exclude_from_rel_destab = VarinfoSet.empty}
 
-type change_status = Unchanged | Changed | ForceReanalyze of Cil.fundec
+(* 'ChangedFunHeader' is used for functions whose varinfo or formal parameters changed. 'Changed' is used only for
+ * changed functions whose header is unchanged and changed non-function globals *)
+type change_status = Unchanged | Changed | ChangedFunHeader of Cil.fundec | ForceReanalyze of Cil.fundec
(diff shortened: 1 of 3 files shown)


src/incremental/compareCIL.ml
```diff
@@ -23,15 +23,17 @@ type change_info = {
   mutable unchanged: global list;
   mutable removed: global list;
   mutable added: global list;
-  mutable force_reanalyze: VarinfoSet.t;
+  mutable exclude_from_rel_destab: VarinfoSet.t;
```
sim642, line 26: Why is `force_reanalyze` removed here, but in `change_status` there's still a case for it?

src/incremental/compareCIL.ml
```diff
@@ -94,16 +95,20 @@ let compareCilFiles ?(eq=eq_glob) (oldAST: file) (newAST: file) =
...
-      let append_to_changed () =
+      let change_status, diff = eq old_global global cfgs in
+      let append_to_changed (unchangedHeader) =
         changes.changed <- {current = global; old = old_global; unchangedHeader; diff} :: changes.changed
       in
       match change_status with
-      | Changed -> append_to_changed ()
+      | ChangedFunHeader f ->
+        changes.exclude_from_rel_destab <- VarinfoSet.add f.svar changes.exclude_from_rel_destab;
+        append_to_changed false
```
sim642, line 105: `append_to_changed false` looks weird, as if the `false` argument indicates not appending. Would be clearer with a named argument: `append_to_changed ~unchangedHeader:false`.

src/incremental/compareCIL.ml
stilscher: The set `force_reanalyze` was up to now only used to exclude the contained functions from reluctant destabilization. Since functions with a changed header should be handled exactly the same way, I collect both in the same set and renamed it to `force_reanalyze` such that it is clear what functions it contains.

In change_status I kept them separate such that it is more obvious for which reason a function is added to the `force_reanalyze` set. One could probably also get rid of this distinction since they are handled the same way and it produces some duplicate code. Its just that, in my opinion, a `changedFunHeader` and a `ForceReanalyze` function do not really share a property that could be considered a `change_status`.

sim642 (approved):

PR #665, opened 2022-03-24
Bump goblint-cil (https://github.com/goblint/cil/pull/85)

michael-schwarz:
To include fix https://github.com/goblint/cil/pull/85 needed to benchmark zstd.

### DIFF
--- a/goblint.opam
+++ b/goblint.opam
@@ -64,7 +64,7 @@ dev-repo: "git+https://github.com/goblint/analyzer.git"
 # on `dune build` goblint.opam will be generated from goblint.opam.template and dune-project
 # also remember to generate/adjust goblint.opam.locked!
 pin-depends: [
-  [ "goblint-cil.1.8.2" "git+https://github.com/goblint/cil.git#c47bbbdb442b4d50dac5a0879404ab3b5592170a" ]
+  [ "goblint-cil.1.8.2" "git+https://github.com/goblint/cil.git#d07f7dbecd1fdf5d4e8b5f1d370b9ff86f256f3b" ]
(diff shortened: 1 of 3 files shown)


sim642:
> readthedocs CI failing is unrelated.

I added a workaround on master.

sim642:
The big issue with not merging inlines is still there, silently breaking our assumptions: https://github.com/goblint/cil/issues/84 and https://github.com/goblint/analyzer/commit/62d759196717734cb01838080cdf5f9f568fe455#commitcomment-67629901. So I don't understand why you really want to avoid `merge_inlines` instead of fixing the merging errors we get with it enabled instead.

michael-schwarz:
> So I don't understand why you really want to avoid merge_inlines instead of fixing the merging errors we get with it enabled instead.

I don't think it is doing one to the exclusion of the other. This fix allows us to go ahead with our benchmarks without any issues. We should at some future point also fix the merging of inlines (even though I personally find the notion of doing that a bit dubious), but CIL should still work when it is disabled.

Note that it would be perfectly sensible for code to provide two different extern inline functions of the same name in two different compilation units when declaring the `gnu_inline` function attribute, so this code is needed regardless.

I don't think I understand in what sense any invariant is broken here? The one-to-one correspondence between a `fundec` and its `varinfo` is still there (as before), the only difference now is that the names of the corresponding `varinfo`s are also globally unique.

michael-schwarz:
And if it is indeed problematic what does it have to do with `inline`? Is it not equally problematic for static functions then?

michael-schwarz:
Waiting for https://github.com/goblint/cil/pull/86 to land, and update then

sim642 (approved):
I suppose this should be good to go now.

PR #666, opened 2022-03-24
Update README.md for installation on M1 processors

ystade:
Home-brew changed its installation location from /usr/local/ to /opt/homebrew/. Some packages search dependencies still in the old location. For them to find their dependencies two symbolic link need to be created.


michael-schwarz:
The CI failure is once more unrelated, readthedocs seems to have some issue today.

PR #669, opened 2022-03-27
Add command line argument completion

sim642:
The ultimate usability feature we've all been waiting for!

You can tab complete:
* command line arguments,
* option names (including partially),
* boolean-only option names for `--enable` & `--disable`,
* string enumerated & boolean option values,
* default option values.

### Usage
https://github.com/goblint/analyzer/blob/bf9c184503973feba19163864c0dd3752bda7bea/scripts/bash-completion.sh#L3-L7

This is just bash completion though, so I guess it won't work with MacOS's zsh. It should be possible to write a similar completion script for zsh, but zsh's completion support is so overengineered that I couldn't quickly figure out how to do this easiest thing.


michael-schwarz (approved):
Seems to work very well, definitely makes working with Goblint easier!

sim642:
I think I'll try to make an opam release of my arg-complete library so this can be merged without opam pins.

PR #674, opened 2022-04-01
Handling relative paths in compilation database processing

stilscher:
For a command from the compilation databases, the field `file` can contain a relative or absolute path. Differences can for example be observed when generating the compilation database with different tools (i.e. `compiledb` and `bear`). With `compiledb`, the file path is only a relative path whereas with bear it is an absolute path. To be able to run goblint with differently generated compilation databases, I adapted the path handling in the compilation database processing slightly, to concat the `file` path with the `directory` path when necessary.

Also, to be able to specify the repository to be analyzed with a relative path, I take the `realpath` of the concatenation of the repository directory and the compilation database file name.


sim642:
Apparently `Unix.realpath` only exists since OCaml 4.13, so our unlocked builds with earlier version are now failing on `master`. So I'll have to rush #642, which doesn't require `Unix.realpath`.

PR #675, opened 2022-04-04
Rename and split `exp.precious_globs` to `exp.exclude_from_earlyglobs` and `exp.exclude_from_invalidation`

### LINKED ISSUE #192: Rename `exp.precious_globs` and clarify what it does
The documentation currently says

~~~OCaml
reg Experimental "exp.precious_globs"    "[]"    "Global variables that should be handled flow-sensitively when using earlyglobs."
~~~

However, Simmo discovered that "precious globs" are also "my favorite things" that exclude things from invalidation:
https://github.com/goblint/analyzer/blob/85536ab5c9cc0ffe5088af1c01eef34be27fa00e/src/analyses/base.ml#L1799-L1803

So one should rethink what `precious_globs` are, give them a sensible name, and describe their behavior.

----

Also, we should rename `exp.extraspecials` to something that better describes what it means

~~~OCaml
;reg Experimental "exp.extraspecials"     "[]"    "List of functions that must be analyzed as unknown extern functions"
~~~

michael-schwarz:
Before, the name was not really telling, and it did two things at the same time:

- Exclude them from flow-insensitive treatment if `earlyglobs` is enabled
- Exclude them from invalidation 

These are conceptually different, so I now pulled them apart and also gave the options are more telling name.

Closes #192. I think we can put the the remarks about special to the side for now, as the library analysis will be completely redone by @jerhard.

### DIFF
--- a/src/analyses/baseUtil.ml
+++ b/src/analyses/baseUtil.ml
@@ -9,11 +9,15 @@ let is_static (v:varinfo): bool = v.vstorage = Static
 
 let is_always_unknown variable = variable.vstorage = Extern || Ciltools.is_volatile_tp variable.vtype
 
-let precious_globs = ref []
-let is_precious_glob v = List.mem v.vname !precious_globs
+let exclude_from_earlyglobs = ref []
+let exclude_from_invalidation = ref []
+
+let is_excluded_from_earlyglobs v = List.mem v.vname !exclude_from_earlyglobs
+let is_excluded_from_invalidation v =  List.mem v.vname !exclude_from_invalidation
 
 let after_config () =
-  precious_globs := get_string_list "exp.precious_globs"
+  exclude_from_earlyglobs := get_string_list "exp.exclude_from_earlyglobs";
+  exclude_from_invalidation := get_string_list "exp.exclude_from_invalidation"
 
 let _ =
   AfterConfig.register after_config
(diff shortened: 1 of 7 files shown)


src/analyses/baseUtil.ml
```diff
@@ -9,11 +9,15 @@ let is_static (v:varinfo): bool = v.vstorage = Static
...
+let exclude_from_earlyglobs = ref []
+let exclude_from_invalidation = ref []
+
+let is_excluded_from_earlyglobs v = List.mem v.vname !exclude_from_earlyglobs
+let is_excluded_from_invalidation v =  List.mem v.vname !exclude_from_invalidation
 
 let after_config () =
-  precious_globs := get_string_list "exp.precious_globs"
+  exclude_from_earlyglobs := get_string_list "exp.exclude_from_earlyglobs";
+  exclude_from_invalidation := get_string_list "exp.exclude_from_invalidation"
```
sim642, line 20: Given the option caching, this is mostly unnecessary (unless we want to convert the list to a hashtable for faster lookups if we exclude a lot).
As is, this would need to be also called in server mode (where `reset_lazy`s are called) to reset the globals from options for secondary analysis requests.

sim642 (approved):

PR #677, opened 2022-04-05
Make directories for incremental data load and store configurable

jerhard:
This patch allows to configure the name of the directory where incremental data is loaded from and stored to. This is just a quick-fix that helps for benchmarking.


sim642 (commented):
I think this applies to `master` as well, so to avoid the branches deviating more it should be rebased there first.
Also I guess this PR now accidentally contains some unintended commits due to me having force pushed `interactive` moments before.

jerhard:
>I think this applies to master as well, so to avoid the branches deviating more it should be rebased there first.

The change is now based on master.

PR #680, opened 2022-04-07
Fix incorrect relative paths when parsing preprocessed files

sim642:
This should fix the issue @karoliineh found, where Goblint provided incorrect relative paths on zstd due to directory changes in the compilation database.

It depends on the corresponding hook to be added to CIL, using which we can transform the paths while still knowing the original locations of the files.

### TODO
- [x] Update opam pin after https://github.com/goblint/cil/pull/89.


karoliineh (approved):
Works with GobPie.

PR #682, opened 2022-04-08
Remove `version_map` that was used for `updateCIL`

### LINKED ISSUE #679: Remove `version_map` that is kept for incremental analysis
The `VersionMap` maintained and stored in the incremental analysis seems to be superfluous now, as long as the `Cil.file` that is stored contains the updated identifiers.

jerhard:
This PR removes  the `version_map` that was kept for `updateCil` in the incremental analysis.
 
The mapping was used to to reset the identifiers in unchanged global (e.g. unchanged function). Namely, for a semantically unchanged global, the identifiers were changed back to those of the oldest version of the global that had the same semantics. This step is important, as there still might be results in the solver hashtables that are associated to the id of the first time we encountered the (semantically same) global.  

We got rid of this map as follows :  
- Now, if an unchanged function is encountered during `compareCIL`, it is recorded together with the version of the global in the old file. In `updateCIL`, this old version of the global is then used for resetting the ids of the global. 
 - We already store the CIL-files only after updating the ids. This way, the *old version of the global* that we load already has the id of the oldest version of the unchanged function. 
 
 Closes #679

### DIFF
--- a/src/incremental/compareCIL.ml
+++ b/src/incremental/compareCIL.ml
@@ -102,4 +102,4 @@ let compareCilFiles ?(eq=eq_glob) (oldAST: file) (newAST: file) =
   (* We check whether functions have been added or removed *)
   Cil.iterGlobals newAST (fun glob -> if not (checkExists oldMap glob) then changes.added <- (glob::changes.added));
   Cil.iterGlobals oldAST (fun glob -> if not (checkExists newMap glob) then changes.removed <- (glob::changes.removed));
-  changes
+  changes, newMap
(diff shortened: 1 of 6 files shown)


src/incremental/compareCIL.ml
```diff
@@ -103,3 +109,8 @@ let compareCilFiles ?(eq=eq_glob) (oldAST: file) (newAST: file) =
   Cil.iterGlobals newAST (fun glob -> if not (checkExists oldMap glob) then changes.added <- (glob::changes.added));
   Cil.iterGlobals oldAST (fun glob -> if not (checkExists newMap glob) then changes.removed <- (glob::changes.removed));
   changes
+
+(** Given an (optional) equality function between [Cil.global]s, an old and a new [Cil.file], this function computes a [change_info],
+  which describes which [global]s are changed, unchanged, removed and added.  *)
```
sim642, line 114: I think the indentation workflow wants the comment to be like this:
```suggestion
    which describes which [global]s are changed, unchanged, removed and added.  *)
```

sim642 (approved):

PR #684, opened 2022-04-12
Fix unsound dereferencing of unknown pointer offset chain in DDVerify

sim642:
In https://github.com/goblint/bench/pull/26, I found a case of unsoundness due to our handling of unknown pointers in dereferencing chains. Namely, when intermediate dereferings contain unknown pointers in addition to known ones, we very eagerly go to supertop and forget all the known pointers.

This causes unnecessary unsoundness by not calling operation function pointers that we could know about. It's especially relevant for DDVerify, where the implementation stubs for the operation struct management are actually provided as plain global arrays.

I'm not very sure about the fix though. When dereferencing, there's a check for downcast vs upcast safety, but that only allows known pointers. I'm not sure how it should extend to `NULL` and unknown pointers, but it somehow needs to such that we don't drop all known pointers so easily.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -720,6 +720,7 @@ struct
                   else
                     false
               end
+            | NullPtr | UnknownPtr -> true (* TODO: are these sound? *)
             | _ -> false
           in
           if AD.for_all cast_ok p then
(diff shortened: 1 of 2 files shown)


sim642:
On a small example extracted from zstd, it seems that this fix is also needed to overcome https://github.com/goblint/analyzer/pull/686#issuecomment-1106579058 (on top of everything else there). I'll see if this also is sufficient on the entire zstd.

sim642:
Merging this since it seems to be crucial for zstd soundness.

PR #688, opened 2022-04-14
Klever concurrency safety support

sim642:
Adds support for Klever/LDV specific concurrency safety model functions as described in https://github.com/goblint/bench/issues/3#issuecomment-1056675513. Needed to analyze https://github.com/goblint/bench/pull/27.


src/analyses/libraryFunctions.ml
```diff
@@ -24,7 +24,8 @@ let classify' fn exps =
     `Unknown fn
   in
   match fn with
-  | "pthread_create" ->
+  | "pthread_create"
+  | "pthread_create_N" -> (* Klever *)
```
michael-schwarz, line 28: At least as per https://forge.ispras.ru/issues/7987, this would create an arbitrary number of instances of the thread in question, so one probably would have to modify our handling to thread creation to allow for this.

sim642: That's true, I didn't bother changing all the `threadenter` etc signatures to add an extra argument. There's supposedly also `pthread_join_N`, which has unclear semantics: should it join an unknown number of threads or all the ones created by `pthread_create_N`? I assumed the former for now and didn't implement `pthread_join_N` at all to overapproximate it by not joining anything.

Anyway, I'm not sure whether we even want to merge these hardcoded Klever specifc functions like this. It might be more reasonable to provide some `klever.c` stub with their implementations via pthread instead. Or maybe even tweak Klever models directly to generate benchmarks with fully standard pthread calls only.

sim642: #843 would allow doing this correctly and cleanly.

michael-schwarz:
We now have #843. Is there still interest in this or shall we close it?

michael-schwarz (approved):
Sorry for the delay in reviewing this, somehow dropped of my radar.

PR #693, opened 2022-04-20
Optimize TD3 in server mode by avoiding relifting

sim642:
Since in server mode the solver data isn't unmarshaled from a file but stays in memory the entire time, it is unnecessary for TD3 to relift the entire solution, which can be time-consuming since the hashtables are semi-deeply copied by `relift`.
Instead it suffices to just shallowly copy the solver hashtables themselves, which is still necessary to allow analysis abort in server mode to revert back to previous data without the aborted run having mutated the same data structures.

Surprisingly, this didn't initially work and always caused complete reanalysis in server mode. This was because each reanalysis calls `get_spec ()` and reconstructs the `Spec`, including reinitializing the hashconsing functor and its table.
Since we have to forbid changes to the data structure anyway, we can assume the `Spec` to be constant throughout (by using `lazy`). This ensures that the hashconsing table is reused properly without the need for relifting.

Additionally, I added a bit of debug printing to identify options on which the `Spec` construction itself depends (by self-observation). In the future we could forbid these options from being modified in server mode/incrementally without manually having to maintain a list of these (which could go out of date).

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -477,7 +477,14 @@ module WP =
          * - If we destabilized a node with a call, we will also destabilize all vars of the called function. However, if we end up with the same state at the caller node, without hashcons we would only need to go over all vars in the function once to restabilize them since we have
          *   the old values, whereas with hashcons, we would get a context with a different tag, could not find the old value for that var, and have to recompute all vars in the function (without access to old values).
          *)
-        if loaded && GobConfig.get_bool "ana.opt.hashcons" then (
+        if loaded && S.increment.server then (
+          data.rho <- HM.copy data.rho;
+          data.stable <- HM.copy data.stable;
+          data.wpoint <- HM.copy data.wpoint;
+          data.infl <- HM.copy data.infl;
+          (* data.st <- data.st; *)
+        )
+        else if loaded && GobConfig.get_bool "ana.opt.hashcons" then (
           let rho' = HM.create (HM.length data.rho) in
           HM.iter (fun k v ->
             (* call hashcons on contexts and abstract values; results in new tags *)
--- a/src/util/gobConfig.ml
+++ b/src/util/gobConfig.ml
@@ -28,6 +28,8 @@ exception ConfigError of string
 (* Phase of the analysis (moved from GoblintUtil b/c of circular build...) *)
 let phase = ref 0
 
+let building_spec = ref false
+
 
 module Validator = JsonSchema.Validator (struct let schema = Options.schema end)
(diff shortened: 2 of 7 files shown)


src/solvers/td3.ml
```diff
@@ -477,7 +477,14 @@ module WP =
          * - If we destabilized a node with a call, we will also destabilize all vars of the called function. However, if we end up with the same state at the caller node, without hashcons we would only need to go over all vars in the function once to restabilize them since we have
          *   the old values, whereas with hashcons, we would get a context with a different tag, could not find the old value for that var, and have to recompute all vars in the function (without access to old values).
          *)
-        if loaded && GobConfig.get_bool "ana.opt.hashcons" then (
+        if loaded && S.increment.server then (
+          data.rho <- HM.copy data.rho;
+          data.stable <- HM.copy data.stable;
+          data.wpoint <- HM.copy data.wpoint;
+          data.infl <- HM.copy data.infl;
+          (* data.st <- data.st; *)
```
jerhard, line 485: Should one remove the commented out line and (maybe) add a comment that only the hashtables (changed during the analysis) need to be copied?

src/util/gobConfig.ml
```diff
@@ -28,6 +28,8 @@ exception ConfigError of string
 (* Phase of the analysis (moved from GoblintUtil b/c of circular build...) *)
 let phase = ref 0
 
+let building_spec = ref false
+
```
jerhard, line 32: Is this flag still needed? Where is it read?

src/util/gobConfig.ml
```diff
@@ -303,10 +305,16 @@ struct
...
     drop memo_int; drop memo_bool; drop memo_string; drop memo_list
 
-  let get_int    = memo_int.get
-  let get_bool   = memo_bool.get
-  let get_string = memo_string.get
-  let get_list   = memo_list.get
+  let wrap_get f x =
+    (* self-observe options, which Spec construction depends on *)
+    (* if !building_spec then
+      Printf.printf "GET during spec building: %s\n" x; *)
```
jerhard, line 311: Commented out? How about removing it or converting it to a tracing?

michael-schwarz: I think it might not be too bad to have it in the beginning such that we can then limit what can be changed in the server mode, this whitelisting is still completely missing, allowing for all sorts of dangerous changes to be made.

michael-schwarz (approved):

src/util/gobConfig.ml
sim642: I replaced the commented out code with tracing below, so now the option has some meaning for now as well.

PR #696, opened 2022-04-25
Improve `memset` analysis

### LINKED ISSUE #691: Zero-initialization using `memset`
zstd uses `memset` to zero-initialize/reset some allocated structs. For example:
```c
static void ZSTD_initCCtx(ZSTD_CCtx* cctx, ZSTD_customMem memManager)
{
    assert(cctx != NULL);
    ZSTD_memset(cctx, 0, sizeof(*cctx));
    cctx->customMem = memManager;
    cctx->bmi2 = ZSTD_cpuSupportsBmi2();
    {   size_t const err = ZSTD_CCtx_reset(cctx, ZSTD_reset_parameters);
        assert(!ZSTD_isError(err));
        (void)err;
    }
}
```

Currently Goblint completely invalidates the variable under `memset`, destroying all precision in the struct (especially if it's a global or an alloc variable).
Instead, `memset`ting to 0 (if the `memset` length equals the struct size) could be handled better to zero-initialize the abstract value, a la `calloc`.

sim642:
This contains additional attempts to improve race detection precision on zstd after the soundness fixes.

I noticed that certain `memset` calls are accessing an unreasonably large number of memory locations and causing write races there. As a special function, the access analysis calculates the set of _reachable_ addresses from its argument(s) and emits writes to all of them. Notably, if a memset struct contains a pointer to something else (another struct), then a write to the latter is also added. But this is unnecessarily imprecise, because `memset` only changes/writes what it immediately points to and doesn't dereference any pointers inside. In general special functions may do that, but `memset` (and probably many others) don't as generic functions, so instead of _reachable_ addresses it suffices to just emit accesses to _may point to_ addresses (i.e. not go transitive).

Additionally, I tried to fix #691, i.e. add explicit `memset` handling to base analysis as well. My hope is that this avoids introducing at least some unknown pointers whose calls later cause spurious races.

### Changes
1. Hack access analysis to not use reachability for `memset`. This is very ad-hoc for the time being. Ideally each invalidation argument would also specify whether it may access transitively or not.
2. Cherry-pick access analysis type-partitioning unsoundness fix from #695 because a similar (but not the same) issue appears here (no alloc variables in the failing test). Instead, the access analysis incorrectly calculates the type of the `memset` argument as `void` because it's implicitly cast to `void*` by CIL for `memset`. Considering how stupidly the type-partitioning fails, I have serious doubts about fixing it.
3. Implement `memset` handling in base analysis to write a zero value of the corresponding type if `memset` argument is 0 (closes #691).
4. Together with the previous, the special handling in other cases (non-zero argument), only writes top values to direct addresses, not all indirect ones via reachability (analogously to access analysis). Therefore it should also improve `memset` precision on non-zero arguments as well by avoiding all reachable invalidation.

### TODO
- [x] Add special `__builtin_memset` handling as well, because zstd's macro calls it directly.
- [x] Check if this helps on zstd.
- [ ] Apply same improvement to `free` instead of #695?

### DIFF
--- a/src/analyses/accessAnalysis.ml
+++ b/src/analyses/accessAnalysis.ml
@@ -35,6 +35,12 @@ struct
     unsafe := 0
 
   let side_access ctx ty lv_opt (conf, w, loc, e, a) =
+    let ty =
+      if Option.is_some lv_opt then
+        `Type Cil.voidType (* avoid unsound type split for alloc variables *)
+      else
+        ty
+    in
     let d =
       if !GU.should_warn then
         Access.AS.singleton (conf, w, loc, e, a)
--- a/src/analyses/accessAnalysis.ml
+++ b/src/analyses/accessAnalysis.ml
@@ -181,7 +181,7 @@ struct
       (* TODO: per-argument reach *)
       let reach =
         match f.vname with
-        | "memset" -> false
+        | "memset" | "__builtin_memset" -> false
         | _ -> true
       in
       List.iter (access_one_top ctx false reach) (arg_acc `Read);
--- a/src/analyses/accessAnalysis.ml
+++ b/src/analyses/accessAnalysis.ml
@@ -172,8 +172,14 @@ struct
         | Some fnc -> (fnc act arglist)
         | _ -> arglist
       in
-      List.iter (access_one_top ctx false true) (arg_acc `Read);
-      List.iter (access_one_top ctx true  true ) (arg_acc `Write);
+      (* TODO: per-argument reach *)
+      let reach =
+        match f.vname with
+        | "memset" -> false
+        | _ -> true
+      in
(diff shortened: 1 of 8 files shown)


sim642:
On top of https://github.com/goblint/analyzer/pull/695#issuecomment-1108339397, this does improve the zstd situation notably:
```diff
 Summary for all memory locations:
-	safe:         2001
-	vulnerable:    181
-	unsafe:        863
-	-------------------
-	total:        3045
+	safe:         1594
+	vulnerable:    112
+	unsafe:        688
+	-------------------
+	total:        2394
```
Also, this seemed to improve the analysis time from ~40min to ~25min, but I'm not yet sure if it's genuine improvement from less invalidation or something is unsound now.

PR #705, opened 2022-04-27
Add template for Taint Analysis Tutorial

michael-schwarz:
This is a second tutorial (which @stilscher and I gave at the 2022 Venice Retreat on Continuous Verification of Cyber-Physical System) that provides a template for a simple taint analysis, describing what needs to be done where.

As part of this PR, we remove `earlybird` from the `make dev` dependencies, as it is once again lagging behind and also is not really used much by us Goblint developers currently. 

A solution is in the following branch: https://github.com/goblint/analyzer/tree/convey_retreat_venice_topsecret, I also attached a set of [slides](https://github.com/goblint/analyzer/files/8569680/slides_Schwarz_Tilscher_Erhard_GoblintWorkshop.pdf) one may refer people to.

### DIFF
--- a/docs/developer-guide/firstanalysis.md
+++ b/docs/developer-guide/firstanalysis.md
@@ -44,7 +44,7 @@ Goblint could verify that this assertion does hold using interval analysis (`--e
 ## Starting point
 
 We begin with the flawed implementation in **`src/analyses/tutorials/signs.ml`**.
-If you immediately try to run Goblint with the new analysis enabled: `--set "ana.activated[+]" signs`. The result will still be that nothing is verified, so you need to fix all the problems in the code.
+If you immediately try to run Goblint with the new analysis enabled: `--set ana.activated[+] signs`. The result will still be that nothing is verified, so you need to fix all the problems in the code.
 
 It may still be useful to use Goblint's HTML output to [see the result](../user-guide/inspecting.md) of the analysis. This will also include Goblint's base analysis, which is needed to deal with function calls.
 
@@ -55,7 +55,7 @@ We first need to design the abstract domain. It may help if you have read some t
 1. `of_int i` should abstract integers to their best representation in our abstract domain. Our sign domain can distinguish positive, negative and zero values, so do it right!
 2. `gt x y` should answer true if the value represented by `x` is definitely greater than the value represented by `y`. There seems to be a crucial case missing here in the otherwise excellent implementation...
 
-We will represent tha abstract state of the program as a map from variables to the newly created sign domain.
+We will represent the abstract state of the program as a map from variables to the newly created sign domain.
 
 ```ocaml
 module D = MapDomain.MapBot (Basetype.Variables) (SL)
--- a/docs/developer-guide/firstanalysis.md
+++ b/docs/developer-guide/firstanalysis.md
@@ -44,7 +44,7 @@ Goblint could verify that this assertion does hold using interval analysis (`--e
 ## Starting point
 
 We begin with the flawed implementation in **`src/analyses/tutorials/signs.ml`**.
-If you immediately try to run Goblint with the new analysis enabled: `--set ana.activated[+] signs`. The result will still be that nothing is verified, so you need to fix all the problems in the code.
+If you immediately try to run Goblint with the new analysis enabled: `--set "ana.activated[+]" signs`. The result will still be that nothing is verified, so you need to fix all the problems in the code.
 
 It may still be useful to use Goblint's HTML output to [see the result](../user-guide/inspecting.md) of the analysis. This will also include Goblint's base analysis, which is needed to deal with function calls.
 
--- a/docs/developer-guide/firstanalysis.md
+++ b/docs/developer-guide/firstanalysis.md
@@ -79,4 +79,4 @@ For example, have a look at the following program: `tests/regression/99-tutorial
 _Hint:_
 The easiest way to do this is to use the powerset lattice of `{-, 0, +}`.
 For example, "non-negative" is represented by `{0, +}`, while negative is represented by `{-}`.
-To do this, modify `SL` by using `SetDomain.Make` instead of `Lattice.Flat` and reimplementing the two functions using `singleton` and `for_all`.
+To do this, modify `SL` by using `SetDomain.FiniteSet` (takes a `struct` with a list of finite elements as second parameter) instead of `Lattice.Flat` and reimplementing the two functions using `singleton` and `for_all`.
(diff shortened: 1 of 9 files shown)


michael-schwarz:
Failing OS X test seems to be unrelated:
`2 test(s) failed: ["02/75 memset", "04/70 memset_indirect_nr"]`

sim642:
> Failing OS X test seems to be unrelated:
> `2 test(s) failed: ["02/75 memset", "04/70 memset_indirect_nr"]`

I hope this fixes them on master: https://github.com/goblint/analyzer/commit/8935b25686d0906e2760b425c499cf0940d95b76. I guess MacOS has some fortification enabled by default.

docs/developer-guide/firstanalysis.md
```diff
@@ -79,4 +79,4 @@ For example, have a look at the following program: `tests/regression/99-tutorial
 _Hint:_
 The easiest way to do this is to use the powerset lattice of `{-, 0, +}`.
 For example, "non-negative" is represented by `{0, +}`, while negative is represented by `{-}`.
-To do this, modify `SL` by using `SetDomain.Make` instead of `Lattice.Flat` and reimplementing the two functions using `singleton` and `for_all`.
+To do this, modify `SL` by using `SetDomain.FiniteSet` (takes a `struct` with a list of finite elements as second parameter) instead of `Lattice.Flat` and reimplementing the two functions using `singleton` and `for_all`.
```
sim642, line 82: I don't recall having an issue with `SL.top` when solving the signs tutorial. Is it a problem on the existing signs tutorial tests or some other programs? If the latter, then maybe it should be covered by a test.

make.sh
```diff
@@ -80,7 +80,7 @@ rule() {
     ;; dev)
       eval $(opam env)
       echo "Installing opam packages for development..."
-      opam install -y utop ocaml-lsp-server ocp-indent ocamlformat ounit2 earlybird
+      opam install -y utop ocaml-lsp-server ocp-indent ocamlformat ounit2
```
sim642, line 83: The debugging documentation refers specifically to `make dev` for installing earlybird. That should be updated as well then.

src/analyses/tutorials/taint.ml
```diff
@@ -0,0 +1,129 @@
...
+
+(** "Fake" variable to handle returning from a function *)
+let return_varinfo = dummyFunDec.svar
+
+module Spec : Analyses.MCPSpec with module D = Lattice.Unit and module C = Lattice.Unit =
+struct
+  include Analyses.DefaultSpec
+
+  let name () = "taint"
+  module D = Lattice.Unit (* TODO: Change such that you have a fitting local domain; also update in the signature above *)
```
sim642, line 25: Why have it in the signature at all?

tests/regression/99-tutorials/03-taint_simple.c
```diff
@@ -0,0 +1,40 @@
+// SKIP PARAM: --set "ana.activated[+]" taint --set "exp.extraspecials[+]" readInt --set "exp.extraspecials[+]" printInt  --disable warn.imprecise
+
+int readInt() __attribute__((__taint_source__));
+void printInt(int x,int y ,int z) __attribute__((__taint_sink__));
```
sim642, line 4: There shouldn't be need to add the functions to `extraspecials` because they don't have definitions anyway.

sim642: Also, why even invent own functions for it? Could just declare additional attributes on `scanf`/`printf`. Or even hardcode them directly in `special`.

src/analyses/tutorials/taint.ml
michael-schwarz: to have a space that people can easily edit. It's non obvious that one has to overwrite something that is defined somewhere else if you are not familiar with the code base.

tests/regression/99-tutorials/03-taint_simple.c
michael-schwarz: `scanf` and `printf` would be what one would use for a proper analysis, but here we want the persons taking the tutorial to solve it without having to think of pointers and (even worse) format strings.

src/analyses/tutorials/taint.ml
sim642: Not sure what you mean. There's a module constraint on the `Spec` signature above, but it's not really necessary to have that. It can be a fully abstract `MCPSpec`, just like in any analysis, and they wouldn't have to change the definition of `module D` in two places, that's all.

tests/regression/99-tutorials/03-taint_simple.c
sim642: The format strings are irrelevant though, there's no need to check them or anything. Just consider all arguments sources/sinks.
Alternatively the standard `getchar` and `putchar` could be a simpler choice then.

docs/developer-guide/firstanalysis.md
michael-schwarz: I think it was one of the existing programs. Anyway, I think that this solution is the nicer one anyway.

tests/regression/99-tutorials/03-taint_simple.c
michael-schwarz: I now use `getchar` and `putchar`, but also left `printInt` to indicate that one can also define custom sinks. Adding to `extraspecials` is more of a cosmetic thing to get rid of ugly warnings about missing implementations (without completely ignoring all `error` messages).

src/analyses/tutorials/taint.ml
michael-schwarz: True, the entire sharing constraint is unnecessary!

michael-schwarz: Yes, I think it would only be this case, so probably it causes more confusion when it is there than when it is not.

tests/regression/99-tutorials/03-taint_simple.c
sim642: `getchar` and `putchar` should have `LibraryFunctions` specifications though, so they shouldn't print missing implementation errors without `extraspecials` either.

michael-schwarz: doesn't look like they do, should I add them here or in a separate PR?

sim642: Oh, `putchar` does, but `getchar` doesn't. Just add it here, such that the extraspecials wouldn't be needed to hide warnings.

sim642 (approved):

PR #709, opened 2022-04-27
Fix incremental warnings with reluctant destabilization

### LINKED ISSUE #708: Warnings not removed with reluctant destabilization and incremental postsolving
@vesalvojdani reported that with both reluctant destabilization and incremental postsolving enabled, warnings from changed functions are not removed when the issue is fixed.

The underlying problem is likely that reluctant destabilization only destabilizes the function return node (if changed), but not its entry node (as with non-reluctant destabilization). Therefore the old body of the function remains stable and notably superstable, so warnings are reused during incremental postsolving.

### Solutions
The possible solutions are:
1. Use `delete_marked` on more TD3 solver data structures, notably `superstable`, to remove warnings from the old copy.

    This should suffice because the lack of `delete_marked` on new data structures is probably just an oversight from when they were added.

3. Destabilize old non-entry nodes (not just return node), so they are removed from `superstable`.

sim642:
Closes #708.

I tried the fix on the one test I added to confirm it works, but didn't experiment on more incremental tests with reluctant destabilization enabled.

### Changes
1. Fix old warnings not being removed by additionally deleting old nodes from superstable.
2. Fix new warnings not being added for reluctantly unchanged functions by additionally querying those unknowns from postsolver.

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -540,6 +540,9 @@ module WP =
 
       start_event ();
 
+      (* reluctantly unchanged return nodes to additionally query for postsolving to get warnings, etc. *)
+      let reluctant_vs: S.Var.t list ref = ref [] in
+
       if GobConfig.get_bool "incremental.load" then (
         let c = S.increment.changes in
         List.(Printf.printf "change_info = { unchanged = %d; changed = %d; added = %d; removed = %d }\n" (length c.unchanged) (length c.changed) (length c.added) (length c.removed));
@@ -854,6 +857,7 @@ module WP =
               )
               else (
                 print_endline "Destabilization not required...";
+                reluctant_vs := x :: !reluctant_vs
               )
             ) old_ret;
 
@@ -1111,7 +1115,7 @@ module WP =
 
       let module Post = PostSolver.MakeIncrList (MakeIncrListArg) in
 
-      Post.post st vs rho;
+      Post.post st (!reluctant_vs @ vs) rho;
 
       print_data data "Data after postsolve";
 
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -852,6 +852,9 @@ module WP =
                 destabilize x;
                 HM.replace stable x ()
               )
+              else (
+                print_endline "Destabilization not required...";
+              )
             ) old_ret;
 
           print_endline "Final solve..."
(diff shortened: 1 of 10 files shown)


sim642:
The now-failing incremental test added by @vesalvojdani actually leads to a bigger problem. There it appears that a warning is removed when it should not be, because the new version of the function also has the same problem. But we won't really be able to take the old warnings at the old nodes and somehow move them to new nodes, since we aren't aware of any correspondence (and there might even not be any).

So the real issue is that warnings from the new version of the function do not appear if reluctant destabilization detects that the return state doesn't change and avoids destabilizing it. I believe we would also have a problem if the old function had no warning, but the new version does, while the return state remains unchanged.

This is due to the unusual structure of reluctant destabilization solving: we solve some intermediate part of the constraint system and the destabilization doesn't lead all the way to the queried return node of the main function. Hence the latter remains superstable and interactive postsolving never does any work: the reluctantly solved function isn't re-evaluated for postsolving at all, so no warnings, no accesses and no verification of it.

sim642:
> I believe we would also have a problem if the old function had no warning, but the new version does, while the return state remains unchanged.

I added a third test also for this direction.


> This is due to the unusual structure of reluctant destabilization solving: we solve some intermediate part of the constraint system and the destabilization doesn't lead all the way to the queried return node of the main function. Hence the latter remains superstable and interactive postsolving never does any work: the reluctantly solved function isn't re-evaluated for postsolving at all, so no warnings, no accesses and no verification of it.

To fix this I added a `reluctant_vs` list, where reluctantly unchanged unknowns are added. This list of unknowns is passed to the postsolver as additional queried variables, which forces the postsolver to directly iterate from those points. This should do all the postsolving tasks for the functions that were solved directly by reluctant destabilization.
On the last two tests it at least fixes the issue.

sim642:
Interestingly, I already had part of this fix already in https://github.com/goblint/analyzer/pull/647/commits/7ad1f541d98a7d47f664662295709a77d56645fe of #647 "just in case". This might've been why I was surprised by the issue: I had seen it under some conditions and fixed it, but it didn't yet lead to a fix like #713, so probably wasn't reviewed.

stilscher (approved):

vesalvojdani (approved):
Works on the race-synthetic examples.

PR #712, opened 2022-04-29
Add `realloc` support

### LINKED ISSUE #701: Complete `realloc` support
In `LibraryFunctions` we have the following classification for `realloc`:
https://github.com/goblint/analyzer/blob/313793681ce98d1ea2db0e5f853615dd3050095f/src/analyses/libraryFunctions.ml#L62-L66
Surprisingly, for some reason `Realloc` is not handled in base analysis at all, so the default invalidation kicks in, wrecking all we knew about the allocated blob.

### TODO
- [x] Handle `Realloc` in base analysis.
    - [x] Create blob with new size.
    - [x] Copy data from old blob to new blob.
    - [ ] Free old blob.
    - [x] Use mallocWrapper analysis to get a `varinfo` for the new blob.
    - [x] Handle `sem.malloc.fail`.
- [x] Handle `Realloc` in access analysis.
- [x] Handle `Realloc` in region analysis.

sim642:
Closes #701.

Also includes a couple of blob and pointer casting changes that were necessary to get chrony's `array.c` to work soundly (i.e. not going to `VD.top`) while not breaking any other tests (particularly with `calloc`).

### TODO
- [x] Tasks from #701.
- [x] Figure out what to do about `cast_addr` changing alloc variable type from `void`. Had to disable to get chrony `array.c` to work (otherwise a cast to `char*` screwed up the type and any following casts to a larger struct type were considered not ok, so supertop got introduced).
- [x] Adapt `realloc` accesses to `free` type after #695 merge.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -715,7 +715,9 @@ struct
           let cast_ok = function
             | Addr (x, o) ->
               begin
-                match Cil.getInteger (sizeOf t), Cil.getInteger (sizeOf (get_type_addr (x, o))) with
+                let at = get_type_addr (x, o) in
+                if M.tracing then M.tracel "evalint" "cast_ok %a %a %a\n" Addr.pretty (Addr (x, o)) CilType.Typ.pretty (Cil.unrollType x.vtype) CilType.Typ.pretty at;
+                match Cil.getInteger (sizeOf t), Cil.getInteger (sizeOf at) with
                 | Some i1, Some i2 -> Cilint.compare_cilint i1 i2 <= 0
                 | _ ->
                   if contains_vla t || contains_vla (get_type_addr (x, o)) then
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2304,23 +2304,38 @@ struct
         | _ -> st
       end
     | `Realloc (p, size) ->
-      let new_heap_var = AD.from_var (heap_var ctx) in
-      let p_addr =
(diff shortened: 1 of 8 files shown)


sim642:
The small `cast_addr` change of not changing `varinfo` types exposes a profound amount of trouble. So many things seem to be based on types (which we don't have for allocated memory) instead of on the structure of the abstract value. These mismatching for alloc variables (and possibly other cases) cause countless operations to just fall all the way back to their default supertop-returning case and transitively spreading those supertops everywhere else as well.

sim642:
I got all tests to pass again by hacking the `varinfo` type-change hack even further: avoid it when casting to `char*` because that's a special pointer cast by the standard and used for byte-based access to memory (like chrony's `array.c` is doing).

src/analyses/base.ml
```diff
@@ -2298,6 +2310,39 @@ struct
...
+        | Some lv ->
+          let ask = Analyses.ask_of_ctx ctx in
+          let p_rv = eval_rv ask gs st p in
+          let p_addr =
+            match p_rv with
+            | `Address a -> a
+            (* TODO: don't we already have logic for this? *)
+            | `Int i when ID.to_int i = Some BI.zero -> AD.null_ptr
+            | `Int i -> AD.top_ptr
+            | _ -> failwith "realloc p_addr"
```
sim642, line 2324: Apparently chrony manages to hit this failure somehow (excessive supertops still?), so I might have to just use `AD.top_ptr` as the fallback.

PR #713, opened 2022-04-30
Incremental: (Cheap!) Full Reachability to Prune 

### LINKED ISSUE #710: Warnings not removed for unreachable code (contexts) with incremental post-solving
@stilscher and I discovered the following problem: Superstable contains too many things which causes spurious warnings to remain after an incremental run in cases where code is no longer reachable after the patch.

The fundamental issue is that we do not destabilize (and don't want to) into function calls (of unchanged functions) that no longer happen (or do at least no longer happen in a given context). Hence the unknowns corresponding to such code remain superstable and the warnings are re-used.

I added an example to this branch where a call to a function is removed so an entire function becomes uncalled: https://github.com/goblint/analyzer/tree/interactive_postsolver_pruning
It behaves as expected when one disables the incremental post-solver.

The same issue happens also when only certain calling contexts go away.

Short of a full reachability or horrible hacks where one destabilizes every unknown that is no longer queried at calls, it does not seem obvious how to fix this.

One way that I am currently investigating (suggested by Helmut) is doing a lightweight full reachability where we record for each unknown the set of unknowns its last evaluation depended on. This way, one does not have to execute all the right-hand sides for reachability. For those things that are reachable & superstable, I can reuse the old warnings and accesses, for those things that are reachable and not superstable, I then need to execute the right-hand sides as I currently do.

This is reasonable if most time in the post-solver is spent executing right-hand sides instead of doing reachability.

michael-schwarz:
This implements a cheap form of reachability that works on a set of `dep` collected during solving (and heavy-weight reachability).

Something that is a bit odd:
- If a side-effect occurs from `x` to `y`, it still is necessary to record that `x` depends on `y`, to avoid those globals that only receive side-effects but are never read from being pruned. This has the effect that globals that are only side-effected to do not depend on anything, which I am not quite sure is the right thing here.

This stronger form of reachability can then be used to additionally filter warnings and accesses to replay: Only those that are superstable and still reachable need to be replayed.

I tested the performance impact on a small program, and this cheap reachability seems to be a lot cheaper than full reachability, meaning that it hopefully is not too expensive also for larger programs:

~~~
TOTAL                           0.676 s
  parse                           0.016 s
  convert to CIL                  0.010 s
  mergeCIL                        0.008 s
  analysis                        0.642 s
    createCFG                       0.007 s
      handle                          0.003 s
      iter_connect                    0.003 s
        computeSCCs                     0.002 s
    global_inits                    0.013 s
    solving                         0.612 s
      S.Dom.equal                     0.004 s
      postsolver                      0.183 s
        potentially_incremental_reach       0.176 s
        cheap_full_reach                0.001 s
    warn_global                     0.002 s
      access                          0.001 s
Timing used
~~~

There are still three incremental tests that fail, all of which use restarting. I'll try to have a look, but I am not so familiar with it: maybe @sim642 has an idea what could go wrong there?

Closes #710

### DIFF
--- a/src/solvers/postSolver.ml
+++ b/src/solvers/postSolver.ml
@@ -7,6 +7,7 @@ module type S =
 sig
   module S: EqConstrSys
   module VH: Hashtbl.S with type key = S.v
+  module VS: BatSet.S with type elt = S.v
 
   val init: unit -> unit
   val one_side: vh:S.Dom.t VH.t -> x:S.v -> y:S.v -> d:S.Dom.t -> unit
@@ -16,15 +17,16 @@ end
 
 (** Functorial postsolver for any system. *)
 module type F =
-  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) ->
-  S with module S = S and module VH = VH
+  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) (VS:  BatSet.S with type elt = S.v) ->
+    S with module S = S and module VH = VH and module VS = VS
 
 (** Base implementation for postsolver. *)
 module Unit: F =
-  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) ->
+  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) (VS: BatSet.S with type elt = S.v) ->
   struct
     module S = S
     module VH = VH
+    module VS = VS
     let init () = ()
     let one_side ~vh ~x ~y ~d = ()
     let one_constraint ~vh ~x ~rhs = ()
@@ -32,10 +34,11 @@ module Unit: F =
   end
 
 (** Sequential composition of two postsolvers. *)
-module Compose (PS1: S) (PS2: S with module S = PS1.S and module VH = PS1.VH): S with module S = PS1.S and module VH = PS1.VH =
+module Compose (PS1: S) (PS2: S with module S = PS1.S and module VH = PS1.VH and module VS = PS1.VS): S with module S = PS1.S and module VH = PS1.VH and module VS = PS1.VS =
 struct
   module S = PS1.S
   module VH = PS1.VH
+  module VS = PS1.VS
 
   let init () =
     PS1.init ();
@@ -53,9 +56,9 @@ end
 
 (** Postsolver for pruning solution using reachability. *)
 module Prune: F =
-  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) ->
+  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) (VS: BatSet.S with type elt = S.v) ->
   struct
-    include Unit (S) (VH)
+    include Unit (S) (VH) (VS)
 
     let finalize ~vh ~reachable =
       if get_bool "dbg.debug" then
@@ -68,9 +71,9 @@ module Prune: F =
 
 (** Postsolver for verifying solution in demand-driven fashion. *)
 module Verify: F =
-  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) ->
+  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) (VS: BatSet.S with type elt = S.v)  ->
   struct
-    include Unit (S) (VH)
+    include Unit (S) (VH) (VS)
 
     let init () =
       Goblintutil.verified := Some true
@@ -98,9 +101,9 @@ module Verify: F =
 
 (** Postsolver for enabling messages (warnings) output. *)
 module Warn: F =
-  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) ->
+  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) (VS: BatSet.S with type elt = S.v)  ->
   struct
-    include Unit (S) (VH)
+    include Unit (S) (VH) (VS)
 
     let old_should_warn = ref None
 
@@ -114,9 +117,9 @@ module Warn: F =
 
 (** Postsolver for save_run option. *)
 module SaveRun: F =
-  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) ->
+  functor (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) (VS: BatSet.S with type elt = S.v)  ->
   struct
-    include Unit (S) (VH)
+    include Unit (S) (VH) (VS)
 
     let finalize ~vh ~reachable =
       (* copied from Control.solve_and_postprocess *)
@@ -172,9 +175,10 @@ end
 module MakeIncr (PS: IncrS) =
 struct
   module S = PS.S
+  module VS = PS.VS
   module VH = PS.VH
 
-  let post xs vs vh =
+  let post xs vs vh dep =
     if get_bool "dbg.verbose" then
       print_endline "Postsolving\n";
 
@@ -211,13 +215,30 @@ struct
       if M.tracing then M.trace "postsolver" "one_constraint %a %a\n" S.Var.pretty_trace x S.Dom.pretty rhs;
       PS.one_constraint ~vh ~x ~rhs
     in
-    List.iter one_var vs;
+    (Stats.time "potentially_incremental_reach" (List.iter one_var)) vs;
 
-    PS.finalize ~vh ~reachable;
+    match dep with
+    | None -> PS.finalize ~vh ~reachable:reachable;
+    | Some dep -> (
+        (* Perform reachability on whole constraint system, but cheaply by using logged dependencies *)
+        (* This only works if the other reachability has been performed before, so dependencies created only during postsolve are recorded *)
+        let reachable' = VH.create (VH.length vh) in
+        let rec one_var' x =
+          if not (VH.mem reachable' x) then (
+            VH.replace reachable' x ();
+            match VH.find_option dep x with
+            | Some vs -> VS.iter one_var' vs
+            | None -> ()
+          )
+        in
+        (Stats.time "cheap_full_reach" (List.iter one_var')) vs;
+        PS.finalize ~vh ~reachable:reachable';
+      )
+      ;
     Goblintutil.postsolving := false
 
-  let post xs vs vh =
-    Stats.time "postsolver" (post xs vs) vh
+  let post xs vs vh dep =
+    Stats.time "postsolver" (post xs vs vh) dep
 end
 
 (** List of postsolvers. *)
@@ -226,8 +247,9 @@ sig
   (* Specify S and VH here to constrain all postsolvers to use the same. *)
   module S: EqConstrSys
   module VH: Hashtbl.S with type key = S.v
+  module VS: BatSet.S with type elt = S.v
   (* Auxiliary module type because first-class module types cannot contain module constraints. *)
-  module type M = S with module S = S and module VH = VH
+  module type M = S with module S = S and module VH = VH and module VS = VS
 
   val postsolvers: (module M) list
 end
@@ -246,6 +268,7 @@ module MakeIncrList (Arg: MakeIncrListArg) =
 struct
   module S = Arg.S
   module VH = Arg.VH
+  module VS = Arg.VS
 
   let postsolver_opt: (module Arg.M) option =
     match Arg.postsolvers with
@@ -256,7 +279,7 @@ struct
       in
       Some (List.reduce compose postsolvers)
 
-  let post xs vs vh =
+  let post xs vs vh dep =
     match postsolver_opt with
     | None -> ()
     | Some (module PS) ->
@@ -267,7 +290,7 @@ struct
       end
       in
       let module M = MakeIncr (IncrPS) in
-      M.post xs vs vh
+      M.post xs vs vh dep
 end
 
 (** Make complete (non-incremental) postsolving function from list of postsolvers.
@@ -292,13 +315,14 @@ sig
 end
 
 (** List of standard postsolvers. *)
-module ListArgFromStdArg (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) (Arg: MakeStdArg): MakeListArg with module S = S and module VH = VH =
+module ListArgFromStdArg (S: EqConstrSys) (VH: Hashtbl.S with type key = S.v) (VS: BatSet.S with type elt = S.v)  (Arg: MakeStdArg): MakeListArg with module S = S and module VH = VH and module VS = VS =
 struct
   open Arg
 
   module S = S
   module VH = VH
-  module type M = S with module S = S and module VH = VH
+  module VS = VS
+  module type M = S with module S = S and module VH = VH and module VS = VS
 
   let postsolvers: (bool * (module F)) list = [
     (should_prune, (module Prune));
@@ -311,5 +335,5 @@ struct
     postsolvers
     |> List.filter fst
     |> List.map snd
-    |> List.map (fun (module F: F) -> (module F (S) (VH): M))
+    |> List.map (fun (module F: F) -> (module F (S) (VH) (VS): M))
 end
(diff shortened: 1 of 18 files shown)


michael-schwarz:
There are more flaws than I thought: Upon reanalysis everything not in the main thread seems to be pruned away (!?)

src/solvers/postSolver.ml
```diff
@@ -211,13 +215,30 @@ struct
...
+        let rec one_var' x =
+          if not (VH.mem reachable' x) then (
+            VH.replace reachable' x ();
+            match VH.find_option dep x with
+            | Some vs -> VS.iter one_var' vs
+            | None -> ()
+          )
+        in
+        (Stats.time "cheap_full_reach" (List.iter one_var')) vs;
+        PS.finalize ~vh ~reachable:reachable';
```
sim642, line 235: I didn't yet dive into this PR, but here's an idea which might minimize the amount of changes (and possibly require _none_ to the postsolver itself): do the `reachable'` computation based on `dep` directly in TD3 and use that as `init_reachable`.

Currently there are two possibilities for `init_reachable` in TD3:
1. Empty, so full everything is re-evaluated (so no incrementality).
2. `superstable`, so there's maximal reuse (which is too much for the issue in question).

But I think any set between the two would also be a completely fine choice and the cheaply computed reachability (or I guess the complement of it) should be a subset of `superstable`.

michael-schwarz: > do the `reachable'` computation based on `dep` directly in TD3 and use that as `init_reachable`.

Interesting idea. However, I am a bit skeptical about this, as the point precisely is that during the computation of `reachable'` one has to descend into parts of the constraint system that TD3 does not have to consider. I think for this to work, we would have to store a set of all transitive(!) dependencies of an unknown with it during the non-incremental initial run and marshal that with the rest, such a set might be huge for larger programs so I am a bit worried that maintaining it might outweigh the savings from not having to descend during the computation of `reachable'`

michael-schwarz:
> There are more flaws than I thought: Upon reanalysis everything not in the main thread seems to be pruned away (!?)

The problem was everyone's favorite thing, hashconsing: I re-lifted only the keys in `dep` but the values need to be re-lifted too.

src/solvers/postSolver.ml
sim642: This `reachable'` computation is only based on `vs` and `dep`, both of which are directly provided by TD3 here. Seems to me like it could directly be copied to TD3 to do the same thing without this `one_var'` having to consider anything else that TD3 uses (`superstable`, etc).

michael-schwarz: A crucial detail is that it needs to happen only after the post-solver has run as otherwise any global the receives values only during post-solving would be pruned away.
The other thing is that `reachable'` should not be supplied as `init_reachable`, as it includes also all new things, and the normal post-solver still needs to iterate over that to generate warnings/accesses.

sim642: > A crucial detail is that it needs to happen only after the post-solver has run as otherwise any global the receives values only during post-solving would be pruned away.

If a global receives a value during postsolving, then it goes into the postsolver's `reachable` and hence should not be pruned.
Looks like here the `dep`-based `reachable'` completely replaces the `reachable`, which the postsolver is computing based on actual evaluations of the right-hand sides, so accounts correctly for all the `Goblintutil.postsolving` differences.


> The other thing is that `reachable'` should not be supplied as `init_reachable`, as it includes also all new things, and the normal post-solver still needs to iterate over that to generate warnings/accesses.

Hence my comment above about the complement. It should be possible to filter `superstable` using this `reachable'` (keep only those in both, giving a subset of `superstable`) and use that as `init_reachable`. Then all the new things would automatically be excluded since they aren't superstable, but we prevent the unnecessary parts of superstable from being reused.

sim642:
I suppose this should also fix the second problem (and the corresponding test) from #647, right? If so, then the test should be cherry-picked here.

src/solvers/td3.ml
```diff
@@ -1036,6 +1044,13 @@ module WP =
           )
 
         let finalize ~vh ~reachable =
+          (* replay superstable messages from unknowns that are still reachable *)
+          if incr_verify then (
+            HM.iter (fun (l:S.v) m ->
+                if HM.mem reachable l then Messages.add m
```
sim642, line 1050: Seems like `var_messages` should also be pruned by `IncrPrune` if this check is necessary. Otherwise these old messages still accumulate and are marshaled, but never printed again.

src/solvers/td3.ml
```diff
@@ -1077,6 +1080,22 @@ module WP =
             in
             VH.add w y d (* intentional add *)
           )
+
+        let finalize ~vh ~reachable =
+          (* retrigger superstable side writes from unknowns that are still reachable *)
+          if incr_verify then (
+            HM.iter (fun x w ->
+                if HM.mem reachable x then
```
sim642, line 1088: Same here, `IncrPrune` should already be doing this. If it doesn't happen, then maybe the list of postsolvers needs to be reordered such that `IncrPrune.finalize` is called before this one.

sim642 (commented):
The reason I'm thinking about the `init_reachable` possibility and simplifications related to it, is to find a solution which minimizes the necessary changes to the postsolvers (and especially the incremental ones in TD3). It's risky to make so big changes here last minute, considering the subtle interactions and bugs in the current version, which has gotten a lot more testing.

src/solvers/td3.ml
```diff
@@ -154,6 +156,9 @@ module WP =
       let abort_verify = GobConfig.get_bool "solvers.td3.abort-verify" in
       let prev_dep_vals = HM.create 10 in
 
+      (* Tracks dependencies between an unknown and the things it depends on AND has side-effects to *)
+      let dep = data.dep in
```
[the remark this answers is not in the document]
michael-schwarz: It seems like side-effects are not always recorded into `side_infl`? Is there some specific reason for that?

michael-schwarz:
Following the suggestions by @sim642 , I have now massively simplified it by passing in the intersection of `superstable` and `reachable'` as `init_reachable`, meaning the postsolvers can remain unchanged.

src/solvers/td3.ml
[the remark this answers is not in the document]
sim642: Well, before this `side_infl` was only populated during postsolving for marshaling, but wasn't needed to collect during the solving itself.
Of course that can be changed, but I'm wondering if that's even necessary since leaves (side effect targets with no right-hand sides) themselves don't have warnings, which are emitted during their rhs evaluation. So there shouldn't even be anything to reuse from them in `var_messages` and `rho_write`.

src/solvers/td3.ml
```diff
@@ -1036,6 +1062,13 @@ module WP =
           )
 
         let finalize ~vh ~reachable =
+          (* replay superstable messages from unknowns that are still reachable *)
+          if incr_verify then (
+            HM.iter (fun (l:S.v) m ->
+                Messages.add m
+              ) var_messages;
+          );
+
```
sim642, line 1071: Does everything still work if these are also moved from `finalize` back to `init`? There's another such change in `IncrWrite`.

src/solvers/postSolver.ml
```diff
@@ -211,7 +211,7 @@ struct
       if M.tracing then M.trace "postsolver" "one_constraint %a %a\n" S.Var.pretty_trace x S.Dom.pretty rhs;
       PS.one_constraint ~vh ~x ~rhs
     in
-    List.iter one_var vs;
+    (Stats.time "potentially_incremental_reach" (List.iter one_var)) vs;
```
sim642, line 214: Is there any point to this timing any more since it's inside `postsolver` timing anyway? Or at least the name should probably be something different now, like just `iter`?

src/solvers/td3.ml
```diff
@@ -980,17 +994,41 @@ module WP =
...
+          (* This only works if the other reachability has been performed before, so dependencies created only during postsolve are recorded *)
+          let reachable' = HM.create (HM.length rho) in
+          let rechable_and_superstable = HM.create (HM.length rho) in
+          let rec one_var' x =
+            if (not (HM.mem reachable' x)) then (
+              if HM.mem superstable x then HM.replace rechable_and_superstable x ();
+              HM.replace reachable' x ();
+              match HM.find_option dep x with
+              | Some vs -> VS.iter one_var' vs
+              | None -> ()
```
sim642, line 1022: `Option.may`

src/solvers/postSolver.ml
michael-schwarz: I renamed it.

src/solvers/td3.ml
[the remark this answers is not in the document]
michael-schwarz: None of the tests seem to fail but I am still a bit skeptical about using `side_infl`: it is explicitly cleared before we start the solver, which strikes me as very odd.

It seems that we would run the risk of removing globals that are only side-effected to from superstable things (but remain reachable) if we remove the things that side-effects are done to from `dep` (though I could not construct an example thus far).

~~~OCaml
        (* reachability will populate these tables for incremental global restarting *)
        HM.clear side_dep;
        HM.clear side_infl;
~~~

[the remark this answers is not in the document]
michael-schwarz: Maybe the better solution is to rename `dep` to `uses` or something that indicates that these are the unknowns that `x` reads or has side-effects to?

michael-schwarz:
> I suppose this should also fix the second problem (and the corresponding test) from https://github.com/goblint/analyzer/pull/647, right? If so, then the test should be cherry-picked here.

Done

src/solvers/td3.ml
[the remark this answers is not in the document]
sim642: > None of the tests seem to fail but I am still a bit skeptical about using `side_infl`: it is explicitly cleared before we start the solver, which strikes me as very odd.

Right, the previous use of `side_infl` was just for restarting, which is all done before starting the solving itself. So to minimize memory usage and not keep around data during the solving, which will be recollected during postsolving anyway.

> It seems that we would run the risk of removing globals that are only side-effected to from superstable things (but remain reachable) if we remove the things that side-effects are done to from `dep` (though I could not construct an example thus far).

Hmm, you may have some point here, but if that's indeed a problem, then it's broader than just this PR. Maybe it becomes a problem if at least two incremental changes are made (so there's an intermediate run, which both loads and saves). Then I suppose that might lose the original data and not be able to perform restarting on the final execution correctly?

Since all the testing of incrementality has always been just one change, such an issue has never had a chance to reveal itself. I guess @stilscher's benchmarks that do longer chains of incremental changes might also expose these.

michael-schwarz:
I now went with a solution where we do not add the things that we cause side-effects to to `dep`, but instead also query `side_infl`, which we turn never reset as its values supposedly need to persist across executions. The other question is whether in the post-solver that iterates over the part that might have changed , ` sides_infl x` should be reset to the empty set when executing the RHS for `x` to make sure that old things that no side-effects are now caused to are pruged from `sides_infl`...

src/solvers/td3.ml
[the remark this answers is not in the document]
sim642: So maybe this sort of structure is needed indeed and `uses` name with fewer mixup opportunities.

Now I'm wondering if it's even necessary to collect this `uses` during the actual solving or again just during postsolving. The postsolver sees all the changed/new uses and all the `reachable_and_superstable` should also remain.

sim642:
> The other question is whether in the post-solver that iterates over the part that might have changed , ` sides_infl x` should be reset to the empty set when executing the RHS for `x` to make sure that old things that no side-effects are now caused to are pruged from `sides_infl`...

That sounds like it should be sufficient. The `one_constraint` postsolver hook could be used to do that since it gets executed exactly when the right-hand side has been executed.

michael-schwarz:
Actually, resetting the set in this manner breaks the test case 11/09, so I'd suggest we leave it as currently in this PR for now and merge that.

src/solvers/td3.ml
```diff
@@ -35,6 +35,7 @@ module WP =
       mutable side_infl: VS.t HM.t; (** Influences to side-effected variables. Not normally in [infl], but used for restarting them. *)
       mutable var_messages: Message.t HM.t;
       mutable rho_write: S.Dom.t HM.t HM.t;
+      mutable dep: VS.t HM.t
```
sim642, line 38: Should do the proposed `dep` → `uses` rename then and add documentation to this field.

src/solvers/td3.ml
```diff
@@ -50,12 +51,13 @@ module WP =
...
       rho_write = HM.create 10;
+      dep = HM.create 10;
     }
 
     let print_data data str =
       if GobConfig.get_bool "dbg.verbose" then
-        Printf.printf "%s:\n|rho|=%d\n|stable|=%d\n|infl|=%d\n|wpoint|=%d\n|side_dep|=%d\n|side_infl|=%d\n"
-          str (HM.length data.rho) (HM.length data.stable) (HM.length data.infl) (HM.length data.wpoint) (HM.length data.side_dep) (HM.length data.side_infl)
+        Printf.printf "%s:\n|rho|=%d\n|stable|=%d\n|infl|=%d\n|wpoint|=%d\n|side_dep|=%d\n|side_infl|=%d\n|dep|=%d\n"
+          str (HM.length data.rho) (HM.length data.stable) (HM.length data.infl) (HM.length data.wpoint) (HM.length data.side_dep) (HM.length data.side_infl) (HM.length data.dep)
```
sim642, line 60: This is annoyingly duplicated below at `print_solver_stats`, which for consistency should also have this printout.

src/solvers/td3.ml
```diff
@@ -980,17 +986,39 @@ module WP =
...
           HM.replace side_infl x (VS.add y (try HM.find side_infl x with Not_found -> VS.empty));
       end
       in
 
+      let reachable_and_superstable =
+        if incr_verify then
+          (* Perform reachability on whole constraint system, but cheaply by using logged dependencies *)
+          (* This only works if the other reachability has been performed before, so dependencies created only during postsolve are recorded *)
+          let reachable' = HM.create (HM.length rho) in
+          let rechable_and_superstable = HM.create (HM.length rho) in
```
sim642, line 1006: Typo `rechable`.

src/solvers/td3.ml
```diff
@@ -980,17 +986,39 @@ module WP =
...
 
         let one_side ~vh ~x ~y ~d =
+          (* Also record side-effects caused by post-solver *)
           HM.replace side_dep y (VS.add x (try HM.find side_dep y with Not_found -> VS.empty));
           HM.replace side_infl x (VS.add y (try HM.find side_infl x with Not_found -> VS.empty));
       end
       in
 
+      let reachable_and_superstable =
+        if incr_verify then
```
sim642, line 1002: I'm thinking maybe we should have an additional option for whether to conservatively use the entire `superstable` for this or have this additional `uses`-reachability phase in between.

sim642:
> Actually, resetting the set in this manner breaks the test case 11/09, so I'd suggest we leave it as currently in this PR for now and merge that.

Would be good to have a TODO in there somewhere about clearing the sets. Otherwise we'll forget unless we immediately figure it out.

src/solvers/td3.ml
michael-schwarz: The way it is implemented now, it is actually the `dep` from the TD3 paper though

sim642: Oh, right, yes!

src/solvers/td3.ml
```diff
@@ -980,17 +986,39 @@ module WP =
...
+          let reachable_and_superstable = HM.create (HM.length rho) in
+          let rec one_var' x =
+            if (not (HM.mem reachable' x)) then (
+              if HM.mem superstable x then HM.replace reachable_and_superstable x ();
+              HM.replace reachable' x ();
+              Option.may (VS.iter one_var') (HM.find_option dep x);
+              Option.may (VS.iter one_var') (HM.find_option side_infl x)
+            )
+          in
+          (Stats.time "cheap_full_reach" (List.iter one_var')) vs;
```
sim642, line 1015: Although there isn't a git merge conflict, this needs to be adapted for #709. Hopefully the only change is having to iterate over `reluctant_vs` here as well.

Not sure how easy it would be to construct an incremental test, where the lack of this iteration would expose an issue.

sim642 (approved):

michael-schwarz:
@stilscher: Attention, this contains breaking changes to the config options!

PR #715, opened 2022-05-01
Invalidate variables based on annotations of asm statements

bottbenj:
Instead of simply ignoring `asm` statements use their annotations to invalidate variables that are potentially modified by them.

- https://github.com/goblint/analyzer/commit/e3af268163594212468a490abf9318c10ee9a803 are the changes required to make this compile after the changes in https://github.com/goblint/cil/pull/92 that are required for this analysis. This commit will have to be replace in the future after i get to properly creating the CFG for `asm goto` statements.
- https://github.com/goblint/analyzer/commit/92b89d58d9652272bb5bb99d6fda4a1b494286d4 introduces an event to invalidate variables and makes the two existing invalidates in base use it.
- https://github.com/goblint/analyzer/commit/e3af268163594212468a490abf9318c10ee9a803 the actual code to collect which variables to invalidate for an asm

### TODO
- [ ] use the invalidate event outside of base
- [ ] proper handling of `asm goto`
- [ ] more tests
- [ ] Update goblint-cil opam pin to https://github.com/goblint/cil/pull/92
- ???

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2401,6 +2401,24 @@ struct
     (* D.join ctx.local @@ *)
     ctx.local
 
+  let asm ctx _ an =
+    let inv_globs, inv_exps = match an with
+      | None ->
+        (* asm basic instruction *)
+        (not (get_bool "sem.asm.basic-preserve-globals"), [])
+      | Some (outs, ins, clobbers) ->
+        (* advanced asm instructions *)
+        (
+          List.mem "memory" clobbers && not (get_bool "sem.asm.memory-preserve-globals"),
+          List.append (if get_bool "sem.asm.readonly-inputs" then [] else List.map (fun (_, _, exp) -> exp, InvalidateTransitive) ins)
+            (List.map (fun (_, c, lval) -> Lval lval, match String.find c "+" with 
+               | _ -> InvalidateSelfAndTransitive
+               | exception Not_found -> InvalidateSelf) outs)
+        )
+    in
+    (if get_bool "sem.asm.enabled" then invalidate_core ~ctx inv_globs inv_exps);
+    ctx.local
+
   let event ctx e octx =
     let st: store = ctx.local in
     let ask = ask_of_ctx ctx in
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1899,34 +1899,34 @@ struct
     in
     List.concat_map do_exp exps
 
-  let invalidate ?ctx ask (gs:glob_fun) (st:store) (exps: exp list): store =
+  type invalidate_mode = InvalidateSelf | InvalidateTransitive | InvalidateSelfAndTransitive [@@deriving show]
+
+  let invalidate_core ~ctx inv_globs inv_exps =
+    let exps_transitive = List.filter_map (function (_, InvalidateSelf) -> None | (exp, _) -> Some exp) inv_exps in
+    let lvals_self = List.filter_map (function (_, InvalidateTransitive) -> None | (Lval lval, _) -> Some lval | _ -> None) inv_exps in
+    let exps_self = List.filter_map (function (_, InvalidateTransitive) | (Lval _, _) -> None | (exp, _) -> Some exp) inv_exps in
+    let exps_global = if inv_globs then foldGlobals !Cilfacade.current_file (fun acc global ->
+        match global with
+        | GVar (vi, _, _) when not (is_static vi) && not (BaseUtil.is_excluded_from_invalidation vi) ->
+          mkAddrOf (Var vi, NoOffset) :: acc
+        (* TODO: what about GVarDecl? *)
+        | _ -> acc
+      ) [] else [] in
+    let addr_list = collect_funargs (ask_of_ctx ctx) ctx.global ctx.local (exps_transitive @ exps_global) @ (List.map (eval_lv (ask_of_ctx ctx) ctx.global ctx.local) lvals_self) in
+    if M.tracing then (
+      M.trace "invalidate" "invalidate_core transitive: [%a]\n" (d_list "," CilType.Exp.pretty) exps_transitive;
+      M.trace "invalidate" "invalidate_core self (exp): [%a]\n" (d_list "," CilType.Exp.pretty) exps_self;
+      M.trace "invalidate" "invalidate_core self (lval): [%a]\n" (d_list "," CilType.Lval.pretty) lvals_self;
+      M.trace "invalidate" "invalidate_core global: [%a]\n" (d_list "," CilType.Exp.pretty) exps_global;
+    );
+    ctx.emit (Events.Invalidate {addr_list})
+
+
+  let invalidate ~ctx ask (gs:glob_fun) (st:store) (exps: exp list): store =
     if M.tracing && exps <> [] then M.tracel "invalidate" "Will invalidate expressions [%a]\n" (d_list ", " d_plainexp) exps;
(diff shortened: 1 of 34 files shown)


src/framework/constraints.ml
```diff
@@ -675,7 +675,7 @@ struct
       | Entry f        -> tf_entry var edge prev_node f
       | Ret (r,fd)     -> tf_ret var edge prev_node r fd
       | Test (p,b)     -> tf_test var edge prev_node p b
-      | ASM (_, _, _)  -> tf_asm var edge prev_node (* TODO: use ASM fields for something? *)
+      | ASM _          -> tf_asm var edge prev_node (* TODO: use ASM fields for something? *)
```
sim642, line 678: The relevant data should be taken from the `ASM` constructor and passed as actual arguments instead of looking at `ctx.edge` in the transfer function implementation.

michael-schwarz:
Sorry we haven't provided feedback yet, we have an upcoming paper deadline on Friday AOE and will do so after for all your PRs :smile:

src/util/options.schema.json
```diff
@@ -1108,6 +1108,38 @@
...
+          "title": "sem.asm",
+          "description": "Inline Assembly Analysis options",
+          "type": "object",
+          "properties": {
+            "enabled": {
+              "title": "sem.asm.enabled",
+              "description": "",
+              "type": "boolean",
+              "default": true
+            },
```
sim642, line 1122: What would be the use case for ever disabling asm semantics (and whatever that would mean)?
Wouldn't enabling the other options, which make additional safety assumptions about asm statements already give the old behavior, where possible changes from asm statements were completely ignored?

src/analyses/base.ml
```diff
@@ -1899,34 +1899,34 @@ struct
...
-      List.exists BaseUtil.is_excluded_from_invalidation (AD.to_var_may x)
-    in
-    let invalids' = List.filter (fun (x,_,_) -> not (is_fav_addr x)) invalids in
-    if M.tracing && exps <> [] then (
-      let addrs = List.map (Tuple3.first) invalids' in
-      let vs = List.map (Tuple3.third) invalids' in
-      M.tracel "invalidate" "Setting addresses [%a] to values [%a]\n" (d_list ", " AD.pretty) addrs (d_list ", " VD.pretty) vs
-    );
-    set_many ?ctx ask gs st invalids'
+    invalidate_core ~ctx false (List.map (fun exp -> (exp, InvalidateTransitive)) exps);
```
sim642, line 1928: The `inv_globs` argument for `invalidate_core` should probably be named, otherwise it's confusing what this `false` means here.

src/analyses/base.ml
```diff
@@ -2430,6 +2437,20 @@ struct
...
+        let nv =  VD.invalidate_value ask t v in
+        (a, t, nv)
+      in
+      let invalids = List.map invalidate_address addr_list in
+      if M.tracing && addr_list <> [] then (
+        let addrs = List.map (Tuple3.first) invalids in
+        let vs = List.map (Tuple3.third) invalids in
+        M.tracel "invalidate" "Setting addresses [%a] to values [%a]\n" (d_list ", " AD.pretty) addrs (d_list ", " VD.pretty) vs
+      );
+      set_many ~ctx ask ctx.global ctx.local invalids
```
sim642, line 2453: What's the reason for converting this into an event instead of simply extracting the functions for all globals invalidation etc inside the base analysis itself?
All the calls to `invalidate_core` look awkward, because they have to delegate outside to the events system to perform these after the actual transfer function completes, but in all three cases the transfer function just returns `ctx.local` as the updated state. This seems like an unnecessarily roundabout way to extract common logic.

jerhard: >What's the reason for converting this into an event instead of simply extracting the functions for all globals invalidation etc inside the base analysis itself?

The idea of making invalidation an event based mechanism was that other analyses could be informed and then do some proper handling. One could e.g. imagine that an relational analysis that handles referenced variables might need to know when some variables are potentially invalidated.

sim642: I'm just afraid that in light of #719 with the need to distinguish non-transitive vs transitive accesses also for reads/frees and writes other than invalidation, this won't be general enough.

It's also quite base analysis specific by using its `ValueDomain.Addr`. Other analyses like apron or var_eq, which currently also do similar invalidation, aren't using base analysis's address abstraction, but have their own ways of working with invalidated lvalues/rvalues.

jerhard: >I'm just afraid that in light of https://github.com/goblint/analyzer/issues/719 with the need to distinguish non-transitive vs transitive accesses also for reads/frees and writes other than invalidation, this won't be general enough.

The idea for invalidate events would be that the `Base` analysis collects all lvals that have to be invalidated. This way, other analyses would not have to implement the collection of transitively reachable lvals.

>It's also quite base analysis specific by using its ValueDomain.Addr

One could add a utility function that takes elements from the `ValueDomain.Addr` and translates them back to the `lval` they are on (without offset information, in case the offset is not concrete).

sim642: > The idea for invalidate events would be that the `Base` analysis collects all lvals that have to be invalidated. This way, other analyses would not have to implement the collection of transitively reachable lvals.

They don't currently reimplement it either, but just use the `ReachableFrom` query where needed. Since #590 those queries are cached, so multiple analyses asking for the same reachability comes at no extra cost.

We will probably have to see how this plays together with the general non-transitive vs transitive accessing changes from #719 and #720.

sim642: I thought about this more and how it compares to #720 and the `Access` event. It might make sense to have the `Invalidate` event after all because:
1. `Access` event is emitted for all writes, not just invalidating ones. Although this could be changed by adding an extra field to that event.
2. `Access` event is emitted only during multithreaded mode, because it's all about race detection. But these invalidations also need to happen in singlethreaded mode. Although this could also be changed (and accesses from singlethreaded mode wouldn't race with anything).

But given that access analysis (and its event) is really about race detection, it might be better to keep it that way.

tests/regression/56-asm/03-basic-discard-global.c
```diff
@@ -0,0 +1,12 @@
+#include <assert.h>
+
+int x = 0;
+
+int main()
+{
+  asm("nop");
+
+  assert(x == 0); // UNKNOWN
```
jerhard, line 9: It would make sense to have additional tests for the configurable options you implement, e.g. you might add a test for what's currently `sem.asm.basic-preserve-global`.

tests/regression/56-asm/60-goto-issue.c
```diff
@@ -0,0 +1,45 @@
...
+	do                            \
+	{                             \
+		asm goto(x);                \
+		asm("");                    \
+	} while (0)
+
+static int arch_static_branch(const int key, const int branch)
+{
+	asm_volatile_goto("1:"
+										".byte " __stringify(BYTES_NOP5) "\n\t" JUMP_TABLE_ENTRY
```
jerhard, line 31: This section might benefit from better indentation.

tests/regression/56-asm/60-goto-issue.c
```diff
@@ -0,0 +1,45 @@
+// see https://github.com/goblint/cil/issues/81
```
jerhard, line 1: Is this test currently only about parsing and not crashing Goblint? Maybe add an explicit comment about that, and the current limitations.

michael-schwarz:
Unassigning myself as I currently do not have the bandwith to invest time here.

michael-schwarz:
Thank you again for this PR. One of the student teams is now working on implementing this more generally for the Goblint practical. Therefore, I am closing this PR.

PR #717, opened 2022-05-03
Interactive: fix expression locations

### LINKED ISSUE #689: Locations for accesses in guards cover entire statement
When accesses occur in loop guards, the access location covers the entire body of the statement. It makes it very unreadable inside GobPie for real-world examples like this:

```c
#include <pthread.h>
#include <stdio.h>

int myglobal;

void *t_fun(void *arg) {
  if (myglobal > 0) {
    printf("Stupid!");
    printf("Stupid!");
    printf("Stupid!");
    printf("Stupid!");
    printf("Stupid!");
    printf("Stupid!");
  }
  return NULL;
}

int main(void) {
  pthread_t id;
  myglobal=myglobal+1;
  pthread_join (id, NULL);
  return 0;
}
```

I tried a fix in 076660b13eed8bd448cefd5a2579a922773341a6, but it's not clear if that causes issues with our use of locations for other things. Are there cases when we'd want the location to cover the entire statement?

sim642:
Closes #689.
This is only a problem on `interactive` due to #533 only being there.

The ideal solution would be to have `UpdateCil` also update expression locations (not just one location for each statement), but that requires a lot more work to distinguish the statements/instructions and store multiple locations for some.

Also, since there's no way to identify specific CFG edges (unlike nodes, which have IDs from either statements or the function), there's no easy way to just update the edge locations incrementally either. PR #689 and this make edge locations largely redundant (at least for messages, maybe something uses them non-incrementally directly).

Therefore this PR just contains a refinement of @vesalvojdani's original fix: simply use expression locations for everything.
This makes it impossible to get and warn about an entire `if`/`while`/`switch` statement, rather than just its expression, but I don't remember any case where we want to do that. If that turns up, a more involved fix is necessary to distinguish the two.

### DIFF
--- a/.semgrep/cil.yml
+++ b/.semgrep/cil.yml
@@ -6,6 +6,7 @@ rules:
       - pattern: Cil.typeOfInit
       - pattern: Cil.typeOffset
       - pattern: Cil.get_stmtLoc
+      - pattern: Cil.get_instrLoc
     paths:
       exclude:
         - cilfacade0.ml
--- a/src/transform/evalAssert.ml
+++ b/src/transform/evalAssert.ml
@@ -106,16 +106,16 @@ module EvalAssert = struct
         let rec instrument_instructions = function
           | i1 :: ((i2 :: _) as is) ->
             (* List contains successor statement, use location of successor for values *)
-            let loc = get_instrLoc i2 in
+            let loc = get_instrLoc i2 in (* TODO: why not using Cilfacade.get_instrLoc? *)
             i1 :: ((instrument i1 loc) @ instrument_instructions is)
           | [i] when unique_succ ->
             (* Last statement in list *)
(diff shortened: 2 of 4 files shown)


vesalvojdani (approved):
This works on the examples I tried. I'm not sure about the TODOs. We don't have tests for assertion generation, do we?

sim642:
> I'm not sure about the TODOs. We don't have tests for assertion generation, do we?

No. I'll leave it for anyone working on assert generation to figure out in the future, what is intended. Because that's more of an AST transformation and deals with `block` and `If`, maybe it intends to use the entire statement's location? Or maybe it doesn't matter because it looks like it puts locations into the AST it generates, where they're irrelevant for printout anyway.
Either way, it's just good to have it noted down, because the semgrep rule for `Cil.get_stmtLoc` didn't catch them since they're used under `open Cil`.

Issue #719, opened 2022-05-04
Redesign of `LibraryFunctions` specification


sim642:
Recently during interactive benchmarking, it has really bothered me how messy and inflexible our `LibraryFunctions` specifications are

# Problems
1. There are two completely separate specification mechanisms in `LibraryFunctions`:
    1. `classify` is for classifying some functions into particular categories for uniform handling by analyses: https://github.com/goblint/analyzer/blob/fb54fa5bc8dabd43077877f4a1794e32093a4312/src/analyses/libraryFunctions.ml#L8-L98
    2. `invalidate_actions` is for specifying what base invalidation and race detection do with arguments: https://github.com/goblint/analyzer/blob/fb54fa5bc8dabd43077877f4a1794e32093a4312/src/analyses/libraryFunctions.ml#L192-L518
2. There is duplication between the two and to avoid confusion there _must_ be duplication. If a function is in `classify`, but not `invalidate_actions`, then analyses can handle it according to the classification, but base analysis still prints an error about missing function definition and starts invalidating things.
3. Specifying `classify` cases using pattern matching is relatively verbose. Five lines of code are used to match one function and its arguments: https://github.com/goblint/analyzer/blob/fb54fa5bc8dabd43077877f4a1794e32093a4312/src/analyses/libraryFunctions.ml#L27-L31
4. `invalidate_actions` specifications like `readsAll`, `writesAll`, etc. are too broad to specify behavior sufficiently precisely. For example:
    1. `Write` is also used to decide, which arguments need to be spawned as threads. This is almost always unnecessary because C standard library functions don't spawn threads (except for C11 threads, which we don't support yet anyway).
    2. `Free` was added in #695 to be able to distinguish memory freeing from `Write`.
    3. All accesses are assumed to be transitive/deep, i.e. they follow pointers. Again, this is almost never the case for C standard library functions (e.g. `free` doesn't recursively first free data pointed to inside the outer struct, it only deallocates the direct memory and it's the programmer's job to free inner data structures first).
    4. Shallow accessing is currently enforced by another duplicate match in an analysis (separately): https://github.com/goblint/analyzer/blob/fb54fa5bc8dabd43077877f4a1794e32093a4312/src/analyses/accessAnalysis.ml#L190-L197
    5. Shallow (direct) vs deep (indirect) accessing cannot be specified per-argument.
5. More `invalidate_actions` specifications use inconsistent argument numbering:
[...]

PR #720, opened 2022-05-04
OCaml DSL for specifying `LibraryFunctions`

sim642:
This is an experiment to address #719 by designing an OCaml DSL, which allows specifying `LibraryFunctions` such that all the problems mentioned in the issue are avoided and all the goals are fulfilled.

The DSL implementation is a mixture of the following:
* So-called first-class patterns inspired by ppxlib's `Ast_pattern`: https://github.com/ocaml-ppx/ppxlib/blob/main/src/ast_pattern.ml. The three-type-parameter continuation-based internals of the patterns are hidden from actually specifying functions.
* A custom list-like type (using the standard list constructors to allow abusing normal OCaml list notation), but as a GADT to allow composing the first-class patterns in a type-safe manner. The GADT internals are hidden from actually specifying functions.
* A set of helper functions to construct such specifications in a readable way.

The DSL in `LibraryDsl` is just a convenient way to construct newly-introduced `LibraryDesc` objects, which combine the corresponding classify function, invalidate actions (with more precision) and other function-related attributes (e.g. being thread unsafe #723 or invalidating all globals).

The usage is (hopefully sufficiently) extensively documented both in the OCaml code and in the developer documentation.
The usage can be seen at the beginning of `LibraryFunctions` and `LibraryDslTest`.

### TODO
- [x] Document `LibraryDesc`.
- [x] Document `LibraryDsl`.
- [x] Implement shallow vs deep distinction in access analysis correctly. Function arguments are always directly read on call, but the access depth actually concerns MPT/reach. So there are actually _three_ levels of access for reading function arguments.
- [x] Fix new "Function definition missing" errors.
- [x] Separate specifications lists:
    - [x] C standard library
    - [x] POSIX standard library
    - [x] Pthread
    - [x] Linux kernel
- [ ] ~~Add selection for separate specifications lists.~~ To be done in the future, when all old-style specifications have been ported.


src/analyses/libraryDsl.ml
```diff
@@ -0,0 +1,104 @@
+open LibraryDesc
+
+(** First-class patterns for arguments matching.
+    @see <https://github.com/ocaml-ppx/ppxlib/blob/main/src/ast_pattern.ml> for inspiration from ppxlib. *)
+module Pattern =
+struct
```
jerhard, line 6: The intention of the `Pattern` module is a bit hard to understand at first sight, and the link does not really offer an explanation. Adding an additional comment might be useful.

unittest/analyses/libraryDslTest.ml
```diff
@@ -0,0 +1,34 @@
+open OUnit2
+
+let memset_desc: LibraryDesc.t = LibraryDsl.(
+  unknown [drop "dest" [w]; drop "ch" []; drop "count" []]
+)
+
+let pthread_mutex_lock_desc: LibraryDesc.t = LibraryDsl.(
+  special [__' [r]] @@ fun e -> Lock { lock = e; try_ = false; write = true; return_on_success = false; } (* TODO: not right *)
```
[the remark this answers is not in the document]
sim642: This is just in a test, not `LibraryFunctions` itself at this point. I think what I meant there was that if it were in a non-test, then the `try_` value would depend on `sem.lock.fail`. I updated the comment accordingly.

src/analyses/mutexAnalysis.ml
```diff
@@ -131,8 +131,9 @@ struct
           if not (Lockset.is_bot ctx.local) then
             let ls = Lockset.filter snd ctx.local in
             let write = match kind with
-              | `Write | `Free -> true
-              | `Read -> false
+              | Write | Free -> true
+              | Read -> false
+              | Spawn -> false (* TODO: nonsense? *)
```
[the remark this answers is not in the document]
sim642: It does happen all the time though, because the third argument of `pthread_create` is a spawn access to the function's own `varinfo` (which is different from the function pointer).
The question here is, what, if anything, does a protecting lockset of a function mean. And that's different from reading the function pointer. It's like reading the function itself to execute it or something. AFAIK it doesn't really matter because we don't privatize the function itself.

src/analyses/libraryDsl.ml
sim642: True, although this module is completely internal to the implementation. `libraryDsl.mli` doesn't expose it at all and documents everything that is intended to be used. Also there's additional `docs/` about it all.

unittest/analyses/libraryDslTest.ml
```diff
@@ -0,0 +1,34 @@
...
+)
+
+let rand_desc: LibraryDesc.t = LibraryDsl.(
+  unknown ~attrs:[ThreadUnsafe] []
+)
+
+let tests =
+  "libraryDslTest" >::: [
+
+  ]
```
jerhard, line 34: What is the purpose of this file, if the list of `tests` is empty?

sim642: Heh, I think at some point I planned to write some tests for checking if `special` and `accesses` of a description return correct things. Right now it just serves as a check that the above descriptions compile, so it's not entirely redundant.

jerhard: It would sense to comment that the definitions in this file are for compile-time checks only.

jerhard (approved):

Issue #703, opened 2022-04-27
Restarting of Access globals as a baseline for the incremental solver


michael-schwarz:
@stilscher discussed this with Helmut yesterday, and at some point I also joined the discussion.

The issue we currently face is that we try to sell the non-restarting of access globals as a contribution (c.f. read-only globals). However, already in the basic setting without the incremental post-solver we are smart about these accesses in that we only really collect them during postsolving, which already is part of the insight we want to sell.

So for comparison, we were thinking about implementing a more naive approach:

- Without this insight that it suffices to collect accesses during postsolving, we would already collect them during the fixpoint iteration.
- Then, to make sure that (at least some) of the warnings can actually go away, we would have to selectively restart those access globals that received side-effects from changed functions (potentially more, but I guess for a practical evaluation it would suffice to do those for now)

If we implement this, this gives us a natural testbed for selective restarting and also shows that just the Seidl, Vogler, Erhard incrementality is not enough, as it (without our insight) requires restarting of access globals, which is not something they do in their paper.

I think this sort of restarting ought to be easy to implement, as we already collect the dependencies during the access analysis anyway. I think to showcase the difference, it would even suffice to reset because that's what is needed conceptually and then still simply collect in the verification phase, as we iterate over the entire system anyway when the incremental post-solver is off.

WDYT @sim642 @vesalvojdani ?

----

And perhaps a stupid side question: In the setting without the incremental post-solver, do we actually reset the access globals to `bot`, or are we truly just accumulating race warnings without it?

sim642:
> However, already in the basic setting without the incremental post-solver we are smart about these accesses in that we only really collect them during postsolving, which already is part of the insight we want to sell.

The crucial difference is that there was no incremental postsolving before, so all the accesses had to be regenerated each time. 

> the Seidl, Vogler, Erhard incrementality is not enough, as it (without our insight) requires restarting of access globals, which is not something they do in their paper.

Prior to incremental postsolving, there were even no unknowns for accesses, but it was all done ad-hoc. So effectively they used to be all restarted (the incrementally loaded run started with no accesses) and all recalculated (by the non-incremental postsolver). So if I understand the idea correctly, it's exactly what they did, but it was all implicit because the access tables were a global OCaml data structure that simply wasn't marshaled.

Or am I misunderstanding something?

> And perhaps a stupid side question: In the setting without the incremental post-solver, do we actually reset the access globals to `bot`, or are we truly just accumulating race warnings without it?

[...]

PR #721, opened 2022-05-04
Option to only destabilize access globals

michael-schwarz:
This adds on option to reset only access globals to bottom. I think that in that case `destabilize_normal` should be enough, there is no need to destabilize into side-effects, right @sim642.

What is very curious is that I have been completely unable to construct an example where the  `incremental.restart.sided.destab-with-sides` would make any sort of difference. Maye this is due to the other `read_only` globals that are not actually tracking accesses that the access analysis seems to collect for some reason?

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -131,6 +131,8 @@ module WP =
       (* If true, incremental side-effected var restart will only restart destabilized globals (using hack).
          If false, it will restart all destabilized side-effected vars. *)
       let restart_only_globals = GobConfig.get_bool "incremental.restart.sided.only-global" in
+      let restart_only_access = GobConfig.get_bool "incremental.restart.sided.only-access" in
+      let restart_destab_with_sides = GobConfig.get_bool "incremental.restart.sided.destab-with-sides" in
 
       (* If true, wpoint will be restarted to bot when added.
          This allows incremental to avoid reusing and republishing imprecise local values due to globals (which get restarted). *)
@@ -580,7 +582,14 @@ module WP =
           let w = HM.find_default side_dep x VS.empty in
           HM.remove side_dep x;
 
-          if not (VS.is_empty w) && (not restart_only_globals || Node.equal (S.Var.node x) (Function Cil.dummyFunDec)) && not (S.Var.is_write_only x) then (
+          let should_restart =
+            if restart_only_access then
+              S.Var.is_write_only x
+            else
+              (not restart_only_globals || Node.equal (S.Var.node x) (Function Cil.dummyFunDec)) && (not (S.Var.is_write_only x))
+          in
+
+          if not (VS.is_empty w) && should_restart then (
             (* restart side-effected var *)
             restart_leaf x;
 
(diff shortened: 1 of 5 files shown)


src/solvers/td3.ml
```diff
@@ -131,6 +131,8 @@ module WP =
       (* If true, incremental side-effected var restart will only restart destabilized globals (using hack).
          If false, it will restart all destabilized side-effected vars. *)
       let restart_only_globals = GobConfig.get_bool "incremental.restart.sided.only-global" in
+      let restart_only_access = GobConfig.get_bool "incremental.restart.sided.only-access" in
```
sim642, line 134: With more of these options, the nicer thing would be to either have one option with three string values if they can be ordered such that they subsume each other, or have to non-only options (so negated) if all combinations are reasonable.
But we can leave it here as-is, the options need to be cleaned up before #391 gets merged to `master` anyway.

src/solvers/td3.ml
```diff
@@ -580,7 +582,14 @@ module WP =
           let w = HM.find_default side_dep x VS.empty in
           HM.remove side_dep x;
 
-          if not (VS.is_empty w) && (not restart_only_globals || Node.equal (S.Var.node x) (Function Cil.dummyFunDec)) && not (S.Var.is_write_only x) then (
+          let should_restart =
+            if restart_only_access then
+              S.Var.is_write_only x
```
sim642, line 587: The `only_access` name is a bit misleading here, since it's not particular to accesses (it doesn't try to crudely look into the unknowns, unlike the globals vs locals hack below). But the name `only_write_only` would also be very confusing.

src/solvers/td3.ml
```diff
@@ -593,7 +602,10 @@ module WP =
                 if tracing then trace "sol2" "stable remove %a\n" S.Var.pretty_trace y;
                 HM.remove stable y;
                 HM.remove superstable y;
-                destabilize_with_side ~front:false y
+                if restart_destab_with_sides then
+                  destabilize_with_side ~front:false y
+                else
+                  destabilize_normal ~front:false y
```
sim642, line 608: This seems a lot like restarting with 1 fuel (#629) in that it only continues destabilizing without restarting sides. Although here it's a bit differently checked at `side_dep` rather than `side_infl`.

For the purposes of some last-minute benchmarking, this is fine and we can afterwards decide whether to maybe integrate this with #629 or strip out entirely.

tests/incremental/13-restart-write/09-mutex-simple-access.c
```diff
@@ -0,0 +1,23 @@
+#include <pthread.h>
```
sim642, line 1: Maybe this test would be better in 11-restart than 13-restart-write since it's about the general restarting mechanism, not the write-only unknowns one.

sim642: Also, I guess this is the same code and patch as 13-restart-write/01-mutex-simple, but just with different conf? Might be worth mentioning that on top of the file, so it's more obvious than manually having to compare it to all the other variations in 13-restart-write.

sim642 (commented):
Is this supposed to be for #703?

michael-schwarz:
I would propose to merge this now such that Sarah can use it in her benchmark runs, and perform the cleanup tasks w.r.t. to options later.

sim642 (approved):

michael-schwarz:
Sarah's experiments indicate that this is broken somehow :frowning_face:

michael-schwarz:
fixed and merged back into interactive

Issue #702, opened 2022-04-26
Strongly update-able `calloc` blobs for more precise zstd analysis


sim642:
## Problem
Our analysis of zstd ends up with unexpectedly many unknown pointers due to the following sequence of coincidences.

### Thread pool and its queue
In zstd's `POOL_create_advanced`, the thread pool and its queue are allocated on the heap as follows (simplified):
```c
ctx = (POOL_ctx*)ZSTD_customCalloc(sizeof(POOL_ctx), customMem);
ctx->queueSize = queueSize + 1;
ctx->queue = (POOL_job*)ZSTD_customMalloc(ctx->queueSize * sizeof(POOL_job), customMem);
ctx->queueHead = 0;
// ...
```
Here `ctx` and both alloc variables are in the local state. Thanks to #686 the alloc variables escape into global state correctly.

When jobs are added to the queue, their function pointers are nicely joined into the queue's alloc variable, which precisely contains known function pointers and no `NULL` or unknown pointer.

However later in the thread pool threads, they retrieve jobs from the queue using `ctx->queue[ctx->queueHead]`, i.e. `*(ctx->queue + ctx->queueHead)` after CIL normalization. Confusingly, the function pointer in the returned job struct contains an additional unknown pointer, despite the queue's alloc variable not containing one.

### Extensive invalidation
Turns out this is because ctx's alloc variable also has a `NULL` pointer possible for the `queue` field. When pointer arithmetic is applied (and `queueHead` is no longer constant either), the known alloc pointer is nicely kept, but the `NULL` pointer is replaced by an unknown pointer using this code:
https://github.com/goblint/analyzer/blob/313793681ce98d1ea2db0e5f853615dd3050095f/src/analyses/base.ml#L208-L214

Analyzing zstd, the unknown function pointer is absolutely disastrous effect:
1. An unknown function pointer call is analyzed by `special` to invalidate its arguments.
2. In zstd, one of the thread pool jobs is passed an argument, which inside it contains a pointer to the same thread pool `ctx`.
3. The resulting invalidation is extensive and invalidates:
    * `ctx` alloc variable (so an unknown pointer is added to `queue`'s address set directly as well),
    * `queue` alloc variable (so `NULL` and unknown pointers are added to the job's function pointer and argument sets),
    * all the argument structs for all the jobs (holding all of the relevant data) are totally invalidated.

### Culprit
[...]

PR #722, opened 2022-05-04
Malloc uniqueness

TDacik:
This PR tries to increase the precision of the representation of heap-allocated variables. 

* 83e413c Each malloc can now allocate up to _n_ unique heap variables, where _n_ is given as the command-line parameter (by default _n_ is 0, which means that only one non-unique heap variable is allocated for each malloc node -- this is for backward compatibility). Heap variables now also contain a thread identifier of a thread where the allocation happened to distinguish memory allocated in different threads. A heap variable is considered unique if its thread identifier is unique and the number of its alocation is lesser than _n_.

* 8e68873 Heap variables that are unique can be strongly (destructively) updated if the size of a written value matches the size of a blob that represents a memory allocated for the variable.


sim642:
The two issues this at least partially relates to are #186 and #702.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -13,21 +14,61 @@ struct
...
+      else
+        remove malloc counter
+        |> add malloc (count + 1)
+  end
+
+  module Domain = struct
+    include Lattice.Prod (MallocCounter) (PL)
+
+    let join (counter1, node1) (counter2, node2) =
+      (MallocCounter.join counter1 counter2, node1)
```
sim642, line 47: Why this? It violates the lattice properties now.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -13,21 +14,61 @@ struct
...
+
+    let has_wrapper_node (_, wrapper_node) = not @@ PL.is_top wrapper_node
```
sim642, line 49: `wrapper_node` is from a flat lattice of nodes, so it could also be bottom (no malloc wrapper present). This function currently returns `true` then, which doesn't sound right.

src/util/options.schema.json
```diff
@@ -814,6 +814,12 @@
                 "kmalloc", "__kmalloc", "usb_alloc_urb", "__builtin_alloca",
                 "kzalloc"
               ]
+            },
+            unique_address_count: {
```
sim642, line 818: JSON syntax error:
```suggestion
            "unique_address_count": {
```

tests/regression/11-heap/09-no_strong_update_not_unique_thread.c
```diff
@@ -0,0 +1,33 @@
+// PARAM: --set ana.malloc.unique_address_count 2
+
+// Strong updates are not possible in thread because it is not unique.
```
sim642, line 3: How about with mallocFresh analysis also enabled? These alloc variables are created thread-local and never escape, so theoretically they should be strongly updateable.

sim642 (commented):
I guess for me the big question is whether this works for the example in #186 and possibly the first solution steps for #702.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -3,8 +3,9 @@
 open Prelude.Ana
 open Analyses
 open GobConfig
+open ThreadIdDomain
 
-module Spec : Analyses.MCPSpec =
+module Spec () : Analyses.MCPSpec =
```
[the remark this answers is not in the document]
michael-schwarz: IIRC, I think it was needed to avoid accessing some configuration things before the config is parsed.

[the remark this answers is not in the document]
sim642: Ah, I guess it's the `let n` depending on the option for construction of `Lattice.Chain`. I suppose the most consistent thing to do would be to change it to `val n: unit -> int` in `Printable.ChainParams`. We already have it like that for `name`, `bot`, `top` and `arbitrary`. It was probably never done for `Chain`, because it's used very little and for constant things.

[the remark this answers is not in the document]
michael-schwarz: Done!

tests/regression/11-heap/09-no_strong_update_not_unique_thread.c
michael-schwarz: The corresponding variables are still created as globals here, so freshness does not help here.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -48,51 +89,73 @@ struct
...
+        else wrapper_node (* if an interesting callee is called by an interesting caller, then we remember the caller context *)
+      else PL.top () (* if an uninteresting callee is called, then we forget what was called before *)
+    in
+    let caller_ctx = (new_counter, wrapper_node) in
+    let callee_ctx = (new_counter, new_wrapper_node) in
+    [(caller_ctx, callee_ctx)]
 
   let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) : D.t =
-    ctx.local
+    D.join ctx.local au
```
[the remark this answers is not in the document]
michael-schwarz: Refactored

src/analyses/mallocWrapperAnalysis.ml
michael-schwarz: Refactored

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -13,21 +14,61 @@ struct
...
+          failwith "Option ana.malloc.unique_address_count has to be non-negative"
+        else p + 1 (* Unique addresses + top address *)
+
+      let names x = if Int.equal x (n - 1) then "top" else Format.asprintf "%d" x
+
+    end)
+
+  (* Map for counting malloc node visits up to n. *)
+  module MallocCounter = struct
+    include MapDomain.MapBot_LiftTop(PL)(Chain)
```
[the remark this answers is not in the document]
michael-schwarz: the counters are also per thread-id, as thread ids form parts of the context, and in `threadenter` the new thread receives a counter component that is ⊥

michael-schwarz:
I started an sv-comp run to see what sort of impact this has.

michael-schwarz:
As it stands, the PR is unsound for calloc, as the size is not properly calculated there.

michael-schwarz:
> As it stands, the PR is unsound for calloc, as the size is not properly calculated there.

This was some ancient bug where when we modified the `Calloc` variant in `LibraryFunctions` to take two arguments, but did not appropriately pull in the multiplication.

src/analyses/mallocWrapperAnalysis.ml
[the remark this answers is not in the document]
sim642: Right, maybe mention that in a code comment here.

tests/regression/11-heap/09-no_strong_update_not_unique_thread.c
sim642: Oh right, mallocFresh analysis only affects race checking, but not anything value-based, which I suppose it also could. Actually the freshness stuff could become part of escape analysis, but there was some issue at the time which prevented that (considering everything non-escaped to be fresh in general).

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -48,42 +75,60 @@ struct
...
-    ctx.local
-
-  let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
-    ctx.local
+    let counter, wrapper_node = ctx.local in
+    let new_wrapper_node =
+      if Hashtbl.mem wrappers f.svar.vname then
+        match wrapper_node with
+        | `Lifted _ ->  wrapper_node (* if an interesting callee is called by an interesting caller, then we remember the caller context *)
+        | _ ->  (`Lifted ctx.prev_node) (* if an interesting callee is called by an uninteresting caller, then we remember the callee context *)
```
sim642, line 83: Why did this suddenly change from `ctx.node` to `ctx.prev_node`?

tests/regression/11-heap/09-no_strong_update_not_unique_thread.c
michael-schwarz: Yes, in the future, we should definitely make that change. I'd suggest doing it in a different PR though.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -48,42 +75,60 @@ struct
...
+    let _, lnode = ctx.local in
+    (counter, lnode)
+
+  let special (ctx: (D.t, G.t, C.t, V.t) ctx) (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    let desc = LibraryFunctions.find f in
+    match desc.special arglist, f.vname with
+    | Malloc _, _ | Calloc _, _ | Realloc _, _ ->
+      let counter, wrapper_node = ctx.local in
+      (MallocCounter.add_malloc counter ctx.prev_node, wrapper_node)
+    | _ -> ctx.local
```
sim642, line 101: The match on `f.vname` is pointless here.

src/analyses/mallocWrapperAnalysis.ml
```diff
@@ -48,42 +75,60 @@ struct
...
+  let query (ctx: (D.t, G.t, C.t, V.t) ctx) (type a) (q: a Q.t): a Q.result =
+    let counter, wrapper_node = ctx.local in
     match q with
     | Q.HeapVar ->
-      let node = match ctx.local with
-        | `Lifted vinfo -> vinfo
-        | _ -> ctx.node
+      let node = match wrapper_node with
+        | `Lifted wrapper_node -> wrapper_node
+        | _ -> ctx.prev_node
```
sim642, line 119: Same question here, why the change to `ctx.prev_node`?

michael-schwarz:
Before we can merge this, I am still waiting on the results of the sv-comp run: I seem to encounter the issue with `benchexec` freezing up again, so I need to babysit it.

michael-schwarz:
For sv-comp the result once again is a resounding it does not matter, but probably it is one of those things where when used selectively may form the key ingredient somewhere.

Interestingly, performance is also not too bad: the x axis is with setting 0, the y axis is with setting 3

x-0-y-3

sim642:
> the y axis is with setting 3

I tried to find what Frama-C used in SV-COMP for the equivalent limit. According to their manual `-eva-mlevel` sounds to be the option for that, but Frama-C's SV-COMP logs don't indicate anything about that (whereas they indicate a bunch of other precision levels, e.g. loop unrolling being much higher). I suppose in it was the default 0 in SV-COMP then.

Given that the Frama-C manual describes this being useful up to certain loop bounds, it probably won't make a real difference with 3, since most loops aren't that short and thus would end up with one weak alloc variable for all the rest anyway.

This doesn't block this PR, but it might be interesting to take a closer look at ReachSafety-Heap with a much higher limit (based on loops actually occurring in those benchmarks) to see if it gets us further. It's probably a matter for #772 to bump the limit based on `malloc`s in loops.

michael-schwarz:
I think to really benefit in loops, we would have to enable some sort of path-sensitivity, which we currently don't do here.
Yes, may be interesting to run it on the heap benchmarks with a higher limit.

I'll merge this for now though, such that it doesn't become stale.

Issue #716, opened 2022-05-02
Different field access on imprecise address set


sim642:
I continued looking at chrony after #712 and found another little bit of unsoundness: the callback function `name_resolve_handler` is never called.
The problem is that at an intermediate level of callbacks in `sched.c`, the eventual callback `name_resolve_handler` is stored in the intermediate callback's opaque argument, which ends up being an ambiguous set of addresses pointing to structs of different types. This is because in other places the same `sched.c` mechanism is used for something else as well. Later when trying to access a particular field of that, we immediately get `VD.top` since simple structs are implemented using a `MapTop` domain, where looking up the other struct's field gives the default `VD.top`.

I've extracted the problem to https://github.com/goblint/analyzer/tree/ambig-field. A minimized test case is the following: https://github.com/goblint/analyzer/blob/ab70c9b0e1301ad6dd90044cc8c4986139429da1/tests/regression/01-cpa/56-ambig-field.c#L21-L38

### Simple structs
Tracing the program reveals the two structs being joined, giving an empty struct (remember, this is `MapTop`, which is reversed):
```
          %%% get: Singlethreaded mode.
          %%% get: var = mapping {
                           f1 ->   {&foo}
                         }, {&s1} = mapping {
                                        f1 ->   {&foo}
                                      }
          %%% get: Singlethreaded mode.
          %%% get: var = mapping {
                           f2 ->   {&bar}
                         }, {&s2} = mapping {
                                        f2 ->   {&bar}
                                      }
          %%% get: Result: mapping {
                             }
```
Then of course accessing either field on the result returns `VD.top` and completely forgets the known pointers.

### Set structs
One might hope that a more powerful structs domain (`sets`) would help here, but no:
```
          %%% get: Singlethreaded mode.
          %%% get: var = {mapping {
                            f1 ->   {&foo}
                          }}, {&s1} = {mapping {
                                           f1 ->   {&foo}
                                         }}
          %%% get: Singlethreaded mode.
          %%% get: var = {mapping {
                            f2 ->   {&bar}
                          }}, {&s2} = {mapping {
                                           f2 ->   {&bar}
                                         }}
          %%% get: Result: {mapping {
                              f1 ->   {&foo}
                            }, mapping {
                                   f2 ->   {&bar}
                                 }}
        %%% cast: (tests/regression/01-cpa/56-ambig-field.c:34:3-34:43)
  cast {mapping {
          f1 ->   {&foo}
        }, mapping {
               f2 ->   {&bar}
[...]

Issue #434, opened 2021-11-05
Symbolic locksets should invalidate using may-equality information


vesalvojdani:
The current implementation of symbolic lockset track sets of addresses known to be definitely equal to the address expressions that were used when locking. It relies on must-equalities to improve precision, but may-equality should be used when these sets are invalidated. When pointers are written to, it does seem like we rely on may information from the base analysis, but when mutexes are unlocked we do this based on the same must-eq query as for locking. This is unsound.

Additionally, we may want to extract some of the very semantic manipulations from the add/remove operations of the domain into the analysis. I'm not sure how normal it is for domain domains to do such value-dependent computations.

Test case was added for this in f3ba049099005b39098390302737e9748411cbad.

PR #724, opened 2022-05-05
Interactive: improvements for chrony story

sim642:
### Changes
1. Fix var_eq and symb_locks such that they are not confused about `IndexPI` expressions.
2. Add `__goblint_assume_join` special Goblint builtin to threadJoins analysis to annotate thread being joined even if Goblint doesn't consider it unique. This can be used as an annotation after a `pthread_join` loop.
3. Make `realloc` argument read and free accesses non-transitive (we really need #719).
4. Add option `sem.unknown_function.read.args` for also ignoring read accesses from unknown function arguments (`sem.unknown_function.invalidate.args` from #707 only concerns invalidation, i.e. writing).

### DIFF
--- a/src/analyses/apron/apronPriv.apron.ml
+++ b/src/analyses/apron/apronPriv.apron.ml
@@ -988,7 +988,7 @@ struct
     let tids = ask.f (Q.EvalThread exp) in
     if force then (
       if ConcDomain.ThreadSet.is_top tids then
-        failwith "TODO"
+        st (* don't consider anything more joined, matches threadJoins analysis *)
       else (
         (* fold throws if the thread set is top *)
         let (lmust', l') = ConcDomain.ThreadSet.fold (fun tid (lmust, l) ->
--- a/src/analyses/apron/apronPriv.apron.ml
+++ b/src/analyses/apron/apronPriv.apron.ml
@@ -987,11 +987,13 @@ struct
     let w,lmust,l = st.priv in
     let tids = ask.f (Q.EvalThread exp) in
     if force then (
-      if ConcDomain.ThreadSet.is_top tids then
-        st (* don't consider anything more joined, matches threadJoins analysis *)
+      if ConcDomain.ThreadSet.is_top tids then (
+        M.info ~category:Unsound "Unknown thread ID assume-joined, Apron privatization unsound"; (* TODO: something more sound *)
+        st (* cannot find all thread IDs to join them all *)
+      )
       else (
         (* fold throws if the thread set is top *)
-        let tids' = ConcDomain.ThreadSet.diff tids (ask.f Q.MustJoinedThreads) in (* avoid unnecessary imprecision by force joining already must-joined threas, e.g. 46-apron2/04-other-assume-inprec *)
+        let tids' = ConcDomain.ThreadSet.diff tids (ask.f Q.MustJoinedThreads) in (* avoid unnecessary imprecision by force joining already must-joined threads, e.g. 46-apron2/04-other-assume-inprec *)
         let (lmust', l') = ConcDomain.ThreadSet.fold (fun tid (lmust, l) ->
             let lmust',l' = G.thread (getg (V.thread tid)) in
             (LMust.union lmust' lmust, L.join l l')
--- a/src/analyses/apron/apronPriv.apron.ml
+++ b/src/analyses/apron/apronPriv.apron.ml
@@ -991,10 +991,11 @@ struct
         st (* don't consider anything more joined, matches threadJoins analysis *)
       else (
         (* fold throws if the thread set is top *)
+        let tids' = ConcDomain.ThreadSet.diff tids (ask.f Q.MustJoinedThreads) in (* avoid unnecessary imprecision by force joining already must-joined threas, e.g. 46-apron2/04-other-assume-inprec *)
         let (lmust', l') = ConcDomain.ThreadSet.fold (fun tid (lmust, l) ->
             let lmust',l' = G.thread (getg (V.thread tid)) in
             (LMust.union lmust' lmust, L.join l l')
-          ) tids (lmust, l)
+          ) tids' (lmust, l)
         in
         {st with priv = (w, lmust', l')}
       )
--- a/src/analyses/apron/apronPriv.apron.ml
+++ b/src/analyses/apron/apronPriv.apron.ml
@@ -37,7 +37,7 @@ module type S =
     val enter_multithreaded: Q.ask -> (V.t -> G.t) -> (V.t -> G.t -> unit) -> apron_components_t -> apron_components_t
     val threadenter: Q.ask -> (V.t -> G.t) -> apron_components_t -> apron_components_t
 
-    val thread_join: Q.ask -> (V.t -> G.t) -> Cil.exp -> apron_components_t -> apron_components_t
+    val thread_join: ?force:bool -> Q.ask -> (V.t -> G.t) -> Cil.exp -> apron_components_t -> apron_components_t
     val thread_return: Q.ask -> (V.t -> G.t) -> (V.t -> G.t -> unit) -> ThreadIdDomain.Thread.t -> apron_components_t -> apron_components_t
     val iter_sys_vars: (V.t -> G.t) -> VarQuery.t -> V.t VarQuery.f -> unit (** [Queries.IterSysVars] for apron. *)
 
@@ -62,7 +62,7 @@ struct
   let lock ask getg st m = st
   let unlock ask getg sideg st m = st
 
-  let thread_join ask getg exp st = st
+  let thread_join ?(force=false) ask getg exp st = st
   let thread_return ask getg sideg tid st = st
 
   let sync ask getg sideg st reason = st
@@ -270,7 +270,7 @@ struct
     {apr = apr_local'; priv = (p', w')}
 
 
-  let thread_join ask getg exp st = st
+  let thread_join ?(force=false) ask getg exp st = st
   let thread_return ask getg sideg tid st = st
 
   let sync ask getg sideg (st: apron_components_t) reason =
@@ -461,7 +461,7 @@ struct
     let apr_local = remove_globals_unprotected_after_unlock ask m apr in
     {st with apr = apr_local}
 
-  let thread_join ask getg exp st = st
+  let thread_join ?(force=false) ask getg exp st = st
   let thread_return ask getg sideg tid st = st
 
   let sync ask getg sideg (st: apron_components_t) reason =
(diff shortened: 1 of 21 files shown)


src/analyses/threadJoins.ml
```diff
@@ -45,6 +45,19 @@ struct
...
+      if TIDs.is_top threads then
+        D.bot () (* consider everything joined, D is reversed so bot is All threads *)
+      else (
+        (* elements throws if the thread set is top *)
+        let threads = TIDs.elements threads in
+        List.fold_left (fun acc tid ->
+            let joined = ctx.global tid in
+            D.union (D.add tid acc) joined
+          ) ctx.local threads
+      )
```
michael-schwarz, line 60: I have a few questions here:

### Apron Privatizations

I think this does currently conflict with the apron privatizations that do not read from must-joined threads (`ana.apron.priv.must-joined`).

We currently do not join these values in the local per-mutex information, but would exclude them in the privatization then. Or am I missing something?

https://github.com/goblint/analyzer/blob/35f5b9e9ce00222e2221605df4262fe3ab6158a8/src/analyses/apron/apronPriv.apron.ml#L982-L998

Also how would it work with unknown thread ids? We would if this privatization is active have to read from all threads created anywhere?

### Threads created after call to `__goblint_assume_join(...)`

What happens if a thread with the same non-unique thread-id is created again after `__goblint_assume_join(...)`? We would need to remove it from must-joined, right?

What happens if a new thread is created after we called `__goblint_assume_join(...)` on something that was the top thread id. We would (unsoundly) already consider this thread joined?

**I think we should re-consider the handling of top here, think about the interaction with apron privatizations and definitely write tests for these things before merging.**

src/cdomains/exp.ml
```diff
@@ -10,6 +12,8 @@ struct
   (* TODO: what does interesting mean? *)
   let rec interesting x =
     match x with
+    | BinOp ((PlusPI | IndexPI), e1, e2, _) ->
+    interesting e1 (* TODO: what about e2? *)
```
michael-schwarz, line 16: Can we actually say anything useful about equalities involving pointer arithmetic?

src/cdomains/exp.ml
```diff
@@ -290,19 +294,22 @@ struct
     in
     let rec helper exp =
       match exp with
+      | BinOp ((PlusPI | IndexPI), e1, e2, _) ->
+        helper e1 (* TODO: what about e2? add to some Index offset to all? *)
```
michael-schwarz, line 298: Sam here, how confident are we that we are doing something sensible for pointer arithmetic?

src/analyses/varEq.ml
```diff
@@ -519,7 +528,10 @@ struct
       D.B.fold add es (Queries.ES.empty ())
 
   let rec eq_set_clos e s =
-    match e with
+    if M.tracing then M.traceli "var_eq" "eq_set_clos %a\n" d_plainexp e;
+    let r = match e with
+    | BinOp ((PlusPI | IndexPI), e1, e2, _) ->
+      eq_set_clos e1 s (* TODO: what about e2? add to some Index offset to all? *)
```
michael-schwarz, line 534: Can you maybe comment on the impact that this has now? Are we tracking equalities involving pointer arithmetic? If yes, how confident are we that we do useful things?

michael-schwarz (changes requested):

src/analyses/threadJoins.ml
sim642: It's an annotation coming from the user, so if the user makes incorrect assumptions, then the analysis is unsound, yes. We cannot check whether the assumption is always correct (if we could, then we would have a powerful enough analysis to make such annotation by the user redundant).
The point of the annotation is to be placed after a threads joining loop, which in reality ensures that all the threads have been joined. If they haven't, then that's on the user for incorrectly using the annotation.
It covers the extremely common case, where we get tons of spurious races with some cleanup code.

Regarding the top, I don't see the problem because prior to this the joined thread set had a lifted top with the literal meaning "All threads". So if the joined thread set is that top, it means _all_ threads (that could possibly be created or exist) are joined. If there's a problem with that "All threads" and it shouldn't be used, then the safest thing would be to remove that top lifting and make such situation unrepresentable.

I also don't understand how it would cause a problem for the apron analysis, because if it's not handled there, then the apron analysis continues without having joined anything, just like it did before, still accounting for the possibility of values from the threads. That would be an overapproximation, no?
I could also crudely add the handling of the annotation to the apron analysis and simply make it fail, if that seems safer.

michael-schwarz: > The point of the annotation is to be placed after a threads joining loop, which in reality ensures that all the threads have been joined. If they haven't, then that's on the user for incorrectly using the annotation.

Fair enough. However, it seems very counterintuitive that this annotation would extend to threads that are created **AFTER** the annotation and just happen to share abstract thread ids. I think that makes it very hard to reason about as a user, who would place it at a program point to tell Goblint that all threads created thus far are joined, but would not expect this to extend to future copies of the thread being created.

As an extreme case, consider a function that spawns `n` threads, does some work and the  joins all `n` threads again. If we call this function twice (with joined in contexts), in the second call we would assume that at the second invocation all threads it creates are already joined. I don't think that this is particularly helpful.

> I also don't understand how it would cause a problem for the apron analysis, because if it's not handled there, then the apron analysis continues without having joined anything, just like it did before, still accounting for the possibility of values from the threads. 

No, it is added to the set of must joined thread ids and the apron privatization queries this set when locking mutexes no longer reading from the ones associated with must-joined threads. This only works  under the assumption that this must-joined set only contains things where all possible values associated with mutexes and must-joined threads were joined into the thread-local information the apron privatization keeps at the threadjoin, which does no longer happen here.


> If there's a problem with that "All threads" and it shouldn't be used, then the safest thing would be to remove that top lifting and make such situation unrepresentable.

There is. I think the only reason this was not done is for compatibility with the old thread ids.

michael-schwarz: I added an example to illustrate the issue, I hope that helps more than the theoretical discussion above.

sim642: Making threadJoins analysis remove things on `threadspawn` should be a simple change that avoids the main confusion about it affecting future thread creations. And it wouldn't hurt the annotation's intended use case, where the thread isn't recreated anyway.

I'm now quite surprised that threadJoins doesn't already have that logic though. And it's not obvious why it wouldn't be a problem for normal joins, considering we have four different thread ID domains available (plain vs history, with or without `include-node`). Is it assuming something about the thread ID domain that makes this impossible with normal thread joins?
For example, if a loop creates a single thread and joins it in the same iteration, then it could also be considered unique (only a single copy of it exists at a time) and thus must-joined. If I remember correctly, the history thread ID domain doesn't consider that unique, but other notions of thread uniqueness (like the one provided by the old thread uniqueness analysis) do AFAIK.

> I added an example to illustrate the issue, I hope that helps more than the theoretical discussion above.

Thanks! I will have a look at it.



> There is. I think the only reason this was not done is for compatibility with the old thread ids.

As for the "All threads" must-joined set, it might also be avoidable by slightly weakening what I do right now for the annotation: if the thread ID set for the argument contains an unknown thread, then just exclude that and assume-join all the known ones. Hopefully this is sufficient for the chrony case.
What this assume-joining doesn't change though, is that the base analysis still invalidates all the stuff from a `free`, even if the access analysis makes all the exemptions for ignoring those. But that means the invalidation puts unknown pointers and threads all over the flow-insensitive invariant of the global, leaking unknowns everywhere from the beginning, killing any precision we might have had there.

sim642: I also changed the assume join behavior with unknown thread IDs now to not use "All threads" for must-joined, hopefully this gives less surprising behavior.

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -983,22 +983,37 @@ struct
...
+      else (
+        (* fold throws if the thread set is top *)
+        let (lmust', l') = ConcDomain.ThreadSet.fold (fun tid (lmust, l) ->
+            let lmust',l' = G.thread (getg (V.thread tid)) in
+            (LMust.union lmust' lmust, L.join l l')
+          ) tids (lmust, l)
+        in
+        {st with priv = (w, lmust', l')}
+      )
+    )
```
michael-schwarz, line 1001: A perhaps surprising effect here is that the `__goblint_assume_join(...)` annotation may destroy precision here, if things have actually been must-joined before and have then definitely been overwritten locally before (I added an example).

I think here it should be sufficient to iterate over those not already consider must-joined anyway?

sim642: Wouldn't the same problem also happen if you repeat `pthread_join(id);` where the thread ID is unique and can normally be must-joined?

src/analyses/varEq.ml
michael-schwarz: I think we are a bit too eager to declare access as non-racy here, consider the example I added:


~~~C
void cache_entry_addref(struct cache_entry *entry) {
  pthread_mutex_lock(&entry->refs_mutex);
  entry->refs++; // NORACE
  (*entry).refs++; // NORACE
  entry[1].refs++; // RACE
  pthread_mutex_unlock(&entry->refs_mutex);
}
~~~

These accesses at ` entry[1]` are racy, as the protecting mutex for that element is not held (I fixed the loop to only run until 9, so it's not UB). Also passing a `t*` and then using array access on it is a very common idiom in C afaik.

src/analyses/apron/apronPriv.apron.ml
michael-schwarz: Yes, but joining things twice is UB, so I don't really think it matters in that case that we become imprecise (we should in fact produce a gigantic warning if anyone does this!)

I would understand this `__goblint_assume_join(...)` not as something that one would put into the program instead of the actual join, but something one additionally can put to give Goblint the hint that the join must have happened. In this case it is surprising if Goblint suddenly becomes less precise because of it. Or did I misunderstand it?

sim642: > Yes, but joining things twice is UB, so I don't really think it matters in that case that we become imprecise (we should in fact produce a gigantic warning if anyone does this!)

That would be part of #176.


> I would understand this `__goblint_assume_join(...)` not as something that one would put into the program instead of the actual join, but something one **additionally** can put to give Goblint the hint that the join must have happened. In this case it is surprising if Goblint suddenly becomes less precise because of it. Or did I misunderstand it?

You are correct, that it's an additional thing. I can add an extra check about already must-joined threads there to avoid this possibility.
Although actually you would only ever add the assume join if at the actual join Goblint couldn't must-join it, so the assume join would be the only one doing its must join.

src/analyses/varEq.ml
michael-schwarz: I think the assumption we should be making is that 
- 1) pointer arithmetic never leaves the corresponding varinfo (or blob for malloced things),
- 2) and never can cross struct boundaries once one has descended  (because the compiler may have inserted padding), so no adding magic offsets to get from struct member `a` to `a` in the next element of the array unless `a` is the first struct element.
That seems not too strong but also not as radical as setting everything to top. (And should be possible as our addresses  maintain offsets and have abstract values at the granularity of variables anyway).

I think if we assume anything more, we cannot seriously claim we are sound anymore. I personally would be extremely unhappy with Goblint accepting this program as race-free! (How do @vesalvojdani @stilscher @jerhard feel here?)

Should we put this on the agenda at the Gobcon?

michael-schwarz: > Now thinking about bad things, if the argument were an array and we copy the array domain into the function and mutated it, do those writes even propagate back to the caller's array domain?

I'd hope so. We pass a pointer to the array, hence the entire(!) array will go to the callee, we update things there and should then copy the value over in combine?

src/analyses/apron/apronPriv.apron.ml
michael-schwarz: I think that would be nice. I can imagine situations where one desperately tries things until something works and I could see it happening that someone inserts such an annotation ever for a thread where we find it is must-joined.

src/analyses/varEq.ml
sim642: Or maybe this is somehow also related to #434? Because the thing is, this change simply allows the `entry[0]` or `entry[1]` expression to preserve its prefix (previously we gave up on index access of pointer). Index access on array type has been previously supported for a long time and this shouldn't be different: symbolic locking happens on the basis of common prefix between the mutex and the access (and we have at least one paper on this!). So I'm quite surprised how this can suddenly be going wrong.

cc: @vesalvojdani

sim642: > I'd hope so. We pass **a pointer to the array**, hence the entire(!) array will go to the callee, we update things there and should then copy the value over in combine?

I haven't dug into this now, but this is exactly where I'm worried. Sure, if you take a pointer of the array and pass that to a function, it gets passed along as reachable and `combine`d back. But what if you pass the array directly (without making a pointer to it first). In the caller, the local domain contains an array abstraction directly under the array variable name, there isn't an address domain containing an address to the array contents. This is analogous to how an integer variable contains the integer abstraction directly. And when they are passed as arguments, they are copied — there's a copy of the integer's abstraction directly under the callee's argument and copy of the array's abstraction directly under the callee's argument. On `combine`, the internal integer abstraction is simply removed and so would the callee's (modified) array one. Or do we have some special logic somewhere there specifically for arrays? I don't recall at least.

> I personally would be extremely unhappy with Goblint accepting this program as race-free!

Indeed in the non-UB test you added, we should not. It's just the symbolic locks that then might be assuming something stronger.

sim642: I just realized there's a granularity difference at play here as well:
* The `cache_entry` tests in our suite use an array of structs, where each struct has a mutex protecting its data.
* The chrony `DNS_Async_Instance` struct, for which this is needed, is a standalone struct (not one in an array), where the single mutex protects an entire array inside the struct.

Because they are structurally different like that, I'd hope we can actually be sound for the tests you added and the story in chrony. I'll have to dig into how exactly these different granularity locking schemes, and per-field vs per-index symbolic locks are supposed to work, because these cases involve both at the same time (but in different order!).

sim642: The case for chrony is easier than the more general one, where the unsoundness problem came in. I now fixed that using very ad-hoc handling that's still sufficient for chrony.

src/analyses/apron/apronPriv.apron.ml
sim642: Apparently the annotation doesn't help on chrony any more for some reason. I'll look into it.

sim642: Due to #716 in chrony the assume-joined thread ID turns out to be top, so by making it safer and ignoring that, the chrony story collapses again. I don't think we'll be able to fix every single issue that causes unknown thread IDs to occur in chrony.

Maybe a workable alternative is to still assume-join everything with unknown thread ID, but emit an unsoundness message when that happens? Then it would be apparent, when you should do the equivalent thing and remove the annotation (which would have no effect anyway), but allow the user to override our super imprecision if they really know what they're doing.

sim642: As discussed in GobCon, I changed it back to allow assume-join of unknown thread ID to consider everything joined, but a warning is printed. Also when any thread is then created, the must-joined set is made empty and another warning is printed.
This now works for the chrony story and hopefully is reasonable enough behavior.

michael-schwarz (approved):

PR #729, opened 2022-05-06
Interactive: improvements for smtprc story

vesalvojdani:
Creating this to get some help. I'll try to create some eleventh hour workaround, and then maybe we can fix this for real.

- But first, I wanted to add some `strncat` or something and realized that `strcat` is already there but different from `strcpy`. I think the first argument is the destination for both, but it might be worth checking this because it is wrong in master as well.  
-  About what goes wrong inside goblint that triggers lots of dead code. At the core of it is the bot-initialization of mallocs, but we do at least some stuff that would avoid unsoundness, so I'm not completely sure what in this example is really the culprit because almost all simplifications of this example and test no longer fails. (It's even possible that the program is wrong...)


vesalvojdani:
This should allow the story for smtprc to proceed. I will check if I can replicate the chrony story with this merged into the chrony branch.

vesalvojdani:
I can replicate the chrony story with these changes, so I think we should merge.

michael-schwarz (approved):

sim642:
This is currently on top of #724, so merging this would just add to that PR, possibly confusing things.

sim642 (approved):

vesalvojdani:
Yeah, that's true... I don't even know why I spoke of merging in order to test this. Let's wait until next week with all the mergers then. We can just package this and the bigger-benchmark branches. We should even keep those names since they are so informative. :)

Issue #357, opened 2021-09-24
Marshal all data into single file


sim642:
In #354 it became apparent that separately marshaling loses some connections or duplicates some data between the files. See
https://github.com/goblint/analyzer/pull/354#discussion_r714703174, https://github.com/goblint/analyzer/pull/354#discussion_r714966485 and https://github.com/goblint/analyzer/pull/354#discussion_r714705319.

Hence it would make more sense to marshal everything (solution, solver hashtables, analysis auxiliary data, etc) connected and unduplicated in a single record.
We can still have options then controlling what is and isn't marshaled and just have them optional in the record. But if they're all present, they should all relate properly.

Probably to be done after #356 and #350, which both change what and where is marshaled.

sim642:
Another reason to do this is https://github.com/goblint/analyzer/pull/731#discussion_r873500774.

PR #737, opened 2022-05-16
Run dead code analysis with reanalyze

sim642:
This is the result of me trying out [reanalyze](https://github.com/rescript-association/reanalyze) on our codebase. Instead of a local unused check based on signatures (which we rarely have) that the OCaml compiler has, this attempts to perform a global dead code analysis.

I manually went through all of its warnings and did one of the following for each one:
1. Removed the dead code.
2. Added a TODO for potentially significant code, which should maybe be used. Feel free to voice your opinions.
3. Ignored the warning for false positives, significant code to keep or ARINC/spec analysis (which I didn't want to unnecessarily touch).

### TODO
- [x] Merge https://github.com/goblint/gobview/pull/13.
- [x] Update git submodule after https://github.com/goblint/gobview/pull/13 merge.


sim642:
Looks like I'll have to open a PR for Gobview as well. Will do that in a moment.

src/analyses/mCP.ml
```diff
@@ -268,9 +268,6 @@ struct
           Result.meet a @@ S.query ctx' q
         in
         match q with
-        | Queries.PrintFullState ->
-          ignore (Pretty.printf "Current State:\n%a\n\n" D.pretty ctx.local);
-          ()
```
michael-schwarz, line 273: I'm not sure, but do we want to keep this one around for debugging?

src/analyses/base.ml
```diff
@@ -73,7 +73,6 @@ struct
 
   let name () = "base"
   let startstate v: store = { cpa = CPA.bot (); deps = Dep.bot (); weak = WeakUpdates.bot (); priv = Priv.startstate ()}
-  let otherstate v: store = { cpa = CPA.bot (); deps = Dep.bot (); weak = WeakUpdates.bot (); priv = Priv.startstate ()}
```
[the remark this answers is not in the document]
sim642: The `otherstate` in `control.ml` is related to `otherfuns`, so that name is appropriate (until we get a better way to harness Linux kernel benchmarks).
The `otherstate` in analysis was a separate thing, which was replaced by `threadenter`.

src/analyses/mCP.ml
sim642: It was added in 3bd9caf5621ab7db1d2eb7fba0bda93c161a484b and never used. Also quite impractical and roundabout way to debug by abusing the in-analysis queries instead of just directly printing/tracing.

src/cdomains/listDomain.ml
```diff
@@ -2,6 +2,8 @@ open Cil
 
 module GU = Goblintutil
```
michael-schwarz, line 4: For domains as generic as lists, I don't think it's an issue if there are unused functions. These form the future building blocks of other domains, so I'd argue we keep these.

michael-schwarz, line 4: For domains as generic as lists, I don't think it's an issue if there are unused functions. These form the future building blocks of other domains, so I'd argue we keep these.

sim642: Sure, but it is suspicious if operations are defined but never used, even for such a simple list domain. We have that in our `ValueDomain` as a variant, but I think I've never seen that list abstraction in our analysis results. It just looks like incomplete beginnings of some analysis that never panned out. Maybe @vesalvojdani has an idea?

Because the more involved shape analysis was removed in #474, so I have no idea what, if anything, this still does. I'm fine with keeping these for now, but it'd be good to have the comment here pointing out that it's not fully used, such that anyone looking at this code doesn't get the wrong idea that this is something significant and maintained.

sim642: Turns out we only use this when `exp.list-type` is enabled. There's only one test for that: it's skipped and if executed fails for 5/6 annotations, which deepens my doubts about it.

michael-schwarz: Ok, Github on the phone is crappy and posts things twice :/

src/analyses/uninit.ml
```diff
@@ -47,6 +47,7 @@ struct
     | _ ->
       M.warn "Access to unknown address could be global"; []
 
+  (* TODO: unused? remove? *)
```
michael-schwarz, line 50: Is it not used via `is_expr_initd` -> `vars` -> `varoffs` -> `access_one_byval`?

src/analyses/varEq.ml
```diff
@@ -119,6 +119,7 @@ struct
       | _ -> false
 
   (* kill predicate for must-equality kind of analyses*)
+  (* TODO: why unused? how different from below? *)
```
michael-schwarz, line 122: They seem to be virtually identical, except the one below being a fair bit more chaotic

src/analyses/varEq.ml
```diff
@@ -151,6 +152,7 @@ struct
     let bt =  unrollTypeDeep (Cilfacade.typeOf b) in
     type_may_change_t a bt
 
+  (* TODO: why unused? how different from below? *)
```
michael-schwarz, line 155: They are very similar, but the code below handles more cases, but is also considerably more weird.

src/util/server.ml
```diff
@@ -129,14 +126,14 @@ let virtual_changes file =
 let increment_data (s: t) file reparsed = match !Serialize.server_solver_data with
   | Some solver_data when reparsed ->
     let changes = CompareCIL.compareCilFiles s.file file in
-    let old_data = Some { Analyses.cil_file = s.file; solver_data } in
+    let old_data = Some { Analyses.solver_data } in
```
michael-schwarz, line 129: This seems a bit weird... Why make a record that has only one member?

src/util/intOps.ml
```diff
@@ -14,8 +14,8 @@ sig
   (* Constants *)
   val zero : t
   val one : t
-  val lower_bound : t option
-  val upper_bound : t option
+  val lower_bound : t option (* TODO: unused *)
+  val upper_bound : t option (* TODO: unused *)
```
michael-schwarz, line 18: Those seem a bit strange here. IIRC, `IntOpsBase` is the module type we use for single values, not modules, ....
Maybe @jerhard knows what they are intended for?

src/util/intOps.ml
```diff
@@ -42,7 +42,7 @@ sig
   val compare : t -> t -> int
   val equal : t -> t -> bool
   val hash : t -> int
-  val top_range : t -> t -> bool
+  val top_range : t -> t -> bool (* TODO: unused *)
```
michael-schwarz, line 45: Same here, I am not so sure what the intended semantics is here

src/util/intOps.ml
```diff
@@ -51,7 +51,7 @@ sig
   val to_int : t -> int
   val of_int64 : int64 -> t
   val to_int64 : t -> int64
-  val of_string : string -> t
+  val of_string : string -> t (* TODO: unused *)
```
michael-schwarz, line 54: I think we can keep the `of_string`, it may be handy sometimes.

src/framework/analyses.ml
```diff
@@ -392,22 +374,19 @@ sig
...
 
   old_data: analyzed_data option;
-  new_file: Cil.file;
   changes: CompareCIL.change_info
 }
 
-let empty_increment_data ?(server=false) file = {
+let empty_increment_data ?(server=false) () = {
   server;
   old_data = None;
```
michael-schwarz, line 389: It's interesting that the file need not be saved here...

src/domains/partitionDomain.ml
```diff
@@ -132,6 +132,7 @@ struct
       in
       fold f xs (empty ())
 
+  (* TODO: unused *)
```
michael-schwarz, line 135: I think we might want to keep this

src/domains/mapDomain.ml
```diff
@@ -31,7 +31,7 @@ sig
   val for_all: (key -> value -> bool) -> t -> bool
   val map2: (value -> value -> value) -> t -> t -> t
   val long_map2: (value -> value -> value) -> t -> t -> t
-  val merge : (key -> value option -> value option -> value option) -> t -> t -> t
+  val merge : (key -> value option -> value option -> value option) -> t -> t -> t (* TODO: unused, remove? *)
```
michael-schwarz, line 34: I think `merge` may also be worth keeping

src/domains/lattice.ml
```diff
@@ -411,6 +411,7 @@ end
 module Prod = ProdConf (struct let expand_fst = true let expand_snd = true end)
 module ProdSimple = ProdConf (struct let expand_fst = false let expand_snd = false end)
 
+(* TODO: unused, missing lattice ops *)
```
michael-schwarz, line 414: I'd be cool with removing it if there is no usage

src/domains/hoareDomain.ml
```diff
@@ -105,7 +104,7 @@ struct
     (* Map.map (List.map f) m *)
     (* since hashes might change we need to rebuild: *)
     apply_list (List.map f) m
-  let filter f m = apply_list (List.filter f) m (* TODO do something better? *)
+  let filter f m = apply_list (List.filter f) m (* TODO do something better? unused *)
```
michael-schwarz, line 107: I think it's worth keeping

src/domains/hoareDomain.ml
```diff
@@ -48,6 +48,7 @@ struct
   let elements m = Map.values m |> List.of_enum |> List.flatten
 
   (* merge elements in x and y by f *)
+  (* TODO: unused, remove? *)
   let merge op f x y =
```
michael-schwarz, line 52: also worth keeping in my opinion

src/cdomains/lval.ml
```diff
@@ -355,11 +346,13 @@ struct
 
   let printXml f x = BatPrintf.fprintf f "<value>\n<data>\n%a\n</data>\n</value>\n" printInnerXml x
 
+  (* TODO: only used for unused musteqdomain *)
```
michael-schwarz, line 349: The `musteqdomain` has some sort of TODO that it should e.g. be used for `regions`

src/framework/analyses.ml
sim642: The file isn't saved in the incremental data, but with a separate variant into a separate file using `Serialize`, so this field was never read from. Actually #357 would probably change this and put the file (along with everything else) into a single record and not have separate files. Depending on the choices taken for #357, it need not be this increment data record that everything goes into, but maybe some outer wrapping record.

src/util/server.ml
sim642: I didn't want to introduce additional refactoring on top of this. Depending on how #357 and #545 are done, the record might gain additional fields again or could be removed entirely. Didn't want to refactor something that is bound to be refactored more extensively anyway.

src/util/intOps.ml
sim642: I suppose at some point (maybe pre-bigints) it was necessary to know the `min_int` and `max_int` for these different types.

src/util/intOps.ml
sim642: Looking at the implementations, it seems to be for checking if the two arguments are `min_int` and `max_int` respectively. Again I guess we might've used that pre-bigints to do some special treatment, but that seems specific to intervals anyway.

src/util/intOps.ml
sim642: Possibly yes. I guess this was used when we still parsed the strings from CIL constants for big values.

src/domains/setDomain.ml
```diff
@@ -36,38 +36,9 @@ sig
   val min_elt: t -> elt
   val max_elt: t -> elt
   val choose: t -> elt
-  val split: elt -> t -> t * bool * t
+  val split: elt -> t -> t * bool * t (* TODO: unused, remove? *)
```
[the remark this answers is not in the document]
sim642: All these functions have just been directly copied from the standard library: https://v2.ocaml.org/api/Set.S.html#VALsplit.

src/analyses/uninit.ml
sim642: You're right. I accidentally added the TODO to the wrong function, the similarly named `access_byval` below is actually unused.

src/analyses/varEq.ml
sim642: I guess it might have been copied there, but without completely rewriting this all, I'd rather not touch this analysis more than necessary.

src/cdomains/addressDomain.ml
```diff
@@ -117,6 +115,7 @@ struct
       | false, false -> join x y
   *)
 
+  (* TODO: overrides is_top, but not top? *)
```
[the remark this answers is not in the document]
sim642: I suppose it's weird because there isn't just a single value that this `is_top` returns true for. Probably good to have a comment about it to avoid surprises until we somehow design a more consistent domain where this inconsistency isn't necessary.

src/analyses/uninit.ml
```diff
@@ -78,6 +78,7 @@ struct
     | Var x, ofs -> access_offset ofs
     | Mem n, ofs -> access_one_byval a false n @ access_offset ofs
 
+  (* TODO: unused? remove? *)
```
michael-schwarz, line 81: Yes, let's remove it!

src/cdomains/exp.ml
```diff
@@ -59,6 +59,7 @@ struct
     in
     cv false e
```
michael-schwarz, line 60: I think we can remove the unused things here

src/domains/setDomain.ml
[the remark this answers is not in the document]
michael-schwarz: Yes, I see that now. I still think it is not very obvious what it supposed to do without consulting the documentation. Would you be in favor of keeping it for consistency with the standard library?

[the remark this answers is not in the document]
sim642: I'll remove it, there's no need to match the standard library and we're not up to date in that regard anymore anyway. `split` is a low-level operation that is used to implement operations like `union` and `inter` efficiently (by splitting the second set according to the root of the first one, recursively).

src/cdomains/exp.ml
sim642: I removed the unused ones and completely split up this file to more analysis-specific ones because the remaining functions weren't necessarily for general-purpose use, but some analysis-specific logic, especially regarding a custom equality implementation.

src/cdomains/lval.ml
sim642: Yup, I moved these few functions from here to there, so all that unused stuff is together.

src/cdomains/listDomain.ml
sim642: Completely removed after GobCon decision.

src/analyses/condVars.ml
```diff
@@ -1,4 +1,5 @@
 (** Must equality between variables and logical expressions. *)
+(* TODO: unused, what is this analysis? *)
```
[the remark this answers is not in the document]
sim642: It's another untested analysis and the only use is in base analysis's `invariant`. But considering that we never activate it, this never affects the base analysis in our usage.
It's another must equality analysis, so I wonder how it's different from var_eq. Given that description, it sounds more specific and thus maybe cheaper than general var_eq?

Along with `MusteqDomain`'s unused code, this sounds like another candidate for eventual removal, but considering the mess of must equality analyses we currently have, it might be good to keep it for now. At some point a nice must equality analysis needs to be implemented, which is maintainable and covers all three of the existing ones, but that can be a future goal.

sim642:
Since https://github.com/goblint/gobview/pull/13 is now merged into Gobview, which now depends on changes from this PR, it would be good to now delay this too much since it blocks any other PRs to Gobview as they would only work on this branch of Goblint.

Regarding all the "worth keeping" comments, I'd suggest not removing those things now, but leaving the comments so it's immediately apparent if some weird code is used for anything or not.

michael-schwarz:
> I'd suggest not removing those things now, but leaving the comments so it's immediately apparent if some weird code is used for anything or not.

Ok, sounds good to me!

PR #738, opened 2022-05-18
Eager destab, cheap abort

michael-schwarz:
This WIP is the Munich idea for aborting. It's still far from polished, just thought I'd open the PR in case someone wants to take a look.

TODOs:

- [x] Cleanup code
- [x] Test impact
- [x] Decide if we want to have it 
- [x] Integrate with interactive solver


michael-schwarz:
How does this relate to your aborting @sim64? It somehow seems less involved?

michael-schwarz:
Somehow, the benefit seems unclear at best: For `./goblint --conf conf/traces.json --set ana.base.privatization write+lock ../bench/pthread/smtprc_comb.c`, we actually observe a slowdown by a factor of 10(!).

```
Timings:
TOTAL                           1.327 s
  parse                           0.038 s
  convert to CIL                  0.028 s
  mergeCIL                        0.006 s
  analysis                        1.256 s
    createCFG                       0.028 s
      handle                          0.012 s
      iter_connect                    0.015 s
        computeSCCs                     0.012 s
    global_inits                    0.005 s
    solving                         1.223 s
      S.Dom.equal                     0.010 s
      cheap_full_reach                0.002 s
      postsolver                      0.380 s
        postsolver_iter                 0.375 s
    warn_global                     0.000 s
      access                          0.000 s
Timing used
Memory statistics: total=3680.82MB, max=37.52MB, minor=3678.05MB, major=44.41MB, promoted=41.64MB
    minor collections=1757  major collections=9 compactions=0
```

vs

```
Timings:
TOTAL                           9.385 s
  parse                           0.050 s
  convert to CIL                  0.028 s
  mergeCIL                        0.006 s
  analysis                        9.303 s
    createCFG                       0.028 s
      handle                          0.012 s
      iter_connect                    0.015 s
        computeSCCs                     0.013 s
    global_inits                    0.004 s
    solving                         9.266 s
      S.Dom.equal                     0.010 s
      cheap_full_reach                0.002 s
      postsolver                      0.366 s
        postsolver_iter                 0.361 s
    warn_global                     0.001 s
      access                          0.001 s
Timing used
Memory statistics: total=32637.48MB, max=351.13MB, minor=32634.65MB, major=418.49MB, promoted=415.67MB
    minor collections=15568  major collections=17 compactions=0
```

src/solvers/td3.ml
```diff
@@ -272,6 +308,9 @@ module WP =
         let tmp = simple_solve l x y in
         if HM.mem rho y then add_infl y x;
         if tracing then trace "sol2" "eval %a ## %a -> %a\n" S.Var.pretty_trace x S.Var.pretty_trace y S.Dom.pretty tmp;
+        if skip_unchanged_rhs then
+          (let (oldv,curr_dep_vals) = HM.find dep_vals x in
+           HM.replace dep_vals x (oldv,((y,tmp) :: curr_dep_vals)));
```
sim642, line 313: I suspect this association list structure could also be very inefficient and memory hungry. If the right-hand side `get`s a single unknown 1000 times and the value doesn't change in between, then the resulting list has 1000 identical pairs.
Maybe this is what is happening with some of the slower and more complicated privatizations since they rely on a lot more global invariants. For normal evaluation it isn't a problem because without changes, the subsequent `get`s of the same variable are direct lookups of a stable variable, so they're as efficient as they possibly could be (the solver itself serves as a cache for `get`).

michael-schwarz:
@jerhard and I managed to solve the performance degradation: It seems like this `dep_vals` was populated to eagerly!

src/solvers/td3.ml
```diff
@@ -122,6 +122,16 @@ module WP =
...
 
+      let skip_unchanged_rhs =
+        let enabled = GobConfig.get_bool "solvers.td3.skip-unchanged-rhs" in
+        if enabled && (restart_sided || restart_only_globals || restart_only_access || restart_wpoint || restart_once) then
+          (M.warn "restarting active, disabling solvers.td3.skip-unchanged-rhs"; false)
+        else
+          enabled
+      in
+      (* TODO: Do we want to marshal dep_vals? *)
+      let dep_vals = HM.create 10 in
```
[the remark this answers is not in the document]
michael-schwarz: I think the issue is that destabilization no longer re-triggers all side-effects which we rely on when restarting e.g. globals.

michael-schwarz:
After this change, we save some 1000 evaluations, and end up with comparable performance in both cases.

sim642:
> @jerhard and I managed to solve the performance degradation: It seems like this `dep_vals` was populated to eagerly!

Looks like after pulling that `dep_vals` populating code from `eval` to around `S.system` in `solve`, it's all very localized as a wrapper around `S.system`, like my `caching-constrsys` attempt.
If it's like that, then it's still my opinion that it's nicer to pull it out as a constraint system functor in order to keep TD3 itself as clean as possible. But that can be future discussion when this becomes fast enough to make a difference.

src/solvers/td3.ml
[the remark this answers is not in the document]
sim642: If something is restarted, then the value in `rho` is changed, so the checking should detect that and inevitably re-evaluate the right-hand side to retrigger all the side effects. I'm just surprised because in #364 this wasn't a problem, as far as I remember.

Or at worst, if it somehow is, then maybe just disable this for the duration of restarting/postsolving. Because with restarting enabled, normal solving should still be able to benefit from this.

src/solvers/td3.ml
```diff
@@ -234,10 +244,45 @@ module WP =
...
         | None -> S.Dom.bot ()
-        | Some f -> f get set
+        | Some f ->
+          if skip_unchanged_rhs then
+            let all_deps_unchanged =
+              match HM.find_option dep_vals x with
+              | None -> None
+              | Some (oldv, deps) ->
+                let deps_inorder = List.rev deps in
+                if List.for_all (fun (var, value) -> S.Dom.equal (get var) value) deps_inorder then
```
sim642, line 257: Intuitively this reversal makes sense to have the dependencies in the same order as they were added, but is it really necessary? Because the only operation on `deps_inorder` is a `List.for_all` for which it doesn't really matter what the order of elements is.

Because this reversal copies the list each time and immediately discards it after `List.for_all`, creating garbage. Or if it is really necessary/useful for some reason, maybe do the reversal when the dependencies are connected. That way it's done once per unchanged unknown, rather than as many times as the unknown is later queried again.

src/solvers/td3.ml
```diff
@@ -234,10 +244,45 @@ module WP =
...
+        
         match S.system x with
         | None -> S.Dom.bot ()
-        | Some f -> f get set
+        | Some f ->
+          if skip_unchanged_rhs then
+            let all_deps_unchanged =
+              match HM.find_option dep_vals x with
+              | None -> None
+              | Some (oldv, deps) ->
```
sim642, line 255: Is it necessary to store `oldv` like this? Might as well take it directly from `rho` if dependencies are unchanged.
At least that worked for #364 where no values where stored for aborting at all. In `caching-constrsys` the only reason I stored a separate `prev_val` was because the functor had no access to solver's `rho`.

sim642:
This probably only matters if this PR provides meaningful speedup and we want to have it, but currently this is on top of #391, where all aborting was now reverted since it wasn't part of the paper. Since it's separate, then it might be more general to have directly on top of `master`.
Or maybe this aborting stuff becomes a bigger venture itself, where we want to integrate all the different approaches for benchmarking comparison.

src/solvers/td3.ml
michael-schwarz: Yes, it is needed (but took us a while to figure out yesterday): Consider the case where the value we are considering is the loop head, at which widening is applied: Thus the old value we find in rho is the widened value. However, for a subsequent narrowing iteration we do not want to use the value AFTER widening, but we want to use the last value that the RHS returned, such that we can then benefit in the narrowing. This is evidenced e.g. by test 44/44. (Your `eq-reuse` in `solve` sometimes fixes it, but to work generally we need `oldv`).

src/solvers/td3.ml
michael-schwarz: I think in principle, the same unknown can be queried in an RHS twice, and in the presence of side-effects  (e.g. from other unknowns that are solved), I think there is no guarantee that both times the same value is returned. If this is the case, it might make a difference because the first read has read an outdated value while later ones have read the current one.
This means that storing as a map (Var -> Val) is definitely out, and I am a bit worried that having the dependencies in another order might also be problematic, though I do not have an example yet.

src/solvers/td3.ml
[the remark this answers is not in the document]
michael-schwarz: >  so the checking should detect that and inevitably re-evaluate the right-hand side to retrigger all the side effect

It might be true in all our examples, but consider a global that is re-started and an unknown that has a side-effect TO it. Thus unknown is de-stabilized, but then we find that none of its dependencies has changed and the RHS is not re-evaluated, right? So we'd be missing a side-effect?!

michael-schwarz:
> If it's like that, then it's still my opinion that it's nicer to pull it out as a constraint system functor in order to keep TD3 itself as clean as possible.

> Since it's separate, then it might be more general to have directly on top of `master`.

The reason i did it here is that we were hoping it could be an generalization of all the reluctant destabilization business that clutters TD3 now. I think having this in there is much less clutter than all of that business.

I also don't think it will stay as separate as it currently is, as my next ambition would be to somehow integrate this with superstable such that warnings etc can be re-used for things that were destabilized, but their RHS never needed to be evaluated. This would IMO bridge the distance between reluctant destab and what we have here.


I think as a name for this nascent feature, `skip-unchanged-rhs` might be better than `aborting` as there really isn't anything being aborted here.

src/solvers/td3.ml
sim642: Ah, yes, that's right! I think in #364 I also had to have a special check that forbid aborting when switching from `Widen` to `Narrow` using `~abortable:false`. I think at the time, I didn't fully understand why, but your explanation makes sense.

I guess it's probably easier to explicitly keep `oldv` than trick around it and understand the exact combination of conditions when it cannot be used.

sim642:
> my next ambition would be to somehow integrate this with superstable such that warnings etc can be re-used for things that were destabilized, but their RHS never needed to be evaluated.

This is an interesting point, but maybe it doesn't have to be directly related to this even. Maybe our definition of `superstable` should be different, i.e. things only get removed when values are actually changed in `solve`/`side`, not when they are immediately destabilized. That would automatically cover the case when the RHS was never re-evaluated and the value remained the same.

src/solvers/td3.ml
sim642: If the same queried unknown changed in between due to a side effect, then it cannot be reused at all, because later when doing these checks `get var` will return the same value in both cases (because between these there are no side effects), so one of the two equality checks must necessarily fail.

michael-schwarz: > because between these there are no side effects

Is this actually guaranteed though? What if one of the `get`s in turn triggers an evaluation (including side-effects) of some other unknown?

michael-schwarz:
> Maybe our definition of `superstable` should be different, i.e. things only get removed when values are actually changed in `solve`/`side`, not when they are immediately destabilized.

Hmm, that's true. There's no need to consider things that became destabilized once no longer superstable now that we have the cheap reachability and unreachable things being superstable is no longer a concern to us.

src/solvers/td3.ml
michael-schwarz: The other point that I forgot to mention is that if I do it in the order that the unknowns were actually queried, I will not issue calls to get (and thus potentially to `solve`) for unknowns that then won't be used in the RHS.

Consider the worst case where I depend on many variables and do this comparison without reversing the list. The changed value for the first unknown the RHS asks for might mean that none of the other unknowns queried previously are consulted anymore. If I reverse the list, I will stop immediately at the first changed unknown, having only asked for things that the RHS will ask for as well. If I don't, I will consider all others to then notice that the endeavour was pointless, having queried things that the RHS will not even ask for anymore.

michael-schwarz:
The performance story is still a bit mucky: sometimes we have big savings, sometimes it doesn't matter, and sometimes it is worse.

abort

michael-schwarz:
I now added marshalling to see how this works together with the incremental solver: Currently, many incremental tests still fail because the fix-point is not reached (`9 test(s) failed: ["00/00 local", "00/09 unreach", "00/11 unreach-reusesuper", "03/03 reluctant-int-annotation-dyn", "13/01 mutex-simple", "13/05 race-call-remove", "13/06 mutex-simple-reluctant", "13/07 mutex-simple-reluctant2", "13/08 mutex-simple-reluctant3"]`).

Any idea what could be wrong here @stilscher @jerhard? It would seem that we have to clear this hash-table for some unknowns if the constraints for them now changed or something like this?

src/solvers/td3.ml
sim642: > Is this actually guaranteed though? What if one of the `get`s in turn triggers an evaluation (including side-effects) of some other unknown?

Suppose one call of `system x` does:
1. `get y` and gets `d_1` (and caches `(y, d_1)`),
2. something else which destabilizes `y`,
3. `get y` again and gets `d_2` (and caches `(y, d_2)`).

After this, `rho y = d_2`.

The following call of `system x` would first check if `get y = d_1`, which wouldn't hold (something might've increased `rho y` further, but it wouldn't go back down to `d_1`). If it somehow did, then the second `system x` evaluation would be causing the exact same increasing transition from `d_1` to `d_2` for `y` again, which seems to violate TD's termination.
So it's not even about what happens between the `get`s of the subsequent call.

Side-effected variables (leaves in TD3 paper terminology) only join and widen contributions, but there's no narrowing for leaves. If there was, then I suppose it might be possible for a leaf to narrow back down to `d_1` again, but otherwise I cannot see how it could happen.

> Consider the worst case where I depend on many variables and do this comparison without reversing the list. The changed value for the first unknown the RHS asks for might mean that none of the other unknowns queried previously are consulted anymore. If I reverse the list, I will stop immediately at the first changed unknown, having only asked for things that the RHS will ask for as well. If I don't, I will consider all others to then notice that the endeavour was pointless, having queried things that the RHS will not even ask for anymore.

That's a good point. In my versions with exception aborting this difference was avoided by only ever doing `get`s that RHS does, so it always matches the order that the RHS queries them in and doesn't query anything additional accidentally.

Intuitively it seems to me that if there are unknowns that are no longer queried, then `get`ting them wouldn't do real work either, but just look them up. If something else changed them, then they are still in use for something else, meaning that even if that `get` triggers a real `solve`, it would solve something that would've been needed to eventually solved anyway.

Continuing from that last point, if such things somehow happen, then if the preliminary `get` for comparison does an actual `solve` underneath, then that could drastically change what is solved in what order. In particular, this could change the order of some side effects, where in turn widening might cause different results because of it (since `widen` is not commutative).
So it's far from just observing the solving, but the preliminary checks do actually change the solving structure.

Of course my entire point about `List.rev` was just a guess. Profiling with perf and memtrace should directly show if that's anything significant at all.

sim642:
> The performance story is still a bit mucky: sometimes we have big savings, sometimes it doesn't matter, and sometimes it is worse.

It would be a good idea to also do SV-COMP runs. It's a much bigger sample size and can very easily see speedup/slowdown trends overall instead of eyeballing.

michael-schwarz:
Apparently, this will be included in a paper we're writing on the Munich side (The Cousot Festschrift), so we should attempt to get it merged.

sim642:
Given the renowned activity on this, I guess you're intending to polish this into a working state.

If so, then we should really think about this point I had in some of my earlier comments:

> Just to have this here, my earlier take on this as a constraint system functor and actually using exceptions for aborting right-hand sides is here: [531b8a6](https://github.com/goblint/analyzer/commit/531b8a60fce17a99a3fc1a6eee49ca757539e295) (also in `caching-constrsys` branch).

> Looks like after pulling that `dep_vals` populating code from `eval` to around `S.system` in `solve`, it's all very localized as a wrapper around `S.system`, like my `caching-constrsys` attempt. If it's like that, then it's still my opinion that it's nicer to pull it out as a constraint system functor in order to keep TD3 itself as clean as possible. But that can be future discussion when this becomes fast enough to make a difference.

If we still want to achieve #545 (which was started in #596), then hard-coding this into TD3 specifically is a step in the wrong direction. Coupling more features into the already monolithic TD3 solver just creates more future work to maintain the even larger solver and more things to decouple.
So it would be highly preferable, if we can avoid adding more coupling then there already is. I recall there being an email chain with Helmut regarding how this should be presented and I think we realized at the time it would've been equivalent to a lifting of the right-hand sides. Of course presenting such technique on slides is easier if it's inlined into the solver, but that's quite different from the real-world implementation of TD3, which is over 1000 lines by now.

Therefore, what we should think about is whether any part of this (as it is by now) necessarily needs to be coupled into TD3 or could it still be achieved as a constraint-system lifting functor.

michael-schwarz:
I think the reason we are opposed to having it as a functor around the constraint system on the Munich side is that any such data structures that are not maintained in the solver a prone to later breakage, and other unpredictable results. I don't think that this wrapping of TD3 adds much to the complexity.

I agree that, in principle, decoupling things as far as possible is beneficial. However, there are easier wins (such as all the incremental pre-processing).

sim642:
> I think the reason we are opposed to having it as a functor around the constraint system on the Munich side is that any such data structures that are not maintained in the solver a prone to later breakage, and other unpredictable results. I don't think that this wrapping of TD3 adds much to the complexity.

Wrapping the constraint system and wrapping TD3 are essentially the same thing: a functor between the solver and the constraint system. As it is right now, it's not really wrapping TD3, but simply inlining the wrapper code right in the middle of TD3.

Also, there's no reason the transfer function wrapping cannot be decoupled and still be in TD3 at the same time. The incremental warning handling does precisely that:
https://github.com/goblint/analyzer/blob/abc2c404658b68084641220d55a9ff5f1601320f/src/solvers/td3.ml#L85-L86
It does a very similar thing: wraps all right-hand side executions with some current constraint variable tracking without having to insert all that right in the middle of TD3, where it's very self-contained instead of blurring the lines between what's TD3 itself and what's the right-hand side aborting feature. And it can be very close to the particular solver: the solver instantiates the functor and has access to the `dep_vals` data structure used in the functor (for example to marshal it).

> I agree that, in principle, decoupling things as far as possible is beneficial. However, there are easier wins (such as all the incremental pre-processing).

These are two completely unrelated wins. Incremental preprocessing is an optimization, modularizing TD3 is a matter of maintenance. The high degree of coupling was a primary reason for reverting #364, despite it providing up to 50% speedup with no regressions, which was more than this, if I recall correctly.
Although there haven't been recent sv-benchmarks runs with this recently, only non-relational traces benchmarks.

Ultimately I will not stop this from being included into TD3 inlined instead of modularized. But in that case I expect to not hear any complaints about the size of the TD3 implementation either.

michael-schwarz:
Let's puts this onto the agenda for Wednesday :smile:

tests/incremental/14-manual-restart/04-restarting_and_aborting.c
```diff
@@ -0,0 +1,16 @@
+int g = 0;
```
sim642, line 1: Required after #845:
```suggestion
#include <goblint.h>

int g = 0;
```

sim642:
> What I meant here was removing the pre-processing of all the solver data structures for the incremental mode from the solver itself and putting it somewhere else.

#545 does that to an extent it's still possible. These data structures are solver-specific and the concrete type only defined within the solver, so they can't be arbitrarily elsewhere, only extracted out of the `solve` function.

sim642:
I did the refactoring currently in a separate branch: https://github.com/goblint/analyzer/tree/eager_destab_cheap_abort-refactor. Since it depends on #909, I didn't yet merge it here in order to not clutter the diff with changes that need to be reviewed separately first.

As far as I can see, the last status of benchmarking this was on the tiny set of traces benchmarks:

> The performance story is still a bit mucky: sometimes we have big savings, sometimes it doesn't matter, and sometimes it is worse.

If this is to become enabled by default, we need comprehensive performance comparison.

michael-schwarz:
> If this is to become enabled by default, we need comprehensive performance comparison.

Yes! I would suggest to postpone this until after the sv-comp deadline.

src/util/options.schema.json
```diff
@@ -31,36 +31,42 @@
     },
     "verify": {
       "title": "verify",
-      "description":
-        "Verify that the solver reached a post-fixpoint. Beware that disabling this also disables output of warnings since post-processing of the results is done in the verification phase!",
+      "description": "Verify that the solver reached a post-fixpoint. Beware that disabling this also disables output of warnings since post-processing of the results is done in the verification phase!",
```
sim642, line 34: Apparently d8b94a5c5f936d9d3a0733003de84b42fa728d32 reformatted all the schema...
Maybe I'll rebase something here.

sim642: Done! Now the only schema change is the single added option.
Interactive rebase with merges is not for the faint-hearted.

sim642:
I now ran this on sv-benchmarks and there it doesn't help, but has ~3% slowdown:
image
([Full table](https://goblint.cs.ut.ee/results/108-all-fast-abort-after/results/results.2022-12-05_13-30-27.table.html#/scatter?toolY=1&columnY=1&filter=0(0*status*(category(notIn(error)))),1(0*status*(category(notIn(error))))&results=All&regression=Linear&scaling=Linear&line=1.2&toolX=0&columnX=1))

michael-schwarz:
Even if there is no speedup everywhere, I think it would make sense to merge this, as it also something we describe in our TD0 to TD3 paper...

sim642:
Sure, I just disabled it by default for now.

sim642 (approved):

PR #740, opened 2022-05-19
Add server command `pre_files`

### LINKED ISSUE #739: Server mode command for just preprocessing to get files list
In #606 a command was added to get the list of involved files in the analysis. However, currently this is only available in the server after doing a complete analysis.

For https://github.com/goblint/GobPie/issues/17 and https://github.com/goblint/GobPie/issues/31, it would be useful to get the list of files already on startup (so GobPie knows which files should even trigger analysis and for which server instance) without having to wait for the entire analysis to complete.
This would be possible because getting the list of files only requires preprocessing and parsing (not even merging),

sim642:
Closes #739.

This is on top of #391, because that's where #625 went. On `master` the server always immediately preprocesses, parses and merges the file, so this isn't actually needed there.

Even the full parsing is a bit excessive for this, but we need to do it because we use the same CIL parsing to collect all the files from preprocessor pragma locations. This could be optimized further by having a special parser for just those that ignores the actual C code, but that would be too premature for now.


michael-schwarz (approved):
LGTM!

michael-schwarz:
@karoliineh Can you have a look so we can merge this?

sim642:
I'll just merge this because the GobPie features it was intended for aren't a priority right now.

PR #742, opened 2022-05-22
Apron: Track relational information for variables that have their address taken

### LINKED ISSUE #660: Handle variables that have their addresses taken in Apron
Our Apron analysis should also work for variables that have their address taken.

This entails the following
- [ ] Handling of escapes
- [ ] Resolving acceses via the base analysis
- [ ] Handling of variables that need to be updated weakly

michael-schwarz:
TODOs:

- [x] Handle escapes
- [x] Handle variables that need to be updated weakly
- [x] Check if it still works with local of outer functions with different names reachable
- [x] Rectify locals of function in which thread is created leaking to created thread
- [x] Test

The strange numbering in `apron2` is for consistency with #391 

Closes #660


michael-schwarz:
Looks like this is not a simple local change but will require extensive changes, as after considering escaping, it no longer is true that a variable is either a global or a local, but instead it may be both depending on whether it escaped. :face_exhaling:

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -121,6 +124,16 @@ struct
...
       )
+    | (Mem v, NoOffset) ->
+      (let r = ask.f (Queries.MayPointTo v) in
+       match r with
+       | `Top ->
+         st
+       | `Lifted s ->
+         let lvals = Queries.LS.elements r in
+         let ass' = List.map (fun lv -> assign_to_global_wrapper ask getg sideg st (Lval.CilLval.to_lval lv) f) lvals in
+         List.fold_right (fun {apr=apr1; priv=priv1} {apr=apr2; priv=priv2} -> {apr = AD.join apr1 apr2; priv = Priv.D.join priv1 priv2}) ass' {apr = AD.bot (); priv = Priv.D.bot ()}
```
sim642, line 135: Aren't these just `D.join` and `D.bot`?

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -139,17 +152,35 @@ struct
...
+      | UnOp (unop, e, typ) -> UnOp(unop, inner e, typ)
+      | BinOp (binop, e1, e2, typ) -> BinOp (binop, inner e1, inner e2, typ)
+      | CastE (t,e) -> CastE (t, inner e)
+      | Lval (Var v, off) -> Lval (Var v, off)
+      | Lval (Mem e, NoOffset) ->
+        (match ask (Queries.MayPointTo e) with
+         | (`Lifted _) as r when (Queries.LS.cardinal r) = 1 ->
+           let lval = Lval.CilLval.to_lval (Queries.LS.choose r) in
+           Lval lval
+         | _ -> Lval (Mem e, NoOffset))
```
sim642, line 167: Unlike the `MayPointTo` in `assign_to_global_wrapper`, this doesn't perform a join over all possible addresses/values. Not sure it matters for anything real, but if all pointed to values are equal according to Apron (or their join could be bounded, etc), then there's extra information that currently wouldn't be used.

Getting it to work would be significantly more involved though. I guess one would need something like `read_globals_to_locals` which reads (and joins) all dereferenced things into temporary locals and then computes with those instead.

sim642:
> considering escaping, it no longer is true that a variable is either a global or a local, but instead it may be both depending on whether it escaped.

But at any particular program point, it's either must-local or may-escaped, no?

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: Yes, it was a very complicated way to write something simple indeed!

michael-schwarz:
I started a run on sv-comp to see if this introduces any unsoundness

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: True! For now, let's skip this and add it if we notice it has significant performance penalties.

sim642:
> Oddly enough, this breaks the new example `57-vectorsize-lazy-downsize`, no idea what's going on there....

That's worrisome, I'll try to look into it. Maybe this is forgetting something (like type bounds) that it shouldn't?

sim642:
The issue was variable naming for the widening threshold expression conversion. I made the small fix here.

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -104,14 +104,17 @@ struct
       {st with apr = apr'}
     )
 
-  let assign_to_global_wrapper ask getg sideg st lv f =
+  let rec assign_to_global_wrapper (ask:Queries.ask) getg sideg st lv f =
     match lv with
     (* Lvals which are numbers, have no offset and their address wasn't taken *)
     (* This means that variables of which multiple copies may be reachable via pointers are also also excluded (they have their address taken) *)
     (* and no special handling for them is required (https://github.com/goblint/analyzer/pull/310) *)
```
sim642, line 111: I'm having hard time understanding what this comment is trying to say: "Lvals which are _numbers_", what?
The second half seems to be about `IsMultiple`, but now we're accounting for it, so that's no longer true, I suppose.

src/analyses/apron/apronAnalysis.apron.ml
sim642: Maybe add a comment/TODO about it here (possibly with a comment link).

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -139,17 +152,35 @@ struct
...
+      | BinOp (binop, e1, e2, typ) -> BinOp (binop, inner e1, inner e2, typ)
+      | CastE (t,e) -> CastE (t, inner e)
+      | Lval (Var v, off) -> Lval (Var v, off)
+      | Lval (Mem e, NoOffset) ->
+        (match ask (Queries.MayPointTo e) with
+         | (`Lifted _) as r when (Queries.LS.cardinal r) = 1 ->
+           let lval = Lval.CilLval.to_lval (Queries.LS.choose r) in
+           Lval lval
+         | _ -> Lval (Mem e, NoOffset))
+      | e -> e
```
sim642, line 168: This wildcard case contains more constructors that may contain subexpressions. Should we really not be recursing there?

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -175,7 +206,14 @@ struct
 
   (* Function call transfer functions. *)
 
+  let pass_to_callee fundec reachable_from_args var =
+    let vname = Var.to_string var in
+    match List.find_opt (fun v -> v.vname = vname) (fundec.sformals @ fundec.slocals) with
```
sim642, line 211: Now that Apron `Var`s for locals have their `vid` appended to their name, so `vname` has it, but `v.vname` from `fundec.slocals` does not. So does this ever find anything?

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -175,7 +206,14 @@ struct
...
+    | None -> true
+    | Some v -> Queries.LS.exists (fun (v',_) -> v'.vid = v.vid) reachable_from_args
```
sim642, line 213: This seems a bit silly on a set. Do we need offsets of reachables for anything in the apron analysis? If not, then could just use a set of `varinfo`s and use `mem`.

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -253,13 +292,16 @@ struct
...
+    let fundec = Node.find_fundec ctx.node in
     if M.tracing then M.tracel "combine" "apron f: %a\n" d_varinfo f.svar;
     if M.tracing then M.tracel "combine" "apron formals: %a\n" (d_list "," d_varinfo) f.sformals;
     if M.tracing then M.tracel "combine" "apron args: %a\n" (d_list "," d_exp) args;
     let new_fun_apr = AD.add_vars fun_st.apr (AD.vars st.apr) in
     let arg_substitutes =
       GobList.combine_short f.sformals args (* TODO: is it right to ignore missing formals/args? *)
-      |> List.filter (fun (x, _) -> AD.varinfo_tracked x)
+      (* Do not do replacement for actuals whose value may be modified after the call *)
+      |> List.filter (fun (x, e) -> AD.varinfo_tracked x && List.for_all (fun v -> not (Queries.LS.exists (fun (v',_) -> v'.vid = v.vid) reachable_from_args)) (Basetype.CilExp.get_vars e))
```
sim642, line 304: `varinfo` set would also be good here.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -73,7 +73,8 @@ struct
   include VarMetadataTbl (VM)
   open VM
 
-  let local x = make_var ~name:x.vname Local
+  let local_name x = x.vname ^ "#" ^ string_of_int(x.vid)
```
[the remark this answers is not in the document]
sim642: Or even now that we use different names for them, in the above `VM` module they could have a `varinfo` argument, because the correspondence is one-to-one. That would cause a quite widespread change though (like simplifying `to_cil_varinfo` below to be direct), so maybe we hold off that refactoring from this PR.

tests/regression/36-apron/07-problem-pointer.c
```diff
@@ -1,19 +1,21 @@
...
 int main() {
     int c = __VERIFIER_nondet_int();
-    g = 3; // Globals are not tracked by apron for now
+    g = 3;
     assert(g != 3); // FAIL
     assert(g == 3);
     int a = 5;
     int *p = &a; // after this apron should put a to top because pointers are not tracked
     change(p);
+    assert(a == 5); //FAIL
```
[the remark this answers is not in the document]
michael-schwarz: When apron made unsound claims before, this interfered with what base computed, leading to some asserts that should succeed no longer succeeding, and some of this code being reported as dead code. All these asserts failed at various stages in the process of extending Apron to values who have their addresses taken, so I think having them is meaningful.

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: :+1:

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -315,6 +336,8 @@ struct
     | `Thread ->
       st
 
+  let escape node ask getg sideg st escaped = (* TODO: Implement *) st
```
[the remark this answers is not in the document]
michael-schwarz: Good point!

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: I removed the comment completely now, it was outdated.

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: Good catch! No, it doesn't, which was actually problematic, as too much was being passed to the caller. Changing it to take the `vid`s into account, we get the effect that if only some part of the locals is passed, relationality is lost between different locals of the caller. I now changed it such that if one local is reachable, all relevant locals are passed.

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: Going to a set of `varinfo`s is quite confusing though because one would need to account for `T`, so I think that this is the way clearer alternative, despite being a bit odd.

The alternative would be:

~~~OCaml
    let module VS = SetDomain.ToppedSet (CilType.Varinfo) (struct let topname = "All" end) in
    let reachable_vinfos e = try List.fold_left (fun s (v,_) -> VS.add v s) (VS.empty ()) (Queries.LS.elements (ctx.ask (ReachableFrom e))) with SetDomain.Unsupported _ -> VS.top ()  in
    let reachable_from_args = List.fold_left (fun ss s -> VS.union ss (rechable_vinfos s) ) (VS.empty ()) args in
~~~

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: Same reasoning as above.

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: I am skeptical if we would benefit here, I added a TODO now.

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -344,14 +407,18 @@ struct
                 | _ -> acc
               ) []
             in
-            List.fold_left invalidate_one st globals
-          )
+            List.fold_left (invalidate_one ask ctx) st globals)
```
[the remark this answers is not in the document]
michael-schwarz: true!

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -373,18 +373,18 @@ struct
...
-    | `Unknown "__goblint_check" -> st
-    | `Unknown "__goblint_commit" -> st
-    | `Unknown "__goblint_assert" -> st
-    | `ThreadJoin (id,retvar) ->
+    | Assert expression, _ -> st
+    | Unknown, "__goblint_check" -> st
+    | Unknown, "__goblint_commit" -> st
+    | Unknown, "__goblint_assert" -> st
+    | ThreadJoin { thread = id; ret_var = retvar }, _ ->
+      (* nothing to invalidate as only arguments that have their AddrOf taken may be invalidated *)
```
sim642, line 398: I guess this comment creeped back in with the merge, but was supposed to be removed by this PR.

src/analyses/apron/apronAnalysis.apron.ml
sim642: I guess it's fine for now as-is so we can quickly move to see if this helps us find any more relational invariants.

Totally off-topic, but `exists` on a topped set being `true` is actually weird, since it claims that `exists (fun _ -> false)` is also true :upside_down_face:

sim642 (approved):

sim642:
I just tried running the relational traces bench on this branch and they all seem to immediately crash with:
```
Fatal error: exception IntDomain.IncompatibleIKinds("ikinds char and int are incompatible. Values: (Unknown int([-7,7]),not {} ([-7,7])) and (Unknown int([-31,31]),not {} ([-31,31]))")
```

sim642:
Only on this branch, the initial problem is that dereferencing doesn't exclude the unknown pointer. This patch fixes that:
```patch
diff --git a/src/analyses/apron/apronAnalysis.apron.ml b/src/analyses/apron/apronAnalysis.apron.ml
index 3ac185d7a..5c5b494ef 100644
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -158,8 +158,8 @@ struct
       | Lval (Var v, off) -> Lval (Var v, off)
       | Lval (Mem e, NoOffset) ->
         (match ask (Queries.MayPointTo e) with
-         | (`Lifted _) as r when (Queries.LS.cardinal r) = 1 ->
-           let lval = Lval.CilLval.to_lval (Queries.LS.choose r) in
+         | a when not (Queries.LS.is_top a || Queries.LS.mem (dummyFunDec.svar, `NoOffset) a) && (Queries.LS.cardinal a) = 1 ->
+           let lval = Lval.CilLval.to_lval (Queries.LS.choose a) in
            Lval lval
          (* It would be possible to do better here, exploiting e.g. that the things pointed to are known to be equal *)
          (* see: https://github.com/goblint/analyzer/pull/742#discussion_r879099745 *)
```

After fixing that, a `TypeOfError` appears because when responding to `EvalInt` queries, the apron analysis tries to get the ikind (through the type) of the expression, but that fails for our wonderful alloc variables that have void type, but may have offsets on them, e.g. `(alloc@sid:138)#in[(int )"unknown"]`.

sim642:
Ok, this goes deep, but I think I've managed to fix all the following crashes. I'll just go ahead and commit my fixes to this branch.

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -300,14 +352,40 @@ struct
...
+        else
+          Some (Queries.LS.join vs st)
+    in
+    List.fold_right reachable es (Some (Queries.LS.empty ()))
+
+
+  let forget_reachable ctx st es =
+    let ask = Analyses.ask_of_ctx ctx in
+    match reachables ask es with
+    | None -> D.top ()
```
sim642, line 381: On knot from bench it seems like this `D.top ()` might have caused a crash since `DHetero.top` is defined to fail. I think it might have to be `D.top_env` (using the previous environment), because we still need to have the environment but forget everything about it.

sim642: Actually one can't just do that because something needs to happen with the `priv` component as well. Maybe we can just use `invalidate_one` on all the variables in the environment then? That does a self-write which privatization is aware of.

michael-schwarz: The question is do we have a convenient way to get all the CIL(!) variables in the environment here? `invalidate_one` expects a CIL Lval...

sim642: I committed a fix that at least avoids the crash. The primed Arg variables and the Return variable aren't actually reachable so shouldn't be invalidated, therefore `to_cil_varinfo` should suffice.

PR #744, opened 2022-05-24
Add YAML witness output

### LINKED ISSUE #741: Option to dump asserts in YAML-based witness format
In a discussion between some Munich Goblint folks and @MartinSpiessl and others from the CPAChecker team on Friday we identified that as a first step of a possible co-operation (e.g. combining the strengths of Goblint as an Abstract Interpreter and Model Checkers such as CPAChecker), it would be helpful for us to produce the YAML-based invariant format (https://github.com/sosy-lab/sv-witnesses).

Since no complicated re-construction of the call-stack is needed here, I'd suggest we implement this as an additional output format for the assert instrumentation from https://github.com/goblint/analyzer/pull/157 as opposed to integrating it with all the weirdness we have for producing automaton-based witnesses.

sim642:
Closes #741.

Adds YAML witnesses with options under `witness.yaml`. These are separate from any of the SV-COMP or GraphML witness options, since we want these witnesses also outside of SV-COMP. However, outside of SV-COMP, the resulting witnesses are "invalid" since they're missing the `specification` key that we normally simply don't have.

### TODO
- [ ] Unify options about where to emit invariants with assert transformation.

### DIFF
--- a/src/witness/yamlWitness.ml
+++ b/src/witness/yamlWitness.ml
@@ -12,6 +12,17 @@ module Make
     (GHT : BatHashtbl.S with type key = EQSys.GVar.t) =
 struct
 
+  module NH = BatHashtbl.Make (Node)
+
+  (* copied from Constraints.CompareNode *)
+  let join_contexts (lh: Spec.D.t LHT.t): Spec.D.t NH.t =
+    let nh = NH.create 113 in
+    LHT.iter (fun (n, _) d ->
+        let d' = try Spec.D.join (NH.find nh n) d with Not_found -> d in
+        NH.replace nh n d'
+      ) lh;
+    nh
+
   let write lh gh =
     let yaml_creation_time = `String (TimeUtil.iso8601_now ()) in
     let yaml_producer = `O [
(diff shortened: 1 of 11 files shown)


src/witness/yamlWitness.ml
```diff
@@ -0,0 +1,114 @@
...
+              scope=Node.find_fundec n;
+              i = -1;
+              lval=None;
+              offset=Cil.NoOffset;
+              deref_invariant=(fun _ _ _ -> Invariant.none) (* TODO: should throw instead? *)
+            }
+          in
+          begin match Spec.D.invariant context local with
+            | Some inv ->
+              let inv = InvariantCil.exp_replace_original_name inv in
```
michael-schwarz, line 77: You might want to do something like `pullOutCommonConjuncts` as I did in the assert transformation here?

michael-schwarz:
> and thus move/rename the `trans.assert` options there

Yes, I think that would make sense!

> `trans.assert.only-at-locks` also apply to these YAML witnesses

Yes, that is a sensible option! However, I am a bit dubious as to how you would achieve this when the dumping of yaml witnesses is not formulated as a transformation of the AST where the AST is easily at hand.

src/witness/yamlWitness.ml
michael-schwarz: The issue is that the asserts become super-unreadable as soon as you have multiple paths, and pulling common conjuncts out makes it a lot nicer to inspect.

sim642: Yup, I just realized I'll need to do it anyway to work around https://github.com/goblint/cil/issues/94.

michael-schwarz (approved):

PR #745, opened 2022-05-25
Add YAML witness validation by invariant checking

sim642:
This is on top of #744.

The validation implemented here is quite primitive: the YAML witness is parsed and the invariants are checked on our final solution using the `EvalInt` query. It is very similar to the expression evaluation transformation used by semantic search. For each invariant, a corresponding message is printed, similarly to the messages printed for in-program assertions.

This primitive form of witness validation is what could be used for relational traces precision benchmarking: one privatization generates the YAML witness and another privatization can then try to prove the same invariants. ~~It doesn't really work for Apron analysis though, because the way `Formatcil` parses expressions produces invalid ASTs for the kind of expressions currently generated by Apron analysis: https://github.com/goblint/cil/issues/95.~~

~~A more reliable solution would be to do something about `Formatcil`, but a quick workaround would also be to change Apron analysis to output expressions that it can directly parse, e.g. by inserting a few extra casts or whatnot.~~

By switching from `Formatcil` to Frontc via https://github.com/goblint/cil/pull/97, it works for Apron analysis and the additional explicit casts might even be unnecessary now.
Moreover, it allows `&&` and `||` to be used in invariants. During normal parsing we ask CIL to convert them to `if` statements, but for parsing these invariants, that is temporarily disabled. I went over a bunch of `Cil.exp` pattern matchings to make them explicit (no wildcard case) such that it would be easier to find where, e.g. `Question` is being/should be handled.

#### Singlethreaded
```console
./regtest.sh 01 04 --enable witness.yaml.enabled
./regtest.sh 01 04 --set witness.yaml.validate witness.yml
```
#### Base privatization
```console
./regtest.sh 13 01 --enable witness.yaml.enabled
./regtest.sh 13 01 --set witness.yaml.validate witness.yml
```
#### Apron privatization
```console
./regtest.sh 36 89 --enable witness.yaml.enabled
./regtest.sh 36 89 --set witness.yaml.validate witness.yml
```
#### Using logical operators
```console
./regtest.sh 01 04 --enable witness.yaml.enabled --enable ana.int.enums --disable witness.invariant.split-conjunction
./regtest.sh 01 04 --set witness.yaml.validate witness.yml
```
PS. This is an interesting example because we can validate an invariant from enums domain without enabling enums during validation!

### TODO
- [x] Merge https://github.com/goblint/cil/pull/96.
- [x] Merge https://github.com/goblint/cil/pull/97.
- [x] Update opam pin after CIL merges.


sim642:
> A more reliable solution would be to do something about `Formatcil`, but a quick workaround would also be to change Apron analysis to output expressions that it can directly parse, e.g. by inserting a few extra casts or whatnot.

I did that to `IntDomain` invariants, but Apron analysis invariants already had it. The mismatching thing there was the ikind of 0 used in the constraint.
That doesn't still work because `Formatcil` does even more stupid things: when it parses `==`, then the result type of the comparison is the type of the arguments (`long long` for Apron), but it should always be `int` by the standard.

src/analyses/tutorials/taint.ml
```diff
@@ -42,7 +42,7 @@ struct
     | CastE (_,e)
     | UnOp (_,e,_) -> is_exp_tainted state e
     | SizeOf _ | SizeOfStr _ | Const _  | AlignOf _ | AddrOfLabel _ -> false
-    | Question _ -> failwith "should be optimized away by CIL"
+    | Question (b, t, f, _) -> is_exp_tainted state b || is_exp_tainted state t || is_exp_tainted state f
```
michael-schwarz, line 45: CIL still compiles them away, no?

src/analyses/varEq.ml
```diff
@@ -135,8 +135,11 @@ struct
       | SizeOfE _
       | SizeOfStr _
       | AlignOf _
-      | AlignOfE _ -> false
-      | UnOp (_,e,_) -> type_may_change_t e bt
+      | AlignOfE _
+      | AddrOfLabel _ -> false (* TODO: some may contain exps? *)
```
michael-schwarz, line 139: What do you mean by the comment? These are for handling of computer gotos, so I'm fairly certain these do not occur inside Goblint.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -335,7 +335,7 @@ struct
 
 
   let cil_exp_of_lincons1 fundec (lincons1:Lincons1.t) =
-    let zero = Cil.zero in
+    let zero = Cil.kinteger ILongLong 0 in
```
michael-schwarz, line 338: Nice catch!

src/analyses/varEq.ml
sim642: There's a list of cases combined here, not just `AddrOfLabel`, and some of those cases, e.g. `AlignOfE` do contain subexpressions as well, but the checks here (and in many other places) used to just shove it all under a default wildcard case, forgetting occurrence checks on some subexpressions.

michael-schwarz (approved):
Nice!

src/cdomains/valueDomain.ml
```diff
@@ -142,7 +142,7 @@ struct
 
   let rec init_value (t: typ): t = (* top_value is not used here because structs, blob etc will not contain the right members *)
     match t with
-    | t when is_mutex_type t -> `Top
+    | t when is_mutex_type t -> `Bot (* use `Bot instead of `Top to avoid unknown value escape warnings *)
```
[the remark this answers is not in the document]
sim642: That's a good point. I'm not sure about the details of the different struct domains, but in a couple of places there's some explicit workaround to this since arrays (and blobs) also initialize to `Bot`: https://github.com/goblint/analyzer/blob/45b2269ca56b1e6244f2c7bbb480cb49fb48a07c/src/cdomains/structDomain.ml#L159-L167

Maybe a safer alternative would be like a `Struct` with no fields? Of course introducing a new `Mutex` variant to the value domain would be the safest, but I suspect it requires similar hacks as for `Thread` since there's some unknown structure still being manipulated (and that structure probably is different for OSX).

The only reason I put this workaround here is to avoid large amounts of invariants about `pthread_mutex_t` contents, which are actually all incorrect, since base analysis never updates them from their initialized values. Hiding them from g2html has been an idea for a while as well, but so far it was only done for accesses, not base.
I can also revert this small change here and split the hiding of mutex contents into a separate PR for a more reliable and involved solution.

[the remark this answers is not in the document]
michael-schwarz: > I can also revert this small change here and split the hiding of mutex contents into a separate PR for a more reliable and involved solution.

Yes, I think this might be the prudent thing to do!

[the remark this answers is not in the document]
sim642: Reverted here and moved to #746.

src/witness/yamlWitness.ml
```diff
@@ -117,3 +154,195 @@ struct
...
+              let result = LvarS.fold (fun ((n, _) as lvar) (acc: VR.t) ->
+                  let d = LHT.find lh lvar in
+                  let fd = Node.find_fundec n in
+                  let vars = fd.sformals @ fd.slocals @ global_vars in
+
+                  let genv = Cabs2cil.genvironment in
+                  let env = Hashtbl.copy genv in
+                  List.iter (fun (v: Cil.varinfo) ->
+                      Hashtbl.replace env v.vname (Cabs2cil.EnvVar v, v.vdecl)
+                    ) (fd.sformals @ fd.slocals);
```
michael-schwarz, line 283: Did you test with locals that got renamed because several locals of the same name (and different scopes) existed in the function?

sim642: Good point, I didn't think about it here, but I think there's not much to do about it either. On the witness generation side, there's `InvariantCil.exp_replace_original_name` that maps things to original names. But here for parsing we'd somehow need to know, which of the variables with the same original name was in scope at a particular `location` inside the function _after_ they've been all pulled up and renamed.

As I mentioned in https://github.com/goblint/cil/pull/97, the normal `Cabs2cil` uses tons of global variables to keep track of things and these block scopes and alpha renaming tables (and undos in alpha renaming tables) are among those. But all those structures are mutated, so we cannot go back to a particular location and have the exact state of those internal globals that normally keep track of this. One way would be to keep the CABS of all functions around and redo `Cabs2cil` for the function where the invariant belongs to, so it reconstructs that state, such that we can convert the invariant expression exactly as it would have been at a particular location and abort `Cabs2cil`, but that seems super ugly.

But here's a maybe radical idea: we make CIL not do those renames at all and just continue with a single `fundec` containing multiple local `varinfo`s with same `vname`. Since we identify variables by `vid`, not `vname`, this should have no effect on the semantics, only results presentation (which is nicer with less renaming anyway). And somehow also eliminate the pulling up of all locals from all block scopes, such that the scopes are also later observable.

Anyway, tackling that whole issue is probably too much to do right here, but for the time being, a variable name in the expression would just find the first `varinfo` that has it unrenamed. All the others would have names with suffixes, which won't be used unless you directly also use that name in the invariant, but that would violate the semantics of it being insertable as an assertion.

sim642: I extracted this problem to #747 for further discussion and future fixes, so this can be merged.

PR #746, opened 2022-05-27
Hide mutex structure contents and standard globals in base analysis

sim642:
Split off from https://github.com/goblint/analyzer/pull/745#discussion_r883347346.

@michael-schwarz:
> I'm a bit worried about having `Bot` here, as it might make some comparisons of the struct members return false, potentially making both branches unreachable.

@sim642:
> Maybe a safer alternative would be like a `Struct` with no fields? Of course introducing a new `Mutex` variant to the value domain would be the safest, but I suspect it requires similar hacks as for `Thread` since there's some unknown structure still being manipulated (and that structure probably is different for OSX).

---

This is to avoid large amounts of invariants about `pthread_mutex_t` contents, which are actually all incorrect, since base analysis never updates them from their initialized values. Hiding them from elsewhere (e.g. g2html) would improve readability as well, but so far such mutex type ignoring was only done for accesses, not base.

Additionally, it (optionally) hides unknown globals from standard headers, e.g. `__tzname`, etc, such that they wouldn't always clutter local states.

### DIFF
--- a/src/cdomains/valueDomain.ml
+++ b/src/cdomains/valueDomain.ml
@@ -854,6 +854,7 @@ struct
       let mu = function `Blob (`Blob (y, s', orig), s, orig2) -> `Blob (y, ID.join s s',orig) | x -> x in
       let r =
       match x, offs with
+      | _, _ when is_mutex_type t -> `Top (* hide mutex structure contents, not updated anyway *)
       | `Blob (x,s,orig), `Index (_,ofs) ->
         begin
           let l', o' = shift_one_over l o in
--- a/src/cdomains/valueDomain.ml
+++ b/src/cdomains/valueDomain.ml
@@ -133,7 +133,7 @@ struct
 
   let rec init_value (t: typ): t = (* top_value is not used here because structs, blob etc will not contain the right members *)
     match t with
-    | t when is_mutex_type t -> `Top
+    | t when is_mutex_type t -> `Bot (* use `Bot instead of `Top to avoid unknown value escape warnings *)
     | TInt (ik,_) -> `Int (ID.top_of ik)
     | TPtr _ -> `Address AD.top_ptr
     | TComp ({cstruct=true; _} as ci,_) -> `Struct (Structs.create (fun fd -> init_value fd.ftype) ci)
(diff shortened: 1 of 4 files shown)


michael-schwarz (commented):
LGTM! We might want to enhance this mutex data type with information such as whether it is initialized or whether it is recursive at some later point, but this already is a step in the right direction!

michael-schwarz (approved):

src/cdomains/valueDomain.ml
```diff
@@ -170,11 +175,13 @@ struct
     | `Array x -> CArrays.is_top x
     | `Blob x -> Blobs.is_top x
     | `Thread x -> Threads.is_top x
+    | `Mutex -> false
```
jerhard, line 178: Same question as above.

src/cdomains/valueDomain.ml
```diff
@@ -128,12 +131,13 @@ struct
     | `Array x -> CArrays.is_bot x
     | `Blob x -> Blobs.is_bot x
     | `Thread x -> Threads.is_bot x
+    | `Mutex -> false
```
jerhard, line 134: Should the value in a singleton lattice not both be considered `bottom` and `top` at the same time? Or is this `false` needed for some specific reason?

src/util/options.schema.json
```diff
@@ -1364,6 +1364,12 @@
             "Sets the unrolling factor for the loopUnrollingVisitor.",
           "type": "integer",
           "default": 0
+        },
+        "hide-std-globals": {
+          "title": "exp.hide-std-globals",
+          "description": "Hide standard extern globals from cluttering local states, e.g. stdout.",
```
jerhard, line 1370:
```suggestion
          "description": "Hide standard extern globals (e.g. `stdout`) from cluttering local states.",
```
Would changing the description like make it a bit easier to parse?

jerhard (changes requested):

src/cdomains/valueDomain.ml
sim642: I think I initially worried that having `is_bot_value Mutex = true` would make some states dead because one variable has a bottom value. I changed both to `true` now and tests indicate no issue with that, so it makes more sense indeed.

jerhard (approved):

PR #748, opened 2022-05-31
Store all incremental data into single file

### LINKED ISSUE #357: Marshal all data into single file

jerhard:
With this PR, the incremental data is stored to a single file.

Now, incremental data is kept in-memory, maintained by `Serialize.Cache`. Data stored on disc can be loaded into memory with `load_data`, and stored from the cache to disk with `store_data`. 
Retrieving the data from there can be done using a GADT, giving some type safety for the stored data with static type. As the the types for solver and analysis data are determined at run time, they are still kept as `Obj.t` in the cache. 

Closes #357.

### DIFF
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -308,7 +308,7 @@ struct
       if get_string "load_run" <> "" then
         Some (Serialize.unmarshal Fpath.(v (get_string "load_run") / "spec_marshal"))
       else if Serialize.results_exist () && get_bool "incremental.load" then
-        Some (Obj.obj (Serialize.Cache.(get_data AnalysisData)))
+        Some (Serialize.Cache.(get_data AnalysisData))
       else
         None
     in
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -677,7 +677,7 @@ struct
       Serialize.marshal marshal Fpath.(v save_run / "spec_marshal")
     );
     if get_bool "incremental.save" then (
-      Serialize.Cache.(update_data AnalysisData (Obj.repr marshal));
+      Serialize.Cache.(update_data AnalysisData marshal);
       Serialize.Cache.store_data ()
     );
     if get_bool "dbg.verbose" && get_string "result" <> "none" then print_endline ("Generating output: " ^ get_string "result");
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -469,7 +469,7 @@ struct
           Goblintutil.should_warn := get_string "warn_at" = "early" || gobview;
           let (lh, gh), solver_data = Stats.time "solving" (Slvr.solve entrystates entrystates_global) startvars' in
           if GobConfig.get_bool "incremental.save" then
-            Serialize.Cache.(update_data (SolverData (Obj.repr solver_data)));
+            Serialize.Cache.(update_data SolverDataRequest (Obj.repr solver_data));
           if save_run_str <> "" then (
             let save_run = Fpath.v save_run_str in
             let analyses = Fpath.(save_run / "analyses.marshalled") in
@@ -677,7 +677,7 @@ struct
       Serialize.marshal marshal Fpath.(v save_run / "spec_marshal")
     );
     if get_bool "incremental.save" then (
-      Serialize.Cache.(update_data (AnalysisData (Obj.repr marshal)));
+      Serialize.Cache.(update_data AnalysisDataRequest (Obj.repr marshal));
       Serialize.Cache.store_data ()
     );
     if get_bool "dbg.verbose" && get_string "result" <> "none" then print_endline ("Generating output: " ^ get_string "result");
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -469,7 +469,7 @@ struct
           Goblintutil.should_warn := get_string "warn_at" = "early" || gobview;
           let (lh, gh), solver_data = Stats.time "solving" (Slvr.solve entrystates entrystates_global) startvars' in
           if GobConfig.get_bool "incremental.save" then
-            Serialize.Cache.(update_data SolverData (Obj.repr solver_data));
+            Serialize.Cache.(update_data SolverData solver_data);
           if save_run_str <> "" then (
             let save_run = Fpath.v save_run_str in
             let analyses = Fpath.(save_run / "analyses.marshalled") in
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -304,7 +304,7 @@ struct
 
     GU.global_initialization := true;
     GU.earlyglobs := get_bool "exp.earlyglobs";
-    let marshal =
+    let marshal: Spec.marshal option =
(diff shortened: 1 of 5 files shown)


src/framework/control.ml
```diff
@@ -304,11 +304,11 @@ struct
 
     GU.global_initialization := true;
     GU.earlyglobs := get_bool "exp.earlyglobs";
-    let marshal =
+    let marshal: Spec.marshal option =
       if get_string "load_run" <> "" then
         Some (Serialize.unmarshal Fpath.(v (get_string "load_run") / "spec_marshal"))
       else if Serialize.results_exist () && get_bool "incremental.load" then
-        Some (Serialize.load_data Serialize.AnalysisData)
+        Some (Obj.obj (Serialize.Cache.(get_data AnalysisDataRequest)))
```
sim642, line 311: Might be better to move the `Obj.obj` inside `get_data`, so from the outside, it is all type-safe. At least previously `Control` didn't have to do those dangerous conversions itself.

src/incremental/serialize.ml
```diff
@@ -1,10 +1,7 @@
 open Prelude
 
 (* TODO: GoblintDir *)
-let version_map_filename = "version.data"
-let cil_file_name = "ast.data"
-let solver_data_file_name = "solver.data"
-let analysis_data_file_name = "analysis.data"
+let incremental_data_file_name = "analysis.data"
 let results_dir = "results"
 let results_tmp_dir = "results_tmp"
```
sim642, line 6: Is the whole business with two directories and renaming them still necessary?

src/incremental/serialize.ml
```diff
@@ -43,49 +40,78 @@ let results_exist () =
...
-    | AnalysisData -> server_analysis_data := Some (Obj.repr data)
-    | _ -> ()
-  else (
+(** Module to cache the data for incremental analaysis during a run, before it is stored to disk, as well as the server mode *)
+module Cache = struct
+  type t = {
+    solver_data: Obj.t option ref;
+    analysis_data: Obj.t option ref;
+    version_data: MaxIdUtil.max_ids option ref;
+    cil_file: Cil.file option ref;
```
sim642, line 49: Instead of `ref`s, the fields could simply be `mutable`.

Also, why are they all `option`s? When would we ever have incremental data that doesn't have all four components present? I think we always need all of them (or have none of them), so the whole record could be optional.

src/incremental/serialize.ml
```diff
@@ -43,49 +40,78 @@ let results_exist () =
...
+    | CilFileRequest : Cil.file data_query
+    | VersionDataRequest : MaxIdUtil.max_ids data_query
+    | AnalysisDataRequest : Obj.t data_query
+
+  (* Data type to pass incremental data into the cache *)
+  type incremental_data =
+    | SolverData of Obj.t
+    | CilFile of Cil.file
+    | VersionData of MaxIdUtil.max_ids
+    | AnalysisData of Obj.t
```
sim642, line 71: This duplication seems unnecessary. I think just `data_query` would suffice and `update_data` could have a type like `type a. a data_query -> a -> unit`.

src/incremental/serialize.ml
jerhard: >Instead of refs, the fields could simply be mutable.

Done.

>Also, why are they all options? When would we ever have incremental data that doesn't have all four components present? I think we always need all of them (or have none of them), so the whole record could be optional.

The issue is that during the initial analysis, these fields would not all become available at the same time. Namely, the Cil file and the max ids would become available after parsing (in `Maingoblint`), while the solver and analysis data only becomes availabe after solving the constraint system. 

If one were to make the fields in the record non-optional, one would have to drag the cil file and max ids around until all of the data is available, so that one can set them in the record here.

src/incremental/serialize.ml
jerhard: No, actually not.

src/framework/control.ml
sim642: Maybe if instead of `| SolverData : Obj.t data_query` it is defined as `| SolverData : _ data_query` (that particular query being of any type, it might work. Of course it cannot have wildcard type in the record, so it'll have to stay `Obj.t` there, but the `Obj.obj`/`Obj.repr` could be used in the corresponding cases in `get_data` and `update_data`.

src/incremental/serialize.ml
sim642: > these fields would not all become available at the same time

Right. At some point this bothered me before, but I think we can keep them optional for now.

At some point it might be useful to have them all in an immutable record that's read and created as a whole, because storing them at different times maybe can cause some inconsistencies during server mode or such. For example, if the new file is parsed and stored, the solving starts but is aborted, then the cache possibly has inconsistent state.

src/framework/control.ml
jerhard: Good idea, that seems to work!

src/framework/control.ml
```diff
@@ -469,7 +469,7 @@ struct
           Goblintutil.should_warn := get_string "warn_at" = "early" || gobview;
           let (lh, gh), solver_data = Stats.time "solving" (Slvr.solve entrystates entrystates_global) startvars' in
           if GobConfig.get_bool "incremental.save" then
-            Serialize.store_data solver_data Serialize.SolverData;
+            Serialize.Cache.(update_data SolverData (Obj.repr solver_data));
```
sim642, line 472: The same `_ data_query` thing could be done for `SolverData` as well I think.

src/framework/control.ml
```diff
@@ -677,8 +677,8 @@ struct
       Serialize.marshal marshal Fpath.(v save_run / "spec_marshal")
     );
     if get_bool "incremental.save" then (
-      Serialize.store_data marshal Serialize.AnalysisData;
-      Serialize.move_tmp_results_to_results ()
+      Serialize.Cache.(update_data AnalysisData (Obj.repr marshal));
```
sim642, line 680: Since `Obj.repr` is inside `update_data` now, it shouldn't be necessary here any more:
```suggestion
      Serialize.Cache.(update_data AnalysisData marshal);
```
It's just that the fully generic GADT type parameter also allows passing in `Obj.t`, so the compiler wouldn't complain about this being unnecessary.

jerhard: Good catch!

sim642 (approved):
I didn't test this, but the code looks like it should.

PR #752, opened 2022-06-06
Add experimental generation and validation of YAML witness invariants with preconditions

sim642:
Adds the experimental YAML witness entry type `precondition_loop_invariant`, which is like `loop_invariant`, but has a second `precondition` expression to match contexts.
Current validation semantics is that the invariant is only checked for contexts where the precondition _must_ hold.

Additionally refactors all the YAML witness type definitions etc to be more type-safe and extensible.


sim642:
Although the entry type for invariants with preconditions is currently unofficial, it would be good to get this merged for its refactoring of the YAML witnesses. The refactoring makes it easy to work with different entry types, possibly including `location_invariant`, which is to be added as well.

src/witness/yamlWitnessType.ml
```diff
@@ -0,0 +1,380 @@
...
+end
+
+module PreconditionLoopInvariantCertificate =
+struct
+  include LoopInvariantCertificate
+  let entry_type = "precondition_loop_invariant_certificate"
+end
+
+(* module type EntryType =
+sig
```
jerhard, line 273: Is it intentional to keep this commented out code?

jerhard (commented):
We discussed this here  a bit in Munich, with @michael-schwarz, @stilscher.
It probably makes sense to implement this grouping of preconditions before merging this, as otherwise the produced precondition-witnesses could be misleading. 

For a function, one would first generate all the invariant expressions for the contexts, and then check which of the start states may satisfy the condition and group them. Subsequently one would generate the precondition-invariants for these groups of contexts.

src/witness/yamlWitness.ml
```diff
@@ -159,7 +179,42 @@ struct
...
+            offset=Cil.NoOffset;
+            deref_invariant=(fun _ _ _ -> Invariant.none) (* TODO: should throw instead? *)
+          }
+          in
+          begin match Spec.C.invariant context c, Spec.D.invariant context local with
+            | Some c_inv, Some inv ->
+              let loc = Node.location n in
+              (* TODO: group by precondition *)
+              let c_inv = InvariantCil.exp_replace_original_name c_inv in (* cannot be split *)
+              let invs = WitnessUtil.InvariantExp.process_exp inv in
```
jerhard, line 200: Is there a particular advantage to divide conjunctions into separate expressions?

src/witness/yamlWitnessType.ml
sim642: Not necessarily. It was an initial attempt to have less duplication and more type safety through a GADT, but for some reason this didn't work out (or at least as nicely).

src/witness/yamlWitnessType.ml
```diff
@@ -0,0 +1,330 @@
+module Producer =
+struct
+  type t = {
+    name: string;
+    version: string;
+    (* TODO: configuration *)
+    command_line: string;
+    (* TODO: description *)
```
jerhard, line 8: Are these todos that need to be taken care of before merging this PR?

jerhard (approved):
Looks good!

src/witness/yamlWitnessType.ml
sim642: No, they were also present (just in a different place) before this PR. Those are just optional metadata fields that we currently don't need/use.

PR #756, opened 2022-06-13
Fix unsound Apron dummy privatization

sim642:
As discussed on Slack.

### Changes
1. Implement `enter_multithreaded` for dummy privatization.
2. Rename dummy privatization to "top" to clarify that it knows nothing about globals.
3. Clean up `PARAM`s of tests that previously used dummy privatization, but didn't actually have any threads, arrays or need for TD3 explicitly. I guess these parameters were at some point blindly copied from somewhere else.

### DIFF
--- a/src/analyses/apron/apronPriv.apron.ml
+++ b/src/analyses/apron/apronPriv.apron.ml
@@ -44,8 +44,9 @@ module type S =
     val finalize: unit -> unit
   end
 
-
-module Dummy: S = functor (AD: ApronDomain.S2) ->
+(** Top privatization, which doesn't track globals at all.
+    This is unlike base's "none" privatization. which does track globals, but doesn't privatize them. *)
+module Top: S = functor (AD: ApronDomain.S2) ->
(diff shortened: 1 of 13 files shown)


src/analyses/apron/apronPriv.apron.ml
```diff
@@ -44,30 +44,73 @@ module type S =
...
+      )
+    in
+    {st with apr = apr_local}
+
+  let threadenter ask getg (st: apron_components_t): apron_components_t =
+    {apr = AD.bot (); priv = startstate ()}
 
-  let enter_multithreaded ask getg sideg st = st
-  let threadenter ask getg st = st
+  (* TODO: remove escaped locals after PR #742 *)
```
sim642, line 113: This PR is on top of master, which doesn't have escaping in apron analysis. This needs to be implemented together with #742. Not sure which we want to merge first.

michael-schwarz (approved):
Nice!

PR #757, opened 2022-06-14
Add `LOr` eval to base for address and enum set validation

sim642:
When generating witness invariants for enums inclusion set or known address set, the resulting expression is a disjunction of all the possible values. But when trying to validate our own witness, we couldn't confirm such invariant, because evaluating each equality in the disjunction separately gives unknown and so does their `LOr`.

This adds special logic to base analysis `eval_rv` to collect disjunctions of equalities together. If there's a common expression in all the sides of the equalities and all the other sides evaluate to definite addresses/ints, then a set-based subset/disjoint check can be used to evaluate such disjunction as a whole. It allows us to confirm some of our own disjunctive invariants.

This isn't complete though, because we generate, but still cannot validate `(p == & i && *p == 8) || (p == & j && *p == 7)`.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -769,6 +769,102 @@ struct
           let a1 = eval_rv a gs st c1 in
           let a2 = eval_rv a gs st c2 in
           evalbinop a st op t1 a1 t2 a2 typ
+      | BinOp (LOr, arg1, arg2, typ) as exp ->
+        let (let*) = Option.bind in
+        (* split nested LOr Eqs to equality pairs, if possible *)
+        let rec split = function
+          (* copied from above to support pointer equalities with implicit casts inserted *)
+          | BinOp (op, (CastE (t1, e1) as c1), (CastE (t2, e2) as c2), typ) when typeSig t1 = typeSig t2 && (op = Eq || op = Ne) ->
+            let te1 = Cilfacade.typeOf e1 in
+            let te2 = Cilfacade.typeOf e2 in
+            let both_arith_type = isArithmeticType te1 && isArithmeticType te2 in
+            let is_safe = VD.is_safe_cast t1 te1 && VD.is_safe_cast t2 te2 && not both_arith_type in
+            M.tracel "cast" "remove cast on both sides for %a? -> %b\n" d_exp exp is_safe;
+            if is_safe then ( (* we can ignore the casts if the casts can't change the value *)
+              let e1 = if isArithmeticType te1 then c1 else e1 in
+              let e2 = if isArithmeticType te2 then c2 else e2 in
+              Some [(e1, e2)]
+            )
+            else
+              Some [(c1, c2)]
+          | BinOp (Eq, arg1, arg2, _) ->
+            Some [(arg1, arg2)]
+          | BinOp (LOr, arg1, arg2, _) ->
+            let* s1 = split arg1 in
+            let* s2 = split arg2 in
+            Some (s1 @ s2)
+          | _ ->
+            None
+        in
+        (* find common exp from all equality pairs and list of other sides, if possible *)
+        let find_common = function
+          | [] -> assert false
+          | (e1, e2) :: eqs ->
+            let eqs_for_all_mem e = List.for_all (fun (e1, e2) -> CilType.Exp.(equal e1 e || equal e2 e)) eqs in
+            let eqs_map_remove e = List.map (fun (e1, e2) -> if CilType.Exp.equal e1 e then e2 else e1) eqs in
+            if eqs_for_all_mem e1 then
+              Some (e1, e2 :: eqs_map_remove e1)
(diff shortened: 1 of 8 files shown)


jerhard:
Can you add an example test case that now works with these changes?

src/analyses/base.ml
```diff
@@ -769,6 +769,102 @@ struct
...
+          (* copied from above to support pointer equalities with implicit casts inserted *)
+          | BinOp (op, (CastE (t1, e1) as c1), (CastE (t2, e2) as c2), typ) when typeSig t1 = typeSig t2 && (op = Eq || op = Ne) ->
+            let te1 = Cilfacade.typeOf e1 in
+            let te2 = Cilfacade.typeOf e2 in
+            let both_arith_type = isArithmeticType te1 && isArithmeticType te2 in
+            let is_safe = VD.is_safe_cast t1 te1 && VD.is_safe_cast t2 te2 && not both_arith_type in
+            M.tracel "cast" "remove cast on both sides for %a? -> %b\n" d_exp exp is_safe;
+            if is_safe then ( (* we can ignore the casts if the casts can't change the value *)
+              let e1 = if isArithmeticType te1 then c1 else e1 in
+              let e2 = if isArithmeticType te2 then c2 else e2 in
```
jerhard, line 785: Given that this is some non-trivial snippet, it would be nice to avoid code duplication and instead extracting a function for it.

src/analyses/base.ml
```diff
@@ -769,6 +769,102 @@ struct
...
+              | `Address a :: vs when AD.is_definite a ->
+                AD.union a (to_definite_ad vs)
+              | _ :: vs ->
+                to_definite_ad vs
+            in
+            let definite_ad = to_definite_ad vs in
+            if AD.leq a definite_ad then (* other sides cover common address *)
+              Some (`Int (ID.of_bool ik true))
+            else if AD.is_empty (AD.meet a definite_ad) then
+              Some (`Int (ID.of_bool ik false))
```
jerhard, line 832: Is this not unsound for `LOr`, in combination with discarding non-definite address in line 826?
Say we have the following expression, with `p`,`q` being pointers and `w`, `x`, `y`,`z` being variables of the corresponding type.
```C
p==&x|| p==q
```
Assume that possible values are: `p = {&w, &y, &z}` and `q = {&y, &z}`. Then the expression may evaluate to `true`. The meet of `a`  (==  {&w, &y, &z}) and `definite_ad = {&x}` would be empty, however; thus yielding a definite `false`. 

Using the join over `vs` in place of `definite_ad` should work here.

src/analyses/base.ml
```diff
@@ -769,6 +769,102 @@ struct
...
+              | _ :: vs ->
+                to_int_set vs
+            in
+            let* incl_list = ID.to_incl_list i in
+            let incl_set = BISet.of_list incl_list in
+            let int_set = to_int_set vs in
+            if BISet.leq incl_set int_set then (* other sides cover common int *)
+              Some (`Int (ID.of_bool ik true))
+            else if BISet.disjoint incl_set int_set then
+              Some (`Int (ID.of_bool ik false))
```
jerhard, line 852: Same issue as for the handling of addresses above.

jerhard:
>I wanted to, but it's not possible because during normal parsing CIL transforms `LOr` away, so currently such expressions can only arise from witnesses. 

Ah, right! I guess at some latter point in time one might consider using witness based tests, but that's not a priority.

src/analyses/base.ml
sim642: You're right! I added that disjoint check at the last moment without thinking it through. For simplicity I just removed both for now since refuting those witness invariants isn't the current goal. In the ints case it shouldn't be then based on just inclusion sets anyway, but could be more general, so I'll leave it for now.

sim642:
> Ah, right! I guess at some latter point in time one might consider using witness based tests, but that's not a priority.

I added tests for these using manually-written YAML witnesses and just `PARAM`. It's a bit awkward, since the invariants are in the witness file, but the expected results are in code comments, but it at least provides some testing for these now.

tests/regression/56-witness/01-base-lor-enums.c
```diff
@@ -0,0 +1,33 @@
...
+      break;
+    case 1:
+      x = 3;
+      break;
+    default:
+      x = 6;
+      break;
+  }
+  ; // SUCCESS (witness)
+  ; // UNKNOWN (witness)
```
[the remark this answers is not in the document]
sim642: The testing script doesn't care about the exclamation mark at all, so it works here as well. And you're right that it makes sense to use `UNKNOWN!` for these, especially since both also contain an invariant that holds, but the given `LOr` handling cannot prove. So I made these changes.

jerhard (approved):
Looks good!

PR #759, opened 2022-06-15
Replace `Printable.invariant` entirely with `Invariant` query

sim642:
When I originally introduced `Printable.invariant` for witness invariant generation, I naïvely added it as a function to all lattices (and printables!). At some point, also a corresponding query was introduced, but this PR completes the transition. This has an abundance of benefits:
1. Defining `Printable`s and `Lattice`s is less annoying, because `invariant` is no longer required.
2. All redundant and base analysis specific fields are removed from `Invariant.context`.
3. Regarding the removal of `deref_invariant` field, a more direct mutually recursive implementation in the base analysis avoids the complicated trick with recursively defined records.
4. Last but most importantly, it allows generating invariants, which aren't only from the local state. In particular, this will allow invariants, which also include relevant global state, e.g. for privatizations that don't eagerly copy global invariant to local state on `lock`.


src/domains/queries.ml
```diff
@@ -37,7 +37,32 @@ end
...
-module LiftedExp = Lattice.Flat(CilType.Exp)(struct let top_name = "Top" let bot_name = "Unreachable" end)
+module LiftedExp =
+struct
+  include Lattice.Flat(CilType.Exp)(struct let top_name = "Top" let bot_name = "Unreachable" end)
+  let to_invariant = function
+    | `Lifted e -> Some e
+    | `Bot | `Top -> None
+  let of_invariant = function
+    | Some e -> `Lifted e
+    | None -> `Top (* TODO: ??? *)
```
michael-schwarz, line 48: Why the `???`? When would this need to be called?

src/domains/queries.ml
```diff
@@ -37,7 +37,32 @@ end
...
+
+  let join x y =
+    match (x,y) with
+    | (`Top, _) -> `Top
+    | (_, `Top) -> `Top
+    | (`Bot, x) -> x
+    | (x, `Bot) -> x
```
michael-schwarz, line 55: delegate all cases except the `Lifted` `Lifted` one to `join`?

michael-schwarz, line 55: delegate all cases except the `Lifted` `Lifted` one to `join`?

src/domains/queries.ml
```diff
@@ -37,7 +37,32 @@ end
...
+    | (`Lifted x, `Lifted y) -> `Lifted (BinOp (LOr, x, y, intType))
+
+  let meet x y =
+    match (x,y) with
+    | (`Bot, _) -> `Bot
+    | (_, `Bot) -> `Bot
+    | (`Top, x) -> x
+    | (x, `Top) -> x
```
michael-schwarz, line 63: same here?

src/domains/queries.ml
sim642: When prototyping the Apron domain lifter, which compares with box to output only non-box invariants, I realized that I'll have to rework this here, because of the disjunction handling w.r.t. paths.
When one path doesn't provide an invariant (`Invariant.none`), it essentially means `true`, but another path does provide an invariant, then `Invariant.(||)` will currently only leave the second path's expression, which doesn't hold in the first path, so is unsound.

I'm going to merge the lifted expression lattice here into `Invariant` itself, such that the top and bottom invariants are properly distinguished there. Because as-is, `Invariant.none` corresponds to both a tautology and contradiction, which is a contradiction in itself! Essentially, `Invariant` should then be a fully symbolic lattice, but doesn't implement `leq` since that requires entailment checking. But `leq` isn't needed on these invariants anyway.

src/domains/queries.ml
sim642: As I realized in https://github.com/goblint/analyzer/pull/759#discussion_r898791795, this is nonsense because `None` is used for both tautologies and dead paths. This will disappear when I do the refactoring.

src/domains/invariant.ml
```diff
@@ -1,6 +1,38 @@
...
+  let bot () = zero (* false *)
+  let top () = one (* true *)
+  let is_bot _ = failwith "ExpLat: is_bot"
+  let is_top _ = failwith "ExpLat: is_top"
+
+  let leq _ _ = failwith "ExpLat: leq"
+  let pretty_diff () _ = failwith "ExpLat: pretty_diff"
+
+  let join x y = BinOp (LOr, x, y, intType)
+  let meet x y = BinOp (LAnd, x, y, intType)
```
michael-schwarz, line 16: Should there not be some type of simplification here? Or at least a handling of joining `x` and `x` to yield `x` instead of `x || x` or meeting `x` and `x` leading to `x` instead of `x&x`?

michael-schwarz:
Other than my small comment, I think this looks good!

src/domains/invariant.ml
sim642: I suppose there could, but this is more or less how the invariant combination happened before and that hasn't been an issue since we don't have multiple analyses outputting syntactically identical invariants. Even if that were the case, in any nontrivial case AC-rewriting would be needed to avoid such redundancy.

Since these operations are fully syntactic and `leq` isn't even implemented (nor needed), the lack of idempotence, etc shouldn't be a problem in any way. But for future reference I'll add a couple of comments here about this.

Issue #734, opened 2022-05-12
Interval Analysis for Floating-point values


Dudeldu:
# Issue

## Current Situation

Currently, Goblint is unable to analyze `double` values (and for that matter, `float` values as well - but we will focus on
`double` for now).

## Motivation

- floating-point models are quite complex and features such as rounding and special numbers (infinities, NaN , etc.) are not always understood by programmers.
- many (embedded) usecases for interaction with the real world (sensor input, motion output, etc.)
- prominent errors with floating points have already led to catastrophic behaviors, such as the Patriot missile accident \[1\]

## Goal

We aim to implement an interval analysis for `double` values. Therefore, we heavily rely on \[1\]. As is written there, roundings
and underflows are tolerated, but not overflows, divisions by zero or invalid operations. For these, we want goblint
to generate a warning or an error.

In a first approach, in order to be able to tackle the four different rounding modes and arbitrary precision of
intermediate results, we will implement a worst-case analysis as proposed in \[1\]. In other words, we won't care about
the given rounding mode and instead work conservatively with the intervals. Like this, we won't necessarily be able to
say that a program is correct, but we will be able to say if might be incorrect.

With this information we aim to provide checking of assertions and dead-branches and warnings for potential usages of invalid `double` values - namely NaN, +/- infinity.

## Example

```c
int main()
{
    double middle;
    double a = DBL_MAX;
    double b = DBL_MAX;

    middle = middle * 1.;  // Of course this should result in a warning, due to no information about middle and
                           // since it could therefore be NaN / infinity / neg_infinity

    middle = (a + b) / 2.; // Naive way of computing the middle, which (might) result in an overflow

    return middle - 100.   // Further uses of the overflown middle value should result in a warning
}
```

\[1\]: [https://www-apr.lip6.fr/~mine/publi/article-mine-esop04.pdf](https://www-apr.lip6.fr/~mine/publi/article-mine-esop04.pdf)

WIP [here](https://github.com/Dudeldu/analyzer)

michael-schwarz:
Nice, it's about time Goblint gets support for floating point numbers!

sim642:
I suppose the current goal is to add this to our base analysis's `ValueDomain`?

I'm just wondering, because the other angle to floating-point would be to do it in the apron analysis, since the Apron library also supports floating-point arithmetic (but we currently only use integer portions of it).

michael-schwarz:
> I suppose the current goal is to add this to our base analysis's `ValueDomain`?

[...]

PR #761, opened 2022-06-16
Interval Analysis for Floating-point values

Dudeldu:
Adds basic interval analysis for float and double values (Related to Issue: #734)

We implemented an interval analysis for floating-point  values. In order to capture operations leading to unrepresentable values and their rounding mode dependent representations in the concrete execution, all abstract operations overestimate the worst case interval by using the rounding mode that creates the largest possible interval on both bounds.

Introduces following settings:
- `ana.float.interval` : Enables the float interval domain (default: `disabled`)
    + arithmetic operations on floats (+, -, *, /)
    + casts between ints and floats (and vice versa)
    + builtins for `is_nan`, `is_normal`, `is_finite`, `sign`, `is_inf` (although we can not use the captured information for more precise information inside branching)
    + abstractions for different math functions like `cos`, `tan`, `sin` ,...
- `__attribute__((goblint_precision("no-float-interval")))`: to disable/enable the float domain on a node/function level
- `warn.float`: Adds warnings related to the analysis results (default: `enabled`)
    + warn on equality/inequality check on floats
    + warn on operations done on values that could be +/- Infinity or NaN
    + warn in places where where loos track of the exact information (the point where the interval no longer only contains one value)


michael-schwarz:
Nice, thanks for the PR! :rocket: I hope to find some time to review it this weekend!

I resolved the merge conflicts on the benchmarking machine, and started a run of all of sv-comp on that machine (2min timeout). I'll send you the results on slack.
I guess in the meanwhile, it would make sense to keep merging the upstream master into yours from time-to-time so they don't diverge too much. (nothing fancy like rebasing needed, we do this with our branches all the time as well).

tests/regression/56-floats/01-base.c
```diff
@@ -0,0 +1,32 @@
...
+{
+    double x, a = 2., b = 3. + 1;
+    float y, c = 2.f, d = 3.f + 1;
+
+    assert(x == 2.);  // UNKNOWN!
+    assert(y == 2.f); // UNKNOWN!
+
+    assert(a == 2.); // SUCCESS!
+    assert(a < 10.); // SUCCESS!
+    assert(a > 10.); // FAIL!
```
sim642, line 17: `SUCCESS` doesn't actually do anything, by default assertions are required to succeed.
And we don't use `FAIL` with exclamation, only `UNKNOWN` (where it means actually unknowable, not unknown because of our imprecision).

src/analyses/base.ml
```diff
@@ -1597,14 +1635,67 @@ struct
...
+      let meet_bin a' b'  = FD.meet a a', FD.meet b b' in
+      let meet_com oi = (* commutative *)
+        meet_bin (oi c b) (oi c a) in
+      let meet_non oi oo = (* non-commutative *)
+        meet_bin (oi c b) (oo a c) in
+      match op with
+      | PlusA  -> meet_com FD.sub
+      | Mult   ->
+        (* refine x by information about y, using x * y == c *)
+        let refine_by x y = if FD.is_exact y then FD.meet x (FD.div c y) else x
```
sim642, line 1648: When doing this refinement and others, have you considered which IEEE 754 floating point identities are actually guaranteed to hold? For example, their addition isn't even associative, so it could be quite likely that the identities used for these refinements aren't correct either. Especially since right now the analysis internally uses the same floating point numbers.

src/analyses/base.ml
```diff
@@ -1597,14 +1635,67 @@ struct
...
+    let rec inv_exp c_typed exp (st:store): store =
       (* trying to improve variables in an expression so it is bottom means dead code *)
-      if ID.is_bot c then raise Deadcode;
-      match exp with
-      | UnOp (LNot, e, _) ->
+      (
+        match c_typed with
+        | `Int c -> if ID.is_bot c then raise Deadcode
+        | `Float c -> if FD.is_bot c then raise Deadcode
+        | _ -> if VD.is_bot c_typed then raise Deadcode
```
sim642, line 1695: Wouldn't `is_bot_value` cover all of these?

sim642 (commented):
It's great to see some unit tests being added!

I'm leaving a bunch of small comments and a couple of big discussion points (that we might have to settle ourselves first).

src/cdomains/floatOps/stubs.c
```diff
@@ -0,0 +1,87 @@
+#include <stdio.h>
+#include <math.h>
+#include <float.h>
+#include <fenv.h>
+#include <assert.h>
+#include <caml/mlvalues.h>
+#include <caml/alloc.h>
```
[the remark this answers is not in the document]
jerhard: >It would be a lot safer to use MPFR to do all this, especially since it supports all the operations (arithmetic and standard library) with rounding mode argument on all of them: https://antoinemine.github.io/Apron/doc/api/ocaml/Mpfr.html. Also would allow analyzing long doubles.

The critical point seems to be that the computations in the abstract need not be performed on more precise float representations. From the [paper, page 6](https://www-apr.lip6.fr/~mine/publi/article-mine-esop04.pdf):
>This semantics frees the analyzer from the job of statically determining the rounding mode of the expressions and allows the analyzer to use, in the abstract, less precise formats that those used in the concrete (however, using a more precise format in the abstract remains unsound).

The abstract operations should be performed with the same precision as in the concrete. Is there a way to set the precision in MPFR? One would think so, but I didn't find anything regarding this.

src/cdomains/floatDomain.mli
```diff
@@ -0,0 +1,96 @@
...
+  val bot_of: Cil.fkind -> t
+
+  val ending : Cil.fkind -> float -> t
+  val starting : Cil.fkind -> float -> t
+
+  val minimal: t -> float option
+  val maximal: t -> float option
+
+  val is_exact : t -> bool
+  val precision : t -> Cil.fkind
```
jerhard, line 91: This `precision` function has different from what `int_precision` does, so it does should be renamed to something like `get_fkind`.

src/analyses/base.ml
```diff
@@ -694,6 +727,11 @@ struct
       | Const (CInt (num,ikind,str)) ->
         (match str with Some x -> M.tracel "casto" "CInt (%s, %a, %s)\n" (Cilint.string_of_cilint num) d_ikind ikind x | None -> ());
         `Int (ID.cast_to ikind (IntDomain.of_const (num,ikind,str)))
+      | Const (CReal (_, (FFloat | FDouble as fkind), Some str)) -> `Float (FD.of_string fkind str) (* prefer parsing from string due to higher precision *)
+      | Const (CReal (num, (FFloat | FDouble as fkind), None)) -> `Float (FD.of_const fkind num)
+      (* this is so far only for DBL_MIN/DBL_MAX as it is represented as LongDouble although it would fit into a double as well *)
```
jerhard, line 732: Can you elaborate, what is meant with this comment? In particular, what is meant with "this"? Is `LongDouble` referring to an OCaml module/type?

jerhard (commented):
Thanks for your PR, it has a pretty professional code style! I left some comments and questions below.

src/cdomains/floatOps/stubs.c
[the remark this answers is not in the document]
michael-schwarz: > These things shouldn't change often, but with OCaml 5.0 around the corner with a major overhaul of the language runtime, it might cause problems for us going forward.

The stubs we use here seem so basic, it is unlikely they'll break with OCaml 5.0. I'd be far more worried about advanced libraries such as apron w.r.t. to the bindings being broken.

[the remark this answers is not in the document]
michael-schwarz: I don't think adding `-no-float-const-prop` would be too much harm to us, we don't heavily rely on floats anywhere but here!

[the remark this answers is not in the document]
sim642: Just to have it documented here: last week at GobCon we decided to let the stubs remain for now and not block this PR. We'll have to figure out future directions ourselves later.

src/util/options.schema.json
```diff
@@ -1669,6 +1697,12 @@
           "type": "boolean",
           "default": true
         },
+        "float": {
+          "title": "warn.float",
+          "description": "float warnings",
+          "type": "boolean",
+          "default": false
```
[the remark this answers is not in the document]
Dudeldu: As we didn't want to break any current behaviour, we disabled all our introduced options by default so that the current behaviour of goblint does not change. However, we could enable the warnings as they are also additionally _guarded_ by the float analysis itself.

src/cdomains/floatDomain.mli
```diff
@@ -0,0 +1,96 @@
...
+
+  val minimal: t -> float option
+  val maximal: t -> float option
+
+  val is_exact : t -> bool
+end
+
+(* Only required for testing *)
+module F64Interval : FloatDomainBase
+module F32Interval : FloatDomainBase
```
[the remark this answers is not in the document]
Dudeldu: changed the misleading comment. Those are actually __exposed__ for testing

src/cdomains/floatDomain.ml
```diff
@@ -0,0 +1,646 @@
...
+  let pretty_diff () (x, y) =
+    Pretty.dprintf "%a instead of %a" pretty x pretty y
+
+  let bot () = failwith "no bot exists"
+
+  let is_bot _ = false
+
+  let top () = None
+
+  let is_top = Option.is_none
```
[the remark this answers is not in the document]
Dudeldu: Representing `top` in the float domain is not simply possible by using a max range interval (as done in the int domains). Therefore we had to introduce some kind of __special__ value for `top`, that covers the whole range of floating point numbers as well as the specials like +/- inf and Nan. Here we went with `None`. On the other side we have no representation of `bot` in our domain.

src/cdomains/floatOps/floatOps.ml
```diff
@@ -0,0 +1,104 @@
...
+let big_int_of_float f =
+  let x, n = Float.frexp f in
+  let shift = min 52 n in
+  let x' = x *. Float.pow 2. (Float.of_int shift) in
+  Big_int_Z.mult_big_int
+    (Big_int_Z.big_int_of_int64 (Int64.of_float x'))
+    (Big_int_Z.power_int_positive_int 2 (n - shift))
+
+module CDouble = struct
+  type t = float [@@deriving eq, ord, to_yojson]
```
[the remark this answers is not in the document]
Dudeldu: I would argue, that when we use the equality specified here, we are not actually interested in checking the equality of the underlying real numbers (what we warn the user about, when he uses those numbers in a program) but rather checking for exact equivalence of the bit-patterns or at least semantic equivalence according to IEEE 754 (e.g. `+0 == -0`). In the tests as well as other places this is the requested behaviour in our abstract domain, however one has to be careful not to misuse this equality

src/cdomains/floatOps/floatOps.ml
```diff
@@ -0,0 +1,104 @@
...
+  val div: round_mode -> t -> t -> t
+  val atof: round_mode -> string -> t
+end
+
+let big_int_of_float f =
+  let x, n = Float.frexp f in
+  let shift = min 52 n in
+  let x' = x *. Float.pow 2. (Float.of_int shift) in
+  Big_int_Z.mult_big_int
+    (Big_int_Z.big_int_of_int64 (Int64.of_float x'))
```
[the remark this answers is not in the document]
Dudeldu: As `x` is defined to be in the range `[0, 1]`, `x'` is definitely in the range `[0, 2^52]` which is representable by a `int64`. This conversion might loose precision, however this is the intended behaviour when casting from a float to an integer

src/cdomains/floatOps/floatOps.ml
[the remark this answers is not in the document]
sim642: That's true, it shouldn't hurt us. At worst we might have two very similar but not bitwise equal abstractions, but as long as the abstract operations are sound it should make no difference.

src/cdomains/floatDomain.ml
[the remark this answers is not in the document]
sim642: I'm confused where this response comment went that was in an email:

> Representing top in the float domain is not simply possible by using a max range interval (as done in the int domains). Therefore we had to introduce some kind of special value for top, that covers the whole range of floating point numbers as well as the specials like +/- inf and Nan. Here we went with None. On the other side we have no representation of bot in our domain.

The NaN-s are a good point that didn't occur to me. To be explicit `Lattice.LiftTop` would be nicer for this sort of thing, so there's so wondering what `None` means.

src/cdomains/floatOps/stubs.c
```diff
@@ -0,0 +1,87 @@
...
+
+CAMLprim value atof_float(value mode, value str)
+{
+    const char *s = String_val(str);
+    volatile float r;
+    change_round_mode(Int_val(mode));
+    r = (float)atof(s);
+    change_round_mode(Nearest);
+    return caml_copy_double(r);
+}
```
sim642, line 77: The macro above was changed to restore the old rounding mode (instead of `Nearest`), but these two are still unchanged.

src/analyses/base.ml
```diff
@@ -2402,6 +2530,54 @@ struct
...
+                  | "__builtin_isnan" -> `Int (ID.cast_to IInt (FD.isnan float_x))
+                  | "__builtin_isnormal" -> `Int (ID.cast_to IInt (FD.isnormal float_x))
+                  | "__builtin_signbit" -> `Int (ID.cast_to IInt (FD.signbit float_x))
+                  | "__builtin_acos" | "acos" -> `Float (FD.acos float_x)
+                  | "__builtin_asin" | "asin" -> `Float (FD.asin float_x)
+                  | "__builtin_atan" | "atan" -> `Float (FD.atan float_x)
+                  | "__builtin_cos" | "cos" -> `Float (FD.cos float_x)
+                  | "__builtin_sin" | "sin" -> `Float (FD.sin float_x)
+                  | "__builtin_tan" | "tan" -> `Float (FD.tan float_x)
+                  | _ -> failwith (f.vname^" should be implemented in goblint but isn't")
```
sim642, line 2553: There's somewhat of a duplication of all these names between `LibraryFunctions` and here. Also the fact that the multiple versions of the same function (without and with `__builtin_`) need to be handled everywhere, which defeats the point of using `LibraryFunctions`.

A nicer approach might be to define an auxiliary type like
```ocaml
type math =
  | Sin of exp
  (* ... *)
  | Atan2 of exp * exp
```
and have the outer classification constructor as 
```ocaml
  | Math of math
```

Not only does that avoid the name duplication, it also avoids all these `failwith` cases that are currently here, cluttering the interesting parts. With proper `LibraryDesc`s, the checking of argument counts would all be done by `LibraryFunctions` with its existing functionality, like it is for `Memset` above.

src/analyses/base.ml
Dudeldu: This was indeed a major unsoundness. This should now be fixed. The idea behind it is an adaption of the proposed method [here](https://hal.inria.fr/inria-00540299/document).

src/analyses/base.ml
```diff
@@ -712,6 +745,11 @@ struct
       | Const (CInt (num,ikind,str)) ->
         (match str with Some x -> M.tracel "casto" "CInt (%s, %a, %s)\n" (Cilint.string_of_cilint num) d_ikind ikind x | None -> ());
         `Int (ID.cast_to ikind (IntDomain.of_const (num,ikind,str)))
+      | Const (CReal (_, (FFloat | FDouble as fkind), Some str)) -> `Float (FD.of_string fkind str) (* prefer parsing from string due to higher precision *)
+      | Const (CReal (num, (FFloat | FDouble as fkind), None)) -> `Float (FD.of_const fkind num)
+      (* we also support long doubles but only with the precision of a double *)
+      | Const (CReal (_, (FLongDouble), Some str)) -> `Float (FD.of_string FDouble str)
+      | Const (CReal (num, (FLongDouble), None)) -> `Float (FD.of_const FDouble num)
```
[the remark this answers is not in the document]
Dudeldu: We treat `long doubles` only with the precision of `double`, which is sound according to [https://www-apr.lip6.fr/~mine/publi/article-mine-esop04.pdf](https://www-apr.lip6.fr/~mine/publi/article-mine-esop04.pdf).
Of course we are not allow to make statements regarding something is __definitely__ `Top`.

src/analyses/base.ml
sim642: Wow, that's some impressive work! You should put that reference link into a comment before this code, such that it's easier to later find what it's based on without having to dig into the comments of this PR.

Also, did any of the existing tests or SV-COMP show this unsoundness? If not, but you had some tests yourself to validate the soundness of the new implementation compared to the old one, it'd be very valuable to have them as tests. Then we could even submit them to SV-COMP.

src/analyses/base.ml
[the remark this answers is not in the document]
sim642: Great! It'd be nice to also have that statement and reference about `double` precision analysis being sound for `long double` somewhere in the code. For example in `FloatDomain` where the `FLong` constructor is defined using 64-bit float intervals.

src/analyses/base.ml
FelixKrayer: Good Idea. It is implemented now in this way.

michael-schwarz:
Sorry for having been a bit absent here the last weeks, we had a paper deadline today at 2pm. I'll start a run on the SV-comp benchmarks so we can see if there are any unsoundnesses or exceptions left.

tests/regression/56-floats/01-base.c
Dudeldu: As we like the expressive `// SUCCESS` statements, we would like to keep them, but removed all `!` in places, where they didn't belong to. Also we tried to make a difference between `UNKNOWN` and `UNKNOWN!` now.

src/analyses/base.ml
Dudeldu: The sv-comp benchmark didn't show this unsoundness in the last run. However, this could also be due to a crash that frequently happened in similar situations.
Nonetheless i have added some more dedicated tests for the respective border cases in regression test `57/11`.

src/analyses/base.ml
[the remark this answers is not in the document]
Dudeldu: now only complex numbers are left for implementation ;)

src/cdomains/floatDomain.ml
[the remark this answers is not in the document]
Dudeldu: This should now be fixed, with also adding an additional, explicit `Top` and `Bot`

src/cdomains/floatDomain.ml
```diff
@@ -0,0 +1,783 @@
...
+
+  let show = function
+    | Top -> "[Top]"
+    | Bot -> "[Bot]"
+    | Interval (low, high) -> Printf.sprintf "[%s,%s]" (Float_t.to_string low) (Float_t.to_string high)
+
+  let pretty () x = text (show x)
+
+  let printXml f (x : t) =
+    BatPrintf.fprintf f "<value>\n<data>\n%s\n</data>\n</value>\n" (show x)
```
sim642, line 98: Could avoid this boilerplate and use `Printable.SimpleShow` here.

src/analyses/libraryFunctions.ml
```diff
@@ -543,6 +569,183 @@ let invalidate_actions = [
     "sema_init", readsAll;
     "down_trylock", readsAll;
     "up", readsAll;
+    "acos", readsAll;
+    "acosf", readsAll;
```
sim642, line 573: Is there a reason `acosf` and all the other type-suffixed variants of these functions aren't defined as special functions above, along with normal, e.g. `acos`?

FelixKrayer: has been added now

tests/regression/57-floats/08-bit_casts.c
```diff
@@ -0,0 +1,31 @@
+// PARAM: --enable ana.float.interval
+
```
jerhard, line 2:
```suggestion
#include <assert.h>

```

tests/regression/57-floats/01-base.c
```diff
@@ -0,0 +1,46 @@
...
+#include <complex.h>
+
+int main()
+{
+    // ensure that complex floats just do not do anything
+    double _Complex c = 0.;
+    assert(c == 0.); // UNKNOWN
+
+    double x, a = 2., b = 3. + 1;
+    float y, c = 2.f, d = 3.f + 1;
```
jerhard, line 15: `c` is defined twice in the same scope; so `gcc` won't compile this.

tests/regression/57-floats/09-svcomp_float_req_bl_1252b.c
```diff
@@ -0,0 +1,90 @@
+// PARAM: --enable ana.float.interval
+
```
jerhard, line 2:
```suggestion
#include <assert.h>

```

tests/regression/57-floats/10-svcomp_floats_cbmc_regression_float11.c
```diff
@@ -0,0 +1,49 @@
+// PARAM: --enable ana.float.interval
+
```
jerhard, line 2:
```suggestion
#include <assert.h>

```

tests/regression/57-floats/12-subtraction_assignmen.c
```diff
@@ -0,0 +1,11 @@
+//PARAM: --enable ana.float.interval
+
```
jerhard, line 2:
```suggestion
#include assert.h

```
It seems like a `t` is missing at the end of the filename.

michael-schwarz:
I started a run of sv-comp with the current master merged into this.

tests/regression/57-floats/04-casts.c
```diff
@@ -0,0 +1,109 @@
...
+    int rnd;
+
+    double value;
+    float value2;
+    long double value3;
+    int i;
+    long l;
+    unsigned u;
+
+    // Casts from double/flout/long double into different variants of ints
```
jerhard, line 25:
```suggestion
    // Casts from double/float/long double into different variants of ints
```

jerhard:
I tried to test your PR and look at the HTML output, see the [documentation on how to do this](https://goblint.readthedocs.io/en/latest/user-guide/inspecting/). However, the abstract values of float variables do not show up in the HTML view.

E.g. in the HTML produced with `./regtest.sh 57 02 --html`, looking at node `79` of the `main` function, the value for the `float` variable `a` empty: 
Bildschirmfoto vom 2022-07-14 11-08-44
This indicates that there might be a problem with the `printXML` (or, rather the `show` it is based on) for floats.

Dudeldu:
The missing content is due to the float domain being disabled at that point. Therefore we have nothing to show. This should be the exact same behaviour as if you disable all integer domains with `--disable ana.int.def_exc`

jerhard:
>The missing content is due to the float domain being disabled at that point. Therefore we have nothing to show. This should be the exact same behaviour as if you disable all integer domains with --disable ana.int.def_exc

Ah, indeed I overlooked that.

michael-schwarz (approved):
Looks good to be merged from my perspective! Any more comments @sim642 @stilscher @jerhard?

sim642 (approved):

src/analyses/base.ml
```diff
@@ -1865,13 +2067,14 @@ struct
         | _ -> false
       in
       let itv = (* int abstraction for tv *)
-        let ik = Cilfacade.get_ikind_exp exp in
+        (* when using floats without explicit cast, we can actually get a non-integer type *)
+        let ik = try Cilfacade.get_ikind_exp exp with Invalid_argument _ -> IInt in
```
jerhard, line 2071: Is it correct to use `IInt` here in case the expression is not an integer?

src/cdomains/floatDomain.ml
```diff
@@ -0,0 +1,821 @@
...
+  let to_int ik = function
+    | Bot -> raise (ArithmeticOnFloatBot (Printf.sprintf "to_int %s" (show Bot)))
+    | Top -> IntDomain.IntDomTuple.top_of ik
+    (* special treatment for booleans as those aren't "just" truncated *)
+    | Interval (l, h) when ik = IBool && (l > Float_t.zero || h < Float_t.zero) -> IntDomain.IntDomTuple.of_bool IBool true
+    | Interval (l, h) when ik = IBool && l = h && l = Float_t.zero -> IntDomain.IntDomTuple.of_bool IBool false
+    | Interval (l, h) when ik = IBool -> IntDomain.IntDomTuple.top_of IBool
+    | Interval (l, h) ->
+      (* as converting from float to integer is (exactly) defined as leaving out the fractional part,
+         (value is truncated towrad zero) we do not require specific rounding here *)
```
jerhard, line 119:
```suggestion
         (value is truncated toward zero) we do not require specific rounding here *)
```

src/cdomains/floatDomain.ml
```diff
@@ -0,0 +1,821 @@
...
+    match (op1, op2) with
+    | Bot, _ | _, Bot -> raise (ArithmeticOnFloatBot (Printf.sprintf "%s op %s" (show op1) (show op2)))
+    | Interval v1, Interval v2 ->
+      let is_exact_before = is_exact op1 && is_exact op2 in
+      let result = norm @@ eval_operation v1 v2 in
+      (match result with
+       | Interval (r1, r2) when not (is_exact result) && is_exact_before ->
+         Messages.warn
+           ~category:Messages.Category.Float
+           ~tags:[CWE 197; CWE 681; CWE 1339]
```
jerhard, line 269: Are [CWE 197](https://cwe.mitre.org/data/definitions/197.html) and [CWE 681](https://cwe.mitre.org/data/definitions/681.html) really applicable for binary operations on floats?

src/cdomains/floatDomain.ml
```diff
@@ -0,0 +1,821 @@
...
+           "The result of this operation is not exact, even though the inputs were exact";
+       | Top ->
+         Messages.warn
+           ~category:Messages.Category.Float
+           ~tags:[CWE 197; CWE 681; CWE 1339]
```
jerhard, line 274: Same question as above.

src/analyses/base.ml
michael-schwarz: Yes it appears harmless here, the only cases where we go here is for pointer sand floats, and they are then correctly handled late in the invariant by` FD.of_int` and `AD.of_int`.

michael-schwarz: Actually after discussing this further with @jerhard, it does in fact turn out to be problematic: In the true case, we erroneously claim that the float must be in the range of `IInt` (and neither `NaN` nor `+/- Inf`).

src/analyses/base.ml
```diff
@@ -2065,17 +2065,23 @@ struct
...
           else
             ID.of_excl_list ik [BI.zero] (* Lvals, Casts, arithmetic operations etc. should work with true = non_zero *)
-          in
-          inv_exp (`Int itv) exp st
-        with Invalid_argument _ ->
-          inv_exp (`Float (FD.top_of (Cilfacade.get_fkind_exp exp))) exp st
+        in
+        inv_exp (`Int itv) exp st
+      with Invalid_argument _ ->
+        let fk = Cilfacade.get_fkind_exp exp in
```
jerhard, line 2078: In C, the expressions in a condition may also have pointer type, i.e. neither be an integer nor a float; so this may throw an exception.

PR #762, opened 2022-06-16
Add Apron box product lifter

sim642:
This adds the `BoxProd` lifter for Apron domains, which allows `invariant` to filter out all non-relational invariants (including trivially derived ones) by filtering out box constraints.
Some refactoring of the interfaces was necessary to abstract away `Man` since the `BoxProd` components use different domain managers.

Also adds options for deciding, which variables should be used in generated Apron invariants.


sim642:
The sad thing is that with all this filtering to find truly relational global invariants, we find none in the relational traces benchmarks.

michael-schwarz (approved):

PR #764, opened 2022-06-21
Add `setjmp` Analysis

bottbenj:
Adds an analysis to handle the [`setjmp`](https://en.cppreference.com/w/c/program/setjmp) and [`longjmp`](https://en.cppreference.com/w/c/program/longjmp) functions.

The setjmp analysis collects local state of the other analyses with the `Longjmp` and `VolatileLocals` query and then returns it when required with the `Setjmp` event. Due to the path insensitive nature of analysis globals it is not necessary to touch them.

The setjump analysis is able to collect arbitrary data from all analyses by using a heterogeneous map, but this also means it is not able to affect any specific state and can only perform domain operations on the whole map. This map domain is based on the [hmap](https://github.com/dbuenzli/hmap) library, but included directly as it is pretty abandoned and it needed to add various functions to be able to implement a domain based on it.

This is currently only used by base, and i am not exactly sure which other analyses also benefit from support for it.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -18,6 +18,7 @@ module LF = LibraryFunctions
 module CArrays = ValueDomain.CArrays
 module BI = IntOps.BigIntOps
 module PU = PrecisionUtil
+module HM = MapDomainHeterogeneous.MapSetJmp
 
 module VD     = BaseDomain.VD
 module CPA    = BaseDomain.CPA
@@ -74,6 +75,10 @@ struct
   let name () = "base"
   let startstate v: store = { cpa = CPA.bot (); deps = Dep.bot (); weak = WeakUpdates.bot (); priv = Priv.startstate ()}
   let exitstate  v: store = { cpa = CPA.bot (); deps = Dep.bot (); weak = WeakUpdates.bot (); priv = Priv.startstate ()}
+
+  let setjmp_return_token = HM.token (module VD) (VD.zero_init_value (TInt (IInt, []))) "base: return"
+  let setjmp_globals_token = HM.token (module D) (D.bot ()) "base: globals"
+  let setjmp_volatiles_token = HM.token (module D) (D.bot ()) "base: volatiles"
 
   (**************************************************************************
    * Helpers
@@ -1180,6 +1185,11 @@ struct
         | _ -> true
       end
     | Q.IsMultiple v -> WeakUpdates.mem v ctx.local.weak
+    | Q.Longjmp exp ->
+      HM.singleton setjmp_return_token (eval_rv (ask_of_ctx ctx) ctx.global ctx.local exp)
+      |> HM.add setjmp_globals_token { ctx.local with cpa = CPA.filter (fun k v -> Basetype.Variables.is_global k) ctx.local.cpa }
+    | Q.VolatileLocals ->
+      HM.singleton setjmp_volatiles_token { ctx.local with cpa = CPA.filter (fun k v -> (not (Basetype.Variables.is_global k)) && Ciltools.is_volatile_tp k.vtype) ctx.local.cpa }
     | _ -> Q.Result.top q
 
   let update_variable variable typ value cpa =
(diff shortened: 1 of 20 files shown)


michael-schwarz:
Thank you for the PR, it is nice to see Goblint getting this feature, I think it sets us apart from a lot of the competition.

src/analyses/base.ml
```diff
@@ -1180,6 +1185,11 @@ struct
         | _ -> true
       end
     | Q.IsMultiple v -> WeakUpdates.mem v ctx.local.weak
+    | Q.Longjmp exp ->
+      HM.singleton setjmp_return_token (eval_rv (ask_of_ctx ctx) ctx.global ctx.local exp)
+      |> HM.add setjmp_globals_token { ctx.local with cpa = CPA.filter (fun k v -> Basetype.Variables.is_global k) ctx.local.cpa }
+    | Q.VolatileLocals ->
+      HM.singleton setjmp_volatiles_token { ctx.local with cpa = CPA.filter (fun k v -> (not (Basetype.Variables.is_global k)) && Ciltools.is_volatile_tp k.vtype) ctx.local.cpa }
```
sim642, line 1192: Here `Basetype.Variables.is_global` is used, which is a purely syntactic check. However, the local state of base analysis can also contain other things which are not syntactically global, but are effectively handled as globals, e.g. alloc variables and escaped variables. Shouldn't some of them also be passed around via these jumps?

sim642 (commented):
Thanks for the support of this weird C feature. Allows us to tick another soundiness box: http://soundiness.org/.

sim642:
> This is currently only used by base, and i am not exactly sure which other analyses also benefit from support for it.

Is it just a question of benefit or is it rather a question of soundness? For example, doesn't the apron analysis also need to handle this and do this extra join to be sound?

Probably a harder question, but what happens with locksets across long jumps (if the combination of threads and long jumps is defined at all)? Because if locksets persist over long jumps, then the mutex analysis would also need such extra join for soundness.

> The setjump analysis is able to collect arbitrary data from all analyses by using a heterogeneous map

The use of hmap is certainly an interesting approach to allow your analysis to manipulate states of other analyses. Likely not worth redoing at this point, but another technique that's used in Goblint is to use a `Spec` lifting functor instead of an analysis. Such a functor would similarly have access to the underlying analysis domains and operate on them as whole. Such a lifter would then apply to the analysis as a whole as well, instead of per-analysis, although some extra operation might still be needed on the analyses to choose the correct data to join across the long jump.

src/analyses/setjmpAnalysis.ml
```diff
@@ -0,0 +1,91 @@
...
+module LF = LibraryFunctions
+module HM = MapDomainHeterogeneous.MapSetJmp
+
+module Spec =
+struct
+  include IdentitySpec
+
+  let name () = "setjmp"
+
+  module D = Queries.LS (* if the local method has performed a setjmp *)
```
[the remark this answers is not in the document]
jerhard: What is the reason for keeping the set of jumpbuffers in the local state?

src/analyses/setjmpAnalysis.ml
```diff
@@ -0,0 +1,91 @@
...
+      if M.tracing then M.traceli "setjmp" "Longjmp vars: [%a]\n" D.pretty vars;
+      sideg ctx vars data;
+      collect_volatiles ctx;
+      if M.tracing then M.traceOutdent();
+      (* longjmp never returns directly it will return from the setjmp instead *)
+      raise Deadcode
+    | _ -> ctx.local
+
+  let enter ctx lval fn args =
+    collect_volatiles ctx;
```
jerhard, line 86: Why do the volatiles have to be collected on every `enter`?

jerhard:
>Is it just a question of benefit or is it rather a question of soundness?

This is indeed a soundness question. With the implementation as an analysis, one would have to add some default join action to all the analyses. Maybe this could be reduced to a one-line for all the analyses by including a module/functor with some default implementation? Otherwise, one would like to come up with a solution where changing the existing analyses would not be necessary. One possible solution would be to change the implementation to be a functor taking a `Spec`, like @sim642  suggest.

sim642:
> One possible solution would be to change the implementation to be a functor taking a `Spec`, like @sim642 suggests.
> 
> You can find an example implementation for such a functor e.g. with `PathSensitive2`.

That requires rethinking how the data is passed though because events are not available for those functors. So it might require adding one or two additional transfer functions to `Spec`.
On the other hand, it allows using a `Either` for `V` and `G` to split globals into inner spec globals and long jump data globals that don't require hmap.

michael-schwarz:
Thank you for your PR, it is has been very helpful to help us explore the design space and highlighted the intricacies of such an analysis. We have decided against merging this for now and will instead try to pursue a more generic approach.

michael-schwarz:
Thank you for your PR, it is has been very helpful to help us explore the design space and highlighted the intricacies of such an analysis. We have decided against merging this for now and will instead try to pursue a more generic approach.

PR #770, opened 2022-06-30
Replace MustBeEqual, MayBeEqual and MayBeLess queries with EvalInt

sim642:
### Changes
1. Removes `MustBeEqual`, `MayBeEqual` and `MayBeLess` queries and their responses, which are all subsumed by `EvalInt`.
2. Add backwards-compatiibility functions for doing those queries through `EvalInt`.
3. Add query cycle detection to partitioned array queries that bypass MCP.
4. Fix interval domain `eq` and `ne` emitting spurious overflow warnings since they used to be implemented via `sub`. Explicit implementations like the other comparison operators avoid the issue.
5. Identify uses of the removed/replaced queries that aren't covered by tests, i.e. everything still passes if those queries return constant top. There's a lot of those, but I ended up adding tests for some of them.
6. Add `ana.base.eval.deep-query` option which allows disabling `EvalInt` queries on subexpressions. This avoids querying Apron at every level, but just querying at the top level expression. Gives a speedup in torture test, although could be less precise in some obscure cases.

### TODO
- [x] Expose `Cabs2cil.doBinOp` in CIL instead of copying code into `Cilfacade`.
- [x] Remove `eval_rv_ask_mustbeequal` layer in base analysis? _Didn't remove because required to handle some added tests that previously would've passed._
- [x] Benchmark (before vs after) on sv-benchmarks for soundness, precision and performance.
- [x] What to do about untested queries? _Added tests for some._


tests/regression/02-base/82-eq-no-overflow.c
```diff
@@ -0,0 +1,17 @@
+// PARAM: --enable ana.int.interval
+#include <assert.h>
+
+int main() {
+  unsigned long x;
+  x = 0;
+
+  int b;
+  b = x == 7; // NOWARN
```
jerhard, line 9: Is there a specific reason for this `NOWARN` annotation? Did we previously produce an overflow warning here?

src/domains/queries.ml
```diff
@@ -329,3 +311,29 @@ struct
 
   let hash x = 31 * order x + hash_arg x
 end
+
+
+let eval_int_binop (module Bool: Lattice.S with type t = bool) binop (ask: ask) e1 e2: Bool.t =
+  let t1 = Cilfacade.typeOf e1 in
+  let t2 = Cilfacade.typeOf e2 in
+  let (_, e) = Cilfacade.doBinOp binop e1 t1 e2 t2 in
```
jerhard, line 319: What does `doBinOp` do in case the expressions have different type?

src/cdomains/arrayDomain.ml
```diff
@@ -353,15 +353,19 @@ struct
...
+                    let b = Q.may_be_less ask exp (Cil.kintegerCilint (Cilfacade.ptrdiff_ikind ()) i) in
                     not b (* !(e <_{may} length) => e >=_{must} length *)
                   | None -> false
                 end
               | _ -> false
             in
             let e_must_less_zero =
-              let b = ask.f (Q.MayBeLess (Cil.mone, exp)) in
-              not b (* !(-1 <_{may} e) => e <=_{must} -1 *)
+              if Cil.isSigned (Cilfacade.get_ikind_exp exp) then ( (* avoid may_be_less comparing unsigned index to -1, which is promoted to max int of ikind via unsigned underflow *)
```
jerhard, line 363: >avoid may_be_less comparing unsigned index to -1, which is promoted to max int of ikind via unsigned underflow

Is this something that one has to take additional care of now, when rephrasing the `MayBeLess` (and related) queries to `EvalInt`? Although it is not so clear what happens in such cases in the current implementation, as the int-domains also consider the `ikind`s. So it probably should be implemented like that here, anyways.

tests/regression/46-apron2/18-evalint-torture.c
```diff
@@ -0,0 +1,30 @@
+// SKIP PARAM: --set ana.activated[+] apron
+// Torture EvalInt queries using large number of (relational) expressions.
+// Ran in ~7s with MustBeEqual, MayBeEqual and MayBeLess queries.
```
jerhard, line 3: Did the peformance change?

src/cdomains/intDomain.ml
```diff
@@ -792,9 +788,28 @@ struct
...
+    | None, None -> bot_of ik
+    | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
+    | Some (x1,x2), Some (y1,y2) ->
+      if Ints_t.compare y2 x1 < 0 || Ints_t.compare x2 y1 < 0 then
+        of_bool ik true
+      else if Ints_t.compare x2 y1 <= 0 && Ints_t.compare y2 x1 <= 0 then
+        of_bool ik false
+      else top_bool
+
+  let eq ik x y =
```
jerhard, line 803: `ne` could be implemented as a negation of the abstract result of `eq`, or vice-versa. Or one could implement a helper method for `ne` and `eq` that takes the boolean that should be the result for the equal and non-equal case as arguments.

Since the implementations are not a lot of code, I don't have a strong opinion on that, though.

jerhard:
>This does provide some speedup for reproving assertions, particularly on ctrace

Interesting! Do you have any intuition why that speed-up occurs?

src/domains/queries.ml
```diff
@@ -329,3 +311,29 @@ struct
...
+  let i = ask.f (EvalInt e) in
+  if ID.is_bot i || ID.is_bot_ikind i then
+    Bool.top () (* base returns bot for non-int results, consider unknown *)
```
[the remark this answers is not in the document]
sim642: Yes. Base itself uses the but result of an `EvalInt` query as indication that the expression actually doesn't evaluate to an integer type. It can then fall back to all the other cases. If it returned top instead, then it would never be able to distinguish, e.g. thread ID sets from actually unknown integers.

One reason for needing such inconsistency is, e.g. for these thread ID sets, because variables of `pthread_t` type are actually just integer values (on Linux), but we want base to evaluate them to `Thread` instead.

src/domains/queries.ml
sim642: The whole point of `doBinOp` is that it does the exact same thing as the C parser: it finds the common type and inserts the required casts. The logic for that is rather involved (as currently can be seen from my copy-paste to `Cilfacade`). This was also required to fix some incompatible ikinds errors that otherwise arise from just putting two incompatible types of integers into a `BinOp`.

src/cdomains/intDomain.ml
sim642: Possibly, I went with the least tricky way here, following what all the other four comparison operators are defined like. There could be some nicer way to define them all with less duplication (but avoid doing that through `sub`).

sim642:
> Interesting! Do you have any intuition why that speed-up occurs?

My original reason for wanting to do this was to get some speedup (the cleanup aspect is nice too). While doing witness validation with ~1000 invariants, it was exceedingly slow (it still isn't very fast, but somewhat better).
One reason I figured was how base calls all the layers of `eval_rv`:
* The `eval_rv_ask_evalint` layer does a an `EvalInt` query with the whole (int) expression, so for `x + y == 0` Apron can answer that.
* If that doesn't give a definite result, evaluation continues to the `eval_rv_ask_mustbeequal` layer, where that expression is split up to `x + y` and `0` and a `MustBeEqual` query was done. But Apron also answers that query by putting the two sides back together and again evaluating `x + y == 0` on the Apron state.

So there could be a significant waste of effort doing double the number of expensive Apron evaluations. Getting rid of `MustBeEqual` (and maybe the corresponding base eval layer, I'm not yet completely sure about it) makes such double evaluation impossible.

Moreover, `MustBeEqual` caused another doubling of work, since base itself also answers that by evaluating both arguments. If they didn't come out as equal, base continues down to the `eval_rv_base` layer where it finally does the structural evaluation, repeating the evaluation of both arguments.

Although now writing this, I realize that this might optimize unknown expressions much more than successful ones. That might be affecting how much speedup this provides.

sim642:
I did before vs after sv-benchmarks runs for this PR. There's no change in resultsl but on average there is a ~4% speedup:
image
The full comparison table is here: https://goblint.cs.ut.ee/results/72-all-fast-evalint-opt-after/results/results.2022-07-04_07-31-46.table.html#/scatter?toolY=1&columnY=1&scaling=Linear&results=All&filter=0(0*status*(status(in(unknown,true)))),1(0*status*(status(in(unknown,true))))&regression=Linear&line=1.2.

To note, this is with our SV-COMP configuration that currently doesn't even use Apron, but just from our default analyses themselves.

tests/regression/02-base/85-evalint-maybeless.c
```diff
@@ -0,0 +1,14 @@
...
+#include <stddef.h>
+
+int main() {
+  int x;
+
+  // expRelation EvalInt
+  assert(!(x + 1 < x));
+  assert(!(x < x + (-1)));
+  assert(!(x - (-1) < x));
+  assert(!(x < x - 1));
```
jerhard, line 12: So `expRelation` does not consider the possibility of overflows?

jerhard:
Impressive optimization, seeing this factor 10 speedup! 

Seeing at this [log](https://goblint.cs.ut.ee/simmo-results/goblint-bench_result-traces-rel-yaml9-shallow-sound/ctrace_comb.cluster12.txt.html) for example, the solving still takes about 5s out of 65s of runtime. Do you have an intuition, where this difference comes from? Is the assertion checking not contained in the `postsolver` runtime?

tests/regression/02-base/85-evalint-maybeless.c
sim642: That's a good point! I think it never has, which I guess could be a problem.
Given that these cases were completely untested before and I reverse engineered the tests for them from the code, maybe we don't really need most of it. I think we'll have to separately decide what to do about it.

sim642:
> Looking at this [log](https://goblint.cs.ut.ee/simmo-results/goblint-bench_result-traces-rel-yaml9-shallow-sound/ctrace_comb.cluster12.txt.html) for example, the solving still takes about 5s out of 65s of runtime. Do you have an intuition, where this difference comes from? Is the assertion checking not contained in the `postsolver` runtime?

That's right, those benchmarks don't use assertions per-se, but YAML witness invariants. Those are currently checked completely after the analysis and postsolver.
It probably doesn't make sense to report on those total times which perform 1448 invariant evaluations in a 1407 line program, especially since those aren't in-program assertions (which have scoping problems in these combined programs).

So the massive speedup just applies to that invariant evaluation, where we are deliberately evaluating just Apron invariant that invoke the worst of the duplicate computation problem of these queries. The impact is much smaller on normal analysis (as also seen on SV-COMP), although it would still be there if these were in-program assertions.

tests/regression/02-base/85-evalint-maybeless.c
sim642: Delegated to #776, as to not block this PR.

jerhard (approved):

PR #772, opened 2022-06-30
Autotuning of Configuration

leunam99:
Adds an option that tries choosing some configurations automatically based on the file that is analyzed.
Already implemented: 
- Enable Congruence domain only in functions that use modulo and related ones
- new widening thresholds
- focus analysis depending on sv-comp specification
- disable thread analyses if program is single threaded
- disable interval contexts in recursive functions 
This is part of my bachelor thesis and still work in progress.


conf/svcomp22.json
```diff
@@ -1,12 +1,12 @@
 {
   "ana": {
+    "autoselect": true,
```
sim642, line 3: This conf file reflects our configuration at SV-COMP 2022 and thus should not be changed.

goblint.opam
```diff
@@ -20,7 +20,7 @@ doc: "https://goblint.readthedocs.io/en/latest/"
 bug-reports: "https://github.com/goblint/analyzer/issues"
 depends: [
   "ocaml" {>= "4.10"}
-  "dune" {>= "2.9.1"}
+  "dune" {>= "2.9" & >= "2.9.1"}
```
sim642, line 23:
```suggestion
  "dune" {>= "2.9.1"}
```

sv-comp/my-bench-sv-comp/goblint-all-fast.sh
```diff
@@ -2,9 +2,9 @@
 
 shopt -s extglob
 
-MYBENCHDIR=/mnt/goblint-svcomp/benchexec/my-bench-sv-comp
-RESULTSDIR=/mnt/goblint-svcomp/benchexec/results/70-all-fast-no-interval
-GOBLINTPARALLEL=14
+MYBENCHDIR=//mnt/goblint-svcomp/sv-comp/goblint/sv-comp/my-bench-sv-comp
+RESULTSDIR=/mnt/goblint-svcomp/benchexec/results/LoopsEnums
+GOBLINTPARALLEL=3
```
sim642, line 7: The changes to all-fast and data-race scripts and definitions should be reverted since they just reflect just some of your own experiments and nothing fundamental about these example scripts.
The no-overflow ones can remain since we didn't have any before.

src/util/wideningThresholds.mli
```diff
@@ -4,3 +4,6 @@ val thresholds_incl_mul2 : unit -> Z.t list
 val exps: Cil.exp list ResettableLazy.t
 
 val reset_lazy : unit -> unit
+val upper_thresholds : unit -> Z.t list
+val lower_thresholds : unit -> Z.t list
+val ocatagon_thresholds : unit -> Z.t list
```
sim642, line 9:
```suggestion
val octagon_thresholds : unit -> Z.t list
```

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -16,7 +16,7 @@ module M = Messages
     - heterogeneous environments: https://link.springer.com/chapter/10.1007%2F978-3-030-17184-1_26 (Section 4.1) *)
 
 let widening_thresholds_apron = ResettableLazy.from_fun (fun () ->
-  let t = WideningThresholds.thresholds_incl_mul2 () in
+  let t = if GobConfig.get_bool "ana.autoselect" then WideningThresholds.ocatagon_thresholds () else WideningThresholds.thresholds_incl_mul2 () in
```
sim642, line 19: The choice of which widening thresholds are used is currently tied together with `ana.autoselect`, but it might be better to have separate options for that, since we might also want to use the more optimized threshold collecting tactics outside of SV-COMP.

src/AutoSelect.ml
```diff
@@ -0,0 +1,209 @@
...
+    tar = All_t;
+    f = Uses_f; (*Does this select only function calls without Parameters? *)
+    str = None_s; (* Only search in the specific function*)
+  }
+*)
+
+module Varinfo = struct 
+  type t = varinfo
+  let compare v1 v2 = Stdlib.compare v1.vid v2.vid
+end;;
```
sim642, line 22: This is just `CilType.Varinfo`.

src/AutoSelect.ml
```diff
@@ -0,0 +1,209 @@
+open GobConfig;;
```
sim642, line 1: The `;;` are excessive and we don't use them in our style.

src/AutoSelect.ml
```diff
@@ -0,0 +1,209 @@
...
+open Cil;;
+
+(*Create maps that map each function to the ones called in it and the ones calling it*)
+(* TODO rewrite with Syntacticsearch?: https://github.com/goblint/cil/blob/develop/src/ext/syntacticsearch/funcFunction.ml
```
sim642, line 5: Would be important to mention that this only detects _statically-known_ calls and none through function pointers (through which also recursion could occur).

Also, I wouldn't rewrite it with syntactic search because its implementation is much uglier than the relatively simple visitor here.

src/AutoSelect.ml
```diff
@@ -0,0 +1,209 @@
...
+      ) 
+      (FunctionSet.filter 
+        (fun x -> isNotExtern x.vstorage) 
+        (neigbourFunction fd.svar) 
+      )
+    ;
+  );;
+
+(*Because we now ignore the total count and only care if it exists, this should be rewritten to stop after the first mod*)
+(* -> No longer a visitor, but a recursive search?*)
```
sim642, line 95: A visitor would still be much shorter since you don't have to rewrite all the structural recursion over the AST. You can use a custom exception to jump out of the visitor and catch the exception outside, which is a fairly common OCaml programming technique for this sort of thing.

src/AutoSelect.ml
```diff
@@ -0,0 +1,209 @@
...
+      print_endline ("function " ^ (f.vname) ^" is recursive, disable interval analysis");
+      f.vattr <- addAttributes (f.vattr) [Attr ("goblint_context",[AStr "base.no-interval"])];
+    )
+  );;
+
+(*If only one Thread is used in the program, we can disable most thread analyses*)
+(*The exceptions are analyses that are depended on by others: base -> mutex -> mutexEvents, access*)
+(*TODO escape is also still enabled, because I do not know if the analysis gets more imprecise if not*)
+
+let toJsonArray list = "[\"" ^ (String.concat "\",\"" list) ^ "\"]";;
```
sim642, line 142: This is very hacky and should be avoided.

src/AutoSelect.ml
```diff
@@ -0,0 +1,209 @@
...
+      ) 
+  in
+
+  (*TODO is there a way to specify only the thread analyses to keep?  *)
+  if not @@ hasThreadCreate () then (
+      print_endline @@ "no thread creation -> disabeling thread analyses \"" ^ (String.concat ", " notNeccessaryThreadAnalyses) ^ "\"";
+      get_string_list "ana.activated" 
+      |> List.filter (fun l -> not (List.mem l notNeccessaryThreadAnalyses))
+      |> toJsonArray
+      |> set_auto "ana.activated"
```
sim642, line 171: Instead of trying to manipulate the entire `ana.activated` list, you can just disable analyses one-by-one:
```ocaml
GobConfig.set_auto "ana.activated[-]" "deadlock"
```

src/AutoSelect.ml
```diff
@@ -0,0 +1,209 @@
...
+      print_endline @@ "Specification: NoDataRace -> enabeling thread analyses \"" ^ (String.concat ", " notNeccessaryThreadAnalyses) ^ "\"";
+      get_string_list "ana.activated" 
+      |> List.filter (fun l -> not (List.mem l notNeccessaryThreadAnalyses)) (*remove duplicates*)
+      |> (@) notNeccessaryThreadAnalyses
+      |> toJsonArray
+      |> set_auto "ana.activated";
+    | NoOverflow -> (*We focus on integer analysis*)
+      set_bool "ana.int.def_exc" true;
+      set_bool "ana.int.enums" true;
+      set_bool "ana.int.interval" true;;
```
sim642, line 189: Instead of hard-coding the options to change for each specification, ultimately the nicer approach would be something like what CPAchecker does: have a conf file for each specification and just load them.

sim642 (commented):
Thanks for the contribution!
Given that you've touched the SV-COMP files here already, I suppose you've already tried this on sv-benchmarks? If so, then does this already show some improvement there?

src/AutoSelect.ml
leunam99: That is a good suggestion. Because I do not know if this is in any way final, I would not implement it yet.

leunam99:
I am testing with the sv-comp tests and there are some improvements.
For example, in the NoOverflow category Goblint is correct 104 instead of 81 times and uses a bit less time and memory (on my machine at least)
In the other categories there are also improvements, but I currently don't know the numbers

src/AutoSelect.ml
michael-schwarz: I would suggest to leave it like this for this PR.

michael-schwarz (approved):

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -125,6 +125,7 @@ struct
       )
       else (
         let v_out = Goblintutil.create_var @@ makeVarinfo false (v.vname ^ "#out") v.vtype in (* temporary local g#out for global g *)
+        v_out.vattr <- v.vattr; (*copy the attributes*)
```
sim642, line 128: Clearly this is copying the attributes, but the comment should explain why this is now needed.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -16,7 +16,7 @@ module M = Messages
     - heterogeneous environments: https://link.springer.com/chapter/10.1007%2F978-3-030-17184-1_26 (Section 4.1) *)
 
 let widening_thresholds_apron = ResettableLazy.from_fun (fun () ->
-  let t = WideningThresholds.thresholds_incl_mul2 () in
+    let t = if GobConfig.get_string "ana.apron.threshold_widening_constants" = "comparisons" then WideningThresholds.octagon_thresholds () else WideningThresholds.thresholds_incl_mul2 () in
   let r = List.map (fun x -> Apron.Scalar.of_mpqf @@ Mpqf.of_mpz @@ Z_mlgmpidl.mpz_of_z x) t in
   Array.of_list r
```
sim642, line 21: Indentation is now screwed up here.

tests/regression/58-array_annotations/01-var_array_annotation.c
```diff
@@ -0,0 +1,20 @@
+// PARAM: --set ana.base.arrays.unrolling-factor 4 --enable annotation.array
+
+
+int main(void);
+
+int main(void)
+{
+    int x[4] __attribute__((goblint_array_domain("unroll")));
```
sim642, line 8: The new annotation should be documented in the user guide as well.

tests/regression/58-array_annotations/03-change_domain_array_annotation.c
```diff
@@ -0,0 +1,69 @@
...
+
+}
+
+int test2(void)
+{
+
+}
+
+
+
```
sim642, line 49:
```suggestion

```

src/util/options.schema.json
```diff
@@ -1306,6 +1344,18 @@
...
           }
+        },
+        "array": {
+          "title": "annotation.array",
+          "description": "Enable manual annotation of arrays with desired domain",
+          "type": "boolean",
+          "default": false
+        },
+        "track_apron": {
+          "title": "annotation.track_apron",
```
sim642, line 1355: Above attribute annotations have their option names exactly match the attribute names. Would be good for consistency to continue that.

src/goblint.ml
```diff
@@ -36,6 +36,7 @@ let main () =
         else
           Analyses.empty_increment_data ()
       in
+      if get_bool "ana.autotune.enabled" && not (get_bool "incremental.load") then AutoTune.chooseConfig file;
```
sim642, line 39: Instead of silently turning off autotune in incremental mode (which in an ideal world should also work), we could just have that combination of options produce an error whereever those other checks are.

src/autoTune0.ml
```diff
@@ -0,0 +1,105 @@
...
+  ignore (visitAction visitor visitedObject);
+  factors
+
+
+
+(*array heuristics*)
+let is_important_type (t: typ): bool = match t with
+  | TNamed (info, attr) -> List.mem info.tname ["pthread_mutex_t"; "spinlock_t"; "pthread_t"]
+  | TInt (IInt, attr) -> hasAttribute "mutex" attr
+  | _ -> false
```
sim642, line 101: Important in what sense?

src/cdomains/valueDomain.ml
```diff
@@ -108,21 +107,23 @@ struct
...
+      let typAttr = typeAttrs ai in
+      `Array (CArrays.make ~varAttr ~typAttr (IndexDomain.bot ()) (bot_value ai))
     | TArray (ai, Some exp, _) ->
+      let typAttr = typeAttrs ai in
       let l = BatOption.map Cilint.big_int_of_cilint (Cil.getInteger (Cil.constFold true exp)) in
-      `Array (CArrays.make (BatOption.map_default (IndexDomain.of_int (Cilfacade.ptrdiff_ikind ())) (IndexDomain.bot ()) l) (bot_value ai))
+      `Array (CArrays.make ~varAttr ~typAttr (BatOption.map_default (IndexDomain.of_int (Cilfacade.ptrdiff_ikind ())) (IndexDomain.bot ()) l) (bot_value ai))
     | t when is_thread_type t -> `Thread (ConcDomain.ThreadSet.empty ())
-    | TNamed ({ttype=t; _}, _) -> bot_value t
+    | TNamed ({ttype=t; _}, _) -> bot_value ~varAttr (unrollType t)
```
sim642, line 126: Why is there a `unrollType` here now? This way the outermost `TNamed` is handled here, but then the inner ones are removed by `unrollType`, which seems unnecessary since `bot_value` does that recursion anyway without having to allocate an intermediate unrolled type for it.

src/analyses/base.ml
```diff
@@ -87,14 +87,50 @@ struct
...
-    | true, true, _ -> VD.project PU.max_int_precision value
-    | true, false, Some p -> VD.project p value
-    | _ -> value
+  (*This is a bit of a hack to be able to change array domains if a pointer to an array is given as an argument*)
+  (*We have to prevent different domains to be used at the same time for the same array*)
+  (*After a function call, the domain has to be the same as before and we can not depend on the pointers staying the same*)
+  (*-> we determine the arrays a pointer can point to once at the beginning of a function*)
+  (*There surely is a better way, because this means that often the wrong one gets chosen*)
+  module VarMap = Map.Make(CilType.Varinfo)
+  let array_map = ref VarMap.empty
```
sim642, line 96: This is just a `Hashtbl`.

src/analyses/base.ml
```diff
@@ -87,14 +87,50 @@ struct
...
+            let possibleVars = PreValueDomain.AD.to_var_may t in
+            List.fold_left (fun map arr -> VarMap.add arr (info.vattr) map) (pointedArrayMap xs) @@ List.filter (fun info -> isArrayType info.vtype) possibleVars
+          )
+          | _ -> pointedArrayMap xs
+      )
+  in
+  match VarMap.find_opt fundec.svar !array_map with
+      (*We already have something -> do not change it*)
+      | Some _ -> ()
+      | None -> array_map := VarMap.add fundec.svar (pointedArrayMap arguments) !array_map
```
sim642, line 113: This indentation looks wrong.

src/analyses/base.ml
```diff
@@ -87,14 +87,50 @@ struct
...
+
+  let attibutes_varinfo info fundec =
```
sim642, line 115:
```suggestion
  let attributes_varinfo info fundec =
```

src/analyses/base.ml
```diff
@@ -87,14 +87,50 @@ struct
...
-  let project_val p_opt value is_glob =
-    match GobConfig.get_bool "annotation.int.enabled", is_glob, p_opt with
-    | true, true, _ -> VD.project PU.max_int_precision value
-    | true, false, Some p -> VD.project p value
-    | _ -> value
+  (*This is a bit of a hack to be able to change array domains if a pointer to an array is given as an argument*)
+  (*We have to prevent different domains to be used at the same time for the same array*)
+  (*After a function call, the domain has to be the same as before and we can not depend on the pointers staying the same*)
+  (*-> we determine the arrays a pointer can point to once at the beginning of a function*)
+  (*There surely is a better way, because this means that often the wrong one gets chosen*)
```
sim642, line 94: This looks very suspicious, especially because it ignores contexts and very likely breaks incrementally (at minimum it should be marshaled or cleared for every analysis? not sure what the scope of this information is supposed to be).

How is this different from the existing int domain projections? Pointers to int variables exist as well.

sim642 (changes requested):
It would be good to have complete SV-COMP before vs after runs with this to see how well this ad-hoc stuff works.

src/analyses/base.ml
leunam99: The array annotations are scoped per array/type and not the function like the int domain. 
I think this is more useful for them because for them their individual usage is more important and opposed to integers there are no problems if two different arrays have different domains.

If we then do want to change them for some function only, my best solution was to do this for the pointer parameters, as array arguments do not exist. As a result of time restraints, this is not really used yet. 

I also dislike that this makes the function somewhat stateful.
The scope of this should be the analysis of a single program. Would it be enough for that to reset it at the same place as all the lazy  maps?
I did not think about the incremental analysis and lack the knowledge to see how this could be accommodated.

src/cdomains/valueDomain.ml
leunam99: unrollType also collects the attributes of the named types, which is why I changed it here

src/autoTune0.ml
leunam99: Here I meant important for thread analysis, The thought was to unroll arrays containing threads and locks so that the analysis has more information about them.

src/autoTune.ml
```diff
@@ -0,0 +1,434 @@
...
+(*The exceptions are analyses that are depended on by others: base -> mutex -> mutexEvents, access*)
+(*escape is also still enabled, because otherwise we get a warning*)
+(*does not consider dynamic calls!*)
+
+let notNeccessaryThreadAnalyses = ["deadlock"; "maylocks"; "symb_locks"; "thread"; "threadflag"; "threadid"; "threadJoins"; "threadreturn"]
+let reduceThreadAnalyses () =
+  let hasThreadCreate () =
+    ResettableLazy.force functionCallMaps
+    |> fun (_,x,_) -> x  (*every function that is called*)
+                      |> FunctionCallMap.exists
```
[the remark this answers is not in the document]
leunam99: It s the one used by icp-indent, because the second chain of |> is inside the function. I thought that at some point I  had put line 142 in brackets to change it into the mostly equivalent version where it is a single chain of |>. Should I just override icp-indent or do it like that again?

src/util/cilfacade.ml
```diff
@@ -133,42 +133,400 @@ class copyandPatchLabelsVisitor(loopEnd,currentIterationEnd) = object
     | _ -> ChangeDoChildrenPost(rename_labels s, after)
 end
 
-class loopUnrollingVisitor = object
+(*loop unroll heuristics*)
+(*used if AutoTune is activated*)
```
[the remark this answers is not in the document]
leunam99: I had part of this file in another module, originally to break a dependency circle. Michael Schwarz moved it back when restructuring ([here](https://github.com/goblint/analyzer/pull/772/commits/4e21fe2379131970f396a58b4262986d2cc17dab))

leunam99:
> It would be good to have complete SV-COMP before vs after runs with this to see how well this ad-hoc stuff works.

I did this in my[ bachelor's thesis](https://github.com/goblint/analyzer/files/9602870/main.pdf): pages 19 and 20 contain tables with the results of the different features and some further configurations. In the text surrounding it I describe a bit further what tests exactly changed.

src/util/cilfacade.ml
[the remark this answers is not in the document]
michael-schwarz: I think the split that was there was a bit hap-hazard. But maybe we can pull out just the loop unrolling bits?

src/autoTune.ml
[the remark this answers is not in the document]
michael-schwarz: I probably re-introduced it when running `ocp-indent`. Feel free to change it again, `ocp-indent` sometimes does strange things, and we are fine with some local deviations from what it suggest.

src/analyses/base.ml
michael-schwarz: > Would it be enough for that to reset it at the same place as all the lazy maps?

I think for now that probably is the best option here. Compatibility with incremental is something that is probably out of scope for your thesis, and we can worry about it later.

Thinking of it, maybe we simply make a check that `annotation.array` and `incremental.load` are not activated at the same time for now?

src/autoTune0.ml
michael-schwarz: Maybe a comment will help here?

src/analyses/base.ml
sim642: > Would it be enough for that to reset it at the same place as all the lazy maps?

Doing the resetting in the `Spec`'s `init` and `finalize` should be enough, not to pull this table into an even more global scope.

> Thinking of it, maybe we simply make a check that `annotation.array` and `incremental.load` are not activated at the same time for now?

I think it might even be fine: if a function doesn't get incrementally reanalyzed, then this won't be looked up again anyway, and if it does, then it gets added to the map again. Resetting it in `init` would also give this behavior in server mode.

src/cdomains/valueDomain.ml
sim642: I think those could also be collected here, right now they're just ignored with wildcard (`_`).

This is in `bot_value`, but other functions like `top_value`/`init_value` have other similar changes, but without the `unrollType`. Shouldn't this attribute collecting be consistent everywhere?

leunam99: I thought it was simpler to use unrollType than to add another list parameter for manually collecting the type attributes.

It should be the same in the other functions, I just reverted it to the wrong version after trying something out.

src/analyses/base.ml
leunam99: I just reset it in finalize. I think normally that should be enough to do it at one point,  not before and after. Does it need both for incremental?

docs/user-guide/annotating.md
```diff
@@ -26,6 +26,29 @@ The following string arguments are supported:
 4. `apron.context`/`apron.no-context` to override the `ana.apron.context` option.
 5. `widen`/`no-widen` to override the `ana.context.widen` option.
 
+### Apron attributes
+The Apron library can be set to only track variables with the attribute `goblint_apron_track`
+
+### Array attributes
+Arrays can be annotated with the domain that should be used for it ("unroll", "parttioned", or "trivial"):
```
michael-schwarz, line 33:
```suggestion
Arrays can be annotated with the domain that should be used for it ("unroll", "partitioned", or "trivial"):
```

docs/user-guide/annotating.md
```diff
@@ -26,6 +26,29 @@ The following string arguments are supported:
...
+	int arr[5] __attribute__((goblint_array_domain("partitioned")));
+};
+```
+It is also possible to annotate a type, so that all arrays of this type without an own attribute will use this one: 
+
+```c
+typedef int unrollInt __attribute__((goblint_array_domain("trivial")));
+unrollInt x[4];
+```
+Lastly, pointer parameters can have this attribute and the array arguments may get converted to the choosen domain. This does not always happen because the pointers can change during the analysis which arrays they could be pointing at.
```
michael-schwarz, line 49: > because the pointers can change during the analysis which arrays they could be pointing at.

That sounds very confusing, maybe try to calrify?

src/analyses/base.ml
michael-schwarz: I think this is still the case. Usually, we align as follows

~~~OCaml
match foo with
| Some Bar -> ...
| Some Baz -> ...
| None -> ...
~~~

docs/user-guide/annotating.md
leunam99: How about this?

`One can also annotate pointer parameters. Inside the function of the parameter, Goblint tries to use the annotated domain for the arrays pointed at by that pointer. This is not guaranteed to work, as following the pointers is done only in the first analyzed function context.`

michael-schwarz: Better!

src/analyses/base.ml
sim642: There's now this:
```ocaml
  module VarMap = Map.Make(CilType.Varinfo)
  let array_map = Hashtbl.create 20
```
This `Hashtbl` is polymorphic, which isn't suitable for CIL types. Instead it should be something like
```ocaml
  module VarH = Hashtbl.Make(CilType.Varinfo)
  let array_map = VarH.create 20
```

src/analyses/base.ml
sim642: That should be enough for this indeed.

sim642:
Besides the couple of unresolved comments, the test group 58-array_annotations needs to be renumbered to 60-array_annotations to avoid conflict on master.

src/analyses/base.ml
leunam99: > This `Hashtbl` is polymorphic, which isn't suitable for CIL types
I do not understand what is meant here.

To clarify: This is a mapping in two steps: array_map maps a function varinfo to a map from an array varinfo to attributes. Do you mean I should make the second map a Hashtbl as well?

src/cdomains/valueDomain.ml
leunam99: When collecting the attributes manually with the code below, I get an Out of memory exception:

```
  let rec bot_value ?(varAttr=[]) ?(typAttr=[]) (t: typ): t =
    match t with
    [...]
    | TArray (ai, None, _) ->
      let typAttr = addAttributes (typeAttrs ai) typAttr in
      `Array (CArrays.make ~varAttr ~typAttr (IndexDomain.bot ()) (bot_value ai))
    [...]
    | TNamed ({ttype=t; _}, attr) -> bot_value ~varAttr ~typAttr:(addAttributes attr typAttr) t
```
I do not understand why this runs out of memory so quickly and the stats do not match. 
```
Memory statistics: total=23.51MB, max=6.63MB, minor=21.73MB, major=5.82MB, promoted=4.04MB
    minor collections=10  major collections=1 compactions=0


```

src/analyses/base.ml
sim642: Ah, right. I think it's fine if the outer hashmap is such `VarH` and the inner map is that `VarMap`.

src/cdomains/valueDomain.ml
sim642: Right now the CI shows out of memory exceptions for many tests even without that change, still with that `unrollType`. So there seems to be some other issue.

When running with `-v`, the full stacktrace of the error is shown, which starts with:
```
Raised by primitive operation at BatHashtbl.find_option.loop in file "src/batHashtbl.mlv", line 218, characters 9-16
Called from Goblint_lib__Base.MainFunctor.add_to_array_map in file "src/analyses/base.ml", line 110, characters 10-51
```
If I were to guess, the problem is the use of the polymorphic `Hashtbl` instead of `VarH` as suggested in another comment. Polymorphic hashtables use polymorphic equality `(=)` to compare elements, but that's incorrect on many CIL types, including `varinfo`. Since CIL data structures contain cyclic references, the default polymorphic equality will recurse forever and very quickly run out of memory (in the C runtime, which is probably why it doesn't show in GC stats). Using a hashtable with our custom equality that avoids the issue should fix the problem.

sim642 (approved):

PR #774, opened 2022-07-04
Detect renamed functions and global variables. 

TimOrtel:
This pull requests builds on top of [my pull request about detection renamed local variables](https://github.com/goblint/analyzer/pull/731).

It implements an algorithm, that can detect renamed functions and global variables on an incremental run. The algorithm is non recursive. 

The following changes have been made:
- Moved eqF from compareCIL.ml to detectGlobals.ml
- Removed logic that would just compare globals with matching names with my algorithm implemented in detectRenamedFunctions.ml
- compareAST.ml is now fully functional, because it cannot have side effects when two wrong functions are compared. Previous side-effective hacks in compareAST have been moved out and are now applied after it is clear that the functions actually match. 

The algorithm works like this (simplified):
1. Find functions and global variables with matching names. 
2. Test these function pairs:
2.1. Foreach function pair detected in 1.: Compare the functions bodies of the functions.
2.2. If a function reports back that it depends on a rename of a global variable/function, those dependecies are checked in the next step.
2.3. Foreach dependency, check if they do not rely on additional dependencies themselves. If this assumption holds for all dependencies, the original function pair from 2.1 is marked as renamed, otherwise it is marked as changed. 
3. Report back the results to compareCIL and apply the change data correctly. 

I added test cases, but they are not verified automatically, meaning they have to be verified manually.


sim642:
What's the explanation for closing this without merging?

TimOrtel:
I do not have time to work on this pull request anymore and I do not want to give the impression that I will be able to work on this in the future by keeping this open. However, if someone else wants to finish this PR feel free to reopen it.

michael-schwarz:
Ok, I re-opened it and assigned it to @stilscher, as she likely knows what needs to be done so it can be merged.

michael-schwarz:
@stilscher: What is the plan here? Having PRs go so stale is a bit problematic. We should either merge or abandon this.

stilscher:
Yes, another round of feedback would be great. I think I also still need to make sure that the new incremental cram tests are also automatically executed in the CI. Other than that I think it is mostly cleaned up and ready for review.

src/incremental/compareAST.ml
```diff
@@ -3,212 +3,284 @@ open CilMaps
 
 module StringMap = Map.Make(String)
 
-type method_rename_assumption = {original_method_name: string; new_method_name: string; parameter_renames: string StringMap.t}
-type method_rename_assumptions = method_rename_assumption VarinfoMap.t
+type method_rename_assumptions = varinfo VarinfoMap.t
+type glob_var_rename_assumptions = varinfo VarinfoMap.t
```
sim642, line 7: What's the meaning of these maps now? An assumption would be that the key `varinfo` is renamed to the value `varinfo`? Might be worth adding a comment about it.

src/incremental/compareAST.ml
```diff
@@ -3,212 +3,284 @@ open CilMaps
...
+    | Some now ->
+      now = name2
+    | None ->
+      name1 = name2 (*Var names differ, but there is no assumption, so this can't be good*)
+  )
+  else name1 = name2
+
+(*Creates the mapping of local renames. If the locals do not match in size, an empty mapping is returned.*)
+let create_locals_rename_mapping (originalLocalNames: string list) (updatedLocalNames: string list): string StringMap.t =
+  if (List.length originalLocalNames) = (List.length updatedLocalNames) then
```
sim642, line 33: I guess semgrep misses this for some reason, but `compare_lengths`.

src/incremental/compareAST.ml
```diff
@@ -3,212 +3,284 @@ open CilMaps
...
 and mem_typ_acc (a: typ) (b: typ) acc = List.exists (fun p -> match p with (x, y) -> a == x && b == y) acc (* TODO: seems slightly more efficient to not use "fun (x, y) ->" directly to avoid caml_tuplify2 *)
 
 and pretty_length () l = Pretty.num (List.length l)
 
-and eq_typ_acc (a: typ) (b: typ) ~(acc: (typ * typ) list) ~(rename_mapping: rename_mapping) =
+and eq_typ_acc ?(fun_parameter_name_comparison_enabled: bool = true) (a: typ) (b: typ) ~(rename_mapping: rename_mapping) ~(acc: (typ * typ) list) : bool * rename_mapping =
+  (* Registers a compinfo rename or a enum rename*)
+  let register_rename_on_success = fun rename_mapping compinfo_option enum_option ->
+    let maybeAddTuple = fun list option ->
+      Option.value ~default:list (Option.bind option (fun elem -> Some(elem :: list)))
```
sim642, line 111: `BatOption.map_default`?

src/incremental/compareAST.ml
```diff
@@ -3,212 +3,284 @@ open CilMaps
...
       )
       else (
         let acc = (a, b) :: acc in
-        let res = eq_compinfo compinfo1 compinfo2 acc rename_mapping && GobList.equal (eq_attribute ~rename_mapping ~acc) attr1 attr2 in
-        if res then begin
-          global_typ_acc := (a, b) :: !global_typ_acc;
-
-          (* Reset cnames and ckeys to the old value. Only affects anonymous structs/unions where names are not checked for equality. *)
-          compinfo2.cname <- compinfo1.cname;
-          compinfo2.ckey <- compinfo1.ckey;
```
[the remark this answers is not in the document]
stilscher: They do. They are executed in `performRenames` after a successful comparison. It was pulled out such that the comparison itself has no side effects in case an assumed match did not pass the comparison and is discarded again.
https://github.com/goblint/analyzer/blob/6ba71fc9a44d593a76b259ac2b6b0a6a209cc899/src/incremental/compareCIL.ml#L90-L95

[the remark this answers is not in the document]
sim642: Right, makes sense. Would the old comment about anonymous structs be still relevant (and should be copied over here) or not?

src/incremental/compareCIL.ml
```diff
@@ -192,28 +192,32 @@ let eqF_check_contained_renames ~renameDetection f1 f2 oldMap newMap cfgs gc_old
...
+          let glob_old = VarinfoMap.find f_old_var var_glob_old in
+          let glob_new = VarinfoMap.find f_new_var var_glob_new in
+          match VarinfoMap.find_opt f_old_var (fst fm) with
+          | None ->
+            let f_old = get_fundec glob_old in
+            let f_new = get_fundec glob_new in (* TODO: what happens if there exists no fundec for this varinfo? *)
+            if acc then
+              eqF_only_consider_exact_match f_old f_new ci fm oldMap newMap var_glob_old var_glob_new
+            else false, ci, fm
+          | Some v -> acc && v.vid = f_new_var.vid && List.mem {old=glob_old; current=glob_new} ci.unchanged, ci, fm
```
sim642, line 210: It seems to me that these `global_col` things may contain `varinfo`s and `fundec`s, which shouldn't be polymorphically compared like `List.mem` does.

`BatList.mem_cmp` or `List.exists` on only the IDs or something would be better.

src/incremental/compareCIL.ml
```diff
@@ -192,28 +192,32 @@ let eqF_check_contained_renames ~renameDetection f1 f2 oldMap newMap cfgs gc_old
...
       let globalDependenciesMatch, change_info, final_matches = VarinfoMap.fold (fun old_var new_var (acc, ci, fm) ->
+          let glob_old = VarinfoMap.find old_var var_glob_old in
+          let glob_new = VarinfoMap.find new_var var_glob_new in
           match VarinfoMap.find_opt old_var (fst fm) with
           | None ->
             if acc then
               compare_varinfo_exact old_var gc_old oldMap new_var gc_new newMap ci fm
             else false, ci, fm
-          | Some v -> v.vid = new_var.vid, ci, fm
+          | Some v -> acc && v.vid = new_var.vid && List.mem {old=glob_old; current=glob_new} ci.unchanged, ci, fm
```
sim642, line 220: Same `List.mem` here.

src/incremental/compareAST.ml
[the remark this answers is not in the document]
stilscher: I think it is still helpful and so I copied it.

michael-schwarz:
@stilscher should we aim to get this merged before we hit the 11 month anniversary of this PR next Monday?

stilscher:
I thought it would be good to wait for an approval in the reviews before merging.

PR #778, opened 2022-07-07
Python scripts for efficiency and precision experiments on big repositories

stilscher:
This PR contains scripts that are used for benchmarking the interactive setups. It can be used to measure efficiency (overall runtime) and precision (precision loss over a series of incrementally analyzed commits). 
The efficiency experiment analyzes each small commit in a given time span first from-scratch and then, base on this, incrementally, incrementally + incremental postsolver and incrementally + incremental postsolver + reluctant. A small commit is a commit that is no merge commit, has relevant changes and fewer relevant CLOC then `maxCLOC`. A relevant changed line of code is a changed line of code in any .c or .h file that has not been excluded with `diff_exclude` (see explanation below).
The precision experiment starts with a depth-first search for maximally long series of consecutive commits (within the given time span), where only the first may be a merge commit. The first commit in a series is analyzed from-scratch, every following commit is analyzed incrementally based on the previous result. The same (optimal) configuration is used for all analyses. The precision of the 1st, 2nd, 5th, 10th and 15th incrementally analyzed commit in a series is compared to a from-scratch run of the same commit. For the comparison, the results are joined over all contexts for each node.

The experiments are parallelized if multiple cores are available to reduce the time required for measuring. The tasks are distributed such that each core analyzes (approximately) the same number of commits, only a one analysis is run on a the same physical core at a time and the load is distributed between the L3 caches (the provided layout works for an Intel machine, when run on a different machine it might be necessary to adjust it).

The setup of the experiment is specified at the top of the script files. This includes 
- the url and name of the repository to be analyzed
- script for creating the compilation database (this varies for different repositories)
- configuration used for analysis
- time span of commits to be covered
- result directory
- maximum number of relevant changed lines of code for a commit to be considered a small commit (`maxCLOC`)
- `diff_exclude`: allows to specify directories that are not relevant (any change in these directories will be considered not relevant and will not be counted)

I settled for specifying them at the top of the file, because when running experiments for a specific setup multiple times it seems cumbersome to provide the specification in the command line each time. The absolute path to the analyzer and the number of cores to be used, however, varies depending on the machine on which the experiment is run on. Therefore, these are provided as command line parameters. 

The python library [pydriller](https://pydriller.readthedocs.io/en/latest/) is used to clone and traverse the commits of a git repository. Additionally a few other python libraries need to be installed.

The PR mainly includes the following scripts:
- `precision.py`: to run the precision experiments and collect the data in a single csv file
- `efficiency.py`: run the efficiency experiments and collect the data in a single json file
- `plot.py`: plot data and output statistical data using the data collected in the experiment runs (it offers some plotting methods f.e. for cumulative distribution graphs that are currently not really used, but might be interesting in future)
- `stats.py`: output statistical data about repository such as number of commits, number of merge commits,...
- (`utils.py`: contains some methods (like analyzing a certain commit) that can be reused)

Further, it introduces a new configuration option `incremental.ingnore-const-args`, that allows to specify functions for which the constant arguments are not compared. This helps to ignore changes that are due to `__LINE__` being passed as argument to `printf` (I added an example test case for this).


jerhard:
Should we maybe move the scripts in this PR to the `bench` repository? They seem mostly relevant for benchmarking.

src/util/options.schema.json
```diff
@@ -966,6 +966,15 @@
           "enum": ["ast", "cfg"],
           "default": "ast"
         },
+        "ignore-const-args": {
+          "title": "incremental.ignore-const-args",
+          "description": "List of functions for which constant parameters are not compared",
```
sim642, line 971: Might be good to have this description say that this is useful for `__LINE__` constants.

scripts/incremental/efficiency.py
```diff
@@ -0,0 +1,232 @@
...
+    if not only_collect_results and numcores > allowedcores:
+        print("Not enough physical cores on this maching (exist: ", avail_phys_cores, " allowed: ", allowedcores, ")")
+        exit()
+    # For equal load distribution, choose a processes to core mapping,
+    # use only physical cores and have an equal number of processes per cache.
+    # The layout of physical/logical cores and sharing of caches is machine dependent. To find out use: 'lscpu --all --extended'.
+    # For our test server:
+    coremapping1 = [i for i in range(numcores - numcores//2)]
+    coremapping2 = [i for i in range(avail_phys_cores//2, avail_phys_cores//2 + numcores//2)]
+    coremapping = [coremapping1[i//2] if i%2==0 else coremapping2[i//2] for i in range(len(coremapping1) + len(coremapping2))]
```
sim642, line 183: Since this sounds very specific to the TUM test server and is deep inside the script, it might be good to point out at the beginning of the script in a comment that `analyze_chunks_of_commits_in_parallel` might require this to be adjusted.

scripts/incremental/utils.py
```diff
@@ -0,0 +1,246 @@
...
+import shutil
+from pathlib import Path
+import subprocess
+from pydriller import Git
+import re
+import pandas
+import json
+import numpy as np
+import brokenaxes
+import matplotlib as mpl
```
sim642, line 11: Might be good to have a `requirements.txt` file for all the non-standard dependencies.

src/framework/constraints.ml
```diff
@@ -1382,10 +1382,10 @@ struct
     nh
 
   let compare (name1, name2) vh1 vh2 =
-    Printf.printf "\nComparing nodes precision of %s (left) with %s (right):\n" name1 name2;
+    let verbose = get_bool "dbg.verbose" in
+    if verbose then Printf.printf "\nComparing nodes precision of %s (left) with %s (right):\n" name1 name2;
     let vh1' = join_contexts vh1 in
     let vh2' = join_contexts vh2 in
-    let verbose = get_bool "dbg.compare_runs.diff" in
```
sim642, line 1388: Are these changes really necessary for the scripts to work?

sim642:
> Should we maybe move the scripts in this PR to the `bench` repository? They seem mostly relevant for benchmarking.

That's a reasonable idea. It probably requires changing some paths in the scripts though. Maybe even the confs can be moved there, such that they would be together with the scripts and there already are confs in the bench repo as well.

src/incremental/compareAST.ml
```diff
@@ -160,10 +160,17 @@ and eq_offset (a: offset) (b: offset) = match a, b with
 and eq_lval (a: lval) (b: lval) = match a, b with
     (host1, off1), (host2, off2) -> eq_lhost host1 host2 && eq_offset off1 off2
 
-let eq_instr (a: instr) (b: instr) = match a, b with
+let eq_instr (a: instr) (b: instr) =
+  let skip_int_args_funs = GobConfig.get_string_list("incremental.ignore-const-args") in
+  let comp_args f1 args1 args2 = match f1 with
+    | Lval (Var v1,_) when List.mem v1.vname skip_int_args_funs ->
```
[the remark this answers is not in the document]
stilscher: In our use case, it contains just a single function name.

scripts/incremental/efficiency.py
```diff
@@ -0,0 +1,232 @@
...
+        gr = Git(repo_path)
+
+        #print("\n" + commit.hash)
+        #print('changed LOC: ', commit.lines)
+        #print('merge commit: ', commit.merge)
+
+        # skip merge commits and commits that have no or less than maxCLOC of relevant code changes
+        relCLOC = utils.calculateRelCLOC(repo_path, commit, diff_exclude) # use this to filter commits by actually relevant changes
+        #print("relCLOC: ", relCLOC)
+        if relCLOC == 0 or (maxCLOC is not None and relCLOC > maxCLOC):
```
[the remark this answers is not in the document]
stilscher: Merge commits are skipped even earlier, because of `only_no_merge=True` in https://github.com/goblint/analyzer/blob/da910e16174240f201d537a230b7e14120a34267/scripts/incremental/efficiency.py#L55

src/framework/constraints.ml
stilscher: Without it, the output comparison log files get huge and very difficult to read or only open. My text editor always crashed whenever I tried which is very annoying. I was also worried, that it might slow down the experiments just by the number of lines that are written to the log.

scripts/incremental/plot.py
```diff
@@ -0,0 +1,143 @@
...
+    diff3 = df[utils.header_runtime_incr_posts_rel_child].astype('float') / df[utils.header_runtime_parent].astype('float')
+    for n, diff in [("incr", diff1), ("+ incr postsolver", diff2), ("+ reluctant", diff3)]:
+        print("80% quantile for", n, "compared to from-scratch analysis:", diff.quantile(q=0.8) * 100, "%")
+        print("75% quantile for", n, "compared to from-scratch analysis:", diff.quantile(q=0.75) * 100, "%")
+
+
+def paper_precision_graph(results_precision, filename, outdir):
+    df = utils.get_data_from_json(os.path.join(results_precision, filename))
+
+    lessprec1 = 'intermediate precision.1.precision.lessprec'
```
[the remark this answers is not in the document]
stilscher: The precision comparison (for measuring the loss of precision after a certain number of incrementally analyzed commits) is very expensive because an additional from-scratch analysis of the commit is necessary. To save some time, we decided, to only do a comparison after 1,2,5,10 and 15 incrementally analyzed commits. These numbers mean just that, that is, the number of nodes with a less precise result after 1 (or 2,...) incrementally analyzed commits compared to the from-scratch analysis.

src/framework/constraints.ml
```diff
@@ -1385,10 +1385,10 @@ struct
     nh
 
   let compare (name1, name2) vh1 vh2 =
-    Printf.printf "\nComparing nodes precision of %s (left) with %s (right):\n" name1 name2;
+    let verbose = get_bool "dbg.verbose" in
```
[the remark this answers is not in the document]
stilscher: No, that is a good point. I will change that. I didn't realize that this option is for the output.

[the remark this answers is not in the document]
stilscher: No, that is a good point. I will change that. I didn't realize that this option is for the output.

scripts/incremental/plot.py
```diff
@@ -0,0 +1,143 @@
...
+    diff2 = df[utils.header_runtime_incr_posts_child].astype('float') / df[utils.header_runtime_parent].astype('float')
+    diff3 = df[utils.header_runtime_incr_posts_rel_child].astype('float') / df[utils.header_runtime_parent].astype('float')
+    for n, diff in [("incr", diff1), ("+ incr postsolver", diff2), ("+ reluctant", diff3)]:
+        print("80% quantile for", n, "compared to from-scratch analysis:", diff.quantile(q=0.8) * 100, "%")
+        print("75% quantile for", n, "compared to from-scratch analysis:", diff.quantile(q=0.75) * 100, "%")
+
+
+def paper_precision_graph(results_precision, filename, outdir):
+    df = utils.get_data_from_json(os.path.join(results_precision, filename))
+
```
jerhard, line 102:
```suggestion
# Compute precision loss after x commits, where x is in {1, 2, 5, 10, 15}
```

sim642:
> In the GobCon today, we decided that we keep it as-is for now, and move them to the `bench` repo at some later point in time.

I guess now that there's no rush to put together an artifact, we could still directly put it to the bench repo.

PR #779, opened 2022-07-12
Pass Escaped Locals to Callee as They Need to Be Considered Globals

michael-schwarz:
Currently, escaped locals (pointer was saved into globals) are not correctly passed to callees, as `enter` filters via `Basetype.Variables.is_global k` instead of `is_global (Analyses.ask_of_ctx ctx) k`.

Looking up the corresponding value in the local state then leads to `bot`, which is ignored when merging over all possible addresses that a pointer may point to.

Minimal example:
~~~C
#include <assert.h>
int* ptr;
int nine = 9;

int other() {
    assert(*ptr == 8); //UNKNOWN!
}

int main()
{
    int g = 8;
    int top;

    if(top) {
        ptr = &g;
    } else {
        ptr = &nine;
    }

    other();
}
~~~

**TODO:** Investigate if this fixes the unsoundnesses for SV-Comp.


src/analyses/base.ml
```diff
@@ -2131,7 +2131,8 @@ struct
           ignore (Priv.enter_multithreaded (Analyses.ask_of_ctx ctx) (priv_getg ctx.global) (priv_sideg ctx.sideg) st);
         Priv.threadenter (Analyses.ask_of_ctx ctx) st
       ) else
-        let globals = CPA.filter (fun k v -> Basetype.Variables.is_global k) st.cpa in
```
sim642, line 2134: I think we should remove `Basetype.Variables.is_global` altogether, it seems to be only used in a few places and can clearly be misleading since it's only `x.vglob` check. The few remaining uses should either use the ask version or explicitly check the field if they for some reason shouldn't account for escapes.

sim642 (approved):
Good catch! I had forgotten that the escaping into globals applies to singlethreaded mode as well. Although the module name is still `ThreadEscape` and its top comment also just mentions escaping to threads. Maybe we should bring that up to date.

src/analyses/base.ml
michael-schwarz: Removed it now!

sim642 (approved):

PR #781, opened 2022-07-12
Modify Library Function Specification for assert to not dereference its argument

### LINKED ISSUE #765: Spurious warning `Write to unknown address: privatization is unsound.` even though no write occurs
~~~C
#include<pthread.h>
#include<assert.h>
int counter = 0;
pthread_mutex_t lock1;

void* producer(void* param) {
    pthread_mutex_lock(&lock1);
    counter = 0;
    pthread_mutex_unlock(&lock1);
}

void* consumer(void* param) {
    pthread_mutex_lock(&lock1);
    assert(counter >= 0);
    pthread_mutex_unlock(&lock1);
}


int main() {
    pthread_t thread;
    pthread_t thread2;
    pthread_create(&thread,NULL,producer,NULL);
    pthread_create(&thread2,NULL,consumer,NULL);
}
~~~

For this innocuous program, Goblint currently produces the following output:

```
./goblint 5.c --enable dbg.debug --enable dbg.regression
[Info][Unsound] Write to unknown address: privatization is unsound. (5.c:15:5-15:25)
Pruning result

        safe:            1
        vulnerable:      0
        unsafe:          0
        -------------------
        total:           1
```

The warning `[Info][Unsound] Write to unknown address: privatization is unsound. (5.c:15:5-15:25)` happening at `counter >= 0` is spurious as no actual write access is happening here.

michael-schwarz:
It seems like there was a mistake in the specification of the library function for `assert` as it was marked to dereference its argument. This caused the warnings in #765.

@sim642 does this fix makes sense like this? Are there other library functions that need to be fixed as well?

Closes #765

### DIFF
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -15,7 +15,7 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("realloc", special [__ "ptr" [r; f]; __ "size" []] @@ fun ptr size -> Realloc { ptr; size });
     ("abort", special [] Abort);
     ("exit", special [drop "exit_code" []] Abort);
-    ("assert", special [__ "cond" [r]] @@ fun cond -> Assert cond);
+    ("assert", special [__ "cond" []] @@ fun cond -> Assert cond);
   ]
 
 (** C POSIX library functions.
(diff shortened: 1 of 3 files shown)


michael-schwarz:
I attempted to have a test for this, but working with //NOWARN as I wanted to did not help here, as the succeeding assert produces a "warning" in itself.

sim642 (approved):
This absolutely makes sense. I guess I must've still been in the mindset of the old specifications.
Other uses of `r` in these descriptions seem appropriate on actual pointer arguments.

sim642:
> I attempted to have a test for this, but working with //NOWARN as I wanted to did not help here, as the succeeding assert produces a "warning" in itself.

Might be able to work around this by disabling `warn.assert` in this test.

PR #782, opened 2022-07-13
Remove erroneous call to `truncateCilint` in `min_for`

### LINKED ISSUE #769: Unsigned long interval unsoundness with refinement
On 37e0d70cd4c6e320c696a906896535a901aba246 we are unsound on the following program:
```c
// PARAM: --enable ana.int.interval --set ana.int.refinement once
#include <assert.h>

int main() {
  for (unsigned long i = 0; i < 1000; i++) {
    assert(1);
  }

  assert(1); // reachable
  return 0;
}
```
We claim that the second assert is dead code. The abstract value for `i` at the loop head is `def_exc:Unknown int([0,64]); intervals [0,255]`.
It's fine with `unsigned int` and just `long`.

michael-schwarz:
Calling `truncateCilint` in `min_for` hides overflows that might happen, so it is incorrect to call it here.

Closes #769

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -393,9 +393,6 @@ module Size = struct (* size in bits as int, range as int64 *)
   open Cil open Big_int_Z
   let sign x = if BI.compare x BI.zero < 0 then `Signed else `Unsigned
 
-  let max = function
-    | `Signed -> ILongLong
-    | `Unsigned -> IULongLong
   let top_typ = TInt (ILongLong, [])
   let min_for x = intKindForValue x (sign x = `Unsigned)
   let bit = function (* bits needed for representation *)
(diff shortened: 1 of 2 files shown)


src/cdomains/intDomain.ml
```diff
@@ -397,7 +397,7 @@ module Size = struct (* size in bits as int, range as int64 *)
     | `Signed -> ILongLong
     | `Unsigned -> IULongLong
   let top_typ = TInt (ILongLong, [])
-  let min_for x = intKindForValue (fst (truncateCilint (max (sign x)) x)) (sign x = `Unsigned)
+  let min_for x = intKindForValue x (sign x = `Unsigned)
```
jerhard, line 400: The function `max` defined above is now unused, one may remove it.

tests/regression/38-int-refinements/02-strange-ulong.c
```diff
@@ -0,0 +1,86 @@
+// PARAM: --enable ana.int.interval --set ana.int.refinement once
+#include <assert.h>
+
+int main() {
+  withint();
+  withuint();
+  withlong();
+  withlonglong();
+  withulonglong();
```
jerhard, line 9: Using functions before they were defined or declared produces a warning, so one might move the `main` down.

jerhard (approved):

PR #785, opened 2022-07-15
Cleanup reporting of dead code

### LINKED ISSUE #94: dbg.print_dead_code is not complete and reports statements inserted by CIL
The behavior of the dead code reporting might confuse people.
~~~c
#include <assert.h>

int main() {
  int x;
  if (1)
    x = 1;
  else
    x = 2;
  if (x)
    x = 1;
  else
    x = 2;
  assert(x > 1 && x < 0);
}
~~~

~~~sh
$ ./goblint --html dead.c
File 'dead.c':
  function 'main' has dead code on lines: 12, 14
Found dead code on 2 lines!
Dead code: the then branch over expression 'x > 1' is dead! (dead.c:13)
Dead code: the else branch over expression 'x' is dead! (dead.c:12)
Writing xml to temp. file: /var/folders/0j/1lq15k654zg7cf1rh9_gqty80000gn/T/ocaml61b1c3tmp
Time needed: 413 ms
~~~
image

It only reports dead code for nodes for which the solver calculated a value, but line 8 above is also dead (like all nodes that have no value).
On the other hand it reports about statements inserted by CIL, like the then-branch of the assert above.

Also, it might be useful to sort the output by line number.

michael-schwarz:
This addresses the shortcomings identified in #94:

- For dead branches, insert a message that the dead branch may have been inserted by CIL if the `synthetic` attribute of the location is set.
- Add new option `alldeadcode` that also reports on dead code in branches over constants that are not taken. Include hint that this option exists when `print_dead_code` is on, but `alldeadcode` is off.

TODO:

- [x] Wait for https://github.com/goblint/cil/pull/103 to be merged

Closes #94.

### DIFF
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -160,10 +160,11 @@ struct
       printf "Total lines (logical LoC): %d\n" (live_count + !count + uncalled_fn_loc); (* We can only give total LoC if we counted dead code *)
     );
     let str = function true -> "then" | false -> "else" in
+    let cilinserted = function true -> "(inserted by CIL) " | false -> "" in
     let report tv (loc, dead) =
       match dead, Deadcode.Locmap.find_option Deadcode.dead_branches_cond loc with
-      | true, Some exp -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "the %s branch over expression '%a' is dead" (str tv) d_exp exp
-      | true, None     -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "an %s branch is dead" (str tv)
+      | true, Some exp -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "the %s branch %sover expression '%a' is dead" (str tv) (cilinserted loc.synthetic) d_exp exp
+      | true, None     -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "an %s branch %sis dead" (str tv)  (cilinserted loc.synthetic)
       | _ -> ()
     in
     if get_bool "dbg.print_dead_code" then (
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -23,7 +23,7 @@ let spec_module: (module Spec) Lazy.t = lazy (
             |> lift (get_bool "ana.sv-comp.enabled") (module HashconsLifter)
             |> lift (get_bool "ana.sv-comp.enabled") (module WitnessConstraints.PathSensitive3)
             |> lift (not (get_bool "ana.sv-comp.enabled")) (module PathSensitive2)
-            |> lift (get_bool "dbg.print_dead_code") (module DeadBranchLifter)
+            |> lift (get_bool "ana.dead-code.lines" || get_bool "ana.dead-code.branches") (module DeadBranchLifter)
             |> lift true (module DeadCodeLifter)
             |> lift (get_bool "dbg.slice.on") (module LevelSliceLifter)
             |> lift (get_int "dbg.limit.widen" > 0) (module LimitLifter)
@@ -82,7 +82,7 @@ struct
     let dead_locations : unit Deadcode.Locmap.t = Deadcode.Locmap.create 10 in
     let module NH = Hashtbl.Make (Node) in
     let live_nodes : unit NH.t = NH.create 10 in
-    let count = ref 0 in (* Is only populated if "dbg.print_dead_code" is true *)
+    let count = ref 0 in (* Is only populated if "ana.dead-code.lines" is true *)
     let module StringMap = BatMap.Make (String) in
     let open BatPrintf in
     let live_lines = ref StringMap.empty in
@@ -149,7 +149,7 @@ struct
       M.msg_group Warning ~category:Deadcode "Function '%s' has dead code" f msgs
     in
     let warn_file f = StringMap.iter (warn_func f) in
-    if get_bool "dbg.print_dead_code" then (
+    if get_bool "ana.dead-code.lines" then (
       if StringMap.is_empty !dead_lines
       then printf "No lines with dead code found by solver.\n"
       else (
@@ -167,7 +167,7 @@ struct
       | true, None     -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "an %s branch %sis dead" (str tv)  (cilinserted loc.synthetic)
       | _ -> ()
     in
-    if get_bool "dbg.print_dead_code" then (
+    if get_bool "ana.dead-code.branches" then (
       let by_fst (a,_) (b,_) = Stdlib.compare a b in
       Deadcode.Locmap.to_list Deadcode.dead_branches_then |> List.sort by_fst |> List.iter (report true) ;
       Deadcode.Locmap.to_list Deadcode.dead_branches_else |> List.sort by_fst |> List.iter (report false) ;
@@ -550,7 +550,7 @@ struct
         | GFun (fn, loc) when is_bad_uncalled fn.svar loc->
             let cnt = Cilfacade.countLoc fn in
             uncalled_dead := !uncalled_dead + cnt;
-            if get_bool "dbg.uncalled" then
+            if get_bool "ana.dead-code.functions" then
               M.warn ~loc ~category:Deadcode "Function \"%a\" will never be called: %dLoC" CilType.Fundec.pretty fn cnt
         | _ -> ()
       in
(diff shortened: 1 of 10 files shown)


src/framework/control.ml
```diff
@@ -151,19 +151,21 @@ struct
     let warn_file f = StringMap.iter (warn_func f) in
     if get_bool "dbg.print_dead_code" then (
       if StringMap.is_empty !dead_lines
-      then printf "No lines with dead code found by solver (there might still be dead code removed by CIL).\n" (* TODO https://github.com/goblint/analyzer/issues/94 *)
+      then printf "No lines with dead code found by solver%s.\n" (if GobConfig.get_bool "alldeadcode" then "" else "(there might still be dead code removed by CIL, try running with --enable alldeadcode to see all) ") (* TODO https://github.com/goblint/analyzer/issues/94 *)
```
sim642, line 154: Is there any reason not to always report all dead code? Having to re-run analyses to _maybe_ find more dead code is not very convenient and creates unnecessary confusion between all the different modes, as opposed to just doing the most sensible thing out of the box.

michael-schwarz: I was thinking of program where maybe large parts are under an `if(0)`, and was worried a bit about degrading performance here.

jerhard: I am not sure about the performance impact of the additional impact of dead code on the analyzer either way. Anyways, if we decide that it's worth having this additional option, I would suggest that one adds remarks about the new option in the description string of `dbg.print_dead_code`.  Maybe one can even group the new option with the existing one in the schema.

sim642: I doubt there's any real-world difference since most programs don't contain a majority of code under `if(0)`. That code is parsed and Cabs is constructed anyway, so the savings are miniscule.

Regarding our dead code options, they should all be refactored to something consistent, e.g. `ana.dead.line`, `ana.dead.branch` and `ana.dead.function`. I also believe that they should all be enabled by default, because forgetting to enable them has hid unsoundness problems for so long. And again, the cost should be insignificant, because they're all basic postprocessing for the parts of code where the expensive transfer functions never even fired.

Also, we should avoid introducing new top-level options, especially for such technical things. Telling the user to try again with `alldeadcode` really has some "try harder" vibes and can potentially be very annoying. Imagine analyzing zstd for over an hour, just to have it essentially say "rerun the whole analysis just to have this extra bit of non-analysis processing that takes an extra fraction of a second". That's just annoying, frustrating and not very user-friendly.

If we really want there to be an option for this, I think we should have it in a new `cil.*` category of options, which essentially just pass through to CIL. It's such an implementation detail, that an user shouldn't have to know if they simply want to know about all of the dead code, as in the spirit of #785 in the first place.

src/framework/control.ml
```diff
@@ -151,19 +151,21 @@ struct
...
       printf "Total lines (logical LoC): %d\n" (live_count + !count + uncalled_fn_loc); (* We can only give total LoC if we counted dead code *)
     );
     let str = function true -> "then" | false -> "else" in
+    let cilinserted = function true -> "(possibly inserted by CIL) " | false -> "" in
     let report tv (loc, dead) =
       match dead, Deadcode.Locmap.find_option Deadcode.dead_branches_cond loc with
-      | true, Some exp -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "the %s branch over expression '%a' is dead" (str tv) d_exp exp
-      | true, None     -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "an %s branch is dead" (str tv)
+      | true, Some exp -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "the %s branch %sover expression '%a' is dead" (str tv) (cilinserted loc.synthetic) d_exp exp
+      | true, None     -> M.warn ~loc ~category:Deadcode ~tags:[CWE (if tv then 570 else 571)] "an %s branch %sis dead" (str tv)  (cilinserted loc.synthetic)
```
sim642, line 168: I'm not sure if `loc.synthetic` is the appropriate thing to base this output on, because the location being synthetic doesn't necessarily correspond to an `If` that CIL created. In some cases it does, e.g. in `a && b`, when `b` has a constant value, then there's a dead branch in that CIL-created `If` and its location is synthetic because there's no program point in the original source code between the evaluation of `a` and `b`, which is why the location is synthetic.

I haven't checked with this PR, but I'm suspicious about cases like the following:
* Again in `a && b`, when `a` has a constant value and there's a dead branch in that CIL-created `If`, then there is a program point (right before the whole expression) where an invariant assert (which the synthetic locations are for) about `a` could be inserted. So that location needn't be synthetic, it might currently be since synthetic locations are probably assigned more crudely than necessary, but if that is ever improved, then the output about a dead branch would actually become more wrong.
* When `for` loops are transformed by CIL, the branching node (for which the branching expression literally exists in the original source code!) has a synthetic location, because an (assert) statement cannot be inserted between the `for` initializer and condition.

Maybe it's something we should address separately in the future, for example by adding `synthetic` fields to CIL statements (such that it's explicit which `If`s were added by CIL) and variables (such that it's explicit which temporaries, etc it came up with) instead of semi-heuristics for these.

michael-schwarz: Yes, that sounds like a good idea. However, I'd think that we should do this in an new issue & PR.

src/framework/control.ml
michael-schwarz: I now removed the `alldeadcode` option again.

src/framework/control.ml
```diff
@@ -23,7 +23,7 @@ let spec_module: (module Spec) Lazy.t = lazy (
             |> lift (get_bool "ana.sv-comp.enabled") (module HashconsLifter)
             |> lift (get_bool "ana.sv-comp.enabled") (module WitnessConstraints.PathSensitive3)
             |> lift (not (get_bool "ana.sv-comp.enabled")) (module PathSensitive2)
-            |> lift (get_bool "dbg.print_dead_code") (module DeadBranchLifter)
+            |> lift (get_bool "ana.dead-code.lines" || get_bool "ana.dead-code.branches") (module DeadBranchLifter)
```
sim642, line 26: Shouldn't this just be the following?
```suggestion
            |> lift (get_bool "ana.dead-code.branches") (module DeadBranchLifter)
```

src/maingoblint.ml
```diff
@@ -51,15 +51,19 @@ let rec option_spec_list: Arg_complete.speclist Lazy.t = lazy (
       set_string "outfile" "result";
     if get_string "exp.g2html_path" = "" then
       set_string "exp.g2html_path" (Fpath.to_string exe_dir);
-    set_bool "dbg.print_dead_code" true;
+    set_bool "ana.dead-code.lines" true;
+    set_bool "ana.dead-code.branches" true;
+    set_bool "ana.dead-code.functions" true;
```
sim642, line 56: Given that these are the defaults, are they really necessary to force with HTML? If the user explicitly disables these and wants HTML, then they should be disabled. Of course the HTML won't report dead lines then, but that's what the user explicitly wanted, so no surprises there.

src/maingoblint.ml
```diff
@@ -51,15 +51,19 @@ let rec option_spec_list: Arg_complete.speclist Lazy.t = lazy (
...
     set_bool "g2html" true;
     set_string "result" "fast_xml"
   in
   let configure_sarif () =
     if (get_string "outfile" = "") then
       set_string "outfile" "goblint.sarif";
-    set_bool "dbg.print_dead_code" true;
+    set_bool "ana.dead-code.lines" true;
+    set_bool "ana.dead-code.branches" true;
+    set_bool "ana.dead-code.functions" true;
```
sim642, line 66: Is this even required at all for SARIF? I guess it was just copied from the HTML case above.

src/util/options.schema.json
```diff
@@ -1433,12 +1452,6 @@
           "type": "boolean",
           "default": false
         },
-        "print_dead_code": {
-          "title": "dbg.print_dead_code",
-          "description": "Print information about dead code",
-          "type": "boolean",
-          "default": false
-        },
```
sim642, line 1441: Just as a post-merge note: the renaming of `dbg.print_dead_code` and `dbg.uncalled` will probably require a bunch of renaming in the bench repo as well.

src/framework/control.ml
michael-schwarz: True!

src/maingoblint.ml
michael-schwarz: I removed it now.

src/util/options.schema.json
michael-schwarz: True!

sim642 (approved):
One last remark since this is supposed to close #94. That issue has an example program. We should check that this PR indeed fixes the problems on that example and possibly add it as a test with some `// WARN`.

michael-schwarz:
It does indeed solve the issues with that example program, but currently our regression test annotations can not currently express the needed properties, so we'll have to do without automatic tests for now.

PR #786, opened 2022-07-18
Precondition Invariant: Disjunct over matching state invariants 

jerhard:
This PR builds on #752. 
As already [discussed](https://github.com/goblint/analyzer/pull/752#pullrequestreview-1036962971), one must take into account that we may discover multiple contexts that fulfill a precondition we generate for a particular context. Thus, the postcondition of a `precondition_loop_invariant` has to be a disjunction over invariants obtained for "matching" unknowns. 

An alternative implementation may use `Union-Find` data structures to group preconditions together that do no exclude each other. While this alternative approach would reduce the number of precondition invariants generated, it may also lose some information.


jerhard:
An example where the invariants generated were previously inconsistent was added [here](https://github.com/goblint/analyzer/pull/786/commits/f58c4ece0ec733f0703325c9d3d38c786258d493).

jerhard:
Now, for the [example](https://github.com/goblint/analyzer/blob/e4fb85604e087c6dfd2b51b3c04b5e6f884e9abb/tests/regression/56-witness/05-prec-problem.c), instead of generating inconsistent/wrong `precondition_loop_invariant`s (see the [commit message](https://github.com/goblint/analyzer/pull/786/commits/f58c4ece0ec733f0703325c9d3d38c786258d493)):

We generate precondition invariants as follows:
```
  loop_invariant:
    string: result == 0 || result == 1
    [...]
  precondition:
    string: '*ptr1 == 5 && *ptr2 == 5'
    [...]
```
However, due to the way these precondition invariants are created, we still create one precondition invariant per context per program point. For example in this case, we generate the precondition invariant twice.

tests/regression/56-witness/05-prec-problem.c
```diff
@@ -0,0 +1,22 @@
...
+#include <stdlib.h>
+
+int foo(int* ptr1, int* ptr2){
+    int result;
+    if(ptr1 == ptr2){
+        result = 0;
+    } else {
+        result = 1;
+    }
+    // Look at the generated witness.yml to check whether there are contradictory precondition_loop_invariant[s]
```
sim642, line 11: This isn't great as a test, but without something like #787 we cannot do any better either.

src/witness/yamlWitness.ml
```diff
@@ -174,43 +218,107 @@ struct
             | None ->
               acc
           end
-        | _ -> (* avoid FunctionEntry/Function because their locations are not inside the function where assert could be inserted *)
+        end else begin
+          (* avoid FunctionEntry/Function because their locations are not inside the function where assert could be inserted *)
```
sim642, line 222: This comment might make more sense in the extracted `is_invariant_node`.

src/witness/yamlWitness.ml
```diff
@@ -174,43 +218,107 @@ struct
...
 
-    (* TODO: deduplicate *)
+
+    (* Generate precondition invariants.
+       We do this in three steps:
+       1. Collect contexts for each function
+       2. For each function context, find "matching"/"weaker" contexts that may satisfy its invariant
+       3. Generate precondition invariants. The postcondition is a disjunction over the invariants for matching states. *)
+
+    (* 1. Collect contexts for each function *)
```
sim642, line 235: Slightly annoying that #391 isn't merged yet, because there's an `IterSysVars` query that directly returns this. Maybe add a TODO here to use that once that's merged.

src/witness/yamlWitness.ml
```diff
@@ -124,6 +152,11 @@ struct
 
   module NH = BatHashtbl.Make (Node)
   module WitnessInvariant = WitnessUtil.Invariant (File) (Cfg)
+  module FMap = Prelude.Hashtbl.Make (CilType.Fundec)
+  module FCMap = Prelude.Hashtbl.Make (Printable.Prod (CilType.Fundec) (Spec.C))
```
sim642, line 156: Why `Prelude`?

src/witness/yamlWitness.ml
```diff
@@ -174,43 +218,107 @@ struct
...
+                    let x = ask_local (c.node, c.context) c.state (Queries.EvalInt c_inv) in
+                    if Queries.ID.is_bot x || Queries.ID.is_bot_ikind x then (* dead code *)
+                      failwith "Bottom not expected when querying context state" (* Maybe this is reachable, failwith for now so we see when this happens *)
+                    else if Queries.ID.to_bool x = Some false then () (* Nothing to do, the c does definitely not satisfy the predicate of current_c *)
+                    else begin
+                      (* Insert c into the list of weaker contexts of f *)
+                      FCMap.modify_def [] (f, current_c.context) (fun cs -> c::cs) fc_map;
+                    end
+                  | None ->
+                    (* If the context invariant is None, we will not generate a precondition invariant. Nothing to do here. *)
```
sim642, line 264: Is there a reason to even add them into this quadratic checking loop then?

sim642:
> However, due to the way these precondition invariants are created, we still create one precondition invariant per context per program point. For example in this case, we generate this precondition invariant twice.

Given the groupings, couldn't we avoid the duplication? Or would that not work to the possible non-symmetry of the relation?

src/witness/yamlWitness.ml
jerhard: Just a noisy way of using `BatHashbtl`. I changed it.

src/witness/yamlWitness.ml
jerhard: Good point, now I first check whether `current_c.invariant` is an actual expression, before iterating over all the other contexts of the function.
Such contexts cannot be excluded from the inner iteration though; thus I still keep them in the `fun_contexts` data structure.

jerhard:
>Given the groupings, couldn't we avoid the duplication? Or would that not work to the possible non-symmetry of the relation?

As the grouping is implemented now, the problem with deduplicating entries is indeed the possibility of non-symmetry. 
However, it should be possible to first group start-states according their invariant-representation: One would put start-states that lead to the same invariant in the same bucket; then, for each such bucket one would only have to generate on precondition-invariant. There one would still have to iterate over all other start states of the function though, to check whether they may make the precondition true. 

For this bucket-per-invariant approach, the invariants should best be usable as keys in some hashtable.

sim642 (approved):
Looks good! For the order of merging things I would propose the following:
1. Get #752 approved itself, because then its diff is easier to look at.
2. Then merge this into #752, so it becomes a complete implementation on its own.
3. Finally merge #752 into master, with changes from both included.

jerhard:
>Get https://github.com/goblint/analyzer/pull/752 approved itself, because then its diff is easier to look at.

What do you mean by that? How is the diff presented differently when the PR is approved?

sim642:
> What do you mean by that? How is the diff presented differently when the PR is approved? Can we not just merge this PR into #752?

I mean if this is merged into #752, then when reviewing #752 the diff also contains all the changes of this PR, as if they were made in #752. It just makes it more difficult to distinguish, which changes we already approved here and thus needn't be re-reviewed in #752.

But you can merge this into #752 right away as well, if you want. It doesn't really matter for me.

PR #788, opened 2022-07-19
Provide Common Configurations

### LINKED ISSUE #14: Provide common configurations
- [ ] Provide `JSON`-configurations for "commonly used configuration variables".
   - [ ] There should be a CI check to ensure they do not become outdated
- [ ] `goblint --help` Should mention example/meaningful configurations in the `conf` directory

michael-schwarz:
This provides three example configurations that someone new to Goblint might want to use.

- `very-precise.json`: Enables some of the more expensive abstract domains and features, especially useful for smaller programs
- `medium-program.json`: Enables some costly features, but staying away from the very expensive ones. This is very close to, e.g. the configuration we use for SV-COMP
- `large-programs.json`: Minimal configuration for larger programs, should run fast even for large programs, but usually needs to be made more precise by adding further features. (based on our `zstd` config)

I came up with these configurations more or less off the top of my head, feel free to suggest tweaks! Are there any other common configurations that we want to offer?

The configs are validated as part of our regression tests, so they do not become stale. The symlink for the tests is needed such that both `regtest.sh` and `update_suite` find the configuration file where they expect it to be.

Closes #14


conf/examples/README.md
```diff
@@ -0,0 +1,7 @@
+# Example Configurations for Goblint
```
sim642, line 1: I think it might be better to have this in our actual documentation instead. There's a "Configuring" page, which is practically empty.

conf/examples/large-program.json
```diff
@@ -0,0 +1,21 @@
+{
+    "ana": {
+      "activated": [
+        "base", "mallocWrapper","escape"
```
sim642, line 4: This is quite extreme, even deactivating `mutex`, which on any multithreaded program will probably produce a silly number of races. There's no criteria for being able to exclude any race.

conf/examples/large-program.json
```diff
@@ -0,0 +1,21 @@
...
+      },
+      "thread": {
+        "domain": "plain",
+        "include-node": false
+      }
+    },
+    "exp": {
+      "earlyglobs": true
+    },
+    "printstats": true
```
sim642, line 20: Slightly off-topic for an example conf.

conf/examples/medium-program.json
```diff
@@ -0,0 +1,50 @@
...
+      ]
+    },
+    "base": {
+      "arrays": {
+        "domain": "partitioned"
+      }
+    }
+  },
+  "exp": {
+    "region-offsets": true
```
sim642, line 47: Might be better to not set this as example: #107.

conf/examples/medium-program.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "solver": "td3"
```
sim642, line 49: Specifying solvers in example confs is also excessive I think. It's quite an internal thing and our default should be good without having it forced in the conf.

conf/examples/medium-program.json
```diff
@@ -0,0 +1,50 @@
...
+        "kmalloc",
+        "__kmalloc",
+        "usb_alloc_urb",
+        "__builtin_alloca",
+        "kzalloc",
+        "ldv_malloc",
+        "kzalloc_node",
+        "ldv_zalloc",
+        "kmalloc_array",
+        "kcalloc"
```
sim642, line 37: Not sure how meaningful this is for an example, these are very benchmark-specific.

src/maingoblint.ml
```diff
@@ -36,6 +36,7 @@ let print_help ch =
   fprintf ch " are used instead of double-quotes (\").\n\n";
   fprintf ch "A <jpath> is a path in a json structure. E.g. 'field.another_field[42]';\n";
   fprintf ch "in addition to the normal syntax you can use 'field[+]' append to an array.\n\n";
+  fprintf ch "Some common configurations to start from can be found in conf/examples/*\n";
```
sim642, line 39: This reminds me of something, which becomes more relevant if we want to provide standard confs: currently they only exist in a locally checked out git repository and are not available when `goblint` is directly installed via opam (or installed into the opam switch from git, but then executed in any other directory than the git repo root).
Maybe we should make them part of the installed package and do the CPAchecker-like thing of also looking `--conf` arguments up from that installation directory.

sim642:
> The configs are validated as part of our regression tests, so they do not become stale. The symlink for the tests is needed such that both `regtest.sh` and `update_suite` find the configuration file where they expect it to be.

The options CI workflow already should do that.

conf/dune
```diff
@@ -0,0 +1,10 @@
+
+(install
+  (section (site (goblint conf)))
+  (files
+    (examples/large-program.json as examples/large-program.json) ; must repeat directory, otherwise gets stripped from target
+    (examples/medium-program.json as examples/medium-program.json) ; must repeat directory, otherwise gets stripped from target
+    (examples/very-precise.json as examples/very-precise.json) ; must repeat directory, otherwise gets stripped from target
+    ; no way to glob here...
```
sim642, line 8: It might make sense to install our svcomp configs also.

src/util/gobConfig.ml
```diff
@@ -351,7 +351,15 @@ struct
 
   (** Merge configurations form a file with current. *)
   let merge_file fn =
-    let v = Yojson.Safe.from_channel % BatIO.to_input_channel |> File.with_file_in (Fpath.to_string fn) in
+    let file = Fpath.to_string fn in
+    let path = if Sys.file_exists file then file else
+        let configs = List.map Fpath.v Goblint_sites.conf in
+        let r = List.find_opt (fun v -> Fpath.filename fn = Fpath.filename v) configs in
```
sim642, line 357: Matching them by only looking at filename parts of paths is quite a weird way of finding files.
The normal way would be to just look `fn` up like this:
https://github.com/goblint/analyzer/blob/4f57a8dadf8e6b4b9579f5116273acc3d8cf4c04/src/maingoblint.ml#L188-L192
https://github.com/goblint/analyzer/blob/4f57a8dadf8e6b4b9579f5116273acc3d8cf4c04/src/maingoblint.ml#L203-L211

sim642 (approved):

PR #789, opened 2022-07-20
Mutex-Meet-TID Privatization for Base

### LINKED ISSUE #631: Enhance non-relational Mutex Meet with MHP
There is no reason we cannot also make use of MHP in the non-relational Mutex-Meet privatization.
This should be added such that we can, e.g. profit from MHP information to gain more precise information about structs.

Very little new code should need to be added to achieve this, as many things are the same between relational and non-relational MM.

michael-schwarz:
This adds a Mutex-Meet-TID based privatization for Base.

TODO:
- [ ] Tests
- [ ] See if more duplication can be avoided between the relational and non-relational analysis

During the process, it became clear that enhancing protection based with MHP should also be fairly trivial and may also pay off, so we should also take a look at that.

Closes #631

### DIFF
--- a/src/analyses/apron/apronPriv.apron.ml
+++ b/src/analyses/apron/apronPriv.apron.ml
@@ -829,49 +829,16 @@ module PerMutexMeetPrivTID (Cluster: ClusterArg): S  = functor (AD: ApronDomain.
 struct
   open CommonPerMutex(AD)
   include MutexGlobals
-  include ConfCheck.RequireThreadFlagPathSensInit
 
   module NC = Cluster(AD)
   module Cluster = NC
   module LAD = NC.LAD
 
-  module LLock =
-  struct
-    include Printable.Either (Locksets.Lock) (CilType.Varinfo)
-    let mutex m = `Left m
-    let global x = `Right x
-  end
-
-  (* Map from locks to last written values thread-locally *)
-  module L = MapDomain.MapBot_LiftTop (LLock) (LAD)
-
-  module LMust = struct
-    include SetDomain.Reverse (SetDomain.ToppedSet (LLock) (struct let topname = "All locks" end))
-    let name () = "LMust"
-  end
+  include PerMutexTidG(LAD)
 
   module D = Lattice.Prod3 (W) (LMust) (L)
-  module GMutex = MapDomain.MapBot_LiftTop (ThreadIdDomain.ThreadLifted) (LAD)
-  module GThread = Lattice.Prod (LMust) (L)
-  module G =
-  struct
-    include Lattice.Lift2 (GMutex) (GThread) (Printable.DefaultNames)
-
-    let mutex = function
-      | `Bot -> GMutex.bot ()
-      | `Lifted1 x -> x
-      | _ -> failwith "PerMutexMeetPrivTID.mutex"
-    let thread = function
-      | `Bot -> GThread.bot ()
-      | `Lifted2 x -> x
-      | _ -> failwith "PerMutexMeetPrivTID.thread"
-    let create_mutex mutex = `Lifted1 mutex
-    let create_global global = `Lifted1 global
-    let create_thread thread = `Lifted2 thread
-  end
 
   module AV = ApronDomain.V
-  module TID = ThreadIdDomain.Thread
 
   module V =
   struct
@@ -884,20 +851,6 @@ struct
 
   let name () = "PerMutexMeetPrivTID(" ^ (Cluster.name ()) ^ (if GobConfig.get_bool "ana.apron.priv.must-joined" then  ",join"  else "") ^ ")"
 
-
-  let compatible (ask:Q.ask) current must_joined other =
-    match current, other with
-    | `Lifted current, `Lifted other ->
-      if (TID.is_unique current) && (TID.equal current other) then
-        false (* self-read *)
-      else if GobConfig.get_bool "ana.apron.priv.not-started" && MHP.definitely_not_started (current, ask.f Q.CreatedThreads) other then
-        false (* other is not started yet *)
-      else if GobConfig.get_bool "ana.apron.priv.must-joined" && MHP.must_be_joined other must_joined then
-        false (* accounted for in local information *)
-      else
-        true
-    | _ -> true
-
   let get_relevant_writes (ask:Q.ask) m v =
     let current = ThreadId.get_current ask in
     let must_joined = ask.f Queries.MustJoinedThreads in
@@ -908,23 +861,7 @@ struct
           acc
       ) v (LAD.bot ())
 
-  let get_relevant_writes_nofilter (ask:Q.ask) v =
-    let current = ThreadId.get_current ask in
-    let must_joined = ask.f Queries.MustJoinedThreads in
-    GMutex.fold (fun k v acc ->
-        if compatible ask current must_joined k then
-          LAD.join acc v
-        else
-          acc
-      ) v (LAD.bot ())
-
-  let merge_all v =
-    GMutex.fold (fun _ v acc -> LAD.join acc v) v (LAD.bot ())
-
   type apron_components_t =  ApronDomain.ApronComponents (AD) (D).t
-
-
-  let startstate () = W.bot (), LMust.top (), L.bot ()
 
   let should_join _ _ = true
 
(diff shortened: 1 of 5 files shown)


sim642:
> During the process, it became clear that enhancing protection based with MHP should also be fairly trivial and may also pay off, so we should also take a look at that.

Here's an idea that may or may not be possible: instead of hardcoding a TID version of each privatization, could one just implement a TID lifting functor which lifts the inner privatization's globals to be partitioned by TID and takes care of reading the right ones?

michael-schwarz:
> implement a TID lifting functor which lifts the inner privatization's globals to be partitioned by TID and takes care of reading the right ones?

That certainly does sound appealing but I don't think it will not be possible in such a general manner.

tests/regression/58-base-mm-tid/21-broken.c
```diff
@@ -0,0 +1,41 @@
+// PARAM: --set ana.activated[+] symb_locks --set ana.activated[+] var_eq --set exp.extraspecials[+] ZSTD_customMalloc  --set ana.base.privatization mutex-meet-tid --set ana.path_sens[+] threadflag
```
sim642, line 1: Why is this test called "broken"? And what does `ZSTD_customMalloc` have to do with it?

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -834,49 +834,19 @@ module PerMutexMeetPrivTID (Cluster: ClusterArg): S  = functor (AD: ApronDomain.
...
-  module LMust = struct
-    include SetDomain.Reverse (SetDomain.ToppedSet (LLock) (struct let topname = "All locks" end))
-    let name () = "LMust"
-  end
+  include PerMutexTidG(struct
+      let exclude_not_started () = GobConfig.get_bool "ana.apron.priv.not-started"
+      let exclude_must_joined () = GobConfig.get_bool "ana.apron.priv.must-joined"
+    end)(LAD)
 
   module D = Lattice.Prod3 (W) (LMust) (L)
```
sim642, line 847: This should already come from `include PerMutexTidG`.

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -834,49 +834,19 @@ module PerMutexMeetPrivTID (Cluster: ClusterArg): S  = functor (AD: ApronDomain.
...
-      | _ -> failwith "PerMutexMeetPrivTID.thread"
-    let create_mutex mutex = `Lifted1 mutex
-    let create_global global = `Lifted1 global
-    let create_thread thread = `Lifted2 thread
-  end
 
   module AV = ApronDomain.V
-  module TID = ThreadIdDomain.Thread
 
   module V =
```
sim642, line 851: This should already come from `include PerMutexTidG`.

src/analyses/commonPriv.ml
```diff
@@ -160,33 +190,95 @@ struct
   end
 end
 
-module ConfCheck =
+module type PerMutexTidGArg = sig
+  val exclude_not_started: unit -> bool
+  val exclude_must_joined: unit -> bool
+end
+
+module PerMutexTidG (Conf:PerMutexTidGArg) (LD:Lattice.S) =
```
sim642, line 198: This doesn't just provide `G`, but `D`, etc as well, so maybe `PerMutexTidBase` is more appropriate.

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -1068,20 +1008,17 @@ struct
...
-      else (
+      match ConcDomain.ThreadSet.elements tids with
+      | [tid] ->
+        let lmust',l' = G.thread (getg (V.thread tid)) in
+        {st with priv = (w, LMust.union lmust' lmust, L.join l l')}
+      | _ ->
+        (* To match the paper more closely, one would have to join in the non-definite case too *)
+        (* Given how we handle lmust (for initialization), doing this might actually be beneficial given that it grows lmust *)
+        st
+      | exception _ ->
```
sim642, line 1019: Should catch the correct exception here.

src/analyses/basePriv.ml
```diff
@@ -1253,6 +1445,9 @@ struct
   let threadenter ask st = time "threadenter" (Priv.threadenter ask) st
   let iter_sys_vars getg vq vf = time "iter_sys_vars" (Priv.iter_sys_vars getg vq) vf
 
+  let thread_join ask get e st = time "thread_join" (Priv.thread_join ask) get e st
+  let thread_return ask get set tid st = time "thread_return" (Priv.thread_return ask) get set tid st
```
sim642, line 1449: We don't want to measure the instant partial application.
```suggestion
  let thread_join ask get e st = time "thread_join" (Priv.thread_join ask get e) st
  let thread_return ask get set tid st = time "thread_return" (Priv.thread_return ask get set tid) st
```

src/analyses/basePriv.ml
```diff
@@ -263,13 +272,8 @@ struct
   let read_global ask getg (st: BaseComponents (D).t) x =
     if is_unprotected ask x then (
       let get_mutex_global_x = get_mutex_global_x_with_mutex_inits getg x in
-      (* None is VD.top () *)
-      match CPA.find_opt x st.cpa, get_mutex_global_x with
-      | Some v1, Some v2 -> VD.meet v1 v2
-      | Some v, None
-      | None, Some v -> v
-      | None, None -> VD.bot () (* Except if both None, needed for 09/07 kernel_list_rc *)
```
sim642, line 271: Interesting that this isn't needed, because it must've been at some point. I think this was doing the "long" meet like `lock` does, but just for a single variable.

src/analyses/basePriv.ml
```diff
@@ -348,6 +352,188 @@ struct
       st
 end
 
+module PerMutexMeetTIDPriv: S =
+struct
+  include PerMutexPrivBase
+  include PerMutexTidG(struct
+      let exclude_not_started () = GobConfig.get_bool "ana.apron.priv.not-started"
+      let exclude_must_joined () = GobConfig.get_bool "ana.apron.priv.must-joined"
```
sim642, line 360: This base privatization still refers to Apron options. I don't think it's worth having separate options for these, rather maybe move them up like `ana.priv.not-started`.

src/analyses/basePriv.ml
```diff
@@ -348,6 +352,188 @@ struct
...
+
+  let long_meet m1 m2 = CPA.long_map2 VD.meet m1 m2
+
+  let get_mutex_global_g_with_mutex_inits inits ask getg g =
+    let get_mutex_global_g = get_relevant_writes_nofilter ask @@ G.mutex @@ getg (V.global g) in
+    let r = if not inits then
+        get_mutex_global_g
+      else
+        let get_mutex_inits = merge_all @@ G.mutex @@ getg V.mutex_inits in
+        let get_mutex_inits' = CPA.filter (fun k v -> k.vid = g.vid) get_mutex_inits in
```
sim642, line 376: Something like this should achieve the same:
```suggestion
        let get_mutex_inits' = CPA.singleton g (CPA.find g get_mutex_inits) in
```

src/analyses/basePriv.ml
```diff
@@ -348,6 +352,188 @@ struct
...
+    let tids = ask.f (Q.EvalThread exp) in
+    match ConcDomain.ThreadSet.elements tids with
+    | [tid] ->
+      let lmust',l' = G.thread (getg (V.thread tid)) in
+      {st with priv = (w, LMust.union lmust' lmust, L.join l l')}
+    | _ ->
+      (* To match the paper more closely, one would have to join in the non-definite case too *)
+      (* Given how we handle lmust (for initialization), doing this might actually be beneficial given that it grows lmust *)
+      st
+    | exception _ ->
```
sim642, line 484: Should catch the exact exception.

src/analyses/basePriv.ml
```diff
@@ -348,6 +352,188 @@ struct
...
+    else
+      let is_in_Gm x _ = is_protected_by ask m x in
+      let tid = ThreadId.get_current ask in
+      let sidev = GMutex.singleton tid (CPA.filter is_in_Gm st.cpa) in
+      sideg (V.mutex m) (G.create_mutex sidev);
+      let lm = LLock.mutex m in
+      let l' = L.add lm (CPA.filter is_in_Gm st.cpa) l in
+      {st with cpa = cpa'; priv = (w',LMust.add lm lmust,l')}
+
+  let thread_join (ask:Q.ask) getg exp (st: BaseComponents (D).t) =
```
sim642, line 473: Maybe this should handle the forceful assume-join case as well, like Apron's.

src/analyses/apron/apronPriv.apron.ml
michael-schwarz: What exactly are you referring to here?

src/analyses/commonPriv.ml
michael-schwarz: Renamed to `PerMutexTidCommon`

src/analyses/basePriv.ml
michael-schwarz: Good point! I added it!

src/analyses/basePriv.ml
michael-schwarz: I duplicated them for now so we still have fine-grained control while we are working on the more-traces paper. Once that is done, I think we can pull things up as you suggested.

src/analyses/basePriv.ml
michael-schwarz: I think it is not even an issue of it being spurious, but there actually was a case where it was unsound, as the comment indicates. Maybe when @stilscher  is done with verifying the side-effecting TD3, we could ask her to verify all our fabulous privatizations to avoid such issues :rofl:

src/analyses/apron/apronPriv.apron.ml
sim642: I mean the included `CommonPriv.PerMutexTidCommon` already contains `module D = Lattice.Prod3 (W) (LMust) (L)`.

src/analyses/apron/apronPriv.apron.ml
sim642: I mean the included `CommonPriv.PerMutexTidCommon` already contains `module V = ...`.

sim642 (approved):

PR #794, opened 2022-07-26
Make `mutex-meet-tid` Sound for Threads Exited by `pthread_exit(...)` & Improve Precision of `threadJoins` Analysis

### LINKED ISSUE #793: `mutex-meet-tid` Privatization Unsound for Threads Terminated with `pthread_exit(...)`
The `mutex-meet-tid` privatization is currently not aware of threads being terminated by `pthread_exit(...)`.
As a result, contributions of threads terminated with `pthread_exit(...)` are not properly taken into account when computing the join-local information for the thread calling `join(...)`.

In most cases this issue is hidden by the fact that threads are usually only exited one way. If that is via `pthread_exit(...)`, the thread return global in threadJoins is not set, which leads to the set of must joined threads being T, which accidently preserves soundness.

A problematic example is:
~~~C
// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --set ana.activated[+] threadJoins --sets ana.apron.privatization mutex-meet-tid
// Copy of 80-tid-toy10 making use of pthread_exit
#include <pthread.h>
#include <assert.h>

int g = 10;
int h = 10;

void *t_benign(void *arg) {
  int top;

  if(top) {
    pthread_mutex_lock(&A);
    g = 10;
    h = 10;
    pthread_mutex_unlock(&A);

    // If the analysis does unsoundly not account for pthread_exit, these writes are lost
    pthread_exit(NULL);
  }

  return NULL;
}

int main(void) {
  int t;

  // Force multi-threaded handling
  pthread_t id2;

  pthread_mutex_lock(&A);
  g = 12;
  h = 14;
  pthread_mutex_unlock(&A);

  pthread_join(id2, NULL);

  pthread_mutex_lock(&A);
  assert(g == 12); //UNKNOWN!
  pthread_mutex_unlock(&A);

  return 0;
}
~~~

Furthermore, `threadJoins` is needlessly imprecise in that it side-effects the set of must-joined threads at every `return` from a function instead of just doing it for actual returns from the thread.

michael-schwarz:
- Consider `pthread_exit` in `mutex-meet-tid` for soundness
- Improve `threadJoins` by only side-effecting set of must-joined TIDs on return from a thread, not just any random **return**

Closes #793

### DIFF
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -421,6 +421,8 @@ struct
     | ThreadExit _, _ ->
       begin match ThreadId.get_current ask with
         | `Lifted tid ->
+          (* value returned from the thread is not used in thread_join or any Priv.thread_join, *)
+          (* thus no handling like for returning from functions required *)
           ignore @@ Priv.thread_return ask ctx.global ctx.sideg tid st;
           raise Deadcode
         | _ ->
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -418,6 +418,14 @@ struct
         | Some lv -> invalidate_one ask ctx st' lv
         | None -> st'
       )
+    | ThreadExit _, _ ->
+      begin match ThreadId.get_current ask with
+        | `Lifted tid ->
+          ignore @@ Priv.thread_return ask ctx.global ctx.sideg tid st;
+          raise Deadcode
+        | _ ->
+          raise Deadcode
+      end
     | _, _ ->
       let lvallist e =
         let s = ask.f (Queries.MayPointTo e) in
--- a/src/analyses/threadJoins.ml
+++ b/src/analyses/threadJoins.ml
@@ -16,18 +16,21 @@ struct
   module G = MustTIDs
   module V = TID
 
-  (* transfer functions *)
-  let return ctx (exp:exp option) (f:fundec) : D.t =
+  let threadexit ctx =
     (
       match ctx.ask CurrentThreadId with
-      | `Lifted tid -> ctx.sideg tid ctx.local
+      | `Lifted tid when ThreadReturn.is_current (Analyses.ask_of_ctx ctx) -> ctx.sideg tid ctx.local
       | _ -> () (* correct? *)
     );
     ctx.local
 
+  (* transfer functions *)
+  let return ctx (exp:exp option) (f:fundec) : D.t = threadexit ctx
+
   let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
     let desc = LibraryFunctions.find f in
     match desc.special arglist with
+    | ThreadExit _ ->  threadexit ctx
     | ThreadJoin { thread = id; ret_var } ->
       let threads = ctx.ask (Queries.EvalThread id) in
       if TIDs.is_top threads then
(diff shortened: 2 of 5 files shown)


src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -418,6 +418,14 @@ struct
         | Some lv -> invalidate_one ask ctx st' lv
         | None -> st'
       )
+    | ThreadExit _, _ ->
+      begin match ThreadId.get_current ask with
+        | `Lifted tid ->
+          ignore @@ Priv.thread_return ask ctx.global ctx.sideg tid st;
```
sim642, line 424: Looking at just this, it seems something `return`-like should be happening here, i.e. to remove locals from `st`. The only reason it's fine right now is that no `Priv.thread_return` uses `st.apr`. Would be worth adding a comment about it here.

tests/regression/46-apron2/19-tid-toy-10-exit.c
```diff
@@ -0,0 +1,45 @@
...
+  int top;
+
+  if(top) {
+    pthread_mutex_lock(&A);
+    g = 10;
+    h = 10;
+    pthread_mutex_unlock(&A);
+
+    // If the analysis does unsoundly not account for pthread_exit, these writes are lost
+    pthread_exit(NULL);
```
sim642, line 20: Wait, if no `thread_return` uses `st.apr`, then how are these writes now being accounted for?

src/analyses/threadJoins.ml
```diff
@@ -16,18 +16,21 @@ struct
...
     );
     ctx.local
 
+  (* transfer functions *)
+  let return ctx (exp:exp option) (f:fundec) : D.t = threadexit ctx
+
   let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
     let desc = LibraryFunctions.find f in
     match desc.special arglist with
+    | ThreadExit _ ->  threadexit ctx
```
sim642, line 33: I think it isn't right now, because `threadexit` only side effects for thread returns, but `pthread_exit` may be called from some inner function in the thread and that should still account for a thread return, which currently would be missed.
The correct structure should be the one used by apron analysis, that `ThreadReturn` is only considered in `return` and `ThreadExit` does it unconditionally.

michael-schwarz: Nice catch!

tests/regression/46-apron2/19-tid-toy-10-exit.c
michael-schwarz: The key is that the join-local component of the more involved privatizations needs to account for writes to globals after the thread that is excited here is must-joined. For this reason, the `thread_return` side effects the local `lmust,l` information to a special unknown.

https://github.com/goblint/analyzer/blob/829e16a044925b0baf08febf3555572d3271c907/src/analyses/apron/apronPriv.apron.ml#L1064-L1067

src/analyses/apron/apronAnalysis.apron.ml
michael-schwarz: Good point, I added a comment!

src/analyses/threadJoins.ml
michael-schwarz: I added an example that was unsound before this fix.

tests/regression/46-apron2/19-tid-toy-10-exit.c
sim642: Right, the `L` component contains the written abstract values, not just `st.apr`.

sim642 (approved):

PR #795, opened 2022-07-27
Fix ARINC testing

### LINKED ISSUE #792: Arinc analysis completely broken
Closely inspecting our ARINC tests reveals that all of them that have any assert other than `//NOWARN`  in other processes than the main one are skipped.
Re-enabling them reveals that no other processes except the main one are live according to our analysis.

~~~C
void P1(void){
  // foo(0);
  // foo(1);
  RETURN_CODE_TYPE r;
  while (r){
    LAP_Se_WaitSemaphore(sem_id,600,&r);
    g = g + 1; // NOWARN!
    LAP_Se_SignalSemaphore(sem_id,&r);
  }
  LAP_Se_PeriodicWait(&r);
  LAP_Se_PeriodicWait(&r);
  return;
}

void P2(void){
  // foo(0);
  // foo(1);
  // here we only get 1 context for foo without bg b/c P2 may run arbitrarily
  // foo(0); bg = 1; foo(0);
  // different contexts for different arguments work fine:
  // foo(0); foo(1);
  RETURN_CODE_TYPE r;
  while (1){
    LAP_Se_WaitSemaphore(sem_id,600,&r);
    if(r == NO_ERROR)
        LAP_Se_PeriodicWait(&r);
    g = g - 1;    // NOWARN!
    g2 = g2 + 1;  // RACE!
    LAP_Se_SignalSemaphore(sem_id,&r);
  }
  return;
}

int main(){
  // here we get 2 contexts for foo with bg=0 and bg=1:
  // foo(0); bg = 1; foo(0);
  RETURN_CODE_TYPE r;
  PROCESS_ID_TYPE pi1, pi2;
  SEMAPHORE_ID_TYPE sem_id_local;
  PROCESS_ATTRIBUTE_TYPE p1, p2;
  LAP_Se_CreateSemaphore("my_mutex",1,1,0,&sem_id_local,&r);
  LAP_Se_GetSemaphoreId("my_mutex",&sem_id,&r);
  p1.NAME = "proc1";
  p1.ENTRY_POINT = (void *) &P1;
  p1.BASE_PRIORITY = 10;
  p1.PERIOD = 600;
  p1.TIME_CAPACITY = 600;
  p2.NAME = "proc2";
  p2.ENTRY_POINT = (void *) &P2;
  p2.BASE_PRIORITY = 10;
  p2.PERIOD = 600;
  p2.TIME_CAPACITY = 600;
  LAP_Se_CreateProcess(&p1,&pi1,&r);
  LAP_Se_CreateProcess(&p2,&pi2,&r);
  LAP_Se_Start(pi1,&r);
  LAP_Se_Start(pi2,&r);
  LAP_Se_SetPartitionMode(NORMAL,&r);
  while(1) {
    g2 = g2 - 1;  // RACE!
  }
  return 0;
}
~~~

Here, neither `P1` nor `P2` is alive. We either need to fix this analysis or remove it, having it is pointless in its current state.

Subsumes #26

sim642:
Closes #792. 17-arinc/01-sema is now unskipped and passes at least.

### Changes
1. Switches the tests from extract_arinc to arinc, which does properly spawn the processes.
2. Fixes the handling of `LAP_Se_SignalSemaphore` and `LAP_Se_WaitSemaphore` in mutexEvents, such that the test passes.
3. Fixes an `IntDomain` crash from arinc analysis.

17-arinc/03-preemt_lock still doesn't pass because it relies on some preemption-based privatization I guess, but that was commented out way back: b2fba1f43c402d3a3811502d6ea1079f22fe8d21.

### DIFF
--- a/src/analyses/mutexEventsAnalysis.ml
+++ b/src/analyses/mutexEventsAnalysis.ml
@@ -73,6 +73,7 @@ struct
       ctx.emit (Events.Lock (verifier_atomic, true))
 
   let special (ctx: (unit, _, _, _) ctx) lv f arglist : D.t =
+    let is_activated a = List.mem a (GobConfig.get_string_list "ana.activated") in (* TODO: proper LibraryFunctions group selection *)
     let remove_rw x = x in
     let unlock remove_fn =
       match f.vname, arglist with
@@ -82,7 +83,7 @@ struct
             ctx.split () [Events.Unlock (remove_fn e)]
           ) (eval_exp_addr (Analyses.ask_of_ctx ctx) arg);
         raise Analyses.Deadcode
-      | "LAP_Se_SignalSemaphore", [Lval arg; _] ->
+      | "LAP_Se_SignalSemaphore", [Lval arg; _] when is_activated "arinc" || is_activated "extract_arinc" ->
         List.iter (fun e ->
             ctx.split () [Events.Unlock (remove_fn e)]
           ) (eval_exp_addr (Analyses.ask_of_ctx ctx) (Cil.mkAddrOf arg));
@@ -101,7 +102,7 @@ struct
         | "spin_lock_irqsave", [arg; _] ->
           (*print_endline @@ "Mutex `Lock "^f.vname;*)
           lock ctx rw failing nonzero_return_when_aquired (Analyses.ask_of_ctx ctx) lv arg
-        | "LAP_Se_WaitSemaphore", [Lval arg; _; _]  ->
+        | "LAP_Se_WaitSemaphore", [Lval arg; _; _] when is_activated "arinc" || is_activated "extract_arinc" ->
           (*print_endline @@ "Mutex `Lock "^f.vname;*)
           lock ctx rw failing nonzero_return_when_aquired (Analyses.ask_of_ctx ctx) lv (Cil.mkAddrOf arg)
         | _ -> failwith "lock has multiple arguments"
--- a/src/analyses/mutexEventsAnalysis.ml
+++ b/src/analyses/mutexEventsAnalysis.ml
@@ -82,6 +82,11 @@ struct
             ctx.split () [Events.Unlock (remove_fn e)]
           ) (eval_exp_addr (Analyses.ask_of_ctx ctx) arg);
         raise Analyses.Deadcode
+      | "LAP_Se_SignalSemaphore", [Lval arg; _] ->
+        List.iter (fun e ->
+            ctx.split () [Events.Unlock (remove_fn e)]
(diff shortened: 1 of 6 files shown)


tests/regression/17-arinc/03-preemt_lock.c
```diff
@@ -1,4 +1,5 @@
-// SKIP PARAM: --set ana.activated[+] extract_arinc --set ana.activated[+] thread --disable ana.arinc.export
+// SKIP PARAM: --set ana.activated[+] arinc --set ana.activated[+] thread --disable ana.arinc.export
+#include <assert.h>
```
michael-schwarz, line 2: Maybe add a comment here on why this does not pass such that we don't re-investigate it every year? Or even mark the assertion as TODO?

src/analyses/mutexEventsAnalysis.ml
```diff
@@ -82,6 +82,11 @@ struct
             ctx.split () [Events.Unlock (remove_fn e)]
           ) (eval_exp_addr (Analyses.ask_of_ctx ctx) arg);
         raise Analyses.Deadcode
+      | "LAP_Se_SignalSemaphore", [Lval arg; _] ->
+        List.iter (fun e ->
+            ctx.split () [Events.Unlock (remove_fn e)]
+          ) (eval_exp_addr (Analyses.ask_of_ctx ctx) (Cil.mkAddrOf arg));
+        raise Analyses.Deadcode
```
michael-schwarz, line 89: Should we maybe make this conditional on (the/an) ARINC analysis being activated?
Just to be safe if we ever encounter a name clash?

src/analyses/mutexEventsAnalysis.ml
```diff
@@ -96,6 +101,9 @@ struct
         | "spin_lock_irqsave", [arg; _] ->
           (*print_endline @@ "Mutex `Lock "^f.vname;*)
           lock ctx rw failing nonzero_return_when_aquired (Analyses.ask_of_ctx ctx) lv arg
+        | "LAP_Se_WaitSemaphore", [Lval arg; _; _]  ->
+          (*print_endline @@ "Mutex `Lock "^f.vname;*)
+          lock ctx rw failing nonzero_return_when_aquired (Analyses.ask_of_ctx ctx) lv (Cil.mkAddrOf arg)
```
michael-schwarz, line 106: Same here

michael-schwarz (approved):

src/analyses/mutexEventsAnalysis.ml
sim642: The name clash already occurs at the level of `LibraryFunctions` where we currently don't have ARINC (and Linux kernel) functions categorized properly for separate activation, where this should really be done. But I added extra checks here nevertheless.

Issue #317, opened 2021-08-09
5306 programs fail with `texpr1_expr_of_cil_exp` with octApron and MutexMeet privatization


michael-schwarz:
When running SV-Comp on the current master with octApron enabled (everything else default settings), around 5000 programs fail with `Fatal error: exception Failure("texpr1_expr_of_cil_exp: globals must be replaced with temporary locals")`.

The most simple example program that fails that I could extract is this one:

~~~C
// SKIP PARAM: --sets ana.activated[+] octApron --enable ana.int.interval
int ARR_SIZE ;

int main() {
    int count = 0;
    while(count<ARR_SIZE) {
        count++;
    }
    return 0 ;
}
~~~

[List of all failing programs](https://github.com/goblint/analyzer/files/6953852/failing.txt)

sim642:
I haven't looked at the code but I suspect the issue might be something as simple as one of the following:

- [ ] Temporary locals not being introduced while still in single threaded mode.
- [ ] Temporary locals not being introduced in `branch`.

I wouldn't even be surprised if there's some TODO in the code.

PR #796, opened 2022-07-27
Unassume witness invariants during solving

sim642:
This PR adds an unassume analysis, which loads a YAML witness before solving and emits `Unassume` events during the analysis at nodes which have invariants. Both base, apron and var_eq analysis handle these events to make their local states less precise to immediately account for all states matching the invariant. Since the parsed invariants should be from a completed analysis, they correspond to the fixpoint. Thus unassuming the invariants allows the analysis to directly jump to that fixpoint. The benefits of this are two-fold:
1. It avoids widening-narrowing iterations to reach the same fixpoint, speeding up the analysis (hopefully).
2. It avoids widening jumping over fixpoints (especially with apron that cannot narrow), making the analysis more precise.

### Changes
Besides the unassuming, many surrounding bigger and smaller changes are included:
1. Extract base analysis branch `invariant` out to a separate `BaseInvariant` module. This is a fairly large part of base analysis code, but sufficiently self-contained that it can be decoupled. The real reason for doing this was to use a slightly modified version of `invariant` for unassuming without a complete copy of `invariant`.
2. Add _widening tokens_, a dynamic way to delay widenings: http://www2.in.tum.de/bib/files/mihaila13widening.pdf. This turned out to be necessary, otherwise unassume itself causes a widening and becomes even less precise than intended.
3. Add `witness.invariant.exact`, which allows disabling the generation of exact equality invariants that are useless for unassuming.
4. Add `witness.invariant.exclude-vars`, which is a list of regexes for variable names that invariants should not be generated for. This can be tuned for sv-benchmarks which have already been CIL-led and contain awful amounts of temporaries.
5. Improve base branch refinement for address sets, interval bound non-equality and disjunctions of address/int equalities.

### TODO
- [x] Emit `Unassume` events after all transfer functions, not just `assign`.
- [x] Support unassuming preconditioned invariants.
- [x] Support unassuming multithreaded invariants about globals (at locations)? This kind of requires us to first generate such invariants properly. (#833)
- [x] More precise unassuming than doing an assume on a top state and joining that in? In base, this probably would be an `invariant`-like recursive computation; in apron, not sure if anything better can be done.
- [x] Clean up and document widening tokens.
- [x] Adapt tests to #845.


src/analyses/base.ml
```diff
@@ -2569,6 +2576,52 @@ struct
...
+          ; local
+          }
+        and query': type a. Queries.Set.t -> a Queries.t -> a Queries.result = fun asked q ->
+          let anyq = Queries.Any q in
+          if Queries.Set.mem anyq asked then
+            Queries.Result.top q (* query cycle *)
+          else (
+            let asked' = Queries.Set.add anyq asked in
+            match q with
+            | MayEscape _ -> false (* HACK: avoids get_var in inv_exp thinking non-first var is global *)
```
sim642, line 2605: There's a major problem here in base for using `invariant` on a top state to construct the unassumed state. Since `invariant` is not executed on `ctx.local`, but requires a `ctx`, a new `ctx` has to be constructed with the altered state. Doing so also means that queries inside `invariant` are only answered by base itself, so multithreadedness and escape queries always return top, causing evaluations inside `invariant` to treat locals as globals (and call privatization), which breaks everything.

sim642:
As discussed in GobCon, the idea of trying to unassume hints only once can be quite problematic and thus will not be added to this. In all the reasonable cases when witnesses have true invariants and they are unassumed conservatively (which isn't currently necessarily the case), narrowing down from an overly imprecise invariant should be unnecessary.

src/analyses/base.ml
sim642: To not be unsound, I made this delegate some queries outside to `ctx.ask`, but others are answered by base using the modified local state.

sim642:
This is also a fairly big PR and has changes that often conflict with other changes on master, so it would be great to also get this merged soon.

bench/zarith/benchZarith.ml
```diff
@@ -0,0 +1,47 @@
+(* dune exec bench/zarith/benchZarith.exe -- -a *)
```
michael-schwarz, line 1: Maybe we can add a comment here on what exactly this is benchmarking and why it is needed?

tests/regression/36-apron/38-branch-global.c
```diff
@@ -1,10 +1,15 @@
-// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.int.interval
```
michael-schwarz, line 1: I am a bit worried that this changes what the test tests. Now there is a concrete upper bound upper available, the  bug in the apron analysis may not be triggered anymore...

If this test is also helpful here, it may make sense to duplicate it.

src/domains/invariantCil.ml
```diff
@@ -18,7 +18,7 @@ let exp_replace_original_name e =
 
 let var_is_in_scope scope vi =
   match Cilfacade.find_scope_fundec vi with
-  | None -> true
+  | None -> vi.vstorage <> Static (* CIL pulls static locals into globals, but they aren't syntactically in global scope *)
```
michael-schwarz, line 21: Nice catch!

src/solvers/localFixpoint.ml
```diff
@@ -0,0 +1,24 @@
+(** Local fixpoint iteration solvers that don't use a constraint system. *)
```
michael-schwarz, line 1: In what sense is this solver local? There is only one unknown?

src/analyses/unassumeAnalysis.ml
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  let invs: inv NH.t = NH.create 100
+
+  let fun_pres: Cil.exp FH.t = FH.create 100
+  let pre_invs: inv EH.t NH.t = NH.create 100
+
+  let init _ =
+    locator := Locator.create (); (* TODO: add Locator.clear *)
+    loop_locator := Locator.create (); (* TODO: add Locator.clear *)
```
michael-schwarz, line 42: Should these TODOs be done or are they obsolete?

src/analyses/unassumeAnalysis.ml
```diff
@@ -0,0 +1,276 @@
+(** Unassume analysis. *)
```
michael-schwarz, line 1: Maybe add a comment that the purpose of this analysis is to emit appropriate unassume events for the analyses to consume?

michael-schwarz:
I think once the comments are addressed and the adaptations necessary because of the new affeq analysis are made, this is ready to be merged.

src/solvers/localFixpoint.ml
sim642: In the sense that it's fixpoint iteration local to a single transfer function evaluation.

This is terminology from Miné (https://hal.sorbonne-universite.fr/hal-01657536), which originates from https://link.springer.com/chapter/10.1007/3-540-56287-7_95.

tests/regression/36-apron/38-branch-global.c
sim642: There was also a concrete upper bound before. Being global, `ARR_SIZE` was implicity 0, so the test was completely trivial by even def_exc, not to mention interval (which is also enabled for no reason) and apron itself.

This test is for #317 and it still checks that apron doesn't crash on global variables in branch expressions. But now it additionally checks that it actually refines them as well.

src/solvers/localFixpoint.ml
michael-schwarz: Ah, ok, I was unaware that there are competing definitions of local in this space, our definition for local solver is that it only computes values for those unknowns needed for a query. Maybe this makes that clear?

```suggestion
(** Fixpoint iteration solvers local to a single transfer function (don't use a constraint system). *)
```

tests/regression/36-apron/38-branch-global.c
michael-schwarz: Ah, yes, I didn't realize it was a global.

michael-schwarz (approved):

PR #802, opened 2022-07-28
Fix Crash When the Index Offset has Abstract Type `Address`

michael-schwarz:
In order to keep `Address` values as long as possible, we keep them even for values that are of integer type sometimes. This caused a crash in https://github.com/goblint/bench/issues/19 when such a value is used to index into an array.


src/analyses/base.ml
```diff
@@ -929,12 +929,14 @@ struct
...
     | Index (exp, ofs) ->
-      let exp_rv = eval_rv a gs st exp in
-      match exp_rv with
-      | `Int i -> `Index (iDtoIdx i, convert_offset a gs st ofs)
-      | `Top   -> `Index (IdxDom.top (), convert_offset a gs st ofs)
-      | `Bot -> `Index (IdxDom.bot (), convert_offset a gs st ofs)
-      | _ -> failwith "Index not an integer value"
+      let rec from_exp e = match e with
+        | `Int i -> `Index (iDtoIdx i, convert_offset a gs st ofs)
+        | `Address add -> from_exp @@ VD.cast (TInt(Cilfacade.ptrdiff_ikind (),[])) e
```
sim642, line 934: Is the cast guaranteed to give a non-`Address`? Since we sometimes keep `Address` for integer-typed variables, could the cast not keep `Address` and go into an infinite loop?

michael-schwarz: I don't think that's an issue here: For `VD.cast (...)` we actually go to `Int`

https://github.com/goblint/analyzer/blob/33c0b5dd0fbe07e12936687a3f6a3fda6f796607/src/cdomains/valueDomain.ml#L352-L363

sim642: Not right now, but if someone in the future for whatever reason changes `VD.cast` such that this assumption no longer holds and isn't aware of this particular use of `VD.cast` (among many), it will introduce a nasty non-termination bug on a program the size of SQLite. It won't even crash with a stack overflow, because this is a tail call.

The least we could do is check that the result is an `Int`, although it might be cleaner to extract part of that logic into `AddressSet.to_int` (there already is a corresponding `of_int`) and call it directly.

sim642 (approved):

PR #805, opened 2022-07-29
Clean up test `PARAM`s

### LINKED ISSUE #798: Remove explicit `ana.activated` lists in tests
Looking at the changes, it's a bit ugly that we have so many reg. tests using `--set ana.activated` instead of `--sets ana.activated[+]`. I guess most just add some analysis.
Maybe sth to change in a sep. PR?

_Originally posted by @vogler in https://github.com/goblint/analyzer/issues/278#issuecomment-878319858_

---

This has been a long-standing issue, which is a major inconvenience every time a new default analysis is added/removed, because there's an ungodly amount of tests that hardcode a list of `ana.activated` instead of using `ana.activated[+]` or `ana.activated[-]` to only specify which analyses they really care about.

This requires manually going through all such `PARAM` definitions and determining which analyses a test intends to (de)activate compared to the defaults (if any!). It's better to do it once and for all, rather than every single time we change the default analyses.

Should be done after #278.

sim642:
Closes #798 among other things:
1. Removes all explicit `ana.activated` lists (except a few intentional ones) and replaces them by relative list changes or removes altogether if the test has nothing to do with deactivated analyses.
2. Remove irrelevant `--set ana.base.privatization none` from tests. In most cases these have nothing to do with the test (which often is a single-threaded program altogether). Overall orthogonal features should work with all privatizations.
3. Remove irrelevant `--set solver` from tests. In most cases these just set `td3`, but all orthogonal features should work soundly with all _generic_ solvers. Only a handful of tests are for a specific solver.
4. Remove `dbg.debug` from tests. This is always enabled by default.
5. Remove irrelevant `--set ana.base.arrays.domain partitioned` from tests. In most cases these have nothing to do with the test (which has no arrays at all).

A large proportion of affected tests are partitioned array ones and all sorts of copies of them that have copied the complete `PARAM` without testing anything related any more.

### DIFF
--- a/tests/regression/01-cpa/28-interval.c
+++ b/tests/regression/01-cpa/28-interval.c
@@ -1,4 +1,4 @@
-// PARAM: --set solver "'new'" --set ana.int.interval true
+// PARAM: --set ana.int.interval true
 #include<stdio.h>
 #include<assert.h>
 
--- a/tests/regression/02-base/08-glob_interval.c
+++ b/tests/regression/02-base/08-glob_interval.c
@@ -1,4 +1,4 @@
-// PARAM: --set solver "'new'" --set ana.int.interval true
+// PARAM: --set ana.int.interval true
 #include<pthread.h>
 #include<assert.h>
 
--- a/tests/regression/02-base/87-casts-dep-on-param.c
+++ b/tests/regression/02-base/87-casts-dep-on-param.c
@@ -1,4 +1,4 @@
-// PARAM: --set ana.ctx_insens[+] 'base'  --set ana.base.privatization none
+// PARAM: --set ana.ctx_insens[+] 'base'
 #include<stdlib.h>
 int f(int, void*);
 
--- a/tests/regression/02-base/87-casts-dep-on-param.c
+++ b/tests/regression/02-base/87-casts-dep-on-param.c
@@ -1,4 +1,4 @@
-// PARAM: --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] 'base'  --set ana.base.privatization none
+// PARAM: --set ana.ctx_insens[+] 'base'  --set ana.base.privatization none
 #include<stdlib.h>
 int f(int, void*);
 
--- a/tests/regression/02-base/87-casts-dep-on-param.c
+++ b/tests/regression/02-base/87-casts-dep-on-param.c
@@ -1,4 +1,5 @@
-// PARAM: --set ana.ctx_insens[+] 'base'
+// PARAM: --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] 'base' --set ana.ctx_insens[+] 'mallocWrapper'
+// should be fully context-insensitive to activate minimal analyses
 #include<stdlib.h>
 int f(int, void*);
 
--- a/tests/regression/02-base/87-casts-dep-on-param.c
+++ b/tests/regression/02-base/87-casts-dep-on-param.c
@@ -1,5 +1,6 @@
-// PARAM: --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] 'base' --set ana.ctx_insens[+] 'mallocWrapper'
+// PARAM: --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] 'base' --set ana.ctx_insens[+] 'mallocWrapper' --set ana.base.privatization none
 // should be fully context-insensitive to activate minimal analyses
+// none privatization because mutex deactivated
 #include<stdlib.h>
 int f(int, void*);
 
--- a/tests/regression/29-svcomp/21-issue-casting.c
+++ b/tests/regression/29-svcomp/21-issue-casting.c
@@ -1,3 +1,5 @@
+// PARAM: --set ana.activated ["'base'","'mallocWrapper'"]
+// minimal analyses to reveal bug
 static long main(void)
 {
   unsigned int cmd;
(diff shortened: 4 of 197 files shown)


michael-schwarz:
While I agree with trying to remove `ana.activated` lists wherever it make sense, I'm a bit skeptical about

> Remove irrelevant `--set ana.base.privatization none` from tests. In most cases these have nothing to do with the test (which often is a single-threaded program altogether). Overall orthogonal features **should** work with all privatizations. [emphasis mine]

and 

>Remove irrelevant `--set ana.base.arrays.domain partitioned` from tests. In most cases these have nothing to do with the test (which has no arrays at all).

It may very well be that these were added to tests for re-appearances of old bugs in precisely those configs. I think we should aim to have more tests for non-standard configs rather than less.

tests/regression/01-cpa/28-interval.c
```diff
@@ -1,4 +1,4 @@
-// PARAM: --set solver "'new'" --set ana.int.interval true
```
michael-schwarz, line 1: Why remove this? i think it is one of the very few (if only) tests of the `new` solver

tests/regression/02-base/08-glob_interval.c
```diff
@@ -1,4 +1,4 @@
-// PARAM: --set solver "'new'" --set ana.int.interval true
```
michael-schwarz, line 1: Why not test the `new` solver?

tests/regression/02-base/87-casts-dep-on-param.c
```diff
@@ -1,4 +1,4 @@
-// PARAM: --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] 'base'  --set ana.base.privatization none
```
michael-schwarz, line 1: This may not test what we looking for when we have other analyses context-sensitive, because they may keep these contextes apart again. I'd vote for keeping this explicit list, unless we have a way to make all analyses context-insensitive.

tests/regression/01-cpa/28-interval.c
sim642: Look at this test, it's for testing fundamental interval analysis that should in no way be specific to some historical solver. If you want to test non-default solvers, then you can change the default and run the tests. The removal of these hardcoded `PARAM`s makes it possible to actually run everything with alternative defaults. As it is hardcoded, there's no way to run this with TD3. "New" is far from the only solver that we're not automatically testing right now. To test all solvers properly, you'd have to do a testing matrix with all the different solvers as default and rerun the whole suite, but that hasn't been nor is the goal right now.

tests/regression/02-base/08-glob_interval.c
sim642: Same here, this is a fundamental test about intervals on global variables.

tests/regression/02-base/87-casts-dep-on-param.c
sim642: Sure, I can make it explicit again. As-is this test had no indication that it is strictly about no context sensitivity at all. By the way, mallocWrapper has context too, so for completeness that should be made insensitive as well.

tests/regression/01-cpa/28-interval.c
michael-schwarz: Yes, but I'd still be opposed to leaving things worse than they are currently :wink:

tests/regression/29-svcomp/21-issue-casting.c
```diff
@@ -1,4 +1,3 @@
-// PARAM: --set ana.activated ["'base'","'mallocWrapper'","'assert'"] --set ana.base.privatization none
```
michael-schwarz, line 1: IIRC this minimal config was also because other analyses masked the bug

tests/regression/46-apron2/08-escape-local-in-pthread-dummy.c
```diff
@@ -1,4 +1,4 @@
-// SKIP PARAM: --set solver td3 --set ana.activated "['base','threadid','threadflag','mallocWrapper','assert','apron','escape']" --set ana.base.privatization none --set ana.apron.privatization top
```
michael-schwarz, line 1: Disabling base privatization is useful in all tests of apron privatization!

michael-schwarz (changes requested):
Other than the issues I mentioned above, I think this is good to merge!

tests/regression/01-cpa/28-interval.c
sim642: I'd prefer we test the solver we actually use, not some random old solver though.

sim642:
> While I agree with trying to remove `ana.activated` lists wherever it make sense, I'm a bit skeptical about
> 
> > Remove irrelevant `--set ana.base.privatization none` from tests. In most cases these have nothing to do with the test (which often is a single-threaded program altogether). Overall orthogonal features **should** work with all privatizations. [emphasis mine]
> 
> and
> 
> > Remove irrelevant `--set ana.base.arrays.domain partitioned` from tests. In most cases these have nothing to do with the test (which has no arrays at all).
> 
> It may very well be that these were added to tests for re-appearances of old bugs in precisely those configs.

These irrelevant removals are in tests that never invoke the features under question. If a program doesn't create any threads (and doesn't have `earlyglobs`), then it doesn't ever use privatization. Tests that create at least one thread still have the argument. If a program doesn't contain any arrays, then it doesn't ever create an array abstraction (partitioned or otherwise). Tests that have at least one array still have the argument.
Each of the removals is a separate commit, so it's easy to check this by looking at the individual commit diff.

By the way, most of the removed `ana.base.privatization` options were not originally in the tests but added by #330. By the same argument, that also changed the original intent of the tests and should not have happened, so this actually is restoring their original intended configuration.

tests/regression/01-cpa/28-interval.c
sim642: I added a copy of this test with alternative `PARAM`.

tests/regression/02-base/08-glob_interval.c
sim642: I added a copy of this test with alternative `PARAM`.

michael-schwarz (approved):
Thank you for addressing my comments!

Issue #803, opened 2022-07-28
Improve HoarePO.equal and leq performance


jerhard:
## Findings

@michael-schwarz and I did some runs of Goblint (`master`) on the SQLite amalgamation, and investigated performance using `perf` and the [firefox-profiler](https://profiler.firefox.com/). We looked at the high run time of (on my system) `BaseDomain.fun_6129`, which we conjectured to be `BaseDomain.equal_basecomponents_t`. 

Further investigation with `Stats.time` on a ~1 minute run of SQLite showed that a significant amount of time is used in `HoarePO.equal`: 

```
TOTAL                          55.185 s
  parse                           0.463 s
  convert to CIL                  0.568 s
  mergeCIL                        0.136 s
  analysis                       54.019 s
    createCFG                       0.836 s
      handle                          0.293 s
      iter_connect                    0.479 s
        computeSCCs                     0.217 s
    global_inits                    0.403 s
      HoareDomain.equal               0.001 s
    solving                        52.741 s
      S.Dom.equal                     0.199 s
      HoareDomain.equal              18.238 s
Timing used
```

<details>
<summary>Command line used for running Goblint on sqlite</summary>

```
/goblint ../sqlite-amalgamation-3370200/sqlite3.c ../sqlite-amalgamation-3370200/sqlite3.h ../sqlite-amalgamation-3370200/sqlite3ext.h ../sqlite-amalgamation-3370200/shell.c -v --set pre.cppflags[+] -DSQLITE_DEBUG --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages
```

</details>
We tried this a few times, and in the first minute of analyzing SQLite 30-50% of the runtime are spent in `HoarePO.equal`.

## Problems With Current Implementation

The current implementation of `HoarePO.equal` relies on checking `HoarePO.leq` in both directions. 
The implementation of `leq` in turn uses
- `for_all`, which transforms the argument Hoare-map to a list, 
- and `mem`, which also transforms its Hoare-map argument to a list. The function `mem` may be called for each element in a Hoare-map.

It thus seems like the current implementation could be optimized considerably. 

## Goals

[...]

PR #809, opened 2022-08-02
Optimize and refactor "Hoare" domains

sim642:
Related to issues #101 and #803.

### TODO
- [x] Benchmark on sv-benchmarks.
- [x] Check performance on SQLite.
- [x] Fix termination issue when `SetDomain.Joined` is used inside `AddressDomain`?
- [x] Reimplement `HoareDomain.MapBot` for witness generation?
- [x] Adapt to #808.
- [x] Rename `SensitiveDomain` to `BucketSetDomain` or something?


sim642:
## SQLite
Analyzing https://www.sqlite.org/2022/sqlite-amalgamation-3370200.zip for ~1 minute with
```console
goblint sqlite3.c sqlite3.h sqlite3ext.h shell.c -v --set pre.cppflags[+] -DSQLITE_DEBUG --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages
```
(Same version and arguments as in #803.)

### Before
At 2f5afc2c23fa12bbc1572f3fef43b5fa8cc0fa5e:
<details>

```
vars = 42983    evals = 216906  

Timings:
TOTAL                          63.343 s
  parse                           0.449 s
  convert to CIL                  0.586 s
  mergeCIL                        0.156 s
  analysis                       62.154 s
    createCFG                       0.964 s
      handle                          0.467 s
      iter_connect                    0.433 s
        computeSCCs                     0.217 s
    global_inits                    0.363 s
    solving                        60.793 s
      S.Dom.equal                     0.235 s
Timing used
Memory statistics: total=191160.42MB, max=534.03MB, minor=191085.51MB, major=1361.80MB, promoted=1286.89MB
    minor collections=91130  major collections=20 compactions=0
```
</details>

### After
At 6f886157414c99ec292998efe45cb7d5b7219ec5:
<details>

```
vars = 59925    evals = 299226  

Timings:
TOTAL                          64.871 s
  parse                           0.482 s
  convert to CIL                  0.574 s
  mergeCIL                        0.158 s
  analysis                       63.660 s
    createCFG                       0.983 s
      handle                          0.487 s
      iter_connect                    0.433 s
        computeSCCs                     0.217 s
    global_inits                    0.376 s
    solving                        62.257 s
      S.Dom.equal                     0.296 s
Timing used
Memory statistics: total=182390.45MB, max=614.13MB, minor=182315.17MB, major=1476.96MB, promoted=1401.67MB
    minor collections=86947  major collections=20 compactions=0
```
</details>

### Change
This is an improvement of 39.4% for vars and 38.0% for evals!

michael-schwarz:
Impressive! Could you check if a short-circuiting `equal` still leads to further improvements here? If yes, one probably wants both.

sim642:
> Impressive! Could you check if a short-circuiting `equal` still leads to further improvements here? If yes, one probably wants both.

Prepending `x == y || ` to `equal` of both `Projective` and `Pairwise` gives `vars = 58865    evals = 290541`, so looks like no clear difference.

michael-schwarz:
> Prepending `x == y ||` to `equal` of both `Projective` and `Pairwise` gives `vars = 58865 evals = 290541`, so looks like no clear difference.

It might be interesting to leave it running for a while longer, one would expect the shortcut to start paying off as soon as address sets become large.

sim642:
> It might be interesting to leave it running for a while longer, one would expect the shortcut to start paying off as soon as address sets become large.

It could also be the `AD.fold`/`AD.add` loops that end up copying and not using a physically equal instance.

sim642:
## sv-benchmarks
### Before vs after with `SetDomain.Joined` for `AddressDomain` buckets
image
([Full results](https://goblint.cs.ut.ee/results/75-all-fast-hoaredomain-after-joined-fix/results/results.2022-08-03_07-04-59.table.html#/scatter?filter=0(0*status*(status(in(unknown,true)))),1(0*status*(status(in(unknown,true))))&toolY=1&columnY=1&scaling=Linear&results=All&line=1.2&regression=Linear))

Overall it's a 6% speedup. Most of the noisy blobs that show some slowdown are ReachSafety-ECA tasks.

#### ldv
Filtering for `ldv` tasks, the speedup is 26% without any slowdown outliers:
image

### `SetDomain.Joined` vs `HoareDomain.Set2` for `AddressDomain` buckets
**TODO**: benchmarking running.

src/cdomains/valueDomain.ml
```diff
@@ -562,11 +562,12 @@ struct
...
     | (`Float x, `Float y) -> `Float (FD.widen x y)
+    (* TODO: symmetric widen, wtf? *)
     | (`Int x, `Address y)
     | (`Address y, `Int x) -> `Address (match ID.to_int x with
-        | Some x when BI.equal BI.zero x -> AD.widen AD.null_ptr y
-        | Some x -> AD.(widen y not_null)
-        | None -> AD.widen y AD.top_ptr)
+        | Some x when BI.equal BI.zero x -> AD.widen AD.null_ptr (AD.join AD.null_ptr y)
+        | Some x -> AD.(widen y (join y not_null))
+        | None -> AD.widen y (AD.join y AD.top_ptr))
```
michael-schwarz, line 570: Is this necessary here, or will not the solver already ensure that this join has happened?

src/cdomains/valueDomain.ml
```diff
@@ -562,11 +562,12 @@ struct
     | (x, `Bot) -> x
     | (`Int x, `Int y) -> (try `Int (ID.widen x y) with IntDomain.IncompatibleIKinds m -> Messages.warn ~category:Analyzer "%s" m; `Top)
     | (`Float x, `Float y) -> `Float (FD.widen x y)
+    (* TODO: symmetric widen, wtf? *)
     | (`Int x, `Address y)
     | (`Address y, `Int x) -> `Address (match ID.to_int x with
```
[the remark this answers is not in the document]
michael-schwarz: I think this actually occurs quite often, because we sometimes try very hard to maintain addresses, and sometimes the path meaning we treat a certain itn like an address may only become live in later iterations.

src/cdomains/valueDomain.ml
```diff
@@ -647,11 +648,12 @@ struct
...
-        | Some x when BI.equal x BI.zero -> AD.widen AD.null_ptr y
-        | Some x -> AD.(widen y not_null)
-        | None -> AD.widen y AD.top_ptr)
+        | Some x when BI.equal x BI.zero -> AD.widen AD.null_ptr (AD.join AD.null_ptr y)
+        | Some x -> AD.(widen y (join y not_null))
+        | None -> AD.widen y (AD.join y AD.top_ptr))
```
michael-schwarz, line 656: Same as above

src/domains/hoareDomain.ml
```diff
@@ -194,7 +193,7 @@ struct
   let narrow = product_bot (fun x y -> if B.leq y x then B.narrow x y else x)
 
   let add x a = if mem x a then a else add x a (* special mem! *)
-  let remove x a = unsupported "Set.remove"
+  let remove x a = filter (fun y -> not (B.leq y x)) a
```
michael-schwarz, line 196: I don't think we should expose this functionality as `remove` as it is very different from the normal `remove` on a set.

Consider e.g. a set from the chain lattice over {a,b,c}. The element {a,b} is represented by {b}, removing {b} now yields {} here, the normal set remove on the other hand would yield {a}, which is larger in the lattice order.
So this "remove" is not a sound abstraction of the "remove" of sets which one would expect.

sim642: This is necessary for `AddressDomain` to work though. `AD.remove` and `AD.remove` are used for some crucial logic, which is why I had to implement those for `SetDomain.Joined` as well (otherwise the reachability loop in base never terminates because it cannot remove anything from the workqueue set, or ever refine any values out of address sets).

`HoarePO` also did it in this way and that's where the behavior is from. I agree these removals can be problematic, so I'll add comments to them warning about this.

src/cdomains/valueDomain.ml
sim642: I think the issue stems from the fact that `widen` seems to be copied and modified from `join`, where the `Int`-`Address` (and vice versa) cases are handled symmetrically together. Widening is not commutative, so the misuse of `AD.widen` probably only happens from one of the cases.

I suppose the proper solution would be to remove the symmetry and handle both cases separately (and appropriately), in which case it might no longer be necessary, but I didn't want to make such delicate changes to these domains here.

src/cdomains/valueDomain.ml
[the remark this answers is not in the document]
sim642: These do occur, e.g. when the old value is `Int`, the new value is `Address` and their `join` gives `Address`, then `widen` is from the old `Int` to the joined `Address`.
The change at https://github.com/goblint/analyzer/pull/809#discussion_r936328669 also confirms that they happen.

tests/regression/00-sanity/33-hoare-over-paths.c
```diff
@@ -0,0 +1,21 @@
...
+  int r; // rand
+  if (r) {
+    r = 0;
+    assert(1); // extra node in cfg
+  }
+  else {
+    pthread_mutex_lock(&m);
+    assert(1); // extra node in cfg
+  }
+  // TODO: even though we're path-sensitive w.r.t mutexes, there's only one path after join
```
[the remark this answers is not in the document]
sim642: Right now, yes, the difference can only be seen by manual inspection. I'll think if it's somehow observable by something we can automatically test.

tests/regression/00-sanity/34-hoare-over-addrs.c
```diff
@@ -0,0 +1,20 @@
...
+  int *p;
+  if (r) {
+    p = &arr[1];
+    assert(1); // extra node in cfg
+  }
+  else {
+    p = &arr[2];
+    assert(1); // extra node in cfg
+  }
+  // TODO: even though we use Hoare for addresses per-varinfo
```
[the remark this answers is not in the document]
sim642: Right now it's also manual, but I think I can make this one automatic using an assertion on possible points to.

src/domains/hoareDomain.ml
michael-schwarz: To me it looks like `AD.remove` is only called with very specific values `null_ptr` and `unknown_ptr` for which this is fine. So I think we should maybe not expose this functionality everywhere, or at least name it something like `remove_all_smaller`.

tests/regression/00-sanity/33-hoare-over-paths.c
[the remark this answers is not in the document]
jerhard: If it cannot be automatically tested, adding some comment on what to check/look at in the HTML output may be useful.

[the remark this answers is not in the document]
sim642: I suspect with our current path-sensitive analyses the difference cannot be observed automatically like that, so I added a longer description to this test.
There's a theoretical and internal difference, which might be observable if some value analysis were partially path-sensitive.

src/domains/hoareDomain.ml
sim642: There's still the more general `diff`, which is also used with non-singleton pointer variants.

The annoying thing about having a separate name for this is that it requires a new module type for such sets and thus duplication of most of the set domains hierarchy we have. Only `SetDomain.Make` would be spared from this, because there `remove_all_smaller` would coincide with `remove`. And that's simply because `SetDomain.Make` (whose elements are `Printable.S`) is just a specific instance of set abstractions (whose elements are `Lattice.S`) where the element lattice is flat/fake, thus all equal and smaller non-bot elements to be removed are just the element itself.

Actually this sort of `remove` is the only meaningful one for set abstractions because in all other non-concrete abstractions you can never remove anything, since you cannot be sure (in general) that the elements of the set abstraction are singletons. That's only the case if the elements belong to a flat/fake lattice, which is implicitly the case for `SetDomain.Make` and in `AddressDomain` elements like `NullPtr` and `UnknownPtr`.

Declaring `remove` to mean this kind of _strong_ removal for all set abstractions wouldn't change anything for `SetDomain.Make`. For `SetDomain.ToppedSet` `diff` actually does a _strong_ removal (the first case):
https://github.com/goblint/analyzer/blob/13da7aca2966d9454a527a382e938dbbbe38e0d2/src/domains/setDomain.ml#L209-L213
If it were a _weak_ removal, then the first case would have to return `x` instead: since an unknown set `y` of elements might not cover all the elements of `x`.

src/cdomains/lval.ml
```diff
@@ -112,8 +112,7 @@ struct
   let rec leq x y =
     match x, y with
     | `NoOffset, `NoOffset -> true
-    | `NoOffset, x -> cmp_zero_offset x <> `MustNonzero
-    | x, `NoOffset -> cmp_zero_offset x = `MustZero
+    | `NoOffset, x -> true
```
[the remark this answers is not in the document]
sim642: This change was made in ff03d1c22661da516c453317de35e3d44076bceb in response to https://github.com/goblint/analyzer/issues/101#issuecomment-1200996946., but this might not be entirely correct indeed.
Since I changed `merge` to be independent of `cmp_zero_offset`, it also seems like `leq` should be. Maybe a more delicate fix is needed though.

tests/regression/00-sanity/34-hoare-over-addrs.c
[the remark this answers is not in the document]
sim642: I actually had to disable this new more precise behavior with Hoare sets in each bucket to everything force-joined in each bucket to stay compatible with #808. Since strings are not `leq` of each other, a Hoare set would keep them all, even with limiting enabled.

jerhard:
I ran Goblint with short-circuit evalution for the `SensitiveDomain.Projective.equal`  (branch `master_hoaredomain-new_short_circuit_equal`) and without it (branch `master_hoaredomain-new`). For the diff, see [here](https://github.com/goblint/analyzer/compare/master_hoaredomain-new...master_hoaredomain-new_short_circuit_equal).

The runtime with with the short-circuit evaluation was 42955.290s, and 43475.717s without it. That makes a difference of 520seconds = 8min40s. The version without short-circuit evaluation was 1.2% slower than the version with short-circuiting.

<details>
<summary>Goblint command line</summary>

```
time ./goblint_master_hoaredomain-new_short_circuit_equal ../../benchmarks/sqlite-amalgamation-3370200/sqlite3.c ../../benchmarks/sqlite-amalgamation-3370200/sqlite3.h ../../benchmarks/sqlite-amalgamation-3370200/sqlite3ext.h ../../benchmarks/sqlite-amalgamation-3370200/shell.c -v --set pre.cppflags[+] -DSQLITE_DEBUG --disable ana.base.context.non-ptr --disable ana.int.def_exc --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper

```

</details>

src/cdomains/lval.ml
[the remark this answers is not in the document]
sim642: I agree that this isn't right, because `a` (which is equivalent to `a[0]`) shouldn't be leq of `a[1]`. The modified `join` behavior is then wrong too, because `a` join `a[1]` shouldn't be `a[1]`, but `a[0..1]` or something.

Meanwhile `a` is also equivalent to `a.fst` (first field), which is fully incomparable with `a.snd` (second field). So I'm now completely unsure of what the lattice and partitioning of individual addresses (and their offsets) is (supposed to be).

Especially because `a.snd` at runtime actually corresponds to, e.g. `a[4]` based on the field widths and paddings. So should they be also considered equivalent in the lattice (and thus belong to the same partition)? If so, then all field offsets are also joinable: `a.fst` join `a.snd` is equivalent to `a[0]` join `a[4]` which could be `a[0..4]`. But right now we consider distinct fields totally separate.

This behavior would be inevitable if we simplify the partitions to be per `varinfo` (ignoring all the offsets), since the force-joined set of `a.fst` and `a.snd`, which currently is `{a.fst, a.snd}` would then become an useless `{a[0..4]}`.

sim642:
> The runtime with with the short-circuit evaluation was 42955.290s, and 43475.717s without it. That makes a difference of 520seconds = 8min40s. The version without short-circuit evaluation was 1.2% slower than the version with short-circuiting.

That's nice, I'll add it here then. The difference only becomes apparent at such long scale instead of a 1 minute benchmark.

sim642:
Also adapting `WitnessConstraints.PathSensitive3` to a similar construction from 8d6b8cff1 to cde436f39 gives a huge speedup in some cases (mostly those ECA tasks):
image
[Full results](https://goblint.cs.ut.ee/results/81-all-fast-hoaredomain-after-map-fp/results/results.2022-08-09_07-42-18.table.html#/scatter?filter=0(0*status*(status(in(true,unknown)))),1(0*status*(status(in(true,unknown))))&toolY=1&columnY=1&regression=Linear&scaling=Linear&results=All)

Somehow it also makes us unsound on 3 of our own tasks:
- [ ] goblint-regression/28-race_reach_06-cond_racing1
- [ ] goblint-regression/28-race_reach_23-sound_unlock_racing
- [ ] goblint-regression/28-race_reach_24-sound_lock_racing

sim642:
To avoid hard problems related to #816, it's possible to just use a partitioning which keeps `a`, `a[0]` and `a.fst` strictly separate. This implies a constant number of additional elements and work in the rare cases an address set indeed contains multiple of these.
[According to an sv-benchmarks run](https://goblint.cs.ut.ee/results/82-all-fast-hoaredomain-after-ad-proj/results/results.2022-08-11_12-42-09.table.html#/scatter?toolY=1&columnY=1&scaling=Linear&regression=Linear&results=All&filter=0(0*status*(category(notIn(error)))),1(0*status*(category(notIn(error))))&line=1.5), there's no impact to performance or results.
If anything, this makes it possible to implement `AddressDomain` just using the very effective `ProjectiveSet` and without any (nested) `PairwiseSet`, since the partitions have unique computable representatives (same address but with all indices replaced with top).

sim642:
I hunted down the reason for these new unsoundnesses in our own race_reach benchmarks. It was a very subtle misplacement of `try`.
The latest run on sv-benchmarks fixes those three, but still leaves one additional unsound benchmark (in addition to the handful already present on `master`) which I believe has an incorrect expected verdict: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1364.
In SV-COMP 2022 we answered `unknown` for it and on `master` we get the same unreachability but an `ERROR (verify)` hid our unsound claim.

src/domains/hoareDomain.ml
michael-schwarz: We should at least add a comment to this here, to indicate that it's usage is dangerous!

src/domains/mapDomain.ml
```diff
@@ -65,6 +65,59 @@ sig
...
+  type key
+  type value
+  val bindings: t -> (key * value) list
+  val iter: (key -> value -> unit) -> t -> unit
+end
+
+(** Reusable output definitions for maps. *)
+module Print (D: Groupable) (R: Printable.S) (M: Bindings with type key = D.t and type value = R.t) =
+struct
+  let show x = "mapping" (* TODO: WTF? *)
```
michael-schwarz, line 81: Address TODO?

src/domains/setDomain.ml
```diff
@@ -366,3 +384,43 @@ struct
...
+  let exists p e = p e
+  let for_all p e = p e
+  let mem e e' = E.leq e e'
+  let choose e = e
+  let elements e = [e]
+  let remove e e' =
+    if E.leq e' e then
+      E.bot ()
+    else
+      e'
```
michael-schwarz, line 405: We should also have comments warnings about these here

src/domains/setDomain.ml
```diff
@@ -366,3 +384,43 @@ struct
...
+  let add e e' = E.join e e'
+  let is_empty e = E.is_bot e
+  let union e e' = E.join e e'
+  let diff e e' = remove e' e
+  let iter f e = f e
+  let cardinal e =
+    if is_empty e then
+      0
+    else
+      1
```
michael-schwarz, line 418: This is also a bit dubious

src/cdomains/addressDomain.ml
```diff
@@ -21,16 +21,32 @@ end
...
+  module J = SetDomain.Joined (Addr)
+  (* module H = HoareDomain.SetEM (Addr) *)
+  (* Hoare set for bucket doesn't play well with StrPtr limiting:
+     https://github.com/goblint/analyzer/pull/808 *)
+  include DisjointDomain.ProjectiveSet (Addr) (J) (Addr.R)
 
-  module Addr = Lval.NormalLat (Idx)
-  include HoareDomain.HoarePO (Addr)
+  (* short-circuit with physical equality,
+     makes a different at long-scale: https://github.com/goblint/analyzer/pull/809#issuecomment-1206174751 *)
```
michael-schwarz, line 32:
```suggestion
     makes a difference at long-scale: https://github.com/goblint/analyzer/pull/809#issuecomment-1206174751 *)
```

src/cdomains/addressDomain.ml
```diff
@@ -21,16 +21,32 @@ end
...
     if M.tracing then M.traceu "ad" "-> %a\n" pretty r;
     r
+  let join x y =
+    if M.tracing then M.traceli "ad" "join %a %a\n" pretty x pretty y;
+    let r = join x y in
+    if M.tracing then M.traceu "ad" "-> %a\n" pretty r;
+    r
+  let leq x y =
+    if M.tracing then M.traceli "ad" "leq %a %a\n" pretty x pretty y;
+    let r = leq x y in
```
michael-schwarz, line 47: should `leq` not also short-circuit here?

src/cdomains/lval.ml
```diff
@@ -41,7 +41,7 @@ struct
     match x, y with
     | `NoOffset , `NoOffset -> true
     | `NoOffset, x
-    | x, `NoOffset -> cmp_zero_offset x = `MustZero (* cannot derive due to this special case *)
+    | x, `NoOffset -> cmp_zero_offset x = `MustZero (* cannot derive due to this special case, special cases not used for AddressDomain any more due to splitting *)
```
michael-schwarz, line 44: This comment confuses me.

michael-schwarz: If the special cases are no longer used, why still have them?

michael-schwarz (commented):
I added a few minor comments here and there, but overall I think this is a great step in the right direction. It looks like our various HoareDomains now have a sound footing in theory, so they'll hopefully cause no more issues (or if they do, it is much better understood than what we currently had)

michael-schwarz:
Does this PR also attempt to fix tests 02/{91, 92, 92} that are also closely related to the address domain?

sim642:
> Does this PR also attempt to fix tests 02/{91, 92, 92} that are also closely related to the address domain?

Not at all. On the contrary, it makes the problem even worse since previously `a` and `a[0]` were more-or-less handled equivalently but now they're split into separate buckets. Same with the `a` and `a.fst` pairing. So this makes #822 even more obvious and urgent.

src/domains/hoareDomain.ml
sim642: I documented this both in the `SetDomain` interface and the specific implementations.

src/domains/setDomain.ml
sim642: I documented this both in the `SetDomain` interface and the specific implementations.

src/domains/setDomain.ml
sim642: True, but this is at least internally consistent, e.g. `elements`/`iter`/etc all operate with this number of elements. `HoareDomain` has always been like this as well, i.e. such operations only account for the actually stored maximal elements, not all the possibly subsumed elements.

To make this behavior more explicit, I documented this both in the `SetDomain` interface and the specific implementations.

src/cdomains/addressDomain.ml
sim642: I guess it could, so I added that here.

src/domains/mapDomain.ml
sim642: For some odd reason `MapDomain.pretty` uses this `show` just as a prefix to distinguish itself from sets: `mapping {...}` vs `{...}`. So far this weird `show` hasn't been an issue, so I wouldn't bother refactoring it now since it'll all end up refactored for #213.

src/cdomains/lval.ml
sim642: Since I separated #816 from this PR, I wanted to minimize any other changes to the lvalue lattice as well because it needs a bigger refactoring/overhaul anyway.

sim642:
> On the contrary, it makes the problem even worse since previously `a` and `a[0]` were more-or-less handled equivalently but now they're split into separate buckets. Same with the `a` and `a.fst` pairing.

I just added the tests for these easier cases as well, which aren't listed in #822. Turns out that the first field one actually accidentally still works due to some of those special cases in lval domain still being present (and thus considering the two buckets, which are represented by themselves, equal). It doesn't work for the zero index since there the index gets turned into top for the bucket representative.

michael-schwarz (approved):

Issue #811, opened 2022-08-02
Replace ARINC with pthread extraction


michael-schwarz:
As discussed at the last Gobcon, we should 
- [x] Drop ARINC support for the next release, and
- [x] instead merge #220 to still have some extraction to Promela

michael-schwarz:
Completed via #220

PR #812, opened 2022-08-02
Remove ARINC analyses

michael-schwarz:
This removes the ARINC analyses from the current version of Goblint, they will remain available in the last release.

References #811


sim642:
On top of my head, I remember there are these (de)obfuscations related to ARINC as well: https://github.com/goblint/analyzer/issues/460#issuecomment-981609754 and https://github.com/goblint/analyzer/issues/460#issuecomment-982617718.

michael-schwarz:
The second one was already removed before, I also removed the first now.

sim642:
> The second one was already removed before, I also removed the first now.

Unless I'm mistaken, it's still there on the current master: https://github.com/goblint/analyzer/blob/26fd6ced7a37bb6af7e5fa921bbbf144461f06bd/src/util/goblintutil.ml#L121-L135

michael-schwarz:
Before your comment is what I meant by before :wink:

sim642:
Ah, sorry, right.

The test fails on MacOS. I guess it's because `memcpy` might be a wrapper/macro for some other internal (fortified?) function there. The solution probably is to handle some extra name variant of it (not sure what exactly).
It's odd though because #391 also contains a similar change, but doesn't show a similar failure.

michael-schwarz:
Yes, it looks like it should be `__builtin___memcpy_chk`. However, for some reason than the assert succeeds instead of definitely failing, so there still is something fishy going on.

sim642:
> However, for some reason than the assert succeeds instead of definitely failing, so there still is something fishy going on.

It was actually `unknown` instead of `fail`, which makes sense, because it fell back to the unknown function handling via `LibraryFunctions`, which handles `__builtin___memcpy_chk` by completely invalidating the destination. The result of that could also be 0 then.

> It is broken if you pull out the `AddrOf` to somewhere else as it does not in fact dereference its argument upside_down_face.

That's what #729 also fixed but it's still just on `interactive`.

sim642 (commented):
There's also `src/extract/pml_arinc.ml`.

sim642 (approved):

PR #816, opened 2022-08-05
Try to fix lval lattice

sim642:
This is an attempt to address https://github.com/goblint/analyzer/issues/101#issuecomment-1200996946 and https://github.com/goblint/analyzer/pull/809#discussion_r935693299 by adding more lval lattice tests for _some_ possibly reasonable structure and getting them to pass.

I've opened this separate from #809 because with the growing complexity and nuance of the domain, I'm no longer sure what the intended behavior should be. Right now I have added a bunch of extra cases to treat `a`, `a[0]` and `a.fst` (first field) equivalently.

One can imagine increasingly complex lvals that this doesn't treat correctly. For example, `a.snd` (second field) and `a[0].snd` are the same memory location, but we don't currently treat them as equivalent, because the above equivalence is only ever applied at the end of the offset, not at every intermediate offset. I suspect the only way to deal with that would be to compute some edit distance between the offsets (i.e. where `[0]` and `.fst` can be inserted or removed to get from one to the other), but that would be insane. I guess this is why some people use automata to describe access paths...

### DIFF
--- a/src/cdomains/lval.ml
+++ b/src/cdomains/lval.ml
@@ -27,6 +27,8 @@ struct
     | [] -> false
     | f :: _ -> CilType.Fieldinfo.equal f x
 
+  let is_zero_index x = Idx.equal_to (IntOps.BigIntOps.zero) x
+
   let rec cmp_zero_offset : t -> [`MustZero | `MustNonzero | `MayZero] = function
     | `NoOffset -> `MustZero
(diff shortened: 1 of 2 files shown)


michael-schwarz:
I would propose the following procedure here for now: We merge #809, #827, and this as they are obviously a step in the right direction and use #822 to think about a more general solution that is sound also for more involved offsets. Wdyt?

sim642:
Indeed, with the splitting approach in #809, this fix attempt becomes unnecessary.

PR #817, opened 2022-08-05
Add server command to merge config by file name

karoliineh:
Related to goblint/GobPie#32

> Currently on Goblint conf file modification, GobPie kills the existing Goblint server and starts a new one:
> ...
> This is quite undesirable since the incremental state of the server is lost. Instead GobPie should ask the Goblint server to just reload the conf file in the same process. A new Goblint server command needs to be added to merge a config by file name, instead of directly passing a JSON object to the server.


sim642:
The config merging has a caveat that only becomes relevant in server mode. For example, consider the following:
1. Initially, by default, `ana.int.interval` is disabled.
2. The initial config from the `--conf` argument doesn't specify a value for `ana.int.interval`, so that's still disabled.
3. Then you change the config file to now enable `ana.int.interval` and load it, so it's now enabled.
4. But then you undo your change to the config file (remove `ana.int.interval` from it) and load it again. Even though you undid your change to the config, `ana.int.interval` still remains enabled!

This is because merging a config, that doesn't specify the option, doesn't change or reset it either (hence the _merge_). It could be surprising/confusing that the config you get isn't determined by the config files you have right now, but also the entire history of them.

Having another command for resetting options to their defaults is trivial, but that also means that `server.enabled` and `server.reparse` (or any other options initially specified on the command line) get disabled in the middle of server mode. So the reasonable consistent behavior is not that simple to achieve.

src/maingoblint.ml
```diff
@@ -539,6 +539,19 @@ let diff_and_rename current_file =
...
+  parse_arguments ();
+  check_arguments ();
+  AfterConfig.run ();
+
+  Sys.set_signal (Goblintutil.signal_of_string (get_string "dbg.solver-signal")) Signal_ignore; (* Ignore solver-signal before solving (e.g. MyCFG), otherwise exceptions self-signal the default, which crashes instead of printing backtrace. *)
+
+  if ops_only then Cilfacade.init_options () else Cilfacade.init ();
+
+  handle_extraspecials ();
+  GoblintDir.init ();
```
sim642, line 552:
```suggestion
```

src/util/server.ml
```diff
@@ -244,6 +244,20 @@ let () =
...
   end);
 
+  register (module struct
+    let name = "read_config"
+    type params = { fname: string } [@@deriving of_yojson]
+    type response = unit [@@deriving to_yojson]
+    let process { fname } _ =
+      try
+        GobConfig.json_conf := Options.defaults;
+        Arg.current := 0;
```
sim642, line 254: Better to move this into `parse_arguments` with a comment that it's necessary to reset in server mode.

src/goblint.ml
```diff
@@ -9,17 +9,11 @@ open Printf
...
-    parse_arguments ();
-    check_arguments ();
-    AfterConfig.run ();
-
-    Sys.set_signal (Goblintutil.signal_of_string (get_string "dbg.solver-signal")) Signal_ignore; (* Ignore solver-signal before solving (e.g. MyCFG), otherwise exceptions self-signal the default, which crashes instead of printing backtrace. *)
-
-    Cilfacade.init ();
-
+    Maingoblint.parse_arguments ();
+    Maingoblint.handle_options false;
```
sim642, line 13: `handle options` might be better inside `parse_arguments`. Right now server's `reset_config` forgets to handle after parsing.

PR #819, opened 2022-08-10
Limit YAML witness invariants to to-be-read and just-written variables

sim642:
As discussed in GobCon, this would be something to somewhat limit the large number of invariants we generate.
Since accesses are to lvalues, not just variables, this also has lvalue granularity, because accessing a field in a large struct will the only emit one invariant about that field instead of all the fields of the struct.

### Changes
1. Add access event set to access analysis local state and `MayAccessed` query to use it in YAML witness generation.
2. Split access analysis by extracting race analysis from it. Now access analysis is only concerned with collecting all accessed lvalues from all expressions in all transfer functions, while race analysis uses that to construct its race tables in the global invariant.
3. Generalize `Access` event and `Invariant` query from optional `varinfo` to lvalue set. ~~Still uses a separate `Invariant.context1` within the base analysis because the existing logic is per-lvalue.~~ Simplify base analysis invariant logic to use a single lvalue instead of invariant context.
4. Fix `Invariant` query offset handling in base analysis.
5. Add `query` tracing.


src/analyses/accessAnalysis.ml
```diff
@@ -1,156 +1,43 @@
...
-(** Access and data race analyzer without base --- this is the new standard *)
+(** Access analyzer without base --- this is the new standard *)
 module Spec =
 struct
   include Analyses.DefaultSpec
 
   let name () = "access"
 
-  module D = Lattice.Unit
+  module D = AccessDomain.EventSet
```
michael-schwarz, line 16: Are we worried about this set becoming large? If not what prevents it from growing for large programs?
Naively, it would seem `leq` checks for this EventSet would be quadratic in it's size?

sim642: It's actually not very accumulating because `sync` empties it between transfer functions, such that the local sets would only have the set of accesses of just the last transfer function. I'll add a few comments to point out this unusual behavior.

Also, we already put these accesses into global access/race invariants, so we already have the same amount around, just stored in a different place.

Ideally it would suffice to collect these only during postsolving, but local constraint variable values cannot be modified during postsolving (unlike write-only globals), so they'd only emit fixpoint not reached errors and not persist. That's the same reason why the GraphML witness data collection happens all the time as well, not just during postsolving.

src/domains/invariant.ml
```diff
@@ -40,12 +40,17 @@ let ( || ) = join
...
+type context1 = {
   path: int option;
   lval: lval option;
 }
 
+type context = {
+  path: int option;
+  lvals: CilLval.Set.t;
+}
+
```
michael-schwarz, line 52: Maybe we can have better names here, given that `context` is already over-used as is? Maybe `context_1lval` and `context_lvalset` or something of the sort?

sim642: Actually, I'll have a look if I can get rid of `context1` altogether: if it's now just used by base analysis internally, then it doesn't need the `path` field at all, but could hopefully just pass the lval around directly.

src/analyses/tutorials/unitAnalysis.ml
```diff
@@ -3,7 +3,9 @@
 open Prelude.Ana
 open Analyses
 
-module Spec : Analyses.MCPSpec with module D = Lattice.Unit and module C = Lattice.Unit =
+(* module Spec : Analyses.MCPSpec with module D = Lattice.Unit and module C = Lattice.Unit and type marshal = unit = *)
+(* No signature so others can override module G *)
+module Spec =
```
michael-schwarz, line 8: Honestly, I think here duplication is less bad then not having a signature here, because that tends to be very helpful for newcomers such that they get specific error messages for some functions instead of a general "Module does not match" message that is complicated to read.

At least that was my impression when teaching 10 people how to use Goblint at the ConVEY retreat in Venice.

michael-schwarz (commented):
As soon as the comment about the `unitAnalysis` is addressed, I think this is good to merge!

src/analyses/tutorials/unitAnalysis.ml
sim642: But before this PR, on master there was the signature at the time. Also, having the signature here doesn't really make the error message simpler: if you've mistyped a transfer function, then the entire module still gets underlined, not just that function, and the command-line compiler error is literally the same.

michael-schwarz: Ok, if the error that is reported is the same without the signature, I guess it doesn't make a difference. I seem to recall people getting more local error messages with the signature as opposed to without it, but maybe I am mistaken.

sim642: I think the localized errors you get would be due to having type annotations on all these transfer function arguments, as long as they are right. Then if you mistype something within the function, it's localized to just that function, not the entire module, because it's typechecked more locally.

In that regard, these transfer functions are good, but in various analyses we don't have all the arguments annotated like that.

sim642:
This will conflict with #833 and I think it'll be easier to have that merged first.

src/analyses/accessAnalysis.ml
sim642: I did an SV-COMP run with this and sadly there's ~15% average slowdown from these changes. I think it might be a bit about the sets, but more because now access events are emitted also in singlethreaded mode.

sim642: I implemented some optimizations now:
1. Per-node accesses are only collected if they're enabled for _YAML_ witness generation (which we currently don't use for SV-COMP). Otherwise the old behavior remains: access events are only emitted in multithreaded mode.
2. Instead of collecting accesses into local states all the time, they're now collected only during postsolving into new write-only global constraint variables. This means they are not available during the analysis itself, but that wasn't required to begin with.

With these, performance on sv-benchmarks is on par with what it was before this PR.

michael-schwarz (approved):

PR #820, opened 2022-08-10
Remove phases

sim642:
Issue #460.


michael-schwarz (approved):

sim642:
> `tests/sv-comp/observer/path_nofun_true-unreach-call.c` still has some references to phases.

Good catch! It didn't even occur to me to search tests because I didn't think it would've ever been used, but turns out that I did for my observer automaton stuff, only to realize how limited it is.

PR #828, opened 2022-08-24
Remove `ID.is_bool` and `ID.is_int`

sim642:
This fixes an assert failing in `sv-benchmarks/c/recursive/MultCommutative-2` here:
https://github.com/goblint/analyzer/blob/a4e917f86c983d7c97138364b07e031a7715e3d5/src/analyses/base.ml#L2216-L2220
The problem was that `ID.is_bool value = true` but `ID.to_bool = None`, which of course shouldn't happen.

The offending `value` is `(0,[1,1])`, which is the result of a sequence of imprecise refinements (documented in TODOs). These don't happen with `--set ana.int.refinement once`, but these imprecisions are nontrivial to avoid (e.g. intervals cannot represent exclusion sets).

The easier way to avoid the inconsistency between `is_bool` and `to_bool` is not to have `is_bool` because `is_bool x` is equivalent to `to_bool x <> None`. Most places where we use `is_bool` also end up calling `to_bool`, like above, which doubles the work. Same goes for `is_int`.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -163,6 +163,11 @@ struct
     | Le -> ID.le
     | Ge -> ID.ge
     | Eq -> ID.eq
+    (* TODO: This causes inconsistent results:
+       def_exc and interval definitely in conflict:
+         evalint: base eval_rv m -> (Not {0, 1}([-31,31]),[1,1])
+         evalint: base eval_rv 1 -> (1,[1,1])
+         evalint: base query_evalint m == 1 -> (0,[1,1]) *)
     | Ne -> ID.ne
     | BAnd -> ID.bitand
     | BOr -> ID.bitor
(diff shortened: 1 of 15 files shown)


sim642:
I did an sv-benchmarks run with this for good measure. That single benchmark went from `ASSERTION` to `TIMEOUT`, but at least there's no regression to results or performance.

src/analyses/base.ml
```diff
@@ -163,6 +163,11 @@ struct
     | Le -> ID.le
     | Ge -> ID.ge
     | Eq -> ID.eq
+    (* TODO: This causes inconsistent results:
+       def_exc and interval definitely in conflict:
+         evalint: base eval_rv m -> (Not {0, 1}([-31,31]),[1,1])
```
michael-schwarz, line 168: I'm not sure I understand this. `(Not {0, 1}([-31,31]),[1,1])` already is an invalid value in the first place!

michael-schwarz (approved):

src/analyses/base.ml
sim642: Yes, but `IntDomTuple` doesn't know that if int domain refinement is not applied: it's an unreduced product, so we don't consider such code dead immediately but possibly only at some later operation. Tracing the SV-COMP program shows how such inconsistent values in the TODOs come about during actual analysis. I think the first inconsistent value there arises from intervals not being able to express the exclusion set of 0 (which comes from a non-equality with 0) for refinement, so it's a top interval instead, which doesn't refine the interval but def_exc can be refined.

PR #831, opened 2022-09-05
Generalize protection logging from apron

sim642:
The protection logging from apron analysis can be generalized into mutex analysis to compute, in traces paper notation, $G_m$-s from $\mathcal{M}[g]$-s. So far this information was just used for some benchmarking stats, but it's also useful for other things, e.g. witness invariants only about globals which are protected.

While moving this code, there are opportunities to also make other changes, to be discussed below.

### DIFF
--- a/src/analyses/mutexAnalysis.ml
+++ b/src/analyses/mutexAnalysis.ml
@@ -93,12 +93,12 @@ struct
 
   module GM = Hashtbl.Make (ValueDomain.Addr)
 
-  let max_cluster = ref 0
+  let max_protected = ref 0
   let num_mutexes = ref 0
   let sum_protected = ref 0
 
   let init _ =
-    max_cluster := 0;
+    max_protected := 0;
     num_mutexes := 0;
     sum_protected := 0
 
@@ -175,7 +175,7 @@ struct
           if GobConfig.get_bool "dbg.print_protection" then (
             let (protected, _) = G.protected (ctx.global g) in (* readwrite protected *)
             let s = VarSet.cardinal protected in
-            max_cluster := max !max_cluster s;
+            max_protected := max !max_protected s;
             sum_protected := !sum_protected + s;
             incr num_mutexes;
             M.info_noloc ~category:Race "Mutex %a read-write protects %d variable(s): %a" ValueDomain.Addr.pretty m s VarSet.pretty protected
@@ -243,7 +243,7 @@ struct
     if GobConfig.get_bool "dbg.print_protection" then (
       M.msg_group Info ~category:Race "Mutex read-write protection summary" [
         (Pretty.dprintf "Number of mutexes: %d" !num_mutexes, None);
-        (Pretty.dprintf "Max number variables of protected by a mutex: %d" !max_cluster, None);
+        (Pretty.dprintf "Max number variables of protected by a mutex: %d" !max_protected, None);
         (Pretty.dprintf "Total number of protected variables (including duplicates): %d" !sum_protected, None);
       ]
     )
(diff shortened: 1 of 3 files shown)


src/analyses/mutexAnalysis.ml
```diff
@@ -9,11 +9,13 @@ open Prelude.Ana
...
 
 
+(* TODO: remove SpecParam because only one implementation? *)
 module type SpecParam =
 sig
   module G: Lattice.S
   val effect_fun: ?write:bool -> Lockset.t -> G.t
   val check_fun: ?write:bool -> Lockset.t -> G.t
+  val export: ?write:bool -> G.t -> Mutexes.t
 end
```
sim642, line 19: Mutex analysis is a functor parametrized about such global domain, of which there is only a single implementation. Maybe it'd be clearer to just inline this instead of adding various conversion functions for everything?
It was functorized because OSEK used a different implementation, but that's now gone.

michael-schwarz: I agree, we should do so!

src/analyses/mutexAnalysis.ml
```diff
@@ -47,6 +49,15 @@ struct
   module G = P.G
   module V = VarinfoV
 
+  module GM = Hashtbl.Make (ValueDomain.Addr)
+  module VarSet = SetDomain.Make (Basetype.Variables)
+  let gm_rw = GM.create 10 (* TODO: marshal? move into global invariant? *)
+  let gm_w = GM.create 10 (* TODO: marshal? move into global invariant? *)
```
sim642, line 55: Currently these hashtables are inside the `Spec`, but maybe they should be moved into the global invariant instead? Right now, this information isn't correctly handled during incremental analysis.

src/analyses/mutexAnalysis.ml
```diff
@@ -138,12 +149,39 @@ struct
...
     | _ ->
       event ctx e octx (* delegate to must lockset analysis *)
+
+  let finalize () =
+    if GobConfig.get_bool "dbg.print_protection" then (
+      let max_cluster = ref 0 in
+      let num_mutexes = ref 0 in
+      let sum_protected = ref 0 in
+      Printf.printf "\n\nProtecting mutexes:\n";
+      (* TODO: what about rw? *)
```
sim642, line 174: This logging output is currently about just write-protection, which is what apron privatizations cared about, but maybe it should be about read-write-protection instead?
Actually, in apron this ignored the write flag, so was a mish-mash of both anyway.

src/analyses/mutexAnalysis.ml
sim642: Done!

There's also the opportunity to refactor the global invariant from (readwrite, write) to (read, write), which is simply more general, or even a map on all `AccessKind`s, but that can wait since it's not really necessary or useful right now.

src/analyses/mutexAnalysis.ml
michael-schwarz: Makes sense!

src/analyses/mutexAnalysis.ml
```diff
@@ -43,9 +86,21 @@ struct
...
-  module G = P.G
-  module V = VarinfoV
+  module V = Arg.V
+  module GProtecting = Arg.GProtecting
+  module GProtected = Arg.GProtected
+  module G = Arg.G
+
+  module GM = Hashtbl.Make (ValueDomain.Addr)
+
+  let max_cluster = ref 0
```
michael-schwarz, line 96: Maybe it makes sense to rename this to `max_protected` ?

michael-schwarz (approved):

PR #833, opened 2022-09-12
Privatized location invariants in YAML witnesses

sim642:
This builds on both #830 and #831.

The idea is that each privatization (both base and apron) can list its privatized globals at a location and witness invariants can be emitted using sound values for them (via `read_global` from privatization).
Previously, invariants from globals were dumped from the local state but depending on the privatization that might not be sound.


src/analyses/apron/apronPriv.apron.ml
```diff
@@ -404,6 +403,7 @@ struct
     {apr = getg (); priv = startstate ()}
 
   let iter_sys_vars getg vq vf = () (* TODO: or report singleton global for any Global query? *)
+  let invariant_vars ask getg st = protected_vars ask (* TODO: is this right? *)
```
michael-schwarz, line 406: I am not very confident here either, but I don't think we ended up using this privatization much anyway...
And if we wanted to, we would need to carefully re-check it either way.

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -1071,6 +1073,8 @@ struct
     | VarQuery.Global g -> vf (V.global g)
     | _ -> ()
 
+  let invariant_vars ask getg st = protected_vars ask
+
```
michael-schwarz, line 1077: Same argument here! Any variant of mutex-meet should never have invalid information in the local state!

src/analyses/basePriv.mli
```diff
@@ -31,6 +31,9 @@ sig
 
       Should account for all unprotected/weak values of global variables. *)
 
+  val invariant_vars: Queries.ask -> (V.t -> G.t) -> BaseDomain.BaseComponents (D).t -> varinfo list
```
michael-schwarz, line 34: Could once again be `[`] for all mutex-meet variants!

src/analyses/apron/apronPriv.apron.ml
```diff
@@ -579,6 +579,8 @@ struct
   let threadenter ask getg (st: apron_components_t): apron_components_t =
     {apr = AD.bot (); priv = startstate ()}
 
+  let invariant_vars ask getg st = protected_vars ask
+
```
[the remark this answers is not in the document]
sim642: Regarding this and the following comments: this still needs to be done, because if it were `[]`, then no globals would ever be in invariants. This is for all globals to include in invariants, not extra globals to include/exclude in addition to the ones in the local state, because it could be both ways depending on the privatization: the local state may contain more or less globals.

src/analyses/apron/apronPriv.apron.ml
sim642: Given that it is named protection-based, I guess there's not many other possibilities. Anyway, there's no _wrong_ implementations really: it just means either too few or too many invariants are emitted, but they're all emitted soundly via `read_global`.

src/analyses/apron/apronPriv.apron.ml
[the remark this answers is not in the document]
sim642: I now implemented a slight variation for mutex-meets, which gets the list of protected variables directly from the local state, rather than with all the queries.

michael-schwarz (approved):

PR #836, opened 2022-09-19
Group globals for comparison and updating of ids

### LINKED ISSUE #627: Function forward-declarations mishandled in incremental compare and update
See https://github.com/goblint/analyzer/pull/609#issuecomment-1057333884 and the following comments.

Special care should be taken of `GVarDecl`s that also have corresponding `GFun`s. The CIL invariant is that if they refer to the same function, they should share a single physical instance of `varinfo`.
Currently incremental compare and update treats them separately and may update the ID twice, while also breaking the sharing. I suppose we'll have to group all the `GVarDecl`s for a `GFun` together and do a single update to the `varinfo`, which is shared by all of them.

stilscher:
This PR a fix for the way globals are handled in the version comparison and when updating the ids. Globals are now identified only by their name. The associated declaration and at most one definition of a global are grouped together, when collecting them for comparison. In case a definition exists, it is sufficient to consider the definition only, since declarations share the same `varinfo` with the defnition.
We checked in `cabs2cil` that also only at most one declaration is created for a global, even though the cil documentation does not state this clearly. Therefore, when grouping entities belonging to a global, only a single declaration can be collected. In case multiple are found in the CIL file, an error is thrown. I updated the cil documentation about `GVarDecl`s in goblint/cil@729a1ff3f300330043af118141229079e0fad499 for future reference.

This PR also fixes the incremental tests 03/02 and 03/03 which previously failed when the cfg comparison was enabled.

Closes #627


sim642:
> We checked in `cabs2cil` that also only at most one declaration is created for a global, even though the cil documentation does not state this clearly. Therefore, when grouping entities belonging to a global, only a single declaration can be collected. In case multiple are found in the CIL file, an error is thrown.

Is that also preserved by `Mergecil`? Because even if `Cabs2cil` on each file does produce at most one declaration (in addition to a definition), then maybe `Mergecil` leaves all the declarations?

michael-schwarz:
We also checked that, and Mergecil preserves that invariant.

src/incremental/compareCIL.ml
```diff
@@ -4,20 +4,32 @@ open CilMaps
 include CompareAST
 include CompareCFG
 
+(* global_type and global_t are implicitly used by GlobalMap to keep GVarDecl apart from GVar and GFun, so do not remove! *)
```
sim642, line 7: I guess this comment is no longer necessary?

src/incremental/updateCil.ml
```diff
@@ -89,8 +90,9 @@ let update_ids (old_file: file) (ids: max_ids) (new_file: file) (changes: change
       List.iter (assign_same_id f.sallstmts) d.unchangedNodes
   in
   let reset_changed_globals (changed: changed_global) =
-    match (changed.current, changed.old) with
-    | GFun (nw, _), GFun (old, _) -> reset_changed_fun nw old changed.unchangedHeader changed.diff
+    match (changed.current.def, changed.old.def) with
+    | Some (Fun nw), Some (Fun old) -> reset_changed_fun nw old changed.unchangedHeader changed.diff
+    (* why ids of global variables not adapted? *)
```
sim642, line 95: We should probably figure this out here. Also regarding just declarations.

michael-schwarz (approved):
As soon as Simmo's comments are addressed, I think we can merge this.

src/incremental/compareCIL.ml
```diff
@@ -4,20 +4,31 @@ open CilMaps
 include CompareAST
 include CompareCFG
 
+module GlobalMap = Map.Make(String)
+
+type global_def = Var of varinfo | Fun of fundec
+and global_col = {decls: varinfo option; def: global_def option}
```
jerhard, line 10:
```suggestion
type global_col = {decls: varinfo option; def: global_def option}
```
The `and` is usually only used for mutually recursive type definitions.

src/incremental/updateCil.ml
stilscher: I discussed with @jerhard  that it is wrong to not change the id of changed globals since then the uniqueness of the ids is not necessarily preserved. We think a change to the global variable will often lead to a lot of re-evaluations anyway and adapting the old ids could lead to a lot of precision loss. Therefore, we decided to generate new ids for changed globals.

jerhard:
The question thus is whether 
- when `merge_inlines` is `false`, there can be multiple inline function definitions with the same name?
- when `merge_inlines` is `true`, there can be multiple inline function definitions with the same name, but different bodies?

sim642:
From what I recall, duplicate inline functions, which are not merged, just get renamed with suffixes. At least I remember seeing large numbers of copies of inline functions in the HTML view of such programs.

michael-schwarz:
> From what I recall, duplicate inline functions, which are not merged, just get renamed with suffixes. At least I remember seeing large numbers of copies of inline functions in the HTML view of such programs.

Yes, I think that is the behavior when `merge_inlines` is `false`.

michael-schwarz:
What is the status here? It would be nice to get this so we can merge #609 which will hopefully give a nice performance boost (that the students from our practical starting on Thursday could measure for instance.

stilscher:
I created a small example to test the open questions that @jerhard posted above. It seems that the merger is not behaving correctly. If `merge_inline` is disabled a function definition and an inline function definition with the same name but in a different translation unit do not get merged but are not renamed either. This is also the case when `merge_inline` is enabled but the function bodies are not the same. I will take a look at the merger and try to fix it, in a way that multiple function definitions either get a unique name or are merged.

src/incremental/updateCil.ml
jerhard: >Also, we were thinking that we should maybe generate new ids for functions with changed headers also

That should be possible. Did you come to a  conclusion?

jerhard:
>If merge_inline is disabled a function definition and an inline function definition with the same name but in a different translation unit do not get merged but are not renamed either. This is also the case when merge_inline is enabled but the function bodies are not the same.

Does this already lead to problems with the current `master` branch? Do the changes in this branch cause additional problematic cases?

src/incremental/updateCil.ml
stilscher: There are a couple of incremental tests failing after implementing this. One reason seems to be that functions included in `exclude_from_rel_destab` (as for example force-reanalyze functions) are destabilized at their entry node to force reevaluation. If the id of these functions (which are assumed to be completely changed and not have an unchanged header) changes, the destabilization can not go through all the old influences of this entry node and therefore does not destabilize enough to actually trigger reevaluation.

src/incremental/updateCil.ml
```diff
@@ -88,9 +89,16 @@ let update_ids (old_file: file) (ids: max_ids) (new_file: file) (changes: change
...
   in
+  let update_var (v: varinfo) =
+    v.vid <- make_vid ()
+  in
   let reset_changed_globals (changed: changed_global) =
-    match (changed.current, changed.old) with
-    | GFun (nw, _), GFun (old, _) -> reset_changed_fun nw old changed.unchangedHeader changed.diff
+    match (changed.current.def, changed.old.def) with
+    | Some (Fun nw), Some (Fun old) -> reset_changed_fun nw old changed.unchangedHeader changed.diff
+    | Some (Var nw), Some (Var old) -> update_var nw
```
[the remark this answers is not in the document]
jerhard: @stilscher, @michael-schwarz  and I discussed this just now, and it turns out one would rather not call `update_var` in that case. Actually, we expect the programs to not use variables and functions that were only declared, but not defined. 

For the case of functions, we came to the conclusion that when a function previously only had a declaration and was called, and then a definition is inserted, the call sites of this now newly declared function would not be destabilized. This is because we do not record dependencies to undefined functions. 
However, as such input programs that use undefined functions and variables are malformed on a level that a compiler would reject, we decide to not introduce unnecessary handling for these cases that would clutter the implementation.

jerhard (approved):

PR #837, opened 2022-09-21
Clean up some special functions

sim642:
This replaces many name-based `special` function handling with proper `LibraryFunctions`.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2506,49 +2506,33 @@ struct
       let dest_a, dest_typ = addr_type_of_exp dest in
       let value = VD.zero_init_value dest_typ in
       set ~ctx (Analyses.ask_of_ctx ctx) gs st dest_a dest_typ value
-    | Unknown, "strcpy"
-    | Unknown, "strncpy"
-    | Unknown, "memcpy"
-    | Unknown, "__builtin___memcpy_chk" ->
-      begin match f.vname, args with
-        | _, [dst; src]
-        | _, [dst; src; _]
-        | "__builtin___memcpy_chk", [dst; src; _; _] ->
-          (* invalidating from interactive *)
-          (* let dest_a, dest_typ = addr_type_of_exp dst in
-             let value = VD.top_value dest_typ in
-             set ~ctx (Analyses.ask_of_ctx ctx) gs st dest_a dest_typ value *)
-          (* TODO: reuse addr_type_of_exp for master *)
-          (* assigning from master *)
-          let get_type lval =
-            let address = eval_lv (Analyses.ask_of_ctx ctx) gs st lval in
-            AD.get_type address
-          in
-          let dst_lval = mkMem ~addr:(Cil.stripCasts dst) ~off:NoOffset in
-          let src_lval = mkMem ~addr:(Cil.stripCasts src) ~off:NoOffset in
+    | Memcpy { dest = dst; src }, _ ->
+      (* invalidating from interactive *)
+      (* let dest_a, dest_typ = addr_type_of_exp dst in
+          let value = VD.top_value dest_typ in
+          set ~ctx (Analyses.ask_of_ctx ctx) gs st dest_a dest_typ value *)
+      (* TODO: reuse addr_type_of_exp for master *)
+      (* assigning from master *)
+      let get_type lval =
+        let address = eval_lv (Analyses.ask_of_ctx ctx) gs st lval in
+        AD.get_type address
+      in
+      let dst_lval = mkMem ~addr:(Cil.stripCasts dst) ~off:NoOffset in
+      let src_lval = mkMem ~addr:(Cil.stripCasts src) ~off:NoOffset in
 
-          let dest_typ = get_type dst_lval in
-          let src_typ = get_type src_lval in
+      let dest_typ = get_type dst_lval in
+      let src_typ = get_type src_lval in
 
-          (* When src and destination type coincide, take value from the source, otherwise use top *)
-          let value = if typeSig dest_typ = typeSig src_typ then
-              let src_cast_lval = mkMem ~addr:(Cilfacade.mkCast ~e:src ~newt:(TPtr (dest_typ, []))) ~off:NoOffset in
-              eval_rv (Analyses.ask_of_ctx ctx) gs st (Lval src_cast_lval)
-            else
-              VD.top_value (unrollType dest_typ)
-          in
-          let dest_a = eval_lv (Analyses.ask_of_ctx ctx) gs st dst_lval in
-          set ~ctx (Analyses.ask_of_ctx ctx) gs st dest_a dest_typ value
-        | _ -> failwith "strcpy arguments are strange/complicated."
-      end
-    | Unknown, "__builtin" ->
-      begin match args with
-        | Const (CStr ("invariant",_)) :: ((_ :: _) as args) ->
-          List.fold_left (fun d e -> invariant ctx (Analyses.ask_of_ctx ctx) ctx.global d e true) ctx.local args
-        | _ -> failwith "Unknown __builtin."
-      end
+      (* When src and destination type coincide, take value from the source, otherwise use top *)
+      let value = if typeSig dest_typ = typeSig src_typ then
+          let src_cast_lval = mkMem ~addr:(Cilfacade.mkCast ~e:src ~newt:(TPtr (dest_typ, []))) ~off:NoOffset in
+          eval_rv (Analyses.ask_of_ctx ctx) gs st (Lval src_cast_lval)
+        else
+          VD.top_value (unrollType dest_typ)
+      in
+      let dest_a = eval_lv (Analyses.ask_of_ctx ctx) gs st dst_lval in
+      set ~ctx (Analyses.ask_of_ctx ctx) gs st dest_a dest_typ value
     | Abort, _ -> raise Deadcode
-    | Unknown, "__builtin_unreachable" when get_bool "sem.builtin_unreachable.dead_code" -> raise Deadcode (* https://github.com/sosy-lab/sv-benchmarks/issues/1296 *)
     | ThreadExit { ret_val = exp }, _ ->
       begin match ThreadId.get_current (Analyses.ask_of_ctx ctx) with
         | `Lifted tid ->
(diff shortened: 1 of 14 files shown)


src/analyses/mutexEventsAnalysis.ml
```diff
@@ -65,41 +62,27 @@ struct
...
-          (*print_endline @@ "Mutex `Lock "^f.vname;*)
-          lock ctx rw failing nonzero_return_when_aquired (Analyses.ask_of_ctx ctx) lv arg
-        | _ -> failwith "lock has multiple arguments"
-      end
-    | Unlock _, "__raw_read_unlock"
-    | Unlock _, "__raw_write_unlock"  ->
+    | Lock { lock = arg; try_ = failing; write = rw; return_on_success = nonzero_return_when_aquired }, _ ->
+      lock ctx rw failing nonzero_return_when_aquired (Analyses.ask_of_ctx ctx) lv arg
+    | Unlock arg, ("__raw_read_unlock" | "__raw_write_unlock") ->
+      (* TODO: why is this needed? *)
```
sim642, line 85: Why is there some special magic when unlocking raw kernel mutexes, but not when locking?

src/analyses/libraryFunctions.ml
```diff
@@ -30,18 +35,37 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+let linux_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
+    ("spin_lock_irqsave", special [__ "lock" []; drop "flags" []] @@ fun lock -> Lock { lock; try_ = get_bool "sem.lock.fail"; write = true; return_on_success = true });
+    ("spin_unlock_irqrestore", special [__ "lock" []; drop "flags" []] @@ fun lock -> Unlock lock);
+    ("_raw_spin_unlock_irqrestore", special [__ "lock" []; drop "flags" []] @@ fun lock -> Unlock lock);
+    ("spinlock_check", special [__ "lock" []] @@ fun lock -> Identity lock);
+    ("_lock_kernel", special [drop "func" [r]; drop "file" [r]; drop "line" []] @@ Lock { lock = big_kernel_lock; try_ = false; write = true; return_on_success = true });
+    ("_unlock_kernel", special [drop "func" [r]; drop "file" [r]; drop "line" []] @@ Unlock big_kernel_lock);
+    ("acquire_console_sem", special [] @@ Lock { lock = console_sem; try_ = false; write = true; return_on_success = true });
+    ("release_console_sem", special [] @@ Unlock console_sem);
+    ("misc_deregister", unknown [drop "misc" [r_deep; s_deep]]); (* TODO: why does this need to spawn to pass 04-mutex/40-rw_lock_rc? *)
```
sim642, line 67: Some races in that test arise from `misc_deregister(&test_dev);` calling functions from the device being freed, but there's also `misc_register(&test_dev);` which should already be doing that as well, so why is this necessary?

src/analyses/varEq.ml
```diff
@@ -509,21 +509,15 @@ struct
       |> remove_reachable ~deep:false ask shallow_args
       |> remove_reachable ~deep:true ask deep_args
 
-  let safe_fn = function
-    | "memcpy" -> true
-    | "__builtin_return_address" -> true
```
michael-schwarz, line 514: Why are we no longer treating these as safe?

src/analyses/libraryFunctions.ml
```diff
@@ -30,18 +35,37 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+let console_sem = AddrOf (Cil.var (Goblintutil.create_var (makeGlobalVar "[console semaphore]" intType)))
 
-]
+(** Linux kernel functions. *)
+(* TODO: conditional on kernel option *)
+let linux_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
+    ("spin_lock_irqsave", special [__ "lock" []; drop "flags" []] @@ fun lock -> Lock { lock; try_ = get_bool "sem.lock.fail"; write = true; return_on_success = true });
+    ("spin_unlock_irqrestore", special [__ "lock" []; drop "flags" []] @@ fun lock -> Unlock lock);
+    ("_raw_spin_unlock_irqrestore", special [__ "lock" []; drop "flags" []] @@ fun lock -> Unlock lock);
+    ("spinlock_check", special [__ "lock" []] @@ fun lock -> Identity lock);
```
michael-schwarz, line 62: Maybe a comment why this is Identity would be helpful here?

src/analyses/libraryDesc.ml
```diff
@@ -38,13 +38,16 @@ type special =
...
+  | CondWait of { cond: Cil.exp; lock: Cil.exp }
   | ThreadCreate of { thread: Cil.exp; start_routine: Cil.exp; arg: Cil.exp; }
   | ThreadJoin of { thread: Cil.exp; ret_var: Cil.exp; }
   | ThreadExit of { ret_val: Cil.exp; }
   | Math of { fun_args: math; }
   | Memset of { dest: Cil.exp; ch: Cil.exp; count: Cil.exp; }
   | Bzero of { dest: Cil.exp; count: Cil.exp; }
+  | Memcpy of { dest: Cil.exp; src: Cil.exp } (* TODO: separate for strcpy, etc *)
   | Abort
+  | Identity of Cil.exp
```
michael-schwarz, line 50: Maybe an explanation in what sense these are understood to be `Identity`?

src/analyses/base.ml
```diff
@@ -2506,49 +2506,33 @@ struct
...
-          let dest_a = eval_lv (Analyses.ask_of_ctx ctx) gs st dst_lval in
-          set ~ctx (Analyses.ask_of_ctx ctx) gs st dest_a dest_typ value
-        | _ -> failwith "strcpy arguments are strange/complicated."
-      end
-    | Unknown, "__builtin" ->
-      begin match args with
-        | Const (CStr ("invariant",_)) :: ((_ :: _) as args) ->
-          List.fold_left (fun d e -> invariant ctx (Analyses.ask_of_ctx ctx) ctx.global d e true) ctx.local args
-        | _ -> failwith "Unknown __builtin."
-      end
```
michael-schwarz, line 2549: What exactly was this doing before and why is it no longer needed now?

src/analyses/varEq.ml
sim642: They now have `LibraryFunctions` descriptions, such that they don't need to be hardcoded into a single analysis.

src/analyses/libraryDesc.ml
sim642: Literally the identity function, there's nothing more to it. Numerous annotating functions like `__builtin_expect` fall into this form to only aid the compiler optimizations.

src/analyses/libraryDesc.ml
```diff
@@ -38,13 +38,16 @@ type special =
...
   | Lock of { lock: Cil.exp; try_: bool; write: bool; return_on_success: bool; }
   | Unlock of Cil.exp
+  | CondWait of { cond: Cil.exp; lock: Cil.exp }
   | ThreadCreate of { thread: Cil.exp; start_routine: Cil.exp; arg: Cil.exp; }
   | ThreadJoin of { thread: Cil.exp; ret_var: Cil.exp; }
   | ThreadExit of { ret_val: Cil.exp; }
   | Math of { fun_args: math; }
   | Memset of { dest: Cil.exp; ch: Cil.exp; count: Cil.exp; }
   | Bzero of { dest: Cil.exp; count: Cil.exp; }
+  | Memcpy of { dest: Cil.exp; src: Cil.exp } (* TODO: separate for strcpy, etc *)
```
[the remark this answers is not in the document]
sim642: I'll add `Strcpy`, but there's no point in adding fields for data we don't actually use. This way it's clear that we don't, but otherwise it might look like we do. Only by looking at the `special` implementations of specific analyses, it would become apparent that we actually don't use it at all.
It's trivial to add it if we decide to do something with it.

src/analyses/libraryDesc.ml
michael-schwarz: Still, I think a comment to this effect may be helpful here.

src/analyses/base.ml
michael-schwarz: Ok, makes sense to remove it then!

src/analyses/libraryFunctions.ml
sim642: Added!

src/analyses/libraryDesc.ml
sim642: Added!

michael-schwarz:
From my perspective it looks good to go, though we should probably wait on @vesalvojdani 's input on the kernel / driver questions that were raised by this PR.

src/analyses/libraryFunctions.ml
vesalvojdani: Yes, de-register should not need to spawn these, but if only register spawns, we get the following accesses on `data1`:

```
[Success][Race] Memory location data1@tests/regression/04-mutex/40-rw_lock_rc.c:8:12-8:17 (safe):
  write with [lock:{read lock rwlock1}, thread:[test_init, my_read@tests/regression/04-mutex/40-rw_lock_rc.c:51:3-51:33]] (conf. 110) (tests/regression/04-mutex/40-rw_lock_rc.c:20:3-20:21)
  write with [lock:{rwlock1}, thread:[test_init, my_write@tests/regression/04-mutex/40-rw_lock_rc.c:51:3-51:33]] (conf. 110) (tests/regression/04-mutex/40-rw_lock_rc.c:27:3-27:27)
```

This is without MHP, but it's similar with it... This test was created before we had this pairwise view and was based on the heuristics that you shouldn't write with a read lock, but this is not a problem in itself. In this case, the above accesses are not conflicting if the threads are unique, but this access on line 20 should conflict with itself or do we asssume unknown-function spawns are unique?

sim642: I made it non-spawning and changed the test to use plain (not history) thread IDs instead, so that it doesn't have the uniqueness information. Unknown function spawns are a good point, tracked now by #843.

vesalvojdani (approved):

PR #838, opened 2022-09-21
Add option `sem.null-pointer.dereference`

sim642:
Adds an option, which determines how we handle `NULL` pointer dereferences: as top value or as bottom value (i.e. ignore the possibility). To my surprise, the latter is our current behavior, probably so we don't massively lose precision on real-world programs. It's also how SV-COMP programs should be, free of undefined behavior.

But we should have the option nevertheless. Interestingly, changing the default has no consequences on the regression suite.

In particular, this came up in https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1364. I (and Goblint also) didn't consider the possibility of the error being reachable due to dereferencing `NULL` pointer and possibly reading anything. The sv-benchmarks task is still wrong, but in a different way then I initially thought.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -416,7 +416,12 @@ struct
       in
       let f = function
         | Addr.Addr (x, o) -> f_addr (x, o)
-        | Addr.NullPtr -> VD.bot () (* TODO: why bot? *)
+        | Addr.NullPtr ->
+          begin match get_string "sem.null-pointer.dereference" with
+            | "assume_none" -> VD.bot ()
+            | "assume_top" -> top
+            | _ -> assert false
+          end
         | Addr.UnknownPtr -> top (* top may be more precise than VD.top, e.g. for address sets, such that known addresses are kept for soundness *)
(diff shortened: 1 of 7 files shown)


src/analyses/base.ml
```diff
@@ -416,7 +416,12 @@ struct
       in
       let f = function
         | Addr.Addr (x, o) -> f_addr (x, o)
-        | Addr.NullPtr -> VD.bot () (* TODO: why bot? *)
+        | Addr.NullPtr ->
```
michael-schwarz, line 419: Should we not also produce a warning here?

michael-schwarz (approved):

src/analyses/base.ml
sim642: We actually already produce warnings about this, but in another place about the whole address set:
https://github.com/goblint/analyzer/blob/d915302fbd008bdc13fafcf954c5f605ffed2c7d/src/analyses/base.ml#L1017-L1022

Issue #800, opened 2022-07-28
Maintain more precise lockset for recursive mutexes


michael-schwarz:
We should aim to support recursive mutexes more thoroughly: We could e.g. maintain an interval expressing how many times a recursive mutex is locked, or alternatively just maintain if it locked once or multiple times.

**This part remains open.**

~~Such an analysis that knows if a mutex is recursive or not would also allow us to warn in case of double-locking of non-recursive mutexes.~~ This part was done as part of #839.

Issue #718, opened 2022-05-04
Switch to HoareSet for MayLocks


michael-schwarz:
Opening this issue so we can refer to it later: It may be advantageous to use a HoareSet for MayLocksets instead of our naive set representation for which the operations are quadratic in the size of the sets.

Independently of this, we need to rethink our approach to MayLocks as it pertains to deadlock detection in general.

sim642:
The deadlock analysis uses a set of may lock _events_, consisting of:
1. the mutex,
2. the node where it was locked,
3. the `MCPAccess` data at that point.

In order to have a Hoare set of these events, there needs to be `leq` (and other lattice operations) defined for the events, but I'm not sure how that could be done because none of the three components are lattices.
The possibility I can imagine right now would be to use near-equivalent flat lattices for each of them, which allows some things to be forcefully joined together to `Top`, to essentially ignore those components and reduce the set size. For example, two lock events for the same lock with the same access data at different nodes might be joined together such that the node is forgotten. The exact conditions of when this happens must then be specified though, because it wouldn't come automatically (the flat lattices with lifted values are incomparable either way). So something like a `should_join` phase is needed to force joining of particular events.


> Independently of this, we need to rethink our approach to MayLocks as it pertains to deadlock detection in general.

What do you mean by this?

michael-schwarz:
> What do you mean by this?

We'll give an update at GobCon today.

sim642:
This probably deserves a separate issue because it's not related to this one, right?

To achieve that, I guess we'd need some kind of var_eq analysis, but with equalities in time, e.g. `*ptr@lock == *ptr@unlock`.
Maybe symb_locks already achieves something of the sort because it also needs to invalidate things at intermediate modifications. Although there's #434.

michael-schwarz:
Yes, might make sense to pull it out.

michael-schwarz:
Unassigning myself, as I am not currently planning on working on this.

PR #839, opened 2022-09-22
MayLock Analysis & Sanity Checks of Mutex Usage & Analysis of Mutex Types

michael-schwarz:
Some in-progress work on a May Lock Analysis that allows us to detect potential may double locking. Also contains some first attempt of tracking whether a mutex is potentially recursive or not, in order to emit appropriate warnings.

Relates (partially) to:
1. #176 
2. #658
3. #718
4. #800


michael-schwarz:
In the interest of reducing the number of stale PRs, I would propose to also merge this soon: It covers a meaningful subset of issues, even when further analyses are still conceivable.

sim642 (commented):
I don't see anything regarding #718. The lock domain is changed to use `SetDomain.Reverse`, completing a TODO, but that's about it.

michael-schwarz:
With the help of Yannick I was able to figure out the issue with OS X:

- `PTHREAD_MUTEX_RECURSIVE` is `1` on Linux
- but `2` on OS X

tests/regression/68-doublelocking/02-unknown.c
```diff
@@ -0,0 +1,43 @@
...
+    if(top) {
+        x = 3;
+    }
+
+    void* ptr2;
+    if(top) {
+        ptr2 = &mut[x];
+    } else {
+        ptr2 = &mut[3];
+    }
```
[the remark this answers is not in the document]
michael-schwarz: removed.

sim642:
> With the help of Yannick I was able to figure out the issue with OS X:
> 
>     * `PTHREAD_MUTEX_RECURSIVE` is `1` on Linux
> 
>     * but `2` on OS X

Interesting. Lucky that it's an enum value, not just a preprocessor define, because the latter would be a lot more annoying to deal with.

michael-schwarz:
Actually, it's only an enum on Linux, OS X does a preprocessor define https://opensource.apple.com/source/libpthread/libpthread-301.30.1/pthread/pthread.h.auto.html :(

michael-schwarz:
> I don't see anything regarding https://github.com/goblint/analyzer/issues/718. The lock domain is changed to use SetDomain.Reverse, completing a TODO, but that's about it.

The actual MayLockset analysis that is now used makes use of a HoareDomain (via the address domain from base). But you are right in that this has nothing to do with DeadlockAnalysis.

sim642:
> The actual MayLockset analysis that is now used makes use of a HoareDomain (via the address domain from base). But you are right in that this has nothing to do with DeadlockAnalysis.

But base's address domain doesn't use a Hoare domain:
https://github.com/goblint/analyzer/blob/bdfc3cb96f2d3bb331e81584243756fd46d01205/src/cdomains/addressDomain.ml#L24-L27

So I'm even less sure what this change is really achieving.

sim642:
> I guess it did when I started the PR, probably it has changed since then.

Ah! Looking at the dates, that address domain change was in august, so before this PR was opened, but #718 itself was opened before.

I was about to say that the change has no impact then, but actually it still does, just at the set level. Previously may locksets were essentially `SetDomain.Make (Addr)`, but `AddressDomain` uses the bucketed `DisjointDomain.ProjectiveSet (Addr)`.
The set implementation itself shouldn't be a performance difference since `SetDomain` operations all efficiently delegate to the standard library implementation (which I guess is $O(n \log n)$ at worst). Rather the difference comes precisely from the fact that the contents of buckets (e.g. different indices) are forcefully joined there, which could keep the set itself smaller.

I suppose that's what you meant at the time of opening #718 because then `AddressDomain` was built on the still-somewhat-inefficient `HoarePO`. And over time with various changes things got confusing.

src/analyses/mayLocks.ml
```diff
@@ -1,18 +1,25 @@
...
+  module D = LockDomain.MayLocksetNoRW
   module G = DefaultSpec.G
   module V = DefaultSpec.V
 
-  let add ctx l =
-    D.add l ctx.local
+  let add ctx (l,r) =
+    if D.mem l ctx.local then
+      match D.Addr.to_var_must l with
+      | Some v when ctx.ask (Queries.IsRecursiveMutex v)->
```
sim642, line 15: `to_var_must` only accepts `NoOffset` to make this work only on direct `varinfo`s, but our locks are `lval`s, even in the non-symbolic case. Maybe the query could also work on lvalues, especially since the base analysis tracks mutex values arbitrarily deep in offsets.

src/cdomains/mutexAttrDomain.ml
```diff
@@ -0,0 +1,26 @@
+module MutexKind =
+struct
+  include Printable.Std
+
+  type t = NonRec | Recursive [@@deriving eq, ord, hash, to_yojson]
```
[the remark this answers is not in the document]
michael-schwarz: I added some comments about the other mutex types, but as we do not have proper support for them yet (i.e. checking return types), I'd leave this as it is for now.

src/analyses/mayLocks.ml
```diff
@@ -1,18 +1,25 @@
...
+        ctx.local
+      | _ ->
+        (M.warn ~category:M.Category.Behavior.Undefined.double_locking "Acquiring a (possibly non-recursive) mutex that may be already held"; ctx.local)
+    else
+      D.add l ctx.local
 
-  let remove ctx l =
-    D.remove (l, true) (D.remove (l, false) ctx.local)
+  let remove ctx l = D.remove l ctx.local
```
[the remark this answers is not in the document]
michael-schwarz: Must not be held in this case. I added an appropriate warning.

tests/regression/71-doublelocking/09-other-dyn.c
```diff
@@ -0,0 +1,55 @@
...
+
+
+void* f2(void* ptr) {
+    pthread_mutex_t* mut = (pthread_mutex_t*) ptr;
+
+    pthread_mutex_lock(mut);
+    pthread_mutex_unlock(mut);
+
+    // To check that this is now actually removed from the may lockset
+    return NULL; //WARN
```
sim642, line 27: Why `// WARN` if it's not recursive?

src/analyses/mayLocks.ml
```diff
@@ -1,18 +1,37 @@
...
-    D.add l ctx.local
+  let add ctx (l,r) =
+    if D.mem l ctx.local then
+      let default () = (M.warn ~category:M.Category.Behavior.Undefined.double_locking "Acquiring a (possibly non-recursive) mutex that may be already held"; ctx.local) in
+      match D.Addr.to_var_must l with
+      | Some v ->
+        (let mtype = ctx.ask (Queries.MutexType v) in
+         match mtype with
+         | `Lifted MutexAttrDomain.MutexKind.Recursive -> ctx.local
+         | `Lifted MutexAttrDomain.MutexKind.NonRec -> (M.warn ~category:M.Category.Behavior.Undefined.double_locking "Acquiring a non-recursive mutex that may be already held"; ctx.local)
```
sim642, line 20:
```suggestion
         | `Lifted MutexAttrDomain.MutexKind.NonRec ->
           M.warn ~category:M.Category.Behavior.Undefined.double_locking "Acquiring a non-recursive mutex that may be already held";
           ctx.local
```

src/analyses/mutexTypeAnalysis.ml
```diff
@@ -0,0 +1,69 @@
+(** An analysis tracking the type of a mutex. *)
```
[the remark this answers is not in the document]
sim642: But mutexes themselves might also be nested arbitrarily deep in structs or arrays, while the global invariant for this analysis only considers top-level `varinfo` mutexes.

I might've already mentioned it somewhere, but the lockset analyses themselves allow addresses more generally. Seems unnecessarily imprecise to not track mutex types as granularly as we track mutexes anyway.

src/cdomains/mutexAttrDomain.ml
```diff
@@ -0,0 +1,43 @@
...
+
+
+let of_int z =
+  if Z.equal z Z.zero then
+    `Lifted MutexKind.NonRec
+  else
+    let recursive_int = Lazy.force recursive_int in
+    match recursive_int with
+    | r when Z.equal z r -> `Lifted MutexKind.Recursive
+    | _ -> `Top
```
sim642, line 43: Since the constant is no longer matched, but detected dynamically, this is just an `if`:
```suggestion
    if Z.equal z recursive_int then
      `Lifted MutexKind.Recursive
    else
      `Top
```

src/cdomains/valueDomain.ml
```diff
@@ -99,9 +102,13 @@ struct
     | `Thread of Threads.t
     | `JmpBuf of JmpBufs.t
     | `Mutex
+    | `MutexAttr of MutexAttrDomain.t
```
sim642, line 105: This case is missing from all the lattice operation definitions. The compiler doesn't warn because they have `_, _` default cases.

michael-schwarz: True, I have fixed it now!

tests/regression/71-doublelocking/07-rec-dyn-osx.c
```diff
@@ -0,0 +1,101 @@
...
+int pthread_mutexattr_settype(pthread_mutexattr_t *, int);
+
+
+int g;
+
+void* f1(void* ptr) {
+    pthread_mutex_t* mut = (pthread_mutex_t*) ptr;
+
+    pthread_mutex_lock(mut);
+    pthread_mutex_lock(mut);
```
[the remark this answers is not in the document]
michael-schwarz: The problem was that it is from the absence of an appropriate enum that we deduce that it is OS X. I managed to get this working now by excluding a header.

src/analyses/mutexTypeAnalysis.ml
```diff
@@ -6,6 +6,13 @@ open Analyses
 module MAttr = ValueDomain.MutexAttr
 module LF = LibraryFunctions
 
+(* Removing indexes here avoids complicated lookups inside the map for a varinfo, at the price that different types of mutexes in arrays are not dinstinguished *)
+let rec offs_no_index o =
+  match o with
+  | `NoOffset -> `NoOffset
+  | `Field (f,o) -> `Field (f, offs_no_index o)
+  | `Index (i,o) -> `Index (Lval.any_index_exp, offs_no_index o)
```
sim642: Isn't `MutexAnalysis.Spec.conv_offset_inv` already this, but slightly better?

src/analyses/mutexTypeAnalysis.ml
```diff
@@ -26,16 +26,30 @@ struct
 
   (* transfer functions *)
   let assign ctx (lval:lval) (rval:exp) : D.t =
+    (* replaces the rightmost offset with r *)
+    let rec replace_no_offset o r = match o with
+      | `NoOffset -> r
+      | `Field (f,o) -> `Field (f, replace_no_offset o r)
+      | `Index (i,o) -> `Index (i, replace_no_offset o r)
```
sim642, line 29: Isn't this just `Lval.Offset.add_offset`?

src/analyses/mutexTypeAnalysis.ml
```diff
@@ -0,0 +1,94 @@
...
+open Analyses
+
+module MAttr = ValueDomain.MutexAttr
+module LF = LibraryFunctions
+
+
+
+module Spec : Analyses.MCPSpec with module D = Lattice.Unit and module C = Lattice.Unit =
+struct
+  include Analyses.DefaultSpec
```
sim642, line 13: `IdentitySpec` would reduce the amount of boilerplate here.

src/cdomains/lval.ml
```diff
@@ -628,3 +628,32 @@ struct
     end
     )
 end
+
+module OffsetNoIdx =
+struct
+  type someidx = SomeIdx [@@deriving eq, ord, hash]
+
+  include Printable.StdLeaf
+  type t = (CilType.Fieldinfo.t, someidx) offs [@@deriving eq, ord, hash]
```
sim642, line 637: I'm really not a fan of yet another offset and value type. We already have way too many of them and all the copy-pasted conversions all over the place...

Address domain bucket representatives already are equivalent to this and avoid defining a custom `unit` type. Also access domain implements another equivalent copy. And there's probably more.

Although I still think it would be best for consistency to just use the same mutex lvalues type for may locks as we already have used for must locks for a decade.

src/domains/queries.ml
```diff
@@ -312,6 +321,11 @@ struct
       | Any (Invariant i1), Any (Invariant i2) -> compare_invariant_context i1 i2
       | Any (InvariantGlobal vi1), Any (InvariantGlobal vi2) -> Stdlib.compare (Hashtbl.hash vi1) (Hashtbl.hash vi2)
       | Any (IterSysVars (vq1, vf1)), Any (IterSysVars (vq2, vf2)) -> VarQuery.compare vq1 vq2 (* not comparing fs *)
+      | Any (MutexType (v1,o1)), Any (MutexType (v2,o2)) -> let r = CilType.Varinfo.compare v1 v2 in
+        if r <> 0 then
+          r
+        else
+          Lval.OffsetNoIdx.compare o1 o2
```
sim642, line 328: As mentioned above, ideally this would just use an already existing type for mutex lvals, so this comparison wouldn't have to be part of queries themselves. Either way, `[%ord: CilType.Varinfo.t * Lval.OffsetNoIdx.t]` is how to just derive these definitions inline.

src/cdomains/lval.ml
michael-schwarz: > Although I still think it would be best for consistency to just use the same mutex lvalues type for may locks as we already have used for must locks for a decade.

The problem is the offsets from `IdxDomain` that are used in these. I do not want to have to iterate over all buckets to see which indexes overlap and the join these.

src/domains/queries.ml
michael-schwarz: I derive them now.

src/cdomains/lval.ml
michael-schwarz: One could use the same type and replace all index offsets with unknown, but I am a bit wary of relying on such things if they are not expressed in the type as someone could easily change that back.

src/analyses/mutexTypeAnalysis.ml
michael-schwarz: The comment refers to an earlier version, we now use `Lval.OffsetNoIdx.of_offs` everywhere.

src/analyses/mutexTypeAnalysis.ml
```diff
@@ -14,7 +21,8 @@ struct
   let name () = "pthreadMutexType"
   module D = Lattice.Unit
   module C = Lattice.Unit
-  module G = MAttr
+
+  module G = MapDomain.MapBot_LiftTop (Lval.CilLval) (MAttr)
```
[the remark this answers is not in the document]
michael-schwarz: Yes, since I have replaced the indexes with unit index, this is possible now!

src/cdomains/lval.ml
sim642: > The problem is the offsets from `IdxDomain` that are used in these. I do not want to have to iterate over all buckets to see which indexes overlap and the join these.

Doesn't that mean the must locks analysis already has some issue with checking index aliasing? Because the improved version I mentioned above specifically keeps constant integer offsets around.

> One could use the same type and replace all index offsets with unknown, but I am a bit wary of relying on such things if they are not expressed in the type as someone could easily change that back.

The bucket representative offsets already implement a fully isomorphic type though with precisely the same meaning of a `unit` index for any index. There's no reason not to just reuse that module.

michael-schwarz: I am not sure how much reuse we can get with that though, as it works for general addresses, and I do not want `NullPtr` or `StrPtr` to be possible values.

~~~OCaml
  (** Representatives for lvalue sublattices as defined by {!NormalLat}. *)
  module R: DisjointDomain.Representative with type elt = t =
  struct
    type elt = t

    (* Offset module for representative without abstract values for index offsets, i.e. with unit index offsets.
       Reason: The offset in the representative (used for buckets) should not depend on the integer domains,
       since different integer domains may be active at different program points. *)
    include Normal (UnitIdxDomain)

    let rec of_elt_offset: (fieldinfo, Idx.t) offs -> (fieldinfo, UnitIdxDomain.t) offs =
      function
      | `NoOffset -> `NoOffset
      | `Field (f,o) -> `Field (f, of_elt_offset o)
      | `Index (_,o) -> `Index (UnitIdxDomain.top (), of_elt_offset o) (* all indices to same bucket *)

    let of_elt (x: elt): t = match x with
      | Addr (v, o) -> Addr (v, of_elt_offset o) (* addrs grouped by var and part of offset *)
      | StrPtr _ when GobConfig.get_bool "ana.base.limit-string-addresses" -> StrPtr None (* all strings together if limited *)
      | StrPtr x -> StrPtr x (* everything else is kept separate, including strings if not limited *)
      | NullPtr -> NullPtr
      | UnknownPtr -> UnknownPtr
  end
~~~

michael-schwarz: @sim642: Is that what you had in mind? https://github.com/goblint/analyzer/pull/839/commits/01b9841b262a911f79d30328d3d9b245408cc309

sim642 (approved):
The last big question is how expensive these things are. Ideally we could just have these activated by default.

michael-schwarz:
I will merge them for now (without changing the defaults), and do an sv-comp run after!

PR #840, opened 2022-09-27
Derive `CilType` with ppx

sim642:
This work stems from profiling #796, where I observed the following:
1. `CilType.Exp` has too many hash collisions from its polymorphic hashing. Notably, all offsets of accessing a struct variable produce an `lval`/`exp` with the same hash. This is probably because the polymorphic hashing gets stuck in the struct type circular references and never hashes the offset itself.
2. `CilType.Typ` comparisons spend most of their time converting the `typ`s to temporary `typsig`s, which involve many allocations and GC. Furthermore, the implementation of `Cil.typeSig`, which does the conversion, isn't the best either: it creates fresh visitor objects at every level of recursion to handle attributes.

Therefore, to get proper identity function (comparison and hashing) for all these recursively related CIL types, this PR uses ppx to derive these functions for the entire type family. This avoids the usual pitfalls of polymorphic comparisons because the usual suspects (`compinfo`, `fieldinfo`, possibly others) still have manual implementations which don't recurse into _all_ the fields.

### TODO
- [x] Performance change on sv-benchmarks.
- [x] Check if physical equality checks help any further.

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -285,7 +285,7 @@ module IntDomLifter (I : S) =
 struct
   open Cil
   type int_t = I.int_t
-  type t = { v : I.t; ikind : (ikind [@equal (=)] [@compare Stdlib.compare] [@hash fun x -> Hashtbl.hash x]) } [@@deriving eq, ord, hash]
+  type t = { v : I.t; ikind : CilType.Ikind.t } [@@deriving eq, ord, hash]
 
   (* Helper functions *)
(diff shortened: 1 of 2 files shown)


sim642:
On sv-benchmarks, this gives an average ~9% speedup:
image
[Full results table](https://goblint.cs.ut.ee/results/94-all-fast-ciltype-deriving/results/results.2022-09-27_19-54-41.table.html#/scatter?toolY=1&columnY=1&results=All&regression=Linear&scaling=Linear&line=1.5&filter=0(0*status*(category(in(unknown,correct)))),1(0*status*(category(in(unknown,correct)))))

sim642:
I added physical equality checks to some and it makes another ~1..2% improvement, so we can have those as well.

michael-schwarz (approved):
Great job! It continues to amaze me where these performance improvements are hiding!

PR #841, opened 2022-09-28
Fixes for the incremental analysis with cfg comparison

### LINKED ISSUE #835: Three incremental tests failing with the cfg comparison
Currently three incremental test cases fail when the cfg comparison is activated. The first two throw a `Not_found` exception, for the third the result is not as expected.

- [ ] 03/02 reluctant-int-annotation
- [ ] 03/03 reluctant-int-annotation-dyn
- [ ] 13/05 race-call-remove

Since the cfg comparison often breaks when new features are introduced and the only the AST comparison is considered when adapting the incremental pre-processing for the solver, I would suggest to run the incremental tests also with the cfg comparison in the regular regression test runs.

```
Testing 03/02 precision-annotation/reluctant-int-annotation      Status: 2 (exception)

  - Timings:
  - TOTAL                           0.007 s
  -   parse                           0.002 s
  -   convert to CIL                  0.003 s
  -   mergeCIL                        0.002 s
  -   compareCilFiles                 0.001 s
  -     createCFG                       0.001 s
  -       handle                          0.001 s
  -       iter_connect                    0.001 s
  -         computeSCCs                     0.001 s

Testing 03/03 precision-annotation/reluctant-int-annotation-dyn  Status: 2 (exception)

  - Timings:
  - TOTAL                           0.008 s
  -   parse                           0.003 s
  -   convert to CIL                  0.003 s
  -   mergeCIL                        0.002 s
  -   compareCilFiles                 0.001 s
  -     createCFG                       0.001 s
  -       handle                          0.001 s
  -       iter_connect                    0.001 s
  -         computeSCCs                     0.001 s

03/02 failed!

Expected assert, but registered nothing on reluctant-int-annotation-dyn:10
  __goblint_check(in == 17);
Expected fail, but registered nothing on reluctant-int-annotation-dyn:21
  __goblint_check(a); // FAIL!
Expected assert, but registered nothing on reluctant-int-annotation-dyn:23
  __goblint_check(a == 17);
03/03 failed!

Expected norace, but registered race on race-call-remove:7
  g++; // NORACE (unique thread)
Expected norace, but registered race on race-call-remove:12
  g++; // NORACE
13/05 failed!

3 test(s) failed: ["03/02 reluctant-int-annotation", "03/03 reluctant-int-annotation-dyn", "13/05 race-call-remove"]
```

stilscher:
This PR arises from several incremental test cases that failed when enabling the cfg comparison (#835). With the grouping of globals (#836) and the bug fix in 6f2486ac0bd85fa97b8a0d7a45f73035975a3d87 a couple of problems were fixed but resulted in more/different test cases failing.

This PR includes:

- fix reluctant destabilization: return nodes of partially changed functions need to be collected for reluctant destabilization (not only when reluctant is turned off!)
- extended `update_suite.rb` with the option `-c` to run incremental tests with cfg comparison
- run `update_suite.rb -c` regularly with regression tests

It remains to do:
- [x] the incremental test 11/15 mutex-simple-wrap1 still fails with cfg comparison because an expected race is not found. It seems that the restarting differs from the run with the ast comparison, but it needs to be investigated further what the actual issue is. 
Update: The reason seems to be that fuel is lost when destabilizing through `side_dep` first before destabilizing through `infl`. I suggest to change the behaviour of the destabilize_with_sides in #862.

Closes #835

### DIFF
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -508,16 +508,17 @@ module WP =
               (* collect function return for reluctant analysis *)
               mark_node obsolete_ret f (Function f)
           ) changed_funs;
-        (* Unknowns from partially changed functions need only to be collected for eager destabilization when reluctant is off *)
+        (* Primary changed unknowns from partially changed functions need only to be collected for eager destabilization when reluctant is off *)
+        (* The return nodes of partially changed functions are collected in obsolete_ret for reluctant analysis *)
         (* We utilize that force-reanalyzed functions are always considered as completely changed (and not partially changed) *)
-        if not reluctant then (
-          List.iter (fun (f, pn, _) ->
+        List.iter (fun (f, pn, _) ->
+            if not reluctant then (
               List.iter (fun n ->
                   mark_node obsolete_prim f n
                 ) pn;
-              mark_node obsolete_ret f (Function f);
-            ) part_changed_funs;
-        );
+            );
+            mark_node obsolete_ret f (Function f);
+          ) part_changed_funs;
 
         let old_ret = HM.create 103 in
         if reluctant then (
--- a/src/solvers/td3.ml
+++ b/src/solvers/td3.ml
@@ -516,8 +516,9 @@ module WP =
               List.iter (fun n ->
                   mark_node obsolete_prim f n
                 ) pn;
-            );
-            mark_node obsolete_ret f (Function f);
+            ) else (
+              mark_node obsolete_ret f (Function f);
+            )
           ) part_changed_funs;
 
         let old_ret = HM.create 103 in
--- a/.github/workflows/locked.yml
+++ b/.github/workflows/locked.yml
@@ -67,6 +67,9 @@ jobs:
       - name: Test incremental regression
         run: ruby scripts/update_suite.rb -i
 
+      - name: Test incremental regression with cfg comparison
+        run: ruby scripts/update_suite.rb -c
+
   extraction:
     if: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' }}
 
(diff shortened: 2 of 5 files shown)


src/solvers/td3.ml
```diff
@@ -508,16 +508,17 @@ module WP =
...
+        List.iter (fun (f, pn, _) ->
+            if not reluctant then (
               List.iter (fun n ->
                   mark_node obsolete_prim f n
                 ) pn;
-              mark_node obsolete_ret f (Function f);
-            ) part_changed_funs;
-        );
+            );
+            mark_node obsolete_ret f (Function f);
```
jerhard, line 520: Could this line be in the `else`-clause? `obsolete_ret` and `old_ret` are only accessed when `reluctant` is on.

stilscher: I think that should be possible.

jerhard (approved):

jerhard:
Looks good!

sim642:
Is there a reason this is to be merged into #836 instead of `master`? I just merged `master` into this to confirm that #862 fixes the test, not realizing it's not against `master`. This made the diff really ugly here, so if there was a reason for this, I'll hard reset this branch back before my merge.

stilscher:
The only reason was, was that I needed #836 to see which tests are actually failing when grouping the globals correctly. I am not sure whether it's now best to wait for #836 to be completed (it should be already, but the cil fix for c99 inlines is still missing) or merge this into master directly.

sim642:
I tried to see how it looks like rebased for `master`, but it gets overly messy and the CFG testing added here due to #836. So I reset it back and lets wait for #836.

PR #844, opened 2022-10-03
Improve timing, add Trace Event Format output

sim642:
### Changes
1. Copy CIL's `Stats` into Goblint as `Timing` and clean it up. Although as-is CIL barely uses `Stats`, there's a large number of usages in some CIL plugins, so I decided not to remove it from CIL itself, but just not use it in Goblint. This is similar to what has been done with tracing at some point: CIL still contains `Trace`, but Goblint contains an improved copy as `Tracing`.
2. Rename `time` to `wrap`, so it's `Timing.wrap` instead of `Stats.time` now.
3. Fix inconsistent intermediate timing printouts. Previously it was possible that the sum of subtree times of a tree was greater than that tree itself if that tree was still in the middle of timing and hadn't exited cleanly. Now printing also adds all the in-progress times.
4. Add and improve measured data (all can be individually toggled):
    1. CPU time now includes CPU time of child processes (preprocessor executions). Previously preprocessing time couldn't be accounted for (even if it was wrapped in timing) because waiting for subprocesses exiting suspends Goblint and it doesn't do any work.
    2. Wall time is now measured as well. This rarely makes a difference (especially with preprocessing CPU time accounted for), but there is one case: when `jobs` is used to run multiple preprocessors in parallel, then wall time (which is what really matters for interactive usage) will be less than CPU time (which is more indicative of actual work done).
    3. Allocated memory is now measured as well. This is an idea I got from the landmarks library.
    5. Call counting existed in CIL already, but was disabled by default and we never used it, but I've recently found it to be quite valuable information. Otherwise if something takes 1s, then it might seem like something very slow is happening, but maybe it just gets executed 10000 times instead of the expected 100.
4. Rename `printstats` to `dbg.timing.enabled`.
6. Timing is generalized into a functor, which allows multiple simultaneous timing hierarchies. This allows to separately time how long we analyze what functions, for example.
7. Add Trace Event Format (TEF) output (activated using `dbg.timing.tef`) for both our normal timings and function analysis times (aka my magic flamegraphs). Documentation for viewing this is included. In TEF additional metadata can be included for timed sections, e.g. parsing timings have their filename attached, so it's easy to find out which files take the longest to parse.

### TODO
- [ ] Update bench repo scripts with `printstats` rename.


src/util/timing/goblint_timing.mli
```diff
@@ -0,0 +1,3 @@
+(** Profiling with custom hierarchical timed sections. *)
+
+include Goblint_timing_intf.Goblint_timing (** @inline *)
```
michael-schwarz, line 3: Any particular reason for this construction?

src/util/timing/goblint_timing.ml
```diff
@@ -0,0 +1,220 @@
+include Goblint_timing_intf
+
+(** Dummy options used for initialization before {!S.setup} is called. *)
```
michael-schwarz, line 3:
```suggestion
(** Dummy options used for initialization before {!S.start} is called. *)
```

`start` ?

src/util/timing/goblint_timing.mli
sim642: It avoids having to duplicate the type and module type definitions both in `.mli` and `.ml`: https://www.craigfe.io/posts/the-intf-trick.

michael-schwarz: I was aware of what it does (but not that it's something Jane Street does in their packages all the time.
I'm just not sure if it is smart to introduce this here given that we don't use this pattern anywhere else....

src/util/timing/goblint_timing.ml
sim642: Yes, good catch!

src/util/timing/goblint_timing.mli
michael-schwarz: Ok, then we can start doing now.

michael-schwarz (approved):

Issue #750, opened 2022-06-02
Unsoundness on svcomp22 reachsafety test


MartinWehking:
The following test becomes true but should actually be false:

- [x] sv-benchmarks/c/ldv-validator-v0.8/linux-stable-8817633-1-144_2a-drivers--staging--media--easycap--easycap.ko-entry_point_ldv-val-v0.8.cil.out.c

I've used the following settings:
```
{
  "ana": {
    "sv-comp": {
      "enabled": true,
      "functions": true
    },
    "int": {
      "def_exc": true,
      "enums": false,
      "interval": true
    },
    "activated": [
      "base",
      "threadid",
      "threadflag",
      "mallocWrapper",
      "mutex",
      "mutexEvents",
      "access",
      "escape",
      "expRelation",
      "symb_locks",
      "region",
      "thread"
    ],
    "context": {
      "widen": false
    },
    "malloc": {
      "wrappers": [
        "kmalloc",
        "__kmalloc",
        "usb_alloc_urb",
        "__builtin_alloca",
        "kzalloc",

        "ldv_malloc",

        "kzalloc_node",
        "ldv_zalloc",
        "kmalloc_array",
        "kcalloc"
      ]
    },
    "base": {
      "arrays": {
        "domain": "partitioned"
      }
    }
  },
  "exp": {
    "region-offsets": true
  },
  "solver": "td3",
  "sem": {
    "unknown_function": {
      "spawn": false
    },
    "int": {
      "signed_overflow": "assume_none"
    }
  },
  "witness": {
    "id": "enumerate",
    "unknown": false
  }
}
```

sim642:
In SV-COMP 2022 we timed out on this task after 960s, by the way.

sim642:
I believe this benchmark has UB. `usb_altnum_to_altsetting` `malloc`s some memory into `alt` and then reads from it without initializing anything. At `ep = & (alt->endpoint + (unsigned long )j)->desc;` due to the bottom in the blob, we somehow get that `ep` is an empty address set (which we might want to change as well), so in the following `ep == 0` condition both branches end up dead (#826 would be useful for these situations). The following dead code might be the reason why we consider `reach_error` unreachable.

I'll make a MR to sv-benchmarks.

michael-schwarz:
Still the case with most recent version of Goblint

sim642:
The sv-benchmarks MR for excluding this program has been merged.

PR #849, opened 2022-10-07
Add warning when both branches are dead

### LINKED ISSUE #826: Sanity check both branches not dead
Every now and then we've had unsoundness issues at branches which have both (true and false) outgoing edges `raise Deadcode` and yield bottom states. We should have a sanity check for this.

After #525, the dead branch reporting uses a flat boolean lattice:
https://github.com/goblint/analyzer/blob/21e59a1ad72c49bf448066123ead989a746334c4/src/framework/constraints.ml#L1118-L1125
Thus the `Bot` case would be when neither outgoing edge is live.
As the comment mentions, only problem is our `Neg(1)` edges, which don't have a corresponding `Pos(1)`, so the single false branch being trivially dead means that there's also a `Bot`.

michael-schwarz:
This adds a warning when both branches of a condition are unreachable, usually hitning at an issue in the analyzer. Does not produce such warnings for edges with guard `1`, as Gpblint adds `Neg(1)` edges to ensure connectedness of CFGs.

Closes #826.


sim642 (approved):
Catches the problem from #750 :+1:

PR #853, opened 2022-10-12
Fix Incorrect Shebangs

just-max:
Fix incorrect shebangs in `trace_{off,on}.sh` scripts
https://www.shellcheck.net/wiki/SC1113


sim642 (approved):

michael-schwarz:
Thanks for fixing this! Those are the tiny things that often get overlooked!

PR #854, opened 2022-10-13
tracing: add option to set pretty printer output width

just-max:
The tracing module uses `Pretty.doc` values, but sets `max_int` as the width of the output. Therefore, any `doc` that uses optional line breaks won't use the opportunity to insert line breaks into the output, since as far as the pretty printer is concerned, there is infinite width.

Adds an option `dbg.trace_width`. The default width is 80, suitable for basically all screens. If set to 0, the old behaviour is restored.

### DIFF
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -40,6 +40,12 @@ let print_help ch =
   fprintf ch "Some common configurations to start from can be found in conf/examples/*\n";
   exit 0
 
+let _ =
+  (* avoid a circular dependency between Tracing and GobConfig by using a reference *)
+  Tracing.get_trace_width :=
+    fun () ->
+      match get_int "dbg.trace_width" with 0 -> Stdlib.Int.max_int | w -> w
+
 (** [Arg] option specification *)
 let rec option_spec_list: Arg_complete.speclist Lazy.t = lazy (
   let add_string l = let f str = l := str :: !l in Arg_complete.String (f, Arg_complete.empty) in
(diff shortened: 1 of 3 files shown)


sim642:
> Therefore, any `doc` that uses optional line breaks won't use the opportunity to insert line breaks into the output, since as far as the pretty printer is concerned, there is infinite width.

That's by design. The `Pretty` module is hilariously bad at respecting the width. For example, in `./regtest.sh 04 01 --trace sol` output only 93% of lines exceed the 80 character limit and it's much worse any non-trivial program.

The intermittent line breaks also introduce unnecessary complication to grepping multi-million line tracing logs. If you want line wrapping, just enable soft wrapping in your terminal/editor and it _guarantees_ wrapping at any width. When looking at real tracing output with hundreds or thousands of level of indentation, line breaks make the nesting structure completely undecipherable as well.

The whole pretty-printing system needs a complete overhaul (#213) and properly reimplementing of every pretty-printer in Goblint and CIL. There's no point in a half-assed solution which only works a fraction of the time.

michael-schwarz:
Still, thank you for making the PR here @just-max, it's something that has been bugging us for a while, but so far we have been too lazy to do the proper solution #213.

If it helps during development for you, I guess the fix is local enough that you can just stash it and then re-apply whenever you pull from master.

sim642:
I suspect that an option lookup (even with caching) at every single trace call will make tracing even slower than it already is. Currently even int domain activation option lookups have a noticeable overhead because they are looked up again and again at every intermediate operation in every transfer function. There's precise evalint tracing around most of that.

Issue #852, opened 2022-10-12
EvalAssert transformation crashes on some programs


jerhard:
When activating the `EvalAssert` transformation, goblint crashes on some programs when the transformation is performed.
In particular, this happens on all of the `bench/coreutil` programs. Example command-line:

```
 ./goblint  --trans.activated '["assert"]' -v ../bench/coreutils/cksum_comb.c
```

Goblint will crash with the following stack trace (on `master`, b4c0057e1d82d9649b627af96f5237b2e4a8e1e0):
<details>
  <summary>Stack trace</summary>

```
Fatal error: exception Not_found
Raised at Stdlib__Hashtbl.MakeSeeded.find in file "hashtbl.ml", line 393, characters 21-36
Called from Goblint_lib__Control.AnalyzeCFG.analyze.solve_and_postprocess.ask.(fun) in file "src/framework/control.ml", line 582, characters 50-83
Called from Goblint_lib__Constraints.DeadBranchLifter.conv.(fun) in file "src/framework/constraints.ml", line 1095, characters 29-49
Called from Goblint_lib__MCP.MCP2.inner_ctx.(fun) in file "src/analyses/mCP.ml", line 341, characters 30-51
Called from Goblint_lib__Base.MainFunctor.priv_getg in file "src/analyses/base.ml", line 70, characters 32-49
Called from Goblint_lib__BasePriv.TracingPriv.read_global.getg in file "src/analyses/basePriv.ml", line 1654, characters 14-20
Called from Goblint_lib__BasePriv.PrecisionDumpPriv.read_global in file "src/analyses/basePriv.ml", line 1625, characters 12-42
Called from Goblint_lib__BasePriv.TracingPriv.read_global in file "src/analyses/basePriv.ml", line 1658, characters 12-42
Called from Goblint_lib__ValueDomain.ValueInvariant.deref_invariant in file "src/cdomains/valueDomain.ml", line 1295, characters 12-19
Called from Goblint_lib__ValueDomain.ValueInvariant.ad_invariant.(fun) in file "src/cdomains/valueDomain.ml", line 1249, characters 28-85
Called from Stdlib__Map.Make.fold in file "map.ml", line 321, characters 19-42
Called from Goblint_lib__ValueDomain.ValueInvariant.ad_invariant in file "src/cdomains/valueDomain.ml", line 1225, characters 16-1023
[...]

PR #855, opened 2022-10-13
Fix EvalAssert transformation to work on coreutil programs

jerhard:
The `EvalAssert` transformation is currently not able to produce C-output for the `bench/coreutil` programs, caused by several problems.

Todo:
- [x] Fix crash that occurs because some globals are not found by during the transformation #852 
- [x] Add declarations of functions before function definitions in the Cil-file that is produced. This way, it is ensured that functions are declared, before references to them are made in asserts (e.g. in asserts about function pointers).
- [x] ~Fix generation of asserts about pointers to malloced-memory. With malloced memory, there are (`tmp`-)variables of type `void*` that point to it. The assertions that are currently generated for these variables are not type correct, e.g. because a field offset is applied to a de-referenced void-pointer.~ Avoid generating invariants for pointers to abstract values of type `TVoid` for now.
- [x] Fix error when running gcc on the output of the transformation generated for `cp_comb.c`: ```extent-scan.c:107:18: error: ‘struct fiemap’ has no member named ‘f’```. Fix generation of invariants for unions.
- [x] Add option to not generate asserts for blobs. 

#### Testing this branch on `bench/coreutil` programs
Run `goblint` on the selected input file, e.g. `cp_comb.c`
```
./goblint  --set pre.cppflags[+] "--std=gnu89" --trans.activated '["assert"]' ../bench/coreutils/cp_comb.c
```
This will generate a `transformed.c`. One can check whether this compiles as follows:
```
gcc transformed.c  -lrt --std=gnu89
```


jerhard:
>I didn't check, but this sounds a lot like something I fixed in https://github.com/goblint/analyzer/pull/796: https://github.com/goblint/analyzer/commit/73df4cac9812c65a3446c7ea53770fa6d45d59d6. If that's the right thing, then you can just cherry pick it.

I needed to additionally add a check whether the pointed to abstract block has type `TVoid`, because otherwise it would still generate invariants without necessary casts. In the future one may want to look at the abstract value more closely, e.g. by inferring from the `field`s of a `Struct` which type it has. Then one could also generate invariants for these cases.

jerhard:
I tested this with `bench/cp_comb.c`: when one sets the `surroundByAtomic` in  `EvalAssert` to `false`, and adds an `#include<assert.h>` to the resulting `transformed.c`, the C-file can be compiled.

However, the executable one obtains crashes with a failed assertion:
 ```
a.out: hash.c:606: hash_initialize: Assertion `table___0->tuning == & default_tuning' failed.
Aborted (core dumped)
```

**Comment by sim642** [comment]
> However, the executable one obtains crashes with a failed assertion:
> 
> ```
> a.out: hash.c:606: hash_initialize: Assertion `table___0->tuning == & default_tuning' failed.
> Aborted (core dumped)
> ```

The code is this:
```c
#line 605
  tmp = malloc(sizeof(*table___0));
#line 605
  table___0 = (Hash_table *)tmp;
#line 606
  if ((unsigned long )table___0 == (unsigned long )((void *)0)) {
#line 607
    return ((Hash_table *)((void *)0));
  }
#line 609
  if (! tuning) {
#line 610
    tuning = & default_tuning;
  }
```
and the `tuning` argument actually always is a `NULL` pointer.

I think the issue is due to flow-insensitivity of the global alloc variable, which initially has all bot fields, but then gets `default_tuning` in there. But we don't distinguish at which point the initialization has happened, so we think the assertion about the global should also hold before.
This is quite an issue that we haven't considered before regarding the invariant generation.

jerhard:
>I think the issue is due to flow-insensitivity of the global alloc variable, which initially has all bot fields, but then gets default_tuning in there. But we don't distinguish at which point the initialization has happened, so we think the assertion about the global should also hold before.
This is quite an issue that we haven't considered before regarding the invariant generation.

Yes, indeed, we came to the same conclusion on the Munich side with @michael-schwarz. For the general case, involving possible multithreading, one would have to perform a `must-initialized` analysis to ensure that any abstract value we have for globals isn't "shadowing" a possible uninitialized value.

Since at least `cp_comb` is a single-threaded program and we assume that no threads are created, I reran the transformation with `--disable sem.unknown_function.spawn`. This should hopefully be sufficient to avoid that the analysis enters the multithreaded mode. When compiling the transformed code and running the resulting binary, the program then does crash with a `seg_fault`, instead of a failed assertion. There may be some assertion that dereferences some pointer that is not set properly.

jerhard:
Looking at this with @michael-schwarz and Helmut, we found that performing a context-insensitive analysis and not generating asserts for blobs gets rid of the asserts that fail during some simple executions of `cp`.

src/cdomains/valueDomain.ml
```diff
@@ -1246,7 +1246,18 @@ struct
                 else
                   Invariant.none
               in
-              let i_deref = deref_invariant ~vs ~lval vi offset (Mem c_exp, NoOffset) in
+              let i_deref =
+                match Cilfacade.typeOfLval (Var vi, offset) with
+                | TVoid _ -> Invariant.none (* TODO: Try to determine type of abstract value and introduce appropriate casts *)
```
sim642, line 1251: Doesn't this defeat the whole point of `i_deref` because all of our blobs have `void` type? The reason for these dereferenced invariants was precisely to be able to say something about allocated memory that otherwise cannot be referred to.

src/util/options.schema.json
```diff
@@ -2102,6 +2102,12 @@
                 "Whether to dump assertions about all local variables or limitting it to modified ones where possible.",
               "type": "boolean",
               "default": true
+            },
+            "blobs": {
+              "title": "witness.invariant.blobs",
```
sim642, line 2107: I think it would be more appropriate to call this `ana.base.invariant.blob` because blobs are a base analysis specific thing.

src/cdomains/unionDomain.ml
```diff
@@ -29,21 +29,22 @@ struct
...
     | Field (f, offset) ->
-      let c_lval = Option.get lval in
       begin match lift_f with
         | `Lifted f' ->
           let v = Values.cast ~torg:f'.ftype f.ftype v in
-          let f_lval = Cil.addOffsetLval (Field (f, NoOffset)) c_lval in
-          value_invariant ~offset ~lval:(Some f_lval) v
+          (* Do not add the field offset to lval here:
+             Otherwise, for pointers to a member type of the union, this would
+             generate an assertion with illegal field offset. *)
```
sim642, line 46: What's the example/test case for this? The pointer to an union field should be equal to the pointer to the union itself.
It's also out of sync with the `NoOffset` case now.

src/cdomains/valueDomain.ml
jerhard: Actually, after checking this again, it does not seem to be an issue anymore, at least with the asserts on blobs being deactivated (with a separate option).

src/cdomains/unionDomain.ml
jerhard: >What's the example/test case for this?

I added test case [`61/01`](https://github.com/goblint/analyzer/blob/fix_evalAssert/tests/regression/61-evalAssert/01-union_evalAssert.c).
https://github.com/goblint/analyzer/blob/a4f0a6e8749227e6eabb8ae0f4af155a95721f5a/tests/regression/61-evalAssert/01-union_evalAssert.c#L4-L23

 If one does not change the implementation here, then there will be invariants generated about `ptr` such as 
`assert(ptr->str.a == 1);`
`gcc` will complain that the `struct s` has no member `str`. 
The fix though is still flawed, as `bench/coreutils/dd_comb.c` shows: There I had to [change some asserts about unions](https://gitlab.com/erhardj/sv-benchmarks/-/commit/d10a2fe9ab4ed29d0d3c5f4258240fbbd7136f60#714c4d12a37819c79ac9db33e779ebd886327525_87516_87513), because the offset was missing.

>The pointer to an union field should be equal to the pointer to the union itself.

When one has a pointer to a union field, then generating the assertion should not add this offset again to the pointer (see example).

jerhard: #861 works on the example, and does not create the problems on `dd_comb.c`.

michael-schwarz:
Something else comes to mind here: Do we need to submit Goblint as a validator as well in order to benefit from us being able to solve such huge tasks in sv-comp?

Otherwise our ability to reverify these will not benefit us when it comes to points.

sim642 (approved):

PR #856, opened 2022-10-14
Server request for getting CFG node state

karoliineh:
Relates to goblint/GobPie#42.

Implements a server request to get CFG states by the node ID.


src/framework/node.ml
```diff
@@ -56,3 +56,7 @@ let find_fundec (node: t) =
   | Statement stmt -> Cilfacade.find_stmt_fundec stmt
   | Function fd -> fd
   | FunctionEntry fd -> fd
+
+let of_id s = 
+  let id = int_of_string s in
+  Statement { dummyStmt with sid = id }
```
sim642, line 62: This needs to handle the cases where `s` begins with `fun` or `ret` as well. To find the `fundec` by its ID for those cases, something hacky like this might work:
```ocaml
Cilfacade.find_varinfo_fundec {dummyFunDec.svar with vid = id}
```

src/framework/control.ml
```diff
@@ -39,6 +39,7 @@ let spec_module: (module Spec) Lazy.t = lazy (
 let get_spec (): (module Spec) =
   Lazy.force spec_module
 
+let current_xml : (Node.t -> Yojson.Safe.t) ref = ref (fun _ -> assert false)
```
sim642, line 42: We should probably rename this to something sensible like `current_node_state_json`.

src/util/server.ml
```diff
@@ -318,6 +318,18 @@ let () =
       (* TODO: also filter and include states info (as json) in the response for the requested function *)
   end);
 
+  register (module struct
+    let name = "node_state"
+    type params = { fname: string }  [@@deriving of_yojson]
```
sim642, line 323: The argument isn't a function (file?) name, but a node ID.

src/util/server.ml
```diff
@@ -318,6 +318,18 @@ let () =
...
+    type response = Yojson.Safe.t [@@deriving to_yojson]
+    let process { fname } serv =
+      let f = !Control.current_xml in
+      let n = Node.of_id fname in
+      let json = f n in
+      (* ignore @@ Pretty.printf "%s\n" (Yojson.Safe.to_string json); *)
+      json
+      (* TODO: also filter and include states info (as json) in the response for the requested function *)
```
sim642, line 331: I guess this copied comment is now outdated here and above?

sim642 (changes requested):

src/framework/node.ml
sim642: I cleaned up the code a bit, but didn't test it, so make sure it still works.

karoliineh: Made sure, still works. Thanks for the cleanup.

sim642 (approved):

Issue #780, opened 2022-07-12
Unsoundness for Two SV-Comp benchmarks


michael-schwarz:
- [x] `ldv-validator-v0.6/linux-stable-af3071a-1-130_7a-drivers--hwmon--s3c-hwmon.ko-entry_point.cil.c`
- [x] `ldv-linux-3.12-rc1/linux-3.12-rc1.tar.xz-144_2a-drivers--input--touchscreen--usbtouchscreen.ko-entry_point.cil`

The first benchmark seem to miss some sort of initialization (setting `s3c_hwmon_driver_group0` to anything other than what `zalloc` yields), which leads to an immediate abort in `s3c_hwmon_probe`, making code further down that is supposed to contain the call to `reach_error` unreachable.

I think it might be mislabled (CPAChecker and PESCO somehow produce a counterexample, I'll try to inspect that)

_Originally posted by @michael-schwarz in https://github.com/goblint/analyzer/issues/755#issuecomment-1181526369_

sim642:
When I did SV-COMP performance comparisons for #770, I actually saw 5 unsound tasks, including the ones you listed above: https://goblint.cs.ut.ee/results/72-all-fast-evalint-opt-after/table-generator.table.html#/table?filter=0(0*status*(category(in(wrong)))).
Have you not seen the others in your recent SV-COMP runs? Maybe those are just in categories you haven't included recently?

michael-schwarz:
The fourth one is the one fixed in #755, the other two might well be there additionally, these two are the ones we noticed a while back with affine equalities.

sim642:
Right now these two are the only two unsoundness cases that haven't been dealt with AFAIK. Are you still planning to look into these?

michael-schwarz:
Both are still there with the most recent Goblint version in the sv-comp config.

Plus we newly have
- [ ] `ldv-linux-3.14-races/linux-3.14--drivers--usb--misc--iowarrior.ko.cil.yml`

sim642:
iowarrior is not new, but it hasn't just been mentioned in any issue. The program itself is problematic, so there's already an sv-benchmarks MR for it: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1374.

sim642:
Both of the two tasks stem from the same problem: due to all zero memory from `ldv_zalloc`, a lot of the code is dead.
[...]

PR #859, opened 2022-10-18
SV-COMP 2023 setup

sim642:
This PR conflicts with #845 due to the changes in stubs structure.

### TODO
- [x] Add autotune conf


conf/svcomp23.json
```diff
@@ -0,0 +1,91 @@
...
+      "activated": [
+        "singleThreaded",
+        "mallocWrappers",
+        "noRecursiveIntervals",
+        "enums",
+        "arrayDomain",
+        "octagon",
+        "wideningThresholds"
+      ]
+    }
```
sim642, line 70: @leunam99 Is this the set of autotune activations that you'd recommend?

leunam99: Yes it is. One caveat is that this configuration also slightly increased the number of test with a wrong result in my benchmarks. 
(` sv-benchmarks/c/systemc/pipeline.cil-1.yml,`
`sv-benchmarks/c/ldv-linux-3.14-races/linux-3.14--drivers--usb--misc--iowarrior.ko.cil.yml`,
`sv-benchmarks/c/ldv-validator-v0.6/linux-stable-af3071a-1-130_7a-drivers--hwmon--s3c-hwmon.ko-entry_point.cil.out.yml`
and another one I would have to look up later today)

sim642: - [x] `systemc/pipeline.cil-1` is fixed by https://github.com/goblint/analyzer/pull/859/commits/2c257b60cde8c6b9214f378d92ee82c4f66ba458.
- [x] `ldv-linux-3.14-races/linux-3.14--drivers--usb--misc--iowarrior.ko.cil` is already wrong on master due to a task problem: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1374.
- [ ] `ldv-validator-v0.6/linux-stable-af3071a-1-130_7a-drivers--hwmon--s3c-hwmon.ko-entry_point.cil.out` is part of #780.

michael-schwarz:
We should also think about whether to enable the analysis that can keep more malloc'ed variables apart #722.

sim642:
> We should also think about whether to enable the analysis that can keep more malloc'ed variables apart #722.

According to https://github.com/goblint/analyzer/pull/722#issuecomment-1196424419, it made no difference, with 3 at least. If we want some higher value, we should do local sv-benchmarks experiments first (because the preruns turnaround time is much longer and they're currently limited to 60s anyway) to find such optimal value. Or maybe autotune needs to be adapted to also tune that.

Since this setup has successfully passed a prerun, the Apron packaging stuff works, which was the main concern. I'll merge this so conflicts in #845 can be resolved.

PR #863, opened 2022-10-21
Introduce new possible widening strategy `sides-local` for globals

michael-schwarz:
During #855 we discovered some asserts cannot be re-established. This is likely due to a different evaluation order.

Helmut and I traced down some surprising cases, where information about a a parameter is lost, despite it's values being from `[0,1]` and given explicitly in the program. This can be remedied in some cases by not designating a global a widening point, and instead selectively applying widening only for contributions from unknowns that have made increasing contributions before.

This seem like an interesting thing one would like to play around with, but it has immediately obvious limitations as well.

This PR adds the new option and an example program highlighting both where it helps and where it does not.

### DIFF
--- a/tests/regression/01-cpa/71-widen-sides.c
+++ b/tests/regression/01-cpa/71-widen-sides.c
@@ -1,4 +1,6 @@
 // PARAM: --set ana.ctx_insens "['base', 'mallocWrapper']"  --enable ana.int.interval --sets solvers.td3.side_widen sides-local
+#include <goblint.h>
+
 int further(int n) {
     // Even sides-local can not save us here :(
     __goblint_check(n <= 1); //TODO
--- a/tests/regression/01-cpa/71-widen-sides.c
+++ b/tests/regression/01-cpa/71-widen-sides.c
@@ -1,0 +1,28 @@
+// PARAM: --set ana.ctx_insens "['base', 'mallocWrapper']"  --enable ana.int.interval --sets solvers.td3.side_widen sides-local
+int further(int n) {
+    // Even sides-local can not save us here :(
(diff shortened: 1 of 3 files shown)


sim642:
Is this also supposed to help with the `instrumented_cp_comb.c` case from Zulip?

What's the trade-off with this? Is there a reason to not make it default? Can it be less precise than `sides`?

michael-schwarz:
> Is this also supposed to help with the `instrumented_cp_comb.c` case from Zulip?

It helps with some of the asserts but not all.

In general, it is incomparable to `sides`, as `sides` allows for two increasing contributions and makes it a `wpoint` **after**, whereas `sides-local` widens any increasing contributions from that unknown (widening for the **second** contribution already).

tests/regression/01-cpa/71-widen-sides.c
```diff
@@ -0,0 +1,28 @@
+// PARAM: --set ana.ctx_insens "['base', 'mallocWrapper']"  --enable ana.int.interval --sets solvers.td3.side_widen sides-local
+int further(int n) {
```
sim642, line 2: Required after #845:
```suggestion
#include <goblint.h>

int further(int n) {
```

sim642 (approved):

sim642:
> In general, it is incomparable to `sides`, as `sides` allows for two increasing contributions and makes it a `wpoint` **after**, whereas `sides-local` widens any increasing contributions from that unknown (widening for the **second** contribution already).

This implicit second-third widening delay built into the solver is a bit odd, because other people implement widening delays such that an arbitrary bound can be specified. At some point we probably should do that in the domain.

Issue #799, opened 2022-07-28
Support for `pthread_cleanup_handlers` and `at_exit`


michael-schwarz:
Those are features we currently do not support yet, but it might be interesting to do so.

sim642:
While looking at Concrat again, I found out that `pthread_cleanup_push` is a macro which doesn't actually store the cleanup routine is some magic place, but implements it using some flavor of `setjmp`. It's just that the `longjmp` isn't explicit, rather but that's what pthread does magically with `__pthread_register_cancel`.

Perhaps it's possible to handle `__pthread_register_cancel` specially as triggering a `longjmp` and unwinding the stack. Actually the `longjmp` can happen anywhere in a nested function call, but maybe that doesn't really matter? What's important is that there is something which unwinds the stack back.

sim642:
I learned today that GCC/Clang also have a `__cleanup__` attribute which gives C++ RAII-like features in C. And it's starting to be used in the Linux kernel: https://lwn.net/Articles/934679/.
It's more fine-grained than the functions listed in the issue (and requires block scope support!), but might become relevant if we try to analyze the modern kernel. Not handling them could easily make us very unsound: we just ignore the attribute and perhaps never consider some mutexes unlocked.

PR #865, opened 2022-10-24
Add definitions for library functions used in `cp` program

michael-schwarz:
Adds definitions for the library functions encountered during #855. The following definitions are still missing:

- [x] `atexit` part of (#799)
- [x] `error`
- [x] `fscanf`

@sim642 Is there currently support for varargs in your library DSL? We would need it here.

### DIFF
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -102,6 +102,8 @@ let glibc_desc_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__readlink_chk", unknown [drop "path" [r]; drop "buf" [w]; drop "len" []; drop "buflen" []]);
     ("__readlink_alias", unknown [drop "path" [r]; drop "buf" [w]; drop "len" []]);
     ("__overflow", unknown [drop "f" [r]; drop "ch" []]);
+    ("__ctype_get_mb_cur_max", unknown []);
+    ("__xmknod", unknown [drop "ver" []; drop "path" [r]; drop "mode" []; drop "dev" [r; w]]);
   ]
 
 let big_kernel_lock = AddrOf (Cil.var (Goblintutil.create_var (makeGlobalVar "[big kernel lock]" intType)))
@@ -119,8 +121,6 @@ let linux_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("acquire_console_sem", special [] @@ Lock { lock = console_sem; try_ = false; write = true; return_on_success = true });
     ("release_console_sem", special [] @@ Unlock console_sem);
     ("misc_deregister", unknown [drop "misc" [r_deep]]);
-    ("__ctype_get_mb_cur_max", unknown []);
-    ("__xmknod", unknown [drop "ver" []; drop "path" [r]; drop "mode" []; drop "dev" [r; w]]);
   ]
 
 (** Goblint functions. *)
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -20,14 +20,15 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("realloc", special [__ "ptr" [r; f]; __ "size" []] @@ fun ptr size -> Realloc { ptr; size });
     ("abort", special [] Abort);
     ("exit", special [drop "exit_code" []] Abort);
-    ("ungetc", unknown [drop "c" []; drop "stream" [w]]);
-    ("fscanf", unknown ((drop "stream" [w]) :: (drop "format" [r]) :: (VarArgs (drop' [w])))); (* TODO: How to specify varargs? *)
+    ("ungetc", unknown [drop "c" []; drop "stream" [r; w]]);
+    ("fscanf", unknown ((drop "stream" [r; w]) :: (drop "format" [r]) :: (VarArgs (drop' [w]))));
     ("__freading", unknown [drop "stream" [r]]);
     ("mbsinit", unknown [drop "ps" [r]]);
-    ("mbrtowc", unknown [drop "pwc" [w]; drop "s" [r]; drop "n" []; drop "ps" [w]]);
+    ("mbrtowc", unknown [drop "pwc" [w]; drop "s" [r]; drop "n" []; drop "ps" [r; w]]);
     ("iswspace", unknown [drop "wc" []]);
     ("iswalnum", unknown [drop "wc" []]);
     ("iswprint", unknown [drop "wc" []]);
+    ("rename" , unknown [drop "oldpath" [r]; drop "newpath" [r];]);
   ]
 
 (** C POSIX library functions.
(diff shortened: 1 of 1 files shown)


michael-schwarz:
This is against `widening_global_issue` by accident, but as that is likely to be merged soon, I don't think  cleanup is needed.

src/analyses/libraryFunctions.ml
```diff
@@ -20,6 +20,14 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("realloc", special [__ "ptr" [r; f]; __ "size" []] @@ fun ptr size -> Realloc { ptr; size });
     ("abort", special [] Abort);
     ("exit", special [drop "exit_code" []] Abort);
+    ("ungetc", unknown [drop "c" []; drop "stream" [w]]);
+    (* ("fscanf", unknown [drop "stream" [w]; drop "format" [r]]); (* TODO: How to specify varargs? *) *)
```
michael-schwarz, line 24: TODO

src/analyses/libraryFunctions.ml
```diff
@@ -50,6 +86,17 @@ let gcc_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_unreachable", special' [] @@ fun () -> if get_bool "sem.builtin_unreachable.dead_code" then Abort else Unknown); (* https://github.com/sosy-lab/sv-benchmarks/issues/1296 *)
     ("__assert_rtn", special [drop "func" [r]; drop "file" [r]; drop "line" []; drop "exp" [r]] @@ Abort); (* gcc's built-in assert *)
     ("__builtin_return_address", unknown [drop "level" []]);
+    (* ("error", unknown [drop "status" []; drop "errnum" []; drop "format" [w]]); (* TODO: varargs *) *)
```
michael-schwarz, line 89: TODO

sim642:
> Is there currently support for varargs in your library DSL? We would need it here.

Already thought of, here's a couple of examples:
https://github.com/goblint/analyzer/blob/24ad290e813aa334b7c89376548c81b9bc8229f8/src/analyses/libraryFunctions.ml#L48
https://github.com/goblint/analyzer/blob/24ad290e813aa334b7c89376548c81b9bc8229f8/unittest/analyses/libraryDslTest.ml#L21-L27

src/analyses/libraryFunctions.ml
```diff
@@ -29,6 +37,34 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("nl_langinfo", unknown [drop "item" []]);
+    ("nl_langinfo_l", unknown [drop "item" []; drop "locale" [r]]);
+    ("getc_unlocked", unknown [drop "stream" [w]]);
+    ("getchar_unlocked", unknown []);
+    ("putc_unlocked", unknown [drop "c" []; drop "stream" [w]]);
+    ("putchar_unlocked", unknown [drop "c" []]);
+    ("lseek", unknown [drop "fd" [w]; drop "offset" []; drop "whence" []]);
+    ("fseeko", unknown [drop "stream" [w]; drop "offset" []; drop "whence" []]);
+    ("iconv_open", unknown [drop "tocode" [r]; drop "fromcode" [r]]);
+    ("iconv", unknown [drop "cd" [r]; drop "inbuf" [r]; drop "inbytesleft" []; drop "outbuf" [w]; drop "outbytesleft" []]);
```
sim642, line 49: According to the man page, it decrements the `bytesleft` arguments, which is why they're `size_t*`. So they should be `r; w` I think.

src/analyses/libraryFunctions.ml
```diff
@@ -29,6 +37,34 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("iconv_close", unknown [drop "cd" [w]]);
```
sim642, line 50: It deallocates, so `f` instead?

src/analyses/libraryFunctions.ml
```diff
@@ -29,6 +37,34 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("strnlen", unknown [drop "s" [r]; drop "maxlen" []]);
+    ("chmod", unknown [drop "pathname" [r]; drop "mode" []]);
+    ("fchmod", unknown [drop "fd" [r]; drop "mode" []]);
+    ("fchown", unknown [drop "fd" [r]; drop "owner" []; drop "group" []]);
```
sim642, line 54: `fd` is `int`, so no pointer accesses.

src/analyses/libraryFunctions.ml
```diff
@@ -29,6 +37,34 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("lchown", unknown [drop "pathname" [r]; drop "owner" []; drop "group" []]);
+    ("clock_gettime", unknown [drop "clockid" []; drop "tp" [w]]);
+    ("gettimeofday", unknown [drop "tv" [w]; drop "tz" [w]]);
+    ("futimens", unknown [drop "fd" [w]; drop "times" []]);
```
sim642, line 58: `fd` is `int`. `times` is an array, so `r`.

src/analyses/libraryFunctions.ml
```diff
@@ -29,6 +37,34 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("utimes", unknown [drop "filename" [r]; drop "times" []]);
```
sim642, line 59: `times` is an array, so `r`.

src/analyses/libraryFunctions.ml
```diff
@@ -29,6 +37,34 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("linkat", unknown [drop "olddirfd" []; drop "oldpath" [r]; drop "newdirfd" []; drop "newpath" [r]; drop "flags" []]);
+    ("dirfd", unknown [drop "dirp" [r]]);
+    ("fdopendir", unknown [drop "fd" []]);
+    ("pathconf", unknown [drop "path" [r]; drop "name" []]);
+    ("rename" , unknown [drop "oldpath" [r]; drop "newpath" [r];]);
+    ("symlink" , unknown [drop "oldpath" [r]; drop "newpath" [r];]);
+    ("ftruncate", unknown [drop "fd" [r]; drop "length" []]);
```
sim642, line 66: `fd` is `int`.

src/analyses/libraryFunctions.ml
```diff
@@ -50,6 +86,17 @@ let gcc_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_unreachable", special' [] @@ fun () -> if get_bool "sem.builtin_unreachable.dead_code" then Abort else Unknown); (* https://github.com/sosy-lab/sv-benchmarks/issues/1296 *)
     ("__assert_rtn", special [drop "func" [r]; drop "file" [r]; drop "line" []; drop "exp" [r]] @@ Abort); (* gcc's built-in assert *)
     ("__builtin_return_address", unknown [drop "level" []]);
+    ("error", unknown ((drop "status" []):: (drop "errnum" []) :: (drop "format" [w]) :: (VarArgs (drop' [r]))));
+    ("fputs_unlocked", unknown [drop "s" [r]; drop "stream" [w]]);
```
sim642, line 90: Why is this `unlocked` function here, while others are above in POSIX?

sim642:
I'm not sure if SV-COMP rules cover the POSIX stuff though, they're not exactly part of GCC itself, but glibc.

src/analyses/libraryFunctions.ml
```diff
@@ -67,6 +114,8 @@ let linux_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("acquire_console_sem", special [] @@ Lock { lock = console_sem; try_ = false; write = true; return_on_success = true });
     ("release_console_sem", special [] @@ Unlock console_sem);
     ("misc_deregister", unknown [drop "misc" [r_deep]]);
+    ("__ctype_get_mb_cur_max", unknown []);
+    ("__xmknod", unknown [drop "ver" []; drop "path" [r]; drop "mode" []; drop "dev" [r]]);
```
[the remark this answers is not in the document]
michael-schwarz: The problem is that these are Linux-specific, and not defined in the POSIX standard.

src/analyses/libraryFunctions.ml
michael-schwarz: Iirc, this one is a GCC thing, and not in the standard.

src/analyses/libraryFunctions.ml
```diff
@@ -50,6 +86,17 @@ let gcc_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_unreachable", special' [] @@ fun () -> if get_bool "sem.builtin_unreachable.dead_code" then Abort else Unknown); (* https://github.com/sosy-lab/sv-benchmarks/issues/1296 *)
     ("__assert_rtn", special [drop "func" [r]; drop "file" [r]; drop "line" []; drop "exp" [r]] @@ Abort); (* gcc's built-in assert *)
     ("__builtin_return_address", unknown [drop "level" []]);
+    ("error", unknown ((drop "status" []):: (drop "errnum" []) :: (drop "format" [w]) :: (VarArgs (drop' [r]))));
+    ("fputs_unlocked", unknown [drop "s" [r]; drop "stream" [w]]);
+    ("gettext", unknown [drop "msgid" [r]]);
+    ("futimesat", unknown [drop "dirfd" [w]; drop "pathname" [r]; drop "times" []]);
```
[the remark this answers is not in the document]
michael-schwarz: Iirc, this one is a GCC thing, and not in the standard.

src/analyses/libraryFunctions.ml
sim642: Right, they're all on the same man page, but I didn't look at the `CONFORMING TO` section carefully enough.
I think the appropriate thing to do with this and the similar ones below is to add a new spec list  for glibc, for things that glibc provides, but are not in C/POSIX standard nor GCC builtins.

src/analyses/libraryFunctions.ml
```diff
@@ -29,6 +37,34 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_bzero", special [__ "dest" [w]; __ "count" []] @@ fun dest count -> Bzero { dest; count; });
     ("explicit_bzero", special [__ "dest" [w]; __ "count" []] @@ fun dest count -> Bzero { dest; count; });
     ("__explicit_bzero_chk", special [__ "dest" [w]; __ "count" []; drop "os" []] @@ fun dest count -> Bzero { dest; count; });
+    ("nl_langinfo", unknown [drop "item" []]);
+    ("nl_langinfo_l", unknown [drop "item" []; drop "locale" [r]]);
```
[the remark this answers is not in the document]
jerhard: At least for glibc, it is a pointer to a struct, see the `__locale_t` [here](https://github.com/lattera/glibc/blob/master/locale/bits/types/__locale_t.h).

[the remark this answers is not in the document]
sim642: Thanks for looking that up! Since the struct contains other pointers that the implementation probably uses in some way, `r_deep` would be the safe thing to do.

src/analyses/libraryFunctions.ml
jerhard: It's a non-standard extension of POSIX that is supported by `glibc`, see the [Conforming To](https://linux.die.net/man/3/fputs_unlocked) here.

jerhard: It's a non-standard extension of POSIX that is supported by `glibc`, see the [Conforming To](https://linux.die.net/man/3/fputs_unlocked) here.

src/analyses/libraryFunctions.ml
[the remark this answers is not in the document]
jerhard: It's a non-standard extension of POSIX that is supported by `glibc`, see the [Conforming To](https://linux.die.net/man/2/futimesat) here.

src/analyses/libraryFunctions.ml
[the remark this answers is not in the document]
jerhard: The `dev` parameter does not have a `const` modifier (see [here](https://refspecs.linuxbase.org/LSB_3.1.0/LSB-generic/LSB-generic/baselib---xmknod-1.html), so I am not sure whether it's read-only.

sim642 (approved):

Issue #868, opened 2022-10-26
Missing library functions for sv-comp


michael-schwarz:
Definitions for the following library functions are missing:

**Floats**

- [x] `__builtin_uaddl_overflow` / `__builtin_uadd_overflow` / `__builtin_umull_overflow` Not floats, but still (#869)
- [x] `__isnan`, `__signbit`, `__signbitf`, `_isinff`, `__isinf`, `__isnanf`, `__finite` (#869)
- [x] `fegetround` / `fesetround` (#869)
- [x] `__fpclassify` / `__fpclassifyl` / `__fpclassifyf`  (#869)
- [x] `__builtin_huge_val` /  `__builtin_huge_valf` (#869)
- [x] `__builtin_isgreater` / `__builtin_isgreaterequal` / `__builtin_isless` / `__builtin_islessequal` / `__builtin_islessgreater` / `__builtin_isunordered` (#869)

**Easy to add**

- [x] `pthread_attr_destroy` #878
- [x] `puts` #878
- [x] `strspn` / `strcspn` #878
- [x] `strtoll` / `strtoull` / `strtod` #878
- [x] `mktime` #878
- [x] `clearerr` #878
- [x] `ctime` #878
- [x] `setbuf` #878
- [x] `ntohs` #878
- [x] `__builtin_popcountl` #878
- [x] `swprintf` #878

**Driver / Kernel internals**

- [x] `b1pcmcia_addcard_m2` /  `b1pcmcia_addcard_m1` /  `b1pcmcia_addcard_b1`  (Reported upstream https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1355)
- [x] `__bad_percpu_size` / `__bad_size_call_parameter` / `__xchg_wrong_size` / `__cmpxchg_wrong_size` / `__xadd_wrong_size` / `__put_user_bad` #878
- [x] `fw_card_release`  (Reported upstream https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1355)
- [x] `usbnet_read_cmd_nopm` / `usbnet_write_cmd_nopm` / `usbnet_write_cmd` / `usbnet_read_cmd`  (Reported upstream https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1355)
- [x] `em28xx_free_device`  (Reported upstream https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1355)
[...]

PR #869, opened 2022-10-27
Equip float domain to handle definite NaN, +Inf, -Inf, and some missing float stubs

michael-schwarz:
In the course of #868 I came across a few tests for sv-comp that are low-hanging fruit that require handling the information that some variable is definitely NaN, +Inf or -Inf. This adds these possibilities.
The resulting lattice is a flat lattice with NaN, +Inf, -Inf, and intervals of actual floats values.

In the course of this PR, I will also add the float stubs sv-comp uses that I identified in #868.


tests/regression/57-floats/15-more-library.c
```diff
@@ -0,0 +1,64 @@
...
+
+    double c = ceil(-0.0);
+    __goblint_check((c == -0.0) && signbit(c)); //TODO (We do not distinguish +0 and -0)
+
+    c = ceil(-INFINITY);
+
+    // On OS X this gets expanded differently than on Linux where it is equivalent to the one below
+    // Might make sense to check what is needed for OS X support in the future, but this is not a deal-breaker
+    // and not high priority for now.
+    // __goblint_check(isinf(INFINITY) && signbit(c));
```
sim642, line 42: So this (and the one below and in 57/17) could just be `// TODO` or does that cause bigger problems than just one check not passing?

sim642 (commented):
I suppose the definite NaN/+inf/-inf domain could be extracted out to a separate float domain and be another component of `FloatDomTuple`, which could decouple things somewhat. But since it's already done, there's no point in redoing it right now.

tests/regression/57-floats/15-more-library.c
michael-schwarz: Good point!

src/analyses/libraryFunctions.ml
```diff
@@ -145,21 +169,57 @@ let zstd_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     Functions and builtin versions of function and macros defined in math.h. *)
 let math_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_nan", special [__ "str" []] @@ fun str -> Math { fun_args = (Nan (FDouble, str)) });
+    ("nan", special [__ "str" []] @@ fun str -> Math { fun_args = (Nan (FDouble, str)) });
```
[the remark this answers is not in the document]
michael-schwarz: Yes, I added it now.

sim642 (approved):

PR #871, opened 2022-10-28
var_eq unsoundness with dereferenced floats

sim642:
In dc17de5f0da6df7299df890b67cca4823871dc97 and a889df0be8621f0c63601f2cc8b347c5d512075f some fixes were made to var_eq to make it sound w.r.t. floats.

I was thinking about deduplicating this hand-written code in favor of #840:
https://github.com/goblint/analyzer/blob/790b3a93d1577d8d3471c89d0798f9f7506cfcbc/src/analyses/varEq.ml#L66-L110
The previous fixes are _ad hoc_ in the middle of this otherwise manual version of `CilType.Exp.equal`, which makes reuse impossible.
While wondering, why those fixes were made where they were (for `varinfo` and for `fieldinfo` comparison), I realized that those are insufficient, e.g. with pointers to floats like `*p == *p`.

I think we should change that implementation to use `CilType.Exp.equal` directly, but have some completely separate check for float problems, which is complete. I'm not precisely sure what that would be, maybe check if any subexpression has `float` type? Other ideas?

**EDIT:** I implemented it now and it seems to work at least.

### DIFF
--- a/src/util/cilfacade.ml
+++ b/src/util/cilfacade.ml
@@ -8,8 +8,13 @@ module GU = Goblintutil
 include Cilfacade0
 
 (** Is character type (N1570 6.2.5.15)? *)
-let isCharType = function
+let isCharType = function (* TODO: also unrollType here? *)
   | TInt ((IChar | ISChar | IUChar), _) -> true
+  | _ -> false
+
+let isFloatType t =
+  match Cil.unrollType t with
+  | TFloat _ -> true
   | _ -> false
 
 let init_options () =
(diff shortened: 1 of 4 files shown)


src/util/cilfacade.ml
```diff
@@ -8,10 +8,15 @@ module GU = Goblintutil
 include Cilfacade0
 
 (** Is character type (N1570 6.2.5.15)? *)
-let isCharType = function
+let isCharType = function (* TODO: also unrollType here? *)
```
michael-schwarz, line 11: Probably. The only usage is in `cast_addr`. Might be worth trying what happens if we just do it.

michael-schwarz (approved):

src/util/cilfacade.ml
sim642: Did it here, no tests break at least. Not unrolling the type in such cases probably just makes us less precise at these checks if typedefs are used instead.

PR #874, opened 2022-10-30
Check `space` option for TD3 and removal of wpoints & `sides-local` widening strategy are not enabled at the same time

### LINKED ISSUE #867: `space` option for TD3 incompatible with removal of wpoints & `sides-local` widening strategy
Helmut observed that those do not work together, as `space` stores values only for things that are in `point`.
We should include a check that these configuration options are mutually exclusive.

michael-schwarz:
Closes #867.

### DIFF
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -122,7 +122,7 @@ and complete args =
 let eprint_color m = eprintf "%s\n" (MessageUtil.colorize ~fd:Unix.stderr m)
 
 let check_arguments () =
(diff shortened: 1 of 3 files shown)


sim642 (approved):
At last GobCon I mentioned that `space` might work work `remove-wpoint` because the latter doesn't remove abstract values from `rho`. Did you come to any conclusion about it or are we just forbidding it for now to be on the safe side?
I guess `space` needs more rethinking anyway.

michael-schwarz:
>  Did you come to any conclusion about it or are we just forbidding it for now to be on the safe side?

We decided to just do this for safety reasons for now, and reconsider the details at some later point.

PR #875, opened 2022-10-31
Add `abortUnless` analysis

### LINKED ISSUE #873: Support for `void assume_abort_if_not(int cond) {   if(!cond) {abort();} }`
The sv-comp people are really fond of encoding conditions that are needed to later establish correctness into calls to this function. We do not benefit from any such information at all, which we should probably fix for this year's SV-Comp.

~~~C
void assume_abort_if_not(int cond) {
  if(!cond) {abort();}
}
~~~

There are ~3000 instances of this pattern in the non-cwe part of the sv-comp `.c` files, and for some folders (e.g. `weaver`) they seem to have encoded everything into such calls :upside_down_face:

~~~C
  assume_abort_if_not(counter != 0);
  reach_error();
~~~

What I would envision would be a very cheap analysis to determine if a function is such an `assume_abort_if_not_function`:

- Takes one argument of type `int`
- Contains no assigns
- Returns only for non-zero values

If this is the case, I would meet the caller state with the condition that is passed as the argument.

michael-schwarz:
Adds an analysis that detects if control flow only returns if the argument is `non-zero`, and issues a corresponding `assert` event.

TODO:

- [x] Exploit in more analyses
- [x] Test

Closes #873


michael-schwarz:
There are no other analysis where adding it would make sense (except maybe `vareq`, but that doesn't even exploit `branch` at all currently).
Adding this analysis to the set of default analyses does not cause any regressions.

michael-schwarz:
Unless anyone opposes, I would propose we do a joint sv-comp run of this & #869 to reduce the number of needed runs, as we currently have a lot of things we would need server time for.

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -618,6 +618,8 @@ struct
       Priv.enter_multithreaded (Analyses.ask_of_ctx ctx) ctx.global ctx.sideg st
     | Events.Escape escaped ->
       Priv.escape ctx.node (Analyses.ask_of_ctx ctx) ctx.global ctx.sideg st escaped
+    | Assert exp ->
+      branch ctx exp true
```
sim642, line 622: `assert_fn` would be more appropriate.

src/analyses/base.ml
```diff
@@ -2810,6 +2810,8 @@ struct
     | Events.AssignSpawnedThread (lval, tid) ->
       (* TODO: is this type right? *)
       set ~ctx (Analyses.ask_of_ctx ctx) ctx.global ctx.local (eval_lv (Analyses.ask_of_ctx ctx) ctx.global ctx.local lval) (Cilfacade.typeOfLval lval) (`Thread (ValueDomain.Threads.singleton tid))
+    | Events.Assert exp ->
+      branch ctx exp true
```
sim642, line 2814: `assert_fn` would be more appropriate.

src/analyses/abortUnless.ml
```diff
@@ -0,0 +1,70 @@
...
+      match args with
+      | [arg] -> ctx.emit (Events.Assert arg)
+      | _ -> ()
+    );
+    false
+
+  let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    let desc = LibraryFunctions.find f in
+    match desc.special arglist with
+    | Abort -> ctx.local
```
sim642, line 60: `Abort` is deadcode anyway, so this doesn't matter at all.

src/analyses/abortUnless.ml
```diff
@@ -0,0 +1,70 @@
...
+  module C = Lattice.Unit
+
+  let context _ _ = ()
+
+  (* transfer functions *)
+  let assign ctx (lval:lval) (rval:exp) : D.t =
+    false
+
+  let branch ctx (exp:exp) (tv:bool) : D.t =
+    ctx.local
```
sim642, line 21: Any branch? Not just `arg` or `!arg`?

src/analyses/abortUnless.ml
```diff
@@ -0,0 +1,70 @@
...
+           | None -> false)
+      | _ ->
+        (* should not happen, ctx.local should always be false in this case *)
+        false
+    else
+      false
+
+  let enter ctx (lval: lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
+    let candidate = match f.sformals with
+      | [arg] when isIntegralType arg.vtype -> true
```
sim642, line 44: This allows calls with similar signature to be called within `assume_abort_if_not`, including (mutually) recursively. It's not obvious why something bad cannot come from that.

src/analyses/abortUnless.ml
michael-schwarz: Yes, branching does not actually modify the value here, so any branching is fine.

src/analyses/abortUnless.ml
michael-schwarz: Nice catch! It's not the recursion though that was problematic, as we go to `false` in `combine` either way but the fact that the assert happens after evaluation of the transfer function for call, so if variables in `arg` are assigned to, asserting might unsoundly yield bot (as in the case you constructed).

michael-schwarz:
For sv-comp this analysis currently does not seem to pay off, so I would not include it in the config this year (unless I can come up with a more powerful analysis where we would benefit).
I would still propose to merge it, it does not add much code and might be handy.

Issue #876, opened 2022-11-01
Support for C11 thread-local variables / `pthread_getspecific` & `pthread_setspecific`


michael-schwarz:
The notion of C11 thread-local variables seems to map onto `pthread_getspecific` & `pthread_setspecific`. So when adding support for C11, support for this should be added to.

sim642:
Besides C11's thread-local specifier there is also GCC's [`__thread`](https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html) that we found with @vesalvojdani in silver searcher. It should be very easy to just ignore races on thread-local things.

michael-schwarz:
I think the good first issue applies mostly to not reporting races for these variables, the support to privatize the values is a bit trickier - one has to think a bit what happens when thread ids are non-unique (it would seem like that should also work, but some thought is needed here).

PR #877, opened 2022-11-01
First approximation of `pthread_setspecific` by escaping things reachable from the argument and invalidating all globals

michael-schwarz:
Our analysis currently is sound here, as the second argument to the function is considered escaped, and treated flow-insensitively from there on. Also, it is invalidated when `sem.unknown_function.invalidate.globals` is set.

Any later writes thus do not lead to the value being changed back from `T` and thus all writes through the pointer obtained back from `pthread_getspecific` are also accounted for.

This adds an assert that fails if the code makes use of `pthread_setspecific` and ``sem.unknown_function.invalidate.globals` is disabled, leading to unsound results.

This a temporary measure before #876 is attacked.

### DIFF
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -75,6 +75,9 @@ let pthread_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("pthread_cond_broadcast", special [__ "cond" []] @@ fun cond -> Broadcast cond);
     ("pthread_cond_wait", special [__ "cond" []; __ "mutex" []] @@ fun cond mutex -> Wait {cond; mutex});
     ("pthread_cond_timedwait", special [__ "cond" []; __ "mutex" []; __ "abstime" [r]] @@ fun cond mutex abstime -> TimedWait {cond; mutex; abstime});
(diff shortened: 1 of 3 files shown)


src/analyses/libraryFunctions.ml
```diff
@@ -346,6 +346,10 @@ let classify fn exps: categories =
   | "pthread_mutex_unlock" | "__pthread_mutex_unlock" | "up_read" | "up_write"
   | "up" | "pthread_spin_unlock"
     -> `Unlock
+  | "pthread_setspecific" as x ->
+    (if not (GobConfig.get_bool "sem.unknown_function.invalidate.globals") then
+       failwith "pthread_setspecific is only analyzed soundly when sem.unknown_function.invalidate.globals is set.");
+    `Unknown x
```
sim642, line 352: I think it could just add it in the new style as
```ocaml
unknown ~attrs:[InvalidateGlobals] ...
```

michael-schwarz: Is there a way to specify that the arguments needs to be considered escaped though?

sim642: We haven't generalized thread escapes like that. The escape analysis explicitly does it for thread creation, so the easiest solution is to just add a case for this special function as well.

sim642 (approved):

PR #878, opened 2022-11-01
Add some further library functions

michael-schwarz:
References #868

### DIFF
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -38,7 +38,8 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("strtoul", unknown [drop "nptr" [r]; drop "endptr" [w_deep]; drop "base" []]);
     ("strtoull", unknown [drop "nptr" [r]; drop "endptr" [w_deep]; drop "base" []]);
     ("mktime", unknown [drop "tm" [r;w]]);
-    ("ctime", unknown [drop "rm" [r]])
+    ("ctime", unknown [drop "rm" [r]]);
+    ("clearerr", unknown [drop "stream" [w]]);
   ]
 
 (** C POSIX library functions.
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -32,6 +32,11 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("puts", unknown [drop "s" [r]]);
     ("strspn", unknown [drop "s" [r]; drop "accept" [r]]);
     ("strcspn", unknown [drop "s" [r]; drop "accept" [r]]);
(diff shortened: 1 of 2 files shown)


src/analyses/libraryFunctions.ml
```diff
@@ -29,6 +29,19 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
     ("iswprint", unknown [drop "wc" []]);
     ("rename" , unknown [drop "oldpath" [r]; drop "newpath" [r];]);
+    ("puts", unknown [drop "s" [r]]);
+    ("strspn", unknown [drop "s" [r]; drop "accept" [r]]);
+    ("strcspn", unknown [drop "s" [r]; drop "accept" [r]]);
+    ("strtod", unknown [drop "nptr" [r]; drop "endptr" [w_deep]]);
+    ("strtol", unknown [drop "nptr" [r]; drop "endptr" [w_deep]; drop "base" []]);
+    ("strtoll", unknown [drop "nptr" [r]; drop "endptr" [w_deep]; drop "base" []]);
+    ("strtoul", unknown [drop "nptr" [r]; drop "endptr" [w_deep]; drop "base" []]);
+    ("strtoull", unknown [drop "nptr" [r]; drop "endptr" [w_deep]; drop "base" []]);
```
sim642, line 39: I think these functions only change which character `endptr` points to (`w`) but don't modify those pointed to characters, which would be a double dereference (`w_deep`). So `**endptr` should not be modified because it's a pointer into the constant `nptr` string.

src/analyses/libraryFunctions.ml
```diff
@@ -128,6 +146,12 @@ let glibc_desc_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__overflow", unknown [drop "f" [r]; drop "ch" []]);
     ("__ctype_get_mb_cur_max", unknown []);
     ("__xmknod", unknown [drop "ver" []; drop "path" [r]; drop "mode" []; drop "dev" [r; w]]);
+    ("__bad_percpu_size",  special' [] @@ fun () -> Abort); (* these do not have defintions so the linker will fail if they are actually called *)
+    ("__bad_size_call_parameter",  special' [] @@ fun () -> Abort);
+    ("__xchg_wrong_size",  special' [] @@ fun () -> Abort);
+    ("__cmpxchg_wrong_size",  special' [] @@ fun () -> Abort);
+    ("__xadd_wrong_size",  special' [] @@ fun () -> Abort);
+    ("__put_user_bad",  special' [] @@ fun () -> Abort);
```
sim642, line 154: These all seem to be part of Linux kernel implementation, not glibc, so they should be in the following list. Since these lists are currently conditional, they will work for SV-COMP there as well, although these `abort()` implementations really should be in sv-benchmarks.

Also, there's no need for the functional form, could just be `special [] Abort`.

sim642 (approved):

Issue #879, opened 2022-11-02
Some observations from SV-comp runs (To be split into separate issues)


michael-schwarz:
I made a run of all of sv-comp with 15min timeout and our config and here are some observations that we might want to think about:

**Resolved**

- ~~31 tasks (all with expected verdict true) fail after the actual analysis is already over and the verdict printout is produced with the exception~~ (fixed by #882)

- ~~2 instances where the analysis succeeds, but some time later a Not_found exception is produced~~ (fixed by #882)

**PR exists**

- ~~A lot of recursive runs still go out of memory despite the interval contexts being disabled. At the same time, apron is enabled though and I'm not sure if anything smart is done there to prevent the issue of infinite contexts happening within the apron domain then. (maybe @leunam99 can advise here?)~~ (no noticeable impact, but with #883 it is more consistent)

**Dedicated issue exists**

- Total of 11 instances of #118

**Open**

- We run out of memory on our own task https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/goblint-regression/04-mutex_05-lockfuns.c. It seems the issue is due to `mutex-meet` privatization being the default for octagons, and the autotuner enabling Apron. In this configuration, `side_widen` would need to be `sides-pp`. It seems there is only one such task though where we run out of memory though, so it probably is not much of an issue.

- Also for this task, autoconf enables octagons for the following variables: What seems odd is that it is activated for `__builtin_bswap64` and what these `val` variables for which it is also enabled. (#921)

> Enabled octagon domain for:
__builtin_bswap64, myglobal, id, __x, top, __x, __x, __x, i, key, count, a, tmp, count, i, j, i___0, j___0, k, size, r, val, val, val, val, val, val, val, val, val, val, val, val, val, val

- [ ] One test fails with `Apron.Manager.Error`: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/ldv-linux-3.16-rc1/43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--nfc--st21nfca--st21nfca_i2c.ko-entry_point.cil.out.c

[...]

PR #882, opened 2022-11-03
Do not deref into function or mutex in invariant

michael-schwarz:
This addresses the two cases from #879 where the analysis crashes after the result has already been obtained. Here, during witness generation, an attempt is made to construct an invariant about dereferencing functions or mutexes.

During this attempt, queries to unknowns that have not received any value are made, which then fail with `Not_found`. In this PR we avoid generating such invariants for values of `ignorable_types` and `functions`.


src/cdomains/valueDomain.ml
```diff
@@ -1239,30 +1239,33 @@ struct
...
                   Invariant.of_exp Cil.(BinOp (Eq, c_exp, addr_exp, intType))
                 else
                   Invariant.none
               in
               let i_deref =
+                (* Avoid dereferencing into functions, mutexes, ..., which will lead to errors as they are not added to the hash table *)
+                let noderef t = try isFunctionType t || Cilfacade.is_ignorable_type t with Not_found -> false in
                 match Cilfacade.typeOfLval (Var vi, offset) with
-                | typ ->
+                | typ when not (noderef vi.vtype) ->
```
sim642, line 1261: Why not check `typ`?

src/cdomains/valueDomain.ml
```diff
@@ -1239,30 +1239,33 @@ struct
...
                       addr_exp, c_exp
-                  in 
+                  in
                   Invariant.of_exp Cil.(BinOp (Eq, c_exp, addr_exp, intType))
                 else
                   Invariant.none
               in
               let i_deref =
+                (* Avoid dereferencing into functions, mutexes, ..., which will lead to errors as they are not added to the hash table *)
+                let noderef t = try isFunctionType t || Cilfacade.is_ignorable_type t with Not_found -> false in
```
sim642, line 1259: Where is this `Not_found` even coming from? It's a boolean function so it might as well handle that itself instead of requiring an extra try at every use.

Also, I think it should be just using `ValueDomain.is_immediate_type`.

src/util/cilfacade.ml
```diff
@@ -7,6 +7,23 @@ module GU = Goblintutil
...
+  | TComp ({ cname = "lock_class_key"; _ }, _) -> true
+  | TInt (IInt, attr) when hasAttribute "mutex" attr -> true
+  | t when hasAttribute "atomic" (typeAttrs t) -> true (* C11 _Atomic *)
+  | _ -> false
+
+let is_ignorable = function
+  | None -> false
+  | Some (v,os) ->
+    try isFunctionType v.vtype || is_ignorable_type v.vtype
+    with Not_found -> false
```
sim642, line 25: These are specific to race detection, which is why they were in `Access`. For example, base analysis still handles variables with `_Atomic` instead of ignoring them completely.

src/cdomains/valueDomain.ml
michael-schwarz: Done!

src/util/cilfacade.ml
michael-schwarz: Done

michael-schwarz:
We identified some sort of root cause that also highlights why the fallback to `bot` in case of `Not_found` in the `ctx` for witnesses is needed.

https://github.com/goblint/analyzer/blob/4f8d9b3eef7f79b20f33ac1d4a8b61992a322ef7/tests/regression/29-svcomp/29-witness.c#L1-L19

Here, the global is never accessed, so no information on accesses and protecting mutexes is collected. Then, when we try to build the witness, we access this information for the first time. During the analysis this would simply yield bottom, so everything is fine here as well.

tests/regression/29-svcomp/30-witness2.c
```diff
@@ -0,0 +1,4536 @@
+// SKIP PARAM: --enable ana.sv-comp.enabled --set ana.specification "CHECK( init(main()), LTL(G ! call(reach_error())) )"
+// SV-COMP marshaling doesn't work
```
sim642, line 2: If it's skipped, then is there any reason to dump a ~4500 LoC program into our _regression_ suite? If there's already a minimized case for it, then that should be enough and we regularly run on SV-COMP programs anyway.

michael-schwarz: Yeah, skipping it is by mistake, and I guess we can do without it if we run sv-comp often enough. I'll remove it again.

sim642 (approved):

michael-schwarz:
This fixed the following things from #879:

- 31 tasks (all with expected verdict true) fail after the actual analysis is already over and the verdict printout is produced with the exception

> Bug: typeOffset: Index on a non-array
> [Error][Analyzer] About to crash!

bug_after_success

-2 instances where the analysis succeeds, but some time later a Not_found exception is produced

not_found

PR #883, opened 2022-11-04
Make autotuner set `apron.no-context` for functions where it sets `base.no-interval`

michael-schwarz:
The option is there to prevent a blow-up of contexts because of intervals. When apron is enabled, the same blowup could still happen because of it. Thus, this disables the apron contexts for those functions where the autotune turns of intervals in contexts.


sim642 (approved):
I suppose this fixes the sv-benchmarks timeouts that are mentioned in #879?

michael-schwarz:
It fixes some that I tested with.

I don't know yet, I will have to run sv-comp for that again. (Currently I am running with keeping 10 allocation sites separately to see if it is worth autotuning that).

michael-schwarz:
There is a slight positive impact (1 additional task correct), but the behavior we get with this PR is also the less surprising one, so I am merging this now.

PR #884, opened 2022-11-04
Add library functions for `__atomic_{store,load}_n`

michael-schwarz:
Calls occur in sv-comp. This is a stop-gap measure before we get proper support for C11 in Goblint. References #868 

(These are indeed builtins despite their name not starting with `__builtin`: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html)

### DIFF
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -132,6 +132,8 @@ let gcc_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_popcount", unknown [drop "x" []]);
     ("__builtin_popcountl", unknown [drop "x" []]);
     ("__builtin_popcountll", unknown [drop "x" []]);
+    ("__atomic_store_n", unknown [drop "ptr" [w]; drop "val" []; drop "memorder" []]);
+    ("__atomic_load_n", unknown [drop "ptr" [w]; drop "memorder" []]);
   ]
 
 let glibc_desc_list: (string * LibraryDesc.t) list = LibraryDsl.[
--- a/tests/regression/41-stdlib/06-atomic.c
+++ b/tests/regression/41-stdlib/06-atomic.c
@@ -1,0 +1,11 @@
+int g = 8;
+
+int main() {
+    int i;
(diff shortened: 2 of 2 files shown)


sim642:
Why can't we just do actual assigning and reading of the pointed variable?

tests/regression/41-stdlib/06-atomic.c
```diff
@@ -0,0 +1,11 @@
+int g = 8;
+
+int main() {
+    int i;
+    __atomic_store_n (&i, 12, __ATOMIC_RELAXED);
+    int i = __atomic_load_n (&i, __ATOMIC_RELAXED);
```
sim642, line 6: This doesn't even compile because of the variable redefinition in the same block.

michael-schwarz:
> Why can't we just do actual assigning and reading of the pointed variable?

We could, but tbh we need proper support at some point anyway, and this is only a stop-gap until then.


> There are people who want to start using proper atomics in SV-COMP as well: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1331.

Yes, i also saw this. Though I am not sure what effect surrounding some accesses to a global with these while leaving others unaffected does. If such things actually are committed past the freeze deadline, I think it is fine to be cautious here for now, until we have proper support.

src/analyses/libraryFunctions.ml
```diff
@@ -132,6 +132,8 @@ let gcc_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_popcount", unknown [drop "x" []]);
     ("__builtin_popcountl", unknown [drop "x" []]);
     ("__builtin_popcountll", unknown [drop "x" []]);
+    ("__atomic_store_n", unknown [drop "ptr" [w]; drop "val" []; drop "memorder" []]);
+    ("__atomic_load_n", unknown [drop "ptr" [w]; drop "memorder" []]);
```
sim642, line 136:
```suggestion
    ("__atomic_load_n", unknown [drop "ptr" [r]; drop "memorder" []]);
```

sim642 (approved):

PR #886, opened 2022-11-04
Add missing `__VERIFIER_nondet_X()`s

michael-schwarz:
Adds the ones missing according to #868, and `__VERIFIER_nondet_int128` which will be added to the rules.

https://github.com/goblint/analyzer/blob/0e49d1ff1727ebfd647eb1227e058d015bcb5785/lib/sv-comp/stub/src/sv-comp.c#L25
I think the issue was that the typedef for `loff_t` only comes in the benchmark itself and might depend on the architecture defined for the benchmark. Although maybe all the Linux tasks have the same architecture, so we can hardcode the specific primitive type.

_Originally posted by @sim642 in https://github.com/goblint/analyzer/issues/868#issuecomment-1292059907_
 
 
 I went with simply specifying this one via library functions to invalidate now, that should side-step this issue.


sim642 (approved):
I added some comments for the ones that the SV-COMP rules don't actually specify.

PR #890, opened 2022-11-06
Do not produce overflow warnings from Apron & Unroll typedefs for `invariant`

### LINKED ISSUE #889: Enabling `apron` leads to additional spurious overflow warnings for `long long` type
@sim642 observed poor performance of Goblint for relatively simple Juliet tasks in #879. Having a closer look, the issue seems to be that enabling the apron analysis (with octagons) on top of intervals somehow leads to more overflow warnings.

~~~C
// PARAM: --enable ana.sv-comp.enabled --enable ana.int.interval --set ana.activated[+] apron --set ana.specification "CHECK( init(main()), LTL(G ! overflow) )"
#include<stdio.h>
typedef long long int64_t;

int main(int argc, char * argv[])
{
    int64_t data;
    data = 0LL;
    fscanf (stdin, "%" "l" "d", &data);
    if (data < 0x7fffffffffffffffLL)
    {
        int64_t result = data + 1;
    }

    return 0;
}
~~~

Here, we produce overflow warnings for both the comparison and the assignment. If apron is disabled, we succeed in verifying the absence of overflows.

michael-schwarz:
Apron sometimes will produce answers that are imprecise in response to EvalInt queries, which then get normalized to top if they do not fit within the specified `ikind`. Currently, we emit an overflow warning in such cases. However, this is not necessary: If there is an actual overflow, the interval domain will emit the required warnings.

The old behavior meant that enabling Apron could lead to spurious overflow warnings. (Closes #889)

Additionally, we were not unrolling `TNamed` in `invariant` which lead to precision loss when an alias for an integer type is used in an invariant.

@sim642: Could you maybe run this with the subset of the Juliet suite you mentioned in #879 to see if there are improvements, if you have capacity in Tartu? The Munich server is at capacity these days.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2069,7 +2069,7 @@ struct
          | _ -> failwith "unreachable")
       | Const _ , _ -> st (* nothing to do *)
       | CastE ((TFloat (_, _)), e), `Float c ->
-        (match Cilfacade.typeOf e, FD.get_fkind c with
+        (match unrollType (Cilfacade.typeOf e), FD.get_fkind c with
          | TFloat (FLongDouble as fk, _), FFloat
          | TFloat (FDouble as fk, _), FFloat
          | TFloat (FLongDouble as fk, _), FDouble
@@ -2082,7 +2082,7 @@ struct
         (match eval e st with
          | `Int i ->
            if ID.leq i (ID.cast_to ik i) then
-             match Cilfacade.typeOf e with
+             match unrollType (Cilfacade.typeOf e) with
              | TInt(ik_e, _)
              | TEnum ({ekind = ik_e; _ }, _) ->
                let c' = ID.cast_to ik_e c in
--- a/src/cdomains/apron/apronDomain.apron.ml
+++ b/src/cdomains/apron/apronDomain.apron.ml
@@ -936,8 +936,8 @@ struct
       else
         match eval_interval_expr d e with
         | (Some min, Some max) -> ID.of_interval ~suppress_ovwarn:true ik (min, max)
-        | (Some min, None) -> ID.starting ik min
-        | (None, Some max) -> ID.ending ik max
+        | (Some min, None) -> ID.starting ~suppress_ovwarn:true ik min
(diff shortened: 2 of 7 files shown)


sim642:
> @sim642: Could you maybe run this with the subset of the Juliet suite you mentioned in #879 to see if there are improvements, if you have capacity in Tartu? The Munich server is at capacity these days.

I started a run now.

> Apron sometimes will produce answers that are imprecise in response to EvalInt queries, which then get normalized to top if they do not fit within the specified `ikind`

This sounds like a bug on the Apron analysis side, no?

michael-schwarz:
> This sounds like a bug on the Apron analysis side, no?

Not necessarily, it can happen that apron is less precise than intervals (e.g. tracking less variables, ...). In this case some of its computation may trigger overflow warnings, but the interval domain that is more precise in this case can exclude the overflow.

sim642:
The ikind should still be right and we take the meet for queries, so we should still be as precise as the most precise one.

sim642:
> I started a run now.

This solves another 680, so we're at 2736/3078 on those.

The remaining 342 are all the ones with "square" in the name, involving code like
```c
abs(data) < (long)sqrt((double)0x7fffffff)
```
where we don't support `abs` (for integers) and `sqrt` (for floats)

michael-schwarz:
> The ikind should still be right and we take the meet for queries, so we should still be as precise as the most precise one.

Yes, but if one of the intermediate values overflows, we set the overflow flag in `ID.of_interval`, ` ID.starting` and `ID.ending` already.

sim642:
> Yes, but if one of the intermediate values overflows, we set the overflow flag in `ID.of_interval`, ` ID.starting` and `ID.ending` already before these changes.

Oh, right, the ones in Apron analysis. This really shows how dirty the quickly added no overflow analysis is, hardcoded straight into the interval domain and thus potentially triggered by any internal computation as well, instead of being generic over `EvalInt` result only or something like that.

sim642 (approved):

PR #891, opened 2022-11-08
Clean up warrowing box

sim642:
Currently the constraint system has a `box` function, which is always defined to just be warrowing, and at one point that is passed on via the `box` argument of `solve` in solvers.
This is redundant because solvers then have access to `box` in two different ways. Also the fixed box doesn't need to be carried around everywhere.

Therefore, this removes `box` both from the constraint system and solver argument.
SLR is the only solver that in some variation overrides `box` to just widen or narrow, making the argument also irrelevant. Since the choice of `box` is such a solver-specific thing, it makes the most sense to just do it directly in the solver in a way that's supported by the solver.


michael-schwarz:
@jerhard or @stilscher, could you have a look? You guys know more about the intricacies of the solver.

src/solvers/sLR.ml
```diff
@@ -500,27 +505,16 @@ module TwoPhased =
   functor (S:EqConstrSys) ->
   functor (HM:Hashtbl.S with type key = S.v) ->
   struct
-    include Make (V) (S) (HM)
-    let narrow = narrow S.Dom.narrow
-    let solve box is iv =
-      let sd = solve (fun _ x y -> S.Dom.widen x (S.Dom.join x y)) is iv in
+    module N = Make0 (V) (SolverBox.NarrowOption) (S) (HM)
```
jerhard, line 508: It seems that the only reference to the `narrow` defined in line 8 in this file, that was previously dealing with  "exp.no-narrow", is removed here. So the definition in line 8 may be removed.

jerhard (changes requested):

src/solvers/sLR.ml
sim642: It is still used with `open SLR`:
```
File "src/solvers/sLRterm.ml", line 23, characters 17-23:
23 |     let narrow = narrow S.Dom.narrow
                      ^^^^^^
Error: Unbound value narrow
File "src/solvers/sLRphased.ml", line 23, characters 17-23:
23 |     let narrow = narrow S.Dom.narrow
                      ^^^^^^
Error: Unbound value narrow
```

jerhard (approved):

PR #894, opened 2022-11-08
Handle overflows in congruence domain more precisely

### LINKED ISSUE #893: Keep information about congruences `mod 2^n` on unsigned overflow
Congruences `mod 2^n` with `n <= bitlength` persist for unsigned integer types even after an overflow.  Thus, the information should be kept. 

This information could e.g. useful for the  `/c/bitvector/jain_5-2.c`, `jain_6-1.c`, `jain_7-2.c` benchmarks.

jerhard:
This PR introduces more precise handling of unsigned overflows for congruences. 
When the integer (in the sense of whole number) result `n` of an operation would be known to be `n === k mod 2^a * b`, we can keep the information `n === k mod 2^a` (and possibly reduce k to normalize it), for an `a <= bitlength`.
As discussed with @michael-schwarz , this information still holds when cutting off the result in the whole numbers into the residue class ring `Z/2^{bitlength}`.  
Fixes #893.

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -2308,7 +2308,7 @@ struct
     res
 
   (* Handle unsigned overflows.
-     From n === k mod (2^a * b), we conclude n === k mod 2^a, for a <= 2^{bitwidth}.
+     From n === k mod (2^a * b), we conclude n === k mod 2^a, for a <= bitwidth.
      The congruence modulo b may not persist on an overflow. *)
   let handle_overflow ik (c, m) =
     if m =: Ints_t.zero then
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -2307,16 +2307,40 @@ struct
     if M.tracing then  M.trace "congruence" "shift_left : %a %a becomes %a \n" pretty x pretty y pretty res;
     res
 
+  let find_power_of_two n =
+    let two = Ints_t.of_int 2 in
+    let rec find n acc =
+      if n %: two =: Ints_t.zero then
+        find (n /: two) (two *: acc)
+      else
+        acc
+    in
+    find n Ints_t.one
+
+  (* From n === k mod (2^a * b), we conclude n === k mod 2^a,
+     for a <= 2^{bitwidth}.
+     The congruence modulo b may not persist on an overflow. *)
+  let handle_overflow ik (c, m) =
+    let max = (snd (Size.range ik)) +: Ints_t.one in
+    let m' = Ints_t.max max (find_power_of_two m) in
+    let res = normalize ik (Some (c, m')) in
+    res
+
   let mul ?(no_ov=false) ik x y =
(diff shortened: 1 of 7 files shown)


src/cdomains/intDomain.ml
```diff
@@ -2307,16 +2307,40 @@ struct
...
 
+  let find_power_of_two n =
+    let two = Ints_t.of_int 2 in
+    let rec find n acc =
+      if n %: two =: Ints_t.zero then
+        find (n /: two) (two *: acc)
+      else
+        acc
+    in
+    find n Ints_t.one
```
sim642, line 2318: What is this really computing? Finding the number of trailing zero bits?
If so, then this could be rather inefficient since there are CPU instructions for this sort of thing and `Z.trailing_zeros` could be a much more efficient way to compute that without all the multiplications and divisions.

jerhard: It computes the largest power of two that n is divisible by. Now this is computed with `Z.trailing_zeros` (as you suggest) and `shift_left`.

src/cdomains/intDomain.ml
```diff
@@ -2307,16 +2307,36 @@ struct
...
     res
 
+  (* Handle unsigned overflows. 
+     From n === k mod (2^a * b), we conclude n === k mod 2^a, for a <= 2^{bitwidth}.
+     The congruence modulo b may not persist on an overflow. *)
+  let handle_overflow ik (c, m) =
+    let max = (snd (Size.range ik)) +: Ints_t.one in
+    (* Find largest m'=2^k (for some k) such that m is divisible by m' *)
+    let tz = Ints_t.trailing_zeros m in
+    let m' = Ints_t.shift_left (Ints_t.of_int 1) tz in
```
sim642, line 2317: We need a special case here since an intdomain property test is failing. This is because of a special case:

>  If `x` is zero, `trailing_zeros x` returns `max_int`.

jerhard: Thanks for the hint! `m=0` is actually only passed to `handle_overflow` when one of the operands of `mul` is `0` and `no_ov = false`. This could e.g. happen when mutual refinement between domains is not active, so it's good to handle this.

src/cdomains/intDomain.ml
```diff
@@ -2307,16 +2307,40 @@ struct
     if M.tracing then  M.trace "congruence" "shift_left : %a %a becomes %a \n" pretty x pretty y pretty res;
     res
 
+  (* Handle unsigned overflows.
+     From n === k mod (2^a * b), we conclude n === k mod 2^a, for a <= 2^{bitwidth}.
```
sim642, line 2311: `a` is already an exponent of 2, so do you mean this?
```suggestion
     From n === k mod (2^a * b), we conclude n === k mod 2^a, for a <= bitwidth.
```

michael-schwarz (approved):

sim642 (approved):

PR #896, opened 2022-11-09
Fix handling of labels in unrolling & add possibility to run CIL checks on CFGs

### LINKED ISSUE #895: Loop Unrolling Fails for `do { } while(0)` Loop contained in a switch
Minimal example:

~~~C
// PARAM: --set exp.unrolling-factor 1
int main()
{
  int m;
  int i =0;

  switch (m)
  {
    default:
      do { } while (0);
  }
  return 0;
}
~~~

fails with

```
Fatal error: exception Not_found
Raised at Stdlib__Hashtbl.MakeSeeded.find in file "hashtbl.ml", line 393, characters 21-36
Called from BatOption.map in file "src/batOption.ml", line 37, characters 19-24
Called from Goblint_lib__Constraints.FromSpec.tf in file "src/framework/constraints.ml", line 710, characters 17-53
Called from BatList.map in file "src/batList.mlv", line 239, characters 23-28
Called from Goblint_lib__Constraints.FromSpec.system.tf in file "src/framework/constraints.ml", line 740, characters 27-47
Called from Goblint_lib__Constraints.EqConstrSysFromGlobConstrSys.conv in file "src/framework/constraints.ml", line 876, characters 4-105
```

This causes issues where some 33 tasks go from correct to `Exception(Not_found)`  in SV-Comp 23 when enabling the auto-tuner for loop unrolling.

michael-schwarz:
Our handling of labels when unrolling loops was still slightly off:

- We did not populate the `labelAlphaTable`, so there were collisions possible with labels defined outside the loop that is being unrolled, as we started from a fresh alpha table.
- We were excessively copying statements (not leaving the original statement in the CFG as well (at the very end), causing `goto`s from outside the CFG to point to  statements that were not included in the current CFG anymore.

To catch such issues early from now on, I added the option `dbg.run_cil_check` that causes CIL integrity checks to run and enabled them for our loop unrolling tests.

**TODO:**

- [x] Wait for merge of https://github.com/goblint/cil/pull/121 and update pins

Closes #895

### DIFF
--- a/src/util/loopUnrolling.ml
+++ b/src/util/loopUnrolling.ml
@@ -446,9 +446,6 @@ class loopUnrollingVisitor(func) = object
           annotateArrays b;
           match s.skind with
           | Loop (b,loc, loc2, break , continue) ->
-            (* We copy the statement to later be able to modify it without worrying about invariants *)
-            let s = { s with sid = s.sid } in
-
             (* top-level breaks should immediately go to the end of the loop, and not just break out of the current iteration *)
             let break_target = { (Cil.mkEmptyStmt ()) with labels = [Label (Cil.freshLabel "loop_end",loc, true)]} in
             (* continues should go to the next unrolling *)
@@ -464,10 +461,11 @@ class loopUnrollingVisitor(func) = object
             mkStmt (Block (mkBlock (copies@[s]@[break_target])))
           | _ -> failwith "invariant broken"
         ) else s (*no change*)
-      in ChangeDoChildrenPost({s with sid = s.sid},duplicate_and_rem_labels);
+      in ChangeDoChildrenPost(s, duplicate_and_rem_labels);
     | _ -> DoChildren
 end
 
 let unroll_loops fd =
+  Cil.populateLabelAlphaTable fd;
   let thisVisitor = new loopUnrollingVisitor(fd) in
   ignore (visitCilFunction thisVisitor fd)
--- a/tests/regression/55-loop-unrolling/09-weird.c
+++ b/tests/regression/55-loop-unrolling/09-weird.c
@@ -1,0 +1,13 @@
+// PARAM: --set exp.unrolling-factor 2 --enable dbg.run_cil_check
+
+void main(void)
+{
+    int j = 0;
+
+    for(int i=0;i < 50;i++) {
+        goto somelab;
+        somelab: j = 8;
+    }
+
+    __goblint_check(j==8);
+}
--- a/tests/regression/55-loop-unrolling/09-weird.c
+++ b/tests/regression/55-loop-unrolling/09-weird.c
@@ -1,4 +1,5 @@
 // PARAM: --set exp.unrolling-factor 2 --enable dbg.run_cil_check
+#include <goblint.h>
 
 void main(void)
 {
--- a/goblint.opam
+++ b/goblint.opam
@@ -74,7 +74,7 @@ dev-repo: "git+https://github.com/goblint/analyzer.git"
 # also remember to generate/adjust goblint.opam.locked!
 available: os-distribution != "alpine" & arch != "arm64"
 pin-depends: [
-  [ "goblint-cil.2.0.0" "git+https://github.com/goblint/cil.git#bfb977b48d6d92e0ecbdab2e04f2576095506107" ]
+  [ "goblint-cil.2.0.0" "git+https://github.com/goblint/cil.git#6eec36947c793e2efe59b4f26dd6cd57498c830a" ]
   # TODO: add back after release, only pinned for optimization (https://github.com/ocaml-ppx/ppx_deriving/pull/252)
   [ "ppx_deriving.5.2.1" "git+https://github.com/ocaml-ppx/ppx_deriving.git#0a89b619f94cbbfc3b0fb3255ab4fe5bc77d32d6" ]
   # TODO: add back after release, only pinned for CI stability
--- a/goblint.opam
+++ b/goblint.opam
@@ -74,7 +74,7 @@ dev-repo: "git+https://github.com/goblint/analyzer.git"
 # also remember to generate/adjust goblint.opam.locked!
 available: os-distribution != "alpine" & arch != "arm64"
 pin-depends: [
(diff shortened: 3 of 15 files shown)


michael-schwarz:
comparison

It has an overwhelmingly positive impact if we do this and enable the loop autotuning. The instances in which we are unsound remain to be investigated, they may already be fixed upstream in SV-Comp.

For overflows, we succeed with an additional 14 tasks, for no-data-race it makes no difference.

tests/regression/55-loop-unrolling/09-weird.c
```diff
@@ -0,0 +1,13 @@
+// PARAM: --set exp.unrolling-factor 2 --enable dbg.run_cil_check
```
sim642, line 1:
```suggestion
// PARAM: --set exp.unrolling-factor 2 --enable dbg.run_cil_check
#include <goblint.h>
```

src/util/loopUnrolling.ml
```diff
@@ -446,9 +446,6 @@ class loopUnrollingVisitor(func) = object
           annotateArrays b;
           match s.skind with
           | Loop (b,loc, loc2, break , continue) ->
-            (* We copy the statement to later be able to modify it without worrying about invariants *)
-            let s = { s with sid = s.sid } in
```
sim642, line 450: Why is this now unnecessary? To the contrary, in https://github.com/goblint/analyzer/pull/563#discussion_r789751464 it was deemed necessary.

michael-schwarz:
The unsound cases are all LDV-related, maybe you can have a look @sim642 since you looked into some similar ones as well.

- c/ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--ppp--ppp_generic.ko-entry_point.cil.out.yml
- c/ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wan--hdlc_cisco.ko-entry_point.cil.out.yml
- c/ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wan--hdlc_fr.ko-entry_point.cil.out.yml
- c/ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wireless--mwifiex--mwifiex_sdio.ko-entry_point.cil.out.yml
- c/ldv-validator-v0.6/linux-stable-431e8d4-1-102_1a-drivers--net--r8169.ko-entry_point.cil.out.yml
- c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--md--dm-cache.ko-entry_point.cil.out.yml
- c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-08_1a-drivers--staging--comedi--comedi.ko-entry_point.cil.out.yml
- c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-32_7a-drivers--md--dm-crypt.ko-entry_point.cil.out.yml
- c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-32_7a-drivers--media--v4l2-core--videodev.ko-entry_point.cil.out.yml

At this point I am fairly convinced that the tasks are at fault, and would propose to merge this regardless (and in the worst case if we can not fix the benchmarks before keep the autotuner option for loop unrolling off for sv-comp).

src/util/loopUnrolling.ml
michael-schwarz: Yes, it was, because it covered up some other bugs. After those are now fixed, it is correct to keep one copy of the statement identical: Since loops are unrolled bottom-up, first the innermost loop gets duplicated (with the last instance remaining the same). Then, the outer loop is unrolled, maintaining the invariant that there is at most one copy.

michael-schwarz:
> Would be good to look into at least some of them to determine what's at fault and whether we will be able to do something about them any more after the tasks are frozen.

Iirc, if we find UB, we can still move to exclude those tasks, right?

tests/regression/55-loop-unrolling/09-weird.c
michael-schwarz: Include `goblint.h` header

sim642:
I looked up if any of those have been modified/have open MRs in sv-benchmarks and added links for those.

It is possible to exclude tasks after freezing by proposing changes, but getting those merged is more difficult since "pull requests after deadline only merged after jury approval".
But either way, since loop unrolling is currently disabled, we can keep it disabled until all of those tasks have been handled.

michael-schwarz:
**c/ldv-linux-3.16-rc1/205_9a_array_unsafes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wan--hdlc_cisco.ko-entry_point.cil.out.i**


- `ldv_check_final_state()` checks that `ldv_set_is_empty()` returns a non-zero value
- `ldv_set_is_empty()` is `return (last_index == 0);`
- `last_index` is a global initalized to zero
-  It is modified from `ldv_set_add` and `ldv_set_remove` (which are both uncalled)

- `ldv_set_add` is uncalled and called
    - from `ldv_skb_alloc` (which is uncalled)
        - it is called from `ldv_dev_alloc_skb_12` (which is uncalled)
            - which is called from `cisco_keepalive_send` (which is uncalled)
                - called from `cisco_rx` (where `skb->len <= 3U` is always true and the call to `cisco_keepalive_send` is never performed)
                    - skb is a pointer to `proto_group1` which is initialized to `0` in `ldv_initialize_hdlc_proto_3` and never assigned any different value

    - and from `ldv_netdev_alloc_skb` (which is uncalled)
        - it is called from `ldv_dev_alloc_skb` (which is completely uncalled)

- `ldv_set_remove` only changes `last_index` if it is positive (and thus does not modify it here)

tests/regression/55-loop-unrolling/09-weird.c
sim642: Nothing was actually changed here.

sim642:
I opened a new issue about these unsound cases to not block this PR.

sim642 (approved):

PR #902, opened 2022-11-14
Cleanup array domain

### LINKED ISSUE #485: Cleanup of the array domains

michael-schwarz:
This cleans up the partitioned array domain such that illegal values cannot be represented by the type anymore.

Closes #485

### DIFF
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -443,10 +443,9 @@ struct
       | Joint v ->
         (match i with
          | Some i when not @@ not_allowed_for_part i ->
-          (*TODO: Lifting crap *)
            let l = if equals_zero i then Val.bot () else join_of_all_parts x in
            let r = if equals_maxIndex i then Val.bot () else join_of_all_parts x in
-          Partitioned (i, (l, a, r))
+           Partitioned (i, (l, a, r))
          | _ -> Joint (Val.join v a)
         )
       | Partitioned (e, (xl, xm, xr)) ->
--- a/src/cdomains/arrayDomain.ml
+++ b/src/cdomains/arrayDomain.ml
@@ -8,8 +8,6 @@ module A = Array
 module Q = Queries
 module BI = IntOps.BigIntOps
 
-module LiftExp = Printable.Lift (CilType.Exp) (Printable.DefaultNames)
-
 type domain = TrivialDomain | PartitionedDomain | UnrolledDomain
 
 (* determines the domain based on variable, type and flag *)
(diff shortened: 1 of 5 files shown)


src/cdomains/arrayDomain.ml
```diff
@@ -195,102 +195,118 @@ sig
...
-module Partitioned (Val: LatticeWithSmartOps) (Idx:IntDomain.Z):SPartitioned with type value = Val.t and type idx = Idx.t =
+module Partitioned (Val: LatticeWithSmartOps) (Idx:IntDomain.Z): SPartitioned with type value = Val.t and type idx = Idx.t =
 struct
-  (* Contrary to the description in Michael's master thesis, abstract values here always have the form *)
-  (* (Expp, (Val, Val, Val)). Expp is top when the array is not partitioned. In these cases all three  *)
-  (* values from Val are identical *)
-  module Expp = ExpDomain
-  module Base = Lattice.Prod3 (Val) (Val) (Val)
-  include Lattice.ProdSimple(Expp) (Base)
+  type t = Joint of Val.t | Partitioned of (CilType.Exp.t * (Val.t * Val.t * Val.t)) [@@deriving eq, ord, hash]
```
sim642, line 198: Just a thought: doesn't the `Joint` case just duplicate the trivial array domain? Even if so, it might be a stretch to try to reuse code between the two.

michael-schwarz: Yes, in some sense it does, though I am bit hesitant to add reuse here. The Trvivial domain is very easy to understand and I would want to avoid complicating it by sharing code with the partitioned domain.

sim642: I didn't mean to reduce trivial domain to a special case of partitioned, but rather doing these cases at the level of `AttributeConfiguredArrayDomain` where we already have cases between different array domains. If it's not partitioned, then it could switch to trivial at that level, removing the case splitting from partitioned altogether. But achieving that would be more refactoring than necessary for now.

src/cdomains/arrayDomain.ml
```diff
@@ -188,109 +188,125 @@ end
...
+    | Joint x, Joint y -> Val.leq x y
+    | Partitioned (e,(xl, xm, xr)), Joint y -> Val.leq xl y && Val.leq xm y && Val.leq xr y
+    | Partitioned (e,(xl, xm, xr)), Partitioned (e',(yl, ym, yr)) ->
+        CilType.Exp.equal e e' && Val.leq xl yl && Val.leq xm ym && Val.leq xr yr
+    | Joint x, Partitioned _ -> Val.is_bot x
+
+  let bot () = Joint (Val.bot ())
+  let is_bot (x:t) = Val.is_bot (join_of_all_parts x)
+  let top () = Joint (Val.top ())
+  let is_top (x:t) = x = top ()
```
sim642, line 235: Polymorphic equality.

sim642 (approved):

PR #903, opened 2022-11-14
Do not set overflow flag during refinement

### LINKED ISSUE #866: `ana.int.refinement` can worsen precision
Here's a minimized case, where @karoliineh used our `very-precise` example conf, which actually made things less precise.

```c
int main() {
  int i, j;
  if (i < j)
    return 1;
  else
    return 2;
}
```

Without int refinement, no warnings are printed:
```console
./goblint --enable ana.int.interval --set ana.int.refinement never
```

With int refinement, a spurious overflow warning is printed at the comparison:
```console
./goblint --enable ana.int.interval --set ana.int.refinement once
```
(Same with fixpoint refinement.)

michael-schwarz:
This is another fix in the same vein as #890, where a call to `norm` caused the overflow flag being set unncessarily.
@karoliineh could you verify this fix behaves as expected.

Closes #866.

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -905,7 +905,7 @@ struct
       let l' = if Ints_t.equal l min_ik then l else shrink Ints_t.add l in
       let u' = if Ints_t.equal u max_ik then u else shrink Ints_t.sub u in
       let intv' = norm ik @@ Some (l', u') in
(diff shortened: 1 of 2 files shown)


karoliineh:
Seems good to me. Although I just stumbled upon it, @sim642 noticed it and constructed the minimized case, so maybe he wants to take a look.

sim642 (approved):

PR #904, opened 2022-11-14
Int Domains: Do not refine on joining or widening

### LINKED ISSUE #864: `exception Lattice.Invalid_widen` with enum domain & refinement
During #855, Helmut and I discovered that enabling the enum domain and refinement leads to crashes.

For the reduced program:
~~~C
int main() {
  int a = 3;
  while (1)
    a += 2;
}
~~~

`./goblint  --enable ana.int.enums --sets ana.int.refinement fixpoint reduce.c` crashes with

```
Fatal error: exception Lattice.Invalid_widen(([Unit:(), Unit:(), Unit:(), Reversed (top or Set (Normal Lvals * booleans)):{}, Unit:(), Unit:(), top or Set (variables):{}, booleans:False, MT mode:Singlethreaded, Thread * lifted created and Unit:([main], bot), value domain * array partitioning deps * Vars with Weak Update * P:(mapping {
                                                                                                                                                                                                                                                                                                                  Local {
                                                                                                                                                                                                                                                                                                                    a ->   (Not {0}([0,7]),{3, 5})
                                                                                                                                                                                                                                                                                                                  }
                                                                                                                                                                                                                                                                                                                }, mapping {
                                                                                                                                                                                                                                                                                                                       }, {}, {}), top or std * lifted node:(mapping {
                                                                                                                                                                                                                                                                                                                                                               }, Unknown node), Unit:()], mapping {
                                                                                                                                                                                                                                                                                                                                                                                             }):
not leq ([Unit:(), Unit:(), Unit:(), Reversed (top or Set (Normal Lvals * booleans)):{}, Unit:(), Unit:(), top or Set (variables):{}, booleans:False, MT mode:Singlethreaded, Thread * lifted created and Unit:([main], bot), value domain * array partitioning deps * Vars with Weak Update * P:(mapping {
                                                                                                                                                                                                                                                                                                    Local {
                                                                                                                                                                                                                                                                                                      a ->   (Not {2}([0,8]),not {2} ([-31,31]))
                                                                                                                                                                                                                                                                                                    }
                                                                                                                                                                                                                                                                                                  }, mapping {
                                                                                                                                                                                                                                                                                                         }, {}, {}), top or std * lifted node:(mapping {
                                                                                                                                                                                                                                                                                                                                                 }, Unknown node), Unit:()], mapping {
                                                                                                                                                                                                                                                                                                                                                                               }) because Map: a = (Not {0}([0,7]),{3, 5}) instead of (Not {2}([0,8]),not {2} ([-31,31])). 
)
```

michael-schwarz:
Currently, we run into issues where due to refinement being applied after joining values we end up with a left argument to `widen` that is not `leq` the right `argument` (according to lattice order, property would still hold for concretizations).

With this PR, we only apply refinements upon transfer functions, meets, and narrows. The intuition here is that the corresponding refinement will be performed later on anyway, and very little actual gain can be hoped for when refining a value resulting from a join of two values that were already refined.

Closes #864

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -2753,12 +2753,15 @@ module IntDomTupleImpl = struct
         , map (r.f1 (module I3)) c
         , map (r.f1 (module I4)) d )
 
-  let map2 ik r (xa, xb, xc, xd) (ya, yb, yc, yd) =
-    refine ik
+  let map2 ?(norefine=false) ik r (xa, xb, xc, xd) (ya, yb, yc, yd) =
+    let r =
       ( opt_map2 (r.f2 (module I1)) xa ya
       , opt_map2 (r.f2 (module I2)) xb yb
       , opt_map2 (r.f2 (module I3)) xc yc
       , opt_map2 (r.f2 (module I4)) xd yd )
+    in
+    if norefine then r else refine ik r
+
 
   (* f1: unary ops *)
   let neg ?no_ov ik =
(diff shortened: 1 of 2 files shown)


sim642:
Wouldn't it suffice to just not refine (i.e. reduce a product) after widening?

[Miné's tutorial](https://hal.sorbonne-universite.fr/hal-01657536/document) at least states the following:

> Indeed, although applying the widening on each component of the product independently ensures termination, this is no longer the case if we apply a reduction in-between widening steps. Intuitively, the reduction may strengthen the abstract information, such as providing finite interval bounds, while the widening enforces convergence by setting bounds to infinity. 
> [...]
> Our solution is to avoid performing a reduction after a widening application.

michael-schwarz:
For termination, that should suffice. However, this would still lead to `Invalid_widen` exceptions being thrown as the check that for `widen a b` `a \leq b` may fail, as `leq` does not consider refinement.

sim642:
True, our issue wasn't termination but these refinements going incomparable. I suspect a refinement-aware `leq` isn't even possible for the case that caused the exception. That might be due to our not-fully-lattice int domains with exclusion sets.

sim642 (approved):

PR #908, opened 2022-11-16
Support `location_invariant` entry type in YAML witnesses

sim642:
Some time ago the corresponding sv-witnesses MR was merged: https://gitlab.com/sosy-lab/benchmarking/sv-witnesses/-/merge_requests/59.

With this change, we generate both `location_invariant` and `loop_invariant` entries. What were previously `loop_invariant`s are now `location_invariant`s instead. At loop heads we generate actual `loop_invariant`s now, because location invariants have no valid locations there.

### TODO
- [x] Update pin after https://github.com/goblint/cil/pull/125.


src/witness/yamlWitness.ml
```diff
@@ -553,13 +589,28 @@ struct
           None
       in
 
+      let validate_location_invariant (location_invariant: YamlWitnessType.LocationInvariant.t) =
+        let loc = loc_of_location location_invariant.location in
+        let inv = location_invariant.location_invariant.string in
+        let entry_certificate = Entry.loop_invariant_certificate in (* TODO: wrong *)
```
michael-schwarz, line 595: What is meant by `TODO: wrong` here?

src/witness/yamlWitness.ml
```diff
@@ -553,13 +589,28 @@ struct
...
+        let msgLoc: M.Location.t = CilLocation loc in
+
+        match Locator.find_opt locator loc with
+        | Some lvars ->
+          validate_lvars_invariant ~entry_certificate ~loc ~lvars inv
+        | None ->
+          incr cnt_error;
+          M.warn ~category:Witness ~loc:msgLoc "couldn't locate invariant: %s" inv;
```
michael-schwarz, line 603:
```suggestion
 
        match Locator.find_opt locator loc with
        | Some lvars ->
          validate_lvars_invariant ~entry_certificate ~loc ~lvars inv
        | None ->
          incr cnt_error;
          M.warn ~category:Witness ~loc:(CilLocation loc) "couldn't locate invariant: %s" inv;
```

`msgLoc` appears to only be used here.

src/witness/yamlWitness.ml
sim642: It's wrong in the sense that there's no `location_invariant_certificate`, nor any general invariant certificate for the time being, so this is the closest thing matching old behavior. Since we don't really care about the certificates, I think it's fine, but I can explain this also in the TODO.

michael-schwarz: Yes, i think that might be helpful for others like me who are not intimately familiar with witnesses.

PR #909, opened 2022-11-16
Move incremental data management out of TD3

sim642:
This is a set of commits cherry-picked from #596.

Moving this out should allow the intended refactoring in #738. With immutable marshaling record it can be easily wrapped to carry other components that a TD3 functor would desire (e.g. `dep_vals`).


src/framework/constraints.ml
```diff
@@ -925,7 +927,7 @@ end
 
 (** Transforms a [GenericEqBoxIncrSolver] into a [GenericGlobSolver]. *)
 module GlobSolverFromEqSolver (Sol:GenericEqBoxIncrSolverBase)
-  : GenericGlobSolver
+  (* : GenericGlobSolver *)
```
jerhard, line 930: Is the result of the functor application still a `GenericGlobSolver`? If not, one may remove the comment (also above)

jerhard:
Besides the small comment above, this looks good.

src/framework/constraints.ml
sim642: It should still be, but I think I removed the signature constraint to expose some internals. I'll check if it's still necessary.

sim642: I removed it altogether. It's not really possible to add a `with type marshal` constraint to this because it's a functor and the actual definition as `Sol'.marshal` depends on multiple functor applications in the functor body.

jerhard (approved):

jerhard (approved):

PR #911, opened 2022-11-17
Autotune: Conditionally activate congruence integer domain

jerhard:
This PR extends the autotuner to be able to enable the congruence domain for a whole program. Previously, the autotuner only activated the congruence domain for functions that contain the `%`-operator (remainder)  or  were called by such functions. 

However, the congruence domain may be useful even in some cases when the program does not contain the remainder operator. Therefore the congruence domain is activated on a program when there is still some "autotuner fuel" left for the configuration. This should enable the domain on small programs. 

The formulas for computing the `cost` and `value` option were chosen as something that seemed sensible, but were not fine-tuned.
 
 This PR also adds the `congruence` option to the autotuner in the `svcomp23.json` configuration. That means that the autotuner may activate the `congruence` domain, on a function level or for a whole program.

### DIFF
--- a/conf/svcomp23.json
+++ b/conf/svcomp23.json
@@ -67,6 +67,7 @@
         "mallocWrappers",
         "noRecursiveIntervals",
         "enums",
+        "congruence",
         "octagon",
         "wideningThresholds"
       ]
--- a/conf/svcomp23.json
+++ b/conf/svcomp23.json
@@ -67,7 +67,6 @@
         "mallocWrappers",
         "noRecursiveIntervals",
         "enums",
-        "congruence",
         "octagon",
         "wideningThresholds"
       ]
--- a/src/autoTune.ml
+++ b/src/autoTune.ml
@@ -315,6 +315,21 @@ class octagonFunctionVisitor(list, amount) = object
 
 end
 
+let congruenceOption factors file =
+  let locals, globals = factors.integralVars in
+  let cost = (locals + globals) * (factors.instructions / 12) + 5 * factors.functionCalls in
+  let value = 5 * locals + globals in
+  let activate () =
+    print_endline @@ "Congruence: " ^ string_of_int cost;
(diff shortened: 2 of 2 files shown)


conf/svcomp23.json
```diff
@@ -67,6 +67,7 @@
         "mallocWrappers",
         "noRecursiveIntervals",
         "enums",
+        "congruence",
```
michael-schwarz, line 70: Maybe let's not activated it in the configuration on master yet, until we have an estimate of the impact (In Manuel's thesis there was a negative impact, but that was largely due to `Error(verify)` which should be fixed now).

Other than that, it looks good to merge.

michael-schwarz (approved):

conf/svcomp23.json
jerhard: I removed the setting from `svcomp23.json`.

src/autoTune.ml
```diff
@@ -315,6 +315,21 @@ class octagonFunctionVisitor(list, amount) = object
...
+  let value = 5 * locals + globals in
+  let activate () =
+    print_endline @@ "Congruence: " ^ string_of_int cost;
+    set_bool "ana.int.congruence" true;
+    print_endline "Enabled congruence domain.";
+  in
+  {
+    value;
+    cost;
+    activate;
```
michael-schwarz, line 330: Alternative to try from #912 

```ocaml
    value = 8000;
    cost = 8000;
```

michael-schwarz:
- For reach safety, the overall impact is still negative, but we succeed on many interesting tasks.
- For no overflow, we see a small positive impact.
- For no data race, there is no impact.

Julian is working on fine-tuning the parameters now.

michael-schwarz:
I started another run with the new optimization.

michael-schwarz:
As this is deactivated by default and the issue we uncovered is not here, I am merging this now.

PR #912, opened 2022-11-17
Enable constant loop unrolling for small files

leunam99:
This adds two small features to the autotuner:
If there are less than 5 loops, they all get unrolled at least 10 times
If the file is small, the congruence domain is enabled.


michael-schwarz:
I moved the congruence changes over to #911 so we have two separate PRs.

michael-schwarz:
With less than 5 loops, for sv-comp this does not trigger anywhere.

michael-schwarz:
To get the results for the tasks we care about to change, we need to apply this if there are less than 17 loops. Will still benchmark that, but it is dubious if this will be positive overall.

michael-schwarz:
We will not have time to benchmark this in time for the deadline, so let's move it to 2024.

michael-schwarz:
I have now moved this to a separate autotuner `forceLoopUnrollForFewLoops` that is disabled by default. I propose to merge this now in the interest of reducing the number of stale PRs. We may then investigate the benefits of this option when it is time to prepare the configuration for this year's SV-Comp.

sim642 (approved):

PR #914, opened 2022-11-17
Warnings repositioning analysis

karoliineh:
This PR contains an implementation of the warnings (alarms) repositioning analysis (`warnPostProcess`) proposed by [Muske et al.](https://dl.acm.org/doi/10.1145/3213846.3213850)


src/analyses/warnPostProc.ml
```diff
@@ -0,0 +1,561 @@
...
+    let group = {group with pieces = pieces} in
+    {message with multipiece = Group group}
+  | _ -> failwith "TODO"
+
+let rec append_unique l1 l2 =
+  match l2 with
+  | [] -> l1
+  | (x::xs) -> if List.mem x l1
+    then append_unique l1 xs
+    else append_unique (x::l1) xs
```
sim642, line 133: This is visible in some profiling with lots of `List.mem` calls. Probably would be better to have sets somewhere. Or maybe only deduplicate at the very end.

Also, `List.mem` does polymorphic equality which should be avoided.

michael-schwarz:
If this is finished now? If yes should we do another round of reviews and then aim to merge it?

sim642:
I extracted #1081 from this. This contains other changes which could be cherry-picked but that we have no use for, so I didn't:
- [ ] Implementing `emit` in `query`, although this is a bit dubious.
- [ ] Evaluation of transfer functions outside of solving.

PR #916, opened 2022-11-18
Fix Apron reading globals with offsets

sim642:
This fixes the case where enabling loop unrolling heuristic for `sv-benchmarks/c/ldv-linux-3.4-simple/32_1_cilled_ok_nondet_linux-3.4-32_1-drivers--staging--speakup--speakup_spkout.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out` caused its analysis time to jump from 0.3s to a 15min timeout.

After minimization of the program and options, the bug in Apron analysis is "obvious", but getting a spurious variable in an Apron environment to cause the jump in analysis time requires a fairly intricate combination of options:
1. Loops unrolled at least once.
2. Signed overflow disabled.
3. SV-COMP witness functor enabled.
4. Apron enabled.
5. Def_exc enabled.

The bug was simply that reading an int field of a struct, e.g. `g.x` added `g` as an int variable to the Apron environment.
This causes incomparable differences between Apron states, which causes some def_exc exclusion sets not to be joined in witness functor dependencies, which in turn causes the exclusion sets to grow instead of collapsing via a join/widen at some point.
The original program has a loop counting down from 100000, which takes a while if exclusion sets count down from that one by one.

Integer fields of structs are quite common, so this could be happening in many benchmarks, but with smaller constant bounds the slowdown would not be as drastic.

### DIFF
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -66,8 +66,8 @@ struct
     let v_ins = VH.create 10 in
     let visitor = object
       inherit nopCilVisitor
-      method! vvrbl (v: varinfo) =
-        if v.vglob || ThreadEscape.has_escaped ask v then (
+      method! vlval = function
+        | (Var v, NoOffset) when v.vglob || ThreadEscape.has_escaped ask v ->
           let v_in =
             if VH.mem v_ins v then
               VH.find v_ins v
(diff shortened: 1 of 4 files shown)


michael-schwarz:
Nice catch!

michael-schwarz (approved):

michael-schwarz:
Screenshot from 2022-11-21 10-30-23

There is a slight performance improvement, though it is not very pronounced.

PR #918, opened 2022-11-20
ArrayDomain: `normalize` after `widen` and `join`

michael-schwarz:
In an sv-comp pre-run, we had an issue where a fixpoint was not reached due to missing normalization upon `widen` and `join`. This fixes the issue.
I'd suggest we wait until we have the definitive sv-comp version before merging this, in case it has non-local effects.

Follow-up to #902

### DIFF
--- a/tests/regression/22-partitioned_arrays/20-more-fixpoint.c
+++ b/tests/regression/22-partitioned_arrays/20-more-fixpoint.c
@@ -1,0 +1,37 @@
+// PARAM: --set ana.base.arrays.domain partitioned 
+#include <stdio.h>
+#include <stdlib.h>
+#include <pthread.h>
+#include <assert.h>
+
+pthread_mutex_t m;
+
+int arr[2];
+void g() {
+  int x = arr;
+  if (!(arr[0] <= arr[0])) {x = 5;}
+}
+
+void f(int i) {
+  pthread_mutex_lock(&m);
+  arr[i]++;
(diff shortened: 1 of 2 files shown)


jerhard:
It seems like the result of `meet` and `narrow` also be de-normalized. Could it have an advantage to also normalize the result of these operations? It seems it's generally better for precision in the implementation to have a normalized value.

tests/regression/22-partitioned_arrays/20-more-fixpoint.c
```diff
@@ -0,0 +1,37 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <pthread.h>
+#include <assert.h>
+
+pthread_mutex_t m;
+
+int arr[2];
+void g() {
+  int x = arr;
```
jerhard, line 11: gcc produces a warning about this line:
```
warning: initialization of ‘int’ from ‘int *’ makes integer from pointer without a cast [-Wint-conversion]
```

Is this intended?

michael-schwarz: No, that was by mistake, the assignment to `x` is not important here.

michael-schwarz:
I was unable to construct an example where it matters, but I added normalization after `meet` and `narrow` as well.

jerhard (approved):

PR #920, opened 2022-11-21
Simplify solution querying

sim642:
### Changes
1. Deduplicate `ctx` construction for querying the solution. This was copy-pasted in `Control`, `Witness` and `YamlWitness`.
2. Extract `SpecSys` and `SpecSysSol` wrapper modules for easier passing around of `Spec`, `EQSys`, `LHT` and `GHT`. Previously passing the solution to somewhere from `Control` required at least those four functor arguments with complex argument module constraints. The solution also holds the actual solution (`lh` and `gh`) instead of passing around two arguments.
3. Add joined contexts solution (a node hashtable) to solution querying lazily. This allows the joined contexts solution to be only computed once, instead of doing the same thing multiple times in various places.

This will be useful in #914 for also getting access to per-node queries.

### TODO
- [x] Fix Gobview.

### DIFF
--- a/src/framework/analyses.ml
+++ b/src/framework/analyses.ml
@@ -709,3 +709,12 @@ sig
   module LHT: BatHashtbl.S with type key = EQSys.LVar.t
   module GHT: BatHashtbl.S with type key = EQSys.GVar.t
 end
+
+module type SpecSysSol =
+sig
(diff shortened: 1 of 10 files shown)


michael-schwarz (approved):
Good to merge after sv-comp.

PR #922, opened 2022-11-21
Base.invariant: Refine congruence information on `x % 2 != c`

### LINKED ISSUE #892: Refine congruence information on `x % 2 != 0`
Currently, no special refinement is performed for the congruences domain when encountering an inequality. Implementing refinement for the general case of inequalities `x % d = c`   would require something like congruence sets, but for the case where `d = 2`, the information is expressible in the domain and should be used.
There seem to be two SV-Comp benchmarks affected by this: `c/array-examples/sanfoundry_24-1.i` and 
`sv-benchmarks/c/array-examples/sanfoundry_24-2.i`.

jerhard:
This PR adds a special check in `Base.invariant` for the case that the condition is of the form `x % 2 != c`, i.e. a remainder computation with divisor `2` where one result value can be excluded. In this case of a inequality, the information obtained is now expressed in the congruence domain.

We should decide whether we want to merge this, since this yields some benefit in precision, but only in very particular cases. The motivating example like `sv-benchmarks/c/array-examples/sanfoundry_24-1.i` is not resolved by this, since the expression that is asserted over there is an access to an array. 

Closes #892

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1794,7 +1794,18 @@ struct
             ID.meet a'' t
           | _, _ -> a''
         in
-        meet_bin a''' b'
+        let a,b = meet_bin a''' b' in
+        (* Special handling for case a % 2 != c *)
+        let a = if PrecisionUtil.(is_congruence_active (int_precision_from_node_or_config ())) then
+            let two = BI.of_int 2 in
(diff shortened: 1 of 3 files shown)


michael-schwarz (approved):
I think it makes sense to have this, might come in handy somewhere to have this extra precision. I would propose to merge this after sv-comp.

PR #923, opened 2022-11-21
Remove hack to `--undef __BLOCKS__`

### LINKED ISSUE #605: Undefining `__BLOCKS__` for preprocessing
In https://github.com/goblint/analyzer/pull/597#discussion_r810470622, it came up again that we have this weird `__BLOCKS__` undefine:
https://github.com/goblint/analyzer/blob/b11a60cd5896b41f9e0c4847825d21f045dc6cc1/src/maingoblint.ml#L161-L163

This is something MacOS specific. So the question is whether it's still required for some reason (and thus should be better commented and consistently applied to compilation database and Makefile analysis as well) or it's a relic that could be removed (maybe it was only needed when using MacOS `gcc` instead of the generic one we ask for now)?

michael-schwarz:
Seems like our Mac CI is happy without this hack. Can you check if this still works on your Mac machine @vesalvojdani?

Fixes #605

### DIFF
--- a/src/maingoblint.ml
+++ b/src/maingoblint.ml
@@ -197,8 +197,7 @@ let basic_preprocess ~all_cppflags fname =
   Preprocessor.FpathH.replace basic_preprocess_counts basename (count + 1);
   let nname = Fpath.append (GoblintDir.preprocessed ()) (Fpath.add_ext ".i" unique_name) in
   (* Preprocess using cpp. *)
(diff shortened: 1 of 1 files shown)


vesalvojdani (approved):
This is not needed on my Mac. I tried some bench repo things as well. Maybe some benchmarks ended up including these extensions if this is defined, but then it should be added to whatever benchmark needs it.

Issue #822, opened 2022-08-11
Unsound address set must-not-equality check in base


sim642:
The use of `AD.meet` to check address set must-not-equality:
https://github.com/goblint/analyzer/blob/3a189a9e85e9f494f40e44012f864e663274181b/src/analyses/base.ml#L331-L334
is unsound in at least the following cases:
- [ ] First struct field and zero index: https://github.com/goblint/analyzer/blob/3a189a9e85e9f494f40e44012f864e663274181b/tests/regression/02-base/91-ad-meet.c#L6-L16
- [ ] Union fields: https://github.com/goblint/analyzer/blob/3a189a9e85e9f494f40e44012f864e663274181b/tests/regression/02-base/92-ad-union-fields.c#L6-L17
- [ ] Non-first struct field and non-zero index with alignment: https://github.com/goblint/analyzer/issues/822#issuecomment-1212175339.

For both programs we report that the assert definitely fails, but when compiled and run, it succeeds.

Using `meet` would only work if addresses were represented in their canonical form using a byte offset from the variable.

jerhard:
The unsoundness of comparing addresses with a struct field and an index as offsets is also exhibited for non-zero offsets: 

https://github.com/goblint/analyzer/blob/712c6cef810aa3c6635a677bc9eef7dd8e7cc599/tests/regression/02-base/93-ad-struct-offset.c#L9-L27

The alignment of fields within objects is implementation defined, thus one generally cannot produce definite answers when comparing integer offsets with field offsets on structs.

sim642:
Since #809 makes this problem even more urgent, we probably need a quicker fix than completely refactoring lvalue representations to carry a canonical form or whatnot. A simpler implementation for just `AD.may_equal a1 a2` would be to pairwise check all lvalues from `a1` and `a2`, and check their may equality. That could be implemented by calculating the canonical offset numerically, a la `Cil.bitsOffset` and checking those integers for may equality. The only problem is that `Cil.bitsOffset` works only with constant indices, but we'll have abstract ones, so it'll have to be partially reimplemented in Goblint. However, CIL can still provide us with the byte/bit offsets of struct fields (using `Machdep`), so it shouldn't be too difficult.

sim642:
A remark from Zulip on optimizing the pairwise may alias checking. Checking all pairs is overly wasteful because addresses may only alias if their `varinfo`s (without `offset`s) are equal, so it suffices to do pairwise checking within each `varinfo`.

[...]

PR #924, opened 2022-11-22
Fix address set meet

### LINKED ISSUE #925: Address set meet unsound for addresses with different int domains active
The meet of the address set is unsound for addresses when different integer domains are active in these addresses. An example is given [here](https://github.com/goblint/analyzer/blob/3dd39e7abbe813ad63fe0137e6a1e17fcaa506c1/tests/regression/29-svcomp/31-dd-address-meet.c).

jerhard:
The meet of addresses is currently unsound when different integer domains may be activated within an analysis run. This is demonstrated by this [example](https://github.com/goblint/analyzer/blob/d35bc54f50531b0f4c98951df55bbcf04b7be04c/tests/regression/29-svcomp/31-dd-address-meet.c). Here, the abstract representation of the address pointed to by `slotvec0.val` contains an offset where all int domains are active, as all int domains are activated for global variables with the `annotation.int.enabled` setting. At the point of the comparison however, only `def_exc` is active. Therefore the abstract addresses are kept in different buckets and the meet that is performed yields `bottom`.

This issue also appears on sv-comp when the congruence auto-tuning is enabled, and for *some* functions the congruence domain is activated, and for others not. 

This PR fixes this issue by not keeping the `index` information in the representative that is used to split address sets.

(Closes #925)

### DIFF
--- a/src/cdomains/lval.ml
+++ b/src/cdomains/lval.ml
@@ -385,22 +385,85 @@ end
 module NormalLatRepr (Idx: IntDomain.Z) =
 struct
   include NormalLat (Idx)
+  module Offset = struct
+    type t = NoOffset | Field of CilType.Fieldinfo.t * t | Index of unit * t  [@@deriving eq, ord, hash]
+
+    let is_first_field x = match x.fcomp.cfields with
+      | [] -> false
+      | f :: _ -> CilType.Fieldinfo.equal f x
+
+    let rec cmp_zero_offset : t -> [`MustZero | `MustNonzero | `MayZero] = function
+      | NoOffset -> `MustZero
+      | Index (_, o)  -> (match cmp_zero_offset o with
+          | `MustNonzero -> `MustNonzero
+          | _ -> `MayZero)
+      | Field (x, o) ->
+        if is_first_field x then cmp_zero_offset o else `MustNonzero
+
+    let rec equal x y =
+      match x, y with
+      | NoOffset , NoOffset -> true
+      | NoOffset, x
+      | x, NoOffset -> cmp_zero_offset x = `MustZero (* cannot derive due to this special case, special cases not used for AddressDomain any more due to splitting *)
+      | Field (f1,o1), Field (f2,o2) when CilType.Fieldinfo.equal f1 f2 -> equal o1 o2
+      | Index (_,o1), Index (_,o2) -> equal o1 o2
+      | _ -> false
+
+    let rec hash = function (* special cases not used for AddressDomain any more due to splitting *)
+      | NoOffset -> 1
+      | Field (f,o) when not (is_first_field f) -> Hashtbl.hash f.fname * hash o + 13
+      | Field (_,o) (* zero offsets need to yield the same hash as `NoOffset! *)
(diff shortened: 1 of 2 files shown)


sim642:
> This PR fixes this issue by not keeping the `index` information in the representative that is used to split address sets.

But that is already the case:
https://github.com/goblint/analyzer/blob/00bfa7f52f303c2bb7007eb3a0f1085b7cae1e46/src/cdomains/lval.ml#L389-L399

Isn't this just another instance of #822?

michael-schwarz:
The issue is `Idx.top ()` that depends on the activated int domains.

src/cdomains/lval.ml
```diff
@@ -385,22 +385,90 @@ end
 module NormalLatRepr (Idx: IntDomain.Z) =
 struct
   include NormalLat (Idx)
+  module Offset = struct
```
michael-schwarz, line 388: Can we maybe change this module name to something such as `ReprOffset` to make it clear that it is distinct from normal offsets?

tests/regression/29-svcomp/31-dd-address-meet.c
```diff
@@ -0,0 +1,25 @@
+// PARAM: --enable annotation.int.enabled
```
michael-schwarz, line 1:
```suggestion
// PARAM: --enable annotation.int.enabled
// This option enables ALL int domains for globals
```

sim642:
> The issue is `Idx.top ()` that depends on the activated int domains.

That seems like a problem in itself because it's supposed to be a real top, not `top_of`. Having multiple incomparable "tops" defies the definition of top.

Also, this is the index domain, which is a separate thing from the intdomain that's actually used for integer values.

> In the example the two address values for which the meet is `\bot` are identical except for the number of activated int domains.

It still seems to me that something deeper is wrong here that such situation can arise at all. Aren't the intdomain projections that are used for converting integer abstractions between different precisions supposed to ensure that we don't have such mixture?
There might be a projection missing somewhere. If it's a problem here, it's a problem elsewhere as well because AFAIK our intdomain operations aren't meant to be used between tuples with different precisions. The fact that such mismatches arise at all is the bug.
If integers with mismatching precisions coexist, then other operations on them could break precisely in the same way.

src/cdomains/lval.ml
```diff
@@ -385,22 +385,94 @@ end
 module NormalLatRepr (Idx: IntDomain.Z) =
 struct
   include NormalLat (Idx)
+  module ReprOffset = struct
```
sim642, line 388: This is all just copy-paste from existing code... `Lval.offs` is polymorphic in both offset types, so the existing type can simply be reused. `Lval.OffsetPrintable` is a functor, so it could simply be instantiated with a different (unit) index domain without copying all this code.

sim642 (changes requested):

src/cdomains/lval.ml
jerhard: Yes, good idea! I now created PR #926 that implements that approach. Since we might want to still merge the current PR first (because we already have a benchmark running), I made it a separate PR.

PR #926, opened 2022-11-22
Make Address Representatives Use Unit-domain for Index-Offsets

### LINKED ISSUE #925: Address set meet unsound for addresses with different int domains active

jerhard:
This PR makes the `NormalLatRepr` use the `Unit`-domain for the index-offset of representatives. This way, the active integer domains do not have an impact on the representative that addresses are grouped with.

Fixes #925. This is the more elegant alternative to #924. I created it separately here because we are already benchmarking #924 and might want to merge #924 first to make it part of the Goblint version submitted to sv-comp.

When we merge #924 first, I will adapt this PR to be able to be merged.

### DIFF
--- a/src/cdomains/lval.ml
+++ b/src/cdomains/lval.ml
@@ -11,14 +11,20 @@ type ('a, 'b) offs = [
 
 
 (** Subinterface of IntDomain.Z which is sufficient for Printable (but not Lattice) Offset. *)
-module type IdxDomain =
+module type IdxPrintable =
 sig
   include Printable.S
   val equal_to: IntOps.BigIntOps.t -> t -> [`Eq | `Neq | `Top]
   val to_int: t -> IntOps.BigIntOps.t option
 end
 
-module OffsetPrintable (Idx: IdxDomain) =
+module type IdxDomain =
+sig
+  include IdxPrintable
+  include Lattice.S with type t := t
+end
+
+module OffsetPrintable (Idx: IdxPrintable) =
 struct
   type t = (fieldinfo, Idx.t) offs
   include Printable.Std
@@ -105,7 +111,7 @@ struct
     | `Index(i,o) -> NoOffset (* array domain can not deal with this -> leads to being handeled as access to unknown part *)
 end
 
-module Offset (Idx: IntDomain.Z) =
+module Offset (Idx: IdxDomain) =
 struct
   include OffsetPrintable (Idx)
 
@@ -173,7 +179,7 @@ sig
   (** Finds the type of the address location. *)
 end
 
-module Normal (Idx: IdxDomain) =
+module Normal (Idx: IdxPrintable) =
 struct
   type field = fieldinfo
   type idx = Idx.t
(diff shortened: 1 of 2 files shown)


src/cdomains/lval.ml
```diff
@@ -105,7 +111,7 @@ struct
     | `Index(i,o) -> NoOffset (* array domain can not deal with this -> leads to being handeled as access to unknown part *)
 end
 
-module Offset (Idx: IntDomain.Z) =
+module Offset (Idx: IdxDomain) =
```
sim642, line 114: It's nice that this can be weakened from `IntDomain.Z`. I was afraid we might need a full `IntDomain.Z` implementation for unit.

src/cdomains/lval.ml
```diff
@@ -382,25 +388,38 @@ struct
...
-    include Normal (Idx)
     type elt = t
 
-    let rec of_elt_offset: Offs.t -> Offs.t =
+    (* Offset module for representative without abstract values for index offsets, i.e. with unit index offsets.
+       Reason: The offset in the representative (used for buckets) should not depend on the integer domains,
+       since different integer domains may be active at different program points. *)
+    include Normal (UnitIdxDomain)
+
+    let rec of_elt_offset (*: (fieldinfo, Idx.t) offs -> (fieldinfo, UnitIdxDomain.t) offs *) =
```
sim642, line 411: Does this type signature not work anymore or?

src/cdomains/lval.ml
```diff
@@ -382,25 +388,38 @@ struct
...
       | `NoOffset -> `NoOffset
       | `Field (f,o) -> `Field (f, of_elt_offset o)
-      | `Index (_,o) -> `Index (Idx.top (), of_elt_offset o) (* all indices to same bucket *)
-    let of_elt = function
+      | `Index (_,o) -> `Index (UnitIdxDomain.top (), of_elt_offset o) (* all indices to same bucket *)
+
+    let of_elt (x: elt): t = match x with
       | Addr (v, o) -> Addr (v, of_elt_offset o) (* addrs grouped by var and part of offset *)
       | StrPtr _ when GobConfig.get_bool "ana.base.limit-string-addresses" -> StrPtr None (* all strings together if limited *)
-      | a -> a (* everything else is kept separate, including strings if not limited *)
```
sim642, line 403: We should keep this comment about the now-three remaining cases.

sim642 (commented):
Since #924 was merged for SV-COMP, this now has merge conflicts.

src/cdomains/lval.ml
jerhard: It actually works, so I re-added it.

src/cdomains/lval.ml
jerhard: I re-added the comment.

sim642 (approved):

PR #928, opened 2022-11-24
Use flambda compiler optimization

mrstanb:
As the title says, we've been working on adding flambda optimization
For this, some changes have been made in files, such as:
* `make.sh`
* `goblint.opam.locked`
* `README.md`

Please **note** that `flambda` is only supported in `ocamlopt`, hence all the flags, passed for compiler optimization, are only passed as `ocamlopt_flags`


mrstanb:
Quick note about tests and checking if semantics were compromised with `flambda`:
* I ran `make test` once with the original setup (i.e., after running `make setup`) -> everything went through smoohtly
* I ran `make test` once with the flambda setup (i.e., after running `make setup-flambda`) -> everything went through smoohtly here as well

Hence, I presume that if `make test` works in both cases, then we shouldn't have messed up anything semantics-wise.

sim642:
Thanks for the PR! There are currently some merge conflicts so the diff also shows some unrelated things that are actually already on master.

Is this just to add the flambda setup or is there already also evidence that flambda indeed provides some speedup?

michael-schwarz:
> is there already also evidence that flambda indeed provides some speedup?

Iirc there was a gain of about 1/4 for the SQLite amalgamation when testing locally, but I am sure @mrstanb et al will have the exact numbers.

mrstanb:
> Thanks for the PR! There are currently some merge conflicts so the diff also shows some unrelated things that are actually already on master.

Is this just to add the flambda setup or is there already also evidence that flambda indeed provides some speedup?

There are indeed some performance improvements. Me and my teammates will make sure to post a follow-up with some stats.

mrstanb:
> Iirc there was a gain of about 1/4 for the SQLite amalgamation when testing locally, but I am sure @mrstanb et al will have the exact numbers.

Indeed, we'll make sure to post the different results that we had as a follow-up.

mrstanb:
From my side the stats are as follows:
* Running Goblint with the base configuration, no optimizations and with the SQLite amalgamation, I got a `walltime` of `8073.680s`
* Running Goblint with the base configuration, with flambda flag `-O3` and with the SQLite amalgamation, I got a `walltime` of `6877.092s`

goblint.opam
```diff
@@ -20,7 +20,7 @@ doc: "https://goblint.readthedocs.io/en/latest/"
 bug-reports: "https://github.com/goblint/analyzer/issues"
 depends: [
   "ocaml" {>= "4.10"}
-  "dune" {>= "3.0"}
+  "dune" {>= "2.9" & >= "3.0"}
```
jerhard, line 23: What is the reason for this change?

jerhard (commented):
Thanks for the PR! It's interesting to see that activating `flambda` already has quite some impact on performance. 

Now, with this additional compiler option, it would be great to extend our Github workflows with one that nightly runs all the regressoin-tests with the `flambda` switch. To add such a workflow, you may add a file in `.github/workflows` that contains the `regression` job, which is already present in both the `locked.yml` (this uses a fixed version of compiler and dependencies) and `unlocked.yml` files.

sim642:
> To add such a workflow, you may add a file in `.github/workflows` that contains the `regression` job, which is already present in both the `locked.yml` (this uses a fixed version of compiler and dependencies) and `unlocked.yml` files.

There's no need to make a new workflow altogether. It suffices to just add one extra compiler version to the unlocked matrix.
A locked version isn't possible with flambda anyway, because the lock file is locked into a non-flambda compiler.

goblint.opam
mrstanb: > What is the reason for this change?

I presume this might have been added as a constraint by mistake while one of us in the team ran `opam` locally.

mrstanb: This should be now resolved by https://github.com/goblint/analyzer/pull/928/commits/1f9674dc07059d9715f9746b29dff04c9e81fa72

mrstanb:
I updated the `.github/workflows/unlocked.yml` file to include the OCaml compiler version `4.14.x+options`. However, I'm not sure if this would be sufficient as a change, since I saw that in the `Downgrade dependencies` step of `lower-bounds-downgrade` job `ocaml-base-compiler` is used which is not the same as the compiler version `ocaml-variants` which we use with `flambda`.
Hence, my question: should there be some further changes to the `unlocked.yml` file? Also, should we specify somewhere the `ocaml-option-flambda` setting as we do in the `golbint.opam.locked` file? Otherwise, `flambda` wouldn't be enabled in the compiler and thus we might remain in the same position as originally.

sim642:
I think just `+options` does nothing on its own. You need to actually specify the options like the setup-ocaml documentation shows: https://github.com/ocaml/setup-ocaml#inputs.

The lower bounds jobs are completely separate and need not be modified.

nathanschmidt:
_TLDR: After testing out the effects of activating different Flambda flags, we opted for using `-O3` without any further options. This leads to a speed-up of approx. 15%. The specification of further inlining options lead to mitigated results, being responsible for alternately improved or worse run-times depending on the system, hence why my teammates and I decided against them._

We started out by running Goblint with different flags over the [SQLite Amalgamation](https://www.sqlite.org/amalgamation.html). On our private machines, the flag combination `-O3 -inline-toplevel=400 -inline-max-depth=1 -inline-max-unroll=0` seemed to be the most promising choice.

To confirm those results, we then proceeded with benchmarking on the [CoolMUC-2 HPC](https://doku.lrz.de/display/PUBLIC/CoolMUC-2) of the LRZ in order to achieve reproducible results. Using [this](https://github.com/goblint/bench/issues/19#issuecomment-1197900679) configuration, the following measurements were made (format: hours:minutes:seconds):
- No optimizations: 02:37:47
- Flambda with `-O2`: 02:22:28
- Flambda with `-O3`: 02:14:31
- Flambda with `-O3 -inline-toplevel=400 -inline-max-depth=1 -inline-max-unroll=0`: 02:20:23

The runs on the HPC-cluster confirmed our assumption of a noticeable time benefit of using Flambda. However, using just `-O3` was the fastest configuration.

To ultimately take a decision on which flags to use, we continued our benchmarking with the [coreutils programs of the `bench` repository](https://github.com/goblint/bench/tree/master/coreutils). We activated some `ana.int.*` domains to see if there was some significant changes in the ordering of the above listed combinations.

On our local systems, we marked down this results: [coreutils_benchmarks.md](https://github.com/goblint/analyzer/files/10202788/coreutils_benchmarks.md). The fastest configuration, again contradicting what we had observed on the HPC, was `-O3 -inline-toplevel=400 -inline-max-depth=1 -inline-max-unroll=0`.

We then made some further batch jobs on the LRZ-cluster with `ana.int.interval`, `ana.int.def_ecx`, `ana.int.enums` and `ana.int.congruence` enabled. Only one CPU was used to achieve greater run-times, which allows for better differentiation between all measurements. The **cumulated run-times** over **all** coreutils programs were:
- No optimizations: 00:03:00
- Flambda with `-O2`: 00:02:36
- Flambda with `-O3`: 00:02:24
- Flambda with `-O3 -inline-toplevel=400 -inline-max-depth=1 -inline-max-unroll=0`: 00:02:33

Once more, on the HPC, the fastest choice was just `-O3`. Therefore, as **no clear advantage of activating the additional inlining options** can be established, we decided that it would be wisest to stick with just `-O3`, which is now reflected in our PR.

goblint.opam.locked
```diff
@@ -64,10 +64,10 @@ depends: [
   "mlgmpidl" {= "1.2.14"}
   "num" {= "1.4"}
   "ocaml" {= "4.14.0"}
-  "ocaml-base-compiler" {= "4.14.0"}
+  "ocaml-variants" {= "4.14.0+options"}
   "ocaml-compiler-libs" {= "v0.12.4"}
   "ocaml-config" {= "2"}
-  "ocaml-options-vanilla" {= "1"}
+  "ocaml-option-flambda" {= "1"}
```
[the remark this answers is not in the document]
mrstanb: Should we possibly consider creating a separate file, specifically for flambda (i.e., `make setup` would use one configuration of the project while `make setup-flambda` would use a different one)?

README.md
```diff
@@ -1,4 +1,5 @@
 # Goblint
+
```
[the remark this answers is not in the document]
mrstanb: Thanks for pointing this out. We'll take care of it :)

goblint.opam.locked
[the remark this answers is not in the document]
sim642: There can be just one opam lock file for each opam package, so that won't work.
Since the performance difference is significant, I think we just need to decide ourselves what should be the default such there's consistency across the board when it comes to benchmarking.

jerhard:
What are the built times on your machines with the flambda compiler with the regular `make` and with the release profile (`make release`) compared to the non-flambda compiler switch? That information would help us decide whether we should only keep the flambda switch, or keep it a separate option.

nathanschmidt:
I've benchmarked `make release` both with Goblint as in the master branch and with flambda -O3 as in this PR. Here are the results for 5 builds each:

No flambda: AVERAGE = 9,777s
- 9,681s
- 9,792s
- 9,812s
- 9,731s
- 9,869s

Flambda with -O3: AVERAGE = 26,075s
- 26,296s
- 26,001s
- 25,992s
- 26,183s
- 25,903s

sim642:
Thanks! Could you also get the numbers for development build (`make`)? Hopefully there the slowdown isn't as significant, because it would be slightly annoying.

nathanschmidt:
So for `make`:

No flambda: AVERAGE = 9,175s
- 9,148s
- 9,245s
- 9,092s
- 9,180s
- 9,208s

Flambda with -O3: AVERAGE = 13,829s
- 13,845s
- 13,765s
- 13,864s
- 13,844s
- 13,828s

michael-schwarz:
I think this sort of slowdown for make is acceptable here.

michael-schwarz:
I would suggest given these numbers that we should default to `flambda`. @sim642 @stilscher @jerhard  :+1: or :-1: ?

sim642:
Sure, and actually there's probably no need to have `-O3` in dev profile anyway because it disables cross-module optimizations. That probably brings the development mode overhead down even further.

jerhard:
Yes, I would also be in favor of making `flambda` the default and removing the `O3` flag for the `dev` profile.

nathanschmidt:
In my latest commit, I removed the `-O3` flag for `make`, which lead to slightly improved compilation times:

Flambda with no options: AVERAGE = 12,200s
- 12,487s
- 11,876s
- 12,164s
- 12,482s
- 11,990s

Furthermore, I made the `ocaml-variants.4.14.0+options` compiler with flambda enabled default (`make setup`). To still be able to use the `ocaml-base-compiler` if wished, I added the `make setup-base` command.

src/dune
```diff
@@ -123,6 +123,11 @@
 )
 
 (env
+  (release
+    (flags (:standard -warn-error -A -w -unused-var-strict -w +9)) ; https://dune.readthedocs.io/en/stable/faq.html#how-to-make-warnings-non-fatal
```
sim642, line 127: An extra warning has been added to the other cases meanwhile:
```suggestion
    (flags (:standard -warn-error -A -w -unused-var-strict -w -unused-functor-parameter -w +9)) ; https://dune.readthedocs.io/en/stable/faq.html#how-to-make-warnings-non-fatal
```

README.md
```diff
@@ -14,6 +14,7 @@ Both for using an up-to-date version of Goblint or developing it, the best way i
 1. Install [opam](https://opam.ocaml.org/doc/Install.html).
 2. Make sure the following are installed: `git`, `patch`, `m4`, `autoconf`, `libgmp-dev`, `libmpfr-dev` and `pkg-config`.
 3. Run `make setup` to install OCaml and dependencies via opam.
+   1. In case Goblint needs to be compiled with `flambda` optimizations, then run `make setup-flambda` instead
```
sim642, line 17: Since flambda is to be the default:
```suggestion
```

README.md
```diff
@@ -27,12 +28,11 @@ Both for using an up-to-date version of Goblint or developing it, the best way i
...
-* **[devcontainer](./.devcontainer/).** Select "Reopen in Container" in VS Code and continue with `make` using Linux instructions in devcontainer.
-* **[Docker (GitHub Container Registry)](https://github.com/goblint/analyzer/pkgs/container/analyzer)**. Run `docker pull ghcr.io/goblint/analyzer:latest` (or `:nightly`).
-* **Docker (repository).** Clone and run `docker build -t goblint .`.
-* **Vagrant.** Clone and run `vagrant up && vagrant ssh`.
-
+- **[opam](https://opam.ocaml.org/packages/goblint/)**. Install [opam](https://opam.ocaml.org/doc/Install.html) and run `opam install goblint`.
+- **[devcontainer](./.devcontainer/).** Select "Reopen in Container" in VS Code and continue with `make` using Linux instructions in devcontainer.
+- **[Docker (GitHub Container Registry)](https://github.com/goblint/analyzer/pkgs/container/analyzer)**. Run `docker pull ghcr.io/goblint/analyzer:latest` (or `:nightly`).
+- **Docker (repository).** Clone and run `docker build -t goblint .`.
+- **Vagrant.** Clone and run `vagrant up && vagrant ssh`.
```
sim642, line 35: GitHub still shows these lines as changed for some reason. Maybe there's a whitespace change that needs to be reverted?

make.sh
```diff
@@ -72,6 +80,12 @@ rule() {
         opam upgrade -y $(opam list --pinned -s)
       }
     ;; setup)
+      echo "Make sure you have the following installed: opam >= 2.0.0, git, patch, m4, autoconf, libgmp-dev, libmpfr-dev, pkg-config"
+      echo "For the --html output you also need: javac, ant, dot (graphviz)"
+      echo "For running the regression tests you also need: ruby, gem, curl"
+      echo "For reference see ./Dockerfile or ./scripts/travis-ci.sh"
+      opam_setup_flambda
+    ;; setup-base)
```
sim642, line 88: It would be better to just remove `make setup-base` and have `opam_setup()` install the flambda option from the lockfile.

make.sh
```diff
@@ -8,7 +8,15 @@ opam_setup() {
   set -x
   opam init -y -a --bare $SANDBOXING # sandboxing is disabled in travis and docker
   opam update
-  opam switch -y create . --deps-only ocaml-base-compiler.4.14.0 --locked
+  opam switch -y create . --deps-only ocaml-base-compiler.4.14.0 --update-invariant
```
sim642, line 11: The lack of `--locked` due to the need for `--update-invariant` ignores the lock file and thus breaks precise reproducibility. That's why it's better to just not offer two variants, but only have one which uses the lock file and doesn't require update-invariant.

README.md
mrstanb: I'm pretty sure that this was something my VSCode linter did automatically by mistake by inserting some spaces here and there in the `README`. I noticed that I had commit https://github.com/adelavais/analyzer/commit/cf0d5c45cdd35ce23090d2c9b6f6a712ac24efdd which fixed some merge conflict of the `README` with upstream and I think the spaces might be stemming from there as well

make.sh
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/928/commits/7013805acce83979b3eea329d454dcb82ff1e158

make.sh
mrstanb: Removed the `setup-base` option in commit https://github.com/goblint/analyzer/pull/928/commits/7013805acce83979b3eea329d454dcb82ff1e158

goblint.opam.locked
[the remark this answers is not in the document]
mrstanb: Can we consider this resolved?

sim642 (approved):

PR #930, opened 2022-11-24
Canonize and cast to

BilelGho:


michael-schwarz:
Wrong repo 😉

BilelGho:
Is there a way to get git to give me the right link in the console ? Because that would be easier than creating a PR manually each time.

PR #935, opened 2022-12-01
Read offsets separately in eval_rv_base for pointer dereference

### LINKED ISSUE #716: Different field access on imprecise address set

jerhard:
This PR changes the treatment of pointer dereference `eval_rv_base` and the `Structs.get` method, to avoid producing top in the cases described in #716. 

The change affects the case that `eval_rv_base` evaluates a pointer that is dereferenced and some offset is taken from the resulting value. Previously, all objects that the pointer could point to without the offset. Then, these memory objects were joined, and the offset was taken from the result. This led to additional imprecision in case of the join of e.g. structs of different type. 

Now, `eval_rv_base` will look at all the possible target addresses, and read for each of them the value at the offset. Then, the values at the offsets are joined, to produce the result. 
The struct domain is changed such that `get` yields a `top_value` of the corresponding type, rather than `'Top`. This way, if a field of pointer type is requested, but the abstract value does not contain the field, a top pointer (`{?, NULL}`) is constructed instead of `'Top`.  By avoiding to produce the `'Top` in this case, addresses are read from structs of the appropriate type are kept in the result of the join.  

Closes #716.


jerhard:
When the list of malloc-wrappers is provided to Goblint, this also fixes the issue of warnings for no suitable function existing for dynamic function calls dereferencing pointers with some field offset, e.g.:

 ```
[Warning][Unsound] Unknown call to function *(inst->handler). (nameserv_async.c:94:4-94:66)
[Warning][Unsound] No suitable function to be called at call site. Continuing with state before call. (nameserv_async.c:94:4-94:66)
```
There remain four cases of `"No suitable function to be called at call site. Continuing with state before call"`, however, these are cases, where the function pointer is only ever set to `Null` (and nothing else). So this is not an issue caused by the analysis.

**Comment by sim642** [comment]
> Now, `eval_rv_base` will look at all the possible target addresses, and read for each of them the value at the offset. Then, the values at the offsets are joined, to produce the result.

This replaces one `eval_offset` with as many as there are elements in the address set, which could have a performance impact.
I suppose the Munich benchmarking server is currently busy with interactive, so I'll do the benchmarks on the Tartu server.

File: src/analyses/base.ml [hunk 1]
```diff
@@ -801,43 +801,49 @@ struct
...
+          v'
         in
-        let v' = VD.cast t v in (* cast to the expected type (the abstract type might be something other than t since we don't change addresses upon casts!) *)
-        if M.tracing then M.tracel "cast" "Ptr-Deref: cast %a to %a = %a!\n" VD.pretty v d_type t VD.pretty v';
-        let v' = VD.eval_offset a (fun x -> get a gs st x (Some exp)) v' (convert_offset a gs st ofs) (Some exp) None t in (* handle offset *)
-        let v' = do_offs v' ofs in (* handle blessed fields? *)
-        v'
+        let addresses = AD.elements p in
+        let values = List.map (fun e -> do_single_address (AD.singleton e)) addresses in
+        List.fold VD.join (VD.bot ()) values
```
sim642, line 846: We should avoid the intermediate lists and just do a `AD.fold`.

Actually I was wondering where the `VD.join`-ing before this change was, because it doesn't show in the diff, but apparently that's part of `get`. This just breaks an address set to a set of singleton address sets, such that in `get` no `VD.join`-ing takes place.

jerhard: >We should avoid the intermediate lists and just do a AD.fold.

Yes, good idea, it now does this.

sim642 (approved):

jerhard:
Thanks for benchmarking this! Good to know that it does in general not have a negative impact on performance.

PR #888, opened 2022-11-04
Add configurations for some benchmark programs

jerhard:
This PR adds configurations to run goblint with on the sqlite3 amalgamation, and figlet. The configurations for sqlite include  a list of malloc-wrappers. 

Todo: Check termination of Goblint (with #885 ) on sqlite with these configs:
- [x] sqlite-minimal (~3-4h)
- [ ] sqlite-ctx-sens-no-priv 
    (similiar configuration without `threadid`, `threadflag` and `escape` analysis took around ~1.5h)
- [ ] sqlite-ctx-sens-priv
- [x] figlet, working, except for #932, non-incremental run-time ~10 seconds,

PR #936, opened 2022-12-01
Add race detection configurations for chrony, zstd, figlet

jerhard:
This PR adds configurations for race detection for chrony and figlet. An existing configuration for zstd was moved to the `conf/custom` folder that is intended to keep project-specific configuration files. This supersedes #888, which added configurations for sqlite, which we currently are not pursuing.


michael-schwarz:
Any reasons this is still a draft? If not, should we merge it?

michael-schwarz:
This should be transferred to base.

PR #941, opened 2022-12-02
CompareCIL.comparCilFiles: Handle multiple declarations of globals

### LINKED ISSUE #940: Server mode crashes on re-analysis
Goblint crashes in server mode on re-analysis of code with:
```
Fatal error: exception Failure("there can only be one definition and one declaration per global")
```

When running goblint in server mode, as outlined [here](https://github.com/goblint/analyzer/issues/678#issuecomment-1335236833) e.g. by:
```
goblint --enable server.enabled --conf goblint.json --enable server.reparse file.c
```
and the submitting commands like:
```
{"id":0,"jsonrpc":"2.0","result":{"status":["Success"]}}
```
The first analysis that is triggered succeeds, while the subsequent one crashes with the Failure above. This happens even when the changes made to the file only affect white-space.

jerhard:
Allow the case that there are multiple declarations of a global. Since we are only interested whether declarations exist or not, and the varinfo we still only keep an optional declaration.

Checks that multiple declarations of a global refer to the same varinfo. When `CompareCIL.compareCilFiles.addGlobal` is used to fold over a `Cil.file`, it should only encounter declarations of one `Cil.file`, where the referenced `varinfo`s should be the same memory object.

Closes #940.

### DIFF
--- a/src/incremental/compareCIL.ml
+++ b/src/incremental/compareCIL.ml
@@ -137,14 +137,23 @@ let compareCilFiles ?(eq=eq_glob) (oldAST: file) (newAST: file) =
         | GFun (f,_) -> f.svar.vname, {decls = None; def = Some (Fun f)}
         | GVarDecl (v,_) -> v.vname, {decls = Some v; def = None}
         | _ -> raise Not_found in
-      let merge_d def1 def2 = match def1, def2 with
+      let merge_decl decl1 decl2 = match decl1, decl2 with
         | Some d, None -> Some d
         | None, Some d -> Some d
         | None, None -> None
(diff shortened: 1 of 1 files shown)


sim642:
Conceptually this makes sense, but I think there's some deeper CIL invariant being broken. Here CIL claims that there should only be one declaration (even if the file contained multiple): https://github.com/goblint/cil/blob/5a3b7bb7f4d9b67678826ab7698f8a9048891553/src/cil.ml#L200-L202.

sim642:
Closing this because the root issue isn't the incremental comparison: https://github.com/goblint/analyzer/issues/940#issuecomment-1337579109. This just suppresses an error that should be solved differently.

PR #943, opened 2022-12-06
IntDomain code optimizations

mrstanb:
This PR tackles some code optimization attempts in the `intDomain.ml` file. Namely:
* There are currently 3 `get_bool` accesses to configuration flags of Goblint which can be stored in variables and looked up instead of being evaluated every time
* Those 3 config flags which are being read with `get_bool` are:
 * `ana.int.interval_threshold_widening`
 * `ana.int.interval_narrow_by_meet`
 * `ana.int.def_exc_widen_by_join`

What we currently do as a solution is to store all 3 `get_bool` calls to mutable `ref` variables and just dereference said variables and use them in place of the original config lookups.

**Note:**
* All runs where done on `bench/coreutils/ls_comb.c` as input
* All runs are with Goblint, compiled with the `flambda` `-O3` flag

For `ana.int.interval_threshold_widening`, the results are as follows (with the flags: `--enable ana.int.interval` and `--enable ana.int.interval_threshold_widening`):
* Original run without code changes took very long and I just stopped it around 15 minutes in
* After modifying the code, the run took `11.858s` of `walltime`
* **Note:** when running `make test`, the two tests which include the same set of flags for Goblint (namely `03/24` and `58/17`) failed. Hence, some semantics might be broken by this change. We decided to keep it for now anyways and ask why the tests are failing exactly

For `ana.int.interval_narrow_by_meet`, the results are as follows (with the flags: `--enable ana.int.interval` and `--enable ana.int.interval_narrow_by_meet`):
* Original run without code changes seemed to take long and I stopped after about 1 minute of running
* After modifying the code, the run took `11.908s` of `walltime`
* **Note:** there were no test cases which were failing here, since there were no tests for this combination of Goblint flags

For `ana.int.def_exc_widen_by_join`, the results are more or less the same with and without code changes (roughly 12 seconds and with the flags `--enable ana.int.def_exc` and `--enable ana.int.def_exc_widen_by_join`):
* Hence, this change could be possibly discarded, since it doesn't bring anything
* We decided to keep it anyways in order to remain consistent with the other two changes, described above
* Also, there were a few regression tests which include the set of these two flags (`--enable ana.int.def_exc` and `--enable ana.int.def_exc_widen_by_join`) which ran successfully both before and after the code changes here
* Furthermore, `make test` ran successfully here both prior to and after the code changes
* Thus, semantics here were most likely not broken (at least from what we saw from the passing `make test` run)

We would like to discuss the stated changes of the PR and see if they're viable and make sense.


mrstanb:
Here are the three instructions that we used to run Goblint for this PR so far:
```
./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable ana.int.def_exc --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.interval --enable ana.int.interval_threshold_widening
```

```
./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable ana.int.def_exc --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.interval --enable ana.int.interval_narrow_by_meet
```

```
./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable ana.int.def_exc --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.def_exc --enable ana.int.def_exc_widen_by_join
```

jerhard (commented):
Thanks for looking into this.  I left some remark explaining why the issues are occurring.

jerhard:
>After modifying the code, the run took 11.858s of walltime

This is probably because the option to use the `threshold widening` (which is expensive) is not picked up, see my [comment above](https://github.com/goblint/analyzer/pull/943#discussion_r1040881581).

mrstanb:
@jerhard Thanks a lot for the input and the feedback! We'll look into the options that you suggested :)

src/cdomains/intDomain.ml
```diff
@@ -609,12 +609,14 @@ struct
 
   let cast_to ?torg ?no_ov t = norm ~cast:true t (* norm does all overflow handling *)
 
+  let interval_threshold_widening_ref = ref @@ get_bool "ana.int.interval_threshold_widening"
```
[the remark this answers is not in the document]
montrie: Thanks for the tip, `make test` now passes all tests after adding lazy evaluation of the bool value.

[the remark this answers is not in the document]
KhalilCK: We also tried to do it with function that takes unit as mentioned above and also make test passes all tests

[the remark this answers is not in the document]
mrstanb: > `lazy` will misbehave in server mode, because the option will only be read once. Our `ResettableLazy` would work around that, but if we really want to optimize this all the way, mutable variables would have the least overhead. The only thing is, they have to be initialized from options in a function, which is called after the options have been handled and reloaded in the server mode.
> 
> Since there are multiple of these cached options, instead of a bunch of `ref`s, it's more convenient to use a record with `mutable` fields, which also avoids the additional indirection of a `ref`.

Thanks for the feedback! We'll make sure to look into the solution with a record in place of a `ref`. Currently we have a solution with a `ref` variable which passes `make test` and has the best running time out of all attempts. Hence, we now have to optimize this with a record, as you suggested.

mrstanb:
We refactored the mutable variable way of reading config options and moved it into its own module in the `configUtil.ml` file.
We also tried to replace all `get_bool` and `get_string` calls in `analyses/base.ml`, however this lead to a slower runtime, compared to when only optimizing in the `cdomains/intDomain.ml` file. We decided to keep the commit for reference for now, but we can get rid of it later.
`make test` runs through for this case as well

src/analyses/base.ml
```diff
@@ -2736,11 +2736,13 @@ struct
...
-        |> List.map (fun sets ->
-            List.fold_left (fun acc (lv, x) ->
+        |> List.to_seq
+        |> Seq.map (fun sets ->
+          BatList.fold_left (fun acc (lv, x) ->
                 set ~ctx (Analyses.ask_of_ctx ctx) ctx.global acc (eval_lv (Analyses.ask_of_ctx ctx) ctx.global acc lv) (Cilfacade.typeOfLval lv) x
               ) st sets
           )
+        |> List.of_seq
         |> BatList.fold_left D.meet st
```
jerhard, line 2746: Could one just use the `fold_left` on `Seq` here?

KhalilCK: Thanks for the tip, I tried it and it seems to work fine, so I will try to run Goblint now and see if we get any improvements.

mrstanb:
I changed the current config value reading optimization as follows:
* Instead of having to perform a bunch of string matches in the config util functions (`get_bool_config_value` and `get_string_config_value`), I know added a type with multiple constructors which aim to replace the strings that were being matched in the guards of these two functions
* I also split the one huge record with various config values into separate config value records, based on common config options
* This was done in commit https://github.com/goblint/analyzer/pull/943/commits/696feb7e42cc2cf3e0766967a52b56838b5824c0
* I went to run the following at this commit:

```
./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable ana.int.def_exc --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.interval --enable ana.int.interval_threshold_widening
```

and the running time was still not very impressive, namely `1430.971s` of `walltime`.

Due to this result, we decided to try and slim down the config optimization code a bit and avoid the `get_bool_config_value` and `get_string_config_value` functions now altogether.

Please let us know in case you also have any remarks

src/analyses/base.ml
sim642: `Seq.map` and `Seq.fold_left` are fine, but one has to be careful using `Seq` functions, because many were just added in OCaml 4.14, while we also support 4.10: https://v2.ocaml.org/api/Seq.html.

mrstanb:
> I changed the current config value reading optimization as follows:
> 
> * Instead of having to perform a bunch of string matches in the config util functions (`get_bool_config_value` and `get_string_config_value`), I know added a type with multiple constructors which aim to replace the strings that were being matched in the guards of these two functions
> * I also split the one huge record with various config values into separate config value records, based on common config options
> * This was done in commit [696feb7](https://github.com/goblint/analyzer/commit/696feb7e42cc2cf3e0766967a52b56838b5824c0)
> * I went to run the following at this commit:
> 
> ```
> ./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable ana.int.def_exc --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.interval --enable ana.int.interval_threshold_widening
> ```
> 
> and the running time was still not very impressive, namely `1430.971s` of `walltime`.
> 
> Due to this result, we decided to try and slim down the config optimization code a bit and avoid the `get_bool_config_value` and `get_string_config_value` functions now altogether.
> 
> Please let us know in case you also have any remarks

This brought us again in the ball-park of ~1456s of `walltime`

mrstanb:
Also tried to put all the `configUtil` module things back to `intDomain` which brought a speed up to around 1317.498s

mrstanb:
According to my last comment (https://github.com/goblint/analyzer/pull/943#issuecomment-1381759493), I'm not sure if changing the way that configs are read is really necessary, considering the fact that the speed-up is not so dramatic.
@jerhard @sim642 could you please also provide your feedback and opinions on this?

sim642:
> * Instead of having to perform a bunch of string matches in the config util functions (`get_bool_config_value` and `get_string_config_value`), I know added a type with multiple constructors which aim to replace the strings that were being matched in the guards of these two functions

I didn't have a look at the current state of this PR yet, but just an interesting tidbit: the compilation of matches over constant strings is highly efficient in OCaml: https://discuss.ocaml.org/t/constant-string-pattern-matching/10801.

src/cdomains/intDomain.ml
```diff
@@ -15,6 +15,59 @@ exception Unknown
...
 
+
+
+
+(** Define records that hold mutable variables representing different Configuration values.
+  * These values are used to keep track of whether or not the corresponding Config values are en-/disabled  *)
+  type ana_int_config_values = {
+    mutable interval_threshold_widening : float_precision option;
+    mutable interval_narrow_by_meet : float_precision option;
+    mutable def_exc_widen_by_join : float_precision option;
```
sim642, line 26:
```suggestion
    mutable interval_threshold_widening : bool option;
    mutable interval_narrow_by_meet : bool option;
    mutable def_exc_widen_by_join : bool option;
```

src/cdomains/intDomain.ml
```diff
@@ -15,6 +15,59 @@ exception Unknown
...
+    interval_narrow_by_meet = None;
+    def_exc_widen_by_join = None;
+    interval_threshold_widening_constants = None;
+    refinement = None;
+  }
+  
+  let extract_from_option x =
+      match x with
+      | None -> failwith "Config option not read"
+      | Some x -> x
```
sim642, line 42: This is just `Option.get`.

src/analyses/base.ml
```diff
@@ -855,7 +854,7 @@ struct
           | [] -> assert false
           | (e1, e2) :: eqs ->
             let eqs_for_all_mem e = List.for_all (fun (e1, e2) -> CilType.Exp.(equal e1 e || equal e2 e)) eqs in
-            let eqs_map_remove e = List.map (fun (e1, e2) -> if CilType.Exp.equal e1 e then e2 else e1) eqs in
+            let eqs_map_remove e = List.of_seq @@ Seq.map (fun (e1, e2) -> if CilType.Exp.equal e1 e then e2 else e1) @@ List.to_seq eqs in
```
sim642, line 857: What are all these `Seq` changes here? Also I'm doubtful that it does anything useful here, there's no chaining of list operations.

sim642 (changes requested):
This needs to be rebased onto the current `master` after #928.

src/analyses/base.ml
mrstanb: @KhalilCK @montrie, could you guys take a look at this, as you mostly worked on it? Thanks!

src/cdomains/intDomain.ml
mrstanb: Thanks for the tip! Should be fixed in https://github.com/goblint/analyzer/pull/943/commits/f5859c39a057ab937860b07c87eceaa217f4b49c

src/cdomains/intDomain.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/943/commits/f5859c39a057ab937860b07c87eceaa217f4b49c

src/analyses/base.ml
montrie: @mrstanb  It looks like we accidentally pushed two commits onto this branch instead of the seq-optimization branch (commits [619b634](https://github.com/goblint/analyzer/pull/943/commits/619b634ade7d40f7f6657f9dd86f5b43f7425238) and [abeffdb](https://github.com/goblint/analyzer/pull/943/commits/abeffdb83f2b8b357f8d850d249784e19022e276)) since they introduce some `Seq` notation and seem to revert changes made in your previous commit ([d7ada1c](https://github.com/goblint/analyzer/pull/943/commits/d7ada1c53e55ee58f56a2458b6d9bcc37ae678d6)).

Can we simply revert these commits or do we need to fix that by hand?

jerhard: @mrstanb @montrie 
It would be good to remove all the changes related to `Seq` from this PR to make reviewing this easier.

src/cdomains/intDomain.ml
```diff
@@ -609,20 +659,35 @@ struct
 
   let cast_to ?torg ?no_ov t = norm ~cast:true t (* norm does all overflow handling *)
 
+  (* This is a container variable for holding the config value of ana.int.interval_threshold_widening *)
```
michael-schwarz, line 662: We should not add commented out code to Goblint!

michael-schwarz:
Please rebase this PR, it is currently very hard to review since it includes both changes due to flambda, Seq, and the actual optimization.

mrstanb:
> Please rebase this PR, it is currently very hard to review since it includes both changes due to flambda, Seq, and the actual optimization.

I just rebased on top of the newest version of `master`

src/analyses/base.ml
mrstanb: Should be fixed now

src/cdomains/intDomain.ml
mrstanb: Should be fixed by https://github.com/goblint/analyzer/pull/943/commits/8ef8dcf7069a13510af4035ad1d26c64b20b7de9. Thanks for the remark!

mrstanb:
Feel free to review now after the rebase.

README.md
```diff
@@ -1,4 +1,5 @@
 # Goblint
+
```
sim642, line 2: There are still some unrelated changes in the PR diff.

src/cdomains/intDomain.ml
```diff
@@ -15,6 +15,54 @@ exception Unknown
...
+
+
+(** Define records that hold mutable variables representing different Configuration values.
+  * These values are used to keep track of whether or not the corresponding Config values are en-/disabled  *)
+type ana_int_config_values = {
+  mutable interval_threshold_widening : bool option;
+  mutable interval_narrow_by_meet : bool option;
+  mutable def_exc_widen_by_join : bool option;
+  mutable interval_threshold_widening_constants : string option;
+  mutable refinement : string option;
```
sim642, line 28: These need to be reset to `None` when server mode resets other lazy values.

sim642 (changes requested):
Was there no need to cache the int domain activation options like `ana.int.interval`? I recall from profiling a while ago that those also were noticeable.

README.md
mrstanb: Should be fixed now

src/cdomains/intDomain.ml
mrstanb: I'm not really sure if I completely know what should be done here, but I did a change in commit https://github.com/goblint/analyzer/pull/943/commits/017a721e231d8778c80595dc8e2b90e5831eb8b6 which hopefully makes sense and solves the noted issue here. Please let me know in case that change is wrong.

mrstanb:
> Was there no need to cache the int domain activation options like `ana.int.interval`? I recall from profiling a while ago that those also were noticeable.

I tried some further optimization for the activation options of the integer domains in the `precisionUtil.ml` file with some changes in the `contextUtil.ml` file. Commit https://github.com/goblint/analyzer/pull/943/commits/d5ae8ead8a88902ad64fdc534d593221790b786c reflects these changes.

---

Below I'll post some runs that I did locally on my machine which show some speed-up for the `ls_comb.c` program as input:














































---
4 runs without optimization (with ls_comb.c as input):
 * with ana.int.interval -> 15.283s
 * with ana.int.def_exc -> 15.214s
 * with ana.int.enums -> 15.370s
 * with ana.int.congruence -> 13.945s
 
4 runs with optimization (with ls_comb.c as input):
 * with ana.int.interval -> 14.466s
 * with ana.int.def_exc -> 14.488s
 * with ana.int.enums -> 14.386s
 * with ana.int.congruence -> 14.507s

---

4 runs without optimization (with ls_comb.c as input):
 * with ana.int.interval + annotation.int.enabled -> 19.078s
 * with ana.int.def_exc + annotation.int.enabled -> 19.136s
 * with ana.int.enums + annotation.int.enabled -> 18.822s
 * with ana.int.congruence + annotation.int.enabled -> 17.605s

4 runs with optimization (with ls_comb.c as input):
 * with ana.int.interval + annotation.int.enabled -> 16.523s
 * with ana.int.def_exc + annotation.int.enabled -> 16.447s
 * with ana.int.enums + annotation.int.enabled -> 16.225s
 * with ana.int.congruence + annotation.int.enabled -> 16.322s

---
**Note 1:** `with optimization` means with the changes from commit https://github.com/goblint/analyzer/pull/943/commits/d5ae8ead8a88902ad64fdc534d593221790b786c
**Note 2:** both the `with optimization` and `without optimization` runs that I did were run without the changes, provided in commit https://github.com/goblint/analyzer/pull/943/commits/017a721e231d8778c80595dc8e2b90e5831eb8b6, since I wasn't sure if this commit is correct and whether it preserves Goblint's soundness.

mrstanb:
**Note:** `make test` fails from commit https://github.com/goblint/analyzer/pull/943/commits/7c2455acd1b76c22d07b36856d921523a8c227f3 onward and it runs through successfully for commit https://github.com/goblint/analyzer/pull/943/commits/017a721e231d8778c80595dc8e2b90e5831eb8b6.

Any ideas or insights as to what could be wrong with the optimization changes in commit https://github.com/goblint/analyzer/pull/943/commits/017a721e231d8778c80595dc8e2b90e5831eb8b6?

sim642:
The `PrecisionUtil` options also need a way to reset them in server mode, although that's not directly the issue because the tests aren't in server mode.

I suspect somewhere something is done with the int domains before the config is read, so the default activations are cached and any options activating more int domains are ignored. Maybe some debug printing/exception raising in those functions which do the caching reveal when it happens.

sim642:
> The `PrecisionUtil` options also need a way to reset them in server mode, although that's not directly the issue because the tests aren't in server mode.

I fixed this by adding a corresponding `reset_lazy` function.

> I suspect somewhere something is done with the int domains before the config is read, so the default activations are cached and any options activating more int domains are ignored. Maybe some debug printing/exception raising in those functions which do the caching reveal when it happens.

The problem was caused by `FloatDomain` eagerly constructing `IntDomain` elements during initialization (when config hasn't been handled yet). I fixed this by converting some values to unit functions.

sim642 (approved):
This PR history contains some old flambda stuff and lots of back-and-forth, so a **squash merge** would be appropriate here.

michael-schwarz (approved):
LGTM, thanks everyone from the team, and thank you @sim642 for the improvements!

I'd suggest we wait with merging this until after ~#966~ #994  is landed. These two PRs conflict and given this one is a lot smaller, it'll be easier than the other way around.

mrstanb:
I ran some further benchmarks on this branch, compared to the current latest state of the `master` branch to see if we have some speed-up from this branch's optimizations at the end of the day. Below you can find the results (input program was `ls_comb.c`):

---

## On `master` branch (commit `22a720c`)

### With `ana.int.def_exc` and `annotation.int.enabled`:
* Command: `./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.def_exc --enable annotation.int.enabled`
* Time: 13.589s

### With `ana.int.interval` and `annotation.int.enabled`
* Command: `./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --disable ana.int.def_exc --enable ana.int.interval --enable annotation.int.enabled`
* Time: 14.251s

### With `ana.int.enums` and `annotation.int.enabled`
* Command: `./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --disable ana.int.def_exc --enable ana.int.enums --enable annotation.int.enabled`
* Time: 13.809s

### With `ana.int.congruence` and `annotation.int.enabled`
* Command: `./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --disable ana.int.def_exc --enable ana.int.congruence --enable annotation.int.enabled`
* Time: 13.369s

---

## On `int-domain-code-optimizations` branch (commit `7fb540e`)

### With `ana.int.def_exc` and `annotation.int.enabled`:
* Command: `./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.def_exc --enable annotation.int.enabled`
* Time: 12.293s

### With `ana.int.interval` and `annotation.int.enabled`:
* Command: `./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --disable ana.int.def_exc --enable ana.int.interval --enable annotation.int.enabled`
* Time: 12.102s

### With `ana.int.enums` and `annotation.int.enabled`:
* Command: `./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --disable ana.int.def_exc --enable ana.int.enums --enable annotation.int.enabled`
* Time: 12.248s

### With `ana.int.congruence` and `annotation.int.enabled`:
* Command: `./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --disable ana.int.def_exc --enable ana.int.congruence --enable annotation.int.enabled`
* Time: 11.522s

PR #944, opened 2022-12-07
Prevent modification of options while the solver runs

### LINKED ISSUE #653: Finalize options, so that they cannot be mutated after the solver started
Pulled out from #28 since most of the TODOs there are now done.

This would gives us an additional safety guarantee that no options are modified after the solver has started, which may lead to inconsistent results, crashing (e.g. if the list of activated analyses is modified), and other undesirable things.

just-max:
Raises an exception if a configuration value is changed while the solver is running. This should never occur.

Draft PR, to be discussed.

Resolves #653.

### DIFF
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -691,7 +691,7 @@ let rec analyze_loop (module CFG : CfgBidir) file fs change_info =
   try
     let (module Spec) = get_spec () in
     let module A = AnalyzeCFG (CFG) (Spec) (struct let increment = change_info end) in
-    A.analyze file fs
+    GobConfig.with_immutable_conf (fun () -> A.analyze file fs)
   with Refinement.RestartAnalysis ->
     (* Tail-recursively restart the analysis again, when requested.
         All solving starts from scratch.
@@ -706,8 +706,7 @@ let compute_cfg file =
 
 (** The main function to perform the selected analyses. *)
 let analyze change_info (file: file) fs =
-  GobConfig.with_immutable_conf @@ fun () ->
-    if (get_bool "dbg.verbose") then print_endline "Generating the control flow graph.";
-    let (module CFG) = compute_cfg file in
-    MyCFG.current_cfg := (module CFG);
-    analyze_loop (module CFG) file fs change_info
+  if (get_bool "dbg.verbose") then print_endline "Generating the control flow graph.";
+  let (module CFG) = compute_cfg file in
+  MyCFG.current_cfg := (module CFG);
+  analyze_loop (module CFG) file fs change_info
--- a/src/framework/control.ml
+++ b/src/framework/control.ml
@@ -706,7 +706,8 @@ let compute_cfg file =
 
 (** The main function to perform the selected analyses. *)
 let analyze change_info (file: file) fs =
-  if (get_bool "dbg.verbose") then print_endline "Generating the control flow graph.";
-  let (module CFG) = compute_cfg file in
-  MyCFG.current_cfg := (module CFG);
-  analyze_loop (module CFG) file fs change_info
+  GobConfig.with_immutable_conf @@ fun () ->
+    if (get_bool "dbg.verbose") then print_endline "Generating the control flow graph.";
+    let (module CFG) = compute_cfg file in
+    MyCFG.current_cfg := (module CFG);
+    analyze_loop (module CFG) file fs change_info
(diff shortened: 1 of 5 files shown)


src/framework/control.ml
```diff
@@ -706,7 +706,8 @@ let compute_cfg file =
...
 let analyze change_info (file: file) fs =
-  if (get_bool "dbg.verbose") then print_endline "Generating the control flow graph.";
-  let (module CFG) = compute_cfg file in
-  MyCFG.current_cfg := (module CFG);
-  analyze_loop (module CFG) file fs change_info
+  GobConfig.with_immutable_conf @@ fun () ->
+    if (get_bool "dbg.verbose") then print_endline "Generating the control flow graph.";
+    let (module CFG) = compute_cfg file in
+    MyCFG.current_cfg := (module CFG);
+    analyze_loop (module CFG) file fs change_info
```
sim642, line 713: I think this should be in `analyze_loop` around `A.analyze` instead because that restarting modifies options for the restarted run.

src/util/gobConfig.ml
```diff
@@ -75,7 +75,17 @@ sig
   (** Merge configurations from a JSON object with current. *)
   val merge : Yojson.Safe.t -> unit
 
-  val json_conf: Yojson.Safe.t ref
+  (* TODO: can bypass immutability by directly setting value *)
+  val json_conf : Yojson.Safe.t ref
```
sim642, line 79: Currently `merge` from this file also bypasses the check. It might be best to not expose `json_conf` at all but just have `set_json_conf` or something. That could just be directly used inside `GobConfig` as well, so the same check applies automatically also to single option set commands.

src/util/gobConfig.ml
```diff
@@ -304,22 +336,25 @@ struct
...
-  let set_int    st i = set_path_string_trace st (`Int i)
-  let set_bool   st i = set_path_string_trace st (`Bool i)
-  let set_string st i = set_path_string_trace st (`String i)
-  let set_list   st l =
-    drop_memo ();
-    set_value (`List l) json_conf (parse_path st)
+  let set_int    st i = set_path_string st (`Int i)
+  let set_bool   st i = set_path_string st (`Bool i)
+  let set_string st i = set_path_string st (`String i)
+  let set_list   st l = set_path_string ~show_trace:false st (`List l)
```
sim642, line 357: I think it's just a historical accident that `set_list` didn't do the tracing. You can just remove this argument and make tracing unconditional.

src/util/gobConfig.ml
```diff
@@ -98,19 +112,26 @@ struct
...
+      | Here             -> []
+      | Select (s, p)    -> Printf.sprintf ".%s" s :: helper p
+      | Index (Int i, p) -> Printf.sprintf "[%d]" i :: helper p
+      | Index (App, p)   -> "[+]" :: helper p
+      | Index (Rem, p)   -> "[-]" :: helper p
+      | Index (New, p)   -> "[*]" :: helper p
+    in
+    String.concat "" (helper p)
+
+  (** raise when an attempt is made to the configuration while it is immutable *)
```
michael-schwarz, line 127:
```suggestion
  (** raise when an attempt is made to modify the configuration while it is immutable *)
```

michael-schwarz (approved):

michael-schwarz:
There's still two regression tests failing, we should investigate what's going on there as well.

just-max:
The failures are caused by the call to `set_bool` here, while the options are finalized:

https://github.com/goblint/analyzer/blob/b3b5c9025810e3f0d2e4c9764761059f5b574963/src/analyses/malloc_null.ml#L240-L242

Is the idea that `sem.malloc.fail` should always be enabled when the `malloc_null` analysis is enabled? Or maybe the finalization still in the wrong place, in general?

michael-schwarz:
> Is the idea that `sem.malloc.fail` should always be enabled when the malloc_null analysis is enabled? 

Yes. I guess the best place to do this is with the other argument consistency checks here
https://github.com/goblint/analyzer/blob/b3b5c9025810e3f0d2e4c9764761059f5b574963/src/maingoblint.ml#L124-L138

and produce a warning that says `semmalloc.fail` implicitly enabled because `malloc_null` is activated like we do for some others.

michael-schwarz (approved):

michael-schwarz:
@sim642 Can you review so we can merge this?

sim642 (approved):

PR #945, opened 2022-12-12
Fix compilation on OCaml 5.0.0~rc1

sim642:
So far our OCaml 5 support was blocked by batteries. Now an OCaml 5 compatible version of batteries has been released: https://github.com/ocaml/opam-repository/pull/22635.
Only one Goblint issue required fixing: the deprecated `String.lowercase` is removed in OCaml 5.

There's still one gap: Apron doesn't yet have an OCaml 5 compatible version, which is why that combination is currently excluded from the unlocked matrix.

### DIFF
--- a/src/util/processPool.ml
+++ b/src/util/processPool.ml
@@ -22,7 +22,7 @@ let run ~jobs ?(terminated=fun _ _ -> ()) tasks =
           Unix.open_process task.command
       in
       let pid = Unix.process_pid proc in
-      Catapult.Tracing.a_begin ~id:(string_of_int pid) "ProcessPool" ~args:[("command", `String task.command)];
+      Catapult.Tracing.a_begin ~id:(string_of_int pid) ~cat:[] "ProcessPool" ~args:[("command", `String task.command)];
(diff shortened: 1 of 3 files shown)


src/util/processPool.ml
```diff
@@ -22,7 +22,7 @@ let run ~jobs ?(terminated=fun _ _ -> ()) tasks =
           Unix.open_process task.command
       in
       let pid = Unix.process_pid proc in
-      Catapult.Tracing.a_begin ~id:(string_of_int pid) "ProcessPool" ~args:[("command", `String task.command)];
+      Catapult.Tracing.a_begin ~id:(string_of_int pid) ~cat:[] "ProcessPool" ~args:[("command", `String task.command)];
```
sim642, line 25: This is not directly related to OCaml 5, but required for the unlocked build to pass since a new version of catapult made this argument mandatory: https://github.com/ocaml/opam-repository/pull/22639.

sim642:
There isn't anything to discuss here, so I'm just merging it such that the unlocked CI would stop failing. I just wanted this PR such that it can be related to the next release.

PR #946, opened 2022-12-13
Add option to assume thread uniqueness

vesalvojdani:
I'm considering adding some options to allow some unsound behavior from Goblint. Mainly, this is useful to assess what features would help in reducing false alarms. Assume we, e.g., implemented symbolic thread ids to handle indexed thread-local data, how much would that help? This pull request allows specifying a list of thead ids by function name that will be assumed unique. With this feature, we can get down to a single (relevant) warning on `smtprc` using [this conf](https://github.com/goblint/bench/blob/master/gobpie-demos/smtprc/smtprc.json). It also focuses on high-confidence races, which I think is okay for a demo on fixing bugs.

There are two questions before we merge. This does not work with histories, and I'm not sure it should override the history judgement, but I think it probably should because the main use-case for this override would be that there is some additional thread-id component (such as the argument) that makes the thread unique and histories do not take that into account either. 

Finally, I'm thinking maybe one should just have a thread mode that just assumes all threads are unique. (Or maybe we have that already, and I could just use that...)

### DIFF
--- a/src/cdomains/threadIdDomain.ml
+++ b/src/cdomains/threadIdDomain.ml
@@ -1,5 +1,6 @@
 open GoblintCil
 open FlagHelper
+open Prelude
 
 module type S =
 sig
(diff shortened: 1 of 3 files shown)


vesalvojdani:
Yes, I'm fine with closing this. This sort of thing would be very useful, but an unsound/benchmarking fork of goblint might be better to avoid confusing real use of goblint with unsound options.

PR #948, opened 2022-12-15
Add OCaml and library versions to `--version`

sim642:
This would be a follow-up to #928 to make it clear which compiler and libraries have been used to build a given binary.

### TODO
- [x] Fix gobview build. Dune bug?


jerhard:
Should this this already print some statically linked libraries with the current version of Goblint?
`./goblint --version --enable dbg.verbose` prints the following for me: 
```
Goblint version: heads/dune-build-info-0-g7b3b3925a-dirty
Cil version:     2.0.1-18-g5a3b7bb
Dune profile:    dev
OCaml version:   4.14.0
OCaml flambda:   false
```

sim642:
That's a quirk of our argument parsing, `/goblint --enable dbg.verbose --version` should work.
It's because `--verbose` exits early in the middle of argument parsing.

michael-schwarz (approved):
If the Gobview issue is solved, I think we can merge this!

Issue #949, opened 2022-12-19
Reluctant incremental analysis does not reach fixpoint and is unsound in some cases


jerhard:
The incremental reluctant analysis does not reach the fix-point and produces unsound results on the test case [here](https://github.com/goblint/analyzer/blob/617cb2fcbe3b17e8548f46cf6884e4ca59fea2a4/tests/incremental/00-basic/15-reluctant-test.c). Note that the configuration enables `exp.earlyglobs`, and `incremental.reluctant.enabled`.

The issue arises due to the fact that the `infl` set is reset after the call to solve for return node of `foo`. This way, the information that the return node of `foo` influences `bar` is lost. After this call to solve for the return node of `foo`, the unknowns relating to `foo` and `bar` are in `stable`. 

When `main` is re-evaluted, at the point when  `g = 4` is evaluated, this destabilizes `foo` -- but as the influence to `bar` is not recorded, `bar` is not re-evaluated. The old, unsound value of `bar` is thus reused by the analysis.

PR #950, opened 2022-12-19
Fix Unreached Fixpoint in Reluctant Incremental Analysis

jerhard:
This PR targets issue #949. This PR changes the behavior of reluctant incremental analysis for in case the abstract value of a reluctantly analyzed unknown `x` changed. 

The sets `infl x` from the previous analysis and the set `infl x` after the call to `solve x` by the reluctant analysis are joined before calling `destabilize x`. Unlike previously, this does not add `x` to the `stable` set after the call to destabilize.

### DIFF
--- a/src/util/options.schema.json
+++ b/src/util/options.schema.json
@@ -1053,14 +1053,6 @@
                 "Destabilize nodes in changed functions reluctantly",
               "type": "boolean",
               "default": false
-            },
-            "compare": {
-              "title": "incremental.reluctant.compare",
-              "description":
-                "In order to reuse the function's old abstract value the new abstract value must be leq (focus on efficiency) or equal (focus on precision) compared to the old.",
-              "type": "string",
-              "enum": ["leq", "equal"],
-              "default": "equal"
             }
           },
           "additionalProperties": false
--- a/conf/zstd-race.json
+++ b/conf/zstd-race.json
@@ -41,9 +41,6 @@
     }
   },
   "incremental": {
-    "reluctant": {
-      "compare": "leq"
-    },
     "restart": {
       "sided": {
         "enabled": false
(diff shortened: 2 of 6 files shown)


jerhard:
Looking at the old and new `infl ret_{foo}`  (`ret_{foo}` referring to the unknown at the return node of `foo`) sets occurring in the reluctant reanalyis of the test case in #949, in turns out that the sets are disjunct: the new set only records the new influence to the newly reachable `bar` function that calls `foo`. This is strange, since the `infl ret_{foo}` set should only grow during a call `solve ret_{foo}`, since the unknown is in called,  and therefor should not be destabilized, and therefore its `infl` set should not be reset.

It is also not quite clear to me yet why the test case fails when adding the unknown (in this case `ret_{foo}`) back to the `stable` set, after destabilizing it.

michael-schwarz (approved):

src/util/options.schema.json
```diff
@@ -1053,14 +1053,6 @@
...
               "type": "boolean",
               "default": false
-            },
-            "compare": {
-              "title": "incremental.reluctant.compare",
-              "description":
-                "In order to reuse the function's old abstract value the new abstract value must be leq (focus on efficiency) or equal (focus on precision) compared to the old.",
-              "type": "string",
-              "enum": ["leq", "equal"],
-              "default": "equal"
```
sim642, line 1063: Just wondering, what was the reason for removing this option? Looks like the `equal` behavior is kept, but would `leq` be incompatible with these fixes?

michael-schwarz: Switching to `leq` would now require some hacking into the solver, as it is the solver itself that causes destabilization also beyond the reluctantly analyzed function. As there was no noticable benefit with this option, we chose not to further complicate the solver and remove the option instead.

PR #953, opened 2022-12-21
Warnings and Dead Code Handling for `noreturn`

### LINKED ISSUE #501: Warn when control flow reaches end of a `_Noreturn` function
Since https://github.com/goblint/cil/pull/58, CIL supports the `_Noreturn` attribute.

As remarked by @sim642  on Slack, we should also do something with `_Noreturn` in Goblint and emit a warning when control flow does indeed reach the end of such a function (i.e. return node is not dead).

This probably also requires investigating which standard library functions terminate execution (e.g. `exit(...)` and friends).

just-max:
Resolves  #501.

Changes:
- on return of a function, warn if the function is marked `noreturn`,
- after call of a function, raise `Deadcode` if the option `sem.noreturn.dead_code` is enabled.

Added an `Other` variant to `undefined_behavior` for the message category, but could also just as well be category `Unknown`.

---

However, from what I can tell there's an issue here with how CIL handles `noreturn`. Here are the CFGs generated for the following function, once marked with `noreturn` and once unmarked:

```c
void without_noreturn(int x) {
  if (x)
    abort();
}

noreturn void with_noreturn(int x) {
  if (x)
    abort();
}
```

CFG for `without_noreturn`

image

CFG for `with_noreturn`

image

My guess is that reaching the end of a `noreturn` function is undefined behaviour, so CIL is allowed to do this.

### DIFF
--- a/goblint.opam
+++ b/goblint.opam
@@ -73,7 +73,7 @@ dev-repo: "git+https://github.com/goblint/analyzer.git"
 # also remember to generate/adjust goblint.opam.locked!
 available: os-distribution != "alpine" & arch != "arm64"
 pin-depends: [
-  [ "goblint-cil.2.0.1" "git+https://github.com/goblint/cil.git#5a3b7bb7f4d9b67678826ab7698f8a9048891553" ]
+  [ "goblint-cil.2.0.1" "git+https://github.com/goblint/cil.git#c03ade107208546ef59cd14dcefa7b55f1506494" ]
   # TODO: add back after release, only pinned for optimization (https://github.com/ocaml-ppx/ppx_deriving/pull/252)
   [ "ppx_deriving.5.2.1" "git+https://github.com/ocaml-ppx/ppx_deriving.git#0a89b619f94cbbfc3b0fb3255ab4fe5bc77d32d6" ]
   # TODO: add back after release, only pinned for CI stability
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1767,6 +1767,8 @@ struct
     set_many ~ctx (Analyses.ask_of_ctx ctx) ctx.global ctx.local inits
 
   let return ctx exp fundec: store =
+    if Cil.hasAttribute "noreturn" fundec.svar.vattr then
+      M.warn ~category:(Behavior (Undefined Other)) "Function declared 'noreturn' should not return";
     let st: store = ctx.local in
     match fundec.svar.vname with
     | "__goblint_dummy_init" ->
@@ -2273,6 +2275,8 @@ struct
       in
       let return_val = project_val (Analyses.ask_of_ctx ctx) (attributes_varinfo (return_varinfo ()) callerFundec) (Some p) return_val (is_privglob (return_varinfo ())) in
       let cpa' = project (Analyses.ask_of_ctx ctx) (Some p) nst.cpa callerFundec in
+
+      if get_bool "sem.noreturn.dead_code" && Cil.hasAttribute "noreturn" f.svar.vattr then raise Deadcode;
 
       let st = { nst with cpa = cpa'; weak = st.weak } in (* keep weak from caller *)
       match lval with
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2017,7 +2017,7 @@ struct
     let st: store = ctx.local in
     let gs = ctx.global in
     let desc = LF.find f in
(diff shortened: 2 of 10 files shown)


just-max:
We found the problem to be that the CFG construction in Goblint checks the number of (real) successors of an `If` statement. In the case of two successors, these would typically be the first statements of the true and of the false blocks. If there is only one successor, Goblint assumes it to be the successor in both branches. That's an assumption that doesn't hold if there is no else block and nothing after the `if` statement in the function. This is the case in a `noreturn` function processed in CIL, which is why the CFG ends up having the Neg edge to the successor in the true block (`abort()`, in this case).

We decided to add an option in CIL so that functions that fall through (can reach the end of the body), yet don't contain return statements, always get a return statement added by CIL—even those marked `noreturn`, which was previously not the case. PR for CIL incoming.

just-max:
still depends on https://github.com/goblint/cil/pull/129, otherwise ready to merge

michael-schwarz:
I merged the CIl PR. Could you update the pin in the opam files so we can see if the CI is happy then?

just-max:
should be ready now

michael-schwarz (approved):
LGTM, thank you!

tests/regression/64-noreturn/01-noreturn-deadcode-on.c
```diff
@@ -0,0 +1,53 @@
...
+
+int main() {
+  // switch over an unknown int, otherwise the first call
+  // to a function that aborts will dead-code the rest of main
+  int unknown;
+  scanf("%d", &unknown);
+  switch (unknown) {
+    case 0:
+      does_not_return();
+      no_op(); // TODO WARN (test script can't handle dead code warnings)
```
just-max, line 41: Could add `__goblint_check(0)` here instead, to test that this is dead. Otherwise not sure how to test dead code detection.

sim642: In some tests we have the following for it:
```c
__goblint_check(0); // NOWARN (unreachable)
```

goblint.opam
```diff
@@ -73,7 +73,7 @@ dev-repo: "git+https://github.com/goblint/analyzer.git"
 # also remember to generate/adjust goblint.opam.locked!
 available: os-distribution != "alpine" & arch != "arm64"
 pin-depends: [
-  [ "goblint-cil.2.0.1" "git+https://github.com/goblint/cil.git#5a3b7bb7f4d9b67678826ab7698f8a9048891553" ]
+  [ "goblint-cil.2.0.1" "git+https://github.com/goblint/cil.git#c03ade107208546ef59cd14dcefa7b55f1506494" ]
```
sim642, line 76: This also needs to be changed in `goblint.opam.template`.

src/analyses/base.ml
```diff
@@ -2274,6 +2276,8 @@ struct
       let return_val = project_val (Analyses.ask_of_ctx ctx) (attributes_varinfo (return_varinfo ()) callerFundec) (Some p) return_val (is_privglob (return_varinfo ())) in
       let cpa' = project (Analyses.ask_of_ctx ctx) (Some p) nst.cpa callerFundec in
 
+      if get_bool "sem.noreturn.dead_code" && Cil.hasAttribute "noreturn" f.svar.vattr then raise Deadcode;
```
sim642, line 2279: This applies to normal defined functions, but something similar should be in `special` I suppose. There could be a function which is declared to be noreturn, but doesn't provide an implementation.

sim642 (changes requested):

src/analyses/base.ml
```diff
@@ -2241,6 +2243,8 @@ struct
 
         (* List.map (fun f -> f (fun lv -> (fun x -> set ~ctx:(Some ctx) ctx.ask ctx.global st (eval_lv ctx.ask ctx.global st lv) (Cilfacade.typeOfLval lv) x))) (LF.effects_for f.vname args) |> BatList.fold_left D.meet st *)
       end
+    in
+    if get_bool "sem.noreturn.dead_code" && Cil.hasAttribute "noreturn" f.vattr then raise Deadcode else st
```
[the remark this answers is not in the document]
sim642: They might indeed, although I don't know if any of those functions would be noreturn. Either way, allowing all of those to happen before and making this check at the end should be on the safe side.

sim642 (approved):

PR #956, opened 2022-12-21
Optimization: monomorphization

adelavais:
**Note: this PR should be merged after the flambda PR #928** 

We monomorphized polymorphic operations for optimization in `src/util/cilType.ml`.

We tested the program against the `bench` repository. 
We considered #928 the original program, and the new program is #928 with the current changes added.

The best improvement is -36% of the running time, while the worst case was a 4% addition to the running time. All the measurements can be found [here](https://docs.google.com/spreadsheets/d/1CuXGHs89sDkrOyH9B8ouraDotZfsnamlHlUP2Q8mt2A/edit?usp=sharing).

### DIFF
--- a/src/util/cilType.ml
+++ b/src/util/cilType.ml
@@ -99,31 +99,28 @@ module Ikind: S with type t = ikind =
 struct
   include Std
 
-  type t = ikind
+  type t = ikind =
+    | IChar
+    | ISChar 
+    | IUChar
+    | IBool
+    | IInt
+    | IUInt
+    | IShort
+    | IUShort
+    | ILong
+    | IULong
+    | ILongLong
+    | IULongLong
+    | IInt128
+    | IUInt128
+  [@@deriving hash]
 
   let name () = "ikind"
 
   (* Identity *)
-  let _equal (xy: t * t) = match xy with
-    | (IChar, IChar) -> true
-    | (ISChar, ISChar) -> true
-    | (IUChar, IUChar) -> true
-    | (IBool, IBool) -> true
-    | (IInt, IInt) -> true
-    | (IUInt, IUInt) -> true
-    | (IShort, IShort) -> true
-    | (IUShort, IUShort) -> true
-    | (ILong, ILong) -> true
-    | (IULong, IULong) -> true
-    | (ILongLong, ILongLong) -> true
-    | (IULongLong, IULongLong) -> true
-    | (IInt128, IInt128) -> true
-    | (IUInt128, IUInt128) -> true
-    | _ -> false;;
-
-  let equal (x: t) (y: t) = _equal (x,y)
+  let equal (x: t) (y: t) = x = y
   let compare (x: t) (y: t) = Stdlib.compare x y
-  let hash (x: t) = Hashtbl.hash x
 
   (* Output *)
   let pretty () x = d_ikind () x
@@ -139,23 +136,20 @@ module Fkind: S with type t = fkind =
 struct
   include Std
 
-  type t = fkind
+  type t = fkind =
+    | FFloat
+    | FDouble
+    | FLongDouble
+    | FComplexFloat
+    | FComplexDouble
+    | FComplexLongDouble
+  [@@deriving hash]
 
   let name () = "fkind"
 
   (* Identity *)
-  let _equal (xy: t * t) = match xy with
-    | (FFloat, FFloat) -> true
-    | (FDouble, FDouble) -> true
-    | (FLongDouble, FLongDouble) -> true
-    | (FComplexFloat, FComplexFloat) -> true
-    | (FComplexDouble, FComplexDouble) -> true
-    | (FComplexLongDouble, FComplexLongDouble) -> true
-    | _ -> false;;
-
-  let equal (x: t) (y: t) = _equal (x,y)
+  let equal (x: t) (y: t) = x = y
   let compare (x: t) (y: t) = Stdlib.compare x y
-  let hash (x: t) = Hashtbl.hash x
 
   (* Output *)
   let pretty () x = d_fkind () x
@@ -171,20 +165,17 @@ module Unop: S with type t = unop =
 struct
   include Std
 
-  type t = unop
+  type t = unop =
+    | Neg
+    | BNot
+    | LNot
+  [@@deriving hash]
 
   let name () = "unop"
 
   (* Identity *)
-  let _equal (xy: t * t) = match xy with
-    | (Neg, Neg) -> true
-    | (BNot, BNot) -> true
-    | (LNot, LNot) -> true
-    | _ -> false;;
-
-  let equal (x: t) (y: t) = _equal (x,y)
+  let equal (x: t) (y: t) = x = y
   let compare (x: t) (y: t) = Stdlib.compare x y
-  let hash (x: t) = Hashtbl.hash x
 
   (* Output *)
   let pretty () x = d_unop () x
@@ -200,39 +191,36 @@ module Binop: S with type t = binop =
 struct
   include Std
 
-  type t = binop
+  type t = binop =
+    | PlusA
+    | PlusPI
+    | IndexPI
+    | MinusA
+    | MinusPI
+    | MinusPP
+    | Mult
+    | Div
+    | Mod
+    | Shiftlt
+    | Shiftrt
+    | Lt
+    | Gt
+    | Le
+    | Ge
+    | Eq
+    | Ne
+    | BAnd 
+    | BXor 
+    | BOr
+    | LAnd
+    | LOr
+  [@@deriving hash]
 
   let name () = "binop"
 
   (* Identity *)
-  let _equal (xy: t * t) = match xy with
-      | (PlusA, PlusA) -> true
-      | (PlusPI, PlusPI) -> true
-      | (IndexPI, IndexPI) -> true
-      | (MinusA, MinusA) -> true
-      | (MinusPI, MinusPI) -> true
-      | (MinusPP, MinusPP) -> true
-      | (Mult, Mult) -> true
-      | (Div, Div) -> true
-      | (Mod, Mod) -> true
-      | (Shiftlt, Shiftlt) -> true
-      | (Shiftrt, Shiftrt) -> true
-      | (Lt, Lt) -> true
-      | (Gt, Gt) -> true
-      | (Le, Le) -> true
-      | (Ge, Ge) -> true
-      | (Eq, Eq) -> true
-      | (Ne, Ne) -> true
-      | (BAnd, BAnd) -> true
-      | (BXor, BXor) -> true
-      | (BOr, BOr) -> true
-      | (LAnd, LAnd) -> true
-      | (LOr, LOr) -> true
-      | _ -> false;;
-
-  let equal (x: t) (y: t) = _equal (x,y)
+  let equal (x: t) (y: t) = x = y
   let compare (x: t) (y: t) = Stdlib.compare x y
-  let hash (x: t) = Hashtbl.hash x
 
   (* Output *)
   let pretty () x = d_binop () x
@@ -248,20 +236,17 @@ module Wstring_type: S with type t = wstring_type =
 struct
   include Std
 
-  type t = wstring_type
+  type t = wstring_type =
+    | Wchar_t
+    | Char16_t
+    | Char32_t
+  [@@deriving hash]
 
   let name () = "wstring_type"
 
   (* Identity *)
-  let _equal (xy: t * t) = match xy with
-    | (Wchar_t, Wchar_t) -> true
-    | (Char16_t, Char16_t) -> true
-    | (Char32_t, Char32_t) -> true
-    | _ -> false;;
-
-  let equal (x: t) (y: t) = _equal(x,y)
+  let equal (x: t) (y: t) = x = y
   let compare (x: t) (y: t) = Stdlib.compare x y
-  let hash (x: t) = Hashtbl.hash x
 
   (* Output *)
   let show = function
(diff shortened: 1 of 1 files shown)


src/util/cilType.ml
```diff
@@ -104,9 +104,24 @@ struct
   let name () = "ikind"
 
   (* Identity *)
-  (* Enum type, so polymorphic identity is fine. *)
-  (* Monomorphize polymorphic operations for optimization. *)
-  let equal (x: t) (y: t) = x = y
```
sim642, line 109: I'm surprised that the old and usual way didn't achieve proper monomorphization. Do you know why it doesn't work? Maybe looking at the produced code (either assembly or OCaml IR) explains this?

sim642: I have one idea that might be worth trying out. One could change the type definition above to something like
```ocaml
type t = ikind =
  | IChar
  | ISChar
  | ...
```
Maybe then the old form actually does monomorphization because the possible constructors are available in this `.ml` file, not just the type alias.

Of course at that point one could also derive all these (including `compare` and `hash`).

adelavais: I made the modification as you suggested in all the 6 types we targeted, one example:
```
  type t = wstring_type =
    | Wchar_t
    | Char16_t
    | Char32_t

  let name () = "wstring_type"

  (* Identity *)
  let equal (x: t) (y: t) = x = y
...
``` 
but the results seem to be worse. I added a new sheet, named `update`, [here](https://docs.google.com/spreadsheets/d/1CuXGHs89sDkrOyH9B8ouraDotZfsnamlHlUP2Q8mt2A/edit?usp=sharing). Now the best case is -7% run time decrease and worst is 5% increase. It is generally worse.

(As an aside, I changed my system and I rebased the PR over `master`, so this is why we have new values for the original program too.)

> I'm surprised that the old and usual way didn't achieve proper monomorphization.

I am unsure why it doesn't work properly, and I will look into the produced code tomorrow.

sim642: When writing out the constructors again, you can also try using `[@@deriving eq, ord, hash]` instead of manually writing `equal`, `compare` and `hash` at all. The derivers essentially generate the same code as you have currently hand-written, so it should still be as fast. It would be better though, because such hand-written match is error-prone, especially if new constructors are added in CIL.

adelavais: Regarding the difference between using `[@@deriving eq, ord, hash]` or manually writing equal, compare and hash, I used godbolt.org, with compiler `x86-64 ocamlopt 4.14.0-flambda` and option `-O3`. The assembly results are different. 

Derived:
```
module Ikind =
struct
  type t = 
    | IChar
    | ISChar 
    | IUChar
    | IBool
    | IInt
    | IUInt
    | IShort
    | IUShort
    | ILong
    | IULong
    | ILongLong
    | IULongLong
    | IInt128
    | IUInt128
    [@@deriving eq, ord, hash] 
end
```
```
camlExample__data_begin:
camlExample__code_begin:
camlExample__gc_roots:
        .quad   0
camlExample:
        .quad   camlExample__Ikind_8
camlExample__Ikind_8:
camlExample__entry:
        movl    $1, %eax
        ret
camlExample__code_end:
camlExample__data_end:
        .quad   0
camlExample__frametable:
        .quad   0
```

Manually written:
```
module Ikind =
struct
  type t = 
    | IChar
    | ISChar 
    | IUChar
    | IBool
    | IInt
    | IUInt
    | IShort
    | IUShort
    | ILong
    | IULong
    | ILongLong
    | IULongLong
    | IInt128
    | IUInt128

  let equal (x: t) (y: t) = x = y
  let compare (x: t) (y: t) = Stdlib.compare x y
  let hash (x: t) = Hashtbl.hash x
end
```
```
camlExample__data_begin:
camlExample__code_begin:
camlExample__hash_36:
camlExample__hash_22_closure:
        .quad   camlExample__hash_22
        .quad   0x100000000000005
camlExample__equal_34:
camlExample__equal_6_closure:
        .quad   caml_curry2
        .quad   0x200000000000007
        .quad   camlExample__equal_6
camlExample__compare_35:
camlExample__compare_14_closure:
        .quad   caml_curry2
        .quad   0x200000000000007
        .quad   camlExample__compare_14
camlExample__gc_roots:
        .quad   0
camlExample__equal_6:
        cmpq    %rbx, %rax
        sete    %al
        movzbq  %al, %rax
        leaq    1(%rax,%rax), %rax
        ret
camlExample__compare_14:
        movq    %rax, %rdi
        cmpq    %rbx, %rdi
        setl    %al
        movzbq  %al, %rsi
        cmpq    %rbx, %rdi
        setg    %al
        movzbq  %al, %rax
        subq    %rsi, %rax
        leaq    1(%rax,%rax), %rax
        ret
camlExample__hash_22:
        subq    $8, %rsp
        movl    $1, %edx
        movl    $201, %esi
        movl    $21, %edi
        movq    %rax, %rcx
        call    caml_hash@PLT
        addq    $8, %rsp
        ret
camlExample:
        .quad   camlExample__Pmakeblock_37
camlExample__Pmakeblock_37:
        .quad   camlExample__equal_6_closure
        .quad   camlExample__compare_14_closure
        .quad   camlExample__hash_22_closure
camlExample__entry:
        movl    $1, %eax
        ret
camlExample__code_end:
camlExample__data_end:
        .quad   0
camlExample__frametable:
        .quad   0
```

jerhard: Have you compared the solution using `deriving` to the implementation you currently have, that uses the explicit match?

michael-schwarz: Does the derive even run on Godbolt? I don't see any code for functions in the derived case... Or am I missing something here?

sim642: > Does the derive even run on Godbolt? I don't see any code for functions in the derived case... Or am I missing something here?

Probably not, the ppx_deriving library needs to be available and also added into the compilation pipeline. If it's not there, then the attributes are simply not interpreted by the OCaml compiler itself.

It's possible to use `dune describe pp src/util/cilType.ml` to get a preprocessed source, which might be helpful.

adelavais: I ran the `dune describe pp src/util/cilType.ml` command and the results are:
- for the current PR
```
type t = encoding
     let name () = "encoding"
     let _equal (xy : (t * t)) =
       match xy with
       | (No_encoding, No_encoding) -> true
       | (Utf8, Utf8) -> true
       | _ -> false
     let equal (x : t) (y : t) = _equal (x, y)
     let compare (x : t) (y : t) = Stdlib.compare x y
     let hash (x : t) = Hashtbl.hash x
```
- for the version using manually written functions
```
type t = encoding =
       | No_encoding 
       | Utf8 
     let name () = "encoding"
     let equal (x : t) (y : t) = x = y
     let compare (x : t) (y : t) = Stdlib.compare x y
     let hash (x : t) = Hashtbl.hash x
```
- for the version using `deriving`
```
type t = encoding =
       | No_encoding 
       | Utf8 [@@deriving (eq, ord, hash)]
     let rec equal : t -> t -> Ppx_deriving_runtime.bool =
       ((let open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"]) in
           fun lhs ->
             fun rhs ->
               match (lhs, rhs) with
               | (No_encoding, No_encoding) -> true
               | (Utf8, Utf8) -> true
               | _ -> false)
       [@ocaml.warning "-A"])[@@ocaml.warning "-39"]
     let rec compare : t -> t -> Ppx_deriving_runtime.int =
       ((let open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"]) in
           fun lhs ->
             fun rhs ->
               match (lhs, rhs) with
               | (No_encoding, No_encoding) -> 0
               | (Utf8, Utf8) -> 0
               | _ ->
                   let to_int = function | No_encoding -> 0 | Utf8 -> 1 in
                   Ppx_deriving_runtime.compare (to_int lhs) (to_int rhs))
       [@ocaml.warning "-A"])[@@ocaml.warning "-39"]
     include
       struct
         let _ = fun (_ : t) -> ()
         let rec (hash : t -> int) =
           ((let open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"]) in
               function | No_encoding -> 0 | Utf8 -> 1)
           [@ocaml.warning "-A"])[@@ocaml.warning "-39"]
         let _ = hash
       end[@@ocaml.doc "@inline"][@@merlin.hide ]
```

sim642: The idea would be to now check if the `deriving` version gives the same assembly as the hand-written one. Hopefully it does and we can use that instead of the hand-written versions.

adelavais: I am having trouble both locally and on Godbolt to compile the code [as is](https://github.com/goblint/analyzer/pull/956#discussion_r1083539333). Godbolt does not have the `ppx_deriving` library, and locally, even if I add the library to the compilation, the functions do not appear in the generated `.s` file.

 If I use Godbolt and I modify the code to not use `Ppx_deriving_runtime` directly, I get, different results for the `equal` function:
- for the derived version:
```
module Encoding  =
struct
type t  =
       | No_encoding 
       | Utf8
     let rec equal : t -> t -> bool =
           fun lhs ->
             fun rhs ->
               match (lhs, rhs) with
               | (No_encoding, No_encoding) -> true
               | (Utf8, Utf8) -> true
               | _ -> false
end
```
```
camlExample__equal_6:
        cmpq    $1, %rax
        je      .L101
        cmpq    $1, %rbx
        je      .L100
        movl    $3, %eax
        ret
.L101:
        cmpq    $1, %rbx
        jne     .L100
        movl    $3, %eax
        ret
.L100:
        movl    $1, %eax
        ret
```
- for the manually written version:
```
let equal (x: t) (y: t) = x = y
```
```
camlExample__equal_6:
        cmpq    %rbx, %rax
        sete    %al
        movzbq  %al, %rax
        leaq    1(%rax,%rax), %rax
        ret
```
For the compare function I had to modify more, the original last 2 matching patterns were creating type errors (expecting `t` type but got `int`), so this is why I am unsure if the next part is 100% correctly modified:
- deriving
```
let rec compare : t -> t -> int =
           fun lhs ->
             fun rhs ->
               match (lhs, rhs) with
               | (No_encoding, No_encoding) -> 0
               | (Utf8, Utf8) -> 0
               | (No_encoding, Utf8) -> -1
               | (Utf8, No_encoding) -> 1
```
```
        cmpq    $1, %rax
        je      .L104
        movl    $4, %eax
        subq    %rbx, %rax
        ret
.L104:
        cmpq    $1, %rbx
        je      .L103
        movq    $-1, %rax
        ret
.L103:
        movl    $1, %eax
        ret
```
- manually written
```
        movq    %rax, %rdi
        cmpq    %rbx, %rdi
        setl    %al
        movzbq  %al, %rsi
        cmpq    %rbx, %rdi
        setg    %al
        movzbq  %al, %rax
        subq    %rsi, %rax
        leaq    1(%rax,%rax), %rax
        ret
```
The `hash` function of the `deriving` version  
```
include
       struct
         let _ = fun (_ : t) -> ()
         let rec (hash : t -> int) =
               function | No_encoding -> 0 | Utf8 -> 1
         let _ = hash
       end
```
or even just 
```
let rec (hash : t -> int) =
               function | No_encoding -> 0 | Utf8 -> 1
```
compiles to just `ret`. It does compile; changing the `Utf8` return value to, say, `5`, changes the assembly code. Manually written, this function needs a `caml_hash@PLT` call:
```
camlExample__hash_22:
        subq    $8, %rsp
        movl    $1, %edx
        movl    $201, %esi
        movl    $21, %edi
        movq    %rax, %rcx
        call    caml_hash@PLT
        addq    $8, %rsp
        ret
```
I don't think my changes would have impacted too much the way the assembly code is generated, so I don't think the two versions generate the same assembly.

sim642: I guess here you meant the old version which this PR removes to be "manually written". What about the new manually written version which this PR adds as it is right now?
The OCaml code for that should be quite close to the derived one, so they should be more similar.

adelavais: Sorry for the confusion. By manually written, I mean the second modification I made, where I added the constructors in the file but kept manually written functions.
```
type t =
| No_encoding
| Utf8
let equal (x: t) (y: t) = x = y
let compare (x: t) (y: t) = Stdlib.compare x y
let hash (x: t) = Hashtbl.hash x

```

sim642: I did some experimenting with Godbolt also myself now and it seems like the following produces the shortest and most direct assembly:
```ocaml
  type t = ikind =
    | IChar       (** [char] *)
    | ISChar      (** [signed char] *)
    | IUChar      (** [unsigned char] *)
    | IBool       (** [_Bool (C99)] *)
    | IInt        (** [int] *)
    | IUInt       (** [unsigned int] *)
    | IShort      (** [short] *)
    | IUShort     (** [unsigned short] *)
    | ILong       (** [long] *)
    | IULong      (** [unsigned long] *)
    | ILongLong   (** [long long] (or [_int64] on Microsoft Visual C) *)
    | IULongLong  (** [unsigned long long] (or [unsigned _int64] on Microsoft
                      Visual C) *)
    | IInt128     (** [__int128] *)
    | IUInt128    (** [unsigned __int128] *)
  [@@deriving hash]
  let equal (x: t) (y: t) = x = y
  let compare (x: t) (y: t) = Stdlib.compare x y
```
With the constructors re-exported, monomorphization should actually be able to work for `equal` and `compare`, and output assembly which directly compares the constructor indices, as opposed to having a long sequence of conditional jumps arising from a big match (whether hand-written or derived).
`hash` doesn't seem to really monomorphize, so it's better to derive it, since the derived implementation is a function which maps constructors to their indices. The compiler seems to recognize that it's an identity function and also produces tiny assembly for it which doesn't call the polymorphic `Hashtbl.hash`. If any of the `hash` results are changed to a value different from its constructor index, it's no longer an identity function and compiles to what looks like a lookup table (which isn't bad either).

adelavais: I updated the pull request to export constructors and to derive the `hash` function (as described above).

sim642 (approved):
I added some comments about why these are the way they are now.

Given the back-and-forth changes, this should be **squash merged**.

michael-schwarz (approved):

PR #963, opened 2023-01-05
ESOP '23 Artifact Description

michael-schwarz:
This adds the ESOP '23 artifact description. I am somewhat opposed to merging https://github.com/goblint/bench/tree/esop23_artifact_new, as it contains a few sloppy things (such as copying scripts many times), which means also the scripts here should probably not be merged.

The question is if we should still merge the artifact description and just point people towards the VM, or not merge this at all.


sim642:
Well yeah, the scripts here aren't really useful as such, but I think we should add the artifact description since that's what we did last time.

docs/artifact-descriptions/esop23.md
```diff
@@ -0,0 +1,129 @@
...
+**For the Set "Watts":**
+ - For a detailed discussion on these benchmarks, see Appendix I.2 of the paper.
+ - As opposed to the other scripts, this one also prints run-times as these are needed to also verify **Table 4** in the supplementary material.
+
+Note: To regenerate just some of the results: Invoke one of `../bench/esop23_table2_set_{our,goblint,watts,ratcop}.rb`.
+
+### Reproducing Duet numbers
+
+This artifact ships Duet (at commit `5ea68373bb8c8cff2a9b3a84785b12746e739cee`) with a bug fix (courtesy of its original author Zach Kincaid) allowing it to run successfully on some of the benchmarks.
+For others, it sill reported a number of reachable asserts that is too low.
```
sim642, line 86:
```suggestion
For others, it still reported a number of reachable asserts that is too low.
```

michael-schwarz:
Remarks by AEC to address:

- [ ] Specify storage, compute, and memory requirements for setting up and running the VM. During a first import attempt there was an error code: `E_INVALIDARG (0x80070057)`, which turned out to be due to a lack of available disk space. There was 19 GB free after the 8 GB VM image download but an additional 21 GB was needed for the import.
- [ ] Include `The runtimes given for the commands in the artifact description are based on runtimes we had in the VM on a host system with an `11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz` with 32GB of RAM with some generous upwards rounding.`
- [ ] Include expected output of kick tires script.
- [ ] Clarify that kick tires script exercises relevant parts of the artifact.
- [ ] Include link back to GitHub repo for browsing source code.

PR #964, opened 2023-01-09
Iter optimization

montrie:
### PR introducing the Iter package

**This PR should be merged after the flambda PR #928 and the int-domain-optimization PR #943** 

We introduce the `Iter` package [https://github.com/c-cube/iter/](url) to replace certain `List` methods within `src/analyses/base.ml` and `src/solvers/generic.ml`

We performed a sanity check using `make test` that passed without errors. Additionally, we did some benchmarking using the following file/flags:
`./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable ana.int.def_exc --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.interval --enable ana.int.interval_threshold_widening`

Master branch goblint:
-   1229.096s -> base line

Flambda goblint:
-   1088.269s -> speed up: 11.46%
-   1109.058s -> speed up:  9.77%
-   1146.942s -> speed up:  6.68%

Iter + flambda goblint:  
-   881.201s -> speed up: 28.30%
-   913.384s -> speed up: 25.68% 
-   998.245s -> speed up: 18.78%

The measurements are done on a local machine. All speed ups are compared to the master branch goblint version.

Edit:
We also created a seperate branch called `seq-optimization` ([https://github.com/adelavais/analyzer/tree/seq-optimization](url)) that implements the same changes using OCaml's built-in `Seq` package. We'll perform benchmarks similar to the one mentioned above and compare the results with the one from this PR. Afterwards we can decide whether the `Iter` or the `Seq` package is better suited for goblint.


montrie:
After some more benchmarking on my local machine, I get the following runtimes for a similar Seq based implementation as seen in #965. The measurements were performed as mentioned above:

Seq + flambda goblint (compared to the baseline measurement of 1229.096s):

- 906.564s -> speedup: 26.24%
- 977.171s -> speedup: 20.50%
- 1073.392s -> speedup: 12.67%
- 1094.149s -> speedup: 10.98%


### Runtimes on the LRZ cluster
We did an additional measurement of the Iter and Seq implementation on the **LRZ cluster**. The measured runtimes are as follows:

- Iter + flambda: 1177,313s
- Seq + flambda: 1173,092s

The runtimes of the Iter and the Seq packages on the cluster are pretty much the same. As per [http://ocamlverse.net/content/iterators.html](url) however, we would expect the combination of `Iter` and flambda's optimization flag `-O3` to achieve better performances than than `Seq` + `-O3`

montrie:
Based on the above measurements, there is _no obvious choice_ for which package to use at the moment. Nevertheless, there are two points to consider:

1. **Is it worthwhile to require the installation of the Iter package given the current measurements?** According to [http://ocamlverse.net/content/iterators.html](url), `Iter` _should_ give us better performances in combination with flambda's `-O3` optimization.
2. **Will goblint implement more complex/longer chains of List manipulation operations?** Since Iter omits data allocation for intermediate List creation in a chain of `Iter` function calls, we can assume that Iter would show better performances in certain cases.

sim642:
If performance-wise they should be more or less the same, then at least I would slightly prefer `Seq` as a more standard thing. We already use `BatEnum` to avoid intermediate lists in many places and `Seq` is essentially the same. It might even be good to slowly move to using `Seq` at those places as well.

Iter is conceptually more efficient, but if the `Seq` overhead is irrelevant in our usage, it might not be worth it to introduce yet another style of iteration in the Goblint codebase.

montrie:
> Iter is conceptually more efficient, but if the `Seq` overhead is irrelevant in our usage, it might not be worth it to introduce yet another style of iteration in the Goblint codebase.

That's what I'm thinking as well, it doesn't seem worth it _at the moment_ to introduce the `Iter` package to goblint.

make.sh
```diff
@@ -11,6 +11,14 @@ opam_setup() {
   opam switch -y create . --deps-only ocaml-base-compiler.4.14.0 --locked
 }
 
+opam_setup_flambda() {
+  set -x
+  opam init -y -a --bare $SANDBOXING # sandboxing is disabled in travis and docker
+  opam update
+  # Note: the `--update-invariant` option is needed for replacement of the previous ocaml compiler switch invariant of Goblint
+  opam switch -y create . --deps-only ocaml-variants.4.14.0+options ocaml-option-flambda --locked --update-invariant
+}
```
jerhard, line 20: Are these changes necessary or are they subsumed by the merged `flambda` PR?

src/analyses/base.ml
```diff
@@ -823,7 +824,7 @@ struct
           | [] -> assert false
           | (e1, e2) :: eqs ->
             let eqs_for_all_mem e = List.for_all (fun (e1, e2) -> CilType.Exp.(equal e1 e || equal e2 e)) eqs in
-            let eqs_map_remove e = List.map (fun (e1, e2) -> if CilType.Exp.equal e1 e then e2 else e1) eqs in
+            let eqs_map_remove e = Iter.on_list (Iter.map (fun (e1, e2) -> if CilType.Exp.equal e1 e then e2 else e1)) eqs in
```
jerhard, line 827: Does changing a single `List` operation, as here, improve performance?

jerhard (commented):
Thanks for your PR! It's good nice that you have taken the time to investigate both approaches using `Seq` and `Iter`.  I see that you have some found additonal places with chained List operations. 

There are some places where a `List.map` is replaced by `Iter.on_list (Iter.map ...`. Can such a change improve performance?

This PR uses  a different version of the configuration optimization compared to #965.
Is there a conclusion on which approach to optimizing the configuration yields a larger improvement?

montrie:
> This PR uses a different version of the configuration optimization compared to #965. It would make sense to decide what changes to the configuration handling one would like to have, and keep them in a separate PR.

Yes, this PR is based on a version of #965 where we hadn't yet decided how to do the configuration handling. We'll have that sorted out and adjust this PR accordingly.

make.sh
montrie: These changes are indeed redundant after the merged `flambda` PR.

src/analyses/base.ml
montrie: Unfortunately, these single List operations don't improve performance. We added them anyway thinking that maybe flambda can do something clever with that.
I'd suggest they can/should be removed before merging this branch.

montrie:
I'll close this PR in favor of a cleaner version in #974.

PR #966, opened 2023-01-13
Add Interval Set domain for Interval Set analysis

### LINKED ISSUE #901: Add a new intDomain that deals with sets of intervals
# Motivation

The actual state of mutually refining integer domains does not allow representing the union of non-overlapping intervals such as  {[0, 3] [7, 10]}

If we had this information, we could improve our analyses by detecting unreachable parts of the code.
[](https://postimg.cc/hhbycjLL)

and hence improving our overflow detection.

# Computational Challenge:

The naive approach of interval sets (i.e. an arbitrarily big set of intervals) is computationally problematic, given that theoretically the number of intervals we store in the set could grow quadratically after applying each operation. A solution to this would be to apply (a possibly parametrized) upper bound to the set cardinality. This should add more complexity to the way the operations are implemented. That is, if the naive operation returns many more intervals than the upper bound, we will have to sacrifice some information to respect the upper bound and avoid an explosion of the number of intervals in the set.

# TODOs:
- [ ] Implement the S interface (while getting inspiration from the Interval domain implementation) including the domain concrete representation and the different operations (arithmetic ops, constructors, lattice ops)
- [ ] Refine the implementation towards being more efficient and keeping a good amount of information (maybe lookup for literature documenting relatable projects/experiences)
 - [ ] Implement regression tests
- [ ] add new flag options in order to run the interval
- [ ] adjust the base.ml analysis accordingly to benefit from the refinement that our implementation requires.

gabryon99:
The following pull request will include the *Interval Set Analysis* into Goblint, as described by the issue #901. 

Closes #901.


src/analyses/tutorials/signs.ml
```diff
@@ -9,6 +9,8 @@ struct
   include Printable.Std
 
   type t = Neg | Zero | Pos [@@deriving eq, ord, hash, to_yojson]
+
+  let elems = [Neg;Zero;Pos]
```
sim642, line 13: Solutions to the tutorials should be removed :)

src/util/precisionUtil.ml
```diff
@@ -1,19 +1,20 @@
...
-let max_int_precision : int_precision = (true, true, true, true)
+let max_int_precision : int_precision = (true, true, true, true, true)
 let max_float_precision : float_precision = (true)
 let int_precision_from_fundec (fd: GoblintCil.fundec): int_precision =
   ((ContextUtil.should_keep ~isAttr:GobPrecision ~keepOption:"ana.int.def_exc" ~removeAttr:"no-def_exc" ~keepAttr:"def_exc" fd),
    (ContextUtil.should_keep ~isAttr:GobPrecision ~keepOption:"ana.int.interval" ~removeAttr:"no-interval" ~keepAttr:"interval" fd),
    (ContextUtil.should_keep ~isAttr:GobPrecision ~keepOption:"ana.int.enums" ~removeAttr:"no-enums" ~keepAttr:"enums" fd),
-   (ContextUtil.should_keep ~isAttr:GobPrecision ~keepOption:"ana.int.congruence" ~removeAttr:"no-congruence" ~keepAttr:"congruence" fd))
+   (ContextUtil.should_keep ~isAttr:GobPrecision ~keepOption:"ana.int.congruence" ~removeAttr:"no-congruence" ~keepAttr:"congruence" fd),
+   (ContextUtil.should_keep ~isAttr:GobPrecision ~keepOption:"ana.int.interval_set" ~removeAttr:"no-interval" ~keepAttr:"interval_set" fd))
```
sim642, line 17:
```suggestion
   (ContextUtil.should_keep ~isAttr:GobPrecision ~keepOption:"ana.int.interval_set" ~removeAttr:"no-interval_set" ~keepAttr:"interval_set" fd))
```

sim642: Also these attributes should be added to the documentation.

tests/regression/61-interval-set-one/0-was_problematic_2.c
```diff
@@ -0,0 +1,24 @@
+// PARAM: --enable ana.int.interval_set  --set ana.base.arrays.domain partitioned
```
sim642, line 1: Our test naming is 0-padded, so this should be `00-was_problematic_2.c`, etc.

goblint.opam
```diff
@@ -20,7 +20,7 @@ doc: "https://goblint.readthedocs.io/en/latest/"
 bug-reports: "https://github.com/goblint/analyzer/issues"
 depends: [
   "ocaml" {>= "4.10"}
-  "dune" {>= "3.0"}
+  "dune" {>= "2.9" & >= "3.0"}
```
sim642, line 23: There are conflicts with `master` that need to be resolved.

sim642 (changes requested):
Thanks for the implementation with lots of tests!
It would be good to have an SV-COMP run with this enabled as well, if that hasn't already been done. That could help identify possible unsoundness errors.

I just had a quick look right now, didn't look at the actual domain implementation yet.

michael-schwarz:
> It would be good to have an SV-COMP run with this enabled as well, if that hasn't already been done. That could help identify possible unsoundness errors.

Indeed, I will start such a run once some of the issues we already identified (e.g. missing threshold widening) are addressed.

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+        let lcy =
+          if Ints_t.equal y max_ik then y else
+            Ints_t.sub y (modulo (Ints_t.sub y c) (Ints_t.abs m)) in
+        if Ints_t.compare rcx lcy > 0 then []
+        else if Ints_t.equal rcx lcy then norm ik @@ Some (rcx, rcx)
+        else norm ik @@ Some (rcx, lcy)
+    | _ -> []
+
+  let refine_with_congruence ik (intvs: t) (cong: (int_t * int_t ) option): t =
+    List.map (fun x -> Some x) intvs |> List.map (refine_with_congruence_interval ik cong) |> List.flatten
```
michael-schwarz, line 1341: There is no reason to pollute the module with `refine_with_congruence_interval` here, it should be defined locally inside `refine_with_congruence`.

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+    let intv1 = norm ik @@ Some (min, Ints_t.sub excl Ints_t.one) in
+    let intv2 = norm ik @@ Some (Ints_t.add excl Ints_t.one, max) in
+    intv1 @ intv2
+
+  let excl_to_intervalset (ik: ikind) ((rl, rh): (int64 * int64)) (excl: int_t): t = 
+    excl_range_to_intervalset ik (Ints_t.of_bigint (Size.min_from_bit_range rl),Ints_t.of_bigint (Size.max_from_bit_range rh)) excl
+
+  let of_excl_list ik (excls: int_t list) = 
+    let excl_list = List.map (excl_range_to_intervalset ik (range ik)) excls in
+    List.fold_left (meet ik) (top_of ik) excl_list
```
michael-schwarz, line 1359: Same here

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+module IntervalSetFunctor(Ints_t : IntOps.IntOps): S with type int_t = Ints_t.t and type t = (Ints_t.t * Ints_t.t) list =
+struct
+
+  module Interval_functor = IntervalFunctor(Ints_t) 
+
+  let name () = "IntervalSets"
+
+  type int_t = Ints_t.t
+
+  type t = (Ints_t.t * Ints_t.t) list [@@deriving eq, hash, ord]
```
michael-schwarz, line 935: Might be worth commenting here that these are given in ascending order and are free of overlaps.

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+   
+  let bot () = []
+  
+  let bot_of ik = bot () 
+
+  let show (x: t) =
+    let show_interval i = Printf.sprintf "[%s, %s]" (Ints_t.to_string (fst i)) (Ints_t.to_string (snd i)) in
+    List.fold_left (fun acc i -> (show_interval i) :: acc) [] x |> List.rev |> String.concat ", " |> Printf.sprintf "[%s]"
+  
+  let cartesian_product l1 l2 =
```
michael-schwarz, line 951: This function is used only once, please pull it down to where it is used.  Also, maybe `BatList.cartesian_product`?

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+        begin
+          match (x, y) with
+          | (Enter _, Exit _) -> -1
+          | (Exit _, Enter _) -> 1
+          | (_, _) -> 0
+        end
+    in
+    List.sort cmp
+        
+  let interval_set_to_events (xs: t) =
```
michael-schwarz, line 977: Does this need to work on unordered intervals or are the argument interval sets guaranteed to be sorted?

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+      Goblintutil.svcomp_may_overflow := true;
+    let sign = if signed then "Signed" else "Unsigned" in
+    match underflow, overflow with
+    | true, true ->
+      M.warn ~category:M.Category.Integer.overflow ~tags:[CWE 190; CWE 191] "%s integer overflow and underflow" sign
+    | true, false ->
+      M.warn ~category:M.Category.Integer.overflow ~tags:[CWE 191] "%s integer underflow" sign
+    | false, true ->
+      M.warn ~category:M.Category.Integer.overflow ~tags:[CWE 190] "%s integer overflow" sign
+    | false, false -> assert false
```
michael-schwarz, line 1047: Can we get reuse for this?

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+        end
+      else 
+        [(x,y)]
+    end
+
+  let leq (xs: t) (ys: t) = match xs, ys with
+    | [], _ -> true
+    | _, [] -> false
+    | _::_, _::_ -> let leq_interval = fun (al, au) (bl, bu) -> Ints_t.compare al bl >= 0 && Ints_t.compare au bu <= 0 in
+      List.for_all (fun x -> List.exists (fun y -> leq_interval x y) ys) xs
```
michael-schwarz, line 1090: This looks overly expensive, I don't think this operation needs to be quadratic!

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+    match underflow, overflow with
+    | true, true ->
+      M.warn ~category:M.Category.Integer.overflow ~tags:[CWE 190; CWE 191] "%s integer overflow and underflow" sign
+    | true, false ->
+      M.warn ~category:M.Category.Integer.overflow ~tags:[CWE 191] "%s integer underflow" sign
+    | false, true ->
+      M.warn ~category:M.Category.Integer.overflow ~tags:[CWE 190] "%s integer overflow" sign
+    | false, false -> assert false
+  
+  let norm ?(cast=false) ik = function
```
michael-schwarz, line 1049: `norm` is maybe an add name, as this does not `norm` the interval set representation but rather works on individual intervals.

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+    | (Some (la,ra), (lb,rb))::ts  when Ints_t.compare lb la < 0 -> (Some (la,ra),(min_ik,rb))::ts
+    | x  -> x)
+    |> List.rev
+    |> (function
+    | [] -> []
+    | (None,(lb,rb))::ts -> (None, (lb,max_ik))::ts
+    | (Some (la,ra), (lb,rb))::ts  when Ints_t.compare ra rb < 0 -> (Some (la,ra),(lb,max_ik))::ts
+    | x  -> x)
+    |> List.rev
+    |> List.map snd
```
michael-schwarz, line 1286: Could you describe the idea behind your algorithm here?

michael-schwarz (changes requested):
Thank you for the PR, it is nice to see progress here!

I left a few comments, I think there are some places where we can make things easier to follow by restructuring/reuse, and in a few other places I think we can do better than quadratic complexity.

src/util/processPool.ml
```diff
@@ -22,7 +22,7 @@ let run ~jobs ?(terminated=fun _ _ -> ()) tasks =
           Unix.open_process task.command
       in
       let pid = Unix.process_pid proc in
-      Catapult.Tracing.a_begin ~id:(string_of_int pid) "ProcessPool" ~args:[("command", `String task.command)];
+      Catapult.Tracing.a_begin ~id:(string_of_int pid) ~cat:[] "ProcessPool" ~args:[("command", `String task.command)];
```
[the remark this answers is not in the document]
sim642: This is a previous compatibility fix from master. I guess it somehow shows in this diff because conflicts prevent an up to date diff from being shown.

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+  let cast_to ?torg ?no_ov ik x = List.map (fun x -> norm ~cast:true ik (Some x)) x |> List.flatten
+
+  let leq_interval x y = match x, y with
+  | None, _ -> true
+  | Some _, None -> false
+  | Some (x1,x2), Some (y1,y2) -> Ints_t.compare x1 y1 >= 0 && Ints_t.compare x2 y2 <= 0
+
+  let join_interval ik x y = match x, y with
+  | None, z | z, None -> z
+  | Some (x1,x2), Some (y1,y2) -> Some (Ints_t.min x1 y1, Ints_t.max x2 y2)
```
jerhard, line 1240: Can you reuse `Interval.leq` and `Interval.join` instead?

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+
+  let rec interval_sets_to_partitions (ik: ikind) (acc : (int_t * int_t) option) (xs: t) (ys: t)=
```
jerhard, line 1242: What does this functions compute? 
If the `acc` is always going to be `None` when this function is called from the outside, you may also hide the parameter, by defining the recursive function locally.

src/cdomains/intDomain.ml
```diff
@@ -923,6 +923,461 @@ struct
...
+
+  let maximal xs = xs |> List.rev |> (function [] -> None | (_, y)::_ -> Some y)
+
+  let of_interval ik (x, y) = norm ik @@ Some (x, y)
+
+  let of_int ik (x: int_t) = of_interval ik (x, x)
+
+  let of_bool _ik = function true -> one | false -> zero
+  
+  let invariant_ikind_interval e ik x = match x with
```
jerhard, line 1302: Is this different from `Interval.invariant_ikind`?

tests/regression/64-interval-set-one/test.py
```diff
@@ -0,0 +1,6 @@
+import os
+for index, filename in enumerate(os.listdir(".")):
+    if ".c" in filename:
+        new_filename = f"{index}{filename[2:]}"
+        #print(new_filename)
+        os.rename(filename, new_filename)
```
michael-schwarz, line 6: I guess this was the script you used to make sure filenames are normalized. I don't think we should commit it though.

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+  let to_int = function [(x, y)] when Ints_t.compare x y = 0 -> Some x | _ -> None
+
+  let zero = [(Ints_t.zero, Ints_t.zero)]
+  let one = [(Ints_t.one, Ints_t.one)]
+
+  let top_bool = [(Ints_t.zero, Ints_t.one)]
+
+  let not_bool (x:t) = 
+    let is_false = fun x -> x = zero in
+    let is_true =  fun x -> x = one in
```
michael-schwarz, line 1129: Why define this with `fun` here?

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+              [Ints_t.max tl x, Ints_t.min tu y]
+            else
+              [range ik]
+          end
+        else 
+          [(x,y)]
+      end
+
+  let rec leq (xs: t) (ys: t) =
+    let leq_interval = fun (al, au) (bl, bu) -> Ints_t.compare al bl >= 0 && Ints_t.compare au bu <= 0 in
```
michael-schwarz, line 1101: why `fun`?

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+    events_to_intervals |>
+    remove_empty_gaps
+
+  let meet ik (x: t) (y: t): t = 
+    two_interval_sets_to_events x y |> 
+    combined_event_list  `Meet |> 
+    events_to_intervals |> 
+    remove_empty_gaps
+
+  let to_int = function [(x, y)] when Ints_t.compare x y = 0 -> Some x | _ -> None
```
michael-schwarz, line 1120: I guess this indentation makes it easier to understand:

```suggestion
  let to_int = function
  | [(x, y)] when Ints_t.compare x y = 0 -> Some x
  | _ -> None
```

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+    | _ -> top_of ik
+
+  let log1 f ik i1 = 
+    match interval_to_bool i1 with
+    | Some x -> of_bool ik (f ik x)
+    | _ -> top_of ik
+
+  let bit f ik (i1, i2) =
+    match (interval_to_int i1), (interval_to_int i2) with
+    | Some x, Some y -> (try norm_interval ik (Interval.of_int ik (f ik x y)) with Division_by_zero -> top_of ik)
```
michael-schwarz, line 1194: Why supply `ik` to `f` if you then jump through hoops to make all `f` ignore the `ik`? Am I missing something?

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+    | _ -> top_of ik
+
+  let bit1 f ik i1 =
+    match interval_to_int i1 with
+    | Some x -> of_int ik (f ik x)
+    | _ -> top_of ik
```
michael-schwarz, line 1200: Only used once? Also, why the `ik` for `f`?

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+
+  let bitcomp f ik (i1, i2) = 
+    match (interval_to_int i1, interval_to_int i2) with 
+    | Some x, Some y -> (try norm_interval ik (Interval.of_int ik (f ik x y)) with Division_by_zero | Invalid_argument _ -> top_of ik)
```
michael-schwarz, line 1204: Same thing with the `ik` here.

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+  let ne ik x y = not_bool @@ eq ik x y
+  let interval_to_int i = Interval.to_int (Some i)
+  let interval_to_bool i = Interval.to_bool (Some i)
+
+  let log f ik (i1, i2) = 
+    match (interval_to_bool i1, interval_to_bool i2) with
+    | Some x, Some y -> of_bool ik (f x y)
+    | _ -> top_of ik
+
+  let log1 f ik i1 =
```
michael-schwarz, line 1187: Only used once & the `ik` thing again.

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+      let x2y1 = (Ints_t.mul x2 y1) in let x2y2 = (Ints_t.mul x2 y2) in
+      norm_interval ik @@ Some ((Ints_t.min (Ints_t.min x1y1 x1y2) (Ints_t.min x2y1 x2y2)), (Ints_t.max (Ints_t.max x1y1 x1y2) (Ints_t.max x2y1 x2y2)))
+    in
+    binary_op x y interval_mul
+
+  let div ?no_ov ik x y = 
+    let rec interval_div ((x1, x2), (y1, y2)) = begin
+      let is_zero v = Ints_t.compare v Ints_t.zero = 0 in
+      match y1, y2 with
+      | l, u when is_zero l && is_zero u -> top_of ik (* TODO warn about undefined behavior *)
```
michael-schwarz, line 1267: What's up with this TODO? The warnings should indeed be generated here, as this domain may be more precise than intervals (same logic as for the overflows).

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+    | [], _ -> [] | _ ,[] -> xs
+    | _, _ ->
+      let min_xs = fst (List.hd xs) in
+      let max_xs = snd @@ BatList.last xs in
+      let min_ys = fst (List.hd ys) in
+      let max_ys = snd @@ BatList.last ys in
+      let min_range,max_range = range ik in
+      let min = if Ints_t.compare min_xs min_range == 0 then min_ys else min_xs in
+      let max = if Ints_t.compare max_xs max_range == 0 then max_ys else max_xs in
+      xs |> (function (_, y)::z -> (min, y)::z | _ -> []) |> List.rev |> (function (x, _)::z -> (x, max)::z | _ -> []) |> List.rev
```
michael-schwarz, line 1308: Maybe add a comment explaining the narrowing here?

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+      let t = List.find_opt (fun x -> Z.compare l x >= 0) ts in
+      BatOption.map_default Ints_t.of_bigint min_ik t
+    in
+    (*obtain partitioning of xs intervals according to the ys interval that includes them*)
+    let rec interval_sets_to_partitions (ik: ikind) (acc : (int_t * int_t) option) (xs: t) (ys: t)= 
+     match xs,ys with 
+    | _, [] -> []
+    |[], (y::ys) -> (acc,y):: interval_sets_to_partitions ik None [] ys 
+    |(x::xs), (y::ys) when  Interval.leq (Some x) (Some y) -> interval_sets_to_partitions ik (Interval.join ik acc (Some x)) xs (y::ys)
+    |(x::xs), (y::ys) -> (acc,y) :: interval_sets_to_partitions ik None  (x::xs) ys
```
jerhard, line 1339: Please leave a space after a "|"

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+        let new_c = function
+          | None -> Some (lower_threshold d, lower_threshold d)
+          | Some (cx,cy) -> Some (lower_threshold d, cy)
+        in
+      if threshold && Ints_t.compare (lower_threshold d) (upper_threshold b) > 1 then 
+        [(new_a a,(fst b, upper_threshold b)); (new_c c, (lower_threshold d, snd d))] else
+        [(Interval.join ik a c, (Interval.join ik (Some b) (Some d) |> Option.get))]
+    in
+    let partitions_are_approaching x y = match x, y with 
+        | (Some (_, ar), (_, br)), (Some (al, _), (bl, _)) -> Ints_t.compare (Ints_t.sub al ar) (Ints_t.sub bl br) > 0
```
jerhard, line 1357: Is this looking at four intervals at a time? Or do these tuples contain something else? Maybe one give more expressive names to `ar`, `al`, `br`, `bl` to make this more clear?

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+    unary_op x neg_interval
+
+  let sub ?no_ov ik x y = 
+    let interval_sub ((x1, x2), (y1, y2)) = norm_interval ik @@ Some (Ints_t.sub x1 y2, Ints_t.sub x2 y1) in
+    binary_op x y interval_sub
+
+  let mul ?no_ov (ik: ikind) (x: t) (y: t) : t = 
+    let interval_mul ((x1, x2), (y1, y2)) = 
+      let x1y1 = (Ints_t.mul x1 y1) in let x1y2 = (Ints_t.mul x1 y2) in
+      let x2y1 = (Ints_t.mul x2 y1) in let x2y2 = (Ints_t.mul x2 y2) in
```
jerhard, line 1258: Putting each definition in a separate line may make this clearer.

src/cdomains/intDomain.ml
```diff
@@ -919,22 +932,535 @@ struct
...
+      let max_xs = snd @@ BatList.last xs in
+      let min_ys = fst (List.hd ys) in
+      let max_ys = snd @@ BatList.last ys in
+      let min_range,max_range = range ik in
+      let min = if Ints_t.compare min_xs min_range == 0 then min_ys else min_xs in
+      let max = if Ints_t.compare max_xs max_range == 0 then max_ys else max_xs in
+      xs |> (function (_, y)::z -> (min, y)::z | _ -> []) |> List.rev |> (function (x, _)::z -> (x, max)::z | _ -> []) |> List.rev 
+
+  (*
+    1. partitions xs' intervals by assigning each of them to the an interval in ys that includes it.
```
jerhard, line 1311: Do you mean `xs` instead of `xs'`?

src/cdomains/intDomain.ml
```diff
@@ -911,30 +938,600 @@ struct
...
+    | Some x -> of_int ik (f ik x) |> unlift
+    | _ -> top_of ik
+
+  let bitcomp f ik (i1, i2) = 
+    match (interval_to_int i1, interval_to_int i2) with 
+    | Some x, Some y -> (try  of_int ik (f ik x y) with Division_by_zero | Invalid_argument _ -> (top_of ik,false,false,false))
+    | _, _ -> (top_of ik,true,true,false)
+
+  let bitand ik x y = 
+    let interval_bitand = bit (fun _ik -> Ints_t.bitand) ik in
```
jerhard, line 1254: You may redefine `bit` such that the `f` that it takes does not take an `ikind`.

src/cdomains/intDomain.ml
BilelGho: I believe we discussed this. Same TODO exists in IntervalFunctior. Let me know if I forgot something.

src/cdomains/intDomain.ml
```diff
@@ -911,30 +939,602 @@ struct
...
+    | [], _ -> [] | _ ,[] -> xs
+    | _, _ ->
+      let min_xs = minimal xs |> Option.get in
+      let max_xs = maximal xs |> Option.get in
+      let min_ys = minimal ys |> Option.get in
+      let max_ys = maximal ys |> Option.get in
+      let min_range,max_range = range ik in
+      let min = if Ints_t.compare min_xs min_range == 0 then min_ys else min_xs in
+      let max = if Ints_t.compare max_xs max_range == 0 then max_ys else max_xs in
+      xs |> (function (_, y)::z -> (min, y)::z | _ -> []) |> List.rev |> (function (x, _)::z -> (x, max)::z | _ -> []) |> List.rev
```
jerhard, line 1359: You may split this line.

michael-schwarz:
Closing in favor of #994.

PR #969, opened 2023-01-22
Value Domain: Rm code duplication caused by "smart" operations

michael-schwarz:
The "smart" `leq`, `join`, and `widen` that are needed for the partitioned array domain contained a lot of code duplication, and implementations even started to diverge (different error messages).

This removes duplication by delegating to the standard `leq`, `join`, `widen` for the cases where no special handling is needed.

### DIFF
--- a/src/cdomains/valueDomain.ml
+++ b/src/cdomains/valueDomain.ml
@@ -517,129 +517,6 @@ struct
       warn_type "join" x y;
       `Top
 
-  let rec smart_join x_eval_int y_eval_int  (x:t) (y:t):t =
-    let join_elem: (t -> t -> t) = smart_join x_eval_int y_eval_int in  (* does not compile without type annotation *)
-    match (x,y) with
-    | (`Top, _) -> `Top
-    | (_, `Top) -> `Top
-    | (`Bot, x) -> x
-    | (x, `Bot) -> x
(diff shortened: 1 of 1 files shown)


src/cdomains/valueDomain.ml
```diff
@@ -587,31 +550,37 @@ struct
...
 
+  let rec smart_join x_eval_int y_eval_int  (x:t) (y:t):t =
+    let join_elem: (t -> t -> t) = smart_join x_eval_int y_eval_int in  (* does not compile without type annotation *)
+    match (x,y) with
+    | (`Struct x, `Struct y) -> `Struct (Structs.join_with_fct join_elem x y)
+    | (`Union (f,x), `Union (g,y)) -> `Union (match UnionDomain.Field.join f g with
+        | `Lifted f -> (`Lifted f, join_elem x y) (* f = g *)
+        | x -> (x, `Top)) (* f <> g *)
+    | (`Array x, `Array y) -> `Array (CArrays.smart_join x_eval_int y_eval_int x y)
+    | (`Blob x, `Blob y) -> `Blob (Blobs.join x y)
```
sim642, line 561: The blob case looks the same, the smart version just had an extra comment before.

sim642 (approved):

src/cdomains/valueDomain.ml
michael-schwarz: Nice catch!

PR #970, opened 2023-01-22
Analysis of longjmp/setjmp

### LINKED ISSUE #887: Soundly handle `setjmp` and `longjmp`
There was an attempt in #764 but we are aiming at a more general solution here at TUM.

Pulled-out from #868

michael-schwarz:
PR to document the ongoing progress towards sound handling of longjmp/setjmp.

Corresponding draft for those with access to versioncontrolseidl: https://versioncontrolseidl.in.tum.de/schwarz/longjmp-and-setjmp

Issues that need to be handled:

- [x] Non-unique jump buffers
- [x] Jumping from several functions deep
- [x]  Jumping in recursive calls
- [x] Returning a value that is potentially zero in `longjmp` call
- [x] Warn for longjmps that propagate out of the current thread
- [x]  ...?
- Detecting misuse
     - [x] Jumping from different thread
     - [x] Accessing non-volatile local variable that has been potentially modified since call
     - [x] ~`setjmp` in invalid place (not directly used for branching)~ (deemed out of scope)
     - [x] `sigsetjmp`

Closes #887


michael-schwarz:
The failing test only fails in the CI but works locally for some reason.

michael-schwarz:
This still is a long way from being done, but I am now convinced our approach will actually work.  I'll try to focus on coming up with a reasonable outline / structure for the paper today.

In the meantime, if anyone is interested in contributing to the implementation, here are some of the things that are still open for this implementation:

- [ ] In `poisonVariables.ml` we currently only check for variables that appear directly in the expression, but we would also need to check if such a variable is read via pointers. Also, we should remove things from this poisonous set again when they are definitely overwritten.
- [ ] Just copying a jump_buf from one place to another and then calling setjmp on that is UB. We should enhance the base domain to mark jmp_buffers as copied when that happen and warn when setjmp is called.
- [ ] The structure is a bit ugly currently, with the `paths_as_set` transfer functions added to all analyses, and the `event` handling also added everywhere. Maybe there's a smarter way to achieve this @sim642? 

If you want to work on any of these things, please edit the comment to say so. I'll continue with whatever remains after I also have a decent draft of the paper.

michael-schwarz:
Currently, we seem to find T, i.e., `all jumpbuf` for all jumpbuffers appearing in the larger benchmark problems. Tomorrow, I will try a solution akin to the one we use for addresses, i.e., keep buffers we know about. Hopefully, this will fix this problem meaning we can claim that:

- It succeeds on crafted examples extracted from real programs
- Works on larger programs, but is hindered by our poor heap handling
- Flags a bug akin to the one there was in ImageMagick if it is inserted manually.

michael-schwarz:
The remaining CI failures here are on MacOS, with incremental & CFG comparison (CC @stilscher)

Probably something there is broken again, but I don't think this is particularly important here: We can finish the implementation and write the paper and fix these incremental weirdnesses later.

michael-schwarz:
For some reason the test fails in the CI only but works locally.

michael-schwarz:
Current state before allowing struct updates:

- Runs for 1.5h, ~40GB RAM
- Produces warnings that all jumptargets may be unknown, and 1 warning about an invalid setjmp
- 280 warnings accessing poisonous variables

michael-schwarz:
It appears as if most of the poisonous variable warnings are spurious, for some reason variables whose scope is left after a longjmp call, e.g. because the function is jumped out of, are also marked as poisonous and then propagated into new calls.

src/domains/events.ml
```diff
@@ -12,6 +12,7 @@ type t =
   | UpdateExpSplit of exp (** Used by expsplit analysis to evaluate [exp] on post-state. *)
   | Assert of exp
   | Unassume of {exp: CilType.Exp.t; uuids: string list}
+  | Poison of Queries.VS.t
```
sim642, line 15: Poison is a rather generic name. I'm guessing this refers to the notion of clobbering?

src/domains/events.ml
```diff
@@ -26,7 +27,8 @@ let emit_on_deadcode = function
   | Assign _
   | UpdateExpSplit _ (* Pointless to split on dead. *)
   | Unassume _ (* Avoid spurious writes. *)
-  | Assert _ -> (* Pointless to refine dead. *)
+  | Assert _ (* Pointless to refine dead. *)
+  | Poison _ -> (* TODO: correct? *)
```
sim642, line 31: This might depend on the precise meaning of poisoning.

src/witness/witness.ml
```diff
@@ -303,7 +303,9 @@ struct
         let i_str = string_of_int i in
         match n with
         | Statement stmt  -> Printf.sprintf "s%d(%d)[%s]" stmt.sid c_tag i_str
+        | LongjmpTo stmt  -> "" (* TODO: Correct? *)
         | Function f      -> Printf.sprintf "ret%d%s(%d)[%s]" f.svar.vid f.svar.vname c_tag i_str
+        | LongjmpFromFunction f -> "" (* TODO: Correct? *)
```
sim642, line 308: Probably not, because empty node name doesn't mean it won't be part of the ARG. But this is to be figured out after we see whether all this CFG change is strictly necessary.

src/witness/witnessConstraints.ml
```diff
@@ -149,6 +149,19 @@ struct
...
+  let map_event ctx e =
+    (* we now use Sync for every tf such that threadspawn after tf could look up state before tf *)
+    let h x (xs, sync) =
+      try
+        let x' = Spec.event (conv ctx x) e (conv ctx x) in
+        (Dom.add x' (step_ctx_edge ctx x) xs, Sync.add x' (SyncSet.singleton x) sync)
+      with Deadcode -> (xs, sync)
+    in
+    let d = Dom.fold_keys h (fst ctx.local) (Dom.empty (), Sync.bot ()) in
+    if Dom.is_bot (fst d) then raise Deadcode else d
```
sim642, line 162: I'll have to think about this sometime later. The meaning of events coming from outside the analysis is a wobbly concept, which is why `event` wasn't part of the `Spec` signature to begin with.

src/framework/node0.ml
```diff
@@ -7,16 +7,22 @@ type t =
...
   (** The statements as identified by CIL *)
   (* The stmt in a Statement node is misleading because nodes are program points between transfer functions (edges), which actually correspond to statement execution. *)
+  | LongjmpTo of CilType.Stmt.t
+  (** The statement as the target of a longjmp *)
   | FunctionEntry of CilType.Fundec.t
   (** *)
   | Function of CilType.Fundec.t
   (** The variable information associated with the function declaration. *)
+  | LongjmpFromFunction of CilType.Fundec.t
+  (** Handles longjumps out of the concerned function *)
```
sim642, line 17: Wasn't the idea to do an analysis without using program transformations? Or did I get the wrong impression?

It seems to me that these dynamic things shouldn't require CFG modifications, but should be doable with their existing call nodes and edges:
* `setjmp` returning twice could just be a join of the two things within the call transfer function instead of requiring two incoming edges that get joined by the CFG shape.
* `longjmp` being an early return should also be doable because we already do thread returns and thread return values just fine.

One could be calling `setjmp`/`longjmp` via function pointers, so CFG transformation cannot be complete. Wasn't there even some real-world example where a custom `longjmp` could be supplied (as a function pointer)?

src/util/goblintutil.ml
```diff
@@ -17,6 +17,8 @@ let svcomp_may_overflow = ref false
 (** The file where everything is output *)
 let out = ref stdout
 
+let assign_is_setjmp = ref false
```
sim642, line 20: This needs documentation at minimum.

src/framework/analyses.ml
```diff
@@ -451,13 +451,18 @@ sig
 
   val special : (D.t, G.t, C.t, V.t) ctx -> lval option -> varinfo -> exp list -> D.t
   val enter   : (D.t, G.t, C.t, V.t) ctx -> lval option -> fundec -> exp list -> (D.t * D.t) list
-  val combine : (D.t, G.t, C.t, V.t) ctx -> lval option -> exp -> fundec -> exp list -> C.t option -> D.t -> Queries.ask -> D.t
+  val combine : (D.t, G.t, C.t, V.t) ctx -> ?longjmpthrough:bool -> lval option -> exp -> fundec -> exp list -> C.t option -> D.t -> Queries.ask -> D.t
```
sim642, line 454: `longjmpthrough` (if really necessary at all) has no value in being an optional argument because every analysis must repeat the default value then. Worse yet, different analyses could even set the default differently.

src/analyses/threadReturn.ml
```diff
@@ -29,9 +29,13 @@ struct
     ctx.local
 
   let enter ctx (lval: lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
-    [ctx.local, false]
+    if ctx.edge = MyCFG.Skip && ctx.node = MyCFG.dummy_node then
+      (* We are inside enter_with inside a startfun, and thus the current function retruning is the main function *)
+      [ctx.local, true]
```
sim642, line 34: Shouldn't this already be covered by `let startstate v = true`?

src/analyses/modifiedSinceLongjmp.ml
```diff
@@ -0,0 +1,92 @@
...
+
+  let return ctx (exp:exp option) (f:fundec) : D.t =
+    ctx.local
+
+  let enter ctx (lval: lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
+    [ctx.local, D.bot ()]
+
+  let combine ctx ?(longjmpthrough = false) (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask:Queries.ask) : D.t =
+    let fromlv = if not longjmpthrough then relevants_from_lval_opt ctx lval else VS.empty () in
+    let taintedcallee = relevants_from_ls (f_ask.f Queries.MayBeTainted) in
```
sim642, line 54: This reminds me that our notion of "tainted" here is not its usual meaning (a la our taint analysis tutorial). We might have to improve this naming sometime separately.

src/analyses/poisonVariables.ml
```diff
@@ -0,0 +1,87 @@
...
+    | Question (b, t, f, _) -> check_exp tainted b; check_exp tainted t; check_exp tainted f
+  and check_lval ?(ignore_var = false) tainted lval = match lval with
+    | (Var v, offset) ->
+      if not ignore_var && not v.vglob && VS.mem v tainted then M.warn "accessing poisonous variable %a" d_varinfo v;
+      check_offset tainted offset
+    | _ -> () (* TODO: Consider mem *)
+  and check_offset tainted offset = match offset with
+    | NoOffset -> ()
+    | Field (_, o) -> check_offset tainted o
+    | Index (e, o) -> check_exp tainted e; check_offset tainted o
```
sim642, line 35: Seems like this would be a lot easier using a CIL visitor with just a single case for the variables.

src/framework/constraints.ml
```diff
@@ -635,19 +669,165 @@ struct
...
+                    in
+                    let r = S.event res_ctx (Events.Poison modified_vars) res_ctx in
+                    sidel (jmptarget node, ctx.context ()) r
+                  | _ -> failwith (Printf.sprintf "strange: %s" (Node.show node))
+                 )
+               else
+                 (if M.tracing then Messages.tracel "longjmp" "Longjmp to somewhere else, side-effect to %i\n" (S.C.hash (ctx.context ()));
+                  sidel (LongjmpFromFunction current_fundec, ctx.context ()) res))
+        in
+        List.iter (handle_longjmp) (JmpBufDomain.JmpBufSet.elements targets)
```
sim642, line 825: Does this not exist?
```suggestion
        JmpBufDomain.JmpBufSet.iter handle_longjmp targets
```

src/framework/constraints.ml
```diff
@@ -635,19 +669,165 @@ struct
...
+            if not (JmpBufDomain.JmpBufSet.mem (Target (node,c)) validBuffers) then
+              M.warn "Longjmp to potentially invalid target! (Target %s in Function %s which may have already returned or is in a different thread)" (Node.show node) (Node.find_fundec node).svar.vname
+            else
+              (let controlctx = ControlSpecC.hash (ctx.control_context ()) in
+               if c = IntDomain.Flattened.of_int (Int64.of_int controlctx) && (Node.find_fundec node).svar.vname = current_fundec.svar.vname then
+                 (if M.tracing then Messages.tracel "longjmp" "Potentially from same context, side-effect to %s\n" (Node.show node);
+                  match node with
+                  | Statement { skind = Instr [Call (lval, exp, args,_, _)] ;_ } ->
+                    let res' = match lval with
+                      | Some lv -> Goblintutil.assign_is_setjmp := true; let r = S.assign path_ctx lv value in Goblintutil.assign_is_setjmp := false; r
```
sim642, line 801: The resetting needs to be under `Fun.protect`

src/framework/constraints.ml
```diff
@@ -635,19 +669,165 @@ struct
       if M.tracing then M.traceu "combine" "combined local: %a\n" S.D.pretty r;
       r
     in
+    let handlelongjmp (cd,fc,fd) =
```
sim642, line 672: Every normal function call also handles longjmps? Maybe I need to read the paper.

My concern is that this really complicates `FromSpec`, which is already complicated enough. I'm wondering if maybe it's possible to do this stuff in a functor.

src/framework/node0.ml
michael-schwarz: This is not a CFG transformation, it just adds (unconnected!) new global(!) unknowns to the analysis (`ret'_f` is `LongjmpFromFunction`), `LongjmpTo` is needed because there can be no direct side-effect to unknowns that have a right-hand side due to a solver limitation.  Technically, this once again is a global queried during RHS evaluation, and not the CFG transformation splitting a `setjmp` into two different transfer functions as we describe in the paper. 

 I'm not sure if there is a better way to do this, as these unknowns need to live outside any particular analysis and outside any lifters, so I thought it might be best to add them to the CFG as unconnected nodes.

> setjmp returning twice could just be a join of the two things within the call transfer function instead of requiring two incoming edges that get joined by the CFG shape.

This is exactly what happens, though it has to happen at the top level, i.e., in `constraints.ml` and cannot happen in any particular analysis.

> longjmp being an early return should also be doable because we already do thread returns and thread return values just fine.

This also is done as a side-effect to an additional node. The new node is needed because this should not be specific to any particular analysis, so once again is handled in `constraints.ml`


> One could be calling setjmp/longjmp via function pointers, so CFG transformation cannot be complete. Wasn't there even some real-world example where a custom longjmp could be supplied (as a function pointer)?

Yes, there is also in the regression tests. Here, we first rely on `constraints.ml` resolving the potentially called functions. A call to `setjmp` or `longjmp` then is handled in `tf_special_call` as are any other calls to special functions via pointers.

tests/regression/66-longjmp/05-counting-return-one-method.c
```diff
@@ -0,0 +1,17 @@
+// PARAM: --enable ana.int.interval --enable ana.int.enums --enable exp.earlyglobs  --set solvers.td3.side_widen never
+#include <goblint.h>
+#include <setjmp.h>
+
+jmp_buf my_jump_buffer;
+
+int main(void)
+{
+    int count = setjmp(my_jump_buffer);
```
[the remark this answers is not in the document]
michael-schwarz: Yes, we could. On the other hand, that makes understanding bugs etc in the analysis harder, as one needs to think about `tmp___0` variables that are pulled out and such. I'll add a comment that many of these are not syntactically valid and this reasoning to some README file for the entire group.

src/analyses/base.ml
```diff
@@ -2286,27 +2322,34 @@ struct
...
       let add_globals (st: store) (fun_st: store) =
         (* Remove the return value as this is dealt with separately. *)
-        let cpa_noreturn = CPA.remove (return_varinfo ()) fun_st.cpa in
+        let cpa_noreturn =
+          if not longjmpthrough then
+            (* Remove the return value as this is dealt with separately. *)
+            CPA.remove (return_varinfo ()) fun_st.cpa
+          else
+            (* Keep the return value as this is not actually the return value but the thing supplied in longjmp *)
+            fun_st.cpa
```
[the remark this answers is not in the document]
michael-schwarz: > Does this not mean that the RETURN variable remains in local state with the setjmp return value? Or why is it necessary to keep it like this?

Yes, that is precisely what we want here such that the correct value gets set here:

https://github.com/goblint/analyzer/blob/af4ce3b6bd85e1f62861906fbeb5ef4b53c0d53e/src/framework/constraints.ml#L694-L702

src/domains/events.ml
michael-schwarz: Yes, though clobbering is somehow related to variables being placed in registers, which is not necessary for it to be Undefined Behavior in the C standard.

src/witness/witness.ml
michael-schwarz: Do nodes that are not of the CFG end up in the Arg at all? Also, what would an automaton-based witness for a function using longjmp look like?

src/framework/constraints.ml
michael-schwarz: > Every normal function call also handles longjmps? 

It needs to - among others to account for modifications to locals that might have happened inside some function before the call to the function that finally executed the `longjmp`.

src/analyses/base.ml
```diff
@@ -1685,7 +1696,7 @@ struct
      | _ -> ()
     );
     match lval with (* this section ensure global variables contain bottom values of the proper type before setting them  *)
-    | (Var v, offs) when AD.is_definite lval_val && v.vglob ->
```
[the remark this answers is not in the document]
michael-schwarz: I don't think it ever was necessary, it probably was just a leftover from some old code.

tests/regression/66-longjmp/18-simple-else.c
```diff
@@ -0,0 +1,39 @@
...
+   val = setjmp( env_buffer );
+
+   __goblint_check(global == 0); //UNKNOWN!
+   __goblint_check(global == 2); //UNKNOWN!
+   __goblint_check(global == 8); //FAIL!
+
+   if( val != 0 ) {
+      printf("Returned from a longjmp() with value = %i\n", val);
+      __goblint_check(val == 2);
+      __goblint_check(global == 2); //TODO (requires path-sensitivity distinguishing between returns)
```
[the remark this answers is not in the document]
michael-schwarz: > Seems like this would be fairly easy to achieve, given that setjmp returns at most twice

Actually, it may return more than twice. But yes, there should be no explosion of paths because of it.

src/util/goblintutil.ml
michael-schwarz: I was able to completely remove this.

tests/regression/66-longjmp/17-loop.c
```diff
@@ -0,0 +1,27 @@
+#include <stdio.h>
```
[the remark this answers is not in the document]
michael-schwarz: I added appropriate annotations.

michael-schwarz:
I now have the number of warnings for potentially poisonous variables down to 150, but there still is something strange going on where we see more variables then one would expect being marked as poisonous (e.g. somehow also things that are not in scope?).

src/framework/node0.ml
sim642: I see, I missed the fact that these aren't actually added into the CFG, but just defined in the CFG edge type. This is rather confusing though because all the CFG-handling code (e.g. Graphviz output) must handle constructs which never occur.

Arbitrary global constraint variables can be introduced as `Spec.V`. That not only includes individual analyses but also lifters and at the outer level here in `FromSpec` (which defines `GVar` of the constraint system from `Spec.V`). In fact, `FromSpec` already introduces a set of its own (write-only) constraint unknowns for some context lookups.
Something similar can be done for these extra things. It would be amazing if it could even be extracted to a lifter since that would avoid a 4-way `Either` of global constraint variables in `FromSpec`.

src/witness/witness.ml
sim642: Since they're never part of the CFG to begin with, they don't, but it's weird to have to handle them here at all.

As to what the witness would look like, probably nobody knows nor has thought of it. I guess conceptually there could be a direct edge from the `longjmp` to the `setjmp`, although in this simple manner it would violate the well-nesting of the function entry/exit edges in the witness automaton.

sim642:
> I now have the number of warnings for potentially poisonous variables down to 150, but there still is something strange going on where we see more variables then one would expect being marked as poisonous (e.g. somehow also things that are not in scope?).

There is `InvariantCil.var_is_in_scope`, but filtering with it after the fact would probably be a real hack.

src/framework/constraints.ml
michael-schwarz: I was able to completely remove `assign_is_setjmp` now.

michael-schwarz:
I started a run of `creduce` in the hope to get a smaller example for why locals that shouldn't be are in the tainted set here. I'll have to see how effective `creduce` is for such long-running programs.

src/framework/node0.ml
michael-schwarz: Come to think of it, these should probably not be added to `Spec.V` but instead be part of `LVar`, as are the entry points and return points of functions, as they also take values from the lattice of locals, and not from some dedicated ones for globals.

michael-schwarz: I looked at changing `VarF`, i.e. `LVar` but that is super ugly and propagates absolutely everywhere, so actually I now think making this modification to the Node type is the right call. These behave like nodes in every possible way, i.e., are enhanced with contexts to get unknowns taking values from `D.t`,  get side-effects via `sideg`, ...

So i think for this conceptual simplicity, I think having some cases where these two new types of nodes are to be ignored is a small price to pay.

michael-schwarz:
I now fixed an issue where for recursive functions, the poisoned set was overestimated. The other issue why poison spreads so eagerly may be due to the fact that we have largely disabled context-sensitivity for `pngtest`. Maybe this fix that means poison-information for unreachable variables is not propagted to callees also helps there.

src/analyses/base.ml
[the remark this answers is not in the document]
sim642: It seems to me that snippet is calling `return` again (which adds the `RETURN` variable) and immediately `combine` without `~longjmpthrough` (which immediately removes the `RETURN` variable), or am I missing something?
Also, could it not be simpler if the `longjmp` value isn't passed off as `RETURN` but some separate `varinfo`, which `combine` leaves untouched? Then the above distinction wouldn't be required and `LONGJMP_RETURN` (or whatever) would simply be passed as any other variable.

On another note, is there a reason that `return` has to be called on the exceptional return state each time instead of storing the result of `return` at that new constraint unknown, just like normal function return nodes store the result of `return`?

src/framework/node0.ml
sim642: > Come to think of it, these should probably not be added to `Spec.V` but instead be part of `LVar`, as are the entry points and return points of functions, as they also take values from the lattice of locals, and not from some dedicated ones for globals.

At this point the distinction between `LVar` and `GVar` is really just historical. The customizability of `GVar` is already used in numerous places for similar things, e.g. as I mentioned somewhere thread returns. Also the dead branch lifter just induces per-`Node.t` unknowns for itself. We now use side-effects for a lot more than just function entry and global value invariants, so the distinction is already blurred and we've used them in a more general way.

What I really don't want is the implementation details of an analysis to leak through the abstractions we have maintained. I might want to have a look at changing this myself, but that can wait until after the paper submission. Paper-wise it doesn't have to change anything but keeping a separation of concerns in the implementation is valuable for all future maintenance and development.

src/analyses/threadReturn.ml
sim642: I see. We even have `morphstate` for some hacky things like this, but that doesn't even help here because it's called before `enter`. This is probably another things that #178 can improve.

src/analyses/base.ml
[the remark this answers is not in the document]
michael-schwarz: Yes, that is what's going on here: The situation here is the following: We are inside a function f that has performed a setjmp and called function g that did the longjmp.

Then, we do the following:
- Call `S.return` for this call to `g` (removes locals of g, and leaves the return value alone, as we pass `None` here)
- Call `S.combine` without `longjmpthrough` as we are in the context where the call is handled, this combines the local state and the callee state to obtain  the new local state, which then is side-effected to the appropriate location.

If the handling is not in the current context, the same thing happens except that we pass `longjmpthrough` in order to leave the return variable alone.

In principle, one could also do the `S.return` for the function itself in the handling of `longjmp` if it is not inside the current procedure and then always perform a return for the procedure if the target is not the caller in combine. I think I decided against it for symmetry reasons inside the handling of `longjmp` calls, but I agree that maybe that is actually the nicer way to structure it.

michael-schwarz:
With the fixes that I did yesterday, I now have the poisonous variable warnings down to 26.

Also there is one warning about a longjmp potentially happening to top and two warnings about longjmps to functions that have potentially returned, and 33 warnings about variables that may be modified between setjmp and longjmp.

Also, I have gotten the runtime down to 50 minutes, with 35GB of memory required.

michael-schwarz:
Together with Julian, I was able to identify another cause of poison being spread to eagerly: Taking the address of a poisonous variable is fine and not in itself already a problem.

michael-schwarz:
We now got down to 16 poisonous variable warnings about 10 different variables, 1 warning about potentially jumping to a top jumpbuffer, and 2 warnings about jumping to potentially invalid buffers.

michael-schwarz:
It may be possible to get the number of warnings further down by introducing path-sensitivity in the return value of setjmp.

tests/regression/66-longjmp/18-simple-else.c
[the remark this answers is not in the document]
michael-schwarz: I added a test that does this!

michael-schwarz:
I now have the number of warnings for poisonous accesses down to 7 thanks to path-sensitivity in the return value of the setjmp call.

src/analyses/modifiedSinceLongjmp.ml
```diff
@@ -0,0 +1,89 @@
...
+    ctx.local
+
+  let return ctx (exp:exp option) (f:fundec) : D.t =
+    ctx.local
+
+  let enter ctx (lval: lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
+    [ctx.local, D.bot ()]
+
+  let combine ctx ?(longjmpthrough = false) (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask:Queries.ask) : D.t =
+    let fromlv = if not longjmpthrough then relevants_from_lval_opt ctx lval else VS.empty () in
```
sim642, line 50: I started trying to refactor things and found this other usage of `longjmpthrough` (in addition to the one in base). However, I don't understand what this is trying to achieve. Both when I hard-code true or false here, all the longjmp tests pass.

sim642:
## sv-benchmarks
I compared this branch to the state after #1014 and here's the slowdown:
image

Note that this is without changing the svcomp config at all, so none of the longjmp-related analyses are activated! So this slowdown is merely from the `FromSpec` changes it seems.

sim642:
Thinking deeper, I found a soundness issue in our implementation when a local `longjmp` is performed with 0. As just an assign is used, this does not get changed 1 one like it does in base's `special` on a non-local `longjmp`.

This is not so easy to fix, especially given how local jumps look up the target node instruction's lvalue to do the `assign`/`combine` directly into the `setjmp` lvalue. I'll try to refactor this to be closer to the paper (such that the value is temporarily held in the return  variable in the `longjmp`-to-unknown).
It seems to me that this would also allow getting rid of `longjmpthrough` to do things even more like the paper presents.

sim642:
> Actually there could be some spurious difference in this plot because this branch is in conflict with `master` so the comparison is not completely adequate. I will do an extra run from the merge base to isolate other changes better.

Comparing fe6672433 (last `master` merged into this PR) with 8c0af0770 (this PR), there's no such performance overhead.

I will also do a run with the new longjmp analyses activated for SV-COMP because we'll probably want to do that.

sim642:
> I will also do a run with the new longjmp analyses activated for SV-COMP because we'll probably want to do that.

At 8c0af0770, without any of the longjmp analyses activated vs with them activated, sv-benchmarks shows 17% slowdown still:
image

I haven't investigated this much, but it may simply be inherent to all the extra analyses used. It might be a good idea to adapt autotuning to toggle these analyses based on the presence of `setjmp`/`longjmp`.

tests/incremental/11-restart/12-mutex-simple-access.c
```diff
@@ -1,3 +1,4 @@
+// SKIP!
```
sim642, line 1: If these are skipped because they fail on this branch, there's a regression that certainly needs to be fixed.

src/analyses/vla.ml
```diff
@@ -0,0 +1,51 @@
...
+  let enter ctx (lval: lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
+    [ctx.local, false]
+
+  let combine ctx ?(longjmpthrough = false) (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask: Queries.ask) : D.t =
+    ctx.local
+
+  let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    ctx.local
+
+  let vdecl ctx (v:varinfo) : D.t = true
```
sim642, line 36: This should actually check if it's a VLA or just a declaration, instead of implicitly relying on the fact that right now we choose to keep `Cabs2cil.alwaysGenerateVarDecl` as false.

Actually, isn't this analysis just a check whether any of the current function's locals is a VLA? I think that could be checked directly via CIL instead of a new analysis.

tests/incremental/11-restart/12-mutex-simple-access.c
michael-schwarz: Yes, that is true! They just seemed to be unrelated to longjmp/setjmp, but rather seems to highlight some issue of the incremental analysis, potentially together with OS X.

sim642: In #1015 I made GitHub Actions upload the test logs and they show stack overflow exceptions. Since the incremental tests don't contain any `setjmp`/`longjmp`, it is highly suspicious why some pre-existing issue would be revealed now. It's more likely that something in the new (activated by default!) analyses is problematic even when the new analyses don't compute anything meaningful. Maybe a polymorphic comparison somewhere?

sim642: I looked at the CI logs and the exception comes from #952:
```
Raised by primitive operation at Stdlib__List.mem in file "list.ml", line 184, characters 12-23
Called from Goblint_lib__TaintPartialContexts.Spec.return.(fun) in file "src/analyses/taintPartialContexts.ml", line 41, characters 73-99
```

The `List.mem` there is implicitly using polymorphic comparison on `varinfo`, which, if has a struct type, will have reference cycles and the default comparison does not terminate.

I made a fix in https://github.com/goblint/analyzer/pull/1015/commits/37c3ce123e0ca8fe9e193c16a0b3def5b38d23fa in #1015. We'll see whether that fixes the issue.

michael-schwarz:
I'll also start a testrun of this with the benchmarks we used for SOAP.

michael-schwarz:
The changes increase the runtime for libpng by ~ 10%. I could not inspect warnings, as they now are in the behavior category, so I restarted a run where I increased the level of output again.

sim642:
Interesting, I'm not sure what might've caused it but I have a few ideas from my changes in #1015:
1. Replace context hashes with `ControlSpecC` (1f7424f05c58bd924cbe6be62f9f9b99f9266ea2). This makes things more reliable but such elements in sets are also more expensive to compare, etc than just `int`s.
2. I also changed the normal `combine` (7e7055316e9ca395c15fe0263cf8836447259cfc). Previously it changed `local` without adapting `ask`.
3. Using `Access` events (ba3437a74cc0aa22850fad91da426146412d6ef3, d7708d3b179bc052c9b59ab67ed5931c054afa92, f2eb6e0e1ace87980e12bfdc9bd85239a3803e41) might be making things more expensive while in single-threaded mode. Normally those events are not emitted, but with these new analyses they are.

michael-schwarz:
I was able to verify that the output is still as expected. W.r.t. the slowdown, I don't think it is critical for this PR, I think it can be addressed in a separate PR.

sim642:
For some reason marshaling tests on the new test group fail across the board: https://github.com/goblint/analyzer/actions/runs/4516943406.

michael-schwarz:
Maybe it's because contexts are not relifted inside domain values?

michael-schwarz:
After relifting the contents of jumpbuffers, it works if done manually:

> ./regtest.sh 68 02 --set save_run save
> ./regtest.sh 68 02 --set load_run save 

For some reason, it still fails when invoked via the script.

sim642:
> Interesting, I'm not sure what might've caused it but I have a few ideas from my changes in #1015:
> 
>    2. I also changed the normal `combine` 7e7055316e9ca395c15fe0263cf8836447259cfc. Previously it changed `local` without adapting `ask`.

I benchmarked this just this change on sv-benchmarks with master and there wasn't a performance difference.
So at least this change, which affects Goblint even without longjmp analyses, doesn't hurt us in general.

sim642:
> After relifting the contents of jumpbuffers, it works if done manually:
> 
> > ./regtest.sh 68 02 --set save_run save
> > ./regtest.sh 68 02 --set load_run save
> 
> For some reason, it still fails when invoked via the script.

That's not how marshaling tests do it though. The second run uses `--conf save/config.json --set save_run '' --set load_run save`.
Exactly using the full commands used by `update_suite.rb`, I can reproduce the fixpoint error still.

michael-schwarz:
For some reason, there are two paths when the incremental analysis checks if the fixpoint is reached, but only one in the marshalled result.

sim642:
I've started looking into this myself already. We have quite many `relift`s missing, but that's hidden by the default one. Making the default one in `Printable.Std` fail allows tracking down any missing ones that we've been too lazy to add. I've done that for everything in this particular example, but there's still an issue, but it's now inside base's jmpbuf domain.

PR #972, opened 2023-01-24
Handle annihilators in interval `logor` and `logand`

sim642:
If `2 <= x` is true according to intervals, but `r` is unknown, then `r || 2 <= x` should still be true. However, previously it was unknown as annihilators for `logor` and `logand` were not specially handled in the interval domain.

This only concerns witness expressions since `LOr` and `LAnd` do not appear in normal analysis.

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -655,18 +655,20 @@ struct
     else
       narrow ik x y
 
-  let log f ik i1 i2 =
+  let log f annihilator ik i1 i2 =
     match is_bot i1, is_bot i2 with
     | true, true -> bot_of ik
     | true, _
     | _   , true -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show i1) (show i2)))
     | _ ->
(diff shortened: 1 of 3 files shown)


src/cdomains/intDomain.ml
```diff
@@ -655,18 +655,20 @@ struct
...
       match to_bool i1, to_bool i2 with
+      | Some x, _ when x = annihilator -> of_bool ik annihilator
+      | _, Some y when y = annihilator -> of_bool ik annihilator
       | Some x, Some y -> of_bool ik (f x y)
       | _              -> top_of ik
 
-  let logor = log (||)
-  let logand = log (&&)
+  let logor = log (||) true
+  let logand = log (&&) false
```
jerhard, line 671: One may consider making `annihilator` a named argument, but I don't have a strong opinion on that.

jerhard (approved):
Looks good!

src/cdomains/intDomain.ml
sim642: Good idea, changed!

PR #965, opened 2023-01-12
Seq optimization

KhalilCK:
We replaced certain List methods with Seq package.
The sanity check passed without errors using make test and the benchmarking using the following flags:
 

`./goblint ../bench/coreutils/ls_comb.c --set pre.cppflags[+] "--std=gnu89" --disable ana.base.context.non-ptr --disable ana.int.def_exc --disable sem.unknown_function.spawn --set ana.thread.domain plain --enable exp.earlyglobs --set ana.base.privatization none --set pre.cppflags[+] -DGOBLINT_NO_BSEARCH --set pre.cppflags[+] -DGOBLINT_NO_ASSERT --set result json-messages --set ana.activated "['base', 'mallocWrapper']" --set ana.ctx_insens[+] base --set ana.ctx_insens[+] mallocWrapper --enable dbg.timing.enabled --enable ana.int.interval --enable ana.int.interval_threshold_widening`

On my local machine the Seq implementation took 820.1s.
We will choose between the Seq and iter packages, which one is better based on the results of the benchmarking on the cluster that you can find in the Iter PR https://github.com/goblint/analyzer/pull/964.

PR #974, opened 2023-01-27
Iter optimization clean up

montrie:
### This PR is a cleaned up version of PR #964 

As discussed, I created a new PR containing only the relevant Iter related changes. This PR is now based on the merged `flambda` PR #928. 
In particular, it is based on the commit [https://github.com/goblint/analyzer/commit/ab2141a00b577cb20d354878a315a2e27e5074c6](url)
.
All relevant information can be found in the comments of #964.

PR #975, opened 2023-01-28
Seq-optimization clean version

KhalilCK:
This PR is a clean up version of https://github.com/goblint/analyzer/pull/965 and contains only the Seq changes.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -107,7 +107,7 @@ struct
         | (info,value)::xs ->
           match value with
           | `Address t when hasAttribute "goblint_array_domain" info.vattr ->
-            let possibleVars = List.to_seq (PreValueDomain.AD.to_var_may t ) in
+            let possibleVars = List.to_seq (PreValueDomain.AD.to_var_may t) in
             Seq.fold_left (fun map arr -> VarMap.add arr (info.vattr) map) (pointedArrayMap xs) @@ Seq.filter (fun info -> isArrayType info.vtype) possibleVars
           | _ -> pointedArrayMap xs
       in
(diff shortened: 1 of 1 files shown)


src/analyses/base.ml
```diff
@@ -107,8 +107,8 @@ struct
         | (info,value)::xs ->
           match value with
           | `Address t when hasAttribute "goblint_array_domain" info.vattr ->
-            let possibleVars = PreValueDomain.AD.to_var_may t in
-            List.fold_left (fun map arr -> VarMap.add arr (info.vattr) map) (pointedArrayMap xs) @@ List.filter (fun info -> isArrayType info.vtype) possibleVars
+            let possibleVars = List.to_seq (PreValueDomain.AD.to_var_may t )in
```
michael-schwarz, line 110: Whitespace is a bit wonky here...

src/analyses/base.ml
```diff
@@ -1257,7 +1257,7 @@ struct
             | _ -> None
           in
           let alen = List.filter_map (fun v -> lenOf v.vtype) (AD.to_var_may a) in
-          let d = List.fold_left ID.join (ID.bot_of (Cilfacade.ptrdiff_ikind ())) (List.map (ID.of_int (Cilfacade.ptrdiff_ikind ()) %BI.of_int) (slen @ alen)) in
+          let d = Seq.fold_left ID.join (ID.bot_of (Cilfacade.ptrdiff_ikind ())) ((Seq.map (ID.of_int (Cilfacade.ptrdiff_ikind ()) %BI.of_int))@@ List.to_seq (slen @ alen)) in
```
michael-schwarz, line 1260: Can one do better than `List.to_seq` after the append here?

michael-schwarz: Also, I'd prefer a space before the `@@`.

KhalilCK: > Can one do better than `List.to_seq` after the append here?

Thanks for the feedback, I will try to refactor the `List.to_seq` here.

sim642:
Compared to #974, it has this change which isn't present here: https://github.com/goblint/analyzer/pull/974/files#diff-969b2ca0b7175f19daa5deae1308a02718e8cd1927a4d8e160541a35e57c31dbL1376-R1380. Is there a reason for that?

Same with: https://github.com/goblint/analyzer/pull/974/files#diff-7c58a9f6dd84f297a28b19f30a5cd93b5a9c13c4bfe3a225c43e8a9c85b9aae5L100-R104.

michael-schwarz (approved):

jerhard (approved):

sim642 (approved):
The changes themselves are fine, I was just wondering about my previous point: https://github.com/goblint/analyzer/pull/975#issuecomment-1410204711.

michael-schwarz:
We discussed this with our students offline yesterday, one of places you mentioned these was added, for the one using sort we didn't see the point as the introduction won't save anything there, as the entire list needs to be considered to sort it anyway.

PR #977, opened 2023-01-31
Thread-modular witnesses benchmarking fixes

sim642:
### TODO
- [ ] Rewrite history to remove all BenchExec stuff, which should go to bench repository.


src/witness/yamlWitness.ml
```diff
@@ -447,7 +447,8 @@ struct
     let loop_locator = Locator.create () in
     LHT.iter (fun ((n, _) as lvar) _ ->
         let loc = Node.location n in
-        if not loc.synthetic then
+        (* TODO: filter synthetic? *)
+        if WitnessInvariant.is_invariant_node n then
```
michael-schwarz, line 451: Why did you remove the filtering for `loc.synthetic` here?

src/analyses/baseInvariant.ml
```diff
@@ -29,6 +29,8 @@ sig
 
   val id_meet_down: old:ID.t -> c:ID.t -> ID.t
   val fd_meet_down: old:FD.t -> c:FD.t -> FD.t
+
+  val contra: D.t -> D.t
```
michael-schwarz, line 33: I think we should explain what contra is, and why we pulled it out here.

michael-schwarz (approved):

PR #978, opened 2023-02-02
Missing races from `free` of `calloc`, special case `calloc` with count 1

sim642:
While debugging thread-modular witness validation, I implemented a special case for `calloc` with count 1, such that it avoids an intermediate array, which we cannot output invariants for. (Also committed here to demonstrate the issue.)

Surprisingly, this broke some zstd thread pool tests since additional races appeared. This PR contains further minimizations of the zstd thread pool example. Somehow the unsoundness appears to be that if there's an extra array domain in between, then accesses from `free` disappear altogether.


sim642:
Everything split between #1089 and #1124.

PR #979, opened 2023-02-06
Dead Code Elimination Transformation

### LINKED ISSUE #850: Dead code elimination transformation
Since we know where dead code is and can output C programs via CIL, we could have a transformation that removes all the dead code from the program.

It wouldn't be super practical on real-world code as CIL also changes all the formatting, but it might be useful for debugging unsoundness on huge programs. For example, the program from #750 has both branches dead in a massive function where Graphviz gives up on the CFG:
image
If instead we could transform the program, it would be much easier to see, which parts of the program we consider live at all and thus where we go wrong.

Such elimination could apply to many things:
- [x] Remove dead statements
- [x] Remove dead functions altogether
- [x] Remove dead (i.e. unreferenced) globals
- [x] Remove dead (i.e. unused) types, like typedefs, struct declarations, etc.

The latter should be determined after the first two, such that dead functions wouldn't keep otherwise unused globals/types around. Cilled LDV programs in sv-benchmarks are notorious for containing large amounts of environment modeling code, which the particular benchmark doesn't actually use.

just-max:
Dead code elimination transformation as suggested by Simmo in #850. ~~The implementation in its current state is still quite messy, but all the important bits should be there.~~

The transformation:
- removes dead code (statements) within functions
- removes functions that are entirely dead
- using CIL Rmtmps/RmUnused, tracks references between globals (functions, variables, named types, struct/union types, enums), then removes any that aren't referenced (possibly transitively) by live code

This PR also makes small changes to the transformation subsystem:
- expand query/ask for access to liveness information
- streamline output to file

Merging will close #850.

TODO:
- [x] Clean up code, move all to own module
- [x] Put transformation behind appropriate option and clean up output to file


src/framework/control.ml
```diff
@@ -595,26 +595,46 @@ struct
...
-            let v' = try Spec.D.join (Hashtbl.find h k) v with Not_found -> v in
-            Hashtbl.replace h k v') e;
+            let v' = Hashtbl.find_option h loc |> Stdlib.Option.fold ~none:v ~some:(Spec.D.join v) in
+            Hashtbl.replace h loc v')
+            e;
           h
         in
+
+        (* Map locations of functions to their IDs, so queries can ask about functions. *)
+        let fid_by_loc = Hashtbl.create 37 in
```
just-max, line 612: Only iterate over `lh` once.

src/framework/control.ml
```diff
@@ -595,26 +595,46 @@ struct
...
+        LHT.enum lh
+          |> Enum.map (fst %> fst) (* grab node from key in key-value pair *)
+          |> Enum.iter (function
+              | (Function fd | FunctionEntry fd) as n -> Hashtbl.replace fid_by_loc (Node.location n) fd.svar.vid
+              | _ -> ()) ;
+
         let ask ~node loc = (fun (type a) (q: a Queries.t) ->
```
just-max, line 619: Rewrite `ask` to use a datatype with a constructor for query, must be dead and must be uncalled.

src/transform/deadCode.ml
```diff
@@ -0,0 +1,70 @@
+open Batteries
+open GoblintCil
+
+
+(** Filter statements out of a block (recursively). CFG fields (prev, next, etc.) are no longer valid after calling.
+    Returns true if anything is left in block, false if the block is now empty.
+    Invariants:
+    - f (goto label) ==> f (labelled stmt), i.e. if a goto statement is not filtered out, the target may not be filtered out either.
+    - block may not contain switch statements *)
+let filter_map_block f (block : Cil.block) : bool =
```
just-max, line 10: Could gotos be spuriously detected as live.

src/util/options.schema.json
```diff
@@ -1357,7 +1357,10 @@
           "description":
             "Lists of activated transformations. Transformations happen after analyses.",
           "type": "array",
-          "items": { "type": "string" },
+          "items": {
+            "type": "string",
+            "enum": ["partial", "expeval", "assert", "remove_dead_code"]
```
just-max, line 1362: Constrain ordering: remove dead code before adding assertions.

just-max: Upon testing, this almost certainly needs to be the other way around, i.e. `assert` before `remove_dead_code`. Otherwise, assertions added by `assert` may reference variables that have already been removed.

And the initial problem as described above does not exist: assertions are not inserted around dead code, by virtue of that code being dead.

just-max:
As discussed with Michael, a final change would be to rename the `Rmtmps` CIL module, since it removes much more than temporaries. But there should be no more material changes to this pull request.

src/transform/deadCode.ml
michael-schwarz: Did you come to a conclusion on this?

michael-schwarz (approved):
As soon as you have convinced yourself that the issue with `goto`s mentioned above is a non-issue, I think this is ready to merge!

src/transform/deadCode.ml
just-max: The case that would cause a problem is if a `goto label;` remains in the block, but `label: ...` is removed. That means the goto must be either (1) live in some context or (2) not appear in the result at all. The labelled statement on the other hand must be dead in all contexts.

In case (1), the goto is live in some context and control flow always passes to the labelled statement, so the labelled statement would also need to be live.

In case (2), the goto would not have a CFG node in the result. This case only occurs for statements without a node in the result, e.g. compound statements. As far as I understand, this case can not occur.

michael-schwarz:
@sim642 any comments before we merge?

src/transform/deadCode.ml
```diff
@@ -0,0 +1,68 @@
...
+      | _ -> ());
+
+    (* step 2: remove globals that are (transitively) unreferenced by live functions
+       - dead functions: removed by answering 'false' for isRoot
+       - unreferenced globals and types: removeUnusedTemps keeps only globals referenced by live functions *)
+    let open GoblintCil.Rmtmps in
+    let keepUnused0 = !keepUnused in
+    Fun.protect ~finally:(fun () -> keepUnused := keepUnused0) (fun () ->
+      keepUnused := false;
+      removeUnusedTemps ~isRoot:(function GFun (fd, _) -> fundec_live fd | _ -> false) file
```
sim642, line 59: Wouldn't it be enough for just the `main` function to be root?

src/transform/deadCode.ml
```diff
@@ -0,0 +1,68 @@
...
+        (* be careful to not short-circuit, since call to impl_block b2 is always needed for side-effects *)
+        let keep_b1, keep_b2 = impl_block b1, impl_block b2 in keep_b1 || keep_b2
+      | Switch _ -> failwith "switch statements must be removed"
+        (* handling switch statements correctly would be very difficult; consider that the switch
+        labels may be located within arbitrarily nested statements within the switch statement's block *)
+      | Loop (b, _, _, _, _) ->
+        impl_block b
+      | Block b ->
+        impl_block b
+      | sk -> true
```
sim642, line 29: This match should enumerate all constructors explicitly and be explicit about each construction and what is assumed to be transformed away. Like `CfgTools` does it.

sim642 (commented):
How does this handle the fact that there isn't a one-to-one correspondance between our CFG nodes (which we compute liveness for) and CIL statements?
That is, our CFG construction skips over many CIL statements, so we don't compute liveness for many of them. Evene moreso, `CfgTools` never puts `Block` statements into nodes (which are handled here), but instead `Instr` statements (which are left into the default `true` case here).

It would be good to have (e.g. cram) tests for this that really show all aspects working, because I'm not entirely convinced right now.
CIL transformations working at the level of statements and locations instead of our CFG nodes has been an issue in the past (other transformations have a bunch of hackery regarding this), so it's odd that this transformation can get away with not needing any such special treatment.

src/transform/deadCode.ml
sim642: Actually, isn't this even biased the incorrect way? If `main` has takes a function pointer to an uncalled function `foo`, then even considering just `main` as root would still mean that `foo` is found to be reachable from the root even when `isRoot` for it returns false, no? Since `Rmtmps` is entirely syntactical, it feels like it wouldn't really be able to remove much at all because it is a conservative syntactic thing, which, to remain on the safe side, cannot know to remove things which our analysis knows more precisely about.
Or am I mistaken about what `Rmtmps` does?

src/transform/deadCode.ml
```diff
@@ -0,0 +1,68 @@
...
+    (* whether a global function (might) still be live, and should therefore be kept *)
+    let fundec_live : fundec -> bool = not % q.must_be_uncalled in
+
+    (* step 1: remove statements found to be dead *)
+    Cil.iterGlobals file
+      (function
+      | GFun (fd, _) ->
+        (* invariants of filter_map_block satisfied: switch statements removed by CFG transform,
+           and a live label implies its target is live *)
+        filter_map_block stmt_live fd.sbody |> ignore
```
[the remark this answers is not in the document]
just-max: A function with an empty body may still be referenced elsewhere, even if it's only, for example, by taking its address. If the function is truly unreferenced, it will be removed in the next phase (by `removedUnused`).

src/transform/transform.ml
```diff
@@ -1,16 +1,42 @@
...
-  T.transform
+let register (module T : S) = Hashtbl.replace h T.name (module T : S)
+
+let run_transformations ?(file_output = true) file names ask =
+  let active_transformations =
+    List.filter_map
+    (fun name ->
+      match Hashtbl.find_option h name with
+      | Some t -> Some (name, t)
+      | None -> M.warn_noloc "Transformation %s does not exist!" name; None)
```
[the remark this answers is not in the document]
just-max: Updated to a hard failure. In a normal Goblint run, a hard failure also occurs on option parsing because of the enumeration of available transformations:

https://github.com/goblint/analyzer/blob/8d862841004ebe7800bea66de0da56e45e074b88/src/util/options.schema.json#L1360-L1363

src/transform/deadCode.ml
just-max: No, your understanding of `Rmtmps` is correct; it is a purely syntactical transformation. The interesting part, removing code that Goblint finds to be semantically dead, happens in the first phase by filtering out statements within a function.

For the purposes of this transformation, I think an uncalled function whose address is nevertheless taken still counts as a usage of that function, and thus it should be retained. I'm also not sure what the reference would be replaced with, as once the address has been stored somewhere, it may be used in non-trivial ways.

As for just having `main` as the root: as long as there is a (chain of) syntactic references from `main` to any live function, using just `main` is enough. Since the analyzed program is considered complete, we don't need to worry about keeping any visible library functions. The CIL docs on `Rmtmps` mention functions with the "constructor" or "destructor" attribute, not sure if those are relevant to Goblint. And finally the CIL global initializer should be called by `main` automatically, as far as I understand. So in short: this is the way we decide which functions are dead for warnings, so I'm more confident keeping it this way. But if there's no edge case I've missed, I'll change this to compare to `mainfun`.

just-max:
> How does this handle the fact that there isn't a one-to-one correspondance between our CFG nodes (which we compute liveness for) and CIL statements?

CIL statements are put in three categories by this transformation: 

- in the (solver) result, dead in all contexts
- in the result, but live in some contexts
- not in the result

Only the first category is removed. Block statements (literaly CIL `Block`s along with e.g. `If` statements) fall under the third category, and are not removed.

(By "result" I mean the `lh` hashtable of solver results by CFG node and context.)

> which are left into the default true case here

The case distinction only handles the recursion. A dead `Instr` is removed above:

https://github.com/goblint/analyzer/blob/8d862841004ebe7800bea66de0da56e45e074b88/src/transform/deadCode.ml#L15-L18

---

> CIL transformations working at the level of statements and locations instead of our CFG nodes has been an issue in the past

Yes, it's not the prettiest. But ultimately the transformation must occur on a syntax level on the original CIL file, short of regenerating the C code from scratch out of the CFG. From my understanding, other transformations need a bit of a hack because they go through the query system. This transformation sort of gets around that with the new `Transform.queries` `must_be_dead` parameter, which directly looks up a statement/location in the map of results joined by statement/location. As long as each CFG node corresponds to one CIL statement, this should work.

src/transform/deadCode.ml
sim642: > For the purposes of this transformation, I think an uncalled function whose address is nevertheless taken still counts as a usage of that function, and thus it should be retained. I'm also not sure what the reference would be replaced with, as once the address has been stored somewhere, it may be used in non-trivial ways.

That's true. Uncalled functions which have their address taken transforming into an empty function seems sensible.

> Since the analyzed program is considered complete, we don't need to worry about keeping any visible library functions. The CIL docs on `Rmtmps` mention functions with the "constructor" or "destructor" attribute, not sure if those are relevant to Goblint.

Goblint has its own `mainfun`, etc stuff similar to that, but the whole Linux kernel auto-harness isn't in a good state, so I think it's fine as is for now.

just-max:
> To do things properly, it looks like it is necessary to traverse through CIL statements in the same way as CFG construction. Then, liveness of If, Loop and Instr/Return can be checked as before, but liveness of anything else would be decided by liveness of the preceding CFG node.

This is essentially what I've implemented now. Statements that are skipped during CFG construction are tracked directly. Then, a statement between two nodes is considered dead if the preceding or following node is dead (with appropriate handling for statements on multiple such paths).

src/framework/cfgTools.ml
```diff
@@ -579,16 +610,17 @@ let fprint_hash_dot cfg  =
   close_out out
 
 
-let getCFG (file: file) : cfg * cfg =
-  let cfgF, cfgB = createCFG file in
+let getCFG (file: file) : cfg * cfg * stmt list CfgEdgeH.t =
+  let cfgF, cfgB, skippedByEdge = createCFG file in
   let cfgF, cfgB =
+    (* TODO: might be broken *)
```
just-max, line 616: This probably breaks the detection of skipped edges.

src/transform/deadCode.ml
```diff
@@ -0,0 +1,217 @@
+open Batteries
+open GoblintCil
+open GobConfig
+open MyCFG
+
+let dpf fmt = Pretty.gprintf (fun doc -> print_endline @@ Pretty.sprint ~width:Int.max_num doc) fmt
```
sim642, line 6: All of the custom debug printing should be removed or replaced by tracing.

src/transform/deadCode.ml
```diff
@@ -0,0 +1,217 @@
...
+
+let empty_block () = { battrs = [] ; bstmts = [] }
```
sim642, line 8: This doesn't need to be a function:
```suggestion
let empty_block = { battrs = [] ; bstmts = [] }
```

src/transform/deadCode.ml
```diff
@@ -0,0 +1,217 @@
...
+  Seq.(ints 0 |> map with_suffix |> filter (not % Hashtbl.mem used_names) |> first)
+
+module RemoveDeadCode : Transform.S = struct
+  let transform (q : Transform.queries) (file : file) : unit =
+
+    (* whether a global function (might) still be live, and should therefore be kept *)
+    let fundec_live : fundec -> bool = not % q.must_be_uncalled in
+
+    (* add an uninitialized global variable to be used as the condition
+       in if-statements where the condition is never checked *)
```
sim642, line 153: Why is this needed? If it's unreachable, then it can be anything, e.g. `1`.
Synthesizing a new global variable (and avoiding name conflicts) is a lot of extra work but that global is still default initialized to 0, so it's not any better than a direct constant condition.

tests/regression/70-transform/transform.sh
```diff
@@ -0,0 +1,33 @@
...
+  done
+
+  # save stdout to fd 3
+  exec 3>&1
+  [ $stdout -eq 1 ] || exec 1>/dev/null
+  [ $stderr -eq 1 ] || exec 2>/dev/null
+
+  # run goblint, then output the transformed file with the
+  # 'Generated by CIL v. X.X.X' header and #line annotations removed
+  "$(which goblint || which ./goblint)" \
```
sim642, line 27: This shouldn't be necessary at all: `goblint` is available directly in cram tests through a dependency in tests/regression/dune.
```suggestion
  goblint \
```

tests/regression/70-transform/deadcode.json
sim642, line 1: Where is this conf used?

tests/regression/70-transform/02-deadcode.c
```diff
@@ -0,0 +1,139 @@
...
+
+// only called with negative n
+int basic2(int n) {
+  int a = 0;
+
+  if (n < 0)
+    return 0;
+
+  for (int i = 0; i < n; i++)
+    a += i + n;  // bug in dead code detection: is not found dead by goblint, because it is not included in the result
```
sim642, line 33: Hmm? Is this comment outdated? Because it is removed by the transformation.

tests/regression/70-transform/03-deadcode-globals.c
```diff
@@ -0,0 +1,50 @@
...
+int struct_pointer_f(struct1_named4 *x) {
+  return x->field1 + 1;
+}
+
+// globals referencing each other
+
+const int global1 = 1;  // referenced (indirectly)
+const int global2 = global1 + 7;  // referenced (directly in main)
+const int global3 = 2;  // referenced by global3 only
+const int global4 = global3 + 9;
```
sim642, line 38:
```suggestion
const int global3 = 2;  // referenced by global4 only
const int global4 = global3 + 9;
```

src/transform/deadCode.ml
just-max: Missed that, thanks.

src/transform/deadCode.ml
just-max: The idea was to document in the resulting code that the condition is dead; there's no functional benefit. But given the amount of work to synthesize the global, it might not be worth it. Really, a comment above the conditional would be enough, but there's no way to do that with CIL.

src/transform/deadCode.ml
just-max: `battrs` and `bstmts` are mutable, hence the function. Currently, these empty blocks are never modified further, so it wouldn't matter, but sharing them seems like a potential future headache.

src/framework/cfgTools.ml
```diff
@@ -135,14 +135,35 @@ let () = Printexc.register_printer (function
...
+module CfgEdge = struct
+  open Batteries
+
+  type t = node * MyCFG.edges * node
+
+  let equal = Tuple3.eq Node.equal (List.equal (Tuple2.eq CilType.Location.equal Edge.equal)) Node.equal
+  let hash =
+    (* map the nodes to their hash, and each (location, edge) pair to their respective hashes;
+       then we have a structure of primitive constructors and ints, and can use polymorphic hash *)
+    Tuple3.map Node.hash (List.map (Tuple2.map CilType.Location.hash Edge.hash)) Node.hash %> Hashtbl.hash
```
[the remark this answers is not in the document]
just-max: Yes, full type needs to be given so `deriving` can find things but otherwise done:

```suggestion
  type t = Node.t * MyCFG.edges * Node.t [@@deriving eq, hash]
```

Also derived eq/hash for `edges` (also requires writing out type names).

tests/regression/70-transform/02-deadcode.c
just-max: The transformation works, but Goblint's existing dead code detection does not:

`./goblint --enable ana.int.interval --enable sem.noreturn.dead_code tests/regression/70-transform/02-deadcode.c`

```text
...
[Warning][Deadcode] Function 'basic2' has dead code:
  on line 32 (tests/regression/70-transform/02-deadcode.c:32-32)
  on line 35 (tests/regression/70-transform/02-deadcode.c:35-35)
...
```

The body of the loop is not reported as dead. This is since the body doesn't appear in the result, and currently Goblint only reports dead code for nodes that are in the result, but dead in all contexts.

tests/regression/70-transform/transform.sh
```diff
@@ -0,0 +1,33 @@
...
+    "${trans_args[@]}" \
+    --set trans.output >(awk '!/^#line / && NR > 3' 1>&3) \
```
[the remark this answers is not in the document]
just-max: I've now disabled `#line` output for all transformations. The only other transformation with file output is assertion-adding, does that need line numbers?

just-max:
Integrated review (https://github.com/goblint/analyzer/compare/ff9aa70...just-max:goblint-analyzer:deadcode_removal).

The regression cram tests are currently failing because apparently cram only sees an empty stdout. Probably some part of file redirection trickery in `transform.sh` not working right.

src/transform/deadCode.ml
sim642: Ah, that's right! A function is indeed safer then. Might be worth adding a comment about it such that nobody in the future is tempted to do the same optimization.

tests/regression/70-transform/transform.sh
sim642: Goblint can be just installed into the current opam switch as well with `make install`.

tests/regression/70-transform/02-deadcode.c
sim642: I've opened #1036 to fix this.

tests/regression/70-transform/transform.sh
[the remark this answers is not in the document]
sim642: I'm not sure, maybe not. But it can remain this way until someone has a specific need to change it.

src/transform/deadCode.ml
sim642: I would just use the constant `1` condition and remove the extra logic for creating these globals, but maybe @michael-schwarz has a different opinion?

michael-schwarz: I think it makes sense to have this special global to draw attention to this fact, that it is in fact never checked. That might make reduced code easier to understand as it is clear that the things inside can only be reachable by jumps etc.

michael-schwarz:
> Edit: Making this work on Mac without having such a system locally is quite frustrating.

I feel you, I have the same issue all the time myself too!

michael-schwarz (approved):

src/transform/deadCode.ml
sim642: A simpler alternative would be to use a string literal like `"dead condition"`. It contains the same kind of description and is constant like the global variable.

just-max:
As Simmo suggested, I've used a string constant in dead conditions. Everything should be ready now.

PR #986, opened 2023-02-12
Short-circuiting operations inside `HConsed`

### LINKED ISSUE #611: Add short-circuit optimization for `leq` to `HConsed`
Values that are physically equal or, for `HConsed` have the same tag, are also `leq`. We should exploit this by implementing a short-circuiting version of `leq` for `HConsed`.

michael-schwarz:
I have now implemented a first prototype of this and started an evaluation on sv-comp.

Closes #611

### DIFF
--- a/src/domains/lattice.ml
+++ b/src/domains/lattice.ml
@@ -152,11 +152,11 @@ module HConsed (Base:S) =
 struct
   include Printable.HConsed (Base)
   let lift_f2 f x y = f (unlift x) (unlift y)
-  let narrow x y = lift (lift_f2 Base.narrow x y)
-  let widen x y = lift (lift_f2 Base.widen x y)
-  let meet x y = lift (lift_f2 Base.meet x y)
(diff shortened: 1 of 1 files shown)


michael-schwarz:
The results are not really encouraging here, it quite clearly does not matter that much. Some 60 tasks go from timeout to unknown, and we succeed on 4 additional tasks but that's about it. [results_shortcircuiting.zip](https://github.com/goblint/analyzer/files/10730077/results_shortcircuiting.zip)

Maybe one would want to try this again for one of the longer running benchmarks, though I am not sure why one would expect a speedup there and none here.

sim642:
> Some 60 tasks go from timeout to unknown, and we succeed on 4 additional tasks but that's about it

Unless there's some downside, that's still something and worth merging for.

michael-schwarz:
> Unless there's some downside

No, there's no downsides. There's the usual jitter one expects, but nothing becomes worse in a relevant way.

sim642 (approved):

PR #987, opened 2023-02-12
Autotuner: Do not activate Apron for `val` variables from Goblint header for `__VERIFIER_nondet_*`

### LINKED ISSUE #921: Autotuner: Do not activate Apron for `val` variables from Goblint header for `__VERIFIER_nondet_*`
Currently, the autotuner activates apron for the variables `val` that are used inside our   `__VERIFIER_nondet_*` stubs. As no information will ever become available about these variables, we should add a check such that the autotuner does not enable Apron for these variables.

michael-schwarz:
Currently, the autotuner enables tracking relational information about variables used to implement `__VERIFIER_nondet_*` functions. 

This PR introduces attribute `goblint_stub` for variables to prevent tracking them relationally and marks the variables with this attribute.

Closes #921

### DIFF
--- a/src/autoTune.ml
+++ b/src/autoTune.ml
@@ -245,7 +245,7 @@ let isComparison = function
 
 let rec extractVar = function
   | UnOp (Neg, e, _) -> extractVar e
-  | Lval ((Var info),_) -> Some info
+  | Lval ((Var info),_) when not (List.exists (fun (Attr(s,_)) -> s = "goblint_stub") info.vattr) ->  Some info
   | _ -> None
 
 let extractOctagonVars = function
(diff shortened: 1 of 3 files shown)


src/autoTune.ml
```diff
@@ -245,7 +245,7 @@ let isComparison = function
 
 let rec extractVar = function
   | UnOp (Neg, e, _) -> extractVar e
-  | Lval ((Var info),_) -> Some info
+  | Lval ((Var info),_) when not (List.exists (fun (Attr(s,_)) -> s = "goblint_stub") info.vattr) ->  Some info
```
jerhard, line 248: You may consider whether you want to extract the condition here and below into a function `is_goblint_stub`.

jerhard (commented):
Looks good! Nice to see that you found a simply fix for this.

Just for reference: There already is a `goblint_stub` attribute for functions: 
It is used to exclude those functions from functions with dead lines of code, and to exclude them from being analyzed as a thread when the `"allfuns"` option is active. I do not think that these different effects of the `goblint_stub` attribute are a problem however, since they are really consequence of the function/variable being a stub / part of a stub.

jerhard (approved):

PR #989, opened 2023-02-14
Handling escaped variables in region analysis

vesalvojdani:
This is meant to address two issues we became aware of about escaped variables and the region analysis. 

- [x] Simply handling pthread arguments and treating escaped variables as globals, so we handle regtest 09-34 without resorting to the empty region map override. 
- [ ] Dealing with escaped owners when a region member escaped. Would be nice if one could trigger an escape event for this, although it would be extremely imprecise because of the non-directional heap abstraction. The alternative would be to re-implement a region-specific escape analysis. 
- [ ] (Optional) Understand what region owners actually mean: probably one should use `AddrOf(i)` as owner in 9-34 when its not a pointer.


jerhard:
Please add a test case for the cases that are fixed with this PR. Then we can consider merging this PR. 
The following issues may be related to the behavior addressed here: #107  #1070 #1298  #1300.
Please check whether any of these is affected.

michael-schwarz:
@copilot Resolve conflicts and add a test case for a regression that this fixes.

michael-schwarz:
@copilot The tests fail

michael-schwarz:
@copilot: Instead of updating individual files, merge master into this.

michael-schwarz:
The failing CI is unrelated.

sim642:
I don't understand what's up with this. The merges have reverted all the fix attempts this contained.

The test added by Copilot is completely irrelevant: it already passes without needing any of the fixes that used to be here, so it's not a regression test for what this was supposed to fix.
Also, that test has nothing to do with region really: it's race-free purely due to very basic mutex analysis, the addition of region cannot even screw it up. It was copied from 09/34 where that was actually checking for soundness, i.e., region doesn't prove something race-free which races.

As-is I would just close this because there's nothing worth merging. But I'm confused about why the merges are reverting the original attempts.

michael-schwarz:
> But I'm confused about why the merges are reverting the original attempts.

They were copilot things. So probably best to hard reset to [d26f8fb](https://github.com/goblint/analyzer/pull/989/commits/d26f8fbdd8dc9ee98c8eddb10b0d4764ec0e7a89).

sim642:
It was actually @vesalvojdani's merge in 78d465a665620522923954e0ebd790c885065587 that reverted everything, maybe accidentally due to the merge conflict?

PR #992, opened 2023-02-16
Use `Z` instead of `Big_int_Z` and `Cilint`

sim642:
`Big_int_Z` is just an old compatibility wrapper for `Z` except all the function names redundantly contain `big_int` as well.
`Cilint` is now just a wrapper for `Z` too except all the function names redundantly contain `cilint` as well.

This just makes much of the bigint handling code much more readable by using `Z` directly.

### DIFF
--- a/src/util/intOps.ml
+++ b/src/util/intOps.ml
@@ -219,16 +219,8 @@ struct
   let add = Z.add
   let sub = Z.sub
   let mul = Z.mul
-
-  (* If the first operand of a div is negative, Zarith rounds the result away from zero.
-     We thus always transform this into a division with a non-negative first operand.
-  *)
-  let div a b = if Z.lt a zero then Z.neg (Z.ediv (Z.neg a) b) else Z.ediv a b
-
-  (* Z.erem computes the Euclidian Modulus, but what we want here is the remainder, as returned by mod on ints
-     -1 rem 5 == -1, whereas -1 Euclid-Mod 5 == 4
-  *)
-  let rem a b = Z.sub a (mul b (div a b))
+  let div = Z.div
+  let rem = Z.rem
 
   let gcd x y = abs @@ Z.gcd x y
   let compare = Z.compare
--- a/src/util/intOps.ml
+++ b/src/util/intOps.ml
@@ -53,8 +53,8 @@ sig
   val to_int64 : t -> int64
   val of_string : string -> t (* TODO: unused *)
   val to_string : t -> string
-  val of_bigint : Big_int_Z.big_int -> t
-  val to_bigint : t -> Big_int_Z.big_int
+  val of_bigint : Z.t -> t
+  val to_bigint : t -> Z.t
 end
 
 module type IntOps =
@@ -206,9 +206,9 @@ struct
   let to_bigint = Z.of_int64
 end
 
-module BigIntOpsBase : IntOpsBase with type t = Big_int_Z.big_int =
+module BigIntOpsBase : IntOpsBase with type t = Z.t =
 struct
-  type t = Big_int_Z.big_int
+  type t = Z.t
   let zero = Z.zero
   let one = Z.one
   let upper_bound = None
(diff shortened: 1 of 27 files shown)


src/util/intOps.ml
```diff
@@ -202,59 +202,59 @@ struct
...
   *)
-  let div a b = if Big_int_Z.lt_big_int a zero then Big_int_Z.minus_big_int (Big_int_Z.div_big_int (Big_int_Z.minus_big_int a) b) else Big_int_Z.div_big_int a b
+  let div a b = if Z.lt a zero then Z.neg (Z.ediv (Z.neg a) b) else Z.ediv a b
 
-  (* Big_int_Z.mod_big_int computes the Euclidian Modulus, but what we want here is the remainder, as returned by mod on ints
+  (* Z.erem computes the Euclidian Modulus, but what we want here is the remainder, as returned by mod on ints
      -1 rem 5 == -1, whereas -1 Euclid-Mod 5 == 4
   *)
-  let rem a b = Big_int_Z.sub_big_int a (mul b (div a b))
+  let rem a b = Z.sub a (mul b (div a b))
```
michael-schwarz, line 231: Is this now just `Z.rem` ?

> Integer remainder. Can raise a Division_by_zero. The result of rem a b has the sign of a, and its absolute value is strictly smaller than the absolute value of b. The result satisfies the equality a = b * div a b + rem a b.

From https://ocaml.org/p/zarith/1.12/doc/Z/index.html#val-rem

src/util/loopUnrolling.ml
```diff
@@ -227,16 +227,16 @@ let rec loopIterations start diff comp =
...
   match comp with
   | BinOp (op, (Const _ as c), var, t) -> loopIterations start diff (BinOp (flip op, var, c, t))
-  | BinOp (Lt, _, (Const (CInt (cint,_,_) )), _) -> if Z.lt diff Z.zero then None else loopIterations' (Cilint.big_int_of_cilint cint) false
-  | BinOp (Gt, _, (Const (CInt (cint,_,_) )), _) -> if Z.gt diff Z.zero then None else loopIterations' (Cilint.big_int_of_cilint cint) false
-  | BinOp (Le, _, (Const (CInt (cint,_,_) )), _) -> if Z.lt diff Z.zero then None else loopIterations' (Z.add Z.one @@ Cilint.big_int_of_cilint cint) false
-  | BinOp (Ge, _, (Const (CInt (cint,_,_) )), _) -> if Z.gt diff Z.zero then None else loopIterations' (Z.pred @@ Cilint.big_int_of_cilint cint ) false
-  | BinOp (Ne, _, (Const (CInt (cint,_,_) )), _) -> loopIterations' (Cilint.big_int_of_cilint cint) true
+  | BinOp (Lt, _, (Const (CInt (cint,_,_) )), _) -> if Z.lt diff Z.zero then None else loopIterations' cint false
+  | BinOp (Gt, _, (Const (CInt (cint,_,_) )), _) -> if Z.gt diff Z.zero then None else loopIterations' cint false
+  | BinOp (Le, _, (Const (CInt (cint,_,_) )), _) -> if Z.lt diff Z.zero then None else loopIterations' (Z.succ @@ cint) false
```
michael-schwarz, line 237:
```suggestion
  | BinOp (Le, _, (Const (CInt (cint,_,_) )), _) -> if Z.lt diff Z.zero then None else loopIterations' (Z.succ cint) false
```

src/util/loopUnrolling.ml
```diff
@@ -227,16 +227,16 @@ let rec loopIterations start diff comp =
...
+  | BinOp (Ge, _, (Const (CInt (cint,_,_) )), _) -> if Z.gt diff Z.zero then None else loopIterations' (Z.pred @@ cint) false
```
michael-schwarz, line 238:
```suggestion
  | BinOp (Ge, _, (Const (CInt (cint,_,_) )), _) -> if Z.gt diff Z.zero then None else loopIterations' (Z.pred cint) false
```

michael-schwarz (approved):

src/util/intOps.ml
sim642: We have our own `div` definition above, but `Z.div` (as opposed to `Z.ediv`) seems to work the same way, so I'll just remove both.

sim642: I'll add a qcheck test to make sure they're the same as ours before.

PR #993, opened 2023-02-17
Fix loading of C files in Gobview

stilscher:
This PR contains changes that were made on the Goblint side to fix issue https://github.com/goblint/gobview/issues/2 and the timing statistics displayed in Gobview:

- When Goblint is run with gobview enabled, all C source files (from warning locations) will be copied to the directory served by Gobview. The files receive a unique name to avoid any clashes between files from different directories. The mapping from the filename used in CIL-locations to this unique file name is marshalled additionally such that it can be accessed by Gobview. This way, the files can be displayed in Gobview without accessing any other parts of the filesystem. I tested this fix for C files located within the Goblint directory as well as for files outside. However, I am assuming, that one might need to extend this solution for it to work with the compilation database setup as well.

- Furthermore, this PR fixes the timing statistics displayed in Gobview. Instead of marshaling incomplete timings within analyze, where the overall analysis time has not been calculated yet, they are now only marshaled after finishing the analysis.

### TODO
- [x] Update gobview submodule after https://github.com/goblint/gobview/pull/19 is merged.


michael-schwarz (approved):

src/maingoblint.ml
```diff
@@ -508,6 +508,24 @@ let do_gobview () =
...
+      let copy_rem p =
+        let _, name = BatFile.open_temporary_out ~prefix:(Fpath.filename p) ~suffix:".c" ~temp_dir:(Fpath.to_string file_dir) () in
+        let gobview_path = match Fpath.relativize ~root:(Fpath.v run_dir) (Fpath.v name) with
+          | Some p -> Fpath.to_string p
+          | None -> failwith "The gobview directory should be a prefix of the paths of c files copied to the gobview directory" in
+        Hashtbl.add file_loc (Fpath.to_string p) gobview_path;
+        BatFile.write_lines name (BatFile.lines_of (Fpath.to_string p)) in
+      let paths = Cil.foldGlobals cilfile (
+          fun acc g -> match g with GFun (_,loc) -> if not (List.mem loc.file acc) then loc.file :: acc else acc | _ -> acc) [] in
+      List.iter copy_rem (List.map Fpath.v paths);
```
michael-schwarz, line 524: Maybe we can use `Fileutils.cp` here, instead of doing it manually linewise

docs/user-guide/inspecting.md
```diff
@@ -11,9 +11,12 @@ Modern browsers' security settings forbid some file access which is necessary fo
 
 ## Gobview
 
+For the initial setup:
```
sim642, line 14: I think this Markdown flavor requires an empty line before a list. At least the PR build doesn't show it correctly right now: https://goblint--993.org.readthedocs.build/en/993/user-guide/inspecting/#gobview.

```suggestion
For the initial setup:

```

docs/user-guide/inspecting.md
```diff
@@ -11,9 +11,12 @@ Modern browsers' security settings forbid some file access which is necessary fo
...
 1. Install Node.js (preferably ≥ 12.0.0) and npm (≥ 5.2.0)
-2. For the initial setup: `make setup_gobview`
-3. Run `dune build gobview` to build the web UI
-4. Run Goblint with these flags: `--enable gobview --set save_run DIR` (where `DIR` is the name of the result directory that Goblint will create and populate)
-5. `cd` into `DIR` and run `python3 -m http.server`
-6. Visit http://localhost:8000
+2. Run `make setup_gobview` in the analyzer directory
+
+To build Gobview (also for development):
```
sim642, line 18:
```suggestion
To build Gobview (also for development):

```

docs/user-guide/inspecting.md
```diff
@@ -11,9 +11,12 @@ Modern browsers' security settings forbid some file access which is necessary fo
...
+1. Run `dune build gobview` in the analyzer directory to build the web UI
+2. Run Goblint with these flags: `--enable gobview --set save_run DIR` (`DIR` is the name of the result directory that Goblint will create and populate, if not specified it is `run`)
+3. `cd` into `DIR` and run `python3 -m http.server`
+4. Visit http://localhost:8000
```
sim642, line 22: Also to make this clickable like the g2html link above.
```suggestion
4. Visit <http://localhost:8000>
```

src/maingoblint.ml
```diff
@@ -508,6 +508,24 @@ let do_gobview () =
...
+      let file_loc = Hashtbl.create 113 in
+      let copy_rem p =
+        let _, name = BatFile.open_temporary_out ~prefix:(Fpath.filename p) ~suffix:".c" ~temp_dir:(Fpath.to_string file_dir) () in
+        let gobview_path = match Fpath.relativize ~root:(Fpath.v run_dir) (Fpath.v name) with
+          | Some p -> Fpath.to_string p
+          | None -> failwith "The gobview directory should be a prefix of the paths of c files copied to the gobview directory" in
+        Hashtbl.add file_loc (Fpath.to_string p) gobview_path;
+        BatFile.write_lines name (BatFile.lines_of (Fpath.to_string p)) in
+      let paths = Cil.foldGlobals cilfile (
+          fun acc g -> match g with GFun (_,loc) -> if not (List.mem loc.file acc) then loc.file :: acc else acc | _ -> acc) [] in
```
sim642, line 523: This only gets the files with functions in them, but there may be files which don't define any functions, but only declare them, or define only global variables/types/etc.

I think the safer route would be to iterate over `Preprocessor.dependencies`, which actually collects all locations from CIL during parsing, including any deeper inside any AST.

src/maingoblint.ml
sim642: `Fileutils.cp`? I don't see us having anything like that.

Could just use `create_symlink` anyway, which this logic does for all the other Gobview files as well. Yes, that means changing the files in their original locations changes the file shown in Gobview for the old results, but that's already the case because everything else is symlinked as well. To be fully persistent, everything needs to be copied instead of symlinking.

sim642 (changes requested):
Once https://github.com/goblint/gobview/pull/19 is merged, this also needs to update the git submodule.

src/maingoblint.ml
michael-schwarz: > `Fileutils.cp`? I don't see us having anything like that.

It is an external library that is part of #995

stilscher: I thought symlinking might not be so good, because one would like to forbid following any symlink and really limit the file access to the served directory to avoid symlink attacks.

stilscher: > Could just use create_symlink anyway, which this logic does for all the other Gobview files as well.

Most other files for Gobview (e.g. the cil file, configs, solver stats) are marshalled directly to the Gobview directory. So they are not symlinked, but there exists a separate copy of them just for Gobview. The only ones that I find that are symlinked are some javascript file for the monaco editor, the `index.html`, `main.js` and one `.ttf` file.

sim642 (approved):

PR #994, opened 2023-02-19
Add Interval Set domain for Interval Set analysis

### LINKED ISSUE #901: Add a new intDomain that deals with sets of intervals

michael-schwarz:
Copy of #966, as that PR does not allow us to commit directly to it.

Original PR:
> The following pull request will include the _Interval Set Analysis_ into Goblint, as described by the issue #901.
> 
> Closes #901.


michael-schwarz:
I have now cleaned up some things as opposed to #994 w.r.t. to code style and some unnecessary complications such as passing around `cast` everywhere. Would be great if someone could have a look to ensure I didn't break anything in the process.

src/cdomains/intDomain.ml
```diff
@@ -911,30 +944,631 @@ struct
...
+  let cmp_events x y =
+    let res = Ints_t.compare (unbox_event x) (unbox_event y) in
+    if res <> 0 then res
+    else
+      begin
+        match (x, y) with
+        | (Enter _, Exit _) -> -1
+        | (Exit _, Enter _) -> 1
+        | (_, _) -> 0
+      end
```
sim642, line 1022: I was about to suggest `[@@deriving ord]`, but I'm guessing it's intentionally comparing the ints before the event constructors? If so, then this deserves a comment, so nobody accidentally cleans this up in the future.

src/cdomains/intDomain.ml
```diff
@@ -911,30 +944,631 @@ struct
...
+        let x1y2n = x1 /. y2 in
+        let x2y1n = x2 /. y1 in
+        let x2y2n = x2 /. y2 in
+        let x1y1p = x1 /. y1 in
+        let x1y2p = x1 /. y2 in
+        let x2y1p = x2 /. y1 in
+        let x2y2p = x2 /. y2 in
+        [(min4 x1y1n x1y2n x2y1n x2y2n, max4 x1y1p x1y2p x2y1p x2y2p)]
+    end
+    in binary_op_with_norm ik x y interval_div
```
sim642, line 1352: I'm surprised by the amount of `IntervalFunctor` code this just duplicates as opposed to delegating to the existing implementation. I thought it was supposed to just do that. Now we have two copies of the massive interval implementation to keep in sync.

src/cdomains/intDomain.ml
michael-schwarz: I'll make sure to add an appropriate comment!

src/cdomains/intDomain.ml
```diff
@@ -522,51 +565,35 @@ struct
...
       if underflow || overflow then (
-        if not suppress_ovwarn then set_overflow_flag ~cast ~underflow ~overflow ik;
         if should_wrap ik then (* could add [|| cast], but that's GCC implementation-defined behavior: https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html#Integers-implementation *)
           (* We can only soundly wrap if at most one overflow occurred, otherwise the minimal and maximal values of the interval *)
           (* on Z will not safely contain the minimal and maximal elements after the cast *)
           let diff = Ints_t.abs (Ints_t.sub max_ik min_ik) in
           let resdiff = Ints_t.abs (Ints_t.sub y x) in
           if Ints_t.compare resdiff diff > 0 then
-            top_of ik
+            (top_of ik, {underflow = underflow && not suppress_ovwarn; overflow = overflow && not suppress_ovwarn})
```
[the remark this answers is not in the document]
michael-schwarz: I have now pulled this out.

src/cdomains/intDomain.ml
michael-schwarz: Added.

src/cdomains/intDomain.ml
michael-schwarz: I have now tried to get more reuse out of this by pulling out operations that are basically textbook interval arithmetic. However, I am not sure how much more we can really pull out here.

src/cdomains/intDomain.ml
```diff
@@ -687,18 +714,18 @@ struct
     | _   , true -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show i1) (show i2)))
     | _ ->
       match to_int i1, to_int i2 with
-      | Some x, Some y -> (try norm ik (of_int ik (f ik x y)) with Division_by_zero -> top_of ik)
```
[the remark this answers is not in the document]
michael-schwarz: `of_int` already performs `norm` here.

src/cdomains/intDomain.ml
```diff
@@ -687,18 +714,18 @@ struct
...
 
   let bitcomp f ik i1 i2 =
     match is_bot i1, is_bot i2 with
-    | true, true -> bot_of ik
+    | true, true -> (bot_of ik,{underflow=false; overflow=false})
     | true, _
     | _   , true -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show i1) (show i2)))
     | _ ->
       match to_int i1, to_int i2 with
-      | Some x, Some y -> (try norm ik (of_int ik (f ik x y)) with Division_by_zero | Invalid_argument _ -> top_of ik)
```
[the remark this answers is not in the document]
michael-schwarz: Dame reasoning here.

michael-schwarz:
I would propose to merge this soon, in the interest of avoiding having such a huge PR go stale (and this becoming the third attempt at interval sets that ends up not being merged).

This does not prevent us from doing refactoring to maximize reuse at a later point.

src/cdomains/intDomain.ml
sim642: Is there a reason interval sets cannot just delegate directly to our interval domain for all operations? I mean not just extracting some common internal arithmetic, but just delegate the whole operation.

michael-schwarz: The problem is that  e.g. normalization is different between both: IntervalSets can represent intervals spanning over zero for unsigned types without going to top. This also means that the type is different, as normalization for IntervalSests would lead to intervals here (to avoid having intervals with an upper bound smaller than the lower bound).

sim642: > IntervalSets can represent intervals spanning over zero for unsigned types without going to top.


So `IntervalSet` is not just a disjunctive completion of intervals, which simply lifts all operations directly, (as described by e.g. Miné's numeric tutorial) but additionally borrows elements from the circular intervals domain that we used to have for such wrap-around intervals?
It would be good to have such details documented because I always had the impression that this is just a lifting to sets (implemented as sorted lists) with a normalization procedure (which joins overlapping intervals).

michael-schwarz: I added a comment to this effect now.

sim642: 030aebc780a815fcf294f3e659515dc4f04ef477 added the comment to the existing single interval domain, not the new interval set domain.

michael-schwarz: True, wasn't paying close enough attention apparently.

sim642 (approved):
All the comments have been resolved, so I guess this should be in good enough state to merge.

I'll adapt #943 accordingly after this is merged.

PR #995, opened 2023-02-19
Skip preprocessing for `.i` files and add option `pre.enabled` to always skip preprocessor

### LINKED ISSUE #898: Do not preprocess `.i` files
`gcc` man page describes its behavior depending on the input file extension:
```
       file.c
           C source code that must be preprocessed.

       file.i
           C source code that should not be preprocessed.
```
We could also follow this convention and skip executing the preprocessor subprocess on `.i` files, where it's completely pointless. This is the case for many SV-COMP tasks, although it probably won't make a performance difference.

Since SV-COMP tasks are always supposed to be preprocessed, even as `.c` files, we could even have an option that assumes any input to be preprocessed regardless of extension. Then we wouldn't even have the GCC dependency in SV-COMP.

michael-schwarz:
Closes #898.


src/maingoblint.ml
```diff
@@ -209,11 +211,16 @@ let basic_preprocess ~all_cppflags fname =
...
+  if ext <> ".i" && not (GobConfig.get_bool "pre.skipcpp") then
+    (* Preprocess using cpp. *)
+    let arguments = all_cppflags @ Fpath.to_string fname :: "-o" :: Fpath.to_string nname :: [] in
+    let command = Filename.quote_command (Preprocessor.get_cpp ()) arguments in
+    if get_bool "dbg.verbose" then print_endline command;
+    (nname, Some {ProcessPool.command; cwd = None})
+  else
+    (* No preprocessing needed. *)
+    (FileUtil.cp [(Fpath.to_string fname)] (Fpath.to_string nname);
+     (nname, None))
```
sim642, line 223: The copying is completely unnecessary here though, we could just use the original file directly. The copy goes to the temporary `.goblint` directory, which is deleted after preprocessing anyway.

So we could avoid dragging in another dependency.

src/util/options.schema.json
```diff
@@ -223,6 +223,12 @@
           "items": { "type": "string" },
           "default": []
         },
+        "skipcpp": {
+          "title": "pre.skipcpp",
+          "description": "Skip running the C preprocessor.",
+          "type": "boolean",
+          "default": false
```
sim642, line 230: We should enable this in the `svcomp` conf, but not the older SV-COMP confs because they didn't have that yet.

src/maingoblint.ml
sim642: We do have the fileutils library from #993 now, but even then this copying can be avoided.

michael-schwarz: I thought keeping it this way would make it more consistent with `pre.keep`: Otherwise some files would be in the corresponding directory, and some would remain at their original path.
I don't think copying files is too expensive anyway.

sim642: The directory is `.goblint/preprocessed` though, which would no longer be precisely true.

The bigger problem with copying is that the file paths in locations become incorrect: if the file isn't actually preprocessed from its original path into that directory, but is copied verbatim, then there are no `#line`, etc pragmas to indicate the original paths for CIL. Instead, it will be using the file's "actual" path, i.e. `.goblint/preprocessed/something.c`, which isn't where the `something.c` file originally is for the user. So the paths will be wrong in messages and not correctly displayable in GobPie.

michael-schwarz: True. I have now removed the copying in this case.

sim642:
I fixed the issue by forcing preprocessing of our stubs. Also flipped the option to `pre.enabled` to make it consistent with other options and avoid the double-negation by default.

PR #996, opened 2023-02-20
Add missing library functions for Concrat benchmarks

sim642:
Adds a bunch of missing library function specifications for https://github.com/goblint/bench/pull/53.

### DIFF
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -109,6 +109,10 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("timer_gettime", unknown [drop "timerid" []; drop "curr_value" [w_deep]]);
     ("timer_getoverrun", unknown [drop "timerid" []]);
     ("lstat", unknown [drop "pathname" [r]; drop "statbuf" [w]]);
+    ("getpwnam", unknown [drop "name" [r]]);
+    ("strndup", unknown [drop "s" [r]; drop "n" []]);
+    ("freeaddrinfo", unknown [drop "res" [f_deep]]);
+    ("getgid", unknown []);
   ]
 
 (** Pthread functions. *)
@@ -201,6 +205,7 @@ let glibc_desc_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("svcerr_systemerr", unknown [drop "xprt" [r_deep; w_deep]]);
     ("svc_sendreply", unknown [drop "xprt" [r_deep; w_deep]; drop "outproc" [s]; drop "out" [r]]);
     ("shutdown", unknown [drop "socket" []; drop "how" []]);
+    ("getaddrinfo_a", unknown [drop "mode" []; drop "list" [w_deep]; drop "nitems" []; drop "sevp" [r; w; s]]);
   ]
 
 let linux_userspace_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -101,6 +101,12 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("shm_open", unknown [drop "name" [r]; drop "oflag" []; drop "mode" []]);
     ("sched_get_priority_max", unknown [drop "policy" []]);
     ("mprotect", unknown [drop "addr" []; drop "len" []; drop "prot" []]);
+    ("ftime", unknown [drop "tp" [w]]);
+    ("timer_create", unknown [drop "clockid" []; drop "sevp" [r; w; s]; drop "timerid" [w]]);
+    ("timer_settime", unknown [drop "timerid" []; drop "flags" []; drop "new_value" [r_deep]; drop "old_value" [w_deep]]);
+    ("timer_gettime", unknown [drop "timerid" []; drop "curr_value" [w_deep]]);
+    ("timer_getoverrun", unknown [drop "timerid" []]);
+    ("lstat", unknown [drop "pathname" [r]; drop "statbuf" [w]]);
   ]
 
 (** Pthread functions. *)
@@ -119,6 +125,8 @@ let pthread_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("pthread_setcanceltype", unknown [drop "type" []; drop "oldtype" [w]]);
     ("pthread_detach", unknown [drop "thread" []]);
     ("pthread_attr_setschedpolicy", unknown [drop "attr" [r; w]; drop "policy" []]);
+    ("pthread_condattr_init", unknown [drop "attr" [w]]);
+    ("pthread_condattr_setclock", unknown [drop "attr" [w]; drop "clock_id" []]);
   ]
 
 (** GCC builtin functions.
--- a/src/analyses/libraryFunctions.ml
+++ b/src/analyses/libraryFunctions.ml
@@ -44,6 +44,7 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("setbuf", unknown [drop "stream" [w]; drop "buf" [w]]);
     ("swprintf", unknown (drop "wcs" [w] :: drop "maxlen" [] :: drop "fmt" [r] :: VarArgs (drop' [])));
     ("assert", special [__ "exp" []] @@ fun exp -> Assert { exp; check = true; refine = get_bool "sem.assert.refine" }); (* only used if assert is used without include, e.g. in transformed files *)
+    ("difftime", unknown [drop "time1" []; drop "time2" []]);
   ]
 
 (** C POSIX library functions.
@@ -88,6 +89,7 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("getsockopt", unknown [drop "sockfd" []; drop "level" []; drop "optname" []; drop "optval" [w]; drop "optlen" [w]]);
     ("gethostbyaddr", unknown [drop "addr" [r_deep]; drop "len" []; drop "type" []]);
     ("gethostbyaddr_r", unknown [drop "addr" [r_deep]; drop "len" []; drop "type" []; drop "ret" [w_deep]; drop "buf" [w]; drop "buflen" []; drop "result" [w]; drop "h_errnop" [w]]);
+    ("sigaction", unknown [drop "signum" []; drop "act" [r_deep; s_deep]; drop "oldact" [w_deep]]);
   ]
 
 (** Pthread functions. *)
@@ -104,6 +106,7 @@ let pthread_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("pthread_key_delete", unknown [drop "key" [f]]);
     ("pthread_cancel", unknown [drop "thread" []]);
     ("pthread_setcanceltype", unknown [drop "type" []; drop "oldtype" [w]]);
+    ("pthread_detach", unknown [drop "thread" []]);
   ]
 
 (** GCC builtin functions.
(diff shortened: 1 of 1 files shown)


src/analyses/libraryFunctions.ml
```diff
@@ -44,6 +44,14 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("setbuf", unknown [drop "stream" [w]; drop "buf" [w]]);
     ("swprintf", unknown (drop "wcs" [w] :: drop "maxlen" [] :: drop "fmt" [r] :: VarArgs (drop' [])));
     ("assert", special [__ "exp" []] @@ fun exp -> Assert { exp; check = true; refine = get_bool "sem.assert.refine" }); (* only used if assert is used without include, e.g. in transformed files *)
+    ("difftime", unknown [drop "time1" []; drop "time2" []]);
+    ("system", unknown [drop "command" [r]]);
+    ("wcscat", unknown [drop "dest" [r; w]; drop "src" [r]]);
+    ("abs", unknown [drop "j" []]);
```
michael-schwarz, line 50: Maybe worth actually handling? Seems simple enough?

src/analyses/libraryFunctions.ml
```diff
@@ -44,6 +44,14 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("localtime_r", unknown [drop "timep" [r]; drop "result" [w]]);
+    ("strsep", unknown [drop "stringp" [r_deep; w]; drop "delim" [r]]);
```
michael-schwarz, line 52: This one should probably go to glibc or linux, it is not standard mandated: https://man7.org/linux/man-pages/man3/strsep.3.html

src/analyses/libraryFunctions.ml
```diff
@@ -44,6 +44,14 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("strcasestr", unknown [drop "haystack" [r]; drop "needle" [r]]);
```
michael-schwarz, line 53: same here: https://man7.org/linux/man-pages/man3/strstr.3.html

src/analyses/libraryFunctions.ml
```diff
@@ -44,6 +44,14 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("strpbrk", unknown [drop "s" [r]; drop "accept" [r]]);
   ]
```
michael-schwarz, line 55: For these it would be nice if we could somehow specify that the resulting pointer still points inside the first blob to avoid having unknown pointers. Just a thought though, no need to do this here.

src/analyses/libraryFunctions.ml
```diff
@@ -88,6 +96,43 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("getsockopt", unknown [drop "sockfd" []; drop "level" []; drop "optname" []; drop "optval" [w]; drop "optlen" [w]]);
     ("gethostbyaddr", unknown [drop "addr" [r_deep]; drop "len" []; drop "type" []]);
     ("gethostbyaddr_r", unknown [drop "addr" [r_deep]; drop "len" []; drop "type" []; drop "ret" [w_deep]; drop "buf" [w]; drop "buflen" []; drop "result" [w]; drop "h_errnop" [w]]);
+    ("sigaction", unknown [drop "signum" []; drop "act" [r_deep; s_deep]; drop "oldact" [w_deep]]);
+    ("tcgetattr", unknown [drop "fd" []; drop "termios_p" [r_deep]]);
```
michael-schwarz, line 100: Is `r_deep` right here for `termios_p`? 

> The tcgetattr() function shall get the parameters associated with
       the terminal referred to by fildes and store them in the termios
       structure referenced by termios_p.

src/analyses/libraryFunctions.ml
```diff
@@ -88,6 +96,43 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("tcsetattr", unknown [drop "fd" []; drop "optional_actions" []; drop "termios_p" [w_deep]]);
```
michael-schwarz, line 101: I think here `r_deep` would suffice for `termios_p`.

src/analyses/libraryFunctions.ml
```diff
@@ -88,6 +96,43 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("timer_getoverrun", unknown [drop "timerid" []]);
+    ("lstat", unknown [drop "pathname" [r]; drop "statbuf" [w]]);
+    ("getpwnam", unknown [drop "name" [r]]);
+    ("strndup", unknown [drop "s" [r]; drop "n" []]);
+    ("freeaddrinfo", unknown [drop "res" [f_deep]]);
+    ("getgid", unknown []);
+    ("pselect", unknown [drop "nfds" []; drop "readdfs" [r]; drop "writedfs" [r]; drop "exceptfds" [r]; drop "timeout" [r]; drop "sigmask" [r]]);
+    ("strncasecmp", unknown [drop "s1" [r]; drop "s2" [r]; drop "n" []]);
+    ("getnameinfo", unknown [drop "addr" [r_deep]; drop "addrlen" []; drop "host" [w]; drop "hostlen" []; drop "serv" [w]; drop "servlen" []; drop "flags" []]);
+    ("strtok_r", unknown [drop "str" [r; w]; drop "delim" [r]; drop "saveptr" [w]]);
```
michael-schwarz, line 120: Does `w` suffice for `saveptr` or would we need `w_deep` here?

src/analyses/libraryFunctions.ml
```diff
@@ -88,6 +96,43 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("kill", unknown [drop "pid" []; drop "sig" []]);
+    ("closelog", unknown []);
+    ("dirname", unknown [drop "path" [r]]);
+    ("setpgid", unknown [drop "pid" []; drop "pgid" []]);
+    ("dup2", unknown [drop "oldfd" []; drop "newfd" []]);
```
michael-schwarz, line 125: is `[]` right for the second argument? I thought we switched to `w` for FDs in such scenarios?

src/analyses/libraryFunctions.ml
```diff
@@ -104,6 +149,13 @@ let pthread_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("pthread_key_delete", unknown [drop "key" [f]]);
     ("pthread_cancel", unknown [drop "thread" []]);
     ("pthread_setcanceltype", unknown [drop "type" []; drop "oldtype" [w]]);
+    ("pthread_detach", unknown [drop "thread" []]);
+    ("pthread_attr_setschedpolicy", unknown [drop "attr" [r; w]; drop "policy" []]);
+    ("pthread_condattr_init", unknown [drop "attr" [w]]);
+    ("pthread_condattr_setclock", unknown [drop "attr" [w]; drop "clock_id" []]);
+    ("pthread_mutexattr_destroy", unknown [drop "attr" [f]]);
```
michael-schwarz, line 156: See also #839

src/analyses/libraryFunctions.ml
```diff
@@ -324,18 +397,44 @@ let svcomp_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
     pthread_descs_list;
     gcc_descs_list;
     glibc_desc_list;
+    linux_userspace_descs_list;
     linux_descs_list;
     goblint_descs_list;
     zstd_descs_list;
     math_descs_list;
     svcomp_descs_list;
+    ncurses_descs_list;
```
michael-schwarz, line 437: Maybe with this addition it is time to actually make these conditional? With `ncurses` and `zstd` we now have two highly specific libraries in here.

michael-schwarz (commented):
Other than the comments, this looks good to me!
It's nice that you put in the legwork here to get all of them properly classified!

src/analyses/libraryFunctions.ml
sim642: We could, yes, given that we handle `fabs` and its variants. I'd leave it as a separate issue/PR though because a new operation needs to be implemented for all int domains.

src/analyses/libraryFunctions.ml
sim642: True, I think I just saw `string.h` and thought it would be standard. I moved it to glibc because it's listed here: https://www.gnu.org/software/gnulib/manual/html_node/Glibc-string_002eh.html.

src/analyses/libraryFunctions.ml
sim642: Indeed, I must've mixed up `tcgetattr` and `tcsetattr`.

src/analyses/libraryFunctions.ml
sim642: These file descriptors are just `int`, so access specifications are meaningless (they only concern pointers passed to functions).

src/analyses/libraryFunctions.ml
```diff
@@ -88,6 +96,43 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("kill", unknown [drop "pid" []; drop "sig" []]);
+    ("closelog", unknown []);
+    ("dirname", unknown [drop "path" [r]]);
+    ("setpgid", unknown [drop "pid" []; drop "pgid" []]);
+    ("dup2", unknown [drop "oldfd" []; drop "newfd" []]);
+    ("pclose", unknown [drop "stream" [w; f]]);
+    ("getcwd", unknown [drop "buf" [w]; drop "size" []]);
+    ("inet_pton", unknown [drop "af" []; drop "src" [r]; drop "dst" [w]]);
+    ("inet_ntop", unknown [drop "af" []; drop "src" [r]; drop "dst" [w]; drop "size" []]);
+    ("inet_aton", unknown [drop "cp" [r]; drop "inp" [w]]);
```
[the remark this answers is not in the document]
sim642: True, I think I just put it here together with the others, but I'll move it to glibc since it's listed here: https://www.gnu.org/software/gnulib/manual/html_node/Glibc-arpa_002finet_002eh.html.

src/analyses/libraryFunctions.ml
```diff
@@ -145,6 +198,9 @@ let gcc_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_popcountll", unknown [drop "x" []]);
     ("__atomic_store_n", unknown [drop "ptr" [w]; drop "val" []; drop "memorder" []]);
     ("__atomic_load_n", unknown [drop "ptr" [r]; drop "memorder" []]);
+    ("__sync_fetch_and_add", unknown (drop "ptr" [r; w] :: drop "value" [] :: VarArgs (drop' [])));
+    ("__sync_fetch_and_sub", unknown (drop "ptr" [r; w] :: drop "value" [] :: VarArgs (drop' [])));
```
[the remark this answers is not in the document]
sim642: Possibly. There's a whole lot of other atomic operations (including C11 ones) that could also be handled similarly.

src/analyses/libraryFunctions.ml
sim642: The zstd ones are certainly very specific since they are used in the zstd implementation itself. ncurses at least has ones used by actual library users and is a widely used library.

I'll create some activation mechanism for these groups in a follow-up PR. Though it won't be precise until all of the old-style things are converted to new ones in appropriate groups:
https://github.com/goblint/analyzer/blob/bcb4de29e22885570ba82f7ec9cee2e661854b7a/src/analyses/libraryFunctions.ml#L359-L400
https://github.com/goblint/analyzer/blob/bcb4de29e22885570ba82f7ec9cee2e661854b7a/src/analyses/libraryFunctions.ml#L500-L980

michael-schwarz (approved):

PR #999, opened 2023-02-22
Render options JSON schema to Readthedocs

sim642:
Uses json-schema-for-humans to render the options JSON schema to HTML and adds it into Readthedocs (options reference link in the left menu bar).

Docs preview for this PR: https://goblint--999.org.readthedocs.build/en/999/.


vesalvojdani:
This would be quite useful, so might be worth making this link more prominent. Typically a link like [`this.c`](http://non-obvious.link) just links to the file. But more importantly, what is the Markdown output like? Would it be rendered better and directly on readthedocs/github? I'm just not at all a fan of these super-padded modern designs, such as https://cs.ut.ee, where it's no longer possible to find anything.

sim642:
> This would be quite useful, so might be worth making this link more prominent. Typically a link like [`this.c`](http://non-obvious.link) just links to the file.

It's also separately in the docs side bar.


> But more importantly, what is the Markdown output like? Would it be rendered better and directly on readthedocs/github?

Markdown doesn't allow such arbitrarily nested structure, so it'd be more or less flat, at which point you might as well search the original schema file directly like we've been doing so far.

Also, I don't know if the Markdown flavor used by json-schema-for-humans is the same one as the one used by Readthedocs (which has some oddities).
There's no nice way of rendering it directly on GitHub because the rendered version would have to be committed into the repository then, but that requires keeping it in sync.


> I'm just not at all a fan of these super-padded modern designs, such as https://cs.ut.ee, where it's no longer possible to find anything.

The CSS for this can be customized to have less padding.

sim642:
I added a Markdown version into the Readthedocs preview now. It's mostly boilerplate and less actually useful information.

Maybe the best option would be to at some point write our own custom Markdown conversion for it (in OCaml), which can omit the irrelevant parts of JSON schemas and emphasize the parts we do use.

michael-schwarz (approved):

sim642:
The question is, which version do we want to have right now: HTML or Markdown? Both would be rather redundant.

michael-schwarz:
I personally prefer the HTML version, but am happy with either one!

PR #1000, opened 2023-02-22
Handle pseudo return nodes in `Node.of_id` and improve their location

### LINKED ISSUE #997: Handle pseudo return nodes in server mode `cfg/lookup` request
Follow-up to #988.

The handling involves two aspects:
- [x] `Node.of_id` working for pseudo return node statement IDs. Currently the `Cilfacade` map does not contain them.
- [x] `Locator` handling pseudo return nodes. Or maybe it shouldn't because those should be synthetic?

sim642:
Closes #997.

### Changes
1. Adds `Cilfacade.pseudo_return_stmt_sids`, which maps pseudo return node statement IDs to the statements themselves. This allows `Cilfacade.find_stmt_sid`, which is used by `Node.of_id` to work for pseudo return nodes.
2. Previously pseudo return nodes were given the location of the entire function definition, which starts already before the function's return type. This PR changes their location to be the end location of the function definition (with 0-width range), corresponding to the `}`. Since `Locator` uses nodes from the CFG, this already includes pseudo return nodes, but now with more reasonable locations for both lookup and display.

There is one unexpected point though as indicated by the change to the cram test: our LLoC counting can increase because the pseudo return node at its new location is on a separate line. Not sure if this weird or fine.

### DIFF
--- a/src/framework/cfgTools.ml
+++ b/src/framework/cfgTools.ml
@@ -233,12 +233,13 @@ let createCFG (file: file) =
          * lazy, so it's only added when actually needed *)
         let pseudo_return = lazy (
           if Messages.tracing then Messages.trace "cfg" "adding pseudo-return to the function %s.\n" fd.svar.vname;
-          let newst = mkStmt (Return (None, fd_loc)) in
+          let fd_end_loc = {fd_loc with line = fd_loc.endLine; byte = fd_loc.endByte; column = fd_loc.endColumn} in
+          let newst = mkStmt (Return (None, fd_end_loc)) in
           newst.sid <- get_pseudo_return_id fd;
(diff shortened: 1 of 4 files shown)


michael-schwarz:
Btw everyone, this is no #1000 :tada:

stilscher (approved):

sim642:
> There is one unexpected point though as indicated by the change to the cram test: our LLoC counting can increase because the pseudo return node at its new location is on a separate line. Not sure if this weird or fine.

It might be possible to avoid this change by making dead code statistics ignore pseudo return nodes. Maybe that's what we should be doing because they don't correspond to anything real in the user code anyway?

stilscher:
I think that is a good idea.

PR #1004, opened 2023-02-28
Fix incremental write-only restarting for side effects from deleted nodes

sim642:
Before this fix, the added test case failed because old accesses from the deleted node (in a completely changed function) are not restarted.

This issue was revealed when re-running some incremental restarting benchmarks. `aget_comb02.patch` with RestartGlob failed to become more precise due to this issue.

### DIFF
--- a/tests/incremental/13-restart-write/09-mutex-self-fix.json
+++ b/tests/incremental/13-restart-write/09-mutex-self-fix.json
@@ -3,7 +3,8 @@
     "restart": {
       "sided": {
         "enabled": false
-      }
+      },
+      "write-only": true
     }
   }
 }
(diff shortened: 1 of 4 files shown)


tests/incremental/13-restart-write/09-mutex-self-fix.json
```diff
@@ -0,0 +1,9 @@
+{
+  "incremental": {
+    "restart": {
+      "sided": {
+        "enabled": false
+      }
```
stilscher, line 6: I think it would be good to include the `incremental.restart.write-only` property in this configuration even though it is set to true by default, because it is essential for this test and these configurations break easily when we make changes to our default configurations.

stilscher, line 6: I think it would be good to include the `incremental.restart.write-only` property in this configuration even though it is set to true by default, because it is essential for this test and these configurations break easily when we make changes to our default configurations.

stilscher (approved):

PR #1007, opened 2023-03-01
ThreadCreate analysis to improve partially context-sensitive threadID analysis

FelixKrayer:
- An option is added that allows performing the existing threadID analysis partially context-sensitively (sensitive with respect to the current threadID but insensitive to the set of encountered create edges).
- A new analysis is added that tracks wether a thread is created during a call (threadCreate). If so, the combine of threadID needs to return the callee state, but if not, we are sure that the caller state is still valid (and possibly more precise than the callee state, if threadID is performed partially context-sensitively).

TODO:
- [ ]  Instead of tracking if a thread is created in a call, track a set of encountered create edges. The combine of the threadID analysis then should only add edges that are encountered in the call to the set of encountered edges -> more often useful than the current approach


FelixKrayer:
> Thank you for the PR! We're a bit busy this week, but I'll be sure to look it over early next week.

Ok, great! I'l try to add some more test cases until then.

src/util/options.schema.json
```diff
@@ -935,6 +935,19 @@
                 "Whether the node at which a thread is created is part of its threadid",
               "type": "boolean",
               "default" : true
+            },
+            "context": {
+              "title": "ana.thread.context",
+              "type": "object",
+              "properties": {
+                "createEdges": {
+                  "title": "ana.thread.context.createEdges",
```
michael-schwarz, line 944: Maybe `create-edges`? We don't really use camelCase for any other option.

src/analyses/threadCreateEdges.ml
```diff
@@ -0,0 +1,52 @@
...
+      | _ -> ctx.local
+
+  let startstate v = D.bot ()
+  let threadenter ctx lval f args = [D.bot ()]
+  let threadspawn ctx lval f args fctx = 
+    match ctx.local with
+    | `Top -> `Top
+    | `Lifted ces -> D.lift (Thread.threadspawn ces ctx.prev_node f)
+
+  let exitstate  v = D.top () (* is this correct? What exactly is exitstate?*)
```
michael-schwarz, line 42: `exitstate` is the initial state of cleanup functions if there are any. This mechanism is currently (no longer) well thought out and we only have one test for it. 
Us core team should probably in the future have another look at `exitfun` altogether, for now `T` is good here!

src/analyses/threadCreateEdges.ml
```diff
@@ -0,0 +1,52 @@
...
+  let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask: Queries.ask) : D.t =
+    (* after a call, join both both sets of encountered create edges from caller and callee*)
+    if M.tracing then M.trace "threadCreateEdges" "caller: %a\ncallee: %a\nboth: %a\n\n" D.pretty ctx.local D.pretty au D.pretty (D.join ctx.local au); 
+    D.join ctx.local au
+
+  let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    let desc = LibraryFunctions.find f in
+    match desc.special arglist with
+      | Unknown -> D.top () (*unknown function: Maybe a unknown thread was created*)
+      | _ -> ctx.local
```
michael-schwarz, line 33: Is this consistent with what the standard thread analysis currently does?

michael-schwarz (commented):
Thanks for the PR! I added a few small questions,once these are addressed, we can move forward with merging this!

src/analyses/threadCreateEdges.ml
FelixKrayer: I agree, changing it to just return `ctx.local` makes it more precise and consistent.

FelixKrayer:
Are there plans to perform SV-Benchmarks on this for the "no-data-race" property to compare computation time and precision?

michael-schwarz:
Sure, I can try to start such runs tomorrow :+1:

michael-schwarz:
Here are the results:

The order is:

- <base conf, e.g. sv-comp23>
- <base conf> --disable ana.thread.context.create-edges
- <base conf> --disable ana.thread.context.create-edges --set ana.activated[+] threadCreateEdges

[no-data-race.zip](https://github.com/goblint/analyzer/files/11048274/no-data-race.zip)
[no-overflow.zip](https://github.com/goblint/analyzer/files/11048276/no-overflow.zip)
[unreach-call.zip](https://github.com/goblint/analyzer/files/11048277/unreach-call.zip)


Unfortunately, we once more don't see a real difference. I think this is still worth having, as it may well be the key ingredient making the analysis of _some_ program feasible in the future.

FelixKrayer:
Thanks, I will have a look at the results

src/util/options.schema.json
```diff
@@ -954,6 +954,19 @@
...
+            },
+            "context": {
+              "title": "ana.thread.context",
+              "type": "object",
+              "properties": {
+                "create-edges": {
+                  "title": "ana.thread.context.create-edges",
+                  "description": "threadID analysis: Encountered create edges in context.",
+                  "type": "boolean",
+                  "default": true
```
sim642, line 966: Isn't this option a bit redundant? It's also misleading because it's `true` by default, but actually that on its own doesn't do anything and also requires threadCreateEdges analysis to be activated.

Seems like the analysis activation could be enough of a toggle. 

---

**EDIT:** Given my confusion in the other comment, it now seems that maybe there should only be this option and no extra analysis?

src/analyses/threadCreateEdges.ml
```diff
@@ -0,0 +1,49 @@
+(* threadCreateEdges: Tracks which create edges have been encountered since the beginning of the current function *)
```
sim642, line 1: I'm a bit puzzled by what this is really doing. This is just using the thread ID domain as its local domain and delegating to the thread ID analysis. It's not tracking any additional thread creation edges, but simply reusing the set from within the thread ID domain. And then the thread ID analysis queries this for the set.

Why this roundabout way? if the data already exists inside the thread ID analysis domain and is tracked by the thread ID analysis, why can't it just directly use the set itself?

sim642 (changes requested):

src/util/options.schema.json
michael-schwarz: This is true by default, as are many context-sensitivity options. Disabling this makes the analysis less precise. If `threadCreateEdges` is enabled some of the precision can be recovered again.
In this way, it is analogous to the mechanism for improving int values for partial contexts Felix implemented for base.

src/analyses/threadCreateEdges.ml
sim642: Could the existing created set in the thread ID domain simply be replaced with a pair of sets then? One tracking the existing thing and another the one which starts out empty.
Or maybe just have the new analysis's local value be added as the third one to thread ID analysis?

michael-schwarz: One could for sure do that. The question just is if this does not cause overhead in the default configuration? Currently we don't have any overhead if these options are disabled.
Also, if I am context-sensitive in the set of encountered thread creation edges, this additional component is not helpful.

src/util/options.schema.json
sim642: Right, so there's three levels of precision.

src/analyses/threadCreateEdges.ml
sim642: I suspect the overhead of an extra analysis has more overhead than an extra component in an existing analysis, but that all depends on which is the default. Since this analysis isn't the default, then it might be slightly better this way.

Although I'm still wondering if the query is necessary. Couldn't the new analysis simply be context-sensitive instead of making threadid query this one for what to use in its context?

michael-schwarz:
Closing in favor of #1077.

PR #1012, opened 2023-03-14
Add `cil/varinfos` request to server mode

sim642:
This returns a list of data for all `varinfo`s known to CIL, including:
1. `vid` to be used for other requests since name is not globally unique,
2. `original_name` to be used for mapping to actual code names (or detecting temporaries if missing),
3. `role` and `function` for determining the scope.

This is provided as a list and a client may construct suitably grouped mappings from it.


sim642:
I added a `richvarinfos` request directly on master: https://github.com/goblint/analyzer/commit/db5843fb17f527591f83d4bc148873f3185774d6. That should return all of those alloc variables.

PR #1013, opened 2023-03-14
Add `global-state` request to server mode

sim642:
Builds on #1012.

This request allows getting the entire global state (not too practical), but also filtering by constraint variable queries (e.g. by variable `vid` or node). The latter could be useful for displaying relevant global invariant states in the abstract debugger, relating to the current node or global variables in multi-threaded mode.


FeldrinH:
I currently don't see much use for this in my use case. `arg/eval` and `cil/varinfos` together supersede most of the functionality this would have provided for abstract debugging i.e. finding and showing global variables.

PR #1015, opened 2023-03-17
Refactor `setjmp`/`longjmp` analysis

sim642:
This is my ongoing attempt to polish #970 for merging. It includes #970 and is currently against master to see the overall diff, but later it might be good to switch this to be against #970 to see the diff only on top of that.


tests/regression/66-longjmp/50-arguments-non-top.c
michael-schwarz, line 1: I think this is an instance where the failing asserts are not indicative of a flaw in the analyses, but highlight an issue with our assert checking: One one path `val == 1` holds, and on the other one it does not. However, our annotations do not really work well with path-sensitive analyses.

michael-schwarz: Ah, good point!

src/analyses/base.ml
```diff
@@ -2242,31 +2250,43 @@ struct
...
       let ensure_not_zero rv = match rv with
-        | `Int i when ID.to_bool i = Some true -> rv
-        | `Int i when ID.to_bool i = Some false -> M.warn "Must: Longjmp with a value of 0 is silently changed to 1"; `Int (ID.of_int (ID.ikind i) Z.one)
-        | `Int i when ID.to_bool i = None -> M.warn "May: Longjmp with a value of 0 is silently changed to 1"; `Int (ID.meet i (ID.of_excl_list (ID.ikind i) [Z.one]))
-        | _ -> M.warn "Arguments to longjmp are strange!"; rv
+        | `Int i ->
+          begin match ID.to_bool i with
+            | Some true -> rv
+            | Some false ->
+              M.error "Longjmp with a value of 0 is silently changed to 1";
```
michael-schwarz, line 2275: I think one might still wants to stress must/may more here rather than just emitting messages of different severity.

src/analyses/activeLongjmp.ml
```diff
@@ -5,49 +5,29 @@ open Analyses
...
+    | Longjmp {env; value}, _ ->
       (* Set target to current value of env *)
       let bufs = ctx.ask (EvalJumpBuf env) in
       bufs, JmpBufDomain.NodeSet.singleton(ctx.prev_node)
     | _ -> ctx.local
 
   let startstate v = D.bot ()
-  let threadenter ctx lval f args = [D.top ()]
-  let threadspawn ctx lval f args fctx = ctx.local
+  let threadenter ctx lval f args = [D.top ()] (* TODO: why other threads start with top? *)
```
michael-schwarz, line 28: I think it doesn't matter here. Initially there are no active `longjmps`. Whenever a `longjmp` happens, these get overwritten, and `longjmps` are only propagated between special nodes, so no joins happen.

src/analyses/base.ml
sim642: I've added that back. I blindly thought that elsewhere we switched to using severity entirely but all of them actually say must/may naturally in the message.

src/analyses/activeLongjmp.ml
sim642: Ah, yes! I've added this explanation as a comment and changed it to `bot` for consistency.

src/framework/constraints.ml
```diff
@@ -1544,6 +1385,239 @@ struct
...
+            local = synced;
+          }
+        in
+        longfd_ctx
+      in
+      let combined = lazy ( (* does not depend on target, do at most once *)
+        (* Globals are non-problematic here, as they are always carried around without any issues! *)
+        (* A combine call is mostly needed to ensure locals have appropriate values. *)
+        (* Using f from called function on purpose here! Needed? *)
+        S.combine cd_ctx lv e f args fc longfd_ctx.local (Analyses.ask_of_ctx longfd_ctx)
```
michael-schwarz, line 1501: Is actually using `lv` correct here? Will this not lead to spurious modifications (invalidations?!) of variables on the left-hand side of calls that are jumped through?

I added an example highlighting this issue!

michael-schwarz:
Impressive refactoring, great job @sim642! :clap: :clap: :clap: 

If we manage to fix the issue with combines in longjmps incorrectly modifying the LHS, I think this is read to be merged into https://github.com/goblint/analyzer/tree/longjmp and that into master.

src/framework/constraints.ml
sim642: Indeed, using `None` instead of the actual lval fixes the problem. I think in #970 this combine was called with very fictitious arguments for the function, etc, so I changed those to still be accurate, but lval is the one to change though.

michael-schwarz (approved):

PR #1021, opened 2023-03-28
Thread Wrappers

### LINKED ISSUE #664: Generalize concepts of `wrappers` to thread creation
Akin to our list of `malloc_wrappers` we should have `thread_create_wrappers` to generate better thread ids when all thread creations happen through a helper function wrapping `pthread_create`. This may, e.g. happen where code can work with different threading libraries with the specific of these threading libraries abstracted away by intermediate helpers.

The general notion is that one can maintain an additional node to use instead of the node at which the actual creation happens for objects for which we are creation-node-sensitive (e.g. `malloc`, `pthread_create`, ...).

just-max:
Extends the concept of `malloc` wrappers to thread creation. Threads receive an ID based on the CFG node of the wrapper function's call, instead of the `pthread_create` call, which in the case of a wrapper is always the same node.

Closes #664.


src/analyses/threadId.ml
```diff
@@ -104,13 +104,19 @@ struct
     else
       None
 
+  let node_for_ctx ctx = match (ctx.ask Queries.ThreadId) with
+    | `Lifted node -> node
+    | _ -> ctx.prev_node
+
   let threadenter ctx lval f args =
-    let+ tid = create_tid ctx.local ctx.prev_node f in
+    (* x *)
```
michael-schwarz, line 112: why the `(* x *)`?

src/analyses/threadId.ml
```diff
@@ -104,13 +104,19 @@ struct
...
+    let+ tid = create_tid ctx.local (node_for_ctx ctx) f in
     (tid, TD.bot ())
 
   let threadspawn ctx lval f args fctx =
     let (current, td) = ctx.local in
-    (current, Thread.threadspawn td ctx.prev_node f)
+    (* x *)
```
michael-schwarz, line 118: Same here

src/analyses/threadId.ml
just-max: The PR is still in a draft state, needs tests and a cleanup still!

src/domains/queries.ml
```diff
@@ -117,6 +126,7 @@ type _ t =
   | IterSysVars: VarQuery.t * Obj.t VarQuery.f -> Unit.t t (** [iter_vars] for [Constraints.FromSpec]. [Obj.t] represents [Spec.V.t]. *)
   | MayAccessed: AccessDomain.EventSet.t t
   | MayBeTainted: LS.t t
+  (* |  *)
```
michael-schwarz, line 129: Why `(* | *)`?

src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -0,0 +1,208 @@
...
+
+    (* Description that gets appended to the varinfo-name in user output. *)
+    let describe_varinfo (v: varinfo) (t, node, c) =
+      let loc = UpdateCil.getLoc node in
+      CilType.Location.show loc
+
+    let name_varinfo (t, node, c) =
+      Format.asprintf (* TODO *) "(alloc@sid:%s@tid:%s(#%s))" (Node.show_id node) (ThreadLifted.show t) (Chain.show c)
+
+  end
```
michael-schwarz, line 64: I don't quite get what this is for yet.

michael-schwarz:
We now have an issue with `Duplicate test ID 66/01` here. The reason is that master got a group 66 in the meanwhile, so we should change the group number to the next free one.

src/domains/queries.ml
```diff
@@ -13,6 +13,15 @@ module TS = SetDomain.ToppedSet (CilType.Typ) (struct let topname = "All" end)
 module ES = SetDomain.Reverse (SetDomain.ToppedSet (CilType.Exp) (struct let topname = "All" end))
 module VS = SetDomain.ToppedSet (CilType.Varinfo) (struct let topname = "All" end)
 
+
+(* TODO: where to put this *)
```
michael-schwarz, line 17: I think here is fine!

src/util/debug.ml
```diff
@@ -0,0 +1,6 @@
+open GoblintCil
+
+let f = Printf.sprintf
+let pf fmt = Printf.ksprintf print_endline fmt
+let df fmt = Pretty.gprintf (Pretty.sprint ~width:max_int) fmt
+let dpf fmt = Pretty.gprintf (fun doc -> print_endline @@ Pretty.sprint ~width:max_int doc) fmt
```
michael-schwarz, line 6: Instead of introducing this extra mechanism, it might be nice to rely on Tracing which we already use in a lot of places throughout Goblint.

src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -0,0 +1,257 @@
...
+
+  let dbg = WrapperArgs.is_wrapped (LibraryDesc.ThreadCreate { thread = Cil.integer 1; start_routine = Cil.integer 1; arg = Cil.integer 1; })
+  let st name ctx = if dbg then dpf"----------------------------------\n[%s] prev_node=%a => node=%a" name Node.pretty ctx.prev_node Node.pretty ctx.node
+  
+  (* TODO:
+    Does it matter if this is node or prev_node? [malloc] analysis used ctx.node and seemed to care.
+    Thread ID analysis is using ctx.prev_node (which makes more sense, since that's where the thread_create edge is,
+    and would keep two wrapper calls apart if they are e.g. both on edges leading into a join point)
+    https://github.com/goblint/analyzer/commit/77c0423640c50bb82e4290bcc97f33d4082715d0
+      *)
```
michael-schwarz, line 39: Intuitively it makes more sense to use `prev_node`. Do things break if one uses that everywhere?

src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -0,0 +1,257 @@
...
+    st "enter" ctx;
+    let counter, wrapper_node = ctx.local in
+    let new_wrapper_node =
+      if Hashtbl.mem wrappers f.svar.vname then
+        begin
+        if dbg then dpf"is wrapper";
+        match wrapper_node with
+        (* if an interesting callee is called by an interesting caller, then we remember the caller context *)
+        | `Lifted _ -> wrapper_node
+          (* if dbg then dpf"interesting caller, keep caller context"; *)
```
michael-schwarz, line 99: Please do not add anymore commented out code than we already have :wink:

src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -0,0 +1,257 @@
...
+        (* if an interesting callee is called by an uninteresting caller, then we remember the callee context *)
+        | _         -> `Lifted (node_for_ctx ctx)
+          (* if dbg then dpf"uninteresting caller, keep callee context"; *)
```
michael-schwarz, line 102: Same here

src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -0,0 +1,257 @@
...
+    (* Keep (potentially higher) counter from callee and keep wrapper node from caller *)
+    let _, lnode = ctx.local in
+    (counter, lnode)
+
+  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc ((counter, _):D.t) (f_ask: Queries.ask) : D.t =
+    ctx.local
+
+  let add_unique_call ctx =
+    let counter, wrapper_node = ctx.local in
+    (* TODO: previously, unique count isn't by wrapper node (e.g. my_malloc_wrapper), but by wrapped node (e.g. malloc). Why, and is it safe to change? *)
```
michael-schwarz, line 121: This is probably incidental and due to the student who made the implementation of the unique counter. Should be safe to change!

src/util/debug.ml
just-max: Oops, this is still temporary debugging code.

src/analyses/wrapperFunctionAnalysis.ml
just-max: The malloc analysis breaks :) You changed it from prev_node in the linked commit, but I can't figure out why the change was made.

just-max: We determined that the malloc analysis doesn't break (I still had an old test run in mind), so I'll change this to always use prev_node.

just-max:
A last note and implementation change I've made: `threadcreate` and `threadspawn` see different contexts: the former before `special` updates it (and increments the unique-counter), the latter after. The fix up until now didn't properly handle reaching `Top`. I've changed it so that the wrapper-function domain stores a pair of counter values `(previous, current)` for each call site; in `threadcreate` we get the current value, in `threadspawn` the previous value. Note that just adding/subtracting 1 to work around this isn't enough, since when the current counter is `Top`, the previous counter could either be `Top - 1` or `Top` with no way of knowing which.

This should be ready now. I've tested this PR with `gobopt='--set ana.activated[+] threadCreateWrapper --set ana.thread.unique_thread_id_count N' scripts/update_suite.rb` for N=0,1,8 with no problems (apart from the obvious failures from two of the new tests that expect N>=1). The new tests are in regression/71... instead of regression/70... anticipating that #850 will be merged first :) What's left: once there are some benchmarks for this, we could consider having it on be default.

Note that, like the malloc wrapper analysis, a given N actually allows distinguishing N+1 calls at the same call site: the first N calls are given unique values, then the next call is given `Top`, but these can still all be distinguished. That might be relevant, for example, if you "know" that your program has 8 threads (or the autotuner says so): then N can be set to 7.

src/util/gobList.ml
```diff
@@ -30,6 +30,13 @@ let rec fold_while_some (f : 'a -> 'b -> 'a option) (acc: 'a) (xs: 'b list): 'a
 
 let equal = List.eq
 
+(** [span p xs] is [take_while p xs, drop_while p xs] but may be more efficient *)
+let span p =
+  let rec span_helper prefix = function
+    | x :: xs when p x -> span_helper (x :: prefix) xs
+    | suffix -> List.rev prefix, suffix
+  in span_helper []
```
sim642, line 38: This already exists as `BatList.span`.

src/domains/printable.ml
```diff
@@ -470,7 +470,7 @@ module type ChainParams = sig
   val names: int -> string
 end
 
-module Chain (P: ChainParams): S with type t = int =
```
sim642, line 473: Why is this signature removal required?

src/cdomains/threadIdDomain.ml
```diff
@@ -55,14 +69,16 @@ struct
...
+  let threadenter l i v: t =
     if GobConfig.get_bool "ana.thread.include-node" then
-      (v, Some l)
+      (v, Some (l, i))
     else
       (v, None)
 
   let is_main = function
-    | ({vname = "main"; _}, None) -> true
+    (* shouldn't this check configured mainfun?? *)
```
sim642, line 80:
```suggestion
```
Since this check was changed.

src/cdomains/threadIdDomain.ml
```diff
@@ -30,21 +31,34 @@ sig
...
-  include Printable.Prod (CilType.Varinfo) (Printable.Option (Node) (struct let name = "no node" end))
+  include
+    Printable.Prod
+      (CilType.Varinfo) (
+      Printable.Option (
+        Printable.Prod
+          (Node) (
+          Printable.Option
+            (Printable.Int)
+            (struct let name = "no index" end)))
```
sim642, line 54: MallocWrapper used `Chain` for this, why doesn't that work here?

src/analyses/threadId.ml
```diff
@@ -87,13 +87,25 @@ struct
...
+    (* [ctx] here is the same as in [special], i.e. before incrementing the unique-counter,
+       thus we want the current counter (previous: false) *)
+    let+ tid = create_tid ctx.local (indexed_node_for_ctx ctx) f in
     (tid, TD.bot ())
 
   let threadspawn ctx lval f args fctx =
     let (current, td) = ctx.local in
-    (current, Thread.threadspawn td ctx.prev_node f)
+    (* here we see the updated counter, so we want the previous counter value *)
+    let node, index = indexed_node_for_ctx ~previous:true ctx in
```
sim642, line 107: Just do the query in `fctx` instead of `ctx`. Then it's querying the counter from the created thread, not the creator thread.

src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -0,0 +1,230 @@
...
+(* Functor argument for determining wrapper and wrapped functions *)
+module type WrapperArgs = sig
+  val wrappers : unit -> string list
+  val is_wrapped : LibraryDesc.special -> bool
+end
+
+(* The main analysis, generic to which functions are being wrapped. *)
+module SpecBase (UniqueCountArgs : UniqueCountArgs) (WrapperArgs : WrapperArgs) =
+struct
+  include Analyses.DefaultSpec
```
sim642, line 26: Might as well switch to `IdentitySpec` here to avoid all the identity transfer functions.

src/cdomains/threadIdDomain.ml
just-max: The problem is that `Chain` is a functor, returning a different module for each `n`. For the query we need a single module, hence this `Int` lattice.

Now that you mention it though, it might be possible to put the thread wrapper chain in a seperate module and reference it in both the thread wrapper analysis and in queries, which would clean things up.

src/domains/printable.ml
just-max: Oops, it's not :) I had added signatures all over the place while tinkering, then removed them again, and I guess also accidentally removed this.

michael-schwarz:
@just-max Do you still plan to address the comments here, or should we do it ourselves?

just-max:
Thanks for the reminder on this ;) I've addressed all the comments (thanks Simmo!)

There's one last problem I can't figure out: if the `unique_thread_id_count` is too small (e.g. set to `0` when there are two threads), Goblint should not be able to distinguish threads created at the same location. After implementing Simmo's suggestion to use `fctx` though, this isn't the case. See the test `71/05`, which specifically checks that a race does occur if `unique_thread_id_count` is set to 0. This test passes before implementing the change, but fails after, and I can't figure out why Goblint considers the two threads in that test case to be distinct.

sim642:
> See the test `71/05`, which specifically checks that a race does occur if `unique_thread_id_count` is set to 0. This test passes before implementing the change, but fails after, and I can't figure out why Goblint considers the two threads in that test case to be distinct.

When run with `--enable allglobs` to also see the non-racing accesses, it shows that the created threads are considered unique and thus joining "works". So Goblint actually only sees one unique thread being created (from the one location), not two. I didn't look into why that's the case though.

just-max:
In the wrapper function analysis' `threadenter`, the new thread receives a fresh view of the current wrapper node and counter. Then, when `threadspawn` goes to use `fctx` to add to the set of created threads, it adds the unwrapped node to the set of created threads.

https://github.com/goblint/analyzer/blob/68b22676ada38dda0aa61df7edc740118d7f283e/src/analyses/wrapperFunctionAnalysis.ml#L85-L87

https://github.com/goblint/analyzer/blob/68b22676ada38dda0aa61df7edc740118d7f283e/src/analyses/threadId.ml#L100-L103

I've changed `threadenter` in the wrapper function analysis to pass `ctx.local` instead of `bot`, but I'm not sure that's really the right approach.

sim642:
> I've changed `threadenter` in the wrapper function analysis to pass `ctx.local` instead of `bot`, but I'm not sure that's really the right approach.

That doesn't seem right indeed because the idea has been that each thread is handled on its own, independent of who created it.

Now looking at it, why is `ThreadId.threadspawn` using `indexed_node_for_ctx` at all? `ThreadId.threadenter` already computed the created thread's correct ID before the wrapper analysis' `threadenter` sets it to `bot`. Couldn't `ThreadId.threadspawn` just use the already computed thread ID from `fctx.local`?

just-max:
The spawned thread's ID from `fctx.local` in `ThreadId.threadspawn` is a stateful thread ID (prefix-set if history is enabled), whereas the set of created threads contains just the thread creation nodes. So there's no way to directly add a node to the set given the abstract thread ID.

I've changed the `ThreadId` domain to also include the (wrapper) node that created the thread, so that it is available in `fctx` in addition to the abstract thread ID.

<details>
<summary>Alternative Idea</summary>

https://github.com/goblint/analyzer/compare/master...just-max:goblint-analyzer:thread_wrappers_stateful_id

Another option is: change `Stateful` so that it takes an abstract thread ID instead of a just a node. The prefix-set implementation uses the top element of the prefix stack for unique threads, and all the nodes in the suffix set for non-unique threads, to update the set of thread creation nodes. I'm not convinced this is a good idea though: for non-unique threads, this is quite an over-approximation.

</details>

just-max:
Conflicts are resolved. Thanks again for the help getting this done!

sim642 (approved):

PR #1022, opened 2023-03-28
Split `combine` into `combine_env` and `combine_assign`

### LINKED ISSUE #1009: Relational analysis (and potentially others) unsound due to eval on pre-state in function calls
Consider the following program

~~~C
// SKIP PARAM: --set ana.activated[+] apron --set ana.base.privatization none --set ana.relation.privatization top
#include <goblint.h>

int g = 0;
int h = 0;
int *ptr = &g;

int change_ptr_and_return_5() {
    ptr = &h;
    return 5;
}

int main() {
    // ptr points to h here, but apron evaluates on pre-state of function call, updating g instead
    *ptr = change_ptr_and_return_5();

    int x = g + 1; // Currently this fails because of operation on bot
    assert(g == h); //FAIL
    assert(h == 5);
    assert(g == 0);
}
~~~

Here, the value of `ptr` is changed inside the function, which happens before the assignment. However, we currently use the pre-state of base to resolve the pointer information. This is unsound.

In the example, this leads to:
- our relational analysis wrongly claiming that `g=5` and `h=0` and
- base correctly claiming `h=5` and `g=0`. 

This inconsistent state can lead to unsound analysis results & crashes in places such as assignments where the analyses aim to benefit from each other and then find the value of a variable to be `\bot`.

sim642:
Closes #1009.

### TODO
- [x] Check (and fix) witness lifter.
- [x] Rethink `IdentitySpec` defaults.
- [x] Document `combine_env` and `combine_assign`.

### DIFF
--- a/src/analyses/accessAnalysis.ml
+++ b/src/analyses/accessAnalysis.ml
@@ -105,7 +105,10 @@ struct
   let enter ctx lv f args : (D.t * D.t) list =
     [(ctx.local,ctx.local)]
 
-  let combine ctx lv fexp f args fc al f_ask =
+  let combine_env ctx lval fexp f args fc au f_ask =
+    ctx.local
+
+  let combine_assign ctx lv fexp f args fc al f_ask =
     access_one_top ctx Read false fexp;
     begin match lv with
       | None      -> ()
--- a/src/analyses/accessAnalysis.ml
+++ b/src/analyses/accessAnalysis.ml
@@ -106,16 +106,17 @@ struct
     [(ctx.local,ctx.local)]
 
   let combine_env ctx lval fexp f args fc au f_ask =
-    ctx.local
+    (* These should be in enter, but enter cannot emit events, nor has fexp argument *)
+    access_one_top ctx Read false fexp;
+    List.iter (access_one_top ctx Read false) args;
+    au
 
   let combine_assign ctx lv fexp f args fc al f_ask =
-    access_one_top ctx Read false fexp;
     begin match lv with
       | None      -> ()
       | Some lval -> access_one_top ~deref:true ctx Write false (AddrOf lval)
     end;
-    List.iter (access_one_top ctx Read false) args;
-    al
+    ctx.local
 
 
   let threadspawn ctx lval f args fctx =
--- a/src/analyses/expsplit.ml
+++ b/src/analyses/expsplit.ml
@@ -46,7 +46,10 @@ struct
   let return ctx (exp:exp option) (f:fundec) =
     emit_splits_ctx ctx
 
-  let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc au (f_ask: Queries.ask) =
+  let combine_env ctx lval fexp f args fc au f_ask =
+    ctx.local
+
+  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc au (f_ask: Queries.ask) =
     let d = D.join ctx.local au in
     emit_splits ctx d
 
--- a/src/analyses/expsplit.ml
+++ b/src/analyses/expsplit.ml
@@ -47,11 +47,11 @@ struct
     emit_splits_ctx ctx
 
   let combine_env ctx lval fexp f args fc au f_ask =
-    ctx.local
-
-  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc au (f_ask: Queries.ask) =
     let d = D.join ctx.local au in
     emit_splits ctx d
+
+  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc au (f_ask: Queries.ask) =
+    emit_splits_ctx ctx
 
   let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) =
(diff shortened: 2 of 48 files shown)


src/analyses/accessAnalysis.ml
```diff
@@ -105,7 +105,10 @@ struct
   let enter ctx lv f args : (D.t * D.t) list =
     [(ctx.local,ctx.local)]
 
-  let combine ctx lv fexp f args fc al f_ask =
+  let combine_env ctx lval fexp f args fc au f_ask =
+    ctx.local
+
+  let combine_assign ctx lv fexp f args fc al f_ask =
     access_one_top ctx Read false fexp;
```
michael-schwarz, line 112: Is this correct if `fexp` is modified inside `f`? I would think this needs to happen before? Potentially even in enter?

src/analyses/expsplit.ml
```diff
@@ -46,10 +46,13 @@ struct
   let return ctx (exp:exp option) (f:fundec) =
     emit_splits_ctx ctx
 
-  let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc au (f_ask: Queries.ask) =
+  let combine_env ctx lval fexp f args fc au f_ask =
     let d = D.join ctx.local au in
     emit_splits ctx d
 
+  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc au (f_ask: Queries.ask) =
+    emit_splits_ctx ctx
```
michael-schwarz, line 54: Is it necessary to do `emit_splits` twice here?

src/framework/constraints.ml
```diff
@@ -441,7 +458,7 @@ struct
 
   let paths_as_set ctx =
     let liftmap = List.map (fun x -> D.lift x) in
-    lift_fun ctx liftmap S.paths_as_set (Fun.id) []
+    lift_fun ctx liftmap S.paths_as_set (Fun.id) [D.bot ()]
```
michael-schwarz, line 461: Why is this change of the default value needed here?

michael-schwarz:
LGTM! Maybe we should do an sv-comp run to see if there are any new issues? Might be nice to uncover them now when this change is fresh on our minds...

src/analyses/accessAnalysis.ml
sim642: It's only read. I thought about moving these to `enter` but it doesn't have the original function expression argument, only the resolved name. It could be changed but I don't want to change every `enter` here as well. It's been fine since forever.

sim642:
> LGTM! Maybe we should do an sv-comp run to see if there are any new issues? Might be nice to uncover them now when this change is fresh on our minds...

I think I started some run. I'll get the results before merging.

src/analyses/accessAnalysis.ml
michael-schwarz: Still, the access should be in `combine_env` right?

sim642:
Verdicts-wise there's no notable difference. CPU-time–wise it's also an identity line.

Surprisingly, this somehow reduces memory usage on some sv-benchmarks:
image
I don't have an explanation for this. Maybe it being two-staged like this allows some garbage to be collected earlier after `combine_env`?

src/framework/constraints.ml
sim642: Without it many of the longjmp tests fail because `combine_env` of the entire `Spec` is never called if the function's normal return node is dead (and thus the list is empty). But it needs to be delegated because there could be a longjmpy return that's not dead.

src/analyses/expsplit.ml
sim642: I wondered it too and thought there could be some corner case where it matters, but didn't spend time trying to construct it.
The idea is that after `combine_env` there could be splits over global variables, which need to be updated/preserved, and after `combine_assign` there could be a split over the assigned variable, which needs to be updated/preserved. Maybe when the assignment is to a global variable which was also modified inside the function (but its new value comes from the return value)?

src/analyses/accessAnalysis.ml
sim642: I'll move it there to be slightly closer to where it should be, although it wouldn't still be fully precise.

src/witness/observerAnalysis.ml
```diff
@@ -65,7 +65,10 @@ struct
     (* ctx.local doesn't matter here? *)
     [ctx.local, step ctx.local ctx.prev_node (FunctionEntry f)]
 
-  let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask: Queries.ask) : D.t =
+  let combine_env ctx lval fexp f args fc au f_ask =
+    ctx.local
```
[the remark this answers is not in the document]
sim642: The observer analysis is consistent with the witness lifter such that both consider the function call edge to be fully completed after the assign has also taken place since the assign is part of that edge.

src/analyses/expsplit.ml
michael-schwarz: I think as long as we are convinced that it might make sense to have to do it twice, it doesn't hurt to do it twice. Probably not worth investing too much time in an example where it matters.

michael-schwarz (approved):

PR #1023, opened 2023-03-28
Fix longjmp marshaling tests, clean up relifting

sim642:
Besides just fixing the necessary, this does broader cleanup with the hope to avoid similar issues again.

### Changes
1. Remove identity `Printable.Std.relift` which hides missing definitions.
2. Add all missing `relift` definitions. Some rather core types like `MapDomain` were not relifting their inner components.
3. Introduce `Printable.StdLeaf` as `Printable.Std` alternative with identity `relift`. This is only to be used with primitive and CIL types which don't contain inner types that could require relifting.
4. Remove default `Printable.Std.name`, which names everything "std". This makes debugging such issues a total pain, e.g. "lifted std and either std or std option" is completely ineligible. All printables must specify a name explicitly.
5. Add all missing `name` definitions. 

### TODO
- [x] Readd `relift`s for all other marshaling tests.

### DIFF
--- a/src/analyses/mCPRegistry.ml
+++ b/src/analyses/mCPRegistry.ml
@@ -184,9 +184,7 @@ struct
     let arbs = map (fun (n, (module D: Printable.S)) -> QCheck.map ~rev:(fun (_, o) -> obj o) (fun x -> (n, repr x)) @@ D.arbitrary ()) @@ domain_list () in
     MyCheck.Arbitrary.sequence arbs
 
-  let relift = unop_map (fun (module S: Printable.S) x ->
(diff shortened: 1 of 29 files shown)


michael-schwarz (approved):

michael-schwarz:
Might be worth doing some of the incremental benchmarks here, if this is very expensive, it might be worth taking steps to avoid relifting where it is not needed due to the current configuration.

michael-schwarz:
Also we might think about writing a deriver for `relift`? For most cases I don't see a reason we cannot do it.

PR #1024, opened 2023-03-29
Fix 00-sanity/35-join-contexts by syncing at function entry

sim642:
This is code-wise a tiny fix, but relatively important, so it's good to have a PR rather than quietly committing to master.

Side effects to function entry nodes are joined by the solver, but simply using joined values is unsound as the example also demonstrated. This inserts a `sync Join` also at the entry node, to be performed before the `body` transfer function is called. This way, whatever is joined is "made sound" for subsequent use.

This happens during solving in the `body` transfer function evaluation, so all new side effects from the sync are correctly handled for. A similar fix wouldn't work for join-over-all-contexts after the analysis, since new side effects may (and do) arise.

### DIFF
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -523,9 +523,11 @@ struct
      2. fundec -> set of S.C  --  used for IterSysVars Node *)
 
   let sync ctx =
-    match Cfg.prev ctx.prev_node with
-    | _ :: _ :: _ -> S.sync ctx `Join
-    | _ -> S.sync ctx `Normal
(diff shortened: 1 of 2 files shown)


michael-schwarz:
I am still not a huge fan of these sync edges as we have no theoretical underpinnings for them, but I guess we must live with them if we do not want to force certain analyses to always be context-sensitive.

sim642:
> I am still not a huge fan of these sync edges as we have no theoretical underpinnings for them

The ASE16 paper has these syncs wrapped around all transfer functions actually. It doesn't describe anything interprocedural though, so essentially there is just one context.

PR #1030, opened 2023-04-03
Refactor to Remove Dependency of `ValueDomain` on `Queries`

jerhard:
#### Changes

This PR refactors the `ValueDomain` such that its compilation no longer depends on  `Queries`. To this goal, the subset of queries that were previously used by the `ValueDomain` (and by the `ArrayDomain` in particular), are now passed as a record to the `ValueDomain` functions. 

#### Benefit
Removing this dependency has the benefit of allowing use of the `ValueDomain` in `Queries`, which would otherwise not be possible as it would introduce a cyclic dependency. One instance where queries passing `ValueDomain.t` values are needed is the implementation of the modular analysis, where values written by functions should be communicated.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -465,7 +465,7 @@ struct
       let f_addr (x, offs) =
         (* get hold of the variable value, either from local or global state *)
         let var = get_var a gs st x in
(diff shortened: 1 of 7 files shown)


sim642:
I'm doing sv-benchmarks comparisons runs just in case the allocation of all these extra `VDQ.t` records has a measurable impact.

sim642 (approved):

PR #1036, opened 2023-04-19
Fix missing dead code warning from strictly dead loop body

sim642:
Fixes the case identified in https://github.com/goblint/analyzer/pull/979#discussion_r1171079392.

If the loop entry edge is already dead, the loop's back edge was never even queried from the solver. Thus it was completely missing from the solver solution (not with dead code lifter's bottom) and no dead code warning was emitted.

This makes the constraint system query all of them, but still ignore the non-strict ones if the entry edge is dead. This way dead code lifter's bottom is propagated through the loop.

Using a cram test because the presence of dead code warnings cannot be checked with update_suite.rb.

### DIFF
--- a/tests/regression/00-sanity/36-strict-loop-dead.c
+++ b/tests/regression/00-sanity/36-strict-loop-dead.c
@@ -6,8 +6,8 @@ int basic2(int n) {
     return 0;
 
   for (int i = 0; i < n; i++) // NOWARN
-    // Bug in dead code warnings: no dead code warning is emitted, because body is not included
-    // in the result. Transformation checks all CFG nodes, and therefore works.
+    // Was bug in dead code warnings: no dead code warning was emitted, because body was not included
+    // in the result. Transformation checks all CFG nodes, and therefore worked.
     a += i + n; // NOWARN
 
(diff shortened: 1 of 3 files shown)


tests/regression/00-sanity/36-strict-loop-dead.c
```diff
@@ -0,0 +1,19 @@
...
+int basic2(int n) {
+  int a = 0;
+
+  if (n < 0)
+    return 0;
+
+  for (int i = 0; i < n; i++) // NOWARN
+    // Bug in dead code warnings: no dead code warning is emitted, because body is not included
+    // in the result. Transformation checks all CFG nodes, and therefore works.
+    a += i + n; // NOWARN
```
michael-schwarz, line 11: Should this description be in the present? Or in the past now as this is fixed?

michael-schwarz (approved):

tests/regression/00-sanity/36-strict-loop-dead.c
sim642: It's fixed, so I reworded it.

PR #1037, opened 2023-04-20
Add automated test for GobView beyond building

stilscher:
This PR adds an automated python test that serves the GobView page and checks that the overall structure of the page is correct, i.e., that the DOM elements for the navbar, sidebars, bottom panel and content view are displayed. Although this is not a very detailed check, it should be sufficient to detect errors during the initialization phase that can cause nothing to be displayed at all even though the build was successful. 
In the ubuntu-latest image for the Github actions, pip3 and chrome should already be preinstalled (https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md).

### DIFF
--- a/.github/workflows/locked.yml
+++ b/.github/workflows/locked.yml
@@ -165,3 +165,9 @@ jobs:
 
       - name: Build Gobview
         run: opam exec -- dune build gobview
+
+      - name: Test Gobview
+        run: |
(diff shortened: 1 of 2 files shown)


michael-schwarz:
It is great to see progress in this direction, it has been very annoying that there were no checks other than whether it still compiles!

PR #1041, opened 2023-05-07
Analysis tracking equality between calls to math library functions to improve invariant

FelixKrayer:
I was part of last year's practical course implementing the interval domain for floats. One issue we were unfortunately unable to solve was using certain math library functions in the invariant. Consider the following piece of code:

~~~C
if (isfinite(f)) {
    __goblint_check(! isnan(f)); 
}
~~~

In this case the check would result in unknown, because the code is transformed into the following where the call to the library function cannot be detected at the if-check. Thus `f` cannot be refined to be finite in this example:

~~~C
tmp__0 = isfinite(f);
if (tmp__0) {
    tmp__1 = isnan(f);
    __goblint_check(! tmp__1);
}
~~~

This pull request adds a new analysis `TmpSpecial` tracking equalities between lvals and calls to library functions together with the corresponding arguments. 
Additionally a set of lval is tracked for each equality. This describes the lvals the equality depends on, i.e., lvals contained within the arguments as well as the lval of the assignment. This set is used to delete equalities when an lval it depends on is written.
The information this analysis collects is accessed in the base Invariant via a query to refine arguments of a call. Some examples of this refinement can be seen in regtest 57/19.

TODO:

- [x] better inversion of `ceil` and `floor`


michael-schwarz:
I started the run with this analysis enabled just now, let's see what the results are!

michael-schwarz:
Something to consider: What is the relationship of this with the `varEq` analysis we already have? Can there be some reuse?

src/analyses/libraryDesc.ml
```diff
@@ -143,3 +143,184 @@ let of_old ?(attrs: attr list=[]) (old_accesses: Accesses.old) (classify_name):
...
+  | Fabs (fk, exp) -> Fabs (CilType.Fkind.relift fk, CilType.Exp.relift exp)
+  | Fmax (fk, exp1, exp2) -> Fmax (CilType.Fkind.relift fk, CilType.Exp.relift exp1, CilType.Exp.relift exp2)
+  | Fmin (fk, exp1, exp2) -> Fmin (fk, CilType.Exp.relift exp1, CilType.Exp.relift exp2)
+  | Acos (fk, exp) -> Acos (CilType.Fkind.relift fk, CilType.Exp.relift exp)
+  | Asin (fk, exp) -> Asin (CilType.Fkind.relift fk, CilType.Exp.relift exp)
+  | Atan (fk, exp) -> Atan (CilType.Fkind.relift fk, CilType.Exp.relift exp)
+  | Atan2 (fk, exp1, exp2) -> Atan2 (CilType.Fkind.relift fk, CilType.Exp.relift exp1, CilType.Exp.relift exp2)
+  | Cos (fk, exp) -> Cos (CilType.Fkind.relift fk, CilType.Exp.relift exp)
+  | Sin (fk, exp) -> Sin (CilType.Fkind.relift fk, CilType.Exp.relift exp)
+  | Tan (fk, exp) -> Tan (CilType.Fkind.relift fk, CilType.Exp.relift exp)
```
michael-schwarz, line 178: As there are no hash-consed things inside this, relifting is not really needed here.

src/analyses/libraryDesc.ml
```diff
@@ -143,3 +143,184 @@ let of_old ?(attrs: attr list=[]) (old_accesses: Accesses.old) (classify_name):
...
+      | Fmax fee1, Fmax fee2 -> cmp_fk_exp_exp fee1 fee2
+      | Fmin fee1, Fmin fee2 -> cmp_fk_exp_exp fee1 fee2
+      | Acos fe1, Acos fe2 -> cmp_fk_exp fe1 fe2
+      | Asin fe1, Asin fe2 -> cmp_fk_exp fe1 fe2
+      | Atan fe1, Atan fe2 -> cmp_fk_exp fe1 fe2
+      | Atan2 fee1, Atan2 fee2 -> cmp_fk_exp_exp fee1 fee2
+      | Cos fe1, Cos fe2 -> cmp_fk_exp fe1 fe2
+      | Sin fe1, Sin fe2 -> cmp_fk_exp fe1 fe2
+      | Tan fe1, Tan fe2 -> cmp_fk_exp fe1 fe2
+      | _ -> failwith "impossible"
```
michael-schwarz, line 263: Could we just derive these here?

src/analyses/baseInvariant.ml
```diff
@@ -691,7 +697,30 @@ struct
...
             in
-            update_lval c x c' ID.pretty
+            let st = update_lval c x c' ID.pretty in
+            (* handle special calls *)
+            begin match t with
+            | TInt (ik, _) ->
+              begin match x with
+                | ((Var v), offs) -> 
+                  if M.tracing then M.trace "invSpecial" "qry Result: %a\n" Queries.ML.pretty (ctx.ask (Queries.TmpSpecial (v, TmpSpecial.Spec.resolve offs))); 
+                  let tv = not (ID.leq c (ID.of_bool ik false)) in
```
michael-schwarz, line 707: Is this correct? Consider, e.g. `c = [0,1]`?
$tv = \neg ([0,1] \not\sqsubseteq [0,0]) \implies tv = \textsf{true}$

src/analyses/baseInvariant.ml
```diff
@@ -691,7 +697,30 @@ struct
...
+                  | `Lifted (Isnan xFloat) when tv -> inv_exp (`Float (FD.nan_of (unroll_fk_of_exp xFloat))) xFloat st
+                  (* should be correct according to C99 standard*)
+                  | `Lifted (Isgreater (xFloat, yFloat)) -> inv_exp (`Int (ID.of_bool ik tv)) (BinOp (Gt, xFloat, yFloat, (typeOf xFloat))) st
+                  | `Lifted (Isgreaterequal (xFloat, yFloat)) -> inv_exp (`Int (ID.of_bool ik tv)) (BinOp (Ge, xFloat, yFloat, (typeOf xFloat))) st
+                  | `Lifted (Isless (xFloat, yFloat)) -> inv_exp (`Int (ID.of_bool ik tv)) (BinOp (Lt, xFloat, yFloat, (typeOf xFloat))) st
+                  | `Lifted (Islessequal (xFloat, yFloat)) -> inv_exp (`Int (ID.of_bool ik tv)) (BinOp (Le, xFloat, yFloat, (typeOf xFloat))) st
+                  | `Lifted (Islessgreater (xFloat, yFloat)) -> inv_exp (`Int (ID.of_bool ik tv)) (BinOp (LOr, (BinOp (Lt, xFloat, yFloat, (typeOf xFloat))), (BinOp (Gt, xFloat, yFloat, (typeOf xFloat))), (TInt (IBool, [])))) st
+                  | `Lifted (Isunordered (xFloat, yFloat)) -> st (* something can probably be done here *)
```
michael-schwarz, line 717: Probably, yes. However, it seems non-obvious what to do here...

src/analyses/baseInvariant.ml
```diff
@@ -701,7 +730,28 @@ struct
...
-            update_lval c x c' FD.pretty
+            let st = update_lval c x c' FD.pretty in
+            (* handle special calls *)
+            begin match t with
+            | TFloat (fk, _) ->
+              begin match x with
+                | ((Var v), offs) -> 
+                  if M.tracing then M.trace "invSpecial" "qry Result: %a\n" Queries.ML.pretty (ctx.ask (Queries.TmpSpecial (v, TmpSpecial.Spec.resolve offs))); 
+                  begin match ctx.ask (Queries.TmpSpecial (v, TmpSpecial.Spec.resolve offs)) with
+                  | `Lifted (Ceil (ret_fk, xFloat)) when FD.is_interval c -> inv_exp (`Float (FD.inv_ceil (FD.cast_to ret_fk c))) xFloat st
```
michael-schwarz, line 741: Why is the check if it is an Interval needed here?

src/domains/queries.ml
```diff
@@ -274,6 +279,7 @@ struct
     | Any ActiveJumpBuf -> 46
     | Any ValidLongJmp -> 47
     | Any (MayBeModifiedSinceSetjmp _) -> 48
+    | Any (TmpSpecial _) -> 42
```
michael-schwarz, line 282: Probably my mistake when merging...

```suggestion
    | Any (TmpSpecial _) -> 49
```

src/analyses/tmpSpecial.ml
```diff
@@ -0,0 +1,151 @@
...
+  let rec ls_of_exp ctx (exp:exp) : LS.t =
+  match exp with
+  | Const _ -> LS.bot ()
+  | Lval lv -> ls_of_lv ctx lv
+  | SizeOf _ -> LS.bot ()
+  | Real e -> ls_of_exp ctx e
+  | Imag e -> ls_of_exp ctx e
+  | SizeOfE e -> ls_of_exp ctx e
+  | SizeOfStr _ -> LS.empty ()
+  | AlignOf _ -> LS.top () (* TODO: what is this*)
```
michael-schwarz, line 38: This is an Align of type, so this should be save
```suggestion
  | AlignOf _ -> LS.empty ()
```

src/analyses/tmpSpecial.ml
```diff
@@ -0,0 +1,151 @@
...
+  | AlignOfE _ -> LS.top () (* TODO: what is this*)
```
michael-schwarz, line 39:
```suggestion
  | AlignOfE e -> ls_of_exp ctx e
```

Not sure if this is needed, but it would be a safe default.

src/analyses/tmpSpecial.ml
```diff
@@ -0,0 +1,151 @@
...
+  let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) f_ask : D.t =
+    D.bot ()
+
+  let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    let d = ctx.local in
+
+    (* Just dbg prints *)
+    (match lval with
+    | Some lv -> if M.tracing then M.tracel "tmpSpecial" "Special: %s with lval %a\n" f.vname d_lval lv
+    | _ -> if M.tracing then M.tracel "tmpSpecial" "Special: %s\n" f.vname);
```
michael-schwarz, line 79: Maybe pull the if `M.tracing` out?

src/analyses/tmpSpecial.ml
```diff
@@ -0,0 +1,151 @@
...
+    ctx.local
+
+  let return ctx (exp:exp option) (f:fundec) : D.t =
+    ctx.local
+
+  let enter ctx (lval: lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
+    [ctx.local, D.bot ()]
+
+  let combine ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) f_ask : D.t =
+    D.bot ()
```
michael-schwarz, line 71: Maybe make a comment explaining that we forget all relationships at the call?

sim642:
> Something to consider: What is the relationship of this with the `varEq` analysis we already have? Can there be some reuse?

`var_eq` tracks equalities between CIL expressions, but function calls are not part of those, so I don't think there's much to be done. If anything, then this is more similar to `condvars`, which I think exists for a similar purpose, but with logical expressions (transformed by CIL) instead of special function calls.

This could be combined with `var_eq` in another (not reuse) way. That is, to track the return value of a special function not just through a single assignment to temporary just before a branch, but through a longer chain of manually written assignments. But that would just involve some extra queries to `var_eq` I suppose.

src/analyses/libraryDesc.ml
FelixKrayer: It is possible but only if the type `math` in `LibraryDesc` is changed to use `CilType.Fkind.t` and `Basetype.CilExp.t` instead of `Cil.fkind` and `Cil.exp`.
If we change this, should we change the whole LibraryDesc to use these types instead of the `Cil.` types for consistency?

src/analyses/tmpSpecial.ml
FelixKrayer: The analysis can surely be expanded to track relationships interprocedurally. I will look into that (in this PR or in a following one), however, the seemingly most important case, where the result of a function call is put into a tmp variable is covered by the current implementation.

src/analyses/baseInvariant.ml
FelixKrayer: It was needed in a previous implementation. However, now this is handled correctly in the float domain, so the whole `is_interval` is no longer needed and I removed it

src/analyses/baseInvariant.ml
FelixKrayer: This was a mistake on my side. Now it is handled correctly with `ID.to_bool`

src/analyses/libraryDesc.ml
sim642: `CilType.Fkind.t` and `CilType.Exp.t` are simply aliases for `Cil.fkind` and `Cil.exp` but come with all the necessary `Printable.S` functions. This is precisely to allow deriving from them. So changing to `CilType` is the way to go here.

FelixKrayer: Ok, thanks for the clarification, I changed it

src/analyses/tmpSpecial.ml
michael-schwarz: Sure, I'm not sure if one even wants to extend it that way, but if we don't we should make this comment here, so people understand it is `D.bot ()` for a good reason!

michael-schwarz:
I did a run on sv-comp but it seems that the analysis sadly does not really make a difference :disappointed: 

[tmpSpecial.zip](https://github.com/goblint/analyzer/files/11522103/tmpSpecial.zip)

michael-schwarz:
I started another run after the fixes now, let's see if that changes something!

sim642:
> I did a run on sv-comp but it seems that the analysis sadly does not really make a difference disappointed

It might be worth checking with grep if sv-benchmarks even contains such pattern in the first place, or whether it's something that only exists in real-world programs. I didn't look deeply at the implementation, but sv-benchmarks contains many cilled programs, where such transformation could already be hard-coded instead of being introduced by our CIL. Maybe that affects the applicability of this analysis?

FelixKrayer:
> > I did a run on sv-comp but it seems that the analysis sadly does not really make a difference disappointed
> 
> It might be worth checking with grep if sv-benchmarks even contains such pattern in the first place, or whether it's something that only exists in real-world programs. I didn't look deeply at the implementation, but sv-benchmarks contains many cilled programs, where such transformation could already be hard-coded instead of being introduced by our CIL. Maybe that affects the applicability of this analysis?

I did a quick search through the sv-comp repository and did only find 1456 files that include `math.h`. From what I could find, most of them do not use any interesting function (fabs, isfinite, isnan) in a way where this analysis is helpful.
Furthermore a search for `fabs` (380 files), `isfinite` (28 files) and `isnan` (150 files) only resulted in me finding many "custom" implementations of these functions and only a few uses of the library functions. Unfortunately none of these were in cases, where this analysis could provide an improvement (mostly the functions were just used in checks, but then there was no way, a refinement could be beneficial).

As for the transformations already being hardcoded, this should not be a problem. The analysis tracks the equality between a variable and the math descriptor and it does not matter, if the transformation was introduced by CIL or if it was already hardcoded in the code.

src/analyses/tmpSpecial.ml
```diff
@@ -0,0 +1,155 @@
...
+open GoblintCil
+open Analyses
+
+module Spec =
+struct
+  include Analyses.IdentitySpec
+
+  let name () = "tmpSpecial"
+  module ML = LibraryDesc.MathLifted
+  module LS = SetDomain.ToppedSet(Lval.CilLval) (struct let topname = "All" end)
```
sim642, line 13: Could just use `Queries.LS` like everything else. That avoids the need for yet another lvalue conversion function (`resolve`). Queries give `Queries.LS`, so just using that avoids need for such conversions altogether.

sim642 (changes requested):

src/analyses/libraryDesc.ml
```diff
@@ -152,3 +152,49 @@ let of_old ?(attrs: attr list=[]) (old_accesses: Accesses.old) (classify_name):
...
+    | Cos (fk, exp) -> Pretty.dprintf "(%a )cos(%a)" d_fkind fk d_exp exp
+    | Sin (fk, exp) -> Pretty.dprintf "(%a )sin(%a)" d_fkind fk d_exp exp
+    | Tan (fk, exp) -> Pretty.dprintf "(%a )tan(%a)" d_fkind fk d_exp exp
+
+    include Printable.SimplePretty (
+      struct
+        type nonrec t = t
+        let pretty = pretty
+      end
+      )
```
sim642, line 194:
```suggestion
  include Printable.SimplePretty (
    struct
      type nonrec t = t
      let pretty = pretty
    end
    )
```

src/analyses/tmpSpecial.ml
```diff
@@ -0,0 +1,105 @@
...
+    invalidate (Analyses.ask_of_ctx ctx) (mkAddrOf lval) ctx.local
+
+  let branch ctx (exp:exp) (tv:bool) : D.t =
+    ctx.local
+
+  let body ctx (f:fundec) : D.t =
+    ctx.local
+
+  let return ctx (exp:exp option) (f:fundec) : D.t =
+    ctx.local
```
sim642, line 38: These should be unnecessary with `IdentitySpec`.
```suggestion
```

sim642 (approved):

FelixKrayer:
Awesome! Glad to see it merged after I now finally finished it up.

PR #1042, opened 2023-05-08
Replace `should_join` with path representative

### LINKED ISSUE #1039: `PathSensitive2`: Avoid quadratic complexity for `map` by grouping/hashing
### Issue:
The `map` operation in the `PathSensitive2` functor performs a quadratic number of `should_join` calls in the number of paths. In a modified version of regression test `08/02` with 9 malloced variables, where the `malloc_null` analysis is activated, there a considerable amount of time is spent in `Constraints.PathSensitive2.map` and in particlar in `BatSet.add` operations:
Bildschirmfoto vom 2023-05-05 15-42-30

In case of 9 malloced variables, there are 2^9 paths, leading to a number of comparisons in the order of magnitude of (2^9)^2. 

### Solution:
The idea is to avoid the comparison of all paths to build the set of paths. To that goal, one could demand that each analysis provides an additional hash function `hash'`, such that `should_join d1 d2` implies `hash' d1 = hash' d2`. Then, one can use the `hash'` to build buckets in which a mutual `should_join` comparison has to be performed.

As this approach would require to maintain a separate `hash'` function for each analysis, a modified solution (proposed by @michael-schwarz) may be preferable:
The `should_join` function would be modified to return the following type ` True | Eq | ShouldJoin of 'a -> 'a -> bool`. 
The `Eq` would mean that the should join is defined via the equality of the analysis states (`D.equal`), `True` would mean that the every pair of states should be joined, and the `ShouldJoin of 'a -> 'a -> bool` could be any custom should join function.

Then, building the set of paths in the `PathSensitive2.map` could rely on this information:
For analysis for which the `shouldjoin` returned `Eq`, the regular hash function of the analysis domain can be used to obtain  buckets, so that joins of paths only occur within buckets. For `ShouldJoin`, one may use a constant hash function.

As there are multiple analyses, the buckets should be split by the combination (e.g. the product) of those hashes.

sim642:
Closes #1039.

On 08/02 with 7 malloced variables, this reduces analysis time 4.8s → 1.5s.

### TODO
- [x] Implement `DisjointDomain.ProjectiveMap`.
- [x] Use path representatives in `PathSensitive3` for witness generation.
- [x] Compare on sv-benchmarks.
- [x] Check if witness automata are constructed correctly.

### DIFF
--- a/src/analyses/stackTrace.ml
+++ b/src/analyses/stackTrace.ml
@@ -4,12 +4,11 @@ open GoblintCil
 open Analyses
 module LF = LibraryFunctions
 
-module Spec (D: StackDomain.S) (P: sig val name : string end)=
+module Spec (D: StackDomain.S) (N: sig val name : string end)=
 struct
-  module ArgP = P
   include Analyses.IdentitySpec
 
-  let name () = ArgP.name
+  let name () = N.name
   module D = D
   module C = D
 
--- a/src/analyses/stackTrace.ml
+++ b/src/analyses/stackTrace.ml
@@ -6,9 +6,10 @@ module LF = LibraryFunctions
 
 module Spec (D: StackDomain.S) (P: sig val name : string end)=
 struct
+  module ArgP = P
   include Analyses.IdentitySpec
 
-  let name () = P.name
+  let name () = ArgP.name
   module D = D
   module C = D
 
--- a/src/analyses/unassumeAnalysis.ml
+++ b/src/analyses/unassumeAnalysis.ml
@@ -23,7 +23,6 @@ struct
   let exitstate _ = D.empty ()
 
   let context _ _ = ()
-  let should_join _ _ = false
 
   module Locator = WitnessUtil.Locator (Node)
 
--- a/src/analyses/activeSetjmp.ml
+++ b/src/analyses/activeSetjmp.ml
@@ -11,8 +11,7 @@ struct
 
   module D = JmpBufDomain.JmpBufSet
   module C = JmpBufDomain.JmpBufSet
-
-  let should_join a b = D.equal a b
+  module P = IdentityP (D)
 
   let combine_env ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask:Queries.ask): D.t =
     ctx.local (* keep local as opposed to IdentitySpec *)
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -21,6 +21,12 @@ struct
     include Priv.V
     include StdV
   end
+  module P =
+  struct
+    include Priv.P
+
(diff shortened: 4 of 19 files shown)


michael-schwarz:
> On 08/02 with 7 malloced variables, this reduces analysis time 4.8s → 1.5s.

Would also be interesting how it behaves, e.g. for 10 malloced variables. Then we would know whether it already is the explosion of paths that breaks our neck or whether this only happens in combination with the old quadratic operations on the set.

sim642:
> Would also be interesting how it behaves, e.g. for 10 malloced variables. Then we would know whether it already is the explosion of paths that breaks our neck or whether this only happens in combination with the old quadratic operations on the set.

I ran this test with various numbers of malloced variables before and after this PR and the results are as follows (log plot):
image
The raw data itself:
Mallocs | Before | After
-- | -- | --
2 | 0.06 | 0.08
3 | 0.095 | 0.141
4 | 0.21 | 0.23
5 | 0.47 | 0.43
6 | 1.27 | 0.9
7 | 4.18 | 1.8
8 | 16.65 | 3.8
9 | 70.9 | 8.1
10 | 321 | 19.6
11 | 1518 | 48.7

Obviously it's still exponential, but the improvement is also exponential and significant.

Actually the after trend line is extremely close to exponential, but the before one seems even superexponential. This is also visible in a custom log plot with quadratic trend lines:
image
I guess the reduction in the quadratic factor corresponds to the reduction of pairwise checking of an exponential number of paths here.

sim642:
On sv-benchmarks there's ~2% improvement inside the noise:
image
This is to be expected because there's a minimal amount of path-sensitivity there.

I still have to manually check that the witness automata are constructed correctly with the new `ProjectiveMap`.

src/analyses/stackTrace.ml
```diff
@@ -6,9 +6,10 @@ module LF = LibraryFunctions
 
 module Spec (D: StackDomain.S) (P: sig val name : string end)=
```
michael-schwarz, line 7: Why not rename the argument of the functor?

src/analyses/unassumeAnalysis.ml
```diff
@@ -23,7 +23,6 @@ struct
   let exitstate _ = D.empty ()
 
   let context _ _ = ()
-  let should_join _ _ = false
```
michael-schwarz, line 26: Do we not need some equivalent version of this?

michael-schwarz (approved):

src/analyses/unassumeAnalysis.ml
sim642: I couldn't figure out why it even was `false` to begin with. I don't recall ever using the unassume analysis in path sensitive mode. Also a constant `false` doesn't make sense I think because it isn't even reflexive.

michael-schwarz:
@stilscher do you want to still review this?

stilscher:
Yes, I will finish my review until Thursday at the latest.

PR #1046, opened 2023-05-12
Make privatizations sound in presence of analyses that can recover from multi-threaded mode

### LINKED ISSUE #1043: Privatizations unsound when enabling the `thread` analysis
While playing around with examples to discover what we want to do for phases, I discovered (some of) the privatizations are unsound after enabling the `thread` analysis from `threadAnalysis.ml`.

~~~C
// PARAM: --set ana.path_sens[+] threadflag --set ana.base.privatization mutex-meet-tid --enable ana.int.interval --set ana.activated[+] threadJoins --set ana.activated[+] thread
#include <pthread.h>
#include <goblint.h>

int g = 10;

void *t_benign(void *arg) {
  pthread_mutex_lock(&A);
  g = 10;
  __goblint_check(g == 10);
  pthread_mutex_unlock(&A);
  return NULL;
}

void *t_benign2(void *arg) {
  pthread_mutex_lock(&A);
  __goblint_check(g == 20);
  g = 10;
  __goblint_check(g == 10);
  pthread_mutex_unlock(&A);
  return NULL;
}

int main(void) {

  pthread_t id2;
  pthread_join(id2, NULL);


  g = 20;
  __goblint_check(g == 20);


  pthread_create(&id2, NULL, t_benign2, NULL);


  pthread_mutex_lock(&A);
  __goblint_check(g == 20); //UNKNOWN!
  pthread_mutex_unlock(&A);

  return 0;
}
~~~

Here, the analysis claims to know that the value of `g` is always `10` in `t_benign2` and in `main` after `t_benign2` is created.

The issue seems to be that the `thread` analysis answers that in the place where `g` is written the analysis is in single-threaded mode again (as `t_benign` exited cleanly). However., the privatizations do not properly support going back to single-threaded mode and then back to multi-threaded mode later.

michael-schwarz:
This is the immediate soundness fix needed for #1043.

The idea here is to distinguish the situation where the program is currently single-threaded but has been multi-threaded before from the situation where no threads have been started yet at all.

The privatizations all assume that no recovery from multi-threaded mode is possible, so if an analysis such as `thread` is activated, they misbehave and become unsound.

I will also attempt to make some of the privatizations support phases, but this requires fundamental changes, e.g., to how protecting locksets are computed, so I will do that in a separate PR.

Closes #1043

### DIFF
--- a/src/analyses/accessAnalysis.ml
+++ b/src/analyses/accessAnalysis.ml
@@ -43,7 +43,7 @@ struct
       + [deref=true], [reach=true] - Access [exp] by dereferencing transitively (reachable), used for deep special accesses. *)
   let access_one_top ?(force=false) ?(deref=false) ctx (kind: AccessKind.t) reach exp =
     if M.tracing then M.traceli "access" "access_one_top %a %b %a:\n" AccessKind.pretty kind reach d_exp exp;
-    if force || !collect_local || !emit_single_threaded || ThreadFlag.is_currently_multi (Analyses.ask_of_ctx ctx) then (
+    if force || !collect_local || !emit_single_threaded || ThreadFlag.has_ever_been_multi (Analyses.ask_of_ctx ctx) then (
       if deref then
         do_access ctx kind reach exp;
       Access.distribute_access_exp (do_access ctx Read false) exp
--- a/src/analyses/accessAnalysis.ml
+++ b/src/analyses/accessAnalysis.ml
@@ -43,7 +43,7 @@ struct
       + [deref=true], [reach=true] - Access [exp] by dereferencing transitively (reachable), used for deep special accesses. *)
   let access_one_top ?(force=false) ?(deref=false) ctx (kind: AccessKind.t) reach exp =
     if M.tracing then M.traceli "access" "access_one_top %a %b %a:\n" AccessKind.pretty kind reach d_exp exp;
-    if force || !collect_local || !emit_single_threaded || ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) then (
+    if force || !collect_local || !emit_single_threaded || ThreadFlag.is_currently_multi (Analyses.ask_of_ctx ctx) then (
       if deref then
         do_access ctx kind reach exp;
       Access.distribute_access_exp (do_access ctx Read false) exp
--- a/src/analyses/mutexAnalysis.ml
+++ b/src/analyses/mutexAnalysis.ml
@@ -209,7 +209,7 @@ struct
 
   let event ctx e octx =
     match e with
-    | Events.Access {exp; lvals; kind; _} when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* threadflag query in post-threadspawn ctx *)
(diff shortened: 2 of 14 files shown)


tests/regression/58-base-mm-tid/24-phases-sound.c
```diff
@@ -0,0 +1,46 @@
...
+#include <goblint.h>
+
+int g = 10;
+
+pthread_mutex_t A = PTHREAD_MUTEX_INITIALIZER;
+
+void *t_benign(void *arg) {
+  pthread_mutex_lock(&A);
+  g = 10;
+  __goblint_check(g == 10); //TODO
```
sim642, line 13: What's the status for these TODOs? Are they related to this single-threadedness stuff or just orthogonal?

src/analyses/accessAnalysis.ml
```diff
@@ -43,7 +43,7 @@ struct
       + [deref=true], [reach=true] - Access [exp] by dereferencing transitively (reachable), used for deep special accesses. *)
   let access_one_top ?(force=false) ?(deref=false) ctx (kind: AccessKind.t) reach exp =
     if M.tracing then M.traceli "access" "access_one_top %a %b %a:\n" AccessKind.pretty kind reach d_exp exp;
-    if force || !collect_local || !emit_single_threaded || ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) then (
+    if force || !collect_local || !emit_single_threaded || ThreadFlag.has_ever_been_multi (Analyses.ask_of_ctx ctx) then (
```
sim642, line 46:
```suggestion
    if force || !collect_local || !emit_single_threaded || ThreadFlag.is_currently_multi (Analyses.ask_of_ctx ctx) then (
```
Because accesses and races only depend on the current state. Or is there some reason for this?

src/analyses/mutexAnalysis.ml
```diff
@@ -209,7 +209,7 @@ struct
 
   let event ctx e octx =
     match e with
-    | Events.Access {exp; lvals; kind; _} when ThreadFlag.is_multi (Analyses.ask_of_ctx ctx) -> (* threadflag query in post-threadspawn ctx *)
+    | Events.Access {exp; lvals; kind; _} when ThreadFlag.has_ever_been_multi (Analyses.ask_of_ctx ctx) -> (* threadflag query in post-threadspawn ctx *)
```
sim642, line 212:
```suggestion
    | Events.Access {exp; lvals; kind; _} when ThreadFlag.is_currently_multi (Analyses.ask_of_ctx ctx) -> (* threadflag query in post-threadspawn ctx *)
```
Or is there some reason it needs to account for history?

src/analyses/accessAnalysis.ml
michael-schwarz: The access event needs to be emitted here, such that protection is computed in a way that coincides with the way the privatizations expect it to be computed.
This does only impact protection but race checking still works!

src/analyses/mutexAnalysis.ml
michael-schwarz: As above, this is needed for protection to be computed as desired.

sim642 (approved):

PR #1047, opened 2023-05-15
Split and remove `Goblintutil`

sim642:
`Goblintutil` is a another bottleneck module in our dependency structure and it consists of many unrelated things. This splits them up into less generic modules and removes the dumping ground of `Goblintutil` altogether.

### DIFF
--- a/conf/incremental.json
+++ b/conf/incremental.json
@@ -28,10 +28,12 @@
     "trace": {
       "context": true
     },
-    "debug": true,
     "timing": {
       "enabled": true
     }
+  },
+  "warn": {
+    "debug": true
   },
   "result": "none",
   "solver": "td3",
--- a/conf/minimal_incremental.json
+++ b/conf/minimal_incremental.json
@@ -27,10 +27,12 @@
     "trace": {
       "context": true
     },
-    "debug": true,
     "timing": {
       "enabled": true
     }
+  },
+  "warn": {
+    "debug": true
   },
   "result": "none",
   "solver": "td3",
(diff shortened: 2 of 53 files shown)


src/util/uniqueType.ml
```diff
@@ -0,0 +1,16 @@
...
+  with Not_found ->
+    let i = Cilfacade.create_var (makeGlobalVar ("{struct "^c.cname^"}") (TComp (c,[]))) in
+    Hashtbl.add type_inv_tbl c.ckey i;
+    i
+
+let find (t:typ): varinfo option =
+  let me_gusta x = List.mem x (GobConfig.get_string_list "exp.unique") in
+  match unrollType t with
+  | TComp (ci,_) when me_gusta ci.cname -> Some (type_inv ci)
+  | _ -> (None : varinfo option)
```
michael-schwarz, line 16: What exactly is this??? Why `me_gusta`?

src/util/uniqueType.ml
```diff
@@ -0,0 +1,16 @@
+open GoblintCil
+
+(* Type invariant variables. *)
+let type_inv_tbl = Hashtbl.create 13
```
michael-schwarz, line 4: Is it ok to not marshal this table?

michael-schwarz (approved):

src/util/uniqueType.ml
sim642: Possibly not, but `exp.unique` probably hasn't been used in the era of incremental and I wasn't going to try to refactor all this here.

PR #1048, opened 2023-05-15
String Literals Domain

nathanschmidt:
This PR extends the base analysis with cases for `strcpy` (adapted), `strncpy`, `strcat`, `strncat`, `strlen`, `strstr`, `strcmp` and `strncmp`. The following scenarios can now be analyzed:

- If the parameter `dest` of `strcpy`, `strncpy`, `strcat` or `strncat` potentially is a string literal, Goblint will now warn about a possible undefined behavior. For this purpose, the previous implementation of `strcpy` had to be adapted.
- If all the potential values of the parameters of the functions `strlen`, `strstr`, `strcmp` and `strncmp` are string literals, Goblint can now compute the exact return value of these functions.

I also added two regression tests, both with assertions that can now be tracked and assertions for which the String Literals Domain can still not make any assertion.
_Note: The regression tests fail on MacOS while they succeed on Linux. I couldn't yet figure out why that is the case._


michael-schwarz:
@sim642 You had some way to debug OS X things iirc? Do you mind sharing how that worked again?

michael-schwarz (commented):
Other than the CI issue, LGTM!

michael-schwarz:
It looks like the warning `Invalid call to builtin_object_size` actually comes from CIL. Maybe you can file an issue against the goblint-cil repository for this issue @nathanschmidt?
And until now, we could exclude the tests for OS X here in the same manner I did it in some other regression tests, i.e., by using `#ifdef`s.

sim642:
> @sim642 You had some way to debug OS X things iirc? Do you mind sharing how that worked again?

The CI summary page has a `suite_result` artifact at the bottom, which contains all the logs from the failed OSX job.

tests/regression/71-strings/01-string_literals.c
```diff
@@ -0,0 +1,84 @@
...
+#include <string.h>
+#include <stdlib.h>
+
+char* hello_world() {
+    return "Hello world!";
+}
+
+void id(char* s) {
+    s = s;
+}
```
sim642, line 13: I'm not sure what this function is trying to achieve. The identity copy of the pointer/string literal is only in the scope of this function, so even something like `s = s + 1` wouldn't influence the passed string in the caller.

tests/regression/71-strings/01-string_literals.c
```diff
@@ -0,0 +1,84 @@
...
+    #ifdef __APPLE__
+        /* the following portion fails on macOS because of a spurious warning:
+         * see issue goblint/cil#143
+         *
+         * remove #ifdef portion and change "should warn" to normal warning as soon as issue fixed */
+    #else
+        strcpy(s1, "hi"); // should warn
+        strncpy(s1, "hi", 1); // should warn
+        strcat(s1, "hi"); // should warn
+        strncat(s1, "hi", 1); // should warn
```
sim642, line 72: If they should warn, then it should be `// WARN`. Otherwise this isn't really testing anything.

src/cdomains/lval.ml
```diff
@@ -256,6 +256,18 @@ struct
...
+      if n > String.length x then
+        Some x
+      else if n < 0 then
+        None
+      else
+        Some (String.sub x 0 n)
+    | _ -> None
+  let to_string_length = function
+    | StrPtr (Some x) -> Some (String.length x)
+    | _ -> None
```
sim642, line 270: I suspect this might not be so easy because C strings and OCaml strings are different. In particular a string literal `"hello\0world"` would have `strlen` 5, but the OCaml string has length 11.
The same probably applies to other length-related functionality as well.

Maybe CIL has some helper functions for manipulating strings with C semantics?

src/cdomains/addressDomain.ml
```diff
@@ -103,7 +103,86 @@ struct
...
   (* strings *)
   let from_string x = singleton (Addr.from_string x)
+      
   let to_string x = List.filter_map Addr.to_string (elements x)
+      
+  let to_string_length x =
+    let transform elem =
+      match Addr.to_string_length elem with
+      | Some x -> Idx.of_int IUInt (Z.of_int x)
+      | None -> Idx.top_of IUInt in
```
sim642, line 113: Isn't this assuming `size_t` is `unsigned int`? `Cil.kindOfSizeOf` would probably be better.

src/cdomains/addressDomain.ml
```diff
@@ -103,7 +103,86 @@ struct
...
+
+  let substring_extraction haystack needle =
+    (* map all StrPtr elements in input address sets to contained strings *)
+    let haystack' = List.map Addr.to_string (elements haystack) in
+    let needle' = List.map Addr.to_string (elements needle) in
+
+    (* helper functions *)
+    let extract_string = function
+      | Some s -> s
+      | None -> failwith "unreachable" in
```
sim642, line 127: This is just `Option.get`.

src/cdomains/addressDomain.ml
```diff
@@ -103,7 +103,86 @@ struct
...
+      | Some s -> from_string s
+      | None -> null_ptr in
+    let compute_substring s1 s2 =
+      try 
+        let i = Str.search_forward (Str.regexp_string s2) s1 0 in
+        Some (String.sub s1 i (String.length s1 - i))
+      with Not_found -> None in
+
+    (* if any of the input address sets contains an element that isn't a StrPtr, return top *)
+    if List.exists ((=) None) haystack' || List.exists ((=) None) needle' then
```
sim642, line 138:
```suggestion
    if List.mem None haystack' || List.mem None needle' then
```

src/cdomains/addressDomain.ml
```diff
@@ -103,7 +103,86 @@ struct
...
+      | None -> Addr.to_string in
+
+    (* map all StrPtr elements in input address sets to contained strings / n-substrings *)
+    let x' = List.map f (elements x) in 
+    let y' = List.map f (elements y) in 
+
+    (* helper functions *)
+    let extract_string = function
+      | Some s -> s
+      | None -> failwith "unreachable" in
```
sim642, line 160: `Option.get`

src/cdomains/addressDomain.ml
```diff
@@ -103,7 +103,86 @@ struct
...
+      let res = String.compare s1 s2 in
+      if res = 0 then
+        Idx.of_int IInt Z.zero
+      else if res > 0 then
+        Idx.starting IInt Z.one
+      else
+        Idx.ending IInt (Z.neg (Z.one)) in
+
+    (* if any of the input address sets contains an element that isn't a StrPtr, return top *)
+    if List.exists ((=) None) x' || List.exists ((=) None) y' then
```
sim642, line 171:
```suggestion
    if List.mem None x' || List.mem None y' then
```

src/cdomains/addressDomain.ml
```diff
@@ -103,7 +103,86 @@ struct
...
+      Idx.top_of IInt
+    else
+      (* else compare every string of x' with every string of y' and return the least upper bound *)
+      BatList.cartesian_product x' y'
+      |> List.map (fun (s1, s2) -> compare (extract_string s1) (extract_string s2))
+      |> List.fold_left Idx.join (Idx.bot_of IInt)
+
+  let string_writing_defined dest =
+    (* if the destination address set contains a StrPtr, writing to such a string literal is undefined behavior *)
+    if List.exists (fun x -> match x with Some _ -> true | None -> false) (List.map Addr.to_string (elements dest)) then
```
sim642, line 181:
```suggestion
    if List.exists Option.is_some (List.map Addr.to_string (elements dest)) then
```

src/analyses/base.ml
```diff
@@ -2059,6 +2059,53 @@ struct
...
+        begin match eval_rv (Analyses.ask_of_ctx ctx) gs st n with
+          | `Int i ->
+            begin match ID.to_int i with
+              | Some x -> Some (Z.to_int x)
+              | _ -> Some (-1)
+            end
+          | _ -> Some (-1)
+        end
+      (* do nothing if all characters are needed *)
+      | _ -> None
```
sim642, line 2088: Why those `-1`s? Shouldn't they also be `None` if the integer is ambiguous or something?

src/analyses/base.ml
```diff
@@ -2059,6 +2059,53 @@ struct
     let st: store = ctx.local in
     let gs = ctx.global in
     let desc = LF.find f in
+    let memory_copying dst src =
+      let dest_a, dest_typ = addr_type_of_exp dst in
+      let src_lval = mkMem ~addr:(Cil.stripCasts src) ~off:NoOffset in
+      let src_typ = eval_lv (Analyses.ask_of_ctx ctx) gs st src_lval
+                    |> AD.get_type in
+      (* when src and destination type coincide, take value from the source, otherwise use top *)
+      let value = if typeSig dest_typ = typeSig src_typ then
```
sim642, line 2068: Previously `Memcpy` used the same type lookup pattern for both `src` and `dst`, but now they're different. What's the reason for changing that?

tests/regression/71-strings/01-string_literals.c
nathanschmidt: Yes, I messed that one up. Thanks for pointing it out!

tests/regression/71-strings/01-string_literals.c
nathanschmidt: The problem is that if I keep `// WARN`, than the test still fails on macOS because no warning was issued as the portion is never reached on an Apple device. Is there any smart way to test if the warnings are correctly generated on Linux while ignoring that annotation on OS X?

src/cdomains/lval.ml
nathanschmidt: That's absolutely right, again thank you for thinking about this case! I couldn't find a CIL function that would help me in this case in a quick search, but I'll look at it again. Otherwise, `String.split_on_char '\x00' x` and taking the first element of the resulting list would do the trick.

src/cdomains/addressDomain.ml
nathanschmidt: In *cil.ml*, `Cil.kindOfSizeOf` is defined as `ref IUInt`, so would this make any difference? There is also a signature mismatch problem because of the `ref` if using `Cil.kindOfSizeOf`

src/analyses/base.ml
nathanschmidt: If they'd be `None`, then when the next function is called, e.g. `string_comparison`, then the equivalent to `strcmp` (without the `n`) would be performed as then instead of using `Addr.to_n_string`, `Addr.to_string` would be picked. However, when returning `Some (-1)`, then it's still `Addr.to_n_string` and the latter transforms negative values to `None`, which leads `string_comparions` to just return top

src/analyses/base.ml
nathanschmidt: That was just a simplification as the operations performed to compute `dest_a` and `dest_typ` originally are just the same as `addr_type_of_exp`. Nothing has changed semantically. I do not use `addr_type_of_exp` for `src` as it would unnecessarily compute an address not needed

tests/regression/71-strings/01-string_literals.c
michael-schwarz: We decided to go with `*0;` in the OS X case so we get a warning. This is hacky, but I guess ok.

michael-schwarz: Nice that you managed to make it work!

michael-schwarz:
@sim642 do you want to do a final pass or are you good with merging?

src/cdomains/lval.ml
sim642: It is now fixed for `strlen`, but I think it still needs to be fixed for some of the other functions. I think it matters for `strstr` and `strcmp` as well. Right now they're just performing the operations on OCaml strings, but all of those operations need to be performed with C semantics, so only up to the NUL byte considered when checking for inclusion etc.

I'm wondering if a better solution is to maybe just do the stripping initially, so it doesn't have to be done for every operation individually. Or maybe CIL is already doing that as well?
I haven't checked, but it would be good to have tests for these things. If sv-benchmarks doesn't contain such cases, it would be nice to also contribute those because this detail is easy to get wrong and become unsound.

nathanschmidt: My first assumption was that CIL would cut the string at the first NULL byte, that's why I didn't do the stripping. After your comment, I tested this and found that's not the case.

As you're right that this also concerns `strstr` and `str(n)cmp`, I introduced a new function `to_c_string` performing the stripping which I now use accross all my other functions. I didn't change `to_string` as some parts of the base analysis not related to my PR rely on it.

I added a new *02-string_literals_with_null.c* test to verify this edge case is now correctly covered.

PR #1049, opened 2023-05-16
Support for recovering from multi-thread mode

michael-schwarz:
This adds support for recovering from multi-threaded mode.

This works as follows: 

- Collect a set of mutexes that are protecting when recovery is possible
- For MutexMeetTID (currently only supported privatization)
     - In case of being single-threaded again, updates happen to appropriate L
     - Upon going multi-threaded again, remove all globals from local state that will be no longer protected.
     
This uncovered a unsoundness in the Mutex-Meet-TID privatizations when a new thread  is created while holding mutexes, because the local cpa was passed as the argument instead of `CPA.bot ()` which is appropriate in case no mutexes are held.

This also uncovers an issue in the following program: The assert below behaves as expected, but if the original one is uncommented, we suddenly run into an issue with computations on bottom?!  Any idea here @sim642?

~~~C
// PARAM: --set ana.path_sens[+] threadflag --set ana.base.privatization mutex-meet-tid --enable ana.int.interval --set ana.activated[+] threadJoins
#include <pthread.h>
#include <goblint.h>

int g = 10;

void *t_benign(void *arg) {
  return NULL;
}

void *t_benign2(void *arg) {
  pthread_mutex_lock(&A);
  int x =  g == 40;
  // Adding this back leads to ArithmeticOnBottom errors ?!?!
  // __goblint_check(g == 40); //TODO
  __goblint_check(x); //TODO
  return NULL;
}

int main(void) {

  pthread_t id2;
  pthread_join(id2, NULL);

  pthread_mutex_lock(&A);
  g = 30;
  pthread_create(&id2, NULL, t_benign2, NULL);
  g = 40;
  pthread_mutex_unlock(&A);

  return 0;
}
~~~


michael-schwarz:
I am starting to wonder if the reason is that upon evaluating the second thread, the side-effect from the first one unlocking `a` again might not have happened yet, and so there is no contribution from that thread yet. On the other hand, this thread knows that `g` has been written for sure, so the initializer value associated with [g] is not read, while for `[A]` this is the only thing that is read before more side-effects have happened. This could lead to bottom somewhere, which might be caught somewhere in one case but not the other?

src/analyses/mutexAnalysis.ml
```diff
@@ -36,17 +36,27 @@ struct
...
-        let name () = "write"
+        let name () = "readwriteRecover"
       end
-      include Lattice.Prod (ReadWrite) (Write)
+      module WriteRecover =
+      struct
+        include G0
+        let name () = "writeRecover"
+      end
+      include Lattice.Prod4 (ReadWriteNoRecover) (WriteNoRecover) (ReadWriteRecover) (WriteRecover)
```
sim642, line 59: Wouldn't something like
```ocaml
module P = Lattice.Prod (ReadWrite) (Write)
include Lattice.Prod (P) (P)
```
be easier because the recovery and non-recovery cases are symmetric?

src/analyses/base.ml
```diff
@@ -1127,6 +1127,7 @@ struct
       | `Bot   -> Queries.ID.top () (* out-of-scope variables cause bot, but query result should then be unknown *)
       | `Top   -> Queries.ID.top () (* some float computations cause top (57-float/01-base), but query result should then be unknown *)
       | v      -> M.debug ~category:Analyzer "Base EvalInt %a query answering bot instead of %a" d_exp e VD.pretty v; Queries.ID.bot ()
+      | exception (IntDomain.ArithmeticOnIntegerBot _) when not !AnalysisState.should_warn -> Queries.ID.top () (* for some privatizations, values can intermediately be bot because side-effects have not happened yet *)
```
sim642, line 1130: Why conditional on `should_warn` though? In `eval_rv` (which this bypasses) it's unconditional.

src/analyses/mutexAnalysis.ml
```diff
@@ -223,22 +238,43 @@ struct
...
+                    let protected =
+                      if write then
+                        (vs_empty, vs_empty, vs_empty, vs)
                     else
-                      (vs, vs_empty)
+                      (vs_empty, vs_empty, vs, vs_empty)
                   in
                   ctx.sideg (V.protected addr) (G.create_protected protected)
-                ) held_locks
+                  ) held_recovery;
```
sim642, line 277: Something funky with the indentation here.

src/domains/queries.ml
```diff
@@ -35,9 +35,9 @@ module MustBool = BoolDomain.MustBool
 module Unit = Lattice.Unit
 
 (* Helper definitions for deriving complex parts of Any.compare below. *)
-type maybepublic = {global: CilType.Varinfo.t; write: bool} [@@deriving ord, hash]
-type maybepublicwithout = {global: CilType.Varinfo.t; write: bool; without_mutex: PreValueDomain.Addr.t} [@@deriving ord, hash]
-type mustbeprotectedby = {mutex: PreValueDomain.Addr.t; global: CilType.Varinfo.t; write: bool} [@@deriving ord, hash]
+type maybepublic = {global: CilType.Varinfo.t; write: bool; recoverable: bool} [@@deriving ord, hash]
+type maybepublicwithout = {global: CilType.Varinfo.t; write: bool; without_mutex: PreValueDomain.Addr.t; recoverable: bool} [@@deriving ord, hash]
+type mustbeprotectedby = {mutex: PreValueDomain.Addr.t; global: CilType.Varinfo.t; write: bool; recoverable: bool} [@@deriving ord, hash]
```
sim642, line 40: The meaning of `recoverable` here (and probably elsewhere) would not be clear at all to anyone who isn't aware of this PR (e.g. a new student). We need to somehow make it clear at all the interfaces what is this recovery that we're talking about.

src/analyses/basePriv.ml
```diff
@@ -579,12 +596,25 @@ struct
     {st with cpa= cpa_local }
 
   let threadenter ask (st: BaseComponents (D).t): BaseComponents (D).t =
-    (* Copy-paste from Base make_entry *)
-    let globals = CPA.filter (fun k v -> is_global ask k) st.cpa in
-    (* let new_cpa = if !earlyglobs || ThreadFlag.is_multi ctx.ask then CPA.filter (fun k v -> is_private ctx.ask ctx.local k) globals else globals in *)
-    let new_cpa = globals in
+    (* We cannot copy over protected things, the thread may start with things privatized that are overwritten before becoming public *)
```
sim642, line 599: Is there an example of this? How can it already privatize something if the created thread (not the creator thread!) starts with no locks?

src/analyses/base.ml
michael-schwarz: The idea was to let this still fail if it happens when a solution has been computed, since I am not sure if it is also benign if it happens there.

src/analyses/basePriv.ml
sim642: The comment sounds backwards then: "the thread may start with things privatized". I guess what it really should say is that the creator thread may have things privatized, but the created thread cannot have anything privatized on entry, so the privatized things of the creator must be removed.

michael-schwarz: Yeah, it was referring to the situation before and why that was wrong which isn't terribly useful after we've fixed it.

src/analyses/mutexAnalysis.ml
michael-schwarz: I am now trying an extensive-ish re-write also of check_fun, let's see if that is more clear then.

michael-schwarz: I think it turned out quite nicely, let me know what you think!

src/domains/queries.ml
```diff
@@ -42,13 +42,13 @@ module Unit = Lattice.Unit
     - whenever g is accessed and there are any threads other than the main thread that are created but not joined yet, m is held
 *)
 module Protection = struct
-  type t = Strong | Weak [@@deriving ord, hash]
+  type protection = Strong | Weak [@@deriving ord, hash]
```
sim642: Why rename it away from `Protection.t`, which is the standard pattern? `Protection.protection` has such redundancy.
If it's just to avoid some `open`ing issues, then that's a clear sign to not use `open` so much.

src/analyses/mutexAnalysis.ml
```diff
@@ -114,39 +159,36 @@ struct
...
-      let locks = Lockset.export_locks ls in
-      if write then (Mutexes.bot (), locks) else (locks, Mutexes.bot ())
-    in
-    let non_overlapping locks1 locks2 =
-      let intersect = GProtecting.join locks1 locks2 in
-      GProtecting.is_top intersect
-    in
+    (* get the set of mutexes protecting the variable v in the given mode *)
+    let protecting ~write mode v = GProtecting.get ~write mode (G.protecting (ctx.global (V.protecting v))) in
+    let non_overlapping locks1 locks2 = Mutexes.is_empty @@ Mutexes.inter locks1 locks2 in
```
sim642, line 164: This changes the `non_overlapping` check back to just `Mutexes` operations. For traces in 1ce6e7889419f8e96acc551537752ea485b23ebc, it was changed from `Mutexes` to the global invariant based check and there was some reason for making the check more abstract.

Are you sure that the `GProtecting` version isn't still needed? It might have to do with the read- and write-lock distinction, which is not captured by `Mutexes`. Could be that a regression test was never added for that and this is making it less precise again.

src/domains/queries.ml
michael-schwarz: It's not avoiding issues per se, but writing `Protection.*` everywhere is annoying, so I want to open them. But then having a type flying around in a module that confusingly comes from some other module is just annoying.

src/analyses/mutexAnalysis.ml
michael-schwarz: I don't think it is necessary here. All values passed to `non-overlapping` were constructed by `check_fun` such that there is only one component that is not the empty lockset. So I think the new code is functionally equivalent but clearer to read now!

src/domains/queries.ml
sim642: I renamed it back because due to type-directed disambiguation there's rarely the need to write out `Queries.Protection.Strong`.

src/analyses/mutexAnalysis.ml
sim642: Looks like originally it was for OSEK mutexes and their priorities. In the remaining instance `G` was referring to the same type of locks, so indeed it is equivalent.

sim642 (approved):

PR #1050, opened 2023-05-17
Use-After-Free (UAF) Analysis First Iteration

mrstanb:
To sum up the PR infos:
* Added a UAF analysis to Goblint
* For this, I also had to add `free` as a supported library function (similar to `malloc`)
* Added a simple C regression test file with a UAF bug

**Note:** Currently the analysis doesn't seem to be really working properly, hence it's only a first iteration. I'm debugging my way through it to try and find out what's going on and what's wrong.

@michael-schwarz feel free to do a review round


src/analyses/libraryFunctions.ml
```diff
@@ -19,6 +19,8 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("strcpy", special [__ "dest" [w]; __ "src" [r]] @@ fun dest src -> Strcpy { dest; src });
     ("malloc", special [__ "size" []] @@ fun size -> Malloc size);
     ("realloc", special [__ "ptr" [r; f]; __ "size" []] @@ fun ptr size -> Realloc { ptr; size });
+    (* TODO: Do we need accessKind of "r" for "free" or is "f" alone already sufficient? *)
```
michael-schwarz, line 22: @sim642: Any opinion on this TODO? You are the one intricately familiar with these invalidation actions!

sim642: `f` should suffice for `free`. The reason `realloc` also has `r` is because it has to read the existing memory to copy it before it frees.

src/analyses/libraryFunctions.ml
```diff
@@ -487,6 +490,16 @@ let classify fn exps: categories =
...
+    (*
+     * TODO: Docs (https://goblint.readthedocs.io/en/latest/developer-guide/extending-library/#library-function-specifications)
+     * say we shouldn't add new specifications in classify
+     * Should I remove the guard for "free"-related functions here in this case?
+    *)
+  | "kfree" | "usb_free_urb" ->
+    begin match exps with
+      | ptr::_ -> `Free ptr
+      | _ -> strange_arguments ()
+    end
```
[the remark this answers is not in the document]
sim642: Where do these two functions even come from? Are you using some kind of Linux kernel benchmarks?

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,188 @@
...
+    | _ -> false
+
+  let check_lval (lval:lval) ctx =
+    let state = ctx.local in
+    match lval with
+    | (Mem e, _) ->
+      begin match ctx.ask (Queries.EvalValue e) with
+        | a when not (Queries.VD.is_top a) ->
+          begin match a with
+            | `Blob (v, s, t) -> if D.mem (v, s, t) state then true else false
```
michael-schwarz, line 38: Here, you are checking that the entire blob abstract value is in your set, including the abstraction of the  value etc...  Probably this checking should happen at a higher level such as the varinfo associated with the blob.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,188 @@
...
+  module D = SetDomain.Make(ValueDomain.Blobs)
+  module C = Lattice.Unit
+
+  (** TODO: What about context-sensitivity? *)
+  let context _ _ = ()
+
+
+  (* HELPER FUNCTIONS *)
+
+  let check_exp (exp:exp) ctx =
```
michael-schwarz, line 21: Here, we would also need to check intermediate expressions. E.g.  `a[*p] = 5` constitutes a UAF if `p` points to dynamically allocated memory, that has been freed.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,188 @@
...
+
+  (* TRANSFER FUNCTIONS *)
+
+  let assign ctx (lval:lval) (rval:exp) : D.t =
+    let state = ctx.local in
+    (* Intuition:
+      * Check if lval and/or lval has an expression that points to a "maybe freed" blob
+      * If yes -> send out a WARNING; otherwise -> don't WARN
+      * In either case above -> don't change the CFG node's state
+    *)
```
michael-schwarz, line 54: :+1: yup, that's exactly the right intuition!

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,188 @@
...
+    match lval with
+    | Some lval ->
+      if check_lval lval ctx then
+        M.warn "WARN: lval in enter contains maybe freed blob";
+      [caller_state, callee_state]
+    | None -> [caller_state, callee_state]
+
+  (* TODO: Unsure about this transfer function. Should we leave it as an identity function? *)
+  let combine_env ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (callee_local:D.t) (f_ask:Queries.ask) : D.t =
+    callee_local
```
michael-schwarz, line 101: That depends: In enter it looks like you filter out things that are unreachable from the callee? These would have to be added back here.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,188 @@
...
+          | a when not (Queries.VD.is_top a) ->
+            begin match a with
+              | `Blob (v, s, t) -> D.add (v, s, t) state
+              | _ -> state
+            end
+          | _ -> state
+        end
+      else
+        (* TODO: Is this way of printing in the else block okayish? *)
+        let () = M.warn "WARN: trying to free an already freed ptr" in
```
michael-schwarz, line 132: Yes, sure!

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,188 @@
...
+            | (Mem e, _) ->
+              begin match ctx.ask (Queries.EvalValue e) with
+                | a when not (Queries.VD.is_top a) ->
+                  begin match a with
+                    | `Blob (v, s, t) ->
+                      (*
+                       * If we're trying to allocate a blob that is maybe freed,
+                       * then we can remove it from the maybe freed set
+                       * TODO: Makes sense?
+                       *)
```
michael-schwarz, line 154: I'm not sure if the corresponding blob is even already created here, so I fear this will usually yield `VD.top ()`. Also, upon newly creating the blob here, the abstract value inside it will usually not be the same as the one when free was called. So `mem` will probably never trigger.

src/analyses/useAfterFree.ml
michael-schwarz: Actually, come to think of it, we should not track the blobs themselves, but rather their varinfos!

src/analyses/libraryFunctions.ml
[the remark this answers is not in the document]
mrstanb: > Where do these two functions even come from? Are you using some kind of Linux kernel benchmarks?

I saw that there are analogous functions for `malloc` and that's why I decided to add their `free`-related counterparts as well. I haven't used any Linux kernel benchmarks thus far.

src/analyses/libraryDesc.ml
```diff
@@ -122,6 +123,7 @@ let special_of_old classify_name = fun args ->
   | `Malloc e -> Malloc e
   | `Calloc (count, size) -> Calloc { count; size; }
   | `Realloc (ptr, size) -> Realloc { ptr; size; }
+  | `Free ptr -> Free ptr
```
michael-schwarz, line 126: Is this needed at all? Is special_of_old ever called for `Free ?

tests/regression/71-use_after_free/01-simple-uaf.c
```diff
@@ -0,0 +1,12 @@
...
+#include <stdio.h>
+
+int main() {
+    int *ptr = malloc(sizeof(int));
+    *ptr = 42;
+
+    free(ptr);
+
+    *ptr = 43; // Should report "Use After Free (CWE-416)"
+    free(ptr); // Should report "Double Free (CWE-415)"
```
michael-schwarz, line 11: Maybe we should try annotating with `//WARN` (maybe disabling some of the other warnings that also warn here. To then get warnings for both different cases one would probably have to duplicate it.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,219 @@
...
+
+  (* HELPER FUNCTIONS *)
+
+  (* Took inspiration from malloc_null *)
+  let may (f:'a -> 'b) (x:'a option) : unit =
+    match x with
+    | Some x -> f x
+    | None -> ()
+
+  (* Also took inspiration from malloc_null *)
```
michael-schwarz, line 31: I don't think it is necessary to comment this everywhere :smile:

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,219 @@
...
+    match exp with
+    | Some ret ->
+      begin match get_concrete_exp ret with
+        | Some v ->
+          if D.mem v state then
+            D.add (freed_var_at_return ()) state
+          else state
+        | None -> state
+      end
+    | None -> state
```
michael-schwarz, line 154: Same here, I don't think you need to add the return variable here.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,219 @@
+(** An analysis for the detection of use-after-free vulnerabilities. *)
+
+open GoblintCil
+open Analyses
+open MessageCategory
+
+(* TODO: Maybe come up with a better name for top at some point? *)
+module ToppedVarInfoSet = SetDomain.ToppedSet(CilType.Varinfo)(struct let topname = "Unknown" end)
```
michael-schwarz, line 8: Will switch to tracking malloc varinfos only.

src/analyses/libraryDesc.ml
mrstanb: Seems like it's not needed. I'll make sure to remove it

mrstanb: I think that I forgot to remove the `` `Free `` constructor of the `categories` type in `libraryFunctions.ml` (cf. https://github.com/mrstanb/analyzer/blob/uaf-analysis-first-iteration/src/analyses/libraryFunctions.ml#L458). I'm assuming `` `Free `` is not needed there either, right?

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,219 @@
...
+  let context _ _ = ()
+
+
+  (* HELPER FUNCTIONS *)
+
+  (* Took inspiration from malloc_null *)
+  let may (f:'a -> 'b) (x:'a option) : unit =
+    match x with
+    | Some x -> f x
+    | None -> ()
```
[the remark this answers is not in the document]
mrstanb: Indeed :)

michael-schwarz:
Thank you for the PR, and sorry I am only now getting around to reviewing it in detail.

Here's a first issue: As expected, this does not soundly account for UAF in the concurrent setting yet:

~~~C
#include <stdlib.h>
#include <stdio.h>
#include <pthread.h>

int* gptr;

// Mutex to ensure we don't get race warnings, but the UAF warnings we actually care about
pthread_mutex_t mtx = PTHREAD_MUTEX_INITIALIZER;

void *t_other(void* p) {
    pthread_mutex_lock(&mtx);
    free(gptr); //WARN
    pthread_mutex_unlock(&mtx);
}

int main() {
    gptr = malloc(sizeof(int));
    *gptr = 42;

    pthread_t thread;
    pthread_create(&thread, NULL, t_other, NULL);

    pthread_mutex_lock(&mtx);
    *gptr = 43; //WARN
    free(gptr); //WARN
    pthread_mutex_unlock(&mtx);

    return 0;
}
~~~

There are two ways how to fix this: We immediately go for a solution to this problem in the multi-threaded context, or we add a simple heuristics to basically make the analysis warn whenever something is accessed in multi-threaded mode, and improve precision in a separate PR.

As a idea: We might want to record  the thread ids of threads  (and potentially which threads have already been joined at that point) that free the memory in a global invariant. Then, upon a potentially multi-threaded access, I need to warn if the thread ids do indicate that the free must have happen only after the current function.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,153 @@
...
+  module D = ToppedVarInfoSet
+  module C = Lattice.Unit
+
+  (** TODO: Try out later in benchmarks to see how we perform with and without context-sensititivty *)
+  let context _ _ = ()
+
+
+  (* HELPER FUNCTIONS *)
+
+  let rec warn_lval_might_contain_freed ?(is_double_free = false) (transfer_fn_name:string) (lval:lval) ctx =
```
michael-schwarz, line 24: Would it make sense to switch the order of  `ctx` and `lval` here? That might make invocations less awkward e.g. in `return`.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,153 @@
...
+  let body ctx (f:fundec) : D.t =
+    ctx.local
+
+  let return ctx (exp:exp option) (f:fundec) : D.t =
+    Option.iter (fun x -> warn_exp_might_contain_freed "return" x ctx) exp;
+    ctx.local
+
+  let enter ctx (lval:lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
+    let caller_state = ctx.local in
+    Option.iter (fun x -> warn_lval_might_contain_freed "enter" x ctx) lval;
```
michael-schwarz, line 98: I think this check should be made in `combine_assign`: Consider a case where the calle calls `free()` for something that `lval` may point to.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,153 @@
...
+    | Free ptr ->
+      begin match ctx.ask (Queries.MayPointTo ptr) with
+        | a when not (Queries.LS.is_top a) && not (Queries.LS.mem (dummyFunDec.svar, `NoOffset) a) ->
+          begin try
+              let (v, _) = Queries.LS.choose a in
+              if ctx.ask (Queries.IsHeapVar v) then
+                D.add v state
+              else state
+            with Not_found -> state
+          end
```
michael-schwarz, line 139: What if there are several things that the pointer could point to? We would have to iterate over all of them and add them to the set of may-freed things.

src/analyses/useAfterFree.ml
```diff
@@ -0,0 +1,153 @@
...
+    let caller_state = ctx.local in
+    D.join caller_state callee_local
+
+  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (callee_local:D.t) (f_ask: Queries.ask): D.t =
+    ctx.local
+
+  let special ctx (lval:lval option) (f:varinfo) (arglist:exp list) : D.t =
+    let state = ctx.local in
+    Option.iter (fun x -> warn_lval_might_contain_freed ("special: " ^ f.vname) x ctx) lval;
+    List.iter (fun arg -> warn_exp_might_contain_freed ~is_double_free:(f.vname = "free") ("special: " ^ f.vname) arg ctx) arglist;
```
michael-schwarz, line 127: Great that you thought of this too!

michael-schwarz:
Something we should also write tests for: Developers like to have custom wrappers around calls to memory-related functions such as `free` or `malloc` to potentially plugin-in a custom allocator. I think we will be precise in presence of such usages, as we are context-sensitive in the pointer argument that is passed to free, but it might still be worth having tests for this.

src/analyses/useAfterFree.ml
mrstanb: I'll do the same for `warn_exp_might_contain_freed` for consistency's sake, as it's essentially dual to `warn_lval_might_contain_freed`

mrstanb: Also done for `warn_exp_might_contain_freed` now (cf. https://github.com/goblint/analyzer/pull/1050/commits/6bfcf02a0e4bbda27846383265a92d7da3c4a062)

src/analyses/useAfterFree.ml
mrstanb: Done in https://github.com/goblint/analyzer/pull/1050/commits/d518665e4d8c0cba299f4d5a34c8a3bf4c1aaa67

src/analyses/useAfterFree.ml
mrstanb: Thanks for noting this! Should be fixed in https://github.com/goblint/analyzer/pull/1050/commits/47eff4483145b24366c8b80089ffd1e8663cb01f

(after https://github.com/goblint/analyzer/pull/1050/commits/47eff4483145b24366c8b80089ffd1e8663cb01f, Goblint should now also warn for each possible pointed-to heap var if it's used after being freed; earlier a warning was only issued for a single heap var, selected via `Queries.LS.choose`)

mrstanb:
> There are two ways how to fix this: We immediately go for a solution to this problem in the multi-threaded context, or we add a simple heuristics to basically make the analysis warn whenever something is accessed in multi-threaded mode, and improve precision in a separate PR.

I think I'd go for the second option and add proper multi-threaded support in a separate PR, if we'd prefer to keep this PR rather lean and not too big. What do you think?

mrstanb:
> Something we should also write tests for: Developers like to have custom wrappers around calls to memory-related functions such as `free` or `malloc` to potentially plugin-in a custom allocator. I think we will be precise in presence of such usages, as we are context-sensitive in the pointer argument that is passed to free, but it might still be worth having tests for this.

Should be now covered by https://github.com/goblint/analyzer/pull/1050/commits/9e5cd7a199a5b04c35215cced5d09727fb48bf61

mrstanb:
> Thank you for the PR, and sorry I am only now getting around to reviewing it in detail.
> 
> Here's a first issue: As expected, this does not soundly account for UAF in the concurrent setting yet:
> 
> ```c
> //PARAM: --set ana.activated[+] useAfterFree
> #include <stdlib.h>
> #include <stdio.h>
> #include <pthread.h>
> 
> int* gptr;
> 
> // Mutex to ensure we don't get race warnings, but the UAF warnings we actually care about
> pthread_mutex_t mtx = PTHREAD_MUTEX_INITIALIZER;
> 
> void *t_other(void* p) {
>     pthread_mutex_lock(&mtx);
>     free(gptr); //WARN
>     pthread_mutex_unlock(&mtx);
> }
> 
> int main() {
>     gptr = malloc(sizeof(int));
>     *gptr = 42;
> 
>     pthread_t thread;
>     pthread_create(&thread, NULL, t_other, NULL);
> 
>     pthread_mutex_lock(&mtx);
>     *gptr = 43; //WARN
>     free(gptr); //WARN
>     pthread_mutex_unlock(&mtx);
> 
>     return 0;
> }
> ```
> 
> There are two ways how to fix this: We immediately go for a solution to this problem in the multi-threaded context, or we add a simple heuristics to basically make the analysis warn whenever something is accessed in multi-threaded mode, and improve precision in a separate PR.
> 
> As a idea: We might want to record the thread ids of threads (and potentially which threads have already been joined at that point) that free the memory in a global invariant. Then, upon a potentially multi-threaded access, I need to warn if the thread ids do indicate that the free must have happen only after the current function.

I now added this example C program as a regression test for UAF in a multi-threaded context (cf. https://github.com/goblint/analyzer/pull/1050/commits/a00814f3cd70d7d04831b36a206497fd5f3db9b7)

I also added a simple heuristic that warns for a potential UAF occuring whenver the program is definitely not single-threaded (cf. https://github.com/goblint/analyzer/pull/1050/commits/9545618a3ccba1dc7bbca051f980c4fc1803bc09)

As already discussed, the proper multi-threaded solution for UAF detection will be implemented in a follow-up PR

michael-schwarz (approved):

PR #1055, opened 2023-05-23
Add odoc API documentation

### LINKED ISSUE #1054: API documentation
While #208 is about documenting high-level usage and development patterns, the OCaml Goblint library also has ocamldoc comments here-and-there (and should obviously have more!) but there's no good documentation for the codebase itself.
Such documentation would also cover some more implementation-specific points from #208.

### TODO
- [x] GitHub Actions job to publish odoc to GitHub pages, as is standard for OCaml packages.
- [x] Overview/structure of the OCaml codebase.

    There is https://github.com/goblint/analyzer/blob/6a41ac5f84fa561d100139e82cbbe28d3461a898/src/main.camldoc, but it is very outdated, unused, incomplete and incompatible with odoc (due to `{!indexlist}`).

sim642:
Closes #1054.

This adds a GitHub Actions workflow that builds Goblint's OCaml API documentation with odoc and publishes it to https://goblint.github.io/analyzer/.

In order to be actually useful, the `Goblint_lib` module is manually written to organize all the modules in our huge flat namespace. Moreover, ocamldoc synopses are added for all those modules.

### TODO
- [x] Change workflow branch to `master` before merging.


src/analyses/commonPriv.ml
```diff
@@ -1,3 +1,5 @@
+(** Thread-modular value analysis utilities. *)
```
michael-schwarz, line 1: Maybe include a reference to the two main modules that use this here?

src/analyses/expRelation.ml
```diff
@@ -1,3 +1,5 @@
+(** Stateless symbolic comparison expression analysis ([expRelation]). *)
+
 (** An analysis specification to answer questions about how two expressions relate to each other.   *)
 (** Currently this works purely syntactically on the expressions, and only for =_{must}.            *)
```
michael-schwarz, line 4: This `=_{must}` needs to be escaped here.

src/analyses/expRelation.ml
```diff
@@ -1,3 +1,5 @@
...
 (** Does not keep state, this is only formulated as an analysis to integrate well into framework    *)
```
michael-schwarz, line 5:
```suggestion
(** Does not keep state, this is only formulated as an analysis to integrate well into the framework.  *)
```

src/analyses/expsplit.ml
```diff
@@ -1,3 +1,5 @@
+(** Path-sensitive analysis according to arbitrary given expressions ([expsplit]). *)
```
michael-schwarz, line 1:
```suggestion
(** Makes the analysis path-sensitive according to what an arbitrary given expression evaluates to ([expsplit]). *)
```

src/analyses/mCP.ml
```diff
@@ -1,4 +1,4 @@
-(** Master Control Program *)
+(** MCP analysis specification. *)
```
michael-schwarz, line 1:
```suggestion
(** MCP analysis specification.  Combines all active analysis into an n-tuple using an n-tuple of abstract values as its domain. *)
```

src/analyses/modifiedSinceLongjmp.ml
```diff
@@ -1,4 +1,6 @@
-(** Locally track the variables that may have been written since the corresponding jumpbuffer was set *)
+(** Analysis of variables modified since [setjmp] ([modifiedSinceLongjmp]). *)
+
+(* TODO: this name is wrong *)
```
michael-schwarz, line 3: Yeah, somehow never noticed that!

src/analyses/tutorials/constants.ml
```diff
@@ -1,3 +1,4 @@
+(** Simple intraprocedural integer constants analysis example ([constants]). *)
```
michael-schwarz, line 1: Maybe add a link to the tutorial?

src/framework/edge.ml
```diff
@@ -1,3 +1,6 @@
+(** CFG edge.
+    Corresponds to a (primitive) program statement between program points (and their states). *)
```
michael-schwarz, line 2: how do they correspond to states?

src/framework/varQuery.mli
```diff
@@ -1,3 +1,5 @@
+(** Queries for constraint variables related to semantic elements. *)
```
michael-schwarz, line 1: What do you mean by `semantic elements`?

src/solvers/generic.ml
```diff
@@ -1,3 +1,5 @@
+(** Various old solvers. *)
```
michael-schwarz, line 1: True?

src/witness/argTools.ml
```diff
@@ -1,3 +1,5 @@
+(** Construction of ARGs from constraint system solutions. *)
```
michael-schwarz, line 1:
```suggestion
(** Construction of  Abstract Reachability Graphs (ARGs) from constraint system solutions. *)
```

src/witness/myARG.ml
```diff
@@ -1,3 +1,5 @@
+(** Abstract reachibility graph. *)
```
michael-schwarz, line 1:
```suggestion
(** Abstract reachability graph. *)
```

michael-schwarz (changes requested):

src/analyses/tutorials/constants.ml
sim642: This isn't even a tutorial, but just an example that you guys added at some point.

src/framework/edge.ml
sim642: There's another fairly common CFG representation where statements are in nodes (not edges) and states are moving between them via edges (not held in nodes).

We need to make this very clear because our nodes are also identified by a CIL statement ID, but we don't use such a CFG representation.

src/framework/varQuery.mli
sim642: Since these are for manual restarting, they're things which the user could reference in a program, e.g. variables, types, program points, as opposed to some Goblint-internal things.

src/solvers/generic.ml
sim642: Well, most of this file is some old things like `DirtyBoxSolver`. I'm not sure what they exactly are.

src/analyses/threadReturn.ml
```diff
@@ -1,4 +1,4 @@
-(** Thread returning analysis. *)
+(** Thread returning analysis using boolean call stack ([threadreturn]). *)
```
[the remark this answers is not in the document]
michael-schwarz: Yes, I read boolean call stack as "a callstack with booleans as stack elements", but there is always only one boolean. Maybe we can find a way to say that?

src/analyses/tutorials/constants.ml
michael-schwarz: Ah, right.

src/solvers/generic.ml
michael-schwarz: `LoadRunIncrSolver` is used for Gobview, and the `SolverStats` functor is used to get solver statistics. I'd argue these two are more important than `DirtyBoxSolver` and the other crap, so I think we should mention them in the summary.

src/analyses/threadReturn.ml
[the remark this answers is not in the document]
sim642: I added the longer explanation there now.

src/solvers/generic.ml
sim642: Ideally we wouldn't have such a generically named module as `Generic` (which seemingly has nothing to do with solvers) at all. `LoadRunIncrSolver` should be separate and `SolverStats` functor could be in the newly introduced `SolverStats` file module, but I won't refactor the code itself here, just expanded the synopsis.

PR #1063, opened 2023-05-26
Add CIL's ptranal for `EvalFunvar` queries

vesalvojdani:
This will add the option to use Cil's Ptranal module to deal with dynamic function calls. This would avoid having to run the base analysis for fast analyses. The problem is that the Ptranal module's points to queries are really literal, so `*fp` (which is how function pointer calls are normalized by CIL) does not point to anything, you have to remove the dererefence and query `fp`, and then for cases like calling through pointers within structs, the analysis is flow-insensitive and that causes some issues.

- [ ] This is currently implemented as a wrapper / unit analysis to answer the queries. Should decide if this rather belongs properly to the base functionality of the framework. Depends on how much code is still needed to handle things.
- [ ] The expressions fed to Ptranal is still not very robust: there are dead code on some bench repo examples.
- [ ] Filter compatible function calls since the pointer analysis is flow-insensitive.


src/analyses/ptranalEvalFunvar.ml
```diff
@@ -0,0 +1,59 @@
...
+    [ctx.local, ctx.local]
+
+  let combine_env ctx lval fexp f args fc au f_ask =
+    au
+
+  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask: Queries.ask) : D.t =
+    ctx.local
+
+  let special ctx (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    ctx.local
```
sim642, line 38:
```ocaml
include UnitAnalysis.Spec
```

michael-schwarz:
What are our plans with this?

sim642:
Since Goblint CIL still has `Ptranal` and this Goblint "analysis" is very simple, there's no reason to not have this around for some modular/base-less analysis experiments.

michael-schwarz:
> Since Goblint CIL still has Ptranal and this Goblint "analysis" is very simple, there's no reason to not have this around for some modular/base-less analysis experiments.

I think in this case it would be a good idea to address the open remarks and then merge this.

michael-schwarz:
I have now brought this up-to-date with master. Should we merge it as-is or address the comments above before?

michael-schwarz:
The failing indentation is already on master.

sim642 (approved):
I made a few small changes to make this merge-ready.

PR #1065, opened 2023-05-30
Also allow `threadid` to recover from multi-threaded mode

michael-schwarz:
Currently, only the `thread` analysis can recover from multi-threaded mode. This allows also the `threadid` and `threadJoins` analysis to achieve this.

To this end, `threadJoins` is equipped with a notion of cleanly joined:

- Locally, this means that for all threads that have been must-joined, all threads transitively created have also been joined.
- At a global unknown for the tid, this means that all threads that have been created by the current thread have been joined.

This is then used to answer the `MustBeSingleThreaded {since_start = false}` queries. 

Follow-up to #1049


michael-schwarz:
The problem is that `created` returns only the unique TIDs that have been created, not all of them.

michael-schwarz:
I have now modified the `CreatedThreads` query to not just return unique threads, but also non-unique threads that have been created.

sim642:
> To this end, `threadJoins` is equipped with a notion of cleanly joined:
> 
> * Locally, this means that for all threads that have been must-joined, all threads transitively created have also been joined.
> 
> * At a global unknown for the tid, this means that all threads that have been created by the current thread have been joined.
> 
> 
> This is then used to answer the `MustBeSingleThreaded {since_start = false}` queries.

Isn't that precisely what the old `thread` analysis is already doing? Or how are they different?
Because we shouldn't be duplicating an analysis.

michael-schwarz:
One might even wonder if with this addition, our new analyses subsume `thread`: After allowing for a path-sensitive `threadid` analysis, all tests for the `thread` analysis pass when enabling `threadJoins`, and turning `threadId` path-sensitive whenever `thread` is.

sim642:
> One might even wonder if with this addition, our new analyses subsume `thread`: After allowing for a path-sensitive `threadid` analysis, all tests for the `thread` analysis pass when enabling `threadJoins`, and turning `threadId` path-sensitive whenever `thread` is.

Something seems amiss here. Our thread ID domains don't even define a sensible `join`, but are simply flat or something. The only reason we got away with it was because the thread's ID doesn't change and the analysis is context-sensitive. If any joining of thread IDs were to happen, there would just be top thread IDs.

sim642 (approved):

PR #1067, opened 2023-05-30
Refactor offsets, lvals and addresses

sim642:
Everything is organized into three modules:
1. `Offset` contains everything for manipulating Goblint offsets with various index types.
2. `Mval` contains everything for manipulating `varinfo * offset` pairs.
3. `AddressDomain` contains everything for manipulating base analysis addresses.

All three now have interface files which makes rebuild times negligible when only modifying the implementation (as opposed to recompiling ~200 modules which transitively depend on them).
Many functions are renamed to make naming consistent and standard.


sim642:
> It may also be worth trying if we can get rid of the polymorphic variants for offsets?

That's a future idea I have, but it requires modifying so many places that I don't want to do it right now.
With organized type signatures in place, it should become doable without tons of module prefixes. Especially because we open CIL in so many places and it has constructors with the same names.

src/cdomains/mval.ml
```diff
@@ -0,0 +1,74 @@
...
+  let narrow = merge Offs.narrow
+
+  include Lattice.NoBotTop
+
+  let pretty_diff () (x,y) =
+    Pretty.dprintf "%s: %a not equal %a" (name ()) pretty x pretty y
+end
+
+module Unit = MakePrintable (Offset.Unit)
+module Exp = MakePrintable (Offset.Exp)
```
michael-schwarz, line 74: I think naming these `UnitOffset` and `ExpOffset` or `UnitO` and `ExpO` makes it more intuitive what they do.

src/cdomains/mval.ml
```diff
@@ -0,0 +1,74 @@
...
+  let prefix (v1,ofs1) (v2,ofs2) =
+    if CilType.Varinfo.equal v1 v2 then
+      Offs.prefix ofs1 ofs2
+    else
+      None
+
+  let to_cil ((v, o): t): lval = (Var v, Offs.to_cil o)
+  let to_cil_exp lv = Lval (to_cil lv)
+
+  let is_definite (_, o) = Offs.is_definite o
```
michael-schwarz, line 38: Maybe call `is_definite_offset`? This will return true also for heap-allocated things etc which might be confusing.

src/cdomains/offset_intf.ml
```diff
@@ -0,0 +1,113 @@
+type 'i t = [
+  | `NoOffset
+  | `Field of CilType.Fieldinfo.t * 'i t
+  | `Index of 'i * 'i t
+] [@@deriving eq, ord, hash]
+
+(* TODO: remove? *)
+type 'i offs = 'i t [@@deriving eq, ord, hash]
```
michael-schwarz, line 8: What is the plan with this TODO? Do we want to do it or not?

src/analyses/base.ml
```diff
@@ -404,20 +404,12 @@ struct
...
-  (* Auxiliary function to append an additional offset to a given offset. *)
-  let rec add_offset ofs add =
-    match ofs with
-    | `NoOffset -> add
-    | `Field (fld, `NoOffset) -> `Field (fld, add)
-    | `Field (fld, ofs) -> `Field (fld, add_offset ofs add)
-    | `Index (exp, `NoOffset) -> `Index (exp, add)
-    | `Index (exp, ofs) -> `Index (exp, add_offset ofs add)
-
+  (* TODO: move to Lval *)
```
michael-schwarz, line 407: Where exactly is this supposed to be moved to?

src/analyses/base.ml
```diff
@@ -619,21 +611,9 @@ struct
...
-        | _ -> Lval.any_index_exp
-      in
-      match o with
-      | `NoOffset -> `NoOffset
-      | `Field (f,o) -> `Field (f,offsNormal o)
-      | `Index (i,o) -> `Index (toInt i,offsNormal o)
-    in
-    match x with
-    | ValueDomain.AD.Addr.Addr (v,o) ->[v,offsNormal o]
+  (* TODO: Lval *)
```
michael-schwarz, line 614: TODO

src/analyses/commonPriv.ml
```diff
@@ -125,20 +125,14 @@ struct
...
-
   let current_lockset (ask: Q.ask): Lockset.t =
     (* TODO: remove this global_init workaround *)
     if !AnalysisState.global_initialization then
       Lockset.empty ()
     else
       let ls = ask.f Queries.MustLockset in
       Q.LS.fold (fun (var, offs) acc ->
-          Lockset.add (Lock.from_var_offset (var, conv_offset offs)) acc
+          Lockset.add (Lock.of_mval (var, Lock.Offs.of_exp offs)) acc
```
michael-schwarz, line 135: As far as I can see it, this changes the semantics where we suddenly now have integer offsets in here?
Can we maybe postpone such major semantic changes to a new PR?

src/analyses/mutexEventsAnalysis.ml
```diff
@@ -18,17 +18,9 @@ struct
...
-
-  (* Currently we care only about concrete indexes. *)
-  let rec conv_offset x =
-    match x with
-    | `NoOffset    -> `NoOffset
-    | `Index (Const (CInt (i,_,s)),o) -> `Index (IntDomain.of_const (i,Cilfacade.ptrdiff_ikind (),s), conv_offset o)
-    | `Index (_,o) -> `Index (ValueDomain.IndexDomain.top (), conv_offset o)
-    | `Field (f,o) -> `Field (f, conv_offset o)
-
+  (* TODO: Lval *)
```
michael-schwarz, line 21: What remains here?

src/analyses/pthreadSignals.ml
```diff
@@ -17,15 +17,9 @@ struct
...
   module G = SetDomain.ToppedSet (MHP) (struct let topname = "All Threads" end)
 
-  let rec conv_offset x =
-    match x with
-    | `NoOffset    -> `NoOffset
-    | `Index (Const (CInt (i,_,s)),o) -> `Index (IntDomain.of_const (i,Cilfacade.ptrdiff_ikind (),s), conv_offset o)
-    | `Index (_,o) -> `Index (ValueDomain.IndexDomain.top (), conv_offset o)
-    | `Field (f,o) -> `Field (f, conv_offset o)
-
+  (* TODO: Lval *)
```
michael-schwarz, line 20: what exactly?

src/analyses/region.ml
```diff
@@ -82,12 +81,8 @@ struct
...
     | Memory {exp = e; _} ->
       (* TODO: remove regions that cannot be reached from the var*)
-      let rec unknown_index = function
-        | `NoOffset -> `NoOffset
-        | `Field (f, os) -> `Field (f, unknown_index os)
-        | `Index (i, os) -> `Index (Lval.any_index_exp, unknown_index os) (* forget specific indices *)
-      in
-      Option.map (Lvals.of_list % List.map (Tuple2.map2 unknown_index)) (get_region ctx e)
+      (* forget specific indices *)
+      Option.map (Lvals.of_list % List.map (Tuple2.map2 Offset.Exp.top_indices)) (get_region ctx e)
```
michael-schwarz, line 85: If we forget them here, do we not want to avoid storing them in the first place?

src/analyses/threadEscape.ml
```diff
@@ -22,16 +22,10 @@ struct
...
-    match x with
-    | `NoOffset    -> `NoOffset
-    | `Index (_,o) -> `NoOffset
-    | `Field (f,o) -> `Field (f, cut_offset o)
-
   let reachable (ask: Queries.ask) e: D.t =
     match ask.f (Queries.ReachableFrom e) with
     | a when not (Queries.LS.is_top a) ->
-      (* let to_extra (v,o) set = D.add (Addr.from_var_offset (v, cut_offset o)) set in *)
+      (* let to_extra (v,o) set = D.add (Addr.of_mval (v, cut_offset o)) set in *)
```
michael-schwarz, line 28: If we leave this here, we probably also should leave `cut_offset`. Otherwise it's hard to understand what this used to do.

src/analyses/uninit.ml
```diff
@@ -29,19 +29,11 @@ struct
...
-  (* NB! Currently we care only about concrete indexes. Base (seeing only a int domain
-     element) answers with Lval.any_index_exp on all non-concrete cases. *)
-  let rec conv_offset x =
-    match x with
-    | `NoOffset    -> `NoOffset
-    | `Index (Const (CInt (i,ik,s)),o) -> `Index (IntDomain.of_const (i,ik,s), conv_offset o)
-    | `Index (_,o) -> `Index (IdxDom.top (), conv_offset o)
-    | `Field (f,o) -> `Field (f, conv_offset o)
-
+  (* TODO: Lval *)
```
michael-schwarz, line 32: What?

src/analyses/uninit.ml
```diff
@@ -91,56 +83,42 @@ struct
...
-    CilType.Varinfo.equal v1 v2 && is_offs_prefix_of ofs1 ofs2
+    List.map Addr.of_mval (varoffs a rval)
 
+  let is_prefix_of m1 m2 = Option.is_some (Addr.Mval.prefix m1 m2)
 
   (* Does it contain non-initialized variables? *)
   let is_expr_initd a (expr:exp) (st:D.t) : bool =
     let variables = vars a expr in
-    let raw_vars = List.filter_map Addr.to_var_offset variables in
+    let raw_vars = List.filter_map Addr.to_mval variables in
```
michael-schwarz, line 93: `mvals`?

src/cdomains/addressDomain_intf.ml
```diff
@@ -0,0 +1,117 @@
...
+    val to_mval: t -> Mval.t option
+    (** Get the offset *)
+
+    val to_exp: t -> GoblintCil.exp
+
+    val type_of: t -> GoblintCil.typ
+    (** Finds the type of the address location. *)
+  end
+
+  (** Lvalue lattice.
```
michael-schwarz, line 50: Is this still an `Lvalue lattice`?

src/cdomains/addressDomain_intf.ml
```diff
@@ -0,0 +1,117 @@
...
+
+      Actually a disjoint union of lattices without top or bottom.
+      Lvalues are grouped as follows:
+
+      - Each {!Addr}, modulo precise index expressions in offset, is a sublattice with ordering induced by {!Offset}.
```
michael-schwarz, line 55: Not `Mval`?

src/cdomains/lvalMapDomain.ml
```diff
@@ -13,7 +13,7 @@ exception Error
 module type S =
```
michael-schwarz, line 13: Should this not be named `mvalMapDomain`?

src/cdomains/musteqDomain.ml
```diff
@@ -6,32 +6,82 @@ open Pretty
...
-    | (`Left x::xs) -> occurs v xs
-    | (`Right x::xs) -> I.occurs v x || occurs v xs
-    | [] -> false
+    | `Field (x, xs) -> occurs v xs
+    | `Index (x, xs) -> I.occurs v x || occurs v xs
+    | `NoOffset -> false
 end
 
-module EquAddr =
+(* TODO: Mval *)
```
michael-schwarz, line 73: Todo?

src/cdomains/symbLocksDomain.ml
```diff
@@ -302,9 +302,10 @@ struct
 
     let equal_to _ _ = `Top
     let to_int _ = None
+    let top () = Unknown
   end
 
-  include Lval.Normal (Idx)
+  include AddressDomain.AddressPrintable (Mval.MakePrintable (Offset.MakePrintable (Idx)))
 
   let rec conv_const_offset x =
```
michael-schwarz, line 310: Are we actually using precise offsets here anywhere?

michael-schwarz (commented):
Thank you for undertaking this gigantic refactoring endeavor. I left comments in a few places. Additionally I was wondering if we want to write something up for the docs while this is fresh on our minds.

The distinction between Mval and Lval really is vital to newcomers to the framework.

src/cdomains/offset_intf.ml
```diff
@@ -0,0 +1,113 @@
...
+
+  (** Special index expression for some unknown index.
+      Weakly updates array in assignment.
+      Used for exp.fast_global_inits. *)
+  val any_index_exp: GoblintCil.exp
+
+  (** Special index expression for all indices.
+      Strongly updates array in assignment.
+      Used for Goblint-specific witness invariants. *)
+  val all_index_exp: GoblintCil.exp
```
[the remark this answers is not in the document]
sim642: We also use them in CIL offsets, not just our exp-indexed ones, but I'll move them anyway since we already have a nice structure here.

src/analyses/base.ml
sim642: I've added longer explanations to the TODOs you've mentioned here. Mostly the idea is to also use `AddressDomain` for queries, such that all those lossy and hacky back-and-forth conversions could be avoided. But I don't want to make this PR bigger than it already is.

src/analyses/uninit.ml
sim642: `Addr.to_mval` still returns one optional mval.

michael-schwarz: Yes, I think i meant the `raw_vars` name on the left... I think what we call a `raw_var` here is precisely an `mval`.

src/analyses/commonPriv.ml
sim642: That was the entire point of the previous `(* TODO: better indices handling *)`. Our must-lockset analysis already tracks definite indices, so there's no reason to suddenly drop them. As far as privatizations are concerned, mutex names are treated opaquely, so `m[1]` is as different from `m[2]` as `m1` is from `m2` or `s1.m` from `s2.m`.

I actually find the old behavior more suspect because `current_lockset` would return `{m}` while events like unlocking would be for `m[2]`. That's odd because we use `Lockset.mem` and `Lockset.remove` in these privatizations and there are no semantic checks on whether `m[2]` and `m` are related/subsumed in any way.

This is another place where using `LockDomain` directly for the query could avoid all such conversions altogether.

src/cdomains/addressDomain_intf.ml
sim642: I just copied the old documentation. I've now updated it.

src/analyses/threadEscape.ml
sim642: These have been commented out since 2013, so I'll just remove them since not dropping index offsets doesn't seem to have been an issue in the last decade.

src/cdomains/mval.ml
sim642: That's true, although in `AddressDomain` `is_definite` isn't just about offsets again, so it would make for inconsistent naming. There we also consider `NullPtr` to be definite. So it seems that the semantics of this check has never been very well defined.

src/cdomains/mval.ml
sim642: I think `UnitOffset` is even more misleading because that sounds like it's just for offsets, while this is not. Since these modules are supposed to be used in context, I think the prefix makes it clear what it is, e.g. `Mval.Unit`, `Offset.Unit` or `Offset.Index.Unit`. To make it fully explicit in the submodule name, it'd have to be `Mval.WithUnitIndexOffset` but that's way too verbose.

I will add more documentation to these things though to make it clear that this module is for mvalues with offsets that use `exp` indices.

sim642:
> Additionally I was wondering if we want to write something up for the docs while this is fresh on our minds.
> 
> The distinction between Mval and Lval really is vital to newcomers to the framework.

I added a few sentences to the top level API docs. That way it hopefully stays better in sync.

src/analyses/region.ml
sim642: Maybe, I just added a TODO about it instead of refactoring the region analysis more deeply.
As noted in #107, region analysis uses `exp`s for some reason, although maybe `lval`s (or now our `mval`s) could actually suffice. In that case one could easily use `Mval.Unit`.

src/analyses/uninit.ml
sim642: I looked at this and simplified it even more: this contained pointless back-and-forth conversions which fuse to identity.

src/cdomains/offset_intf.ml
sim642: At some point I thought it could be removed but actually it cannot be. Some inner signatures define their own `type t` but need to refer to this outer `t`, so `offs` is the alias necessary to do that.
I've replaced the TODO with this note.

michael-schwarz (approved):

PR #1068, opened 2023-05-31
Remove polymorphic variants for VD.Compound

michael-schwarz:


src/analyses/baseInvariant.ml
```diff
@@ -547,12 +547,12 @@ struct
...
     in
     let eval e st = eval_rv a gs st e in
-    let eval_bool e st = match eval e st with `Int i -> ID.to_bool i | _ -> None in
+    let eval_bool e st = match eval e st with Int i -> ID.to_bool i | _ -> None in
     let rec inv_exp c_typed exp (st:D.t): D.t =
       (* trying to improve variables in an expression so it is bottom means dead code *)
       if VD.is_bot_value c_typed then contra st else
       match exp, c_typed with
-      | UnOp (LNot, e, _), `Int c ->
+        | UnOp (LNot, e, _), Int c ->
```
sim642, line 555: Why are all these cases indented now? There's no `begin`..`end` around the `match`, so this shouldn't be necessary. Also the bodies of the cases aren't indented deeper, so it's all off now.

The indentation check can be wrong about these since we only ask it to check and indent modified lines in isolation. So it can make suggestions which it wouldn't if the whole file is reindented.

sim642 (approved):

src/analyses/baseInvariant.ml
michael-schwarz: Yeah, that is indeed strange. I'll fix indentation manually.

PR #1069, opened 2023-06-01
Improve mallocFresh analysis

karoliineh:
Some false positive warnings from [the-silver-searcher](https://github.com/goblint/bench/tree/master/gobpie-demos/silver-searcher) could be removed with a more precise `mallocFresh` analysis. The changes I have done are:

1. Instead of a set of only fresh variables, also track the set of not fresh variables in the domain. That is because previously the domain was a reversed set domain of fresh variables, in which case if the variable was fresh in one branch, but unassigned in another, it was considered not fresh. Also added two test cases for that scenario.
2. Using `Escape` events. Previously all fresh variables became not fresh once anything escaped. The analysis now tracks whether the escaped variables are the same as the fresh variables.
3. Made the `Escape` events return the `escaped_to` variables in addition to `escaped` variables, to track whether a variable escaped into itself and thus remained fresh.

Note: I noticed that there is no `combine_assign` in `threadEscape`. Why is that? Should there be one?

### DIFF
--- a/src/analyses/mallocFresh.ml
+++ b/src/analyses/mallocFresh.ml
@@ -6,21 +6,35 @@ module Spec =
 struct
   include Analyses.IdentitySpec
 
-  (* must fresh variables *)
-  module D = SetDomain.Reverse (SetDomain.ToppedSet (CilType.Varinfo) (struct let topname = "All variables" end)) (* need bot (top) for hoare widen *)
+  (* must fresh (or may not fresh) variables *)
+  module DF = SetDomain.ToppedSet (CilType.Varinfo) (struct let topname = "All variables" end) (* need bot (top) for hoare widen *)
+  module D =
+  struct 
+    include Lattice.Prod (DF) (DF)
+
+    let join (df1, dnf1) (df2, dnf2) =
+      let dnf = DF.union dnf1 dnf2 in
+      let df = DF.diff (DF.union df1 df2) dnf in
+      (df, dnf)
+
+    let leq x y = equal (join x y) y
+
+    let widen = join
+    let narrow a b = a
+  end
   module C = D
 
   let name () = "mallocFresh"
 
-  let startstate _ = D.empty ()
-  let exitstate _ = D.empty ()
+  let startstate _ = (DF.empty (), DF.empty ())
+  let exitstate _ = (DF.empty (), DF.empty ())
 
   let assign_lval (ask: Queries.ask) lval local =
-    match ask.f (MayPointTo (AddrOf lval)) with
-    | ls when Queries.LS.is_top ls || Queries.LS.mem (dummyFunDec.svar, `NoOffset) ls ->
-      D.empty ()
-    | ls when Queries.LS.exists (fun (v, _) -> not (D.mem v local) && (v.vglob || ThreadEscape.has_escaped ask v)) ls ->
-      D.empty ()
+    match ask.f (MayPointTo (AddrOf lval)), local with
+    | ls, (df, dnf) when Queries.LS.is_top ls || Queries.LS.mem (dummyFunDec.svar, `NoOffset) ls ->
+      (DF.empty (), DF.union df dnf)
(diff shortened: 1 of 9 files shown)


src/analyses/threadEscape.ml
```diff
@@ -75,14 +75,16 @@ struct
     else
       ctx.local
 
+  (* TODO: why is there no combine_assign? *)
+
```
michael-schwarz, line 79: You are right that it probably would to have handling similar to `assign`, rather than identity from `IdentitySpec`.

src/analyses/mallocFresh.ml
```diff
@@ -6,34 +6,38 @@ module Spec =
...
+      let df_both = DF.meet df1 df2 in
+      let dnf_any = DF.union dnf1 dnf2 in
+      let df_one = DF.diff (DF.union df1 df2) dnf_any in
+      (DF.union df_both df_one, dnf_any)
+
+    let leq x y = equal (join x y) y
+
+    let widen = join
+    let narrow a b = a
+  end
```
michael-schwarz, line 25: I don't think I completely understand this yet... What actually are the components here?

I guess the first one is supposed to be the set of must fresh variables. But what exactly is the second one?

Also, why do we not need to redefine the meet here? Does it just happen to be compatible with the new leq and join?

src/analyses/mallocFresh.ml
```diff
@@ -42,19 +46,26 @@ struct
...
-    D.empty ()
+    ctx.local
+
+  let event ctx e octx =
+    let local = ctx.local in
+    match e, local with
+    | Events.Escape {escaped; escaped_to}, (df, dnf) when DF.is_top escaped ->
+      (DF.empty (), DF.union df dnf)
+    | Events.Escape {escaped; escaped_to}, (df, dnf) when not (DF.subset escaped_to df) ->
+      (DF.diff df escaped, DF.union dnf escaped)
```
michael-schwarz, line 66: Do we need to add any things that aren't malloced to the domain here?

src/analyses/mallocFresh.ml
sim642: I think the second is the set of variables which may not be fresh, i.e. they have become unfresh by escaping.

`meet` hasn't been redefined because nothing uses it if the analysis itself does not. But it's most likely wrong, so at least should be replaced with
```ocaml
let meet _ _ = assert false
```
Also the `leq` is very naive. I think one could figure out what it should be directly.

src/analyses/threadEscape.ml
```diff
@@ -63,10 +63,10 @@ struct
     if M.tracing then M.tracel "escape" "assign vs: %a\n" D.pretty vs;
     if D.exists (fun v -> v.vglob || has_escaped ask v) vs then (
       let escaped = reachable ask rval in
-      let escaped = D.filter (fun v -> not v.vglob) escaped in
+      let escaped = D.filter (fun v -> not v.vglob || ctx.ask (IsHeapVar v)) escaped in
```
sim642, line 66: The potential thing to be extracted from this PR before closing is this (and similar changes below). It also considers heap variables to be escaped, besides just locals.
It's not clear whether this was some soundness fix, without a corresponding test, or something specific to this PR. Maybe it was specific to other changes which are about heap memory.

Also, since this, there's now an `IsAllocVar` query, which might be more appropriate if this still has any purpose.

sim642: We thought about it for a while and concluded that such extra query is not necessary to fix anything on `master`.
It was introduced in a4ac9c765b44111c2f739471e44d694c28fcf345 only because the `event` handler added to the new mallocFresh cared about alloc variables having escaped. Since the current version doesn't care about the event at all, this makes no difference.
The escape analysis itself already behaves correctly with allocated memory and we have regression tests for it.

michael-schwarz: I guess this means the PR can be closed?

sim642: Yup, @karoliineh will still add `combine_assign` to the escape analysis, but that's a separate PR, nothing to reuse from here.

PR #1071, opened 2023-06-01
Ignore accesses to variables with `__thread` keyword

karoliineh:
Relates to #876.

### DIFF
--- a/src/domains/access.ml
+++ b/src/domains/access.ml
@@ -20,6 +20,7 @@ let is_ignorable_type (t: typ): bool =
 
 let is_ignorable = function
   | None -> false
+  | Some (v,os) when hasAttribute "thread" v.vattr -> true (* Thread-Local Storage *)
   | Some (v,os) ->
     try isFunctionType v.vtype || is_ignorable_type v.vtype
     with Not_found -> false
--- a/tests/regression/04-mutex/82-thread-local-storage.c
+++ b/tests/regression/04-mutex/82-thread-local-storage.c
@@ -1,0 +1,23 @@
+#include <pthread.h>
+#include <stdio.h>
+
+__thread int myglobal;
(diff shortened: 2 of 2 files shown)


sim642 (approved):

michael-schwarz:
I think this is a bit overzealous in claiming race-freedom and thus sadly unsound.

Consider, e.g., the following program where the address of the global escapes to another thread.

~~~C
#include <pthread.h>
#include <stdio.h>

__thread int myglobal;

void *t_fun(void *arg) {
  int* ptr = (int*) arg;
  pthread_mutex_lock(&mutex1);
  *ptr=*ptr+1; //RACE
  pthread_mutex_unlock(&mutex1);
  return NULL;
}

int main(void) {
  pthread_t id;
  pthread_create(&id, NULL, t_fun, (void*) &myglobal);
  pthread_mutex_lock(&mutex2);
  myglobal=myglobal+1; //RACE
  pthread_mutex_unlock(&mutex2);
  pthread_join (id, NULL);
  return 0;
}
~~~

The GCC manual says the following:

> When the address-of operator is applied to a thread-local variable, it is evaluated at run time and returns the address of the current thread’s instance of that variable. An address so obtained may be used by any thread. When a thread terminates, any pointers to thread-local variables in that thread become invalid.

(from https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html)

So I don't think there's anything forbidding code like this. Maybe we should only claim race-freedom if the global does not have its address taken for now? (Or do the proper thing and do escape tracking for such globals).

sim642:
That's a good catch!

@karoliineh If the ignore is only limited to when `v.vaddrof` is false, is this still useful for silver searcher?

karoliineh:
> @karoliineh If the ignore is only limited to when `v.vaddrof` is false, is this still useful for silver searcher?

Yes, still removes 10 warnings.

sim642:
You can then add that fix to `master`.

PR #1075, opened 2023-06-06
`ThreadEscape`: Collect flow-insensitive set of variables reachable from globals

### LINKED ISSUE #1074: Escaped variables are not communicated between threads
In [this test case](
https://github.com/goblint/analyzer/blob/8dcc0b00406839ce0782e52132f91a4ec06a7629/tests/regression/74-escape/01-local-esacpe.c), Goblint, with the default configuration, does not relax information about the escaped local variables `x`, even though another thread may be writing into it. In particular, Goblint unsoundly claims that the second `__goblint_check` will succeed:

https://github.com/goblint/analyzer/blob/8dcc0b00406839ce0782e52132f91a4ec06a7629/tests/regression/74-escape/01-local-esacpe.c#L9-L17

jerhard:
In #1074 it was discovered that in the analysis it is not communicated to a thread that variable may escape (via a global), after the thread has been created.
Previously, the analysis was adding variables that may become reachable via the argument passed to `pthreadcreate`. This does not take into account that the escaping after the thread creation may also happen via global variables.
This PR collects variables that are assigned to global variables separately under a flow-insensitive variable `global_var`.
In the `threadenter`, the set of variables escaped to globals, that was collected flow-insensitively, is then joined into the local state.

This increases the imprecision of the escape analysis, due to the flow-insensitive nature of the handling of variables escaped via globals. Such variables will be considered escaped even in unique threads before they have actually escaped. 
   
Fixes #1074.

### DIFF
--- a/src/analyses/threadEscape.ml
+++ b/src/analyses/threadEscape.ml
@@ -61,6 +61,8 @@ struct
       D.mem v ctx.local
     | _ -> Queries.Result.top q
 
+  let global_var = Cil.dummyFunDec.svar
+
   (* transfer functions *)
   let assign ctx (lval:lval) (rval:exp) : D.t =
     let ask = Analyses.ask_of_ctx ctx in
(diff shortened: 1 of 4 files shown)


michael-schwarz:
Two ideas:

- Can we maybe add the thread id of the thread that caused it to escape? Then we would still be able to consider things as not escaped when it is known that that has not happened yet.
- is `threadenter` ever called for the main thread, i.e., does it also see these things that escaped?

jerhard:
Closed in favor of #1078

PR #1076, opened 2023-06-06
Null Byte Array Domain

nathanschmidt:
This PR implements a Null Byte Array Domain tracking all indexes of array elements thay *may* be null and *must* be null. Additionally, it also memorizes the size of the array. It is always run as part of the `AttributeConfiguredArrayDomain`, which allowed an easy extension of the cases for the functions listed above in the base analysis.

This domain makes it possible to gain meaningful information for strings, especially concerning the effects of the functions `strcpy`, `strncpy`, `strcat`, `strncat`, `strlen`, `strstr`, `strcmp` and `strncmp`.  On top, the domain tries to identify possible and sometimes even certain buffer overflows.

**TODOs:**
- [x] map all array elements that are read-in to `null ()` resp. `not_null ()`
- [x] tackle TODO annotations in code
- [x] write test cases for both sound uses of the functions and scenarios for which warnings should be triggered


src/analyses/base.ml
```diff
@@ -2043,24 +2030,66 @@ struct
...
+        begin match get (Analyses.ask_of_ctx ctx) gs st s1_a None, get (Analyses.ask_of_ctx ctx) gs st s2_a None with
+          | Array array_s1, Array array_s2 -> set ~ctx (Analyses.ask_of_ctx ctx) gs st lv_a lv_typ (op_array array_s1 array_s2)
+          | Array array_s1, _ when s2_typ = charPtrType -> 
+            let s2_null_bytes = List.map CArrays.to_null_byte_domain (AD.to_string s2_a) in
+            let array_s2 = List.fold_left CArrays.join (CArrays.bot ()) s2_null_bytes in
+            set ~ctx (Analyses.ask_of_ctx ctx) gs st lv_a lv_typ (op_array array_s1 array_s2)
+          | _, Array array_s2 when s1_typ = charPtrType ->
+            (* if s1 is string literal, str(n)cpy and str(n)cat are undefined *)
+            if op_addr = None then
+              let _ = AD.string_writing_defined s1_a in
```
michael-schwarz, line 2085: I think it might be worth commenting why we ignore the result here: Function is evaluated for its side-effects only.

src/cdomains/arrayDomain.ml
```diff
@@ -39,15 +39,12 @@ let get_domain ~varAttr ~typAttr =
 
 let can_recover_from_top x = x <> TrivialDomain
 
-module type S =
+module type SMinusDomainAndRet =
```
michael-schwarz, line 42: Why `SMinusDomainAndRet`? You are removing `get`, right? Also maybe just naming it `S0` or something like this is more transparent.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +994,746 @@ struct
...
+    | None -> None
+
+  let get ?(checkBounds=true) (ask: VDQ.t) (must_nulls_set, may_nulls_set, size) (e, i) =
+    let rec all_indexes_must_null i max =
+      if Z.gt i max then
+        true
+      else if MustNulls.mem i must_nulls_set then
+        all_indexes_must_null (Z.succ i) max
+      else
+        false in
```
michael-schwarz, line 1022: This looks expensive with all the `mem` queries. Maybe we can do a cardinality based thing first?

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +994,746 @@ struct
...
+    let min interval = match Idx.minimal interval with
+      | Some min_num when Z.geq min_num Z.zero -> min_num
+      | _ -> Z.zero in (* assume worst case minimal natural number *)
+
+    let min_i = min i in
+    let max_i = idx_maximal i in
+    let min_size = min size in
+
+    (* warn if index is (potentially) out of bounds *)
+    if checkBounds then (array_oob_check (module Idx) ((must_nulls_set, may_nulls_set), size) (e, i));
```
michael-schwarz, line 1032: Is this needed given that you always have another array domain that is active?

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +994,746 @@ struct
...
+    match max_i, idx_maximal size with
```
michael-schwarz, line 1033: Is it maybe easier to understand if you pull all the cases where accesses are out of bounds and we thus return top out and the only deal with the well-defined cases in here?

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +994,746 @@ struct
...
+  let must_nulls_remove i must_nulls_set min_size =
+    let rec compute_set acc i =
+      if Z.geq i min_size then 
+        acc
+      else
+        compute_set (MustNulls.add i acc) (Z.succ i) in
+    if MustNulls.is_bot must_nulls_set then
+      MustNulls.remove i (compute_set (MustNulls.empty ()) Z.zero)
+    else
+      MustNulls.remove i must_nulls_set
```
michael-schwarz, line 1074: Please add a linebreak after this function (and similar below).

src/cdomains/arrayDomain.ml
michael-schwarz, line 1: Do you have access to the coverage reports generated by the CI? These might be helpful to see if all cases are triggered (as there's quite the list of cases). If not, try to login into the coverage website with GitHub. If that doesn't help, ping me, and I'll promote your access rights.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +994,746 @@ struct
...
+  let must_nulls_filter cond must_nulls_set min_size = 
+    let rec compute_set acc i =
+      if Z.geq i min_size then 
+        acc
+      else
+        compute_set (MustNulls.add i acc) (Z.succ i) in
```
michael-schwarz, line 1080: This should be possible to do more concisely, e.g. `List.init` and then `MustNulls.from_list`

src/cdomains/arrayDomain.ml
michael-schwarz: Also maybe you can put these into a new module (e.g. `IndexMustMaySet`) to encapsulate this more?

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +994,746 @@ struct
...
+    let set_exact i =
+      match idx_maximal size with
+      (* if size has no upper limit *)
+      | None -> 
+        (* ..., i < minimal size and value = null, add i to must_nulls_set and may_nulls_set *)
+        if Z.lt i min_size && Val.is_null v then
+          (MustNulls.add i must_nulls_set, MayNulls.add i may_nulls_set, size)
+        (* ..., i < minimal size and value <> null, remove i from must_nulls_set and may_nulls_set *)
+        else if Z.lt i min_size then
+          (must_nulls_remove i must_nulls_set min_size, MayNulls.remove i may_nulls_set, size)
```
michael-schwarz, line 1139: Why can you remove from `ManyNulls` here?

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +994,746 @@ struct
...
+      | _ -> Z.zero in (* assume worst case minimal natural number *)
+
+    let min_size = min size in
+    let min_i = min i in
+    let max_i = idx_maximal i in
+
+    let set_exact i =
+      match idx_maximal size with
+      (* if size has no upper limit *)
+      | None ->
```
michael-schwarz, line 1133: You explained to me at some point why `Val.is_not_null` is not needed, but I am now confused about this again: It seems there are three cases:

- Value is definitely `\0`
- Value is definitely not `\0`
- Value may be `\0` or not

I don't quite follow why you only need a boolean function here...

src/cdomains/arrayDomain.ml
michael-schwarz: I don't think `value <> null` is a valid conclusion here, as discussed above.

src/cdomains/arrayDomain.ml
nathanschmidt: You're right, we can simply rely on the other array domain for that.

src/cdomains/arrayDomain.ml
nathanschmidt: Is this what you have in mind?

```ocaml
let all_indexes_must_null i max =
      let rec check_all_indexes i =
        if Z.gt i max then
          true
        else if MustNulls.mem i must_nulls_set then
          check_all_indexes (Z.succ i)
        else
          false in
      if Z.lt (Z.sub max i) (Z.of_int (MustNulls.cardinal must_nulls_set)) then
        false
      else
        check_all_indexes i
```

michael-schwarz: yup

src/cdomains/arrayDomain.ml
nathanschmidt: I personally don't think it makes things much easier to read, but I'm biased of course. Do you find this function difficult to understand in its current form?

src/cdomains/arrayDomain.ml
nathanschmidt: Nope I don't see the reports. What website do you mean?

src/cdomains/arrayDomain.ml
nathanschmidt: Ah, I think I messed up because I implemented the domain before implementing the `is_null` function in `Compound` of *valueDomain*, assuming that `is_null` returning `false` would mean it's definitely not `\0`. For my assumption, no `is_not_null` was needed. Of course, that changes things. I'll add a `is_not_null` function after all.

src/analyses/base.ml
```diff
@@ -2069,24 +2056,67 @@ struct
...
+            (* check if s1 is potentially a string literal as writing to it would be undefined behavior; then return top *)
+            let _ = AD.string_writing_defined s1_a in
+            set ~ctx (Analyses.ask_of_ctx ctx) gs st s1_a s1_typ (VD.top_value (unrollType s1_typ))
+        end
+      (* else compute value in array domain *)
+      else 
+        let lv_a, lv_typ = match lv with
+          | Some lv_val -> eval_lv (Analyses.ask_of_ctx ctx) gs st lv_val, Cilfacade.typeOfLval lv_val
+          | None -> s1_a, s1_typ in
+        begin match get (Analyses.ask_of_ctx ctx) gs st s1_a None, get (Analyses.ask_of_ctx ctx) gs st s2_a None with
```
michael-schwarz, line 2102: Calling this for blobs lead to a problem. @michael-schwarz

michael-schwarz:
TODOs:

- Invalidate upon string function should go to top of type and not supertop

src/analyses/base.ml
```diff
@@ -2075,6 +2075,7 @@ struct
       let s2_a = address_from_value s2_v in
       let s2_typ = AD.type_of s2_a in
       (* compute value in string literals domain if s1 and s2 are both string literals *)
+      (* TODO: comparing types structurally should not be done (use typSig instead!) *)
```
michael-schwarz, line 2078: @nathanschmidt Please compare `typeSig` here instead of types; in this specific case it is fine, but in general types can be cyclic.

src/analyses/base.ml
```diff
@@ -2113,7 +2128,8 @@ struct
               let s1_null_bytes = List.map CArrays.to_null_byte_domain (AD.to_string s1_a) in
               let array_s1 = List.fold_left CArrays.join (CArrays.bot ()) s1_null_bytes in
               set ~ctx (Analyses.ask_of_ctx ctx) gs st lv_a lv_typ (op_array array_s1 array_s2)
-          | _ -> set ~ctx (Analyses.ask_of_ctx ctx) gs st lv_a lv_typ (VD.top_value (unrollType lv_typ))
+          | vals1, _ ->
```
michael-schwarz, line 2131: `vals1` is not needed here, sorry about that.

tests/regression/73-strings/08-cursed.c
```diff
@@ -0,0 +1,32 @@
...
+    __goblint_check(len2 == 5);
+
+    strcpy(s6,s5);
+    int len3 = strlen(s6);
+    __goblint_check(len3 == 5);
+
+    // Why does this not know the string length after the copy?
+    // This goes into the array/array case, so it seems unrelated to blob problems.
+    strcpy(s5, "badabingbadaboom");
+    len2 = strlen(s5); // no must 0 bytes anywhere?
```
michael-schwarz, line 29: Any idea why this does not work? I think it should?!

michael-schwarz: Actually, the problem is likely the weak updates for blobs :/

src/analyses/base.ml
sim642: `CilType.Typ.equal` is fine too. Possibly even better because it doesn't allocate visitor objects on each call.

tests/regression/73-strings/08-cursed.c
michael-schwarz: Should be fixed now.

src/analyses/base.ml
```diff
@@ -2093,16 +2094,30 @@ struct
...
+            let empty_array = CArrays.make (ID.cast_to (Cilfacade.ptrdiff_ikind ()) s_id) (Int (ID.top_of IChar)) in
+            set ~ctx (Analyses.ask_of_ctx ctx) gs st lv_a lv_typ (op_array empty_array array_s2)
+          | Bot , _ when s2_typ = charPtrType ->
+            (* If we have bot inside here, we assume the blob is used as a char array and create one inside *)
+            let size = ctx.ask (Q.BlobSize s1) in
+            let s_id = ValueDomainQueries.ID.unlift (Fun.id) size in
+            let empty_array = CArrays.make (ID.cast_to (Cilfacade.ptrdiff_ikind ()) s_id) (Int (ID.top_of IChar)) in
+            let s2_null_bytes = List.map CArrays.to_null_byte_domain (AD.to_string s2_a) in
+            let array_s2 = List.fold_left CArrays.join (CArrays.bot ()) s2_null_bytes in
+            set ~ctx (Analyses.ask_of_ctx ctx) gs st lv_a lv_typ (op_array empty_array array_s2)
```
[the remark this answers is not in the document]
nathanschmidt: When running Goblint now on some programs of the coreutils, it often crashes because of the exception `Queries.ID.unlift`, I suppose that is related to `s_id`.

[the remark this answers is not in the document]
michael-schwarz: Yes, maybe we can handle this by catching and creating an `ID.top ()`? But still weird that `ValueDomainQueries.ID` would not be an `ID.t` here...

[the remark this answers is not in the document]
michael-schwarz: Might make sense to log for which element we call Blobsize here?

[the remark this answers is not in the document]
michael-schwarz: Or do you think this could be the same issue you had with blobs @mrstanb ?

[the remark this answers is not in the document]
mrstanb: > Or do you think this could be the same issue you had with blobs @mrstanb ?

I'm not entirely sure what's going on here, but I think it could be a similar issue. In my case `BlobSize` was always giving `top` before we introduced the changes for `BlobSize`. Might be worth checking if this PR's branch is up-to-date with the `BlobSize` changes that should already be on `master`?

In this case `unlift` is probably complaining due to `bot` or `top`. If `BlobSize` is giving `bot` as a result, it might be worth checking if there's some offset in the abstract address set. In that case `BlobSize` will answer with `bot`.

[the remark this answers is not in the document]
michael-schwarz: @nathanschmidt Were you able to take a look? If yes, I could do a final review tomorrow and then hopefully merge this.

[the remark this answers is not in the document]
nathanschmidt: @michael-schwarz The branch is up to date with master. As you suggested, I caught the exception and created a top element. I suppose it's not ideal and should be fixed in the future, but it does the trick for now. I've also added some further regression tests after looking at coverage reviews, so feel free to review tomorrow!

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
 
-module AttributeConfiguredArrayDomain(Val: LatticeWithSmartOps) (Idx:IntDomain.Z):S with type value = Val.t and type idx = Idx.t =
+module HelperFunctionsIndexMustMaySets =
+struct
+  module MustSet = SetDomain.Reverse (SetDomain.ToppedSet (IntDomain.BigInt) (struct let topname = "All indexes" end))
+  module MaySet = SetDomain.ToppedSet (IntDomain.BigInt) (struct let topname = "All indexes" end)
+
+  let compute_set len =
+    List.init (Z.to_int len) (Fun.id)
+    |> List.map Z.of_int
```
michael-schwarz, line 1013:
```suggestion
    List.init (Z.to_int len) (Z.of_int)
```

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+    if MaySet.is_top may_nulls_set then
+      MaySet.filter cond (compute_set max_size)
+    else
+      MaySet.filter cond may_nulls_set
+
+  let may_nulls_min_elt may_nulls_set =
+    if MaySet.is_top may_nulls_set then
+      Z.zero
+    else
+      MaySet.min_elt may_nulls_set
```
michael-schwarz, line 1050: Why not put them inside your may/must modules respectively?

~~~ocaml

module MustSet = struct 
  include SetDomain.Reverse (SetDomain.ToppedSet (IntDomain.BigInt) (struct let topname = "All indexes" end))
  (* the functions for MustSet *)
  
  let filter cond must_nulls_set min_size =
    if MustSet.is_bot must_nulls_set then
      MustSet.filter cond (compute_set min_size)
    else
      MustSet.filter cond must_nulls_set
end
~~~

Then you can directly call things like `MustSet.filter` which should make the code nicer to read.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+
+  let name () = "arrays containing null bytes"
+  type idx = Idx.t
+  type value = Val.t
+
+  type ret = Null | NotNull | Top
+
+  (* helper: returns Idx.maximal except for Overflows that are mapped to None *)
+  let idx_maximal i = match Idx.maximal i with
+    | Some i -> (try Some (Z.of_int (Z.to_int i)) with Z.Overflow -> None)
```
michael-schwarz, line 1070: What exactly is this doing? It goes from `Z -> int -> Z` ? If we need this, then

```suggestion
    | Some i when Z.fits_int i -> Some i
```

may be more appropriate. Also, if you want to make sure the value fits into `int`, why is the return type still `Z.t`? Something doesn't seem to quite add up here...

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+
+  let get ?(checkBounds=true) (ask: VDQ.t) (must_nulls_set, may_nulls_set, size) (e, i) =
+    let all_indexes_must_null i max =
+      let rec check_all_indexes i =
+        if Z.gt i max then
+          true
+        else if MustNulls.mem i must_nulls_set then
+          check_all_indexes (Z.succ i)
+        else
+          false in
```
michael-schwarz, line 1081: Since this only used in the last else case, it might make sense to move it there for clarity.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+  type value = Val.t
+
+  type ret = Null | NotNull | Top
+
+  (* helper: returns Idx.maximal except for Overflows that are mapped to None *)
+  let idx_maximal i = match Idx.maximal i with
+    | Some i -> (try Some (Z.of_int (Z.to_int i)) with Z.Overflow -> None)
+    | None -> None
+
+  let get ?(checkBounds=true) (ask: VDQ.t) (must_nulls_set, may_nulls_set, size) (e, i) =
```
michael-schwarz, line 1073: Sine we don't actually do bounds checking here it may make sense to drop the `checkBounds` optional argument here and in the module type `Str`.

michael-schwarz (changes requested):
Thanks again for the PR, looks mostly good! I left a couple of style suggestions that I would hope are easy to implement and might make things easier to read, by reducing the amount of code somewhat.

The biggest blocker to me is that the NullByte domain does not seem to work for malloc'ed things unless their value is set via `strcpy`, which seems very strange and makes the analysis unsound (see example `73/09`). The problem seems to be that the value inside the blob is not made an actual array when the writes happens. This needs to be addressed before it makes sense for us to merge this.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+    (* if there is no maximum size *)
+    | Some max_i, None when Z.geq max_i Z.zero ->
+      (* ... and maximum value in index interval < minimal size, return Null if all numbers in index interval are in must_nulls_set *)
+      if Z.lt max_i min_size && all_indexes_must_null min_i max_i then
+        Null
+        (* ... return NotNull if no number in index interval is in may_nulls_set *)
+      else if not (MayNulls.exists (fun x -> Z.geq x min_i && Z.leq x max_i) may_nulls_set) then
+        NotNull
+      else
+        Top
```
michael-schwarz, line 1114: According to coverage reports, this code is never exercised. I think we should definitely add such cases. 
https://coveralls.io/builds/62854622/source?filename=src%2Fcdomains%2FarrayDomain.ml

(I guess by accessing at unknown indexes?)

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+          (must_nulls_remove i must_nulls_set min_size, may_nulls_set, size)
+          (* ..., i < minimal size and value = null, add i to must_nulls_set and may_nulls_set *)
+        else if Z.lt i min_size && Val.is_null v then
+          (MustNulls.add i must_nulls_set, MayNulls.add i may_nulls_set, size)
+          (* ..., i >= minimal size and value = null, add i only to may_nulls_set *)
+        else if Val.is_null v then
+          (must_nulls_set, MayNulls.add i may_nulls_set, size)
+          (* ... and value unknown, remove i from must_nulls_set and add it to may_nulls_set *)
+        else
+          (must_nulls_remove i must_nulls_set min_size, MayNulls.add i may_nulls_set, size)
```
michael-schwarz, line 1158: Never exercised by tests.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+       else
+         match Idx.minimal size, idx_maximal size with
+         (* ... and size unknown, modify both sets to top *)
+         | None, None -> (MustNulls.top (), MayNulls.top (), size)
+         (* ... and only minimal size known, remove all indexes < minimal size from must_nulls_set and modify may_nulls_set to top *)
+         | Some min_size, None -> (must_nulls_filter (Z.gt min_size) must_nulls_set min_size, MayNulls.top (), size)
+         (* ... and only maximal size known, modify must_nulls_set to top and add all i from minimal index to maximal size to may_nulls_set *)
+         | None, Some max_size -> (MustNulls.top (), add_indexes min_i (Z.pred max_size) may_nulls_set, size)
+         (* ... and size is known, remove all indexes < minimal size from must_nulls_set and add all i from minimal index to maximal size to may_nulls_set *)
+         | Some min_size, Some max_size -> (must_nulls_filter (Z.gt min_size) must_nulls_set min_size, add_indexes min_i (Z.pred max_size) may_nulls_set, size))
```
michael-schwarz, line 1229: Never exercised by tests.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+          let min_size2 = match Idx.minimal size2' with
+            | Some min_size2 -> min_size2
+            | None -> Z.zero in
+          must_nulls_filter (Z.gt min_size1) must_nulls_set2' min_size2
+          |> MustNulls.union (must_nulls_filter (Z.leq max_len2) must_nulls_set1 min_size1) in
+        let may_nulls_set_result = 
+          (* get all may nulls from src string as no maximal size of dest *)
+          may_nulls_set2'
+          |> MayNulls.union (may_nulls_filter (Z.leq min_len2) may_nulls_set1 (Z.succ min_len2))  in
+        (must_nulls_set_result, may_nulls_set_result, size1)
```
michael-schwarz, line 1465: Never exercised.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+            |> MayNulls.filter (Z.gt max_size1) 
+          else if not (MayNulls.is_top may_nulls_set1) && not (MayNulls.is_top may_nulls_set2') && maxlen1_exists && maxlen2_exists then
+            MayNulls.filter (fun x -> if maxlen1_exists && maxlen2_exists then Z.leq x (Z.add maxlen1 maxlen2) else true) may_nulls_set1
+            |> MayNulls.elements
+            |> BatList.cartesian_product (MayNulls.elements may_nulls_set2')
+            |> List.map (fun (i1, i2) -> Z.add i1 i2)
+            |> MayNulls.of_list
+            |> MayNulls.union (MayNulls.filter (Z.lt (Z.add minlen1 minlen2)) may_nulls_set1)
+          else
+            MayNulls.top () in
```
michael-schwarz, line 1570: Never exercised.

src/cdomains/arrayDomain.ml
```diff
@@ -941,7 +1003,762 @@ struct
...
+            |> MayNulls.filter (fun x -> if max_size1_exists then Z.gt max_size1 x else true) 
+          else if not (MayNulls.is_top may_nulls_set1) then
+            MayNulls.filter (fun x -> if maxlen1_exists && maxlen2_exists then Z.leq x (Z.add maxlen1 maxlen2) else true) may_nulls_set1
+            |> MayNulls.elements
+            |> BatList.cartesian_product (MayNulls.elements may_nulls_set2'_until_min_i2)
+            |> List.map (fun (i1, i2) -> Z.add i1 i2)
+            |> MayNulls.of_list
+            |> MayNulls.union (MayNulls.filter (Z.lt (Z.add minlen1 minlen2)) may_nulls_set1)
+          else
+            MayNulls.top () in
```
michael-schwarz, line 1614: Never exercised.

tests/regression/73-strings/09-malloc.c
```diff
@@ -0,0 +1,16 @@
...
+
+int main () {
+    char* s1 = malloc(50);
+    s1[0] = 'a';
+
+    char s2[50];
+    s2[0] = 'a';
+
+    // Use size_t to avoid integer warnings hiding the lack of string warnings
+    size_t len1 = strlen(s1); //WARN
```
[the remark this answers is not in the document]
nathanschmidt: I have made the same observation. The domain is only called after an initialization through `memcpy` or `strcpy`. However, I don't really think I have enough insights into the other parts of Goblint to fix this. I am happy to abide by any suggestion you might have @michael-schwarz. Or just feel free to make some changes directly if that is easier.

src/cdomains/valueDomain.ml
```diff
@@ -713,6 +758,21 @@ struct
     | _, Bot -> Bot (* Leave uninitialized value (from malloc) alone in free to avoid trashing everything. TODO: sound? *)
     |                 t , _             -> top_value t
 
+  let invalidate_abstract_value = function
+    | Top -> Top
+    | Int i -> Int (ID.top_of (ID.ikind i))
+    | Float f -> Float (FD.top_of (FD.get_fkind f))
+    | Address _ -> Address (AD.top_ptr)
+    | Struct _ -> Struct (Structs.top ())
```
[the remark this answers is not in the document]
nathanschmidt: Works for `Struct` and `Array`. However `Unions` doesn't have a `map` function. Any alternative approach?

src/cdomains/arrayDomain.ml
nathanschmidt: `Idx.maximal` doesn't map $\infty$ to `None` but to an invalid value of `Z.t`, that whenever used causes the exception `Z.overflow`. Therefore I want to filter out such invalid values and set them to `None`. From my understanding, `Z.fits_int` also works. I think this wrapper makes the logic easier to understand in the rest of the code.

src/cdomains/arrayDomain.mli
```diff
@@ -53,29 +46,108 @@ sig
...
+  val string_concat: t -> t -> int option -> t
+  (** [string_concat s1 s2 n] returns a new abstract value representing the string 
+    * concatenation of the input abstract values [s1] and [s2], taking at most [n] bytes of
+    * [s2] if present *)
+
+  val substring_extraction: t -> t -> bool * bool
+  (** [substring_extraction haystack needle] returns [is_null_ptr, is_offset_0], i.e. 
+    * [true, false] if the string represented by the abstract value [needle] surely isn't a 
+    * substring of [haystack], [false, true] if [needle] is the empty string, 
+    * else [false, false] *)
```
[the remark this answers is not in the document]
nathanschmidt: I've added the `type substr = IsNotSubstr | IsSubstrAtIndex0 | IsMaybeSubstr`. However, OCaml says that `N.substr <> substr` in the AttributeConfiguredAndNullByteArrayDomain, meaning I had to write:
```ocaml
let substring_extraction (_, t_n1) (_, t_n2) = match N.substring_extraction t_n1 t_n2 with
    | IsNotSubstr when get_bool "ana.base.arrays.nullbytes" -> IsNotSubstr
    | IsSubstrAtIndex0 when get_bool "ana.base.arrays.nullbytes" -> IsSubstrAtIndex0
    | _ -> IsMaybeSubstr
```
Is there any way around this I'm not aware of?

src/cdomains/arrayDomain.ml
nathanschmidt: I've added an example where there is no maximum value in the index interval, but I don't know how to write a short example where there is no maximum size as initialization by an `int` variable or non-constant length leads to an error. This case rather is here for completeness.

src/cdomains/arrayDomain.ml
nathanschmidt: Same as above: I don't know how to enforce that `size` has no upper bound.

src/cdomains/arrayDomain.ml
nathanschmidt: Same as above

src/cdomains/arrayDomain.ml
nathanschmidt: Same as above.

src/analyses/base.ml
```diff
@@ -2078,24 +2065,88 @@ struct
...
+        begin match (get (Analyses.ask_of_ctx ctx) gs st s1_a None), get (Analyses.ask_of_ctx ctx) gs st s2_a None with
+          | Array array_s1, Array array_s2 -> set ~ctx ~blob_destructive:true (Analyses.ask_of_ctx ctx) gs st lv_a lv_typ (op_array array_s1 array_s2)
+          | Array array_s1, _ when CilType.Typ.equal s2_typ charPtrType ->
+            let s2_null_bytes = List.map CArrays.to_null_byte_domain (AD.to_string s2_a) in
+            let array_s2 = List.fold_left CArrays.join (CArrays.bot ()) s2_null_bytes in
+            set ~ctx ~blob_destructive:true (Analyses.ask_of_ctx ctx) gs st lv_a lv_typ (op_array array_s1 array_s2)
+          | Bot, Array array_s2 ->
+            (* If we have bot inside here, we assume the blob is used as a char array and create one inside *)
+            let ptrdiff_ik = Cilfacade.ptrdiff_ikind () in
+            let size = ctx.ask (Q.BlobSize s1) in
```
nathanschmidt, line 2120: I am getting the following compilation error:
`This form is not allowed as the type of the inlined record could escape.`

jerhard: The reason is that `Q.BlobSize` is not taking an `exp` but a record `{exp: Cil.exp; base_address: bool}`.
So changing it to `{exp = s1; base_address = b}`, with `b` $\in \{true, false\}$ compiles.

michael-schwarz: Resolved.

src/cdomains/valueDomain.ml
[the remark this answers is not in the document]
michael-schwarz: Actually it doesn't really make too much sense for Unions, as which component is used to access the union is lost here anyway. However, it does make sense for blobs where I added it.

src/cdomains/arrayDomain.mli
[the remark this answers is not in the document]
michael-schwarz:
~~~OCaml
type substr = N.substr = IsNotSubstr | IsSubstrAtIndex0 | IsMaybeSubstr
~~~
helped here.

src/analyses/base.ml
```diff
@@ -2229,53 +2293,50 @@ struct
...
           let dest_typ = Cilfacade.typeOfLval lv_val in
-          let lval = mkMem ~addr:(Cil.stripCasts s) ~off:NoOffset in
-          let address = eval_lv (Analyses.ask_of_ctx ctx) gs st lval in
-          let (value:value) = Int(AD.to_string_length address) in
+          let v = eval_rv (Analyses.ask_of_ctx ctx) gs st s in
+          let a = address_from_value v in
+          let value:value =
+            (* if s string literal, compute strlen in string literals domain *)
+            if AD.type_of a = charPtrType then
+              Int (AD.to_string_length a)
```
sim642, line 2308: Polymorphic comparison of types. But also, I'm not sure how reliable this type-based check is because looks like it's trying to check for `StrPtr`. But there may be a `char*` which isn't `StrPtr`.

src/analyses/base.ml
```diff
@@ -2189,24 +2189,88 @@ struct
...
+    in
+    let string_manipulation s1 s2 lv all op_addr op_array =
+      let s1_v = eval_rv (Analyses.ask_of_ctx ctx) gs st s1 in
+      let s1_a = address_from_value s1_v in
+      let s1_typ = AD.type_of s1_a in
+      let s2_v = eval_rv (Analyses.ask_of_ctx ctx) gs st s2 in
+      let s2_a = address_from_value s2_v in
+      let s2_typ = AD.type_of s2_a in
+      (* compute value in string literals domain if s1 and s2 are both string literals *)
+      if CilType.Typ.equal s1_typ charPtrType && CilType.Typ.equal s2_typ charPtrType then
```
sim642, line 2213: Same type-based `StrPtr` check here.

src/analyses/base.ml
```diff
@@ -2189,24 +2189,88 @@ struct
...
+        let rec lo = function
+          | `Index (i, `NoOffset) -> `NoOffset
+          | `NoOffset -> `NoOffset
+          | `Field (f, o) -> `Field (f, lo o)
+          | `Index (i, o) -> `Index (i, lo o) in
+        let rmLastOffset = function
+          | Addr.Addr (v, o) -> Addr.Addr (v, lo o)
+          | other -> other in
+        AD.map rmLastOffset a
+      | _ -> raise (Failure "String function: not an address")
```
sim642, line 2203: Why can the last index just be ignored unconditionally. I find it odd because this looks like `Offset.remove_offset` but that one also consistently removes fields.

Also, this is just called `address_from_value` and exposed in quite large scope, which is prone to misuse. If the logic is indeed exactly as needed, then this should have a different name or be withint `string_manipulation` or whereever.

src/cdomains/valueDomain.ml
```diff
@@ -19,11 +19,12 @@ sig
   include Lattice.S
   type offs
   val eval_offset: VDQ.t -> (AD.t -> t) -> t-> offs -> exp option -> lval option -> typ -> t
-  val update_offset: VDQ.t -> t -> offs -> t -> exp option -> lval -> typ -> t
+  val update_offset: ?blob_destructive:bool -> VDQ.t -> t -> offs -> t -> exp option -> lval -> typ -> t
   val update_array_lengths: (exp -> t) -> t -> Cil.typ -> t
   val affect_move: ?replace_with_const:bool -> VDQ.t -> t -> varinfo -> (exp -> int option) -> t
   val affecting_vars: t -> varinfo list
   val invalidate_value: VDQ.t -> typ -> t -> t
+  val invalidate_abstract_value: t -> t
```
sim642, line 27: Why a different thing here? If we're dealing with strings, surely we know the type, no?

src/analyses/base.ml
michael-schwarz: The idea is that the framework has tacked on a `[0]` offset to string pointer somewhere, but we do want the raw pointer here. You are probably right about naming and moving to a different scope.

tests/regression/73-strings/01-string_literals.c
```diff
@@ -10,28 +11,49 @@ char* hello_world() {
 void id(char* s) {
     char* ptr = NULL; // future usage of cmp should warn: workaround for macOS test
     #ifdef __APPLE__
-        #define ID int i = strcmp(ptr, "trigger warning")
+        #define ID int i = *ptr
```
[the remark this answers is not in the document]
michael-schwarz: `strcmp` does not longer unconditionally warn on `OS X` (something about how the checks are compiled), so this was switched to be a direct null pointer dereference to ensure we produce a warning.

src/cdomains/nullByteSet.ml
```diff
@@ -0,0 +1,200 @@
...
+        else if M.mem i set then
+          check_all_indexes (Z.succ i)
+        else
+          false in
+      check_all_indexes l
+end
+
+module MaySet = struct
+  module M = SetDomain.ToppedSet (IntDomain.BigInt) (struct let topname = "All Null" end)
+  include M
```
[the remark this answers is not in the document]
sim642: Indeed, we already have interval sets and they're simply a generalization of this, so that should be possible but is probably too much work for now.

src/analyses/base.ml
sim642: May be, but there must be a more reliable way to know when this needs to be dropped. Because this also drops such indices when they explicitly occur in the program, rather than being inserted by CIL, and then we would likely do something wrong. For example, what about an array of strings (i.e. multi-dimensional array of `char`s).

michael-schwarz: I have not investigated this in detail, but agree someone should ideally do so.

PR #1077, opened 2023-06-06
ThreadCreate analysis to improve partially context-sensitive threadID analysis 

michael-schwarz:
This is a reimplementation of #1007 to play nice with the new uniqueness counter integrating the analysis into the `threadId` analysis as suggested in https://github.com/goblint/analyzer/pull/1007#discussion_r1149398038.

Original implementation by @FelixKrayer.


tests/regression/74-threadCreate/01-simple.c
sim642, line 1: Instead of creating yet another group with a handful of tests, maybe these could simply go into 40-threadid. Especially since it's now entirely part of the threadid analysis.

src/util/options.schema.json
```diff
@@ -989,8 +989,21 @@
...
               "type": "integer",
               "default": 0
+            },
+            "context": {
+              "title": "ana.thread.context",
+              "type": "object",
+              "properties": {
+                "create-edges": {
+                  "title": "ana.thread.context.create-edges",
+                  "description": "threadID analysis: Encountered create edges in context.",
```
sim642, line 999: It's not just the context though but also the local state. So why not have it simply as `ana.thread.create-edges`?

src/util/options.schema.json
```diff
@@ -989,8 +989,21 @@
...
+                  "type": "boolean",
+                  "default": true
+                }
+              },
+              "additionalProperties": false
             }
-          },
+        },
           "additionalProperties": false
```
sim642, line 1007:
```suggestion
          },
          "additionalProperties": false
```

tests/regression/74-threadCreate/01-simple.c
```diff
@@ -0,0 +1,32 @@
+// PARAM: --disable ana.thread.context.create-edges
```
sim642, line 1: All these tests disable the new option, so aren't they just testing the old behavior without the edges?
Seems like there should be tests for the cases where enabling the new option actually makes us more precise than before as well.

src/analyses/threadId.ml
```diff
@@ -32,24 +32,30 @@ struct
   module N = Lattice.Flat (VNI) (struct let bot_name = "unknown node" let top_name = "unknown node" end)
   module TD = Thread.D
 
-  module D = Lattice.Prod3 (N) (ThreadLifted) (TD)
+  module D = Lattice.Prod3 (N) (ThreadLifted) (Lattice.Prod(TD)(TD))
```
sim642, line 35: Should document here the meaning of different components because the difference between the two `TD`s is not clear.

tests/regression/74-threadCreate/01-simple.c
michael-schwarz: The option is active by default as that corresponds to the old behavior. What the changes in here are about is holding on to more precision, when reducing the amount of context-sensitivity.

src/util/options.schema.json
michael-schwarz: The set needs to be tracked anyway, the question really is whether they get put into the context or not.

michael-schwarz:
@sim642 When you get a chance, can you have a look? The comments should be addressed, and it would be nice to have this merged before it becomes stale again.

sim642 (approved):

PR #1078, opened 2023-06-07
ThreadEscape: Collect set of escaping threads flow-insensitively

### LINKED ISSUE #1074: Escaped variables are not communicated between threads

jerhard:
This PR changes the `ThreadEscape` analysis to track which threads escaped a variable in a flow-insensitive invariant. A thread escapes a variable if it assigns it to an lvalue that may already be escaped.

The escape information in the local state now collects only the escapes that happened in the threads own past. That means, there is no joining of escaped variables from the flow-insensitive invariant into the local state. 

When queried whether a variable is escaped, the analysis checks the flow-insensitive set of threads that escaped it. If the set is empty, it is not escaped. If the set is not empty, it is checked whether any of the other threads that escaped it may already be started. If so, the variable may be escaped. If none of the other threads escaped the variable, then it is checked whether the current thread may have escaped the variable. This is done by looking it up in the local state. 

This approach in particular handles the case when variables escape via globals after some threads have already started (see #1074), while not treating such variables as escaped at all program points (which was one issue with #1075, as well as the issue with the implementation in `threadenter` mentioned here: https://github.com/goblint/analyzer/pull/1075#issuecomment-1578756437).

Fixes #1074

### DIFF
--- a/src/analyses/threadEscape.ml
+++ b/src/analyses/threadEscape.ml
@@ -42,10 +42,15 @@ struct
       if M.tracing then M.tracel "escape" "mpt %a: %a\n" d_exp e Queries.LS.pretty a;
       D.empty ()
 
+  let emit_escaped ctx escaped =
+    if not (D.is_empty escaped) then (* avoid emitting unnecessary event *)
+      ctx.emit (Events.Escape escaped)
+
   (* queries *)
   let query ctx (type a) (q: a Queries.t): a Queries.result =
     match q with
-    | Queries.MayEscape v -> D.mem v ctx.local
+    | Queries.MayEscape v ->
+      D.mem v ctx.local
     | _ -> Queries.Result.top q
 
   (* transfer functions *)
@@ -57,8 +62,8 @@ struct
       let escaped = reachable ask rval in
       let escaped = D.filter (fun v -> not v.vglob) escaped in
       if M.tracing then M.tracel "escape" "assign vs: %a | %a\n" D.pretty vs D.pretty escaped;
-      if not (D.is_empty escaped) && ThreadFlag.has_ever_been_multi ask then (* avoid emitting unnecessary event *)
-        ctx.emit (Events.Escape escaped);
+      if ThreadFlag.has_ever_been_multi ask then (* avoid emitting unnecessary event *)
+        emit_escaped ctx escaped;
       D.iter (fun v ->
           ctx.sideg v escaped;
         ) vs;
@@ -73,8 +78,7 @@ struct
     | _, "pthread_setspecific" , [key; pt_value] ->
       let escaped = reachable (Analyses.ask_of_ctx ctx) pt_value in
       let escaped = D.filter (fun v -> not v.vglob) escaped in
-      if not (D.is_empty escaped) then (* avoid emitting unnecessary event *)
-        ctx.emit (Events.Escape escaped);
+      emit_escaped ctx escaped;
       let extra = D.fold (fun v acc -> D.join acc (ctx.global v)) escaped (D.empty ()) in (* TODO: must transitively join escapes of every ctx.global v as well? *)
       D.join ctx.local (D.join escaped extra)
     | _ -> ctx.local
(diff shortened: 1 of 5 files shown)


jerhard:
Some changes to the `RelationAnalysis.threadenter` were necessary. The reason is that the `RelationAnalysis` relied on variables that were reachable for the created thread via an argument to the `pthread_create` to be considered escaped right away, otherwise, it would crash. This becomes a problem now that the information that the parameter to the `pthread_create` escapes is only computed later during the evaluation of the constraint system (via a side-effect in `threadspawn`.
Therefore, I changed the `RelationAnalysis.threadenter` such that it maintains variables reachable from passed argument in the relation `rel` of the created thread.

src/analyses/threadEscape.ml
```diff
@@ -42,76 +44,124 @@ struct
...
+          | `Bot -> false
+        in
+        match ctx.ask Queries.CurrentThreadId with
+        | `Lifted current ->
+          let possibly_started = ThreadIdSet.exists (possibly_started current) threads in
+          if possibly_started then
+            true
+          else if ThreadIdSet.exists (equal_current_not_unique current) threads then
+            (* Another instance of the non-unqiue current thread may have escaped the variable *)
+            true
```
michael-schwarz, line 96: Maybe first check current thread for uniqueness?

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -296,12 +293,17 @@ struct
...
-    let ask = Analyses.ask_of_ctx ctx in
-    let new_rel = List.fold_left (fun new_rel (var, e) ->
-        assign_from_globals_wrapper ask ctx.global {st with rel = new_rel} e (fun rel' e' ->
-            RD.assign_exp rel' var e' (no_overflow ask e)
-          )
-      ) new_rel arg_assigns
+    let new_rel =
+      if thread then
+        (* TODO: Why does test 63/16 not reach fixpoint without copy here? *)
+        RD.copy new_rel
```
michael-schwarz, line 299: That is indeed worth investigating, any mutable state should be hidden!

michael-schwarz (approved):
Apart from the comments, i think this is good to merge!

src/analyses/threadEscape.ml
jerhard: Good idea.

src/analyses/apron/relationAnalysis.apron.ml
jerhard: This `RD.copy` was necessary due to the destructive `RD.remove_filter_with` happening later. I now moved the `RD.copy` up, to ensure that it is done before the `RD.remove_filter_with`.

michael-schwarz:
@sim642 Do you still want to review this or are we good to merge?

src/analyses/apron/relationAnalysis.apron.ml
sim642: `RD.add_vars` isn't a `_with` function, so it should already do a copy. Unless there's something wrong there.

jerhard: The `add_vars` of the `AffineEqualityDomain` does not do any explicit copying. It calls `SharedFunctions.EnvOps.add_vars`, which is defined as follows:
https://github.com/goblint/analyzer/blob/a651cccf69af65581222d7e8e046b8ad67510cf4/src/cdomains/apron/sharedFunctions.apron.ml#L303-L310

I assume that in case the list of vars that is added is empty, the `add_vars` for affine equalities will return the same physical object as its argument.

sim642: In that case I think such optimization in `AffineEqualityDomain` is incorrect because as far as I am aware, all non-`_with` functions should provide a pure functional interface with no unintended sharing between the inputs and outputs. That's what makes it possible to write any imperative code with `_with` functions without having to worry about that modifying something which isn't a direct argument of the imperative function.

jerhard: I added an explicit copy operation to the `add_vars`, `drop_vars`, `keep_vars` and `keep_filter`  operations in `AffineEqualityDomain`. The `copy` in `RelationAnalysis.make_calleee_rel` thus was removed.

michael-schwarz:
@sim642 Since the remaining issue was addressed, is this now good to merge?

sim642 (approved):

PR #1081, opened 2023-06-09
Add location to group messages

sim642:
This is cherry-picked from #914 with some small changes.

Race warning groups themselves now also have locations of the memory location variable. This was previously already in the group message text, but simply after the variable name with `@`. A separate location field makes this neater and usable by other tools like GobPie.


michael-schwarz (approved):

sim642:
I might currently hold off merging this to do the bigger race analysis changes in #1084 and #1089 without introducing so many conflicts.

PR #1083, opened 2023-06-13
Ignore indentation for blank lines suggested by ocp-indent

stilscher:
With our usual vscode settings trailing whitespaces at the end of a line are removed (whenever a file is saved). This is also the case for blank lines. Ocp-indent on the other hand complains if blank lines between two indented lines are not indented in the same way. I did not find a configuration option for ocp-indent that allows to have no indentation on blank lines.
To avoid spurious ocp-indent warnings that are created by this differing behaviour of the IDE and ocp-indent  I would suggest to ignore the indentation required by ocp-indent on blank lines in our pre-commit hook, because I think removing trailing whitespaces at the end of lines is desirable.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -28,3 +28,6 @@ ec8611a3a72ae0d95ec82ffee16c5c4785111aa1
 
 # Set up end-of-line normalization.
 78fd79e7f4d15c4412221b155971fac2e0616b90
+
+# fix indentation in baseInvariant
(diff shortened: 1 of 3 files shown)


michael-schwarz:
Does this actually solve the issue though? The CI would still complain, no?

stilscher:
The same script (`scripts/hooks/pre-commit`) is run in the indentation workflow. So it should prevent the indentation action from failing just because of blank line indentations.

sim642 (approved):

PR #1084, opened 2023-06-13
Refactor and fix race access distribution

sim642:
This is the result of reverse-engineering the existing race access distribution code. That code is cleaned up, simplified and commented.

This does not yet do anything clever to avoid distributing accesses to many memory locations, but makes the rules for doing so a lot clearer, hopefully simplifying the redesign.
However, the number of accesses, memory locations and races may change as the result of this PR, in particular due to some included soundness fixes.

## Changes
1. **Define a new `Access.Memo` type for precisely representing memory locations as used by the race analysis.**

    The previous representation used two `options`, of which not all four combinations were possible, and both of them could contain offsets, which were duplicated.
    The new representation has a `varinfo` or `typ` as root, followed by unit-indexed offsets. This also allows reusing features from #1067.

2. **Completely rewrite `Access.add_propagate` as `add_distribute_outer`.**

    Its previous logic was very confusing and completely unexplained. In some ways it had an overly specific special case, while the general case was both too precise or too imprecise, depending on the situation. The new logic generalizes this to two general principles according to which type-based accesses are distributed:
    1. A type-based access is distributed to all global variables of that type.
    2. A type-based access is distributed to all struct fields of that type. This aspect is recursive, i.e. those are also distributed to global variables, etc.

    The new logic fixes both soundness and precision issues that existed previously.

3. **Fix unsoundness regarding distribution to struct fields with array type.**

    Previously accesses via `int*` didn't account for the possibility that it might point to and thus race with an element of a struct field with type `int[2]`. Now arbitrarily deep multi-arrays are unwrapped.

4. **Rename and flip `ana.mutex.disjoint_types` to `ana.race.direct-arithmetic`.**

    This should somewhat clarify the meaning of the option and avoid double negation (`not !unsound`). Disjoint types is rather nondescriptive since it really just controls the behavior of direct (i.e. non-field) accesses to arithmetic-type–based memory locations.

5. **Moves the handling of `ana.race.direct-arithmetic` to just one place.**

    Previously two partial checks were in two different places, obfuscating the logic.

6. **Exclude accesses to some anonymous pthread type internals, which show up as `__anonstruct` and `__anonunion`.**

    Existing exclusions are based on `TNamed`, so simply `typedef` names. But anonymous structs are precisely the ones that don't have such names.

7. **Remove polymorphic `Hashtbl` usage from `Access`.**

    Although it didn't cause issues as the hashtables were with `typsig` keys, it's still nicer to avoid the possibility explicitly. `typsig`s are still used to make sure that all type-based accesses are collected regardless of any `typedef`s.

8. **Add `access` tracing.**

9. **Add option `warn.deterministic`.**

    This makes the order of printed messages deterministic, by delaying the printing and later printing them in sorted (by text, etc) order. This is useful for writing cram tests such that the test doesn't spuriously break (e.g. on OSX) due to different `varinfo` IDs being in different order in some hashtables that get iterated.

## TODO
- [x] Remove system-dependent cram tests.


src/analyses/accessAnalysis.ml
```diff
@@ -42,13 +42,15 @@ struct
...
   let access_one_top ?(force=false) ?(deref=false) ctx (kind: AccessKind.t) reach exp =
-    if M.tracing then M.traceli "access" "access_one_top %a %b %a:\n" AccessKind.pretty kind reach d_exp exp;
+    if M.tracing then M.traceli "access" "access_one_top %a (kind = %a, reach = %B, deref = %B)\n" CilType.Exp.pretty exp AccessKind.pretty kind reach deref;
     if force || !collect_local || !emit_single_threaded || ThreadFlag.has_ever_been_multi (Analyses.ask_of_ctx ctx) then (
       if deref then
         do_access ctx kind reach exp;
-      Access.distribute_access_exp (do_access ctx Read false) exp
+      if M.tracing then M.tracei "access" "distribute_access_exp\n";
+      Access.distribute_access_exp (do_access ctx Read false) exp;
+      if M.tracing then M.traceu "access" "distribute_access_exp\n";
```
michael-schwarz, line 51: Maybe make the second message different from the first one so it's easier to interpret tracing output.

src/analyses/raceAnalysis.ml
```diff
@@ -78,14 +71,14 @@ struct
         | `Left g' -> (* accesses *)
           (* ignore (Pretty.printf "WarnGlobal %a\n" CilType.Varinfo.pretty g); *)
           let accs = G.access (ctx.global g) in
-          let (lv, ty) = g' in
-          let mem_loc_str = GobPretty.sprint Access.d_memo (ty, lv) in
-          Timing.wrap ~args:[("memory location", `String mem_loc_str)] "race" (Access.warn_global safe vulnerable unsafe g') accs
+          let memo = g' in
```
michael-schwarz, line 74: Why not just `Left memo` in the case?

src/domains/access.ml
```diff
@@ -413,7 +434,7 @@ let race_conf accs =
 let is_all_safe = ref true
 
 (* Commenting your code is for the WEAK! *)
-let incr_summary safe vulnerable unsafe (lv, ty) grouped_accs =
+let incr_summary safe vulnerable unsafe _ grouped_accs =
```
michael-schwarz, line 437: Why have an ignored argument here?

src/util/options.schema.json
```diff
@@ -2125,6 +2117,12 @@
           "description": "Races with confidence at least threshold are warnings, lower are infos.",
           "type": "integer",
           "default": 0
+        },
+        "deterministic": {
+          "title": "warn.deterministic",
+          "description": "Output messages in deterministic order. Useful for cram testing.",
+          "type": "boolean",
+          "default": false
```
michael-schwarz, line 2125: How expensive is this sorting for larger benchmarks? If it is not excessive, we might just make it the default and avoid adding a further option.

michael-schwarz (commented):
Definitely looks a lot more reasonable now. I am a bit unhappy with the number of TODO comments we're adding here though! I don't think we'll ever be in a better situation to address these than now.

src/analyses/accessAnalysis.ml
sim642: One's indenting and the other's unindenting. The latter ones still are printed with the indent, so they are already different. That's how we do intented tracing elsewhere as well.

src/util/options.schema.json
sim642: Message order is irrelevant in benchmarks though, so that would just be silly overhead to add.

The bigger reason to not make it default is usability. Right now messages are printed as transfer functions are evaluated, which happens in nice program order and in the order messages are created within a transfer function. It's actually quite nice to be able to read messages for linear code in the order of the code itself. Sorting them by text completely shuffles this up and makes `M.debug` things particularly useless since those often appear right before or after some meaningful messages.

Another reason is that I'm not confident of the reliability of this. This works fine for messages printed during transfer functions, but we also output messages from so many other stages of the analysis that some of them might not be properly accounted for. E.g. some messages may get created after one round of analysis and thus never get printed. This concerns incremental and server mode in particular.

michael-schwarz: Fair points!

michael-schwarz (approved):

PR #1087, opened 2023-06-15
Document regression test annotations fully

sim642:


docs/developer-guide/testing.md
```diff
@@ -24,8 +24,44 @@ gobopt='--set ana.base.privatization write+lock' ./scripts/update_suite.rb
...
+#### First line
+A comment on the first line can contain the following:
+
+| Annotation | Comment |
+| ---------- | ------- |
+| `PARAM: ` <br> (NB! space) | The following command line parameters are added to Goblint for this test. |
+| `SKIP` | The test is skipped (except when run with `./scripts/update_suite.rb group`). |
+| `NOMARSHAL` | Marshaling and unmarshaling of results is not tested on this program. |
+| `TERM` | The expected Goblint result is that the program terminates. |
+| `NONTERM` <br> or `NOTERM` | The expected Goblint result is that the program does not terminate. |
```
michael-schwarz, line 38: I think this checked in terms of the end of main being reachable, which has very little to do with termination. I think we should rename this while we're documenting it.

sim642: Well, it has something to do with very primitive termination analysis, although that check is not even sound because it doesn't account for early exit via `exit` or such. It shouldn't be too hard to make such a simple yet sound termination analysis.

Actually, I'll just remove these annotations and replace them with explicit checks for (un)reachability at the end of `main` like we check for (un)reachability elsewhere. Then there's no need to even rename it.

michael-schwarz: Yes, our students are working on a simple yet sound termination analysis, though primarily with the goal to prover termination, whereas non-reachability of all exit points would be prove definite non-termination. Though I think that there are only few cases where that would work.

docs/developer-guide/testing.md
```diff
@@ -24,8 +24,42 @@ gobopt='--set ana.base.privatization write+lock' ./scripts/update_suite.rb
...
-* Annotate lines inside the regression test with comments: `arr[9] = 10; // WARN`
+Regression tests use single-line comments (with `//`) as annotations.
+
+#### First line
+A comment on the first line can contain the following:
+
+| Annotation | Comment |
+| ---------- | ------- |
+| `PARAM: ` <br> (NB! space) | The following command line parameters are added to Goblint for this test. |
+| `SKIP` | The test is skipped (except when run with `./scripts/update_suite.rb group`). |
```
michael-schwarz, line 35:
```suggestion
| `SKIP` | The test is skipped (except when run with `./scripts/update_suite.rb group`).  This is used e.g. in the tests for Apron, as not all CI jobs have the Apron library installed. |
```

michael-schwarz (approved):

docs/developer-guide/testing.md
sim642: I added this note above where we describe the test running scripts. Too long text stretches out these tables oddly.

docs/developer-guide/testing.md
```diff
@@ -24,8 +25,42 @@ gobopt='--set ana.base.privatization write+lock' ./scripts/update_suite.rb
...
+| `NOWARN` | No warning | — | Precision |
+| `WARN` | Some warning | — | Soundness |
+
+#### Other
+Other useful constructs are the following:
+
+| Code with annotation | Comment |
+| -------------------- | ------- |
+| `__goblint_check(1); // reachable` | Checks that the line is reachable according <br> to Goblint results (soundness). |
+| `__goblint_check(0); // NOWARN (unreachable)` | Checks that the line is unreachable (precision). |
```
stilscher, line 63: So we do not output warnings for definitely unreachable dead code in Goblint, right? And that is the reason, we can use `__goblint_check(1)` and `__goblint_check(0)` to check the reachability of a program location? Just out of interest, where is this logic of not warning in case the code is dead implemented in Goblint?

sim642: Yes, if the code is dead, then Goblint does not output an assertion failure error for `0`, which it trivially can, if it were reached.

This logic is implemented nowhere, which is precisely why it's not easy to change. It's an implicit consequence of analysis transfer functions not being evaluated on dead code. The transfer functions not being evaluated means the `M.warn` calls never happen.

PR #1089, opened 2023-06-15
Use TrieDomain to distribute accesses to contained fields

karoliineh:
Builds on #1084.

Instead of distributing an access-to-a-struct to all of the fields in that struct, the implementation distributes to those fields only that had direct access(es) themselves. Meaning that if there is access to a struct only, there will be one warning for that struct instead of a separate warning for each of the struct fields. And if there was access to a field of a struct, a race between the access to that struct's field and the access to the whole struct is not lost.

For that, the accesses are collected in a `trie` structure representing a struct. In the `trie` node, there is a set of accesses to that offset and a map of that offset's children with the child offset as a key and a sub-`trie` as a value. Later, the accesses of each node's ancestors are also checked for races with the node accesses, doing the distribution.

Also fixes soundness issue described in #978.
Previously `Access.add_distribute_inner` was based on the type. Since alloc variables have `void` type, contained fields for a blob could not be iterated and distributed to. Since this delays the inner distribution and flips the direction, the issue is automatically avoided: an access to `(alloc).foo` will be checked against an access to just `(alloc)` simply due to the trie structure.

### DIFF
--- a/src/domains/access.ml
+++ b/src/domains/access.ml
@@ -80,11 +80,34 @@ let reset () =
 type acc_typ = [ `Type of CilType.Typ.t | `Struct of CilType.Compinfo.t * Offset.Unit.t ] [@@deriving eq, ord, hash]
 (** Old access type inferred from an expression. *)
 
+module MemoRoot =
+struct
+  include Printable.StdLeaf
+  type t = [`Var of CilType.Varinfo.t | `Type of CilType.Typ.t] [@@deriving eq, ord, hash]
+  (* Can't use typsig for `Type because there's no function to follow offsets on typsig. *)
+
+  let name () = "memoroot"
+
+  let pretty () vt =
+    (* Imitate old printing for now *)
+    match vt with
+    | `Var v -> Pretty.dprintf "%a@@%a" CilType.Varinfo.pretty v CilType.Location.pretty v.vdecl
+    | `Type (TComp (c, _)) -> Pretty.dprintf "(struct %s)" c.cname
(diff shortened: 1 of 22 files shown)


michael-schwarz:
Where is the PR-dependency here? Does it make sense to review this as of now?

sim642:
> Where is the PR-dependency here? Does it make sense to review this as of now?

The dependency was #1084, which is now merged, so this can be reviewed.

src/domains/access.ml
```diff
@@ -80,11 +80,34 @@ let reset () =
 type acc_typ = [ `Type of CilType.Typ.t | `Struct of CilType.Compinfo.t * Offset.Unit.t ] [@@deriving eq, ord, hash]
 (** Old access type inferred from an expression. *)
 
+module MemoRoot =
+struct
+  include Printable.StdLeaf
+  type t = [`Var of CilType.Varinfo.t | `Type of CilType.Typ.t] [@@deriving eq, ord, hash]
```
michael-schwarz, line 86: will deriving `eq` work here? Type is (potentially) recursive

michael-schwarz (approved):
Apart from the comment of deriving `eq` for `type`, LGTM!

src/domains/access.ml
sim642: `CilType.Typ` exists precisely to take care of it. The recursion in `Cil.typ` structural comparison comes from `compinfo`s and `CilType.Compinfo` has an implementation that only compares `cid`s.

michael-schwarz:
Thanks for the clarification!

PR #1092, opened 2023-06-20
Towards fixing GobView search: Avoid truncation of very large integers

stilscher:
When working on https://github.com/goblint/gobview/issues/7#issue-1162653541 I tried the following semantic query on the regression test case `tests/regression/00-sanity/01-assert.c`:
{"kind":["var"],"target":["name","fail"],"find":["uses"],"expression":"fail == 0","mode":["Must"]}

It results in a fixpoint not found. The reason for this is that integers larger than 2<sup>32</sup>, namely the start id for varinfos created by Goblint and the ids of pseudo return nodes are truncated (OCaml int is converted to a 32bit int in JavaScript with js_of_ocaml) and therefore differ from what was previously stored in the solver data. My proposed easy fix (at least for not too large projects) would be to decrease the start id by a factor of 10.

Todo
- [x] allow for syntactic queries (which are missing the expression field) to be handed to GobView as query
- [x] add test case for the search interface in GobView (requires https://github.com/goblint/gobview/pull/25/commits/d577fbded6939321ede573b3e8bf63fed68ec772)
- [x] update CIL-pin to include https://github.com/goblint/cil/pull/147#issue-1765053407
- [x] update GobView submodule to include https://github.com/goblint/gobview/pull/25


sim642:
Maybe we could just use negative IDs for what we generate in Goblint? Start counting down from -1. That way they don't conflict with anything CIL generates, regardless of imaginable program size and are also small.

Or is there some reason the IDs shouldn't be negative? Maybe @michael-schwarz knows?

michael-schwarz:
> Or is there some reason the IDs shouldn't be negative? Maybe @michael-schwarz knows?

Off the top of my head, I cannot think of any reason against it.

Tbh, I am bit surprised we end up with 32-bit integers on the javascript side, normally this should be 54 bit integers...

stilscher:
[Here](https://github.com/ocsigen/js_of_ocaml#data-representation) it states that OCaml `int`s are converted to JavaScript 32-bit integers. Also, the truncation occurs for integers >= 2147483648 (2^31). This is observable because an error is thrown when transpiling:
`Warning: integer overflow: integer 0x80000000 (2147483648) truncated to 0x80000000 (-2147483648); the generated code might be incorrect.`
Where did you find the information that it should be 54 bit integers?

michael-schwarz:
Very strange. Might have to make some experiments after the holidays...

I still like simmo's idea to go for negative numbers in these cases.

<OT>The thing is that JavaScript doesn't really have integers: on that vpage it says they're mapped to number (as is float), which is 64bit IEEE 754 floating points, which would give you 54 bit integers. But maybe they do some bitwise crap with every instruction to make sure it is actually 32 bit operations?</OT>.

src/transform/expressionEvaluation.ml
```diff
@@ -60,7 +60,11 @@ struct
 
       val global_variables =
         file.globals
-        |> List.filter_map (function Cil.GVar (v, _, _) -> Some (v.vname, Cil.Fv v) | _ -> None)
+        |> List.filter_map (function
+            | Cil.GVar (v, _, _) -> Some (v.vname, Cil.Fv v)
+            | Cil.GFun (f, l) -> Some (f.svar.vname, Cil.Fv f.svar)
+            | Cil.GVarDecl (v, l) -> Some (v.vname, Cil.Fv v)
+            | _ -> None)
```
sim642, line 67: Just a thought for future: at some point this old expression evaluation should be switched from the odd Format-CIL parser (see goblint-cil issues) to the normal expression parser that's used for witness invariants. That avoids fixing the same issues twice.

At that point one might reconsider this being a transformation altogether, because it isn't. Simply using `EvalInt` or `EvalValue` queries directly from GobView should be possible.

michael-schwarz:
What is the status here?

src/util/cilfacade.ml
```diff
@@ -8,11 +8,9 @@ include Cilfacade0
 
 (** Command for assigning an id to a varinfo. All varinfos directly created by Goblint should be modified by this method *)
 let create_var (var: varinfo) =
-  (* TODO Hack: this offset should preempt conflicts with ids generated by CIL *)
-  let start_id = 10_000_000_000 in
+  (* Hack: using negative integers should preempt conflicts with ids generated by CIL *)
   let hash = Hashtbl.hash { var with vid = 0 } in
-  let hash = if hash < start_id then hash + start_id else hash in
-  { var with vid = hash }
+  { var with vid = - hash }
```
sim642, line 13: I'm not sure if there's any guarantee that `Hashtbl.hash` be positive. If it returns something negative, then we'll flip it into a positive `vid` still.

Although I'm now wondering, why we have this business at all. `Cilfacade.create_var` is called with things from `Cil.makeGlobalVar` etc anyway, which already create unique `vid`s that don't conflict with anything else from CIL.

src/framework/cfgTools.ml
```diff
@@ -123,9 +123,7 @@ let rec pretty_edges () = function
   | (_,x)::xs -> Pretty.dprintf "%a; %a" Edge.pretty_plain x pretty_edges xs
 
 let get_pseudo_return_id fd =
-  let start_id = 10_000_000_000 in (* TODO get max_sid? *)
-  let sid = Hashtbl.hash fd.svar.vid in (* Need pure sid instead of Cil.new_sid for incremental, similar to vid in Cilfacade.create_var. We only add one return stmt per loop, so the hash from the functions vid should be unique. *)
-  if sid < start_id then sid + start_id else sid
+  - (Hashtbl.hash fd.svar.vid) (* Need pure sid instead of Cil.new_sid for incremental, similar to vid in Cilfacade.create_var. We only add one return stmt per loop, so the hash from the functions vid should be unique. *)
```
sim642, line 126: Same thing here, `Hashtbl.hash` might not be positive. Although in this case we could just use `-fd.svar.vid`, no?

src/util/cilfacade.ml
jerhard: The idea is that in a subsequent run of Goblint, the variable will get the same `vid` again, if it is created again.

sim642: Right, but having our own counter assign `vid`s -1, -2, -3, ... would be fine then? It would avoid the issue of hash being potentially negative.

stilscher: Having our own counter would however require that the variables are created in the same order and this is not necessarily the case, is it?

src/framework/cfgTools.ml
stilscher: I think that should work.

src/util/cilfacade.ml
sim642: Ah, right, the point is to treat the variable name as unique, such that multiple creations yield the same `vid`. That's also why https://github.com/goblint/analyzer/pull/1312#discussion_r1447416392 had a problem.

Actually [`Hashtbl.hash`](https://v2.ocaml.org/api/Hashtbl.html#VALhash) returns a _non-negative_ integer, so my initial concern is a non-issue. We should have comments about these things though.

PR #1093, opened 2023-06-22
TUM Practical Course Summer 2023: Termination Analyses

Nimmseasy:
Implementation of a sound termination analyses for single-threaded programs.
We handle 2 possible causes for non-termination:
1. loops inside a function: We add a loop-counter variable to the CIL code after any loop head found in the CIL code and check it by assigning it in a special statment to another variable wich then is then locatet by goblints transfer functions. We run the apron analyses (either with polyhedra or octagon domain so far) to get the interval of our loop-counter Variable and see if its bounded => if its bounded we dont have an infinit loop.
For goto statements we cant proof non termination, if they jump up in the codelines.
2. recursion or call graph cycles: We build a context-sensitive call graph by defining a global invariant 
   (callee_function-> (callee-context -> (caller_function, caller_context) and filling it as we analyse the program, and then search 
   for cycles in this build callgraph.

If the program gets multithreaded we cant proof non termination.


michael-schwarz:
Thank you for this PR! I'll try to review it after my vacation (Wednesday), or before if I find the time

tests/regression/80-termination/15-complex-loop-combination-terminating.c
```diff
@@ -0,0 +1,144 @@
+// SKIP LOCAL_TERM PARAM: --set "ana.activated[+]" termination --enable warn.debug --set ana.activated[+] apron --enable ana.int.interval
```
jerhard, line 1: Why is this test skipped?

tests/regression/80-termination/09-complex-for-loop-terminating.c
```diff
@@ -0,0 +1,107 @@
+// SKIP LOCAL_TERM PARAM: --set "ana.activated[+]" termination --enable warn.debug --set ana.activated[+] apron --enable ana.int.interval --set ana.apron.domain polyhedra
```
jerhard, line 1: Is there a reason this test is skipped? Is it because Goblint cannot determine the precise verdict? In that case, it may be helpful to introduce an annotation for tests that indicates that termination is the proper verdict but that Goblint is currently unable to determine that.

jerhard (commented):
Thank you for your pull request! Nice work! It's great to see that you have implemented the handling of both loops and recursion, given the time constraints of the practical :)
I added some first comments with suggestions and some questions, that you may want to have a look at.

src/analyses/libraryFunctions.ml
```diff
@@ -302,6 +302,7 @@ let goblint_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__goblint_assert", special [__ "exp" []] @@ fun exp -> Assert { exp; check = true; refine = get_bool "sem.assert.refine" });
     ("__goblint_split_begin", unknown [drop "exp" []]);
     ("__goblint_split_end", unknown [drop "exp" []]);
+    ("__goblint_bounded", special [__ "exp"[]] @@ fun exp -> Assert { exp; check = true; refine = false });
```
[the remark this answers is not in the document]
sim642: There's also no documentation for it, so what is it even supposed to mean? All C expressions are bounded in value, so one might just assert that the expression is in the type bounds. But that's probably not what is meant here?

tests/regression/80-termination/01-simple-loop-terminating.c
```diff
@@ -0,0 +1,15 @@
+// LOCAL_TERM PARAM: --set "ana.activated[+]" termination --enable warn.debug --set ana.activated[+] apron --enable ana.int.interval --set ana.apron.domain polyhedra
```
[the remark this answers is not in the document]
LageTs: Until commit 29568d2b8e3c29442152a223d4f1e5086231f1c5 the keywords `TERM` and `NONTERM` were already used by the old termination implementation. Since that commit is now complete, I think we can switch to `TERM` and `NONTERM`. I just needed to use some wording to separate our termination from the old termination.

If you know a better way than `TERM` and `NONTERM` it would probably be fine for me/us.

runningGob.sh
```diff
@@ -0,0 +1,26 @@
...
+#./goblint $cfile_loops $options_apron --enable justcil > output.txt
+#./goblint $cfile_loops $options_apron --html 
+
+# run analysis, write cil output to file and enable visualization via html
+#./goblint -v $cfile_loops $options_term --enable justcil > output.txt
+./goblint $cfile_loops $options_term --html 
+
+# set up server to see visualizatino
+python3 -m http.server --directory result 8080
+#./goblint --enable dbg.debug tests/regression/55-loop-unrolling/01-simple-cases.c --enable ana.int.interval --set "ana.activated[+]" signs --enable justcil > output.txt
```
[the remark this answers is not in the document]
serenita: Yes, of course! Thanks for the reminder.

src/analyses/tutorials/signs.ml
```diff
@@ -59,7 +59,7 @@ struct
 
   (* This should now evaluate expressions. *)
   let eval (d: D.t) (exp: exp): SL.t = match exp with
-    | Const (CInt (i, _, _)) -> SL.top () (* TODO: Fix me! *)
+    | Const (CInt (i, _, _)) -> SL.of_int i (* TODO: Fix me! *)
     | Lval (Var x, NoOffset) -> D.find x d
     | _ -> SL.top ()
```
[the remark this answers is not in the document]
serenita: Done!

src/analyses/termination_new.ml
```diff
@@ -0,0 +1,116 @@
...
+    match q with
+    | Queries.MustTermLoop loop_statement ->
+      (match G.find_opt (`Lifted loop_statement) (ctx.global ()) with
+         Some b -> b
+       | None -> false)
+      && must_be_single_threaded_since_start ctx
+    | Queries.MustTermProg ->
+      G.for_all (fun _ term_info -> term_info) (ctx.global ())
+      && no_upjumping_gotos ()
+      && must_be_single_threaded_since_start ctx
```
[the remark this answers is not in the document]
serenita: Ok, that's good to know. Your suggested fix is basically what I wrote in Zulip (https://goblint.zulipchat.com/#narrow/stream/379241-tum-practical-summer23/topic/.E2.9C.94.20Check.20for.20thread.20creations.20efficiently/near/368214682), right?

.github/workflows/build_testing.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Setup OCaml ${{ matrix.ocaml-compiler }}
+        env:
+          OPAMLOCKED: locked
+        uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: ${{ matrix.ocaml-compiler }}
+          
+      - name: Install opam dependencies
+        run: opam install . --deps-only --locked
```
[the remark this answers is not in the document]
serenita: Ok, I got his confirmation. The file will be removed.

tests/regression/80-termination/01-simple-loop-terminating.c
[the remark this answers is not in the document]
jerhard: `TERM` and `NONTERM` are fine with me.

jerhard:
@Nimmseasy @serenita @LageTs Can you please include a description of your approach in the original post of this PR? This would be helpful for future generations of Goblint developers and also our coworkers in Tartu to better understand your code :)

src/analyses/termination_new.ml
```diff
@@ -0,0 +1,116 @@
+(** Work in progress *)
```
jerhard, line 1: How about naming this something like `terminationLoops` or `loopTermination`?

src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -4,6 +4,7 @@ open Analyses
```
michael-schwarz, line 4: The only thing touched in this file is unrelated whitespace. It should not be modified in this PR.

src/analyses/loop_termination.ml
```diff
@@ -0,0 +1,116 @@
...
+(*
+let loop_heads () =
+  let module FileCfg =
+  struct
+    let file = !Cilfacade.current_file
+    module Cfg = (val !MyCFG.current_cfg)
+  end in
+  let module WitnessInvariant = WitnessUtil.Invariant (FileCfg) in
+  WitnessInvariant.loop_heads (* TODO: Unused *)
+*)
```
michael-schwarz, line 18: What is this commented out code good for?

src/analyses/loop_termination.ml
```diff
@@ -0,0 +1,116 @@
...
+let upjumping_gotos : location list ref = ref []
+
+(** Indicates the place in the code, right after a loop is exited. *)
+let loop_exit : varinfo ref = ref (makeVarinfo false "-error" Cil.intType)
+
+let is_loop_counter_var (x : varinfo) =
+  VarToStmt.mem x !loop_counters
+
+let is_loop_exit_indicator (x : varinfo) =
+  x = !loop_exit
```
michael-schwarz, line 33: Iirc, CIL `varinfo` should not be compared with physical equality, as this may fail for the `type`s that are in it.

src/analyses/loop_termination.ml
```diff
@@ -0,0 +1,116 @@
...
+  include Analyses.IdentitySpec
+
+  let name () = "termination"
+
+  module D = Lattice.Unit
+  module C = D
+  module V = UnitV
+  module G = MapDomain.MapBot (Statements) (BoolDomain.MustBool)
+
+  let startstate _ = D.bot ()
```
michael-schwarz, line 72:
```suggestion
  let startstate _ = ()
```

makes it a bit more explicit what's going on.

src/analyses/loop_termination.ml
```diff
@@ -0,0 +1,116 @@
...
+  let assign ctx (lval : lval) (rval : exp) =
+    (* Detect assignment to loop counter variable *)
+    match lval, rval with
+      (Var y, NoOffset), Lval (Var x, NoOffset) when is_loop_exit_indicator y ->
+      (* Loop exit: Check whether loop counter variable is bounded *)
+      (* TODO: Move to special *)
+      let is_bounded = check_bounded ctx x in
+      let loop_statement = VarToStmt.find x !loop_counters in
+      let () = ctx.sideg () (G.add (`Lifted loop_statement) is_bounded (ctx.global ())) in
+      ctx.local
```
michael-schwarz, line 84:
```suggestion
      ctx.sideg () (G.add (`Lifted loop_statement) is_bounded (ctx.global ()));
      ()
```

src/analyses/loop_termination.ml
michael-schwarz: Same for all references to `ctx.local`.

src/analyses/loop_termination.ml
```diff
@@ -0,0 +1,116 @@
...
+  module D = Lattice.Unit
+  module C = D
+  module V = UnitV
+  module G = MapDomain.MapBot (Statements) (BoolDomain.MustBool)
+
+  let startstate _ = D.bot ()
+  let exitstate = startstate
+
+  let assign ctx (lval : lval) (rval : exp) =
+    (* Detect assignment to loop counter variable *)
```
michael-schwarz, line 76: Is it necessary to always do this or does it suffice to do this once at the end? During iteration, precision may be lost and only restored later during narrowing.

src/autoTune.ml
```diff
@@ -216,6 +216,7 @@ let focusOnSpecification () =
   | NoDataRace -> (*enable all thread analyses*)
     print_endline @@ "Specification: NoDataRace -> enabling thread analyses \"" ^ (String.concat ", " notNeccessaryThreadAnalyses) ^ "\"";
     enableAnalyses notNeccessaryThreadAnalyses;
+  | NoTermination -> ()
```
michael-schwarz, line 219: Maybe enable your termination analysis and one of the relational analyses here?

src/cdomains/intDomain.ml
```diff
@@ -3263,6 +3263,7 @@ struct
   let refine_with_incl_list ik a b = a
 
   let project ik p t = t
+
 end
 
 module SOverflowLifter (D : S) : SOverflow with type int_t = D.int_t and type t = D.t = struct
```
michael-schwarz, line 3269: whitespace change only, should not be part of PR.

src/cdomains/intDomain.mli
```diff
@@ -412,10 +412,10 @@ module IntervalSetFunctor(Ints_t : IntOps.IntOps): SOverflow with type int_t = I
...
-  sig
-    include Printable.S with type t = Z.t (* TODO: why doesn't this have a more useful signature like IntOps.BigIntOps? *)
-    val cast_to: Cil.ikind -> Z.t -> Z.t
-  end
+sig
+  include Printable.S with type t = Z.t (* TODO: why doesn't this have a more useful signature like IntOps.BigIntOps? *)
+  val cast_to: Cil.ikind -> Z.t -> Z.t
+end
 
 module Interval : SOverflow with type int_t = IntOps.BigIntOps.t
```
michael-schwarz, line 421: whitespace change only, should not be part of PR.

src/framework/analyses.ml
```diff
@@ -119,6 +118,100 @@ struct
     | x -> BatPrintf.fprintf f "<analysis name=\"fromspec\">%a</analysis>" printXml x
 end
 
+(* Tuple of fundec and S.C*)
+module T (Base1: Printable.S) (Base2: Printable.S) =
```
michael-schwarz, line 122: This seems to be mostly a `Printable.Prod`? We should be able to get some reuse there... Also `T` is a very generic name if you then have `fundec` and `context`

src/framework/analyses.ml
```diff
@@ -119,6 +118,100 @@ struct
...
+  include Printable.Std
+  type t = (Base1.t * Base2.t)
+
+  let fundec (a,_) = a
+  let context (_,b) = b
+  let equal (a1, b1) (a2, b2) = if (Base1.equal a1 a2 && Base2.equal b1 b2) then true else false
+  let show (a,b) = (Base1.show a) ^ (Base2.show b)
+  let name () = "Tuple"
+  let to_yojson x = `String (show x)
+  let relift (a,b) = (a,b) (*Todo: is this correct?*)
```
michael-schwarz, line 133: No, you should call the relift of the inner things here.

src/framework/analyses.ml
```diff
@@ -119,6 +118,100 @@ struct
...
+    <key>caller_context</key>\n<value>%a</value>\n\n
+    </map></value>\n" Base1.printXml a Base2.printXml b
+
+  let compare (a1,b1) (a2,b2) = (*Todo: is this ok?*)
+    if equal (a1, b1) (a2, b2) then 0
+    else(
+      let val_a a = if (a > 0) then 1 else -1 in
+      let val_b b = if (b > 0) then 3 else -3 in
+      val_a (Base1.compare a1 a2) + val_b (Base2.compare b1 b2)
+    )
```
michael-schwarz, line 147: Can this not be derived?

src/framework/analyses.ml
```diff
@@ -119,6 +118,100 @@ struct
...
+
+  let pretty () x = text (show x)
+  let hash (a,b) = Hashtbl.hash (Base1.hash a * Base2.hash b) (*Todo: is this ok?*)
```
michael-schwarz, line 150: Can this be derived?

src/framework/analyses.ml
```diff
@@ -119,6 +118,100 @@ struct
...
+      BatPrintf.fprintf f "<value>\n<set>";
+      iter (Base.printXml f) a;
+      BatPrintf.fprintf f "</set>\n</value>\n"
+  end
+
+  (* Make the given module Goupable*)
+  module C_Printable (C: Printable.S) =
+  struct
+    include C
+    include Printable.Std (* To make it Groupable *)
```
michael-schwarz, line 169: This seems confused, you are overwriting the definitions from `C` with standard ones here...

src/framework/analyses.ml
```diff
@@ -119,6 +118,100 @@ struct
...
+  module CMap =
+  struct
+    include MapDomain.MapBot (C_Printable (C)) (CSet)
+    let printXml f c = BatPrintf.fprintf f "<value><map>
+    <key>ContextTupleMap</key>\n
+    <value>%a</value>\n\n
+    </map></value>" printXml c
+  end
+
+  include Lattice.Lift2 (G) (CMap) (Printable.DefaultNames)
```
michael-schwarz, line 185: This all seems to be massively complicated, I think it should be possible to have this much more concise.

src/framework/analyses.ml
```diff
@@ -119,6 +118,100 @@ struct
...
+    | `Lifted1 x -> G.printXml f x
+    | `Lifted2 x -> BatPrintf.fprintf f "<analysis name=\"recTerm-context\">%a</analysis>" CMap.printXml x
+    | x -> BatPrintf.fprintf f "<analysis name=\"recTerm\">%a</analysis>" printXml x
+
+  let s = function
+    | `Bot -> G.bot ()
+    | `Lifted1 x -> x
+    | _ -> failwith "RecursionTerm.s"
+
+  let create_s s = `Lifted1 s
```
michael-schwarz, line 208: What is the difference between `create_s` and `create_spec`?

src/framework/constraints.ml
```diff
@@ -1692,6 +1694,142 @@ struct
...
+    Therefore:
+         g    -> {c' -> {(f, c)}}
+    in case f, c --> g, c'  *)
+
+struct
+  include S
+  module V =
+  struct
+    include GVarF(S.V)
+  end
```
michael-schwarz, line 1715:
```suggestion
  module V = GVarF(S.V)
```

Any reason why that doesn't work?

src/framework/control.ml
```diff
@@ -15,28 +15,30 @@ module type S2S = functor (X : Spec) -> Spec
 let spec_module: (module Spec) Lazy.t = lazy (
   GobConfig.building_spec := true;
   let arg_enabled = get_bool "ana.sv-comp.enabled" || get_bool "exp.arg" in
+  let arg_termination = List.mem "termination" (get_string_list "ana.activated") in (* check if loop termination analysis is enabled*)
```
michael-schwarz, line 18: Why `arg`? Are you constructing Abstract Reachability Graphs here?

src/framework/control.ml
```diff
@@ -617,35 +619,35 @@ struct
...
+         let skipped_statements from_node edge to_node =
+           CfgTools.CfgEdgeH.find_default skippedByEdge (from_node, edge, to_node) []
+         in
+
+         Transform.run_transformations file active_transformations
+           { ask ; must_be_dead ; must_be_uncalled ;
+             cfg_forward = Cfg.next ; cfg_backward = Cfg.prev ; skipped_statements };
       );
 
       lh, gh
```
michael-schwarz, line 653: All these places where only the indentation is changed should not be place of this PR. it makes the diff hard to review.

src/util/terminationPreprocessing.ml
```diff
@@ -0,0 +1,54 @@
+open GoblintCil
+include Printf
+
+module VarToStmt = Map.Make(CilType.Varinfo);; (* maps varinfos (= loop counter variable) to the statement of the corresponding loop*)
```
michael-schwarz, line 4:
```suggestion
module VarToStmt = Map.Make(CilType.Varinfo) (* maps varinfos (= loop counter variable) to the statement of the corresponding loop*)
```

michael-schwarz:
I left a few comments. If anything is unclear, just @ me in that comment, and I'll be happy to elaborate! And sorry if many of these things overlap with the things y'all discussed today.

src/analyses/loop_termination.ml
serenita: I considered using it later but probably am not going to do that. It will of course be removed in the end if it stays unused.

src/analyses/loop_termination.ml
```diff
@@ -0,0 +1,116 @@
...
+  upjumping_gotos.contents = []
+
+(** Checks whether a variable can be bounded *)
+let check_bounded ctx varinfo =
+  let open IntDomain.IntDomTuple in
+  let exp = Lval (Var varinfo, NoOffset) in
+  match ctx.ask (EvalInt exp) with
+  | `Top -> false
+  | `Lifted v -> not (is_top_of (ikind v) v)
+  | `Bot -> raise (PreProcessing "Loop variable is Bot")
```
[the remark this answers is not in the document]
serenita: Sorry, I don't completely understand what you mean. When I wrote this, I thought `\`Bot` would be the answer if the variable has not been initialized/ever seen yet (which would be an issue with the pre-processing). Currently, I'm not sure. What would you suggest to put there (in the `\`Bot` match branch)? To be honest, I can't think of any example where this exception was ever thrown.

src/analyses/loop_termination.ml
serenita: Done!

src/analyses/loop_termination.ml
serenita: Originally, we did this check directly after the loop. The reason we moved it into the loop was that the analysis might be unsound in cases like your "pathological loop".

src/analyses/loop_termination.ml
serenita: Done!

serenita: Done!

src/analyses/loop_termination.ml
serenita: Originally, we placed this check right after the loop. The reason we moved it inside the loop was that the analysis might be unsound in cases like your "pathological loop" otherwise.

src/analyses/termination.ml
```diff
@@ -5,22 +5,31 @@ open GoblintCil
 open Analyses
```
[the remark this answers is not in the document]
serenita: The `J: `-prefixed comments were written by Johanna in order for us all to better understand what's going on in that file. We already discussed with @jerhard that they will be removed or polished.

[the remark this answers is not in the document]
serenita: The first question is of course more relevant. But to be honest, I think we don't really know. Looking at the bottom of the filed, it seemed to me that the functionality of this old analysis is currently commented out anyway. I just tried reactivating it and get some compilation errors which was my suspicion already back then.

src/cdomains/intDomain.ml
serenita: I can restore this too.

src/cdomains/intDomain.mli
serenita: For some other files (like this one) it is more complicated. This change has been made by my ocp-indent, and when I try to revert it, I shall not commit because ocp-indent complains about the changes.

serenita: Should I force the commit anyway?

src/framework/control.ml
serenita: @michael-schwarz, please see my reply to your other comment regarding this issue.

src/framework/control.ml
SchiJoha: we just didn't get the naming convention here :)
Done

src/framework/analyses.ml
SchiJoha: We actually searched for a module that implements tuples, but we haven't found it... 
I now deleted module T and reused the Printable.Prod

SchiJoha: Done

src/framework/analyses.ml
SchiJoha: Done

src/framework/analyses.ml
SchiJoha: Done

src/framework/analyses.ml
SchiJoha: Done

src/framework/analyses.ml
SchiJoha: done

src/framework/analyses.ml
SchiJoha: We use this new module for:
include MapDomain.MapBot (C_Printable (C)) (CSet)

the problem is that MapDomain.MapBot needs a module of type Groupable for its first parameter. If we just write MapDomain.MapBot (C) (CSet), we get a compiler error, since the Groupable module is not fulfilled. Therefore, we thought extending C to be groupable might be the right choice here.

Is there a better way to make C groupable? @michael-schwarz

sim642: The point is to do it in the other order:
```ocaml
    include Printable.Std (* To make it Groupable *)
    include C
```
Because if `C` redefines a function, then you don't want `Std` to override it back to some default.

SchiJoha: ah I see! I wasn't aware of that! Thanks

SchiJoha: Done

src/framework/analyses.ml
SchiJoha: @Nimmseasy and I discussed this comment, but we can't think of an easier way to implement it right now. Do you have any hint or approach how it can be done more concisely?

src/analyses/termination_new.ml
serenita: Done!

src/autoTune.ml
serenita: @Nimmseasy took care of this.

tests/regression/80-termination/15-complex-loop-combination-terminating.c
michael-schwarz: Apron?

src/analyses/loop_termination.ml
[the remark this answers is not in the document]
michael-schwarz: raising an exception makes sense, it's just weird that it's a preprocessing exception here, when it has nothing to do with preprocessing.

src/analyses/loop_termination.ml
michael-schwarz: We discussed & resolved this at the meeting today.

src/framework/analyses.ml
michael-schwarz: We discussed this today/

src/analyses/everMultiThreaded.ml
```diff
@@ -0,0 +1,47 @@
...
+
+  (** Sets the global invariant to true when a thread is spawned *)
+  let threadspawn ctx lval f args fctx =
+    ctx.sideg () true;
+    ()
+
+  let query ctx (type a) (q: a Queries.t) : a Queries.result =
+    match q with
+    | Queries.IsEverMultiThreaded ->
+      (match ctx.global () with
```
jerhard, line 35: Do you query `IsEverMultiThreaded` anywhere?

serenita: We have enabled it now.

serenita: It's used by then loop termination analysis.

src/framework/analyses.ml
sim642: Since #1112 this trick can be removed altogether.

src/analyses/loopTermination.ml
```diff
@@ -0,0 +1,130 @@
...
+         with Not_found ->
+           (* This should not happen as long as __goblint_bounded is only used
+            * for this analysis. *)
+           ())
+      | _ -> ()
+    else ()
+
+  (** Checks whether a new thread was spawned some time. We want to discard
+   * any knowledge about termination then (see query function) *)
+  let must_be_single_threaded_since_start ctx =
```
jerhard, line 100: Does the `_since_start` suffix in the name still fit to the implementation? Maybe one can just remove that suffix in the name?

serenita: We can. I think it is still fitting as the information should not be confused with single-threadedness at a certain point of execution.

src/witness/svcompSpec.ml
```diff
@@ -16,6 +17,8 @@ let of_string s =
       NoDataRace
     else if global_not = "overflow" then
       NoOverflow
+    else if global_not = "no-termination" then
+      Termination
```
michael-schwarz, line 21: This is still incorrect. The formula for termination is `CHECK( init(main()), LTL(F end) )`.

michael-schwarz:
The results on the SV-COMP benchmarks are quite encouraging, there's only one case where it is unsound, but 460 where it works :tada: 

[term_results.zip](https://github.com/goblint/analyzer/files/12474798/term_results.zip)


The case where it is unsound is this one:


~~~C
int main(void) {
  int i, j;
 L0:
  i = 0;
 L1:
  while( __VERIFIER_nondet_int() && i < 1000000){
    i++;
  }
  if(i >= 100) STUCK: goto STUCK;
  j = 0;
 L2:
  while( __VERIFIER_nondet_int() && i < 1000000 ){
    i++;
    j++;
  }
  if(j >= 100) goto STUCK;
  __VERIFIER_assert( i < 200 );
  return 0;
}
~~~

michael-schwarz:
Something is phishy here, the output for the unsound example is as follows:

```
./goblint --conf conf/svcomp23_term.json --sets ana.specification /home/goblint2/svcomp/sv-benchmarks/c/properties/termination.prp --sets exp.architecture 32bit /home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i


--------------------------------------------------------------------------------


--sets is deprecated, use --set instead.
--sets is deprecated, use --set instead.
Option warning: termination analysis implicitly activates evermultithreaded analysis and set sem.int.signed_overflow to assume_none
SV-COMP specification: CHECK( init(main()), LTL(G ! no-termination) )
[Warning][Deadcode] Function 'reach_error' is uncalled: 1 LLoC (/home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i:12:1-12:159)
[Warning][Deadcode] Function 'assume_abort_if_not' is uncalled: 2 LLoC (/home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i:14:1-16:1)
[Warning][NonTerminating] Possibly non terminating loops:
  The program might not terminate! (Upjumping Goto) (/home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i:29:3-31:3)
[Warning][NonTerminating] Possibly non terminating loops:
  The program might not terminate! (Upjumping Goto) (/home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i:32:23-32:33)
[Warning][NonTerminating] Possibly non terminating loops:
  The program might not terminate! (Upjumping Goto) (/home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i:35:3-38:3)
[Warning][NonTerminating] Possibly non terminating loops:
  The program might not terminate! (Upjumping Goto) (/home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i:39:16-39:26)
[Warning][Deadcode] Function '__VERIFIER_assert' has dead code:
  on line 19 (/home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i:19-19)
[Warning][Deadcode] Logical lines of code (LLoC) summary:
  live: 16
  dead: 4 (3 in uncalled functions)
  total lines: 20
[Warning][Deadcode][CWE-570] condition '! cond' is always false (/home/goblint2/svcomp/sv-benchmarks/c/loop-invgen/string_concat-noarr.i:18:7-18:14)
SV-COMP result: true
```

michael-schwarz:
Ok, a partial the reason for the encouraging results seems to be that upjumping gotos currently do cause the sv-comp flag to be set to non-terminating.

michael-schwarz:
Requested @stilscher for review as the PR of her student #1145 builds on this.

michael-schwarz:
After also considering upjumping gotos, the number of successful tasks is down to 305, but we remain sound. Thus, it seems improving the handling of upjumping gotos is a worthwhile thing if we want to get more precise.

[results_newer.zip](https://github.com/goblint/analyzer/files/12483495/results_newer.zip)

src/analyses/loopTermination.ml
```diff
@@ -0,0 +1,109 @@
...
+           if not (is_bounded) then (
+             let msgs =
+               [(Pretty.dprintf
+                   "The program might not terminate! (Loop analysis)",
+                 Some (M.Location.CilLocation (Cilfacade.get_stmtLoc loop_statement))
+                );] in
+             M.msg_group Warning ~category:NonTerminating "Possibly non terminating loops" msgs);
+           ()
+         with Not_found ->
+           (* This should not happen as long as __goblint_bounded is only used
```
jerhard, line 75: If this is not expected to happen, there one may rethrow an exception here when it happens.

src/witness/witness.ml
```diff
@@ -442,6 +442,36 @@ struct
         in
         (module TaskResult:WitnessTaskResult)
       )
+    | Termination -> (* TODO: implement this properly*)
```
jerhard, line 445: What should be changed here?

src/witness/svcompSpec.ml
```diff
@@ -42,5 +45,6 @@ let to_string spec =
     | UnreachCall f -> "call(" ^ f ^ "())"
     | NoDataRace -> "data-race"
     | NoOverflow -> "overflow"
+    | Termination -> "no-termination"
```
jerhard, line 48: For termination, the `to_string` should be `CHECK( init(main()), LTL(F end) )`, see [comment above](  https://github.com/goblint/analyzer/pull/1093/files#r1307143458)

jerhard (changes requested):

src/analyses/loopTermination.ml
jerhard: Rethinking this, it may not be a good idea to throw an exception here, as one may be interested in analyzing already preprocessed programs that contain calls to `__goblint_bounded`.

src/util/terminationPreprocessing.ml
```diff
@@ -0,0 +1,65 @@
...
+        let v = (Cil.makeLocalVar fd name typ) in (*Not tested for incremental mode*)
+        let init_stmt = mkStmtOneInstr @@ Set (var v, min_int_exp, loc, eloc) in
+        let inc_stmt = mkStmtOneInstr @@ Set (var v, increm (Lval (var v)) 1, loc, eloc) in
+        let inc_stmt2 = mkStmtOneInstr @@ Set (var v, increm (Lval (var v)) 1, loc, eloc) in
+        let exit_stmt = mkStmtOneInstr @@ Call (None, f_bounded, [Lval (var v)], loc, locUnknown) in
+        (match b.bstmts with
+         | s :: ss ->   (*duplicate increment statement here to fix inconsistencies in nested loops*)
+           b.bstmts <- exit_stmt :: inc_stmt :: s :: inc_stmt2 :: ss;
+         | ss ->
+           b.bstmts <- exit_stmt :: inc_stmt :: ss;
```
sim642, line 50: What inconsistencies? This seems odd.

src/util/terminationPreprocessing.ml
```diff
@@ -0,0 +1,65 @@
...
+
+let extract_file_name s =                    (*There still may be a need to filter more chars*)
+  let ls = String.split_on_char '/' s in    (*Assuming '/' as path seperator*)
+  let ls = List.rev ls in
+  let s' = List.nth ls 0 in
+  let ls = String.split_on_char '.' s' in
+  let s' = List.nth ls 0 in
+  let without_spaces = String.split_on_char ' ' s' in
+  let s' = String.concat "" without_spaces in
+  s'
```
sim642, line 13: Why all this? Besides reimplementing standard path-manipulation functions, this is also error-prone. Basenames are not always unique.

src/util/options.schema.json
```diff
@@ -2099,6 +2099,12 @@
           "type": "boolean",
           "default": true
         },
+        "nonTerminating": {
+          "title": "warn.nonTerminating",
+          "description": "Non Termination warning",
+          "type": "boolean",
+          "default": true
+        },
```
sim642, line 2107: This naming is inconsistent with the rest. `warn.termination` seems good enough. We don't have separate categories for negations, that's what we use message severities for.

src/util/cilfacade.ml
```diff
@@ -85,6 +85,18 @@ let do_preprocess ast =
...
+let register_preprocess_cil name visitor_fun =
+  visitors_cil := !visitors_cil @ [name, visitor_fun]
+
+let do_preprocess_cil ast =
+  let f fd (name, visitor_fun) =
+    (* this has to be done here, since the settings aren't available when register_preprocess is called *)
+    if List.mem name (get_string_list "ana.activated") then
+      ignore @@ visitCilFunction (visitor_fun fd) fd
+  in
+  iterGlobals ast (function GFun (fd,_) -> List.iter (f fd) !visitors_cil | _ -> ())
```
sim642, line 99: Should just fix `register_preprocess` then. It's only used by the old termination analysis, which probably also needs this to happen before CFG, no? It might be that `do_preprocess` was at one point accidentally moved around and nothing we test uses it.

src/framework/control.ml
```diff
@@ -36,6 +37,7 @@ let spec_module: (module Spec) Lazy.t = lazy (
                Also must be outside of deadcode, because deadcode splits (like mutex lock event) don't pass on tokens. *)
             |> lift (get_bool "ana.widen.tokens") (module WideningTokens.Lifter)
             |> lift true (module LongjmpLifter)
+            |> lift termination_enabled (module RecursionTermLifter) (* Always activate the recursion termination analysis, when the loop termination analysis is activated*)
```
sim642, line 40: `RecursionTermLifter` being the outermost made me suspicious but maybe it has to be for some reason? Does this ordering correctly account for how `longjmp` interacts with termination?

src/framework/constraints.ml
```diff
@@ -1694,6 +1694,162 @@ struct
...
+=
+(* two global invariants:
+   - V -> G
+     Needed to store the previously built global invariants
+   - fundec -> Map (S.C) (Set (fundec * S.C))
+     The second global invariant maps from the callee fundec to a map, containing the callee context and the caller fundec and context.
+     This structure therefore stores the context-sensitive call graph.
+     For example:
+      let the function f in context c call function g in context c'.
+      In the global invariant structure it would be stored like this: g -> {c' -> {(f, c)}}
```
sim642, line 1711: Why not use `fundec * S.C -> Set (fundec * S.C)` as the invariant?

src/framework/constraints.ml
```diff
@@ -1694,6 +1694,162 @@ struct
...
+      let tup: (fundec * S.C.t) = (fd_r, c_r) in
+      let t = CallGraphSet.singleton (tup) in
+      side_context ctx.sideg fd_e (c_e) t;
+      S.combine_env (conv ctx) r fe f args fc es f_ask
+    else
+      S.combine_env (conv ctx) r fe f args fc es f_ask
+  let combine_assign ctx = S.combine_assign (conv ctx)
+  let special ctx = S.special (conv ctx)
+  let threadenter ctx = S.threadenter (conv ctx)
+  let threadspawn ctx lv f args fctx = S.threadspawn (conv ctx) lv f args (conv fctx)
```
sim642, line 1846: This doesn't handle calls that create threads. What happens if a thread recursively creates itself?

src/framework/constraints.ml
```diff
@@ -1694,6 +1694,162 @@ struct
...
+        CallGraphSet.iter (fun to_call ->
+            iter_call new_path_visited_calls to_call
+          ) callers;
+      end
+    in
+    let gmap = G.callGraph (ctx.global (v)) in
+    CallGraphMap.iter(fun key value ->
+        let call = (v', key) in
+        iter_call LS.empty call
+      ) gmap (* try all fundec + context pairs that are in the map *)
```
sim642, line 1791: The `G.callGraph`, ... `CallGraphSet.iter` logic is duplicated here. If the entry point of the recursive function were slightly different, this duplication could maybe be avoided.

Deadlock analysis does something similar but doesn't have such duplication.

src/autoTune.ml
```diff
@@ -216,6 +216,7 @@ let focusOnSpecification () =
   | NoDataRace -> (*enable all thread analyses*)
     print_endline @@ "Specification: NoDataRace -> enabling thread analyses \"" ^ (String.concat ", " notNeccessaryThreadAnalyses) ^ "\"";
     enableAnalyses notNeccessaryThreadAnalyses;
+  | Termination -> ()
```
sim642, line 219: This should activate the termination analysis (and Apron I guess?).

src/analyses/loopTermination.ml
```diff
@@ -0,0 +1,109 @@
...
+           let is_bounded = check_bounded ctx x in
+           ctx.sideg () (G.add (`Lifted loop_statement) is_bounded (ctx.global ()));
+           (* In case the loop is not bounded, a warning is created. *)
+           if not (is_bounded) then (
+             let msgs =
+               [(Pretty.dprintf
+                   "The program might not terminate! (Loop analysis)",
+                 Some (M.Location.CilLocation (Cilfacade.get_stmtLoc loop_statement))
+                );] in
+             M.msg_group Warning ~category:NonTerminating "Possibly non terminating loops" msgs);
```
sim642, line 72: One-message group should just be a normal message.

sim642 (changes requested):

sim642:
There seems to be a lot of CIL-specific trouble with this and gotos seem to be an issue. Why not find the loops from CFG then? `WitnessUtil.find_loop_heads` already does that and it works for arbitrarily messy nests of gotos.

The criterion of whether a goto jumps up or down in program source line number is a surprisingly non-semantic property. In fact, `#line` pragmas can arbitrarily mess with line numbers, so that doesn't seem like a reliable criterion for semantic behavior.

src/analyses/everMultiThreaded.ml
```diff
@@ -0,0 +1,47 @@
+(** Analysis to register whether any additional thread has ever been spawned ([evermultithreaded]). *)
```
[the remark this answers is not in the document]
serenita: Actually, using `MustBeSingleThreaded {since_start = true}` was our initial approach. But the answer to a `MustTermProg` query should not depend on the current program location.

See https://github.com/goblint/analyzer/pull/1093#discussion_r1242429988

src/analyses/loopTermination.ml
```diff
@@ -0,0 +1,109 @@
...
+    if not (!single_thread) then (
+      M.warn ~category:NonTerminating "The program might not terminate! (Multithreaded)\n"
+    )
+
+  (** Recognizes a call of [__goblint_bounded] to check the EvalInt of the
+   * respective loop counter variable at that position. *)
+  let special ctx (lval : lval option) (f : varinfo) (arglist : exp list) =
+    if !AnalysisState.postsolving then
+      match f.vname, arglist with
+        "__goblint_bounded", [Lval (Var x, NoOffset)] ->
```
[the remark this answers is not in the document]
serenita: In line 62 we need the `varinfo` to retrieve the related loop statement.

src/analyses/loopTermination.ml
```diff
@@ -0,0 +1,109 @@
...
+             let msgs =
+               [(Pretty.dprintf
+                   "The program might not terminate! (Loop analysis)",
+                 Some (M.Location.CilLocation (Cilfacade.get_stmtLoc loop_statement))
+                );] in
+             M.msg_group Warning ~category:NonTerminating "Possibly non terminating loops" msgs);
+           ()
+         with Not_found ->
+           (* This should not happen as long as __goblint_bounded is only used
+            * for this analysis. *)
```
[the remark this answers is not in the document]
serenita: The `Not_found` can be raised by `VarToStmt.find x !loop_counters` in line 62 in case we encounter a call of `__goblint_bounded` where the `x` in the argument `[Lval (Var x, NoOffset)]` is not registered as a loop counter variable.

src/util/terminationPreprocessing.ml
serenita: I don't remember the details. Probably, @Nimmseasy knows best about it.

src/framework/constraints.ml
serenita: Currently, this should be covered by the fact that any multi-threaded program is reported as possibly non-terminating. From what you say, I think that we would only fail in not reporting recursion as a cause for non-termination then.

serenita:
> There seems to be a lot of CIL-specific trouble with this and gotos seem to be an issue. Why not find the loops from CFG then? `WitnessUtil.find_loop_heads` already does that and it works for arbitrarily messy nests of gotos.
We considered using `WitnessUtil.find_loop_heads`. But it seems natural to locate the loops in CIL when we need to use CIL functionality to insert new variables into certain places in the loops anyway.
> The criterion of whether a goto jumps up or down in program source line number is a surprisingly non-semantic property. In fact, `#line` pragmas can arbitrarily mess with line numbers, so that doesn't seem like a reliable criterion for semantic behavior.
I'm sorry, I don't exactly understand what you mean by that.

src/analyses/everMultiThreaded.ml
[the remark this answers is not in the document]
sim642: I see. Since this is such a tiny analysis, I think it could even be done within the existing threadflag analysis. It shouldn't incur any extra penalty to always have it, especially if it's never queried.

src/analyses/loopTermination.ml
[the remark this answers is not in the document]
sim642: Yes, but the comment says it should never happen in the first place, unless there's some bug. Currently this would hide the bug, but it would be better to just not catch it at all. If there's ever an issue, Goblint will crash and we'll know that something with `loop_counters` is buggy.

sim642:
Due to the module moves in #1206 GitHub shows conflicts that are "too complex" to solve in the web editor. Actually you can locally just merge master into this branch and there shouldn't be any conflicts. GitHub's merge thing seems to have some move detection disabled that git should do by default.

tests/regression/78-termination/29-do-while-continue-nonterminating.c
```diff
@@ -0,0 +1,22 @@
...
+  {
+    printf("Inside the do-while loop\n");
+    i++;
+
+    if (i % 2)
+    {
+      printf("Continue as %i is odd\n", i);
+      continue;
+    }
+  } while (i >= 2); // NONTERMLOOP termination analysis shall mark beginning of while as non-terminating loop
```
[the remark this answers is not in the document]
michael-schwarz: Same question as above. How can a claim of "we don't know if it terminates" be unsound?

tests/regression/78-termination/37-recursion-nonterminating.c
```diff
@@ -0,0 +1,25 @@
...
+  if (n == 30)
+  {
+    printf("Terminating recursion\n");
+    return;
+  }
+
+  printf("Recursive call with n = %d\n", n);
+
+  // Recursive call: Decrement n and call the function again
+  recursiveFunction(n - 1);
```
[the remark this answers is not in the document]
michael-schwarz: Same question as in the other cases.

src/framework/control.ml
michael-schwarz: It is outermost because it should observe the control contexts and construct a call graph. It does indeed currently not account for loops caused by invocations of `longjmp`. For now, I'll add a check that makes no claims in case of longjmp.

michael-schwarz: I fixed the issue with `longjmp` now.

src/analyses/loopTermination.ml
michael-schwarz: Will these still get grouped then? Currently messages about multiple loops (and any other reason) are neatly nested in one group.

sim642: These one-group messages don't get merged in any way (there's no post-processing). The neatly grouped warning is outputted from `control.ml` for some other thing where a non-singleton group is explicitly constructed for it.

src/framework/constraints.ml
michael-schwarz: Fixed!

tests/regression/78-termination/14-do-while-nonterminating.c
```diff
@@ -0,0 +1,16 @@
...
+
+int main()
+{
+  int i = 1;
+
+  do
+  {
+    printf("Inside the do-while loop\n");
+    i++;
+  } while (i >= 2); // NONTERMLOOP termination analysis shall mark while as non-terminating loop
```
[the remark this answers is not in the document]
sim642: > In what sense? The claim here is not that the program never terminates. Instead, we claim here that we do not know that it must terminate.

True, I think I got confused by the negated annotations (`NORACE` is always a sure claim). Maybe these annotations should also be like `RACE` and `RACE!` to distinguish exact annotations from those that are wrong but we only want an imprecise result at this point.

src/framework/constraints.ml
michael-schwarz: Fixed.

src/analyses/loopTermination.ml
```diff
@@ -0,0 +1,107 @@
...
+(** Contains all loop counter variables (varinfo) and maps them to their corresponding loop statement. *)
+let loop_counters : stmt VarToStmt.t ref = ref VarToStmt.empty
+
+(** Checks whether a variable can be bounded. *)
+let check_bounded ctx varinfo =
+  let open IntDomain.IntDomTuple in
+  let exp = Lval (Var varinfo, NoOffset) in
+  match ctx.ask (EvalInt exp) with
+  | `Top -> false
+  | `Lifted v -> not (is_top_of (ikind v) v)
```
michael-schwarz, line 20: This is now problematic as far as I can see. We execute `set_string "sem.int.signed_overflow" "assume_none";` in `maingoblint.ml` and still rely on an overflow yielding top here. This may currently work due to some peculiarities of the implementation, but there is no reason to believe that it always will.

src/analyses/termination.ml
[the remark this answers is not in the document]
michael-schwarz: I have now removed the old analysis.

src/witness/witness.ml
michael-schwarz: There is nothing to be done here, as correctness witnesses are not need for termination.

michael-schwarz:
I have now resolved all issues that are older than October. At some point we need to split things into new issues that we do not intend to address before SV-COMP such as a more proper handling of upjumping gotos.

sim642:
> have now resolved all issues that are older than October

That conveniently excludes my latest review from Oct 4... And that's excluding the dubious upjumping logic that I'm fine with postponing because SV-COMP tasks may not have `#line` directives.

michael-schwarz:
> have now resolved all issues that are older than October

> That conveniently excludes my latest review from Oct 4.

Sorry, should've been more clear: I only wrote this so people don't need to unfold the over 300 comments from before October (that are now all resolved) and click through all of them.
Some of your comments (as well as Julian's and mine) need to be resolved before we can merge.

src/util/terminationPreprocessing.ml
michael-schwarz: I simplified it now by keeping the line numbers plus a unique counter within this function.

src/analyses/everMultiThreaded.ml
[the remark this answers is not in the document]
michael-schwarz: I moved it to `threadflag`.

src/analyses/loopTermination.ml
```diff
@@ -0,0 +1,109 @@
...
+open Analyses
+open GoblintCil
+open TerminationPreprocessing
+
+(** Stores the result of the query whether the program is single threaded or not
+    since finalize does not have access to ctx. *)
+let single_thread : bool ref = ref false
+
+(** Contains all loop counter variables (varinfo) and maps them to their corresponding loop statement. *)
+let loop_counters : stmt VarToStmt.t ref = ref VarToStmt.empty
```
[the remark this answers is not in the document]
michael-schwarz: I removed `single_thread` as it was unneeded and added a check to fail if termination and incremental analysis are activated at the same time (The transformation would require additional considerations for incremental analysis).

src/util/cilfacade.ml
```diff
@@ -699,3 +711,7 @@ let add_function_declarations (file: Cil.file): unit =
   let fun_decls = List.filter_map declaration_from_GFun functions in
   let globals = upto_last_type @ fun_decls @ non_types @ functions in
   file.globals <- globals
+
+(** Contains the locations of the upjumping gotos and the respective functions
+ * they are being called in. *)
+let upjumping_gotos : (location * fundec) list ref = ref []
```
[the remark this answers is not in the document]
michael-schwarz: I changed the data structure to use a map from fundec to a set of locations. 

> And I don't see why the keys are just locations (which come with almost no guarantees), when they could be the actual goto statements themselves.

The locations are only used to emit warnings, as we have no good way to look up whether a goto is live (they are gone after CFG construction, so looking them up in our hashtable will always yield `Bot`). After SV-COMP we probably want to refactor this anyway, so I think this should do for now.

scripts/update_suite.rb
```diff
@@ -306,6 +330,21 @@ def parse_tests (lines)
...
+        todo << -1
+      when /TERM/
+        tests[-1] = "term"
+        todo << -1
+      end
+    when /NONTERM/
+      tests[-1] = "nonterm"
+    when /TERM/
+      tests[-1] = "term"
+    end
```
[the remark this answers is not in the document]
jerhard: I deduplicated the logic

scripts/update_suite.rb
```diff
@@ -145,29 +145,33 @@ def collect_warnings
...
+                  when /^\[Info\]/                                       then "warn"
+                  when /^\[Success\]/                                    then "success"
+                  when /(Upjumping Goto)/                                then "goto"
+                  when /(Fundec \w+ is contained in a call graph cycle)/ then "fundec"
+                  when /(Loop analysis)/                                 then "loop"
+                  when /\[Debug\]/                                       then next # debug "warnings" shouldn't count as other warnings (against NOWARN)
+                  when /^  on line \d+ $/                                then next # dead line warnings shouldn't count (used for unreachability with NOWARN)
+                  when /^  on lines \d+..\d+ $/                          then next # dead line warnings shouldn't count (used for unreachability with NOWARN)
+                  else "other"
+                end
```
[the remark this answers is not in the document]
jerhard: I changed the indentation back of this block.

lib/goblint/runtime/include/goblint.h
```diff
@@ -6,3 +6,5 @@ void __goblint_assume_join(/* pthread_t thread */); // undeclared argument to av
 
 void __goblint_split_begin(int exp);
 void __goblint_split_end(int exp);
+
+void __goblint_bounded(int exp);
```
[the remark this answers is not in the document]
jerhard: I now changed it to `unsigned long long`. Introducing a loop counter of a signed type and at the same time using `sem.int.signed_overflow assume_none` as it was done previously seemed a bit iffy, as the no-overflow assumption would afffect the counter. Using a unsigned type for the counting resolves that problem. There are some test cases that do no longer pass then, however.

src/util/terminationPreprocessing.ml
jerhard: I changed the code to only introduce one increment statement, as I could not think of a case where the additional would be necessary. There was also no test case that became unsound due to this change. I also started a run on sv-comp.

src/analyses/loopTermination.ml
jerhard: The type of the counter variables introduced is now unsigned, and thus they are not affected by the setting of `sem.int.signed_overflow assume_none`. So it should no longer be a problem.

src/goblint_lib.ml
```diff
@@ -124,6 +124,7 @@ module RelationPriv = RelationPriv
 module ThreadEscape = ThreadEscape
 module PthreadSignals = PthreadSignals
 module ExtractPthread = ExtractPthread
+module LoopTermination = LoopTermination
```
sim642, line 127: This isn't a concurrency analysis, so it should go to below where `Termination` used to be.

conf/svcomp23.json
```diff
@@ -70,7 +70,8 @@
         "congruence",
         "octagon",
         "wideningThresholds",
-        "loopUnrollHeuristic"
+        "loopUnrollHeuristic",
+        "termination"
```
sim642, line 74: This shouldn't change the svcomp23 conf, but the general svcomp one only.

jerhard: I changed the configs accordingly.

src/goblint_lib.ml
jerhard: I moved it to the analyses / other category, where `Termination` previously was.

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -196,15 +196,14 @@ struct
   let assert_type_bounds ask rel x =
     assert (RD.Tracked.varinfo_tracked x);
     match Cilfacade.get_ikind x.vtype with
-    | ik when not (IntDomain.should_ignore_overflow ik) -> (* don't add type bounds for signed when assume_none *)
+    | ik -> (* don't add type bounds for signed when assume_none *)
```
sim642, line 199: This comment is now wrong.

But also, what's the reason for removing this? Weren't the loop counters changed to `unsigned long long`?

sim642:
What's with the test failures in CI?

src/analyses/apron/relationAnalysis.apron.ml
jerhard: Thanks, I removed the comment. 

The reason for removing this was that the termination for some simple examples could not be established, when the loop counter in the original program was some `signed int` variable that was decremented. See this [example](https://github.com/goblint/analyzer/blob/d9f39105b35e4c65ac01b457cdb19ab82ac17c87/tests/regression/78-termination/50-decreasing-signed-int.c). 

With this change, we now succeed to show termination on 359 programs instead of 303 in sv-comp.

sim642:
> What's with the test failures in CI?

I see that the test failures are timeouts somehow related to function size. But our regression tests are tiny. Is this termination analysis really that inefficient that it struggles so much on 100-line programs?

jerhard:
>I see that the test failures are timeouts somehow related to function size. But our regression tests are tiny. Is this termination analysis really that inefficient that it struggles so much on 100-line programs?

The analysis relies on our relational analyses, and for most test cases uses polyhedra. The termination analysis introduces one counter variable per loop, so a polyhedra-based termination analysis does scale well with functions containing multiple loops.

sim642:
I suppose we'll have to implement variable packing which everyone else has been doing for decades.

If nothing concerning shows up in our own preruns with termination analysis, I think this should be good enough to finally merge.

sim642 (approved):

jerhard (commented):
I had to make a change so that the termination analysis is activated earlier by the autotuner if the autotuner is active and we have to analyze for termiation, as the termiation analysis needs to be activated before the preprocessing is done (to instrument the program with loop counter variables).

Additionally I had to add a handling if the loop unrolling encounters a all to `goblint_bounded` (the special function where the boundedness of the loop is checked); now loops that contain such a check are also unrolled.

jerhard:
I reran this PR with the `svcomp.json` configuration on the latest version of sv-comp termination tasks and this yields 567 correct verdicts (and no incorrect ones). The increased number of correct tasks may be due to the different configuration (previous runs did not use the autotuner).

Regarding sv-comp: the programs analyzed are contained some artificial loop counters. Since the Termination category is excluded from the validation of correctness witnesses, this should not be a problem, right?

jerhard (approved):

sim642:
All the non-Apron CI jobs are now failing on master: https://github.com/goblint/analyzer/actions/runs/6917734602. A condition might be missing somewhere.

michael-schwarz:
Should be fixed, there was a check missing to only run the termination regression tests in configs in which Apron is present.

PR #1094, opened 2023-06-27
Memory Out-Of-Bounds Access Analysis

mrstanb:
This PR is for the addition of an analysis for the detection of memory out-of-bounds (OOB) accesses.


mrstanb:
I did a rewrite of the majority of the logic that I had previously added for the memory OOB access checks. Still need to try and make it more precise.

Also, for some reason I can't get the `BlobSize` query for a pointer, which has dynamically allocated memory, to give me a meaningful result (i.e., not `top`). We should probably discuss this and see what's wrong there. I'm pretty sure I'm doing something wrong in my code.

@michael-schwarz Nonetheless, feel free to take a look at the PR and leave some feedback :)

mrstanb:
Did yet another rewrite of the logic for this PR:
* `make test` runs through successfully.
* An open issue that's left is handling memory OOB accesses due to loops, such as the one in test case `77/03`

Feel free to review

michael-schwarz (changes requested):
We discussed that this should happen at he address level, potentially using the mechanism by @jerhard from #967. Also, it should only warn for dereferences, and not already when constructing such invalid pointers.

src/analyses/memOutOfBounds.ml
```diff
@@ -0,0 +1,187 @@
...
+    ctx.local
+
+  let special ctx (lval:lval option) (f:varinfo) (arglist:exp list) : D.t =
+    Option.iter (fun x -> check_lval_for_oob_access ctx x) lval;
+    List.iter (fun arg -> check_exp_for_oob_access ctx arg) arglist;
+    ctx.local
+
+  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (callee_local:D.t) (f_ask:Queries.ask) : D.t =
+    Option.iter (fun x -> check_lval_for_oob_access ctx x) lval;
+    List.iter (fun arg -> check_exp_for_oob_access ctx arg) args;
```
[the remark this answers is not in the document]
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1094/commits/6ed476e3216d808fb6e6a16a6b15ca19c93efbfd

src/analyses/memOutOfBounds.ml
```diff
@@ -0,0 +1,336 @@
...
+              (*
+                Offset should be the same for all elements in the points-to set.
+                Hence, we can just pick one element and obtain its offset
+                TODO: Does this make sense?
+              *)
+              let (_, o) = VDQ.LS.choose a in
+              let rec to_int_dom_offs = function
+                | `NoOffset -> `NoOffset
+                | `Field (f, o) -> `Field (f, to_int_dom_offs o)
+                | `Index (i, o) -> `Index (VDQ.ID.unlift (fun x -> x) @@ ctx.ask (Queries.EvalInt i), to_int_dom_offs o)
```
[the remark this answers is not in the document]
mrstanb: Should be resolved in https://github.com/goblint/analyzer/pull/1094/commits/37fb4e8173a3e1d1b3b6b404219bc9b681e01721

mrstanb:
> We discussed that this should happen at he address level, potentially using the mechanism by @jerhard from https://github.com/goblint/analyzer/pull/967. Also, it should only warn for dereferences, and not already when constructing such invalid pointers.

Thanks for the tips and ideas! I updated the analyis, s.t.:
* We now only warn upon dereferences
* We check offset at the address level

There's one thing which is bugging me currently, namely:
* Queries.BlobSize` answers with `bot` if there's some offset in a pointer's abstract address set
* Hence, when we do pointer arithmetic, even when we're within the memory bounds, we still get a WARN, because there is some offset from the pointer's start address and we're comparing something with `bot` which is always true

Should we rethink the `Queries.BlobSize` query answer in `base.ml` again?

P.S. Feel free to drop a review :)

mrstanb:
> Should we rethink the `Queries.BlobSize` query answer in `base.ml` again?

After a discussion with @michael-schwarz, I added the possibility in `Queries.BlobSize` to ask for a blob's size from the base address that contains the blob. The idea behind is to be able to optionally ignore address offsets in `Queries.BlobSize` and in particular to be able to avoid `Queries.BlobSize` giving us `bot` when asking for something has an offset from its base abstract address.

mrstanb:
I also added a check for implicit dereferencing of pointers in special functions, similar to how it's done in #1123. I also added a test case for this which is passing after these changes.

mrstanb:
The new test case `77/05` for checking implicit dereferencing seems to be failing only on the MacOS CI runner (for line 10 of the test case):
* It's failing for a call to `memset` by emitting a warning when `//NOWARN` is expected
* However I can't see anywhere in the logs any of the warnings from this PR for this particular line in the test case which is failing

Is there a way to enable more logs for the MacOS runner? I checked the raw logs, but still couldn't find any warnings for the line which is failing. I don't have a MacOS machine to test it on either, unfortunately.

Could it also be that `memset` is causing some other warning on MacOS?

michael-schwarz:
On Linux vs the things in the logs for OS X, the only difference I see is this

```
[Info][Imprecise] Invalidating expressions: AddrOf(Var(tmp___0, NoOffset)) (05-oob-implicit-deref.c:10:4-10:35)
```

I guess our CI is stupid and interprets this as an actual warning? As to why it is different on OS X, I seem to recall that OS X uses a higher fortification level by default.

If I manually rewrite it as

```
    size_t  tmp = __builtin_object_size(ptr, 0); //NOWARN
    __builtin___memset_chk(ptr, 0, four * sizeof(int), tmp); //NOWARN
```

I get the same warning that we see in the CI also in Linux. The CI seems to take this invalidation warning to be a violation of the `//NOWARN` annotation. I'll make an issue for it, this is stupid.

michael-schwarz:
As a workaround, I would suggest disabling the `info` warnings category for this test with `--disable warn.info`, leaving a comment that this is not a longterm solution.

src/analyses/memOutOfBounds.ml
```diff
@@ -0,0 +1,344 @@
+open GoblintCil
+open Analyses
+open MessageCategory
+
+module AS = AnalysisState
+module VDQ = ValueDomainQueries
+
+module Spec =
```
michael-schwarz, line 8: Maybe add a comment that this is an analysis only to be able to have a separation of concerns and to be able to conveniently turn it on or off, but that it doesn't track any internal state.

michael-schwarz (commented):
Other than the small remarks, LGTM!

mrstanb:
> As a workaround, I would suggest disabling the `info` warnings category for this test with `--disable warn.info`, leaving a comment that this is not a longterm solution.

This is now added in https://github.com/goblint/analyzer/pull/1094/commits/2b4549905c80bbca10d82e04a7d2f21b9978143e

src/analyses/memOutOfBounds.ml
```diff
@@ -0,0 +1,344 @@
...
+struct
+  include Analyses.IdentitySpec
+
+  module D = Lattice.Unit
+  module C = D
+
+  (* TODO: Check out later for benchmarking *)
+  let context _ _ = ()
```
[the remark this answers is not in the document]
mrstanb: Removed the `TODO` in https://github.com/goblint/analyzer/pull/1094/commits/924130132a766149cc97745629fd52e08ec6b798

src/analyses/memOutOfBounds.ml
mrstanb: Added an explanation comment in https://github.com/goblint/analyzer/pull/1094/commits/924130132a766149cc97745629fd52e08ec6b798

src/domains/queries.ml
```diff
@@ -91,7 +91,9 @@ type _ t =
   | EvalStr: exp -> SD.t t
   | EvalLength: exp -> ID.t t (* length of an array or string *)
   | EvalValue: exp -> VD.t t
-  | BlobSize: exp -> ID.t t (* size of a dynamically allocated `Blob pointed to by exp *)
+  | BlobSize: exp * bool -> ID.t t
```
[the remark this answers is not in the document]
mrstanb: Thanks for the tip! I now made `Queries.BlobSize` to work with a record in place of the tuple that was used. Cf. https://github.com/goblint/analyzer/pull/1094/commits/0bae455057d108b3d2e5f368622384ccdf9f8d44

michael-schwarz:
It seems like the current implementation gets confused by string constants, causing quite a few normally simple CWE tests to fail, because they contain code like this:

~~~C
// PARAM: --set ana.activated[+] memOutOfBounds --enable ana.int.interval
#include <stdlib.h>
#include <stdio.h>

void myPrint(char *str) {
    printf("%s", str);
}

    myPrint("starting test");
    // Some stuff
    myPrint("ending test");    
}
~~~

```
[Warning][Unknown] Pointer str has an empty points-to-set (tests/regression/77-mem-oob/04-oob-deref-after-ptr-arith.c:6:5-6:22)
[Warning][Behavior > Undefined > MemoryOutOfBoundsAccess][CWE-823] Size of pointer str not known. Memory out-of-bounds access might occur due to pointer arithmetic (tests/regression/77-mem-oob/04-oob-deref-after-ptr-arith.c:6:5-6:22)
```

This should be easy to fix as we have a special representation of such string constants in the address domain, for which we could just consider all accesses to be within bound.

mrstanb:
> It seems like the current implementation gets confused by string constants, causing quite a few normally simple CWE tests to fail, because they contain code like this:
> 
> ```c
> // PARAM: --set ana.activated[+] memOutOfBounds --enable ana.int.interval
> #include <stdlib.h>
> #include <stdio.h>
> 
> void myPrint(char *str) {
>     printf("%s", str);
> }
> 
> int main(int argc, char const *argv[]) {
>     myPrint("starting test");
>     // Some stuff
>     myPrint("ending test");    
> }
> ```
> 
> ```
> [Warning][Unknown] Pointer str has an empty points-to-set (tests/regression/77-mem-oob/04-oob-deref-after-ptr-arith.c:6:5-6:22)
> [Warning][Behavior > Undefined > MemoryOutOfBoundsAccess][CWE-823] Size of pointer str not known. Memory out-of-bounds access might occur due to pointer arithmetic (tests/regression/77-mem-oob/04-oob-deref-after-ptr-arith.c:6:5-6:22)
> ```
> 
> This should be easy to fix as we have a special representation of such string constants in the address domain, for which we could just consider all accesses to be within bound.

This should be now fixed by commit https://github.com/goblint/analyzer/commit/3258cce90b88ea14db99c5e6466b43b7fd8a01a5 in the `staging_memsafety` branch. If you think it makes sense, we can cherry-pick this, as well as other commits from the `staging-memsafety` branch, which have introduced some changes and fixes to `memOutOfBounds.ml`

michael-schwarz:
> If you think it makes sense, we can cherry-pick this, as well as other commits from the staging-memsafety branch, which have introduced some changes and fixes to memOutOfBounds.ml

Probably a good idea, yes!

mrstanb:
> This should be now fixed by commit [3258cce](https://github.com/goblint/analyzer/commit/3258cce90b88ea14db99c5e6466b43b7fd8a01a5) in the `staging_memsafety` branch. If you think it makes sense, we can cherry-pick this, as well as other commits from the `staging-memsafety` branch, which have introduced some changes and fixes to `memOutOfBounds.ml`

We discussed that these changes will be taken after this PR's branch is merged into `master`, since cherry-picking them now would cause a lot of conflicts.

mrstanb:
The CI job on MacOS is failing again, this time due to the `04/58` regr. test case. I'm strongly assuming that it's due to the issue, documented in #1151

michael-schwarz (approved):

michael-schwarz:
> The CI job on MacOS is failing again, this time due to the 04/58 regr. test case. 

Probably fixed as of 7f9ec9a4ef10ee1221b62cd7b99d867b668a0204. I'll merge master into this again and if the CI passes then, I'll merge the PR.

mrstanb:
> > The CI job on MacOS is failing again, this time due to the 04/58 regr. test case.
> 
> Probably fixed as of [7f9ec9a](https://github.com/goblint/analyzer/commit/7f9ec9a4ef10ee1221b62cd7b99d867b668a0204). I'll merge master into this again and if the CI passes then, I'll merge the PR.

Awesome, thanks!

src/domains/queries.ml
```diff
@@ -335,7 +337,7 @@ struct
       | Any (EvalLength e1), Any (EvalLength e2) -> CilType.Exp.compare e1 e2
       | Any (EvalMutexAttr e1), Any (EvalMutexAttr e2) -> CilType.Exp.compare e1 e2
       | Any (EvalValue e1), Any (EvalValue e2) -> CilType.Exp.compare e1 e2
-      | Any (BlobSize e1), Any (BlobSize e2) -> CilType.Exp.compare e1 e2
+      | Any (BlobSize {exp = e1; _}), Any (BlobSize {exp = e2; _}) -> CilType.Exp.compare e1 e2
```
sim642, line 340: This also needs to compare `base_address` fields!

mrstanb: I'll open a small separate PR for this. Thanks for the note!

mrstanb: Should be fixed in #1196

PR #1095, opened 2023-06-28
Add PPX derivers for lattices of tuple and record types

sim642:
This implements a big portion of #31 using my experimental [ppx_easy_deriving](https://github.com/sim642/ppx_easy_deriving) library (**EDIT:** Necessary parts are now just vendored here for simplicity). The goal of the library is to provide a simpler [ppx_type_directed_value](https://github.com/janestreet/ppx_type_directed_value/)-like interface with no performance penalty.
It is a revival of my two-year-old [`ppx-lattice`](https://github.com/goblint/analyzer/tree/ppx-lattice) branch in more polished state.

This PR changes record-valued domains to use `[@@deriving lattice]` and also changes the implementation of `Lattice.Prod` to do so. Over time, the amount of `Lattice.Prod` usage could be reduced by using records with meaningful field names instead.

### TODO
- [ ] ~~Publish ppx_easy_deriving on opam.~~ Vendored.
- [ ] ~~Fix opam pin in CI.~~ Vendored.
- [ ] ~~Derive `pretty_diff`.~~ Needs `leq` to define `pretty_diff`.
- [x] Check ppxlib best practices.
- [x] Benchmark.
- [x] Investigate outliers.


src/cdomains/baseDomain.ml
```diff
@@ -97,14 +97,6 @@ struct
     let tr = QCheck.quad (CPA.arbitrary ()) (PartDeps.arbitrary ()) (WeakUpdates.arbitrary ()) (PrivD.arbitrary ()) in
     QCheck.map ~rev:to_tuple of_tuple tr
 
-  let bot () = { cpa = CPA.bot (); deps = PartDeps.bot (); weak = WeakUpdates.bot (); priv = PrivD.bot ()}
-  let is_bot {cpa; deps; weak; priv} = CPA.is_bot cpa && PartDeps.is_bot deps && WeakUpdates.is_bot weak && PrivD.is_bot priv
-  let top () = {cpa = CPA.top (); deps = PartDeps.top ();  weak = WeakUpdates.top () ; priv = PrivD.bot ()}
```
michael-schwarz, line 102: I think the generated version will be different w.r.t. to `priv` here, so we might have to leave this one!

src/cdomains/apron/relationDomain.apron.ml
```diff
@@ -191,26 +189,11 @@ struct
     let tr = QCheck.pair (RD.arbitrary ()) (PrivD.arbitrary ()) in
     QCheck.map ~rev:to_tuple of_tuple tr
 
-  let bot () = {rel = RD.bot (); priv = PrivD.bot ()}
-  let is_bot {rel; priv} = RD.is_bot rel && PrivD.is_bot priv
-  let top () = {rel = RD.top (); priv = PrivD.bot ()}
```
michael-schwarz, line 196: We have some top/bot confusion here and would need to do this manually here.

src/cdomains/floatDomain.ml
```diff
@@ -986,10 +986,6 @@ module FloatDomTupleImpl = struct
     Option.map_default identity ""
       (mapp { fp= (fun (type a) (module F : FloatDomain with type t = a) x -> F.name () ^ ":" ^ F.show x); } x)
 
-  let hash x =
-    Option.map_default identity 0
-      (mapp { fp= (fun (type a) (module F : FloatDomain with type t = a) -> F.hash); } x)
-
```
michael-schwarz, line 992: Is the derived one really equivalent here?

src/cdomains/baseDomain.ml
sim642: I'm pretty sure that's a copy-paste typo because on the following line, it's all tops:
```ocaml
let is_top {cpa; deps; weak; priv} = CPA.is_top cpa && PartDeps.is_top deps && WeakUpdates.is_top weak && PrivD.is_top priv
```
Or do you know of a reason why it should be that way? It doesn't make sense to me that it's `PrivD.bot` in both `bot` and `top` because that could only be if it's a unit lattice.

So I think this is the precise reason to have a PPX deriver to avoid such typos.

src/cdomains/floatDomain.ml
sim642: It might have different integer values but that's beside the point. `equal` and `compare` are already being derived, so manually defining `hash` is just error-prone if things ever go out of sync.

src/cdomains/baseDomain.ml
michael-schwarz: I think the idea is that even in top state, there's nothing privatized in the hopes that some additional precision can be regained later. Do the regression tests still work? And if yes, how about the tests we ran for traces and more-traces? Does this change the results there?

sim642: The hand-written version was fixed in #1463.

jerhard:
At the Gobcon today you mentioned that there could be another attempt at this by you on this, that would be easier to maintain. Could you have a look, whether this indeed exists?

src/cdomains/apron/relationDomain.apron.ml
sim642: #1463 fixed a similar thing in base. This is probably the same typo copied.

michael-schwarz (approved):

sim642:
I did a sv-benchmarks run with 60s timeout and here's the CPU time before vs after plot:
image
On average there's no change to performance but there are some outliers (in both directions!)

All verdict changes except one are close to timeout jitter. One `true` changes into a quick `OUT OF MEMORY` which is odd.

I might try looking into some of the biggest outliers before clicking merge.

sim642:
> All verdict changes except one are close to timeout jitter. One `true` changes into a quick `OUT OF MEMORY` which is odd.

This task is `c/ldv-linux-3.4-simple/43_1a_cilled_ok_nondet_linux-43_1a-drivers--input--joystick--turbografx.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.i`. Running with `-v` to see OCaml allocation stats actually shows the memory usage being lower now. This might've just been that slight changes cause GC to trigger a bit later now and there's a moment with higher memory usage.

sim642:
> On average there's no change to performance but there are some outliers (in both directions!)

One of the biggest improvements (2× in time) is on `/c/hardware-verification-bv/btor2c-lazyMod.driving_phils.5.prop1-func-interl.c` where also the allocated memory amount is halved. This is actually orthogonal to the new deriver: it's just because in this PR I've also replaced the `IntDomain` `hash` function with a derived one, which seems to have this huge effect. Maybe the hand-written one has too many collisions or temporary stuff.

In others, e.g. `c/goblint-regression/28-race_reach_71-funloop_racing.i` there's a slowdown 29s → 36s, which seems to be due to the same `hash` change. So there seems to be surprisingly large variation in runtime from a single `hash` function. This might need a separate investigation.

PR #1096, opened 2023-06-29
Changes for Pull Request bench#85: Automatic Generation of Test Cases for Incremental Static Analysis

J2000A:
Changes needed for the "**Test Automation for Incremental Analysis - TAIA**" from the [pull request in the bench repository](https://github.com/goblint/bench/pull/58):

- Add new goblint check annotation `NOFAIL` which expresses the knowledge the incremental analysis should have: After an incremental analysis the check should either be successful or unknown, but not fail. The `TODO` annotation is similar but not sufficient, as the `update_suite.rb` script ignores these annotations.

- Add new goblint check annotation `NOTINPRECISE` which checks that the incremental analysis did not get less precise. After an incremental analysis the check should either be successful or failing, but not unknown. Failing tests are ignored to only warn in case if inprecision, as we intend to check only for the precision and not the correctness.

- Add a reference to the "Test Automation for Incremental Analysis - TAIA" in the documentation.

- Improve the regex in the `update_suite.rb` script for matching annotations. The annotations are sometimes part of code generated by cil (e.g. `SOME_CIL_VARIABLE_FAIL`). To prevent false matchings the annotation should start with a word boundary `\b` or a `/` for the case that the annotation was written directly after a comment (e.g. `//SUCCESS`).

- When there is an exception in the `update_suite.rb` print the content of the `warn-file` for easier debugging.

- In the  `update_suite.rb` script use the `// PARAM:` string at the first line also for the incremental tests.

- Add a bool option `trans.goblint-check` for writing out `__goblint_check()`s instead of `__VERIFIER_assert`s.

- Change the depreciated parameter `--sets` in the regression test `46/26` to `--set`

- Change `MAYFAIL` to `MAY FAIL` for regression test 36/86, 58/15 and 67/16. Otherwise the `FAIL` would not be recognized by the tester due to the changed regex matching of the annoations.


michael-schwarz:
Do we intend to merge this or should we close it?

J2000A:
As far as I know @stilscher meant that she wants to merge it after reviewing.

src/transform/evalAssert.ml
```diff
@@ -26,8 +26,11 @@ module EvalAssert = struct
   (* should asserts be surrounded by __VERIFIER_atomic_{begin,end}? *)
   let surroundByAtomic = true
 
+  (* variable for generating __goblint_check(exp) instead of __VERIFIER_assert(exp) *)
+  let goblintCheck () = GobConfig.get_bool "trans.goblint-check"
```
michael-schwarz, line 30: This option should be renamed.

tests/regression/36-apron/86-branched-thread-creation.c
```diff
@@ -40,7 +40,7 @@ int main(void) {
 
   if(!mt) {
     pthread_mutex_lock(&mutex);
-    __goblint_check(g==h); //MAYFAIL
+    __goblint_check(g==h); //MAY FAIL
```
stilscher, line 43: Even though, this was not the content of this PR, this annotation should actually be a `FAIL` annotation since the check should always fail on all paths.

stilscher:
Since the number of commits is very large, but only few lines of code have been changed, I would suggest to squash merge this PR after the review is finished.

michael-schwarz (approved):

scripts/update_suite.rb
```diff
@@ -300,33 +302,43 @@ def parse_tests (lines)
...
         tests[i] = "unknown"
-      elsif obj =~ /(assert|__goblint_check).*\(/ then
-        if obj =~ /FAIL/ then
-          tests[i] = "fail"
-        elsif obj =~ /UNKNOWN/ then
-          tests[i] = "unknown"
-        else
-          tests[i] = "assert"
+      elsif obj =~ /(\b|\/)(assert|__goblint_check).*\(/ then
+        unless obj =~ /^\s*extern\b/
```
sim642, line 334: What's this `extern` about?

docs/developer-guide/testing.md
```diff
@@ -53,6 +53,7 @@ Comments at the end of other lines indicate the behavior on that line:
 | `DEADLOCK` | Deadlock warning | Deadlock is possible | Soundness |
 | `NOWARN` | No warning | — | Precision |
 | `WARN` | Some warning | — | Soundness |
+| `NOFAIL` | Assertion is unknown <br> or succeeds | Everything except fail | Incremental analysis |
```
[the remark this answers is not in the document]
stilscher: It may succeed or be unknown. It is what one expects for an incremental run if the assert was proven in a from-scratch run.

scripts/update_suite.rb
stilscher: It is about an extern definition of 'goblint_check', which would otherwise also be matched.

stilscher (approved):

docs/developer-guide/testing.md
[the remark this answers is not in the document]
sim642: That's in our abstract semantics, but in the concrete the assertion either fails or it succeeds. So it should be "Assertion may both succeed or fail"?

jerhard:
@sim642 In the GobCon today, you mentioned that you want to have a look at this PR and potentially cherry-pick some changes, before closing this. (As we are currently not working on the incremental).

PR #1099, opened 2023-07-08
Warn for {d,r}eallocation of non-dynamically allocated memory

mrstanb:
This PR adds checks in the `base.ml` analysis for calls to the special functions `realloc` and `free`:
* Whenever the pointer argument to either of these functions does not point to dynamically allocated memory, we need to WARN appropriately
* Cf. https://wiki.sei.cmu.edu/confluence/display/c/MEM34-C.+Only+free+memory+allocated+dynamically
* Cf. [CWE 590](https://cwe.mitre.org/data/definitions/590.html)

Also added a few regression test cases that should cover these newly introduced checks.
`make test` runs through successfully (including the new regression tests)

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1997,6 +1997,20 @@ struct
     let st' = invalidate ~deep:false ~ctx (Analyses.ask_of_ctx ctx) gs st shallow_addrs in
     invalidate ~deep:true ~ctx (Analyses.ask_of_ctx ctx) gs st' deep_addrs
 
+  let check_free_of_non_heap_mem ctx special_fn ptr =
+    match ctx.ask (Queries.MayPointTo ptr) with
+    | a when not (Queries.LS.is_top a) ->
+      let warn_if_not_heap_var special_fn var =
+        if not (ctx.ask (Queries.IsHeapVar var)) then
+          M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Free of non-dynamically allocated memory in function %s for pointer %a" special_fn.vname d_exp ptr
+      in
+      let pointed_to_vars =
+        Queries.LS.elements a
+        |> List.map fst
+      in
+      List.iter (warn_if_not_heap_var special_fn) pointed_to_vars
+    | _ -> ()
+
   let special ctx (lv:lval option) (f: varinfo) (args: exp list) =
     let invalidate_ret_lv st = match lv with
       | Some lv ->
@@ -2277,6 +2291,8 @@ struct
         | _ -> st
       end
     | Realloc { ptr = p; size }, _ ->
+      (* Realloc shouldn't be passed non-dynamically allocated memory *)
+      check_free_of_non_heap_mem ctx f p;
       begin match lv with
         | Some lv ->
           let ask = Analyses.ask_of_ctx ctx in
@@ -2308,6 +2324,10 @@ struct
         | None ->
           st
       end
+    | Free ptr, _ ->
+      (* Free shouldn't be passed non-dynamically allocated memory *)
+      check_free_of_non_heap_mem ctx f ptr;
+      st
     | Assert { exp; refine; _ }, _ -> assert_fn ctx exp refine
     | Setjmp { env }, _ ->
       let ask = Analyses.ask_of_ctx ctx in
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2003,7 +2003,7 @@ struct
       let points_to_set = addrToLvalSet a in
       if Q.LS.is_top points_to_set then
         M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Points-to set for pointer %a in function %s is top. Potential free of non-dynamically allocated memory may occur" d_exp ptr special_fn.vname
-      else if Q.LS.exists (fun (v, _) -> not (ctx.ask (Q.IsHeapVar v))) points_to_set then
+      else if (Q.LS.exists (fun (v, _) -> not (ctx.ask (Q.IsHeapVar v))) points_to_set) || (AD.mem Addr.UnknownPtr a) then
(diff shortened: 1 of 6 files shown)


src/analyses/base.ml
```diff
@@ -1997,6 +1997,20 @@ struct
...
+      let warn_if_not_heap_var special_fn var =
+        if not (ctx.ask (Queries.IsHeapVar var)) then
+          M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Free of non-dynamically allocated memory in function %s for pointer %a" special_fn.vname d_exp ptr
+      in
+      let pointed_to_vars =
+        Queries.LS.elements a
+        |> List.map fst
+      in
+      List.iter (warn_if_not_heap_var special_fn) pointed_to_vars
+    | _ -> ()
```
michael-schwarz, line 2012: When `a`  is top, we should probably also warn!

src/analyses/base.ml
```diff
@@ -1997,6 +1997,20 @@ struct
     let st' = invalidate ~deep:false ~ctx (Analyses.ask_of_ctx ctx) gs st shallow_addrs in
     invalidate ~deep:true ~ctx (Analyses.ask_of_ctx ctx) gs st' deep_addrs
 
+  let check_free_of_non_heap_mem ctx special_fn ptr =
+    match ctx.ask (Queries.MayPointTo ptr) with
+    | a when not (Queries.LS.is_top a) ->
+      let warn_if_not_heap_var special_fn var =
+        if not (ctx.ask (Queries.IsHeapVar var)) then
+          M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Free of non-dynamically allocated memory in function %s for pointer %a" special_fn.vname d_exp ptr
+      in
```
michael-schwarz, line 2006: Does it suffice to emit one warning if such an element exists instead of calling `warn` for every var? For large points-to sets, this may be a bit more efficient.

michael-schwarz (approved):
Thank you for the PR, this is something we should have added a long time ago. As soon as the comments are addressed, I think this is good to merge!

src/analyses/base.ml
mrstanb: Makes sense, yes! Thanks for pointing out! Should be covered in https://github.com/goblint/analyzer/pull/1099/commits/acb1ca83d00ae5f09167e1b30b775a465736ecd1

src/analyses/base.ml
mrstanb: Good point. I changed this to now warn once for the entire points-to set in case there exists a var that is not a heap var (cf. https://github.com/goblint/analyzer/pull/1099/commits/acb1ca83d00ae5f09167e1b30b775a465736ecd1)

src/analyses/base.ml
```diff
@@ -1998,18 +1998,14 @@ struct
...
-      in
-      let pointed_to_vars =
-        Queries.LS.elements a
-        |> List.map fst
-      in
-      List.iter (warn_if_not_heap_var special_fn) pointed_to_vars
-    | _ -> ()
+    let points_to_set = ctx.ask (Queries.MayPointTo ptr) in
+    let exists_non_heap_var =
+      Queries.LS.elements points_to_set
```
michael-schwarz, line 2003: This will throw an exception for top, iirc. So one probably has to check for top separately.

mrstanb: Yes, I completely forgot we might get `Unsupported` here :)
I added a `try ... with` now. (cf. https://github.com/goblint/analyzer/pull/1099/commits/ee80fd1761a7a780e46c2e9af25389bd268af65e)
Not sure if the `WARN` in the `with` of the `try` makes sense (when the points-to set is top), but decided to keep it there as a precaution, just in case.

mrstanb: With the `try ... with`, `make test` now runs through again locally

michael-schwarz:
@sim642 Do you still want to review, or are we good to merge?

src/analyses/base.ml
```diff
@@ -1997,6 +1997,20 @@ struct
...
 
+  let check_free_of_non_heap_mem ctx special_fn ptr =
+    let points_to_set = ctx.ask (Queries.MayPointTo ptr) in
+    begin try
+        let exists_non_heap_var =
+          (* elements throws Unsupported if the points-to set is top *)
+          Queries.LS.elements points_to_set
+          |> List.map fst
+          |> List.exists (fun var -> not (ctx.ask (Queries.IsHeapVar var)))
+        in
```
sim642, line 2008: This is in base, so a query is kind of roundabout. Other parts of base just use `eval_rv_address` or something to look up points-to sets and avoid the unnecessary back-and-forth conversions with `Queries.LS`.

src/analyses/base.ml
```diff
@@ -1997,6 +1997,20 @@ struct
...
+        if exists_non_heap_var then
+          M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Free of non-dynamically allocated memory in function %s for pointer %a" special_fn.vname d_exp ptr
+      with _ -> M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Points-to set for pointer %a in function %s is top. Potential free of non-dynamically allocated memory may occur" d_exp ptr special_fn.vname
```
sim642, line 2011: Catch-all exception handlers shouldn't be used. If `eval_rv_address` is used above, then the top pointer case can be handled without exceptions like elsewhere in base.

sim642 (changes requested):

src/analyses/base.ml
mrstanb: Should also be fixed in https://github.com/goblint/analyzer/pull/1099/commits/99fb013038d3b6c333f236537668e3513d8faceb

mrstanb:
Regression test case 75/02 was failing after the changes in https://github.com/goblint/analyzer/pull/1099/commits/99fb013038d3b6c333f236537668e3513d8faceb.

I added a check for `Addr.Unknown` in https://github.com/goblint/analyzer/pull/1099/commits/e23c42d570ee2539a0f04c10ebd175254eab5065 similarly to how it's done in the `MayPointTo` query answer in `base.ml`.

All the tests are passing again now

sim642 (approved):

PR #1101, opened 2023-07-08
Consider thread-local variables non-unique

### LINKED ISSUE #1072: Address comparisons unsound for thread-local variables
Consider the following program:

~~~C
#include <pthread.h>
#include <stdio.h>
#include <assert.h>

__thread int myglobal;

void *t_fun(void *arg) {
  int* ptr = (int*) arg;
  __goblint_check(&myglobal == ptr); //FAIL
  return NULL;
}

int main(void) {
  pthread_t id;
  pthread_create(&id, NULL, t_fun, (void*) &myglobal);
  pthread_join (id, NULL);
  return 0;
}
~~~

This comparison should fail (or be unknown), but Goblint claims to be able to prove it.

michael-schwarz:
Fixes #1072 and the unsoundness described there.


jerhard:
Does marking thread-locals variables as non-unique also lead to them being only weakly updated? If so, could one treat thread-local variables as non-unique only when they are escaped, to remedy this issue? 

It should not be possible to have address-comparisons between multiple instances of a particular type of thread-local variable without one of them having escaped.

michael-schwarz:
> Does marking thread-locals variables as non-unique also lead to them being only weakly updated? If so, could one treat thread-local variables as non-unique only when they are escaped, to remedy this issue?

The weak update mechanism applies mainly to locals, here the variables are still treated as globals (pending a more systematic approach). However, I added a test whether this thread-local variable has its address taken, which was also the proxy we used for having escaped in #1071. I tried for a bit, but could not really construct any examples where that really makes a difference. I still think it is good to have it in the interest of consistency though.

sim642 (approved):

tests/regression/04-mutex/84-thread-local-eq.c
sim642, line 1: This test needs to be renumbered for master due to #1089.

PR #1107, opened 2023-07-11
Add special function for `rand()` to improve precision

### LINKED ISSUE #1088: Add `rand` special function
As pointed out in https://github.com/goblint/GobPie/pull/60, Goblint's current library function handling for `rand` is not precise. By returning a completely unknown value, it also appears to return negative values, which `rand` does not, according to the C standard.

Therefore, adding a special function type for `rand` would allow it (and possibly some other `rand` variations) more precisely.

michael-schwarz:
Negative values can be excluded. Closes #1088

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2345,6 +2345,13 @@ struct
       let rv = ensure_not_zero @@ eval_rv ask ctx.global ctx.local value in
       let t = Cilfacade.typeOf value in
       set ~ctx ~t_override:t ask ctx.global ctx.local (AD.of_var !longjmp_return) t rv (* Not raising Deadcode here, deadcode is raised at a higher level! *)
+    | Rand, _ ->
+      begin match lv with
+      | Some x -> 
+        let result:value = (Int (ID.starting IInt Z.zero)) in
+        set ~ctx (Analyses.ask_of_ctx ctx) gs st (eval_lv (Analyses.ask_of_ctx ctx) ctx.global st x) (Cilfacade.typeOfLval x) result
+      | None -> st
+      end
     | _, _ ->
       let st =
         special_unknown_invalidate ctx (Analyses.ask_of_ctx ctx) gs st f args
--- a/tests/regression/01-cpa/74-rand.c
+++ b/tests/regression/01-cpa/74-rand.c
@@ -1,0 +1,13 @@
+//PARAM: --enable ana.int.interval
+#include <stdio.h>
+#include <stdlib.h>
+#include <goblint.h>
+
+int main () {
+  int r = rand();
+
+  __goblint_check(r >= 0);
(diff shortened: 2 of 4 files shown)


tests/regression/01-cpa/74-rand.c
```diff
@@ -0,0 +1,13 @@
+//PARAM: --enable ana.int.interval
+#include <stdio.h>
+#include <stdlib.h>
+#include <goblint.h>
+
+int main () {
+  int r = rand();
+
+  __goblint_check(r >= 0);
+  __goblint_check(r <= RAND_MAX);
```
sim642, line 10: `r` has the abstract value `ID.starting IInt Z.zero` which isn't sufficient to prove that it's at most `RAND_MAX`. This accidentally passes only when `RAND_MAX` happens to be `INT_MAX` but that's not required. In all other cases this test will spuriously fail.

src/analyses/base.ml
```diff
@@ -2345,6 +2345,13 @@ struct
       let rv = ensure_not_zero @@ eval_rv ask ctx.global ctx.local value in
       let t = Cilfacade.typeOf value in
       set ~ctx ~t_override:t ask ctx.global ctx.local (AD.of_var !longjmp_return) t rv (* Not raising Deadcode here, deadcode is raised at a higher level! *)
+    | Rand, _ ->
+      begin match lv with
+      | Some x -> 
+        let result:value = (Int (ID.starting IInt Z.zero)) in
+        set ~ctx (Analyses.ask_of_ctx ctx) gs st (eval_lv (Analyses.ask_of_ctx ctx) ctx.global st x) (Cilfacade.typeOfLval x) result
+      | None -> st
+      end
```
sim642, line 2354: While we're at it, maybe relational analysis should also handle this.

tests/regression/01-cpa/74-rand.c
michael-schwarz: I think `RAND_MAX` is `INT_MAX` on all practical systems though.

sim642: We shouldn't be making such an incorrect assumption though, especially since it's explicitly implementation-defined by the standard. Even if it's true in practice, then this check isn't really checking anything that Goblint doesn't trivially already know by default. So it's also misleading as it appears like we handle `RAND_MAX`.

The safest thing would be to just remove this check because it adds nothing and just makes us prone to opam CI failures on some less common architectures, machines and distributions. Otherwise we should handle `RAND_MAX` properly, but that's a pain (having to do something machdep-like to preprocess a tiny program with `RAND_MAX` to find out what it expands to).

michael-schwarz: I'll remove this test.

sim642 (approved):

PR #1109, opened 2023-07-11
`UnionDomain.meet`: Throw `Uncomparable` exception for abstract unions with different active fields

### LINKED ISSUE #1105: Use of `UnionDomain.meet` in refinement causes unsoundness and crashes
The refinement that is done when encountering branching uses `UnionDomain.meet`, which currently causes unsound behavior and crashes. 

In the following example, Goblint considers both branches of the `if` to be dead:

https://github.com/goblint/analyzer/blob/b45d527d1e4d6f66fe0e0e035376bfbd8c45d0cc/tests/regression/27-inv_invariants/18-union-float-int.c#L9-L18

The reason is that the meet of a: $(field: x, value: 12)$ and b: $(field: f, value: 129.0)$ yields $(field: \bot, value: \bot)$.

In an example where the union type contains either a `char` or an `int`, Goblint crashes with a `IncompatibleIkind`-exception, because the abstract values with `ikinds`  of `Char` and `Int` are passed to a `meet`.
https://github.com/goblint/analyzer/blob/b45d527d1e4d6f66fe0e0e035376bfbd8c45d0cc/tests/regression/27-inv_invariants/19-union-char-int.c#L3-L18

jerhard:
**Edit:** This PR changes the `meet` of the `UnionDomain` such that it throws a `Uncomparable` exception, as suggested below: https://github.com/goblint/analyzer/pull/1109#discussion_r1260266161. The `BaseInvariant` that uses the `VD.meet` catches the `Uncomparable` exception and does not do a refinement in case it is thrown.

On test case `27/18` both branches of an `if` comparing two unions are no longer considered dead, and test `27/19` does no longer crashes with the `IncompatbileIkindException`. 
   
Fixes #1105.

<details> 
<summary>Previous Approach</summary>

~~This adds a function `ValueDomain.try_meet`, a version of the `meet` that may fail for unions with different active fields and use it in BaseInvariant.  @michael-schwarz and I discussed this today.~~
    
~~The refinement by BaseInvariant to handle branching used VD.meet, which caused issue #1105, due to the behavior of UnionDomain.meet: The meet of unions with different active fields can result in a bottom value.
This led to e.g. spuriously dead code.~~
    
 ~~This adds a function `try_meet` to the `VvalueDomain`, the union domain, and those domains that may contain a union domain element: structs, arrays, blobs. The function `try_meet` in the union domain throws an exception if the fields of the unions for which a meet was tried do not coincide.~~
 </details>

### DIFF
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -63,7 +63,7 @@ struct
     (*   Address (AD.join o n) *)
     (* | Address o, Address n when AD.mem (Addr.unknown_ptr ()) o -> Address n *)
     (* | Address o, Address n when AD.mem (Addr.unknown_ptr ()) n -> Address o *)
-    | _ -> VD.try_meet oldv newv
+    | _ -> VD.meet oldv newv
 
   let refine_lv_fallback ctx a gs st lval value tv =
     if M.tracing then M.tracec "invariant" "Restricting %a with %a\n" d_lval lval VD.pretty value;
@@ -83,11 +83,7 @@ struct
       in
       let state_with_excluded = set a gs st addr t_lval value ~ctx in
       let value =  get a gs state_with_excluded addr None in
-      let new_val =
-        try
-          apply_invariant oldval value
-        with PreValueDomain.NotMeetable -> oldval
-      in
+      let new_val = apply_invariant oldval value in
       if M.tracing then M.traceu "invariant" "New value is %a\n" VD.pretty new_val;
       (* make that address meet the invariant, i.e exclusion sets will be joined *)
       if is_some_bot new_val then (
@@ -107,11 +103,7 @@ struct
       let oldv = map_oldval oldv var.vtype in
       let offs = convert_offset a gs st o in
       let newv = VD.update_offset (Queries.to_value_domain_ask a) oldv offs c' (Some exp) x (var.vtype) in
-      let v =
-        try
-          VD.try_meet oldv newv
-        with PreValueDomain.NotMeetable -> oldv
-      in
+      let v = VD.meet oldv newv in
       if is_some_bot v then contra st
       else (
         if M.tracing then M.tracel "inv" "improve variable %a from %a to %a (c = %a, c' = %a)\n" CilType.Varinfo.pretty var VD.pretty oldv VD.pretty v pretty c VD.pretty c';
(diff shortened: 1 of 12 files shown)


jerhard:
The exact behavior on `27/19` may be investigated further in the future, as currently, Goblint claims that for the following `if` always evaluates to `true`:

https://github.com/goblint/analyzer/blob/14929ca3bf9219edd9c19d15a29bf0448784b211/tests/regression/27-inv_invariants/19-union-char-int.c#L3-L18

Whether this actually holds may e.g. depend on the endianness of the machine. But this is a separate issue.

src/analyses/baseInvariant.ml
```diff
@@ -83,7 +83,11 @@ struct
       in
       let state_with_excluded = set a gs st addr t_lval value ~ctx in
       let value =  get a gs state_with_excluded addr None in
-      let new_val = apply_invariant oldval value in
+      let new_val =
+        try
+          apply_invariant oldval value
+        with PreValueDomain.NotMeetable -> oldval
```
sim642, line 89: Seems to me like `Lattice.Uncomparable` already exists for this precise meaning:
https://github.com/goblint/analyzer/blob/3af00a86da65ab5caac033b0ecff02dc69c3bfbf/src/domains/lattice.ml#L14-L15
Just using that exception would avoid the need for littering interfaces with some special `try_meet` because `meet` is allowed to raise `Uncomparable`.

Then `try_meet` could just be a helper function here which has such fallback. Right now that code is repeated multiple times.

jerhard: Yes, that is much more concise indeed. I adapted the implementation accordingly.

src/cdomains/unionDomain.ml
```diff
@@ -15,16 +15,29 @@ sig
...
 module Simple (Values: Arg) =
 struct
   include Lattice.Prod (Field) (Values)
   type value = Values.t
 
+  let meet (f, x) (g, y) =
+    let field = Field.meet f g in
+    let value = Values.meet x y in
+    (field, value)
+
```
sim642, line 40: Isn't this the default `Lattice.Prod.meet` anyway?

jerhard: This is right, I removed the explicit definition.

sim642 (approved):

jerhard:
I had to readd the explicit `meet` on `master` since otherwise it would not be ensured that the `meet` is first performed on the `Field`, leading to `IncompatibleIkind` exceptions, where `Lattice.Uncomparable` exceptions should be thrown.

PR #1112, opened 2023-07-12
Remove most `MapDomain.Groupable`

### LINKED ISSUE #1097: Remove `MapDomain.Groupable`
`MapDomain` keys require more than just a `Printable.S`, which is constantly annoying:
https://github.com/goblint/analyzer/blob/a651cccf69af65581222d7e8e046b8ad67510cf4/src/domains/mapDomain.ml#L58-L65
AFAIK only base analysis' `CPA` uses non-trivial grouping and just in its debug `pretty` printing (not even in XML), so the grouped printing code could just be specialized in `CPA` while all other uses of `MapDomain` avoid the extra complication.

sim642:
Closes #1097. The interface is kept such that base's `CPA` could still explicitly use grouped pretty-printing.

### DIFF
--- a/src/cdomains/baseDomain.ml
+++ b/src/cdomains/baseDomain.ml
@@ -15,13 +15,6 @@ struct
   include M
 end
 
-
-module Glob =
-struct
-  module Var = Basetype.Variables
-  module Val = VD
-end
-
 (* Keeps track of which arrays are potentially partitioned according to an expression containing a specific variable *)
 (* Map from variables to sets of arrays: var -> {array} *)
 module PartDeps =
--- a/src/cdomains/baseDomain.ml
+++ b/src/cdomains/baseDomain.ml
@@ -6,13 +6,14 @@ module BI = IntOps.BigIntOps
 
 module CPA =
 struct
+  module M0 = MapDomain.MapBot (Basetype.Variables) (VD)
   module M =
   struct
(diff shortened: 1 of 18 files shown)


src/cdomains/baseDomain.ml
```diff
@@ -6,20 +6,14 @@ module BI = IntOps.BigIntOps
 
 module CPA =
 struct
+  module M0 = MapDomain.MapBot (Basetype.Variables) (VD)
```
michael-schwarz, line 9: Why did you remove `HashCashed` here?

michael-schwarz (approved):
Once my comment about HashCached is addressed, this can be merged.

src/cdomains/baseDomain.ml
sim642: It's still there below on line 15. This is just the innermost `MapDomain` whose printing is changed to preserve the old grouping. After that it's wrapped the same way.

michael-schwarz: But the wrapping in line 15 only applies to `Glob`, not the `CPA` itself. Or am I missing something?

michael-schwarz: ah, nvm the diff is just a bit confusing. sorry about that

sim642: There's no `Glob`, it was just some old unused module that I removed.

michael-schwarz: Yeah, sorry. It's good to merge then.

PR #1114, opened 2023-07-15
Multi-threaded UAF Analysis

mrstanb:
This PR:
* Extends the already existing Use-After-Free (UAF) analysis by extending with proper checks and warnings for UAF bugs in multi-threaded programs
* Removes the previously added simple heuristic for UAF warning in multi-threaded programs.

For reference, see also the idea outlined here: https://github.com/goblint/analyzer/pull/1050#issuecomment-1601670330

The multi-threaded UAF regression test case now passes successfully after the changes from this PR.

### DIFF
--- a/src/analyses/useAfterFree.ml
+++ b/src/analyses/useAfterFree.ml
@@ -24,9 +24,51 @@ struct
 
   (* HELPER FUNCTIONS *)
 
+  let get_current_threadid ctx =
+    ctx.ask Queries.CurrentThreadId
+
   let warn_for_multi_threaded ctx behavior cwe_number =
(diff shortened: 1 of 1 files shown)


michael-schwarz:
> The multi-threaded UAF regression test case now passes successfully after the changes from this PR.

Should not this test receive new annotations then?

mrstanb:
> > The multi-threaded UAF regression test case now passes successfully after the changes from this PR.
> 
> Should not this test receive new annotations then?

The test case already had annotations prior, so nothing should be added or changed to it

sim642 (approved):

PR #1117, opened 2023-07-25
Replace direct `stdout` printing with logging

sim642:
Apparently this is something I started but never opened a PR for. The problem is that there's lots of direct printing to `stdout` throughout Goblint, which makes output uncontrollable and unpredictable. For example, server mode cannot use `stdout` because some random printing might get mixed in and break the whole JSON stream.

The idea is to add a `Logs` library, which is a bit like `Messages`. It has various logging levels that can be toggled globally (unlike direct printing). The difference is that `Messages` is for things related to analysis results and shown in IDE, while logging is about Goblint itself, e.g. those TD3 incremental changes lines.
`Messages` are unordered and incrementally preserved, while `Logs` just come out in the order they happen and aren't preserved like that.

### TODO
- [x] Fix conflicts with `master`. (latest: 2024-02-13)
- [x] Replace all new direct printing on `master` with `Logs`. (latest: 2024-02-13)
- [x] Add log level option.
- [x] `Maingoblint.check_arguments`.
- [x] Remove `dbg.verbose`.
- [x] What to do with GobView printing hacks?
- [x] Document `Logs` in developer guide.
- [x] https://github.com/goblint/gobview/pull/40.


sim642:
Does anyone still want to review this? We should get this merged as soon as possible to prevent it from becoming too stale.

src/analyses/extractPthread.ml
```diff
@@ -850,7 +850,7 @@ module Codegen = struct
 
     Writer.write "promela model" "pml" promela ;
     Writer.write "graph" "dot" dot_graph ;
-    print_endline
+    Logs.info
```
michael-schwarz, line 853: I think this a bit too important to just stick it into a `Logs.info`. It is not something to be logged but directly informs the user about what to do.

michael-schwarz:
This appears as if all logging output now goes into `stderr`??? This seems very strange, I would not expect logging information to be on `stderr` unless it is related to an error?!

src/analyses/threadId.ml
```diff
@@ -156,12 +156,13 @@ struct
     let non_uniques = List.filter_map (fun (a,b) -> if not (Thread.is_unique a) then Some a else None) tids in
     let uc = List.length uniques in
     let nc = List.length non_uniques in
-    Printf.printf "Encountered number of thread IDs (unique): %i (%i)\n" (uc+nc) uc;
-    Printf.printf "unique: ";
-    List.iter (fun tid -> Printf.printf " %s " (Thread.show tid)) uniques;
-    Printf.printf "\nnon-unique: ";
-    List.iter (fun tid -> Printf.printf " %s " (Thread.show tid)) non_uniques;
-    Printf.printf "\n"
+    Logs.debug "Encountered number of thread IDs (unique): %i (%i)" (uc+nc) uc;
```
michael-schwarz, line 159: These should be printed independently of the logging level, as it already is controlled by `dbg.print_tids`. Also, as this not some random debugging information, maybe these should be issued via the messaging system?

src/cdomains/valueDomain.ml
```diff
@@ -501,8 +501,7 @@ struct
 
 
   let warn_type op x y =
-    if GobConfig.get_bool "dbg.verbose" then
-      ignore @@ printf "warn_type %s: incomparable abstr. values %s and %s at %a: %a and %a\n" op (tag_name (x:t)) (tag_name (y:t)) CilType.Location.pretty !Tracing.current_loc pretty x pretty y
+    Logs.debug "warn_type %s: incomparable abstr. values %s and %s at %a: %a and %a" op (tag_name (x:t)) (tag_name (y:t)) CilType.Location.pretty !Tracing.current_loc pretty x pretty y
```
michael-schwarz, line 504: We issue thousands of these for large programs, maybe checking the log level before makes sense?

src/util/precCompare.ml
```diff
@@ -129,6 +129,7 @@ struct
     |> List.filter (fun ((i1, _), (i2, _)) -> i1 <> i2)
     |> List.map (Tuple2.map snd snd)
     |> List.map (uncurry compare_dumps)
-    |> List.iter (fun (_, msg) -> ignore (Pretty.printf "%t\n" (fun () -> msg)));
-    ignore (Pretty.printf "\nTotal locations: %d\nTotal %s: %d\n" locations_count (Key.name ()) location_vars_count)
+    |> List.iter (fun (_, msg) -> Logs.info "%t" (fun () -> msg));
+    Logs.newline ();
+    Logs.info "Total locations: %d\nTotal %s: %d" locations_count (Key.name ()) location_vars_count
```
michael-schwarz, line 134: This is the main output of the corresponding executable, why not just write to stout as usual?

src/witness/witness.ml
```diff
@@ -272,7 +272,7 @@ let write_file filename (module Task:Task) (module TaskResult:WitnessTaskResult)
 
 
 let print_svcomp_result (s: string): unit =
-  Printf.printf "SV-COMP result: %s\n" s
+  Logs.info "SV-COMP result: %s" s
```
michael-schwarz, line 275: This is a main result for the user. Migrate to messages?

michael-schwarz:
> Does anyone still want to review this? We should get this merged as soon as possible to prevent it from becoming too stale.

I reviewed this now. I think this requires considerable changes to all of our benchmarking scripts as well, is there a companion PR to fix all these?

src/cdomains/valueDomain.ml
michael-schwarz: Yes, it is more about the intermediate things being computed before it is determined that no print is necessary because verbose logging is not on.

src/framework/analyses.ml
```diff
@@ -306,8 +306,7 @@ struct
         let f = BatIO.output_channel out in
         write_file f (get_string "outfile")
     | "sarif" ->
-      let open BatPrintf in
-      printf "Writing Sarif to file: %s\n%!" (get_string "outfile");
+      Logs.info "Writing Sarif to file: %s" (get_string "outfile");
```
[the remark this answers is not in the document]
michael-schwarz: Yes. Still, the user will usually not have specified the file name himself (instead using the defaults), thus it is helpful to print where the find is to be found, independent of logging or not logging.

michael-schwarz:
> Also, it's standard for logging frameworks to go to stderr, e.g. in [Java](https://stackoverflow.com/q/34156052/854540) or Python.

I wasn't aware, thank you!


> If logging output went to `stdout`, it would still leave `stdout` as completely unusable by server mode.

`stdout` is user-facing standard output, I don't think we should limit ourselves to only printing cryptic json messages for the sever mode to `stdout` and instead print any things relevant to the normal user there. This certainly doesn't include most of the crap we write. Nevertheless, people will often redirect `stdout` to some log file and not redirect `stderr` (or redirect it to `/dev/null`) (especially if it contains tons of junk when running with high log level), so I think  it is wise to print things like fixpoint errors or warnings on options being magically (de-)activated also to `stdout`. (Maybe via the messages?)

src/witness/witness.ml
michael-schwarz: Also, does this require a rewrite of our tool info module?

michael-schwarz:
I'd suggest to discuss this PR at the next Gobcon.

sim642:
> Nevertheless, people will often redirect `stdout` to some log file and not redirect `stderr` (or redirect it to `/dev/null`) (especially if it contains tons of junk when running with high log level)

Again, that's the point of separating the two: choice! If everything is just directly printed to `stdout`, then it's impossible to separate user-facing warnings from all sorts of debug output. There's lots of things that still go to `stderr` though, so one almost always already has to redirect both together into a file, which isn't that hard: `&>` instead of `>`.


> so I think it is wise to print things like fixpoint errors or warnings on options being magically (de-)activated also to `stdout`. (Maybe via the messages?)

This is precisely what #1191 did to make these issues clearly visible, not just on the right output stream, but also in a visible place, not in the middle of thousands of warnings.

The other major benefit of logging to `stderr` is that it's unbuffered, so output appears as it's produced by Goblint. This matters for timing-sensitive information like logging, but not for user-facing warnings where buffering causes no confusion. With the current mixed output policy, Goblint output can be very confusing and misleading. For example, some CIL merging warnings are often printed much later than the actual merging takes place, between or after Goblint warnings.

This is also an issue when tracing (which correctly goes to `stderr`): non-tracing logging is buffered and appears out of sequence w.r.t. the tracing output, which again is misleading and leads to wasted debugging time. The only way to make things consistent is to duplicate the logging output also with tracing, such that it also appears at the correct place between tracing output.

sim642:
There's now a "result" log level that by default goes to `stdout` for the handful of places that were pointed out. In server mode, it is also directed to `stderr` to allow clean JSON-RPC communication.

src/witness/witness.ml
```diff
@@ -284,7 +284,7 @@ let init (module FileCfg: MyCFG.FileCfg) =
     let specification = Svcomp.Specification.of_option ()
   end
   in
-  Printf.printf "SV-COMP specification: %s\n" (Svcomp.Specification.to_string Task.specification);
+  Logs.info "SV-COMP specification: %s" (Svcomp.Specification.to_string Task.specification);
```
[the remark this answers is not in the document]
sim642: It's not a result, so it shouldn't abuse that level. It's just repeating the specification given by the user on the command line.

src/framework/constraints.ml
```diff
@@ -1713,22 +1713,22 @@ struct
         f_eq ()
       else if b1 then begin
         if get_bool "dbg.compare_runs.diff" then
-          ignore (Pretty.printf "Global %a is more precise using left:\n%a\n" Sys.GVar.pretty_trace k G.pretty_diff (v2,v1));
+          Logs.info "Global %a is more precise using left:\n%a" Sys.GVar.pretty_trace k G.pretty_diff (v2,v1);
```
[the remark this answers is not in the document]
sim642: This isn't the result of the analysis (which is printed separately anyway), this is just extra `dbg`-option-enabled comparison of the results. It's not giving results about the program, but rather for debugging Goblint itself.

src/analyses/threadId.ml
```diff
@@ -175,12 +175,9 @@ struct
...
     let nc = List.length non_uniques in
-    Printf.printf "Encountered number of thread IDs (unique): %i (%i)\n" (uc+nc) uc;
-    Printf.printf "unique: ";
-    List.iter (fun tid -> Printf.printf " %s " (Thread.show tid)) uniques;
-    Printf.printf "\nnon-unique: ";
-    List.iter (fun tid -> Printf.printf " %s " (Thread.show tid)) non_uniques;
-    Printf.printf "\n"
+    M.debug_noloc ~category:Analyzer "Encountered number of thread IDs (unique): %i (%i)" (uc+nc) uc;
+    M.msg_group Debug ~category:Analyzer "Unique TIDs" (List.map (fun tid -> (Thread.pretty () tid, None)) uniques);
+    M.msg_group Debug ~category:Analyzer "Non-unique TIDs" (List.map (fun tid -> (Thread.pretty () tid, None)) non_uniques)
```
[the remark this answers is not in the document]
michael-schwarz: If the check is there, I'm good with it!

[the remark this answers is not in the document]
sim642: There was one for `dbg.print_wpoints` but I also added one for `dbg.print_tids` now.

sim642:
Indeed, but I cannot check if GobView still works because of
- [ ] https://github.com/goblint/gobview/pull/25#issuecomment-1931542589
- [ ] https://github.com/goblint/cil/pull/165, https://github.com/goblint/analyzer/pull/1350, https://github.com/goblint/gobview/pull/39

sim642:
On top of GobView fixes, I confirmed that it still works.

PR #1118, opened 2023-07-25
Relational Substring Domain

nathanschmidt:
This PR implements a relational domain tracking substring relations between string variables. It adapts [Arceri et al.'s generic proposition](https://link.springer.com/chapter/10.1007/978-3-030-94583-1_2) to fit the special design of C-strings. The domain can be understood as an enhancement to the null byte array domain, taking advantage of the tracked relations to analyze `strlen`, `strcmp`, `strncmp`, and especially `strstr` with greater precision. Indeed, to ensure all functions behave as expected, we use the query system to gain necessary information on a string's allocated memory size and its length.

The PR also introduces a new analysis using the new domain, with cases for `strcpy`, `strncpy`, `strcat`, `strncat`, `strlen`, `strstr`, `strcmp`, and `strncmp`. 

**This PR depends on #1076 and hence should only be merged afterwards.**

*TODOs:*

- [x] Implement `special` in the substring analysis
- [ ] Answer new queries and restructure `special` in `base` 
- [ ] Add regression tests


src/analyses/apron/relationalStringAnalyses.apron.ml
```diff
@@ -0,0 +1,47 @@
...
+      (module Spec)
+    )
+
+  let get_spec (): (module MCPSpec) =
+    Lazy.force spec_module
+
+  let after_config () =
+    let module Spec = (val get_spec ()) in
+    MCP.register_analysis (module Spec : MCPSpec);
+    GobConfig.set_string "ana.path_sens[+]"  (Spec.name ())
```
michael-schwarz, line 30: Is this needed?

src/dune
```diff
@@ -47,6 +47,14 @@
...
       (-> relationPrecCompareUtil.no-apron.ml)
     )
+    (select relationalStringAnalyses.ml from
+      (apron -> relationalStringAnalyses.apron.ml)
+      (-> relationalStringAnalyses.no-apron.ml)
+    )
+    (select relationalStringDomains.ml from
+      (apron -> relationalStringDomains.apron.ml)
+      (-> relationalStringDomains.no-apron.ml)
+    )
```
sim642, line 57: The fact that a relational string analysis, which has nothing to do with Apron, needs to do these hacks, means that we really need to refactor our generic relational analysis first to decouple it from Apron.

michael-schwarz:
We'll place reviews on hold until the implementation has progressed further, please ping us once that is the case @nathanschmidt !

michael-schwarz:
We have now merged #1076, so this hopefully is no longer blocked!

nathanschmidt:
> We have now merged #1076, so this hopefully is no longer blocked!

Yes, thank you for finishing the other PR! I'll try my best to get around to it, but I currently have a lot of other work, so unfortunately no guarantees.

jerhard:
While this is a very interesting approach, it is not yet integrated into the other analyses. Thus, we close this for now.

PR #1123, opened 2023-07-31
Improve multi-threaded UAF analysis and add SVComp result generation

mrstanb:
This PR:
* Updates the multi-threaded UAF analysis by now globally tracking a set of tuples of the form `(tid, {must-joined threads})`, where `tid` denotes the tid of a thread which has freed the memory being tracked for UAF
* Adds SVComp result generation for UAF-related memory safety bugs


src/analyses/useAfterFree.ml
```diff
@@ -27,40 +28,61 @@ struct
   let get_current_threadid ctx =
     ctx.ask Queries.CurrentThreadId
 
+  let get_joined_threads ctx =
+    ctx.ask Queries.MustJoinedThreads
+
+  let warn_about_deactivated_thread_joins () =
+    if not @@ List.mem "threadJoins" @@ get_string_list "ana.activated" then
+      M.warn "Running without thread joins analysis. Multi-threaded UAF detection might be imprecise!"
```
sim642, line 36: This "warning" is currently causing some CI failures. I think we don't need this because that applies generally to many of our (multi-threaded) analyses.

src/witness/svcomp.ml
```diff
@@ -52,6 +52,7 @@ struct
         | UnreachCall _ -> "unreach-call"
         | NoOverflow -> "no-overflow"
         | NoDataRace -> "no-data-race" (* not yet in SV-COMP/Benchexec *)
+        | MemorySafety -> "memory-safety"
```
sim642, line 55: There is no `memory-safety` property in SV-COMP: https://sv-comp.sosy-lab.org/2023/rules.php. There's `valid-free`, `valid-deref`, `valid-memtrack` and `valid-memcleanup`.

I'm not sure if use-after-free directly maps to any one of them. I think it's part of `valid-deref` but that means we have to soundly find all other possible invalid dereferences as well, not just those after `free`. So we might not be able to support this property and compete in those SV-COMP categories still.

src/analyses/useAfterFree.ml
```diff
@@ -125,9 +149,25 @@ struct
...
+    let side_effect_globals_to_heap_var heap_var globals_to_add =
+      let (tid_to_add, joined_threads_to_add) = globals_to_add in
+      let current_globals = ctx.global heap_var in
+      (*
+        * Check if there are tuples with the same first component (i.e., tid)
+        * If no, side-effect the globals that we receive here and be done
+        * If yes, join all second components together and side-effect a single global with
+        the tid as first component and the joined second components as a single second component
+      *)
+      let globals_with_same_tid = G.filter (fun (tid, _) -> ThreadIdDomain.ThreadLifted.equal tid tid_to_add) current_globals in
```
sim642, line 162: This sounds like `G` should be a `MapDomain`, not a `SetDomain` of pairs where a complete iteration is done to check equality on first component. Or am I mistaken?

sim642 (changes requested):

src/witness/svcomp.ml
mrstanb: UAF should be part of the `valid-free` category (although it's technically also part of `valid-deref`).

The idea we discussed with @michael-schwarz here was to add SVComp result generation for the entire "Memory Safety" group of properties (i.e., `valid-free`, `valid-deref`, `valid-memtrack`), so that we could let the UAF analysis run against the SVComp benchmarks and see if the UAF analysis is sound and covering all cases. For the rest of the cases, we assumed it'll be normal to not be sound at the moment, as we're only attempting to test the UAF detection functionality.

mrstanb: If `memory-safety` doesn't make any sense as a property, I can change it to `valid-free` (resp. also add the cases for `valid-deref`, `valid-memtrack`).

What do you think?

mrstanb: Should we also add a `valid-memsafety.prp` file in `tests/sv-comp` and write some test code for this?

src/analyses/useAfterFree.ml
```diff
@@ -125,9 +142,18 @@ struct
...
-  let side_effect_mem_free ctx freed_heap_vars threadid =
-    let threadid = G.singleton threadid in
-    D.iter (fun var -> ctx.sideg var threadid) freed_heap_vars
+  let side_effect_mem_free ctx freed_heap_vars threadid joined_threads =
+    let side_effect_globals_to_heap_var heap_var =
+      let current_globals = ctx.global heap_var in
+      let joined_threads_to_add = match G.find_opt threadid current_globals with
+        | Some threads -> ConcDomain.ThreadSet.inter joined_threads threads
+        | None -> joined_threads
+      in
```
michael-schwarz, line 151: Any reason this accumulation cannot happen inside the side-effect by default?

tests/regression/74-use_after_free/13-multi-threaded-uaf-with-joined-thread.c
```diff
@@ -0,0 +1,33 @@
...
+#include <pthread.h>
+
+int* gptr;
+
+// Mutex to ensure we don't get race warnings, but the UAF warnings we actually care about
+pthread_mutex_t mtx = PTHREAD_MUTEX_INITIALIZER;
+
+void *t_use(void* p) {
+    pthread_mutex_lock(&mtx);
+    *gptr = 0; //NOWARN
```
michael-schwarz, line 13: I think the reason we still produce a warning here is the following: We check the set of potentially started threads calling `free`, removing those which will definitely be must-joined at the point when `free` is called.

Here, however it is the **current thread** that will be must joined at the point at which free is called.

That seems like it is somehow a separate issue from checking whether any of the threads calling free can already have been started?

michael-schwarz: Ideally, we'd also want to add a test for the case which your current implementation now catches.

src/analyses/useAfterFree.ml
mrstanb: In case a `join` occurs whenever one performs a side-effect via `ctx.sideg`, then I guess this accumulation here is indeed redundant. Furthermore, since the range of the map lattice, representing `G`, is a must set, I guess that a set intersection would be performed inside the side-effect by default.

If my assumptions are correct, then I guess we could it remove this.

src/witness/svcomp.ml
michael-schwarz: > Should we also add a `valid-memsafety.prp` file in `tests/sv-comp` and write some test code for this?

Sounds like a good idea! 

> does this make (at least some) sense?

@sim642 did you get around to having a look at Stanimir's changes?

sim642: Not yet, but I'll try to soon.

src/witness/svcompSpec.ml
```diff
@@ -23,6 +27,16 @@ let of_string s =
...
+  else if Str.string_match regexp s 0 then
+    let global = Str.matched_group 1 s in
+    if global = "valid-free" then
+      ValidFree
+    else if global = "valid-deref" then
+      ValidDeref
+    else if global = "valid-memtrack" then
+      ValidMemtrack
+    else
+      failwith "Svcomp.Specification.of_string: unknown global expression"
```
sim642, line 39: The problem with this is that the property file in SV-COMP doesn't contain just one, but all three at once: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/3d65c76d8521ef5bc79077a31e7b7e41dd077309/c/properties/valid-memsafety.prp.
But we can leave it for now if the idea is to at least test one of them specifically. I can do larger refactoring of the SV-COMP stuff to allow for multi-property.

It just means that as-is, this isn't ready for `valid-memsafety.prp` from SV-COMP.

src/framework/analysisState.ml
```diff
@@ -7,6 +7,15 @@ let should_warn = ref false
...
 let svcomp_may_overflow = ref false
 
+(** Whether a Use-After-Free (UAF) happened *)
+let svcomp_may_use_after_free = ref false
+
+(** Whether an invalid pointer dereference happened *)
+let svcomp_may_invalid_deref = ref false
+
+(** Whether an invalid memtrack happened *)
+let svcomp_may_invalid_memtrack = ref false
```
sim642, line 17: This still treats use-after-free as equivalent to invalid-free, but as I mentioned above, that's not the case. A UAF constitutes as an invalid-deref of memory which has been freed, while the freeing itself was completely valid.

So UAF is a subclass of invalid-deref. But being a subclass means we can't soundly analyze that unless we also find every other invalid-deref, not just UAF.
Also, competing in valid-memsafety requires verifying all three properties simultaneously. If we don't analyze the other two, we also cannot output sound results. So I'm doubtful that we'll be able to do anything meaningful in the actual SV-COMP scenario.

sim642 (changes requested):

tests/regression/74-use_after_free/06-uaf-struct.c
```diff
@@ -17,13 +17,15 @@ int main(int argc, char **argv) {
...
+        printf("[ auth = %p, service = %p ]\n", auth, service); //NOWARN
 
         if (fgets(line, sizeof(line), stdin) == NULL) break;
         
         if (strncmp(line, "auth ", 5) == 0) {
-            auth = malloc(sizeof(auth)); //WARN
-            memset(auth, 0, sizeof(auth)); //WARN
+            // No deref happening in either of the 2 lines below => no need to warn
+            auth = malloc(sizeof(auth)); //NOWARN
+            memset(auth, 0, sizeof(auth)); //NOWARN
```
michael-schwarz, line 28: Isn't the memset in fact equivalent  to a deref here? I think probably we need to do something in the special function now where we check for use-after-free when the library description says `w_deep` or `r_deep`?

tests/regression/74-use_after_free/11-wrapper-funs-uaf.c
```diff
@@ -27,12 +27,14 @@ int main(int argc, char const *argv[]) {
     my_free2(p);
 
     *(p + 42) = 'c'; //WARN
-    printf("%s", p); //WARN
-
-    char *p2 = p; //WARN
+    // No dereferencing in the line below => no need to warn
+    printf("%s", p); //NOWARN
```
michael-schwarz, line 31: Doesn't the %s implicitly dereference here?

tests/regression/74-use_after_free/06-uaf-struct.c
michael-schwarz: Or maybe even `r`, I don't recall if `r` refers to the pointer being dereferenced to read in our library description. Maybe @sim642 or @karoliineh recalls?

mrstanb: Good point, thanks for the tip! I saw that `memset` has a `w` access kind in `libraryFunctions.ml`. It should be treated as a dereference imo, since it internally dereferences the pointer anyways.

tests/regression/74-use_after_free/11-wrapper-funs-uaf.c
mrstanb: I believe it could be considered a dereference for reading the pointer's contents.

src/framework/analysisState.ml
mrstanb: The issue with the correct classification of UAF as `invalid-deref` is addressed and should be now fixed by https://github.com/goblint/analyzer/pull/1123/commits/d49db71fee1a9ffc125cf373eaa0bf2dab0baa58

src/witness/svcompSpec.ml
mrstanb: I'd suggest that we discuss and implement support for supporting all three properties, so that we can try and let all relevant analyses for `memory safety` run together in order to be able to evaluate how we are actually performing in SV-COMP for `memory safety`.

tests/regression/74-use_after_free/11-wrapper-funs-uaf.c
mrstanb: I looked into a `printf` implementation and it looks like the pointer is indeed dereferenced.

tests/regression/74-use_after_free/13-multi-threaded-uaf-with-joined-thread.c
mrstanb: > That seems like it is somehow a separate issue from checking whether any of the threads calling free can already have been started?

I think I found a solution for removing the WARN now. Cf. https://github.com/goblint/analyzer/pull/1123/commits/cd95e8d45dbcaf68adacd3508cc5938eec4d500c for the solution idea.

mrstanb: > Ideally, we'd also want to add a test for the case which your current implementation now catches.

I believe we already had one covering the implementation before https://github.com/goblint/analyzer/commit/cd95e8d45dbcaf68adacd3508cc5938eec4d500c, i.e., with the "current implementation" from a few weeks ago, namely test case `74/12`.

tests/regression/74-use_after_free/06-uaf-struct.c
mrstanb: Implicit dereferences, such as `memset` should be now covered by https://github.com/goblint/analyzer/pull/1123/commits/66f0c9da414be2dc0d989a9dd22db35978bf00ef

tests/regression/74-use_after_free/11-wrapper-funs-uaf.c
mrstanb: `printf` should now also be treated as an implicit dereference (cf. https://github.com/goblint/analyzer/pull/1123/commits/66f0c9da414be2dc0d989a9dd22db35978bf00ef)

michael-schwarz (approved):
LGTM!

michael-schwarz:
@sim642 If you still want to review this, it would be nice to do it soon, we hope to merge most of Stanimir's things before he defends his thesis in two weeks.

src/witness/svcompSpec.ml
```diff
@@ -23,6 +27,16 @@ let of_string s =
         UnreachCall f
       else
         failwith "Svcomp.Specification.of_string: unknown global not expression"
+  else if Str.string_match regexp s 0 then
```
mrstanb, line 30: As an info:
* This part here has been changed in a separate branch, used for the SV-COMP "Memory Safety" benchmarks
* It will be added to Goblint as part of another PR

src/analyses/useAfterFree.ml
```diff
@@ -24,83 +24,110 @@ struct
 
   (* HELPER FUNCTIONS *)
 
+  let set_global_svcomp_var is_double_free =
+    if is_double_free then
+      AnalysisState.svcomp_may_invalid_free := true
+    else
+      AnalysisState.svcomp_may_invalid_deref := true
+
```
mrstanb, line 32: This is also changed in the branch for the SV-COMP "Memory Safety" benchmarks and will be added to Goblint in a separate PR

src/analyses/useAfterFree.ml
```diff
@@ -175,9 +207,16 @@ struct
...
     let desc = LibraryFunctions.find f in
+    let is_arg_implicitly_derefed arg =
+      let read_shallow_args = LibraryDesc.Accesses.find desc.accs { kind = Read; deep = false } arglist in
+      let read_deep_args = LibraryDesc.Accesses.find desc.accs { kind = Read; deep = true } arglist in
+      let write_shallow_args = LibraryDesc.Accesses.find desc.accs { kind = Write; deep = false } arglist in
+      let write_deep_args = LibraryDesc.Accesses.find desc.accs { kind = Write; deep = true } arglist in
+      List.mem arg read_shallow_args || List.mem arg read_deep_args || List.mem arg write_shallow_args || List.mem arg write_deep_args
+    in
+    Option.iter (fun x -> warn_lval_might_contain_freed ("special: " ^ f.vname) ctx x) lval;
+    List.iter (fun arg -> warn_exp_might_contain_freed ~is_implicitly_derefed:(is_arg_implicitly_derefed arg) ~is_double_free:(f.vname = "free") ("special: " ^ f.vname) ctx arg) arglist;
```
sim642, line 219: `f.vname = "free"` is a very hard-coded check. It should instead look up `LibraryFunctions` for `Free` accesses.

src/witness/svcompSpec.ml
sim642: This stuff was run on SV-COMP but that was done per-specification or? because there's no handling of all three simultaneously here still.

mrstanb:
> This stuff was run on SV-COMP but that was done per-specification or? because there's no handling of all three simultaneously here still.

We added a workaround by having an internal "memory-safety" property which unites all three. This way we were able to let all three run and be tested simultaneously for the benchmarks

src/analyses/useAfterFree.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1123/commits/75f15808590ae9e595bc3857ac353cd0cf595a57

src/witness/svcompSpec.ml
mrstanb: We added a workaround by having an internal "memory-safety" property which unites all three. This way we were able to let all three run and be tested simultaneously for the benchmarks

michael-schwarz: To clarify, this will be fixed in a follow-up PR, which will contain commits from the https://github.com/goblint/analyzer/tree/staging_memsafety branch, we just don't want to include them here yet to avoid polluting the history.

sim642 (approved):

michael-schwarz:
@mrstanb can you resolve the merge conflict with master? Then we should be good to go!

mrstanb:
> @mrstanb can you resolve the merge conflict with master? Then we should be good to go!

Sure, right away.

PR #1124, opened 2023-08-02
Unassume benchmarking fixes

sim642:
### Changes
1. Disable witness lifter if GraphML witness generation disabled (but SV-COMP is still enabled for BenchExec use).
2. Special case `calloc` with count 1 in base (cherry-picked from #978).
3. Fix var_eq `may_change` for other argument being constant.
4. Fix fixpoint issue from #1126.
5. Handle `all_index` in `evalbinop_base`.
6. Disable CIL check in unassume because variables of typedef types fail.
7. Fix two unassume literature witnesses.
8. Improve some tracing.

### DIFF
--- a/src/cdomains/valueDomain.ml
+++ b/src/cdomains/valueDomain.ml
@@ -115,7 +115,7 @@ struct
     | _ -> false
 
   let is_mutex_type (t: typ): bool = match t with
-    | TNamed (info, attr) -> info.tname = "pthread_mutex_t" || info.tname = "spinlock_t" || info.tname = "pthread_spinlock_t"
+    | TNamed (info, attr) -> info.tname = "pthread_mutex_t" || info.tname = "spinlock_t" || info.tname = "pthread_spinlock_t" || info.tname = "pthread_cond_t" || info.tname = "pthread_mutexattr_t"
     | TInt (IInt, attr) -> hasAttribute "mutex" attr
     | _ -> false
 
--- a/src/cdomains/valueDomain.ml
+++ b/src/cdomains/valueDomain.ml
@@ -115,7 +115,7 @@ struct
     | _ -> false
 
   let is_mutex_type (t: typ): bool = match t with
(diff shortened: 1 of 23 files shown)


src/cdomains/valueDomain.ml
```diff
@@ -115,7 +115,7 @@ struct
     | _ -> false
 
   let is_mutex_type (t: typ): bool = match t with
-    | TNamed (info, attr) -> info.tname = "pthread_mutex_t" || info.tname = "spinlock_t" || info.tname = "pthread_spinlock_t"
+    | TNamed (info, attr) -> info.tname = "pthread_mutex_t" || info.tname = "spinlock_t" || info.tname = "pthread_spinlock_t" || info.tname = "pthread_cond_t"
```
michael-schwarz, line 118: Why are we treating `pthread_cond_t` as a mutex type here? That is confusing.

michael-schwarz (approved):
Other than one small comment, LGTM!

src/cdomains/valueDomain.ml
sim642: It wouldn't be more consistent though because it still uses `Mutex` for `ValueDomain`. I don't think it's worth it to introduce `Cond` into `ValueDomain` just for this. Mutex analyses treat condition variables via unlock-lock anyway.

michael-schwarz: > Mutex analyses treat condition variables via unlock-lock anyway.

What do you mean by this? If there is an analysis that does this, it is wrong. (Or are you talking about the second argument of `pthread_cond_wait`? That seems to be unrelated, no?

michael-schwarz: And we do something with condition variables, see #520.

sim642: Well yes, but we don't care about the implementation detail struct of a condition variable in base, just like we don't care about it for mutexes. All this does is make the contents opaque instead of tracking a struct whose contents our `special` functions don't update and thus those contents are outright wrong.

Issue #1119, opened 2023-07-26
Queries.BlobSize returns the contents of a blob instead of its size


mrstanb:
## Description
I noticed that when calling `Queries.BlobSize` on a pointer, containing dynamically-allocated memory, `BlobSize` returns the contents of the blob that the pointer points to instead of the blob's size.

Take the following C program as an example:
```c
#include <stdlib.h>

int main(int argc, char const *argv[]) {
    char *ptr = malloc(5 * sizeof(char));

    *ptr = 'a';

    free(ptr); //BlobSize

    return 0;
}
```

## Observed Behavior
Calling `BlobSize` for `ptr` in the program above for the line with the comment `//BlobSize` returns the contents of `ptr`'s blob, namely `97` (i.e., the char `a` in decimal ASCII encoding).

## Expected Behavior
Calling `BlobSize` for `ptr` in the program above for the line with the comment `//BlobSize` should return `5`, since `ptr` points to a heap memory block with size `5 * sizeof(char)` which should evaluate to `5`

## Further remarks
I noticed that in the `match` in `base.ml`'s `query` function [here](https://github.com/goblint/analyzer/blob/master/src/analyses/base.ml#L1266-L1268), the second `match` case is always taken, i.e., `BlobSize` always gives back `top` and never matches the `Blob (_,s,_)` case.

sim642:
Nothing right now uses `BlobSize`, so it's not surprising that it broke at one point when modifying the blob domain.

PR #1125, opened 2023-08-03
Fix `Queries.BlobSize`

mrstanb:
This PR is an attempt to fix the issue described in #1119:
* `make test` seems to pass just fine locally -> hence, hopefully nothing is broken
* With the test program from #1119 (cf. below), calling `Queries.BlobSize` for the pointer `ptr` in the line with the comment `//BlobSize` now seems to return the size `5`, respectively `[5, 5]` when using intervals.

Test program from #1119:
```c
#include <stdlib.h>

    char *ptr = malloc(5 * sizeof(char));

    *ptr = 'a';

    free(ptr); //BlobSize

    return 0;
}
```

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1261,11 +1261,22 @@ struct
         (* ignore @@ printf "BlobSize %a MayPointTo %a\n" d_plainexp e VD.pretty p; *)
         match p with
         | Address a ->
-          let r = get ~full:true (Analyses.ask_of_ctx ctx) ctx.global ctx.local a  None in
-          (* ignore @@ printf "BlobSize %a = %a\n" d_plainexp e VD.pretty r; *)
-          (match r with
-           | Blob (_,s,_) -> `Lifted s
-           | _ -> Queries.Result.top q)
+          let s = addrToLvalSet a in
+          let has_offset = function
+            | `NoOffset -> false
+            | `Field _
+            | `Index _ -> true
+          in
+          (* If there's a non-heap var or an offset in the lval set, we answer with bottom *)
+          if ValueDomainQueries.LS.exists (fun (v, o) -> (not @@ ctx.ask (Queries.IsHeapVar v)) || has_offset o) s then
+            Queries.Result.bot q
+          else (
+            let r = get ~full:true (Analyses.ask_of_ctx ctx) ctx.global ctx.local a  None in
+            (* ignore @@ printf "BlobSize %a = %a\n" d_plainexp e VD.pretty r; *)
+            (match r with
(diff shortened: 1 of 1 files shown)


src/analyses/base.ml
```diff
@@ -462,7 +462,7 @@ struct
         let var = get_var a gs st x in
         let v = VD.eval_offset (Queries.to_value_domain_ask a) (fun x -> get a gs st x exp) var offs exp (Some (Var x, Offs.to_cil_offset offs)) x.vtype in
         if M.tracing then M.tracec "get" "var = %a, %a = %a\n" VD.pretty var AD.pretty (AD.of_mval (x, offs)) VD.pretty v;
-        if full then v else match v with
+        if full then var else match v with
           | Blob (c,s,_) -> c
           | x -> x
```
sim642, line 467: It's not clear to me whether this is the right thing. It looks to me that the `~full:true` argument is just supposed to avoid the usual `Blob` unwrapping because `BlobSize` needs to take the size from there.
Using `var` instead of `v` means that offsets are actually ignored which seems odd. Is that really the right thing in general?

sim642: I thought about this a bit more and maybe have some answer to my own question.
The variable and its offsets here are already resolved mvalues in an address set, e.g. `x.foo` or `(alloc@...).bar`. Offsets within the blob don't matter for the size of the blob, so I suppose it's safe to ignore them.

In this light, it becomes questionable what a `BlobSize (alloc@...).bar` should even be. The least error prone thing would be to forbid offsets in the query altogether (if not by crashing, then by returning `bot`). Similarly, if the variable isn't a heap variable, then the same thing could apply.

michael-schwarz: >  The least error prone thing would be to forbid offsets in the query altogether (if not by crashing, then by returning bot). Similarly, if the variable isn't a heap variable, then the same thing could apply.

If we do this, I think it'll be good to merge after!

mrstanb: > > The least error prone thing would be to forbid offsets in the query altogether (if not by crashing, then by returning bot). Similarly, if the variable isn't a heap variable, then the same thing could apply.
> 
> If we do this, I think it'll be good to merge after!

I added some changes in https://github.com/goblint/analyzer/pull/1125/commits/15a5b1a1470014017df01ba51e1ddf41c35a157f which should hopefully reflect what was suggested in the discussion here.

`make test` runs through successfully, so hopefully the changes make sense

michael-schwarz (approved):

sim642 (approved):

mrstanb:
Is this good to merge?

PR #1127, opened 2023-08-04
Memory Leak Analysis

mrstanb:
This PR contains a first simple version of a memory leak analysis:
* Should work for single-threaded programs provided that `ana.malloc.unique_address_count` is set to a value > 0
* Currently uses a simple heuristic to warn for potential memory leaks whenever the program is multi-threaded

`make test` fails for test case `76/02` if one uses a comment `//WARN` on line 7 of this test case. The error message is:
`Expected warn, but registered other`
Because of this I've removed the `//WARN` from this test case for now

### DIFF
--- a/src/analyses/memLeak.ml
+++ b/src/analyses/memLeak.ml
@@ -36,7 +36,7 @@ struct
     let state = ctx.local in
     (* TODO: Is this too hacky of a solution? *)
     if f.svar.vname = "main" && not @@ D.is_empty state then
-      M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "Memory leak from function \"%s\": %a\n" f.svar.vname D.pretty state;
+      M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "Memory leak from function \"%s\": %a" f.svar.vname D.pretty state;
     state
 
   let enter ctx (lval:lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
--- a/src/analyses/memLeak.ml
+++ b/src/analyses/memLeak.ml
@@ -25,10 +25,9 @@ struct
   let check_for_mem_leak ?(assert_exp_imprecise = false) ?(exp = None) ctx =
     let state = ctx.local in
     if not @@ D.is_empty state then
-      if assert_exp_imprecise && Option.is_some exp then
-        M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "assert expression %a is unknown. Memory leak might possibly occur for heap variables: %a" d_exp (Option.get exp) D.pretty state
-      else
-        M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "Memory leak detected for heap variables: %a" D.pretty state
+      match assert_exp_imprecise, exp with
+      | true, Some exp -> M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "assert expression %a is unknown. Memory leak might possibly occur for heap variables: %a" d_exp exp D.pretty state
+      | _ -> M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "Memory leak detected for heap variables: %a" D.pretty state
 
   (* TRANSFER FUNCTIONS *)
   let assign ctx (lval:lval) (rval:exp) : D.t =
--- a/src/analyses/memLeak.ml
+++ b/src/analyses/memLeak.ml
@@ -8,7 +8,7 @@ module ToppedVarInfoSet = SetDomain.ToppedSet(CilType.Varinfo)(struct let topnam
 
 module Spec : Analyses.MCPSpec =
 struct
-  include Analyses.DefaultSpec
+  include Analyses.IdentitySpec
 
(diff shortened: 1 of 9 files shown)


tests/regression/76-memleak/02-simple-mem-leak.c
```diff
@@ -0,0 +1,8 @@
+//PARAM: --set ana.malloc.unique_address_count 1 --set ana.activated[+] memLeak
+#include <stdlib.h>
+
+int main(int argc, char const *argv[]) {
+    int *p = malloc(sizeof(int));
+    // No free => memory is leaked
+    return 0; //TODO: `make test` detects OTHER and not WARN here
```
michael-schwarz, line 7: Did you inspect what causes the test to report other? Maybe by tweaking the printing options a little bit, we can get it to produce the warning we want?

src/analyses/memLeak.ml
```diff
@@ -0,0 +1,87 @@
...
+      warn_for_multi_threaded ctx;
+      begin match ctx.ask Queries.HeapVar with
+        | `Lifted var -> D.add var state
+        | _ -> state
+      end
+    | Free ptr ->
+      begin match ctx.ask (Queries.MayPointTo ptr) with
+        | a when not (Queries.LS.is_top a) && not (Queries.LS.mem (dummyFunDec.svar, `NoOffset) a) ->
+          (* TODO: Need to always set "ana.malloc.unique_address_count" to smth > 0 *)
+          let unique_pointed_to_heap_vars =
```
michael-schwarz, line 68: Is this correct? Not **all** potentially pointed to HeapVars should be removed here, but only those for which `free` is definitely called, i.e., when this points-to set is singleton.

src/analyses/memLeak.ml
```diff
@@ -0,0 +1,87 @@
...
+
+  let branch ctx (exp:exp) (tv:bool) : D.t =
+    ctx.local
+
+  let body ctx (f:fundec) : D.t =
+    ctx.local
+
+  let return ctx (exp:exp option) (f:fundec) : D.t =
+    let state = ctx.local in
+    (* TODO: Is this too hacky of a solution? *)
```
michael-schwarz, line 37: I think it is missing other ways in which a function may be left, such as `exit` and so on. Probably, handling of the `Abort` special function is needed here.

src/analyses/memLeak.ml
mrstanb: Yes, I had completely forgotten about this. Thanks for the tip!

src/analyses/memLeak.ml
mrstanb: Thanks for the tip! I previously tried matching the `Abort` function in the `special` transfer function and debug printing some stuff to see if anything happens upon termination via `exit`, but I didn't get anything meaningful there and so I didn't even consider it anymore afterwards.

I'll try `Abort` in the `return` transfer function

michael-schwarz: No, it should in fact be handled by the `special` transfer function. Strange that it didn't show up!

mrstanb: Forget about the previous comment. I got confused. I'll add the checks for `Abort` in `special` as I had tried earlier.

mrstanb: I'll also add a simple check for `Assert` in case assert's expression can be evaluated to something which is 0, i.e., false.

michael-schwarz: > Should we consider supporting them as well?

Yes, that would be a good idea! I'd propose to do it in a separate PR.

michael-schwarz (approved):
LGTM!

src/analyses/memLeak.ml
```diff
@@ -0,0 +1,112 @@
...
+
+  (* TRANSFER FUNCTIONS *)
+  let assign ctx (lval:lval) (rval:exp) : D.t =
+    ctx.local
+
+  let branch ctx (exp:exp) (tv:bool) : D.t =
+    ctx.local
+
+  let body ctx (f:fundec) : D.t =
+    ctx.local
```
sim642, line 40: Could use `IdentitySpec` to avoid a bunch of these.

sim642 (approved):

src/analyses/memLeak.ml
mrstanb: Thanks for the hint! Should be covered in https://github.com/goblint/analyzer/pull/1127/commits/c3c4bb1f96077c17276a4bd1ba5aec52ef7c42dd

src/analyses/memLeak.ml
mrstanb: > I'd propose to do it in a separate PR

Opened #1135 for `quick_exit` support.

mrstanb:
Feel free to merge.

I'll open a new PR to update the `memLeak` analysis with support for `quick_exit` as soon as #1135 is also ready and merged.

mrstanb:
> I'll open a new PR to update the memLeak analysis with support for quick_exit as soon as https://github.com/goblint/analyzer/pull/1135 is also ready and merged.

Actually, I think it's unnecessary to open a new PR for this. I'll just update this one with a merge commit from `master`

mrstanb:
Merged with `master` after #1135 was done. Added a regression test which now also covers a memory leak with `quick_exit()`. `make test` runs through fine with it.

PR #1137, opened 2023-08-16
Upgrade to OCaml 5.0 as default

### LINKED ISSUE #1003: Upgrade to OCaml 5.0.0
Upgrade to OCaml 5.0.0 to support new language features e.g., [shared memory parallelism](https://v2.ocaml.org/releases/5.0/manual/parallelism.html).

List of tasks for goblint/analyzer

- [x] mlgmpodl 1.2.14 -> 1.2.15 (not yet in opam, https://github.com/nberth/mlgmpidl/issues/17)
- [x] ocaml-config 2 -> 3 
- [x] apron v0.9.13 -> latest (https://github.com/ocaml/opam-repository/pull/24090)

List of tasks for goblint/gobview

- [x] jsoo-react.dev -> latest

### LINKED ISSUE #1133: Make setup fails due to "invalid extra arguments" to opam switch create
This happens when one runs make setup:
+ opam switch -y create . --deps-only ocaml-variants.4.14.0+options ocaml-option-flambda --locked
[ERROR] Invalid extra arguments ocaml-variants.4.14.0+options ocaml-option-flambda
gmake: *** [Makefile:4: setup] Error 2

sim642:
Closes #1003.

We can finally do this because mlgmpidl and apron have been released on opam for OCaml 5.0.

### TODO
- [ ] Merge https://github.com/goblint/gobview/pull/26.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,7 +18,7 @@ jobs:
         os:
           - ubuntu-latest
         ocaml-compiler:
-          - ocaml-variants.4.14.0+options,ocaml-option-flambda # matches opam lock file
+          - ocaml-variants.5.0.0+options,ocaml-option-flambda # matches opam lock file
           # don't add any other because they won't be used
 
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,7 +23,7 @@ jobs:
         os:
           - ubuntu-latest
         ocaml-compiler:
-          - ocaml-variants.4.14.0+options,ocaml-option-flambda # matches opam lock file
+          - ocaml-variants.5.0.0+options,ocaml-option-flambda # matches opam lock file
           # don't add any other because they won't be used
 
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/indentation.yml
+++ b/.github/workflows/indentation.yml
@@ -12,7 +12,7 @@ jobs:
         os:
           - ubuntu-latest
         ocaml-compiler:
-          - 4.14.x
+          - 5.0.x
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/locked.yml
+++ b/.github/workflows/locked.yml
@@ -20,7 +20,7 @@ jobs:
           - ubuntu-latest
           - macos-latest
         ocaml-compiler:
-          - ocaml-variants.4.14.0+options,ocaml-option-flambda # matches opam lock file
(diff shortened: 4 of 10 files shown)


sim642:
I did an sv-benchmarks run comparing OCaml 4.14 with OCaml 5.0 and the results are concerning. We might have to postpone making OCaml 5.0 the default and simply cherry-pick the changes from here to at least make Goblint OCaml 5.0 compatible, even if we don't use it by default.

Memory usage is significantly higher, almost 10× in some cases:
image

CPU time is also affected but that could be simply due to working with more memory:
image

([Full results](https://goblint.cs.ut.ee/results/127-all-fast-ocaml5-after/results/results.2023-09-12_07-54-42.table.html#/))

michael-schwarz:
Should we investigate what the performance is like with OCaml 5.1?

michael-schwarz:
Sounds like 5.2 (released today) also won't fix our problems: https://discuss.ocaml.org/t/ocaml-5-2-0-released/14638

> OCaml 5.2.0 is still a somewhat experimental release compared to the OCaml 4.14 branch. In particular
>-  [...]
> - Ephemeron performances need to be investigated.

FelixKrayer:
Just as a quick note here:
The version of the dependency of `camlidl` will have to be changed to version `1.12`, as the current `camlidl = 1.11` requires `ocaml < 5.0`.

sim642:
All the dependencies in the Goblint lock file are for 4.14 anyway: the lock file also specifies OCaml 4.14. The non-locked dependencies allow appropriate versions to be installed on OCaml 5.0, 5.1 and 5.2.

sim642:
I might have to redo the 4.14 vs 5.x (5.3 now) runs now because a lot seems to have changed (in Goblint). I tried a few worst cases from my previous runs to see if they remain, but couldn't find an example of regression due to OCaml. Rather, some just got faster/slower on OCaml 4.14 even.

Having some significant examples of regression would allow us to open an issue with OCaml. The maintainers are interested in cases where performance regressions from OCaml 5 are holding things back: https://discuss.ocaml.org/t/ocaml-5-3-0-released/15916/4.

DrMichaelPetter:
If it helps, I can set up a regular job on our servers, e.g. comparing runtimes of a sample set. We would need some way of building 4.14 and 5.x in an unsupervised fashion, though.

sim642:
I now opened an issue with OCaml, including an updated benchmark for the current Goblint and OCaml 5.3: https://github.com/ocaml/ocaml/issues/13733. The large blob of 10× memory usage cases seems gone, but performance isn't clearly on par with 4.14 either.

PR #1139, opened 2023-08-18
Detect memory deallocation at offset from memory start

mrstanb:
This PR is supposed to improve Goblint's detection of `invalid-free` bugs by adding a check for memory deallocation which happens at an offset from the memory's start address instead of at the start address itself.

Added a few more regression test cases for this as well. `make test` runs through successfully

Cf. https://cwe.mitre.org/data/definitions/761.html

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2016,7 +2016,7 @@ struct
         M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Points-to set for pointer %a in function %s is top. Potentially invalid memory deallocation may occur" d_exp ptr special_fn.vname
       else if (Q.LS.exists (fun (v, _) -> not (ctx.ask (Q.IsHeapVar v))) points_to_set) || (AD.mem Addr.UnknownPtr a) then
         M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Free of non-dynamically allocated memory in function %s for pointer %a" special_fn.vname d_exp ptr
-      else if Q.LS.exists (fun (_, o) -> o <> `NoOffset) points_to_set then
+      else if Q.LS.exists (fun (_, o) -> Offset.Exp.cmp_zero_offset o <> `MustZero) points_to_set then
         M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 761] "Free of memory not at start of buffer in function %s for pointer %a" special_fn.vname d_exp ptr
     | _ -> M.warn ~category:MessageCategory.Analyzer "Pointer %a in function %s doesn't evaluate to a valid address." d_exp ptr special_fn.vname
 
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2013,14 +2013,21 @@ struct
     let st' = invalidate ~deep:false ~ctx (Analyses.ask_of_ctx ctx) gs st shallow_addrs in
     invalidate ~deep:true ~ctx (Analyses.ask_of_ctx ctx) gs st' deep_addrs
 
-  let check_free_of_non_heap_mem ctx special_fn ptr =
+  let check_invalid_mem_dealloc ctx special_fn ptr =
+    let has_offset = function
+      | `NoOffset -> false
+      | `Field _
+      | `Index _ -> true
+    in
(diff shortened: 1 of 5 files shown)


src/analyses/base.ml
```diff
@@ -138,6 +138,11 @@ struct
   let project ask p_opt cpa fundec =
     CPA.mapi (fun varinfo value -> project_val ask (attributes_varinfo varinfo fundec) p_opt value (is_privglob varinfo)) cpa
 
+  let has_offset = function
+    | `NoOffset -> false
+    | `Field _
+    | `Index _ -> true
+
```
sim642, line 145: It might actually even be cleaner to not have this function at all and just use `o <> `NoOffset` whereever it's needed.

mrstanb: Even better in this case, yes. I wasn't sure if one could do such comparisons in OCaml. Thanks for the tip!

mrstanb: Applied change in https://github.com/goblint/analyzer/pull/1139/commits/cb811e1f6d6ff42c4f3f62f93d18c34c767fad75

tests/regression/75-invalid_dealloc/07-free-at-struct-offset.c
```diff
@@ -0,0 +1,15 @@
...
+typedef struct custom_t {
+    char *x;
+    int y;
+} custom_t;
+
+int main(int argc, char const *argv[]) {
+    custom_t *struct_ptr = malloc(sizeof(custom_t));
+    struct_ptr->x = malloc(10 * sizeof(char));
+    free(struct_ptr->x); //NOWARN
+    free(struct_ptr->y); //WARN
```
sim642, line 12: I suspect this isn't checking the right thing. This is freeing the unknown integer in the field (which also probably comes with some unknown address warning) but it's not checking whether the freeing is of the struct field's address. The latter would be `free(&struct_ptr->y)` I think, which should also warn.

A more interesting case is `free(&struct_ptr->x)`. I suspect that might currently warn, although it shouldn't, because the first field address is the same as the entire struct. It's still a weird thing to do (and we could warn about it), but the freeing would be fine.
So more precise handling might even be to use `Mval.Exp.cmp_zero_offset` to check the offset.

mrstanb: > A more interesting case is free(&struct_ptr->x). I suspect that might currently warn, although it shouldn't, because the first field address is the same as the entire struct. It's still a weird thing to do (and we could warn about it), but the freeing would be fine.

It does indeed perform a warning for CWE 761, although it should be totally fine in this case, because it was something that was `malloc`'d and should be `free`'d

mrstanb: > So more precise handling might even be to use Mval.Exp.cmp_zero_offset to check the offset.

I'll check it out

mrstanb: I used `cmp_zero_offset` in https://github.com/goblint/analyzer/pull/1139/commits/92c0dcae14badbb2be23a62d7c5e15951f877bc2 and it seems to improve the precision.

I see an issue with a false positive warning now, however. More specifically, for the following case::
```c
typedef struct custom_t {
    int y;
    char *y;
} custom_t;
```

If we call `free` on the struct's first field -> no warning happens (so it's even unsound as it seems in this case)
If we call `free` on the struct's second field (provided that it was dynamically allocated) -> a warning pops up (a false positive)

I guess we should try to recursively check for invalid deallocations in order to cover such cases. What do you think?

sim642: It's not clear to me why recursive checking should be the right thing. In C freeing the outer structure doesn't recursively free other dynamic memory that's pointed to by some field. That needs to be explicitly and separately deallocated.

sim642: Is there a test case where the improved precision from `cmp_zero_offset` can be seen? If not, then that could be added.

mrstanb: > Is there a test case where the improved precision from cmp_zero_offset can be seen? If not, then that could be added.

Yep, both `75/07` and `75/08` were warning for both struct fields before the `cmp_zero_offset` changes. After the `cmp_zero_offset` changes, they both warn only for the 2nd struct field, as expected.

michael-schwarz (approved):

sim642 (approved):

PR #1142, opened 2023-08-22
Use `AddressDomain` for `MayPointTo` and `ReachableFrom` queries

karoliineh:
### Changes
1. Change `MayPointTo` and `ReachableFrom` query result from `LS` to `AddressDomain`. This has numerous benefits:

    1. Avoid lots of conversions when using these queries, e.g. `AD` → `LS` → `AD`/`Set (Varinfo)`/....
    2. `LS` had two different top representations (one for `ToppedSet` entirely and another when `dummyFundec.svar` was a member). The former allows no known pointers along with unknown pointer, the latter is easy to forget/not know about. `AD` has only one explicit representation by containing `UnknownPtr`.
    3. `NullPtr`s are not implicitly excluded from query results, allowing other analyses to know about them at all. For example, now we warn about unlocking `NULL` pointer.
    4. More precision due to fewer conversions. Addresses have `ID` indices which were dropped (if not definite) when converting to `exp` indices in `LS`. Now interval, etc offsets can be seen by other analyses.

2. Fix unsoundness in analyses which previously forgot to check for `dummyFundec` in query results: relation, fileUse, poisonVariables, uninit, var_eq (in `special`).

### TODO
- [x] Use `MayPointToA` and `ReachableFromA` in `taintPartialContexts`.
- [x] Remove queries with `LS`.
- [ ] Add missing `Address`/`Mval`/`Offset` conversion functions.
- [x] Rename variables to match new type, e.g. `ls` → `ad`, `lv` → addr.
- [ ] Review added TODOs in code.


src/analyses/extractPthread.ml
```diff
@@ -877,16 +877,9 @@ module Spec : Analyses.MCPSpec = struct
...
-        if Queries.AD.mem UnknownPtr ad
-        then (* UNSOUND *)
-          Queries.AD.remove UnknownPtr ad
-          |> Queries.AD.to_var_may
-        else
-          Queries.AD.to_var_may ad
-      else
-        []
+      ctx.ask (Queries.MayPointTo exp)
+      |> Queries.AD.remove UnknownPtr
```
michael-schwarz, line 881: Maybe leave the comment that this is unsound?

sim642:
@michael-schwarz Do you intend to review this again or was that all?

michael-schwarz:
It all seems reasonable at first glance, anything you would like me to look into in particular?

sim642:
Not really, I just wasn't sure since you left one small comment previously.

PR #1143, opened 2023-08-28
fix: copying of files to GobView in projects with subdirectories

sallto:
Fixes the copying of source files to the GobView directory. 

The original solution didn't check whether a filesystem entry is a directory which caused the counter (ex. file.**5**.c) to mismatch on projects with subdirectories.

Additionally fixed the filter for special "files" like `<command-line>`. It now skips all entries starting and ending with brackets, since the specific strings (ex. "command-line") differ for languages and OS. (can be split into a separate PR, but I felt that would be too small)


michael-schwarz (approved):

michael-schwarz:
Thank you for the PR @sallto!

PR #1145, opened 2023-08-29
Add marshalling for function calls

sallto:
Depends on #1093 

Adds marshalling of function calls to use in Gobview


michael-schwarz (changes requested):
This.contains all changes from #1093, and can thus not be merged. Please clean it up, or, if @stilscher prefers, rebase your changes onto master once #1093 lands.

michael-schwarz:
As the only relevant commit here is 457273a, and it only has a few lines, it makes little sense to keep this PR around. If the goal is to merge this, please open a new PR with this commit cherry-picked onto master.

PR #1146, opened 2023-08-29
Selenium

sallto:
depends on #1143 and https://github.com/goblint/gobview/pull/28

adds a E2E test for compilation databases
also expands test to check wether files can be opened and the graph tab and breadcrumb exist


michael-schwarz:
Thank you for the contribution. As we are currently not pursuing this, we decided to close it.

Issue #1153, opened 2023-09-09
Tracking bit status information in goblint


edwintorok:
Maybe I am doing something wrong but the congruence analysis doesn't seem to be able to track this.

Here I am getting a random value, telling goblint to assume the state of the LSB, and then immediately assert it, and goblint says the assertion status is unknown (in a real program of course the assertion would be far away from the assume call):
https://gist.github.com/edwintorok/8c6f3de4f717a3145945705cdcbe863b

(I could attempt to implement an analysis specific to tracking the LSB bit, similar to the sign analysis, but if goblint already has code to handle something similar I'd prefer to reuse or perhaps fix that part. For context: tracking the state of this bit is useful for OCaml C stubs where Is_block() and Is_long() macros check the bit).

michael-schwarz:
Thank you, I'll have a look! This should definitely be possible for the congruence domain (after some small tweaks).

sim642:
Besides implementing `bitand` in the congruence domain, `__goblint_assume(x&1)` also requires specific changes to `BaseInvariant` to actually assume that information.

edwintorok:
I tried using modulo (see updated gist), which seems to have some more code, but perhaps needs some improvement too for assume?
BTW I don't *have* to use assume, if there is another way to tell goblint the status of the low bit I'd be happy to change my code to do that (the goal is to tell goblint that memory allocated by OCaml is aligned to at least 2 bytes or not, when using OCaml integers).

michael-schwarz:
What already works on the master branch is if the program is rewritten to explicitly use `x % 2 == 1`, as in this modified example:

~~~C
//PARAM: --enable ana.int.congruence --set sem.int.signed_overflow assume_none
#include <goblint.h>

int foo(long x)
{
  __goblint_assume(x%2 == 1);
  __goblint_check(x%2 == 1);
}

int main()
{
  foo(__VERIFIER_nondet_int());
}
~~~

PR #1154, opened 2023-09-10
Handle some special cases of `bitand` in `invariant` and `Interval`, `DefExc`, and `Congruences`

michael-schwarz:
This handles some special cases of `bitand` where the second argument can be proven to be `1` or `0`, allowing for tracking the value of the LSB as suggested in #1153.

This focuses on this special use case for now, there is lots of potential to become more precise in the remaining cases as well, but for now it is not clear to me where that would be beneficial (and to what extent). 

(Though a principled handling of all bit-wise operations in all supported domains and comparison of the results before and after on sv-comp might be a good topic for a Bachelor's thesis)

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -835,7 +835,8 @@ struct
     | _ ->
       match to_int i1, to_int i2 with
       | Some x, Some y -> (try of_int ik (Ints_t.bitand x y) |> fst with Division_by_zero -> top_of ik)
-      | _, Some y  when Ints_t.equal y Ints_t.one -> of_interval ik (Ints_t.zero, Ints_t.one) |> fst
+      | _, Some y when Ints_t.equal y Ints_t.zero -> of_int ik Ints_t.zero |> fst
+      | _, Some y when Ints_t.equal y Ints_t.one -> of_interval ik (Ints_t.zero, Ints_t.one) |> fst
       | _ -> top_of ik
 
   let bitor  = bit (fun _ik -> Ints_t.bitor)
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -826,7 +826,18 @@ struct
       | _              -> (top_of ik,{underflow=true; overflow=true})
 
   let bitxor = bit (fun _ik -> Ints_t.bitxor)
-  let bitand = bit (fun _ik -> Ints_t.bitand)
+
+  let bitand ik i1 i2 =
+    match is_bot i1, is_bot i2 with
+    | true, true -> bot_of ik
+    | true, _
+    | _   , true -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show i1) (show i2)))
+    | _ ->
+      match to_int i1, to_int i2 with
+      | Some x, Some y -> (try of_int ik (Ints_t.bitand x y) |> fst with Division_by_zero -> top_of ik)
+      | _, Some y  when Ints_t.equal y Ints_t.one -> of_interval ik (Ints_t.zero, Ints_t.one) |> fst
+      | _ -> top_of ik
+
   let bitor  = bit (fun _ik -> Ints_t.bitor)
 
   let bit1 f ik i1 =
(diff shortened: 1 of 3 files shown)


src/cdomains/intDomain.ml
```diff
@@ -3144,7 +3176,19 @@ struct
...
+    | None, None -> None
+    | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
+    | Some (c, m), Some (c', m') ->
+      if (m =: Ints_t.zero && m' =: Ints_t.zero) then
+        (* both arguments constant *)
+        Some (Ints_t.bitand c c', Ints_t.zero)
+      else if m' =: Ints_t.zero && c' =: Ints_t.one && Ints_t.rem m (Ints_t.of_int 2) =: Ints_t.zero then
+        (* x & 1  and  x == c (mod 2*z) *)
+        (* Value is equal to LSB of c *)
+        Some (Ints_t.bitand c c', Ints_t.zero)
```
michael-schwarz, line 3189: @sim642 I would appreciate it of you can also think this part through again: It seems correct now, but I am not 100% confident.

src/cdomains/intDomain.ml
```diff
@@ -2282,7 +2293,28 @@ struct
...
-  let bitand = lift2 BigInt.bitand
+
+  let bitand ik x y = norm ik (match x,y with
+      (* We don't bother with exclusion sets: *)
+      | `Excluded _, `Definite i ->
+        (* Except in two special cases *)
+        if BigInt.equal i BigInt.zero then
+          `Definite BigInt.zero
+        else if BigInt.equal i BigInt.one then
+          of_interval IBool (BigInt.zero, BigInt.one)
```
sim642, line 2304: This is giving the `bitand` result `IBool` ikind which doesn't seem right. I think it should still be the original ikind, just the range is smaller. That's how you've implemented it for intervals as well.

src/cdomains/intDomain.ml
```diff
@@ -2282,7 +2293,28 @@ struct
...
 
   let bitnot = lift1 BigInt.bitnot
-  let bitand = lift2 BigInt.bitand
+
+  let bitand ik x y = norm ik (match x,y with
+      (* We don't bother with exclusion sets: *)
+      | `Excluded _, `Definite i ->
+        (* Except in two special cases *)
+        if BigInt.equal i BigInt.zero then
+          `Definite BigInt.zero
```
sim642, line 2302: Couldn't intervals also have this special case? It only has the 1 case right now.

src/cdomains/intDomain.ml
sim642: This sounds right, although since `c' = 1`, this really is just `bitand c 1`.

Another way to do this might be to just delegate `x & 1` to `x % 2`. If `m` is even, then this seems to match with `rem` implementation. But `rem` also handles the other case where `m` is odd, so maybe we can just delegate to that?

src/cdomains/intDomain.ml
michael-schwarz: I don't think it's problematic, inside this domain the `ikind` is not stored and only used to compute the ranges:

https://github.com/goblint/analyzer/blob/4d18300af5f6ae1b04bdbf29b29775af07fbc96c/src/cdomains/intDomain.ml#L2157

https://github.com/goblint/analyzer/blob/4d18300af5f6ae1b04bdbf29b29775af07fbc96c/src/cdomains/intDomain.ml#L1989

src/cdomains/intDomain.ml
michael-schwarz: I think that would be incorrect, though?

~~~C

#include<stdlib.h>
#include<stdio.h>

int main(){
  int i  = (-5) %2; // -1
  int j = -5&1; // 1
  printf("%i\n",i);
  printf("%i",j);
  return 0;
}
~~~

michael-schwarz: Turns out we are unsound for `x %2` for signed types when x is maybe negative and not constant. See: https://github.com/goblint/analyzer/issues/1156

src/cdomains/intDomain.ml
sim642: Hmm, that makes it quite misleading though. The result is based on `IBool`, not `(BigInt.zero, BigInt.one)` at all. So right now this is the only way anyway.

But we should open an issue about fixing `DefExc.of_interval` to actually derive a bit range from the actual given interval and intersect it with the ikind's range. I didn't check but `starting` and `ending` should work similarly. It might make a difference because base analysis itself uses `of_interval`/`starting`/`ending`.

sim642 (approved):

src/cdomains/intDomain.ml
michael-schwarz: I guess I'll leave the code as it currently is in light of #1159.

Issue #1156, opened 2023-09-11
Goblint unsound for `x % 2` for `x` possibly negative but not constant


michael-schwarz:
~~~C
  int top;

  int c = -5;
  if(top) { c= -7;}
  __goblint_check(c % 2 == 1); // Doesn't hold at runtime
~~~

Goblint claims that this would hold.

For constants it works correctly:

~~~C
  __goblint_check(-5 % 2 == -1);
~~~

michael-schwarz:
One way to fix this is to have the check `if (x % 2 == 1) {` refine x to be positive.

michael-schwarz:
@hseidl and I constructed another example of this issue.

~~~C
// PARAM: --enable ana.int.congruence --enable ana.int.interval --set sem.int.signed_overflow assume_none
#include <goblint.h>

int main()
{
	int x,y;

	if( y > 0) {
	if(x % 3 == 0){
		if(x + y == 5) {
                       // We have  x:  2+3ℤ
                       // But do not make use of it
                        
			if(y % 3 == 2) {
                                 // Neither by showing this to be the case in the __goblint_check
				__goblint_check(y % 3 == 2);
			} else {
				// Nor by finding this to be unreachable
			}

		}
	}
	}
}
~~~

PR #1161, opened 2023-09-12
Congruences: Make `%` sound by restricting cases where we return a constant

michael-schwarz:
For `(c1, m1) % k` with `k` constant only return a constant when

-  `k` divides `m1` (already here before), i.e. result will  be `+/- c1%k`

**and** **one** of the following hold:
- `c1%k` is `0` (sign does not matter)
- `(c1, m1)` is a constant (sign will be correct)
- operation is on unsigned values, i.e., both arguments will be positive.

This addresses the soundness part of #1156, leaving it open how to become more precise again (by making use of the sign of the left argument where it is known).

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -3199,8 +3199,8 @@ struct
     | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
     | Some (c1, m1), Some(c2, m2) ->
       if m2 =: Ints_t.zero then
-        if (c2 |: m1) then
-          Some(c1 %: c2,Ints_t.zero)
+        if (c2 |: m1) && (c1 %: c2 =: Ints_t.zero || m1 =: Ints_t.zero || not (Cil.isSigned ik)) then
+          Some(c1 %: c2, Ints_t.zero)
         else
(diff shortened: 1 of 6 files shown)


sim642 (approved):

ManuelLerchner:
Hi @michael-schwarz! I hope you're enjoying your time in Singapore :)

Shameless plug / heads-up: I used this PR as a small worked example in my [Voblint](https://github.com/ManuelLerchner/Voblint-Verification-Master-Thesis) project.

I think it makes a good example because it is tiny, local to one transfer function, and my toy language can represent it directly. If you know of any better example I could use, I'm interested in suggestions!

For some more context: here is the relevant section: https://manuellerchner.github.io/Voblint-Verification-Master-Thesis/#stakes

michael-schwarz:
Hi @ManuelLerchner, I heard about your project from Michael Petter and Julian Erhard. I think it is an exciting project and great to see progress in the direction of a verified abstract interpreter!

When the time for your defense comes, I'd be interested in hearing your talk (provided you schedule it no later than noon German time, so I can listen to it). Another nudge: Maybe you should thinking about publishing this work? (Just in case you're currently not considering that).  While there is the Verasco project which is hard to compete with, I guess your work is the first to consider mixed-flow sensitivity.

PR #1170, opened 2023-09-14
Use record as `Access.A` type instead of a quintuple

karoliineh:
The quintuple has become difficult to manage. The contents are now named better and make code more easy to read.

### DIFF
--- a/src/analyses/raceAnalysis.ml
+++ b/src/analyses/raceAnalysis.ml
@@ -225,9 +225,9 @@ struct
     | _ ->
       ()
 
-  let side_access ctx (conf, w, loc, e, a) ((memoroot, offset) as memo) =
+  let side_access ctx Access.A.{conf; kind; node; exp; acc} ((memoroot, offset) as memo) =
     if !AnalysisState.should_warn then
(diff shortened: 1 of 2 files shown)


src/analyses/raceAnalysis.ml
```diff
@@ -225,9 +225,9 @@ struct
     | _ ->
       ()
 
-  let side_access ctx (conf, w, loc, e, a) ((memoroot, offset) as memo) =
+  let side_access ctx Access.A.{conf; kind; node; exp; acc} ((memoroot, offset) as memo) =
     if !AnalysisState.should_warn then
-      ctx.sideg (V.access memoroot) (G.create_access (OffsetTrie.singleton offset (`Lifted (Access.AS.singleton (conf, w, loc, e, a)))));
+      ctx.sideg (V.access memoroot) (G.create_access (OffsetTrie.singleton offset (`Lifted (Access.AS.singleton {conf; kind; node; exp; acc}))));
```
sim642, line 230: Looks like there's no need to destruct and reconstruct the record but just pass it on as a single variable.

sim642 (approved):

Issue #1147, opened 2023-08-30
Unsoundness for reading values from partially initialized objects via char pointers


michael-schwarz:
~~~C
#include <stdlib.h>

int main(){
  unsigned char* in = alloca(10);
  in[9] = 0;

  for (; *in; ++in) {
    __goblint_check(1);
  }

  return 0;
}
~~~

According to the standard, there's no Undefined Behavior here, and the assertion `__goblint_check(1)` is reachable, while we claim it is not.

Issue #1029, opened 2023-04-03
Convert all library functions to new specifications


sim642:
Library activations (#1028) won't be precise until all of the old-style things are converted to new ones in appropriate groups:
https://github.com/goblint/analyzer/blob/bcb4de29e22885570ba82f7ec9cee2e661854b7a/src/analyses/libraryFunctions.ml#L359-L400
https://github.com/goblint/analyzer/blob/bcb4de29e22885570ba82f7ec9cee2e661854b7a/src/analyses/libraryFunctions.ml#L500-L980

_Originally posted by @sim642 in https://github.com/goblint/analyzer/pull/996#discussion_r1143067617_

PR #1174, opened 2023-09-18
Add and fix library functions for Concrat benchmarks

sim642:
### Changes
1. Fix incorrect and unsound `return_on_success` value for some pthread locking functions. This caused most of uthash benchmark being dead.
2. Convert remaining old library function `classify` to new specifications. In particular, some benchmark needed precise accesses for `pthread_join` to not deeply invalidate everything via the second argument.
3. Remove now-unnecessary bridge code for old library function `classify`.
4. Add missing library functions for nnn benchmark. Apparently #996 didn't add everything.


michael-schwarz:
> Fix incorrect and unsound `return_on_success` value for some pthread locking functions. This caused most of uthash benchmark being dead.

Might be worth adding tests for this.

src/analyses/libraryFunctions.ml
```diff
@@ -530,6 +568,10 @@ let linux_kernel_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__cmpxchg_wrong_size", special [] Abort);
     ("__xadd_wrong_size", special [] Abort);
     ("__put_user_bad", special [] Abort);
+    ("kmalloc", special [__ "size" []; drop "flags" []] @@ fun size -> Malloc size);
+    ("__kmalloc", special [__ "size" []; drop "flags" []] @@ fun size -> Malloc size);
+    ("kzalloc", special [__ "size" []; drop "flags" []] @@ fun size -> Calloc {count = Cil.one; size});
+    ("usb_alloc_urb", special [__ "iso_packets" []; drop "mem_flags" []] @@ fun iso_packets -> Malloc iso_packets); (* TODO: iso_packets is size in bytes? *)
```
michael-schwarz, line 574: No, it is not. Might make sense to malloc an unknown size here.

src/analyses/libraryFunctions.ml
```diff
@@ -117,12 +118,14 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
     ("wcstombs", unknown ~attrs:[ThreadUnsafe] [drop "dst" [w]; drop "src" [r]; drop "size" []]);
     ("wcsrtombs", unknown ~attrs:[ThreadUnsafe] [drop "dst" [w]; drop "src" [r_deep; w]; drop "size" []; drop "ps" [r_deep; w_deep]]);
+    ("mbstowcs", unknown [drop "dest" [w]; drop "src" [r]; drop "n" []]);
     ("abs", unknown [drop "j" []]);
     ("localtime_r", unknown [drop "timep" [r]; drop "result" [w]]);
     ("strpbrk", unknown [drop "s" [r]; drop "accept" [r]]);
     ("_setjmp", special [__ "env" [w]] @@ fun env -> Setjmp { env }); (* only has one underscore *)
     ("setjmp", special [__ "env" [w]] @@ fun env -> Setjmp { env });
     ("longjmp", special [__ "env" [r]; __ "value" []] @@ fun env value -> Longjmp { env; value });
+    ("atexit", unknown [drop "function" [s]]);
```
michael-schwarz, line 128: I am worried this might be a footgun, as it makes programs that are single-threaded multi-threaded. I would propose to classify this as a special `Atexit` and then prominently warn if we encounter it, and maybe offer an option to spawn this (that should be off by default).

src/analyses/libraryFunctions.ml
```diff
@@ -292,12 +297,33 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("getaddrinfo", unknown [drop "node" [r]; drop "service" [r]; drop "hints" [r_deep]; drop "res" [w]]); (* only write res non-deep because it doesn't write to existing fields of res *)
     ("fnmatch", unknown [drop "pattern" [r]; drop "string" [r]; drop "flags" []]);
     ("realpath", unknown [drop "path" [r]; drop "resolved_path" [w]]);
+    ("memccpy", special [__ "dest" [w]; __ "src" [r]; drop "c" []; drop "n" []] @@ fun dest src -> Memcpy {dest; src});
```
michael-schwarz, line 300: This exposes a shortcoming that we should handle: We do not handle the size argument of `Memcpy`. Also I am a bit worried about classing this as `memcpy` even if that is fixed, as it simply hides the complications here.

michael-schwarz: I created #1177 to track this.

src/analyses/libraryFunctions.ml
```diff
@@ -292,12 +297,33 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("dprintf", unknown (drop "fd" [] :: drop "format" [r] :: VarArgs (drop' [r])));
+    ("vdprintf", unknown [drop "fd" []; drop "format" [r]; drop "ap" [r_deep]]); (* TODO: what to do with a va_list type? is r_deep correct? *)
+    ("mkdtemp", unknown [drop "template" [r; w]]);
+    ("mkstemp", unknown [drop "template" [r; w]]);
+    ("regcomp", unknown [drop "preg" [w_deep]; drop "regex" [r]; drop "cflags" []]);
```
michael-schwarz, line 305: Class as alloc?

src/analyses/libraryFunctions.ml
```diff
@@ -292,12 +297,33 @@ let posix_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("regexec", unknown [drop "preg" [r_deep]; drop "string" [r]; drop "nmatch" []; drop "pmatch" [w_deep]; drop "eflags" []]);
+    ("regfree", unknown [drop "preg" [f_deep]]);
```
michael-schwarz, line 307: Class as free?

src/analyses/libraryFunctions.ml
sim642: It's far from `Malloc` etc though because it doesn't return the allocated stuff. Rather it may just allocate things into a given struct. We'd need a more general resources mechanism to correctly model that.

src/analyses/libraryFunctions.ml
sim642: It doesn't free the given struct itself, rather only things `regcomp` might've allocated deeper down. Treating it as freeing the outermost struct itself would thus be incorrect.

src/analyses/libraryFunctions.ml
sim642: I'll move it to the C group, although we're decades from C23 at this point.

src/analyses/libraryFunctions.ml
sim642: This is how all other library functions with delayed callbacks are specified because that's the only sound way we have right now. Yes, it's imprecise, but as long as nobody wants to immediately implement more precise support for it, this will have to do. That's tracked by already #799 anyway.

If there's no proper support, I don't want to have a special `Atexit` for it which is unhandled and doesn't make the callback code live. That would be worse than not specifying `atexit` at all. Without this, there's a standard "Function definition missing" warning but a custom warning a lot easier to miss when grepping tens of thousands of lines of logs.

By not having been specified, we have been treating it as `r_deep; w_deep; s_deep` implicitly, so this imprecision has always existed. So this specification is slightly more precise to avoid invalidating all globals for example.
To force things to be single-threaded despite this, we also have `exp.single-threaded`.

sim642: Also sv-benchmarks has actual uses of `atexit`, including in our own submitted `goblint-coreutils`, so we cannot afford to be unsound with this.

michael-schwarz: You are of course morally correct. The notable difference is that we usually have `sem.unknown.spawn` off for larger programs, and this will now spawn the things passed to `atexit` unconditionally.  Given how many knobs we have in all other places, I would propose we also add one here, to be able to disable this behavior when we want it.

To me this is an instance of soundiness, where it may be ok to ignore `atexit`, just like we do all the others in #799 if we set an appropriate flag and are very clear about the fact that this option is enabled. Otherwise, I think we are inflicting self-caused wounds when we benchmark, which I don't think is too smart strategically.

What do the others think? @vesalvojdani @karoliineh @stilscher @jerhard?

michael-schwarz: > Also sv-benchmarks has actual uses of atexit, including in our own submitted goblint-coreutils, so we cannot afford to be unsound with this.

Maybe we should add these to `exitfuns` then, it seems needlessly imprecise to go to multi-threaded mode just because of a call to `atexit`?
(Even with marking it as spawning, I'm not sure if that makes it sound for svcomp, as the autotuner will disable all threading analyses unless there's call to `pthread_create`, and I'm not sure what happens then)

sim642: > Otherwise, I think we are inflicting self-caused wounds when we benchmark, which I don't think is too smart strategically.

I'm strongly against compromising soundness for better benchmark results. That's precisely what we complain about every time we look closer at some other tools and its benchmarks. Making ourselves unsound _by default_ just opens us up to the very same criticism that we've been trying to avoid so hard. It should be an explicit and deliberate act by the user to make Goblint unsound in cases where we have a sound approach, e.g. by modifying options, adding annotations or redefining `atexit` to be identity in this case.

Moreover, if we're intentionally unsound by default, then that goes against how we expect others to benchmark Goblint. We cannot expect external users to know that when we claim to be sound, then they have to change whatever number of options to actually get soundness, instead of it being the default. External users will just use a default/example configuration and can easily point out that we're not sound and thus invalidating our claims.

Even if `atexit` spawns by default, then disabling that for our own benchmarking creates a completely unfair advantage. If we want others to benchmark us in our default configuration, then they won't be able to achieve the same results which we have achieved by going against our own principles. That's just hypocritical.

This handling of `atexit` is not a question of soundiness at all (unlike #1175, which is about behavior _after undefined behavior_). In this case we have the current spawning solution which is sound and we've used this for a decade. It's not hard in any way to be sound with `atexit`. It's harder to be more precise than spawning, but the only way to get better is to actually implement a precise solution to #799. Soundiness concerns issues that have been widely deemed to be hard in some way to solve. This isn't one of them because we have a sound solution already! We cannot claim our laziness to not implement #799 to be soundiness. Nobody's going to buy that!

> Maybe we should add these to `exitfuns` then, it seems needlessly imprecise to go to multi-threaded mode just because of a call to `atexit`?

I, nor @vesalvojdani I think, is very confident about `exitfuns` doing the right thing, especially with all the new privatizations, etc. The solution to #799 isn't `exitfuns` anyway because `atexit`/etc handlers are installed dynamically, not ahead of time.

> (Even with marking it as spawning, I'm not sure if that makes it sound for svcomp, as the autotuner will disable all threading analyses unless there's call to `pthread_create`, and I'm not sure what happens then)

That's just an oversight in the autotuner, not any inherent limitation. I'll open a separate issue about it.

michael-schwarz: Ok, seems like you feel more passionately about it than I do, and I see that your argument has some merit. I would still at the very least want an option `sem.atexit.ignore` in order to remain consistent with all our previous results (whichs tends to have `sem.unknown.spawn` off for all larger benchmarks) and allow reproducing these benchmarks, which now would be impossible to have after this benchmark.

Also, if this is a hill we are willing to die on, we should add many more examples to sv-comp that use `atexit` to highlight why it's an issue.

Re: Soundiness I disagree: To me, not handling exit handlers feels like, e.g., not handling interrupt handlers, and would seem to fall into the category of often-ignored features.

Spoken aside, we have much more egregious soundness issues than ignoring `atexit` (#1177, #1147)

michael-schwarz (approved):
LGTM! Two things I would still want:

- A test for the unsoundness due to the wrong return values of `pthread` functions you mentioned
- An option `sem.atexit.ignore` to get the behavior w.r.t. these that is the current default on master.

src/analyses/libraryFunctions.ml
sim642: > I would still at the very least want an option `sem.atexit.ignore` in order to remain consistent with all our previous results (whichs tends to have `sem.unknown.spawn` off for all larger benchmarks) and allow reproducing these benchmarks, which now would be impossible to have after this benchmark.

The behavior here is perfectly consistent with that:
* By default, the function is spawned, just like on `master`.
* With `--disable sem.unknown_function.spawn`, the function is not spawned, just like on `master`. I even added a second test demonstrating that.

So all existing configurations and benchmarking will behave exactly as they did with `atexit`. Thus, all of our old results are still reproducible and we're not inflicting any more self-caused wounds than we have so far.

Therefore, I see no reason to add an option like `sem.atexit.ignore` right now. In fact, adding it here would require us to adapt all existing confs with `sem.unknown_function.spawn` disabled to also ignore `atexit` (which wouldn't be ignored by default to remain consistent with confs that don't disable that).
If there's ever a need to separately control these, then such option can be added when we actually need it. We haven't needed it in the past nor do the Concrat benchmarks need that, so I'll consider such option out of scope of this PR.

michael-schwarz: Interesting! My understanding thus far was that the ` sem.unknown_function.*` rules only apply for those functions for which we do not have a library description, which `atexit` has after this change?

I think at least for `sem.unknown_function.invalidate.globals` that actually is the behavior.

sim642: Ideally it should be, yes, but there's a legacy reason for this:
https://github.com/goblint/analyzer/blob/70f267b1bb50c66a0e0332982506de963c6f619e/src/analyses/base.ml#L1948-L1951
Until #1029 is done, this cannot be changed. If #1029 is done and this hack is removed, then the second `atexit-no-spawn` test will start failing, so we know that we'll have to do a separate option or something. So changes to `atexit` behavior won't go unnoticed when other seemingly unrelated improvements will get made.

PR #1178, opened 2023-09-20
Data-race challenges extracted from benchmarks

sim642:


sim642:
Replaced by https://github.com/goblint/bench/pull/63.

PR #1179, opened 2023-09-20
Support stack-allocating functions when allocating heap vars

mrstanb:
During the SV-COMP benchmark runs, we noticed that calls to `alloca()` and `__builtin_alloca()` result in heap variables being allocated during analysis. However, when passing memory, allocated with these function, to `free()` no warning was produced, since Goblint couldn't differentiate between stack- and heap-allocated heap vars.

This PR:
* Updates `Queries.HeapVar`, s.t. it now takes a boolean flag indicating whether the heap variable in the query answer is allocated on the stack or the heap
* Adds `Queries.IsDynamicallyAlloced` which works like `Queries.IsHeapVar` with the difference that it also differentiates whether the heap var is one abstracting stack or heap memory

TODOs:
* [x] Track memory, allocated via `alloca()`-like functions, for each function that calls `alloca()`-like functions
* [x] Make sure that `useAfterFree` can free memory, allocated via `alloca()`-like functions, as soon as a function that called `alloca()` is returned from

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -150,8 +150,8 @@ struct
 
   let longjmp_return = ref dummyFunDec.svar
 
-  let heap_var ctx =
-    let info = match (ctx.ask Q.HeapVar) with
+  let heap_var on_stack ctx =
+    let info = match (ctx.ask (Q.HeapVar {on_stack = on_stack})) with
       | `Lifted vinfo -> vinfo
       | _ -> failwith("Ran without a malloc analysis.") in
     info
@@ -1254,7 +1254,7 @@ struct
         | Address a ->
           (* If there's a non-heap var or an offset in the lval set, we answer with bottom *)
           if AD.exists (function
-              | Addr (v,o) -> (not @@ ctx.ask (Queries.IsHeapVar v)) || o <> `NoOffset
+              | Addr (v,o) -> (not @@ ctx.ask (Queries.IsHeapVar v)) || (ctx.ask (Queries.IsHeapVar v) && not @@ ctx.ask (Queries.IsDynamicallyAlloced v)) || o <> `NoOffset
               | _ -> false) a then
             Queries.Result.bot q
           else (
@@ -1995,7 +1995,7 @@ struct
 
   let check_invalid_mem_dealloc ctx special_fn ptr =
     let has_non_heap_var = AD.exists (function
-        | Addr (v,_) -> not (ctx.ask (Q.IsHeapVar v))
+        | Addr (v,_) -> not (ctx.ask (Q.IsHeapVar v)) || (ctx.ask (Q.IsHeapVar v) && not @@ ctx.ask (Q.IsDynamicallyAlloced v))
         | _ -> false)
     in
     let has_non_zero_offset = AD.exists (function
@@ -2263,13 +2263,14 @@ struct
     | Unknown, "__goblint_assume_join" ->
       let id = List.hd args in
       Priv.thread_join ~force:true (Analyses.ask_of_ctx ctx) (priv_getg ctx.global) id st
-    | Malloc size, _ -> begin
+    | Malloc size, fname -> begin
         match lv with
         | Some lv ->
+          let is_stack_alloc = fname = "alloc" || fname = "__builtin_alloca" in
           let heap_var =
             if (get_bool "sem.malloc.fail")
-            then AD.join (AD.of_var (heap_var ctx)) AD.null_ptr
-            else AD.of_var (heap_var ctx)
+            then AD.join (AD.of_var (heap_var is_stack_alloc ctx)) AD.null_ptr
+            else AD.of_var (heap_var is_stack_alloc ctx)
           in
           (* ignore @@ printf "malloc will allocate %a bytes\n" ID.pretty (eval_int ctx.ask gs st size); *)
           set_many ~ctx (Analyses.ask_of_ctx ctx) gs st [(heap_var, TVoid [], Blob (VD.bot (), eval_int (Analyses.ask_of_ctx ctx) gs st size, true));
@@ -2279,7 +2280,7 @@ struct
     | Calloc { count = n; size }, _ ->
       begin match lv with
         | Some lv -> (* array length is set to one, as num*size is done when turning into `Calloc *)
-          let heap_var = heap_var ctx in
+          let heap_var = heap_var false ctx in
           let add_null addr =
             if get_bool "sem.malloc.fail"
             then AD.join addr AD.null_ptr (* calloc can fail and return NULL *)
@@ -2322,7 +2323,7 @@ struct
           let p_addr_get = get ask gs st p_addr' None in (* implicitly includes join of malloc value (VD.bot) *)
           let size_int = eval_int ask gs st size in
           let heap_val:value = Blob (p_addr_get, size_int, true) in (* copy old contents with new size *)
-          let heap_addr = AD.of_var (heap_var ctx) in
+          let heap_addr = AD.of_var (heap_var false ctx) in
           let heap_addr' =
             if get_bool "sem.malloc.fail" then
               AD.join heap_addr AD.null_ptr
@@ -2563,6 +2564,7 @@ struct
                 | MayBeThreadReturn
                 | PartAccess _
                 | IsHeapVar _
+                | IsDynamicallyAlloced _
                 | IsMultiple _
                 | CreatedThreads
                 | MustJoinedThreads ->
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -151,7 +151,7 @@ struct
   let longjmp_return = ref dummyFunDec.svar
 
   let heap_var on_stack ctx =
-    let info = match (ctx.ask (Q.HeapVar {on_stack})) with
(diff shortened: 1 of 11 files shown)


src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -133,17 +133,20 @@ module MallocWrapper : MCPSpec = struct
...
       in
       let count = UniqueCallCounter.find (`Lifted node) counter in
       let var = NodeVarinfoMap.to_varinfo (ctx.ask Q.CurrentThreadId, node, count) in
       var.vdecl <- UpdateCil.getLoc node; (* TODO: does this do anything bad for incremental? *)
+      if on_stack then var.vattr <- addAttribute (Attr ("stack_alloca", [])) var.vattr; (* If the call was for stack allocation, add an attr to mark the heap var *)
       `Lifted var
     | Q.IsHeapVar v ->
       NodeVarinfoMap.mem_varinfo v
+    | Q.IsDynamicallyAlloced v ->
+      NodeVarinfoMap.mem_varinfo v && not @@ hasAttribute "stack_alloca" v.vattr
```
michael-schwarz, line 149: In what sense are stack things allocated by `alloca` not dynamically allocated? I think the terminology is a bit confusing here.

src/analyses/libraryFunctions.ml
```diff
@@ -64,6 +64,7 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("__builtin_strcmp", special [__ "s1" [r]; __ "s2" [r]] @@ fun s1 s2 -> Strcmp { s1; s2; n = None; });
     ("strncmp", special [__ "s1" [r]; __ "s2" [r]; __ "n" []] @@ fun s1 s2 n -> Strcmp { s1; s2; n = Some n; });
     ("malloc", special [__ "size" []] @@ fun size -> Malloc size);
+    ("alloca", special [__ "size" []] @@ fun size -> Malloc size); (* TODO: Maybe define a new special type [Alloca], just like [Malloc]? *)
```
michael-schwarz, line 67: Yes, having a special type here makes sense!

src/analyses/wrapperFunctionAnalysis.ml
mrstanb: You're right, `alloca` is indeed dynamic allocation as well. I got `HeapVar` and `IsDynamicallyAlloced` the other way around.

mrstanb: `IsHeapVar` and `IsDynamicallyAlloced` should be now implemented the other way around. All of their usages in Goblint should be also adapted to these changes now. Cf. https://github.com/goblint/analyzer/pull/1179/commits/d9df607e99ea78600b81f879955208ec00391bcb

src/analyses/libraryFunctions.ml
mrstanb: Added `Alloca` type in https://github.com/goblint/analyzer/pull/1179/commits/aa97af86c23911bd05822149a7d9c9fc4f819ae2. I also moved `__builtin_alloca`, s.t. it's now classified only as `Alloca` and not as `Malloc` anymore: https://github.com/goblint/analyzer/pull/1179/commits/6e9bb2f121114a9c91bf1bb19557bbcfd0d3ffca

mrstanb:
I'm only wondering about one thing, namely `__builtin_alloca` is now defined
* Once as part of `c_descs_list`
* And once as part of `gcc_descs_list`

Iirc, `__builtin_*` functions are GCC-specific and thus it's probably worth it to keep `__builtin_alloca` only in `gcc_descs_list`. What do you think?

sim642:
Indeed there should be just one definition of `__builtin_alloca` and it should be under GCC. Also `alloca` itself should go there because it's not part of any C standard. According to the man page, it's not part of any standard, but is "machine- and compiler-dependent".

src/analyses/base.ml
```diff
@@ -150,8 +150,8 @@ struct
 
   let longjmp_return = ref dummyFunDec.svar
 
-  let heap_var ctx =
-    let info = match (ctx.ask Q.HeapVar) with
+  let heap_var on_stack ctx =
+    let info = match (ctx.ask (Q.HeapVar {on_stack = on_stack})) with
```
sim642, line 154:
```suggestion
    let info = match (ctx.ask (Q.HeapVar {on_stack})) with
```

src/analyses/base.ml
```diff
@@ -1249,7 +1249,7 @@ struct
           (* If there's a non-heap var or an offset in the lval set, we answer with bottom *)
           (* If we're asking for the BlobSize from the base address, then don't check for offsets => we want to avoid getting bot *)
           if AD.exists (function
-              | Addr (v,o) -> (not @@ ctx.ask (Queries.IsHeapVar v)) || (if not from_base_addr then o <> `NoOffset else false)
+              | Addr (v,o) -> (not @@ ctx.ask (Queries.IsDynamicallyAlloced v)) || (ctx.ask (Queries.IsDynamicallyAlloced v) && not @@ ctx.ask (Queries.IsHeapVar v)) || (if not from_base_addr then o <> `NoOffset else false)
```
sim642, line 1252: Should use parenthesis instead of `@@`s here because it's not obvious where the `@@` ends.
Also, the query is duplicated, so it would be nice to just use a variable for it and make this line more readable.

mrstanb:
> Indeed there should be just one definition of `__builtin_alloca` and it should be under GCC. Also `alloca` itself should go there because it's not part of any C standard. According to the man page, it's not part of any standard, but is "machine- and compiler-dependent".

Both `alloca` and `__builtin_alloca` should be now part of `gcc_descs_list` only. Cf. https://github.com/goblint/analyzer/pull/1179/commits/35513d09c2ed2a0fb6f00a1e0dd1854f9dd79329

src/domains/queries.ml
```diff
@@ -101,8 +101,9 @@ type _ t =
   | IterVars: itervar -> Unit.t t
   | PathQuery: int * 'a t -> 'a t (** Query only one path under witness lifter. *)
   | DYojson: FlatYojson.t t (** Get local state Yojson of one path under [PathQuery]. *)
-  | HeapVar: VI.t t
+  | HeapVar: {on_stack: bool} -> VI.t t (* If on_stack is [true], then alloca() or a similar function was called *)
   | IsHeapVar: varinfo -> MayBool.t t (* TODO: is may or must? *)
+  | IsDynamicallyAlloced: varinfo -> MayBool.t t (* [true] if heap var represents dynamically alloced memory *)
```
sim642, line 106: The naming is very confusing. If I understand correctly, `alloca` also uses `HeapVar`. But `alloca` doesn't allocate on the heap, it allocates on the stack!
And the difference between `IsHeapVar` and `IsDynamicallyAlloced` is completely unclear to me. Both `malloc` and `alloca` variables are created by `HeapVar`, so it seems like `IsHeapVar` is true for both. But both are also dynamically allocated.

Since these "heap" variables get `(alloc@...)` as their name, maybe their creation query should be `AllocVar` and the query corresponding to it `IsAllocVar`. Then `IsHeapVar` would distinguish those by `malloc` from those by `alloca`, or something.

mrstanb: I agree that the naming might be confusing. `HeapVar` should be now referring to any dynamic allocation-site abstraction (i.e., `(alloca@...)`), which can in particular be `alloca` (on the stack) and `malloc` (on the heap).
Hence, I agree with the `AllocVar` name in place of `HeapVar`.

`IsDynamicallyAlloced` essentially checks if a var refers to a dynamic allocation-site abstraction, hence we could change it to `IsAllocVar`. `IsHeapVar` should be already be able to distinguish between `alloc@...` vars from `alloca` and from `malloc`. Iirc, I added an extra `vattr` to the `alloc@...` vars for this which is implemented in `HeapVar`.

In short, my suggestion:
* Rename `HeapVar` -> `AllocVar`
* Rename `IsDynamicallyAlloced` -> `IsAllocVar`
* Keep `IsHeapVar` as is
* (not sure if it'd be needed, but we could also add `IsStackVar` which is dual to `IsHeapVar`)

Please let me know, what do you think. CC @michael-schwarz, because we also discussed the naming with him before.

michael-schwarz: Yes, seems reasonable!

src/analyses/base.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1179/commits/3bdc92d7c6e58520f6ed552cb687fa158adcf429

src/domains/queries.ml
mrstanb: I did the renaming in https://github.com/goblint/analyzer/pull/1179/commits/7b1d29850036fb1287b10c8aeb962a162b7fac11

michael-schwarz (approved):

michael-schwarz:
Sorry for not waiting on your review @sim642, but things are starting to get urgent in order for @mrstanb's changes being all set and integrated with the rest in time for us to still run benchmarks without opening another staging branch and the hassle that creates.
If it turns out later that there's anything that should still be fixed here Stanimir (or I) will address it on master / in a follow-up PR.

PR #1180, opened 2023-09-20
First steps towards a Proof-of-Concept for analyzing threadpools

michael-schwarz:
Since we constantly talk about it (today again in Munich), and I started thinking about it on the subway, I spent a few minutes implementing a Proof-of-Concept of the first part of an analysis:

In it's current state it can determine that `pthread_join` has been called for all members of an array, by lugging around a boolean and employing the partitioned array domain.

The other piece that is missing is an analysis that shows that all created threads of a given abstract thread id is indeed contained in the array.

After we have this, we would be able to handle thread pools of constant size.

This is not meant to be merged, but as a simple Proof-of-Concept that anyone is free to adopt and develop into a proper analysis.


jerhard:
Close this for now. Future investigations on handling of thread pools may build up on this.

michael-schwarz:
Since we discussed this again at the Gobcon, I re-open this now.

PR #1187, opened 2023-09-24
Spawn threads created from unknown functions as non-unique

### LINKED ISSUE #843: Non-unique thread spawns
In some special cases it's desirable to spawn threads as non-unique right off the bat:
1. Unknown functions which spawn create unique threads: https://github.com/goblint/analyzer/pull/837#discussion_r984570708. This is unsound and we should instead create the new thread immediately as non-unique.
2. Klever stubs have a non-unique thread spawning function: https://github.com/goblint/analyzer/pull/688#discussion_r851624003.

An extra uniqueness argument needs to be added and passed from `ctx.spawn` to the threadid analysis through the numerous transfer functions involved.

michael-schwarz:
Closes #843.


src/framework/analyses.ml
```diff
@@ -444,7 +444,7 @@ sig
   val paths_as_set : (D.t, G.t, C.t, V.t) ctx -> D.t list
 
   (** Returns initial state for created thread. *)
-  val threadenter : (D.t, G.t, C.t, V.t) ctx -> lval option -> varinfo -> exp list -> D.t list
+  val threadenter : ?multiple:bool -> (D.t, G.t, C.t, V.t) ctx -> lval option -> varinfo -> exp list -> D.t list
```
sim642, line 447: The argument being optional here means that the default value has to be duplicated in all analyses, which is inconvenient and error-prone if someone doesn't. This could just be a named argument.

`ctx.spawn` can still have an optional argument. Using `ctx.spawn` is as convenient as before but the default value is just concentrated to the definition of `ctx.spawn` itself.

src/analyses/threadId.ml
```diff
@@ -56,9 +56,9 @@ struct
...
     (N.bot (), `Lifted (tid), (TD.bot (), TD.bot ()))
 
-  let create_tid (_, current, (td, _)) ((node, index): Node.t * int option) v =
+  let create_tid ?(multiple=false) (_, current, (td, _)) ((node, index): Node.t * int option) v =
     match current with
-    | `Lifted current ->
+    | `Lifted current when not multiple ->
       let+ tid = Thread.threadenter (current, td) node index v in
       if GobConfig.get_bool "dbg.print_tids" then
         Hashtbl.replace !tids tid ();
```
sim642, line 64: I don't think this is right. Now if `multiple = true`, it falls into the second case where it uses `Thread.threadinit` instead of `Thread.threadenter` (and doesn't register the existence of such thread ID for debugging).
The problem is that `threadinit` creates a new history thread ID without parent. It should still use `threadenter`, but that should handle the non-uniqueness with the set.

Not having the right parent possibly breaks some MHP thing.

michael-schwarz: The idea here was to treat these sorts of threads as created by the system (same as if `main` is not unique), as the moment we are spawning some of these via an unknown function, we are pretty much lost anyway.

I don't think it makes sense to have multiple instances of these thread around that differ in their creation history, given that we sort of lost all meaningful precision about threads anyway.

sim642: > The idea here was to treat these sorts of threads as created by the system (same as if `main` is not unique)

That was not the idea of #843 at all because it doesn't work for either motivating case. The requirement there is to do the spawning like we've done so far, but instead of `[main]` spawning `[main; t_fun]` it spawns `[main]-{t_fun}`.

The intended semantics is `while (nondet()) pthread_create(...)`. The history is very well-defined there.

Treating them as parentless if very different from what we've done so far. It's like our Linux kernel auto-harnessing with multiple `mainfuns`, which is wildly unsound. So changing our treatment of unknown spawns to this would make us a lot more unsound on real-world things.

> I don't think it makes sense to have multiple instances of these thread around that differ in their creation history, given that we sort of lost all meaningful precision about threads anyway.

There's no multiple instances with different histories. There's one instance (in the Goblint sense) that simply doesn't return true for `MustBeUniqueThread`, etc. There's no loss of all meaningful precision, it's just one bit of information that's different about the thread. Creating threads in a loop is the normal thing, not losing all precision.

michael-schwarz: > There's no multiple instances with different histories.

If `fun` is spawned from an unknown function by thread `main{}` and by thread `main:f1{}` and by thread `main:f2{}`, I would end up with three instances of it   `main{fun}` `main:f1{fun}`  `main:f2{fun}` (which are all non-unique).

sim642: That's just a matter of efficiency, but efficiency can't come at the expense of soundness. AFAIK the new privatizations (at least the more precise ones) haven't been designed nor validated with multiple initial threads. Also, I think the MHP logic falls apart if new threads appear out of thin air with no history. Some prefix-based checks could make unsound claims.

For example, here with multiple initial threads we unsoundly claim both assertions hold:
```c
// PARAM: --set mainfun '["t1", "t2"]' --set ana.base.privatization mutex-meet-tid --set ana.path_sens[+] threadflag
#include<pthread.h>

int myglobal;
pthread_mutex_t A_mutex = PTHREAD_MUTEX_INITIALIZER;

void t1() {
  pthread_mutex_lock(&A_mutex);
  int g = myglobal;
  __goblint_check(g == 0); // UNKNOWN!
  myglobal = 1; //RACE!
  pthread_mutex_lock(&A_mutex);
}

void t2() {
  pthread_mutex_lock(&A_mutex);
  int g = myglobal;
  __goblint_check(g == 0); // UNKNOWN!
  myglobal = 2; //RACE!
  pthread_mutex_lock(&A_mutex);
}
```

This issue and PR should just focus on the non-uniqueness aspect, not simultaneously try to optimize things as well. Changing two things at once together is a slippery slope. Unless we immediately want to validate that all of the mainfuns/otherfuns/exitfuns/allfuns things work with all the privatizations and MHP, we shouldn't try to simultaneously optimize.

michael-schwarz: Doing it like this also requires adding a `multiple` argument to `threadspawn`, which I have now done.

michael-schwarz:
@sim642: Any more comments or can we merge this?

sim642:
I intend to have another look at some point but this isn't a priority right now.

michael-schwarz:
Ok, then I'll move ahead and merge :)

PR #1191, opened 2023-09-26
Add final messages about unsound results

sim642:


michael-schwarz (approved):
Seems reasonable, after addressing the new TODOs this introduces to the code, I think we're good to merge this.

PR #1197, opened 2023-09-29
Add out-of-bounds check for `memset` and `memcpy`

mrstanb:
This PR adds a check for the `count` and `n` parameter of `memset` and `memcpy`
* `base` should now be able to warn in case `count` or `n` is bigger than the smallest memory, pointed-to by `dest`
* Added a regession test case for this as well


src/analyses/base.ml
```diff
@@ -2026,6 +2026,80 @@ struct
...
+    | true, _ ->
+      AnalysisState.svcomp_may_invalid_deref := true;
+      M.warn ~category:(Behavior behavior) ~tags:[CWE cwe_number] "Size of dest %a in function %s is unknown. Memory out-of-bounds access might occur" d_exp dest fun_name
+    | _, true ->
+      AnalysisState.svcomp_may_invalid_deref := true;
+      M.warn ~category:(Behavior behavior) ~tags:[CWE cwe_number] "Count parameter, passed to function %s is unknown. Memory out-of-bounds access might occur" fun_name
+    | false, false ->
+      if dest_size < eval_n then begin
+        AnalysisState.svcomp_may_invalid_deref := true;
+        M.warn ~category:(Behavior behavior) ~tags:[CWE cwe_number] "Size of dest in function %s is %a (in bytes). Count is %a (in bytes). Memory out-of-bounds access must occur" fun_name ValueDomainQueries.ID.pretty dest_size ValueDomainQueries.ID.pretty eval_n
```
michael-schwarz, line 2099: Is it really must occur here, or may occur?

src/analyses/base.ml
```diff
@@ -2026,6 +2026,80 @@ struct
...
+    let eval_n = ctx.ask (Queries.EvalInt n) in
+    match ValueDomainQueries.ID.is_top dest_size, ValueDomainQueries.ID.is_top eval_n with
+    | true, _ ->
+      AnalysisState.svcomp_may_invalid_deref := true;
+      M.warn ~category:(Behavior behavior) ~tags:[CWE cwe_number] "Size of dest %a in function %s is unknown. Memory out-of-bounds access might occur" d_exp dest fun_name
+    | _, true ->
+      AnalysisState.svcomp_may_invalid_deref := true;
+      M.warn ~category:(Behavior behavior) ~tags:[CWE cwe_number] "Count parameter, passed to function %s is unknown. Memory out-of-bounds access might occur" fun_name
+    | false, false ->
+      if dest_size < eval_n then begin
```
michael-schwarz, line 2097: Is this a polymorphic comparison of things of type `ID.t` here? Then it's not what we want: We want to check if it may be smaller here, which may e.g. be done by `ID.lt`.

src/analyses/base.ml
```diff
@@ -2026,6 +2026,80 @@ struct
...
+        in
+        (* Map each points-to-set element to its size *)
+        let pts_sizes = List.map pts_elems_to_sizes pts_list in
+        (* Take the smallest of all sizes that ptr's contents may have *)
+        begin match pts_sizes with
+          | [] -> `Bot
+          | [x] -> x
+          | x::xs -> List.fold_left (fun acc elem ->
+              if ValueDomainQueries.ID.compare acc elem >= 0 then elem else acc
+            ) x xs
```
michael-schwarz, line 2077: I don't think this is correct here, comparison of ID.t values does not need to coincide with any sort of order. The better thing here would be `ID.join`.

src/analyses/base.ml
```diff
@@ -2026,6 +2026,80 @@ struct
...
+    let points_to_heap_only =
+      match ctx.ask (Queries.MayPointTo dest) with
+      | a when not (AD.is_top a) ->
+        AD.for_all (function
+            | Addr (v, _) -> ctx.ask (Queries.IsHeapVar v)
+            | _ -> false
+          ) a
+      | _ -> false
+    in
+    if points_to_heap_only then
```
michael-schwarz, line 2043: Why do I not need to use this for all of those things that are on the heap but only if all destinations are on the heap?

src/analyses/base.ml
```diff
@@ -2026,6 +2026,80 @@ struct
...
+      (* Ask for BlobSize from the base address (the second field set to true) in order to avoid BlobSize giving us bot *)
+      ctx.ask (Queries.BlobSize {exp = dest; base_address = true})
```
michael-schwarz, line 2045: Does this not ignore offsets in `dest`? If I have a pointer to the end of something that is `80` bytes large, I cannot copy `80` bytes into it...

tests/regression/77-mem-oob/06-memset-memcpy-oob.c
```diff
@@ -0,0 +1,21 @@
+// PARAM: --set ana.activated[+] memOutOfBounds --enable ana.int.interval --disable warn.info
+// TODO: The "--disable warn.info" part is a temporary fix and needs to be removed once the MacOS CI job is fixed
+#include <stdlib.h>
+#include <string.h>
+
+int main(int argc, char const *argv[]) {
```
michael-schwarz, line 6: It maybe be a good idea to also play around with sizes that are not constant here, that might have exposed some of the issues above.

michael-schwarz (changes requested):
I think there's still quite a few kinks to be ironed out, just let me know if any of my comments are cryptic or unclear!

src/analyses/base.ml
mrstanb: It's a may occur, true. Should be fixed now.

src/analyses/base.ml
mrstanb: Thanks for the tip! Should be fixed in https://github.com/goblint/analyzer/pull/1197/commits/fae72563a48b8b7a344395e8f6d75d76630bc8c9

src/analyses/base.ml
mrstanb: Changed the `fold_left` to use `join` accumulating the sizes together in https://github.com/goblint/analyzer/pull/1197/commits/655362e70a7d0c2a42c4dba910e5c67de1aee1c9.

src/analyses/base.ml
mrstanb: Because `BlobSize` currently answers with `bot` if the points-to set does not contain heap variables only. Hence if the points-to set is mixed in the sense that it contains both heap and non-heap memory, we would get `bot` here. This is why we have this check here.

Alternatively, I guess we could skip the check and just join the result of `BlobSize` with the other sizes that we collect for the points-to set elements (i.e., the sizes, calculated in the `else` expression)?

src/analyses/base.ml
mrstanb: True, yes. I'll make sure to set `base_address` to `false`.

mrstanb: Fixed in https://github.com/goblint/analyzer/pull/1197/commits/2534945eec78d99658cc2bf5b420af35580b7fe3

tests/regression/77-mem-oob/06-memset-memcpy-oob.c
mrstanb: Absolutely agree. I added some further cases in https://github.com/goblint/analyzer/pull/1197/commits/87dadd265c5a8af8f8087307ef31aa41de51ff56

src/analyses/base.ml
michael-schwarz: And maybe a second question: Why are you not working with the `to_index` of the offset in the Blob case here? I think something similar to what you did in the other PR should be possible, where for Blobs you get the length of the base address, add the numerical offset, and then check whether `offset + n <= size`?

src/analyses/base.ml
```diff
@@ -2026,6 +2026,100 @@ struct
...
+      let dest_size_lt_count =
+        begin
+          try ID.lt casted_ds casted_en
+          with IntDomain.ArithmeticOnIntegerBot _ -> ID.bot_of @@ Cilfacade.ptrdiff_ikind ()
+        end
+      in
+      begin match ID.to_bool dest_size_lt_count with
+        | Some true ->
+          AnalysisState.svcomp_may_invalid_deref := true;
+          M.warn ~category:(Behavior behavior) ~tags:[CWE cwe_number] "Size of dest in function %s is %a (in bytes). Count is %a (in bytes). Memory out-of-bounds access may occur" fun_name ValueDomainQueries.ID.pretty dest_size ValueDomainQueries.ID.pretty eval_n
```
michael-schwarz, line 2115:
```suggestion
          M.warn ~category:(Behavior behavior) ~tags:[CWE cwe_number] "Size of dest in function %s is %a (in bytes). Count is %a (in bytes). Memory out-of-bounds access must occur" fun_name ValueDomainQueries.ID.pretty dest_size ValueDomainQueries.ID.pretty eval_n
```

In the `Some true` case, we can now say it's a must now!

src/analyses/base.ml
```diff
@@ -2026,6 +2026,100 @@ struct
...
+    | _, `Bot ->
+      M.warn ~category:(Behavior behavior) ~tags:[CWE cwe_number] "Count parameter, passed to function %s is bottom" fun_name
+    | `Lifted ds, `Lifted en ->
+      let casted_ds = ID.cast_to (Cilfacade.ptrdiff_ikind ()) ds in
+      let casted_en = ID.cast_to (Cilfacade.ptrdiff_ikind ()) en in
+      let dest_size_lt_count =
+        begin
+          try ID.lt casted_ds casted_en
+          with IntDomain.ArithmeticOnIntegerBot _ -> ID.bot_of @@ Cilfacade.ptrdiff_ikind ()
+        end
```
michael-schwarz, line 2110: Preferably we should catch `bot` earlier than here (and closer to where they arise). Probably immediately after the BlobSize query?

mrstanb: Should be changed to now catch `bot` earlier for all `lt` call-sites.

src/analyses/base.ml
mrstanb: Thanks for the tip! Should be changed in all relevant spots now.

src/analyses/base.ml
mrstanb: After the refactoring in the latest few commits, this should be fixed and we should be now able to account for offsets as well.

mrstanb:
Hopefully all remarks are now fixed. I also migrated the `memset`/`memcpy` checks completely from `base` to `memOutOfBounds` for consistency's sake. Feel free to review :)

src/analyses/base.ml
```diff
@@ -2093,7 +2094,6 @@ struct
     in
     let st = match desc.special args, f.vname with
     | Memset { dest; ch; count; }, _ ->
-      (* TODO: check count *)
```
michael-schwarz, line 2096: Might make sense to keep the `check_count` TODO, I am not sure the `set` here does the right thing, even if the size is within bounds.

michael-schwarz (approved):
Should be good to go after adding back the one TODO I commented on.

src/analyses/base.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1197/commits/89abb1a9ade5352f19fc2c042c4ef2b387c5fa38

PR #1201, opened 2023-10-02
SV-COMP "Memory Safety" benchmark additions

mrstanb:
This PR:
* Builds upon changes from the https://github.com/goblint/analyzer/tree/staging_memsafety branch that was used for the first benchmark runs of Goblint against the SV-COMP "Memory Safety" benchmarks
* Adds (first and hacky) support for recognizing all three memory safety properties of SV-COMP (i.e., `valid-free`, `valid-deref`, `valid-memtrack`)
* Adds support for the `valid-memcleanup` property of SV-COMP

TODO:
 - [x] Try out the changes from https://github.com/goblint/cil/pull/155 for a further benchmark run as well


src/autoTune.ml
```diff
@@ -226,6 +226,8 @@ let focusOnSpecification () =
   (* TODO: Finish these two below later *)
   | ValidDeref
   | ValidMemtrack -> ()
+  | MemorySafety -> () (* TODO: This is here for now just to complete the pattern match *)
+  | ValidMemcleanup -> ()
```
sim642, line 230: Should probably implement these as well to enable the required analyses because they're not default in the SV-COMP conf (and shouldn't be for other categories).

tests/regression/78-use_after_free/01-simple-uaf.c
sim642, line 1: Why is `74-use_after_free` being renamed to `78-use_after_free` here? It's already 74 on master, so there's no conflict.

src/witness/svcompSpec.ml
```diff
@@ -9,10 +9,13 @@ type t =
   | ValidFree
   | ValidDeref
   | ValidMemtrack
+  | MemorySafety (* Internal property for use in Goblint; serves as a summary for ValidFree, ValidDeref and ValidMemtrack *)
```
sim642, line 12: This is fine for now. I think I promised somewhere to rework this myself anyway.

src/autoTune.ml
mrstanb: The memory safety analyses should be now enabled in the auto-tuner: https://github.com/goblint/analyzer/pull/1201/commits/00cc9b541ae6e6270906c16cbba94498692ab6f3

tests/regression/78-use_after_free/01-simple-uaf.c
mrstanb: Sorry, this probably occured during the merge with `master` and went unnoticed. Should be fixed now, cf. https://github.com/goblint/analyzer/pull/1201/commits/2c8e927f4e5f88b0a4eb003da6fd24ac55a0e004

src/autoTune.ml
```diff
@@ -223,9 +223,18 @@ let focusOnSpecification () =
...
-  | ValidDeref
-  | ValidMemtrack -> ()
+  | ValidDeref -> (* Enable the memOutOfBounds analysis *)
+    let memOobAna = ["memOutOfBounds"] in
+    print_endline @@ "Specification: ValidDeref -> enabling memOutOfBounds analysis \"" ^ (String.concat ", " memOobAna) ^ "\"";
+    enableAnalyses memOobAna
+  | ValidMemtrack
+  | ValidMemcleanup -> (* Enable the memLeak analysis *)
+    let memLeakAna = ["memLeak"] in
+    print_endline @@ "Specification: ValidDeref and ValidMemcleanup -> enabling memLeak analysis \"" ^ (String.concat ", " memLeakAna) ^ "\"";
```
michael-schwarz, line 233:
```suggestion
    print_endline @@ "Specification: ValidMemtrack and ValidMemcleanup -> enabling memLeak analysis \"" ^ (String.concat ", " memLeakAna) ^ "\"";
```

src/autoTune.ml
```diff
@@ -223,9 +223,18 @@ let focusOnSpecification () =
...
-  (* TODO: Finish these two below later *)
-  | ValidDeref
-  | ValidMemtrack -> ()
+  | ValidDeref -> (* Enable the memOutOfBounds analysis *)
+    let memOobAna = ["memOutOfBounds"] in
+    print_endline @@ "Specification: ValidDeref -> enabling memOutOfBounds analysis \"" ^ (String.concat ", " memOobAna) ^ "\"";
+    enableAnalyses memOobAna
+  | ValidMemtrack
+  | ValidMemcleanup -> (* Enable the memLeak analysis *)
+    let memLeakAna = ["memLeak"] in
```
michael-schwarz, line 232: Maybe we should also set `ana.malloc.unique_address_count` to `1` here, otherwise this analysis won't work properly iirc?

src/framework/analysisState.ml
```diff
@@ -16,6 +16,9 @@ let svcomp_may_invalid_deref = ref false
 (** Whether an invalid memtrack happened *)
 let svcomp_may_invalid_memtrack = ref false
 
+(** Whether an invalid memcleanup happened *)
+let svcomp_may_invalid_memcleanup = ref false
```
michael-schwarz, line 20: The comments are a bit weird: `an invalid memtrack happened` is probably something like `a memleak occurred` and similar for `invalid_memcleanup`.

src/witness/witness.ml
```diff
@@ -505,14 +505,14 @@ struct
         in
         (module TaskResult:WitnessTaskResult)
       )
-    | ValidFree ->
-      let module TrivialArg =
+    | ValidFree (*->*)
```
michael-schwarz, line 508: Before merging this, we'll need to cleanup the commented out code.

src/autoTune.ml
mrstanb: This probably remained due to my copy pasta. Thanks!

src/autoTune.ml
mrstanb: Yes, absolutely! Added this in https://github.com/goblint/analyzer/pull/1201/commits/edaef4230d6db75def47f41b4a748c9cbd5aa83d

src/framework/analysisState.ml
mrstanb: Fixed in https://github.com/goblint/analyzer/pull/1201/commits/7f0b43ccf1db2d6c3bec95ce1ea759dff84e4fba

src/witness/witness.ml
mrstanb: Should be cleaned up now (https://github.com/goblint/analyzer/pull/1201/commits/a975702f700511c3c1f0b71979502462192d9bc2)

src/autoTune.ml
```diff
@@ -230,6 +230,8 @@ let focusOnSpecification () =
   | ValidMemtrack
   | ValidMemcleanup -> (* Enable the memLeak analysis *)
     let memLeakAna = ["memLeak"] in
+    print_endline "Setting \"ana.malloc.unique_address_count\" to 1";
+    set_int "ana.malloc.unique_address_count" 1;
```
michael-schwarz, line 234: Now that I think of it, this should probably only set this value to 1 if it is not set to a larger value before?

mrstanb: Yes, it probably makes sense to change it only if it's < 1. I adapted the `set_int` call in https://github.com/goblint/analyzer/pull/1201/commits/b26010ac042143b4cdeb806b2a35b67054db1d76

mrstanb:
I also added a `valid-memcleanup.prp` file similar to `valid-memsafety.prp`, since we now also consider `valid-memcleanup`. Cf. https://github.com/goblint/analyzer/pull/1201/commits/89d68af6d7c68a647e13dd77e876580dbc8e71b1

michael-schwarz:
Without the scoping check, the results seem quite encouraging:

[results_ignore_scopes.zip](https://github.com/goblint/analyzer/files/12791930/results_ignore_scopes.zip)

```
Statistics:           3862 Files
  correct:            1238
    correct true:     1238
    correct false:       0
  incorrect:            27
    incorrect true:     27
    incorrect false:     0
  unknown:            2597
  Score:              1612 (max: 5857)
```

- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_struct_68_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_struct_67_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_struct_45_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_long_68_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_long_67_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_long_45_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_int_68_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_int_67_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_int_45_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_int64_t_68_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_int64_t_67_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_int64_t_45_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_char_68_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_char_67_bad.yml
- [ ] Juliet_Test/CWE415_Double_Free---s01---CWE415_Double_Free__malloc_free_char_45_bad.yml
- [ ] memsafety-ext3/scopes5.yml
- [ ] memsafety-ext3/scopes4-2.yml
- [ ] memsafety-ext3/scopes3.yml
- [ ] memsafety-ext3/scopes1.yml
- [ ] memsafety-ext3/getNumbers4-2.yml
- [ ] memsafety-ext3/getNumbers2.yml
- [ ] memsafety-ext3/derefInLoop1.yml
- [ ] ldv-memsafety-bitfields/test-bitfields-2.1-2.yml
- [ ] ldv-memsafety-bitfields/test-bitfields-2-1.yml
- [ ] ldv-memsafety-bitfields/test-bitfields-1-2.yml
- [ ] list-ext-properties/test-0158_1-3.yml
- [ ] termination-crafted/LexIndexValue-Pointer-3.yml

michael-schwarz:
@mrstanb when you add the scope stuff, I think it would make sense to add an option to switch it off, to make it easy to experiment.

michael-schwarz:
We looked into some of these CWE ones, and it seems there's a bug in `enter` for the `userAfterFree` analysis that causes it to not consider values reachable via global variables.

michael-schwarz:
```
Statistics:           3862 Files
  correct:            1233
    correct true:     1233
    correct false:       0
  incorrect:            12
    incorrect true:     12
    incorrect false:     0
  unknown:            2617
  Score:              2082
```

[no-scope-no-filter.zip](https://github.com/goblint/analyzer/files/12793284/no-scope-no-filter.zip)

Looks like it only cost us 5 correct verdicts to fix this the naive way, so we should probably go with that for now.

mrstanb:
So, I found a small typo of mine which caused us to be unsound for the `termination-crafted/LexIndexValue-Pointer-3` case.

I fixed the typo and added this case from SV-COMP as a regression test case. I'll mark this case as done as well

michael-schwarz:
After running this on the server, it seems that it does not fix `termination-crafted/LexIndexValue-Pointer-3`.

tests/regression/77-mem-oob/10-oob-two-loops.c
```diff
@@ -0,0 +1,20 @@
...
+		p[i] = __VERIFIER_nondet_int(); //NOWARN
+	}
+
+	int *q = p;
+
+	while (*q >= 0 && q < p + 1048 * sizeof(int)) { //WARN
+		if (__VERIFIER_nondet_int()) {
+			q++;
+		} else {
+			(*q)--; //WARN
```
michael-schwarz, line 15: Can you explain why this is an invalid deref?

mrstanb: My assumption here is that the condition `q < p + 1048 * sizeof(int)` could lead to an overflow, but I might be wrong. My line of thought is that the pointer arithmetic with `p` should be only `p + 1048` in order to be correct.

michael-schwarz: Ah, I see!

mrstanb: I think that the `q++` part could keep incrementing beyond the bounds of `q`'s contents, which could lead to an invalid-deref in `(*q)--`

michael-schwarz:
This is a minimal counterexample extracted from LexIndexValue, there seems to be some connection with `--set sem.int.signed_overflow assume_none`

~~~C
// PARAM: --set ana.activated[+] memOutOfBounds --enable ana.int.interval --disable warn.info --set sem.int.signed_overflow assume_none
int main() {
    int *p = malloc(2 * sizeof(int));
    int *q = p;
    int x;

    if (x) {
        q++;
        q++;
        q++;
        x = *q;  //We warn here
    }

    x = *q;  // But not here :(
    return 0;
}
~~~

src/autoTune.ml
```diff
@@ -223,9 +223,24 @@ let focusOnSpecification () =
...
     enableAnalyses uafAna
-  (* TODO: Finish these two below later *)
-  | ValidDeref
-  | ValidMemtrack -> ()
+  | ValidDeref -> (* Enable the memOutOfBounds analysis *)
+    let memOobAna = ["memOutOfBounds"] in
+    print_endline "Setting \"cil.addNestedScopeAttr\" to true";
+    set_bool "cil.addNestedScopeAttr" true;
+    print_endline @@ "Specification: ValidDeref -> enabling memOutOfBounds analysis \"" ^ (String.concat ", " memOobAna) ^ "\"";
+    enableAnalyses memOobAna
```
michael-schwarz, line 231: This needs to be done for mem-safety too!

mrstanb: I think we should also add the `ana.malloc.unique_address_count` option to `1` for `MemorySafety`, right?

michael-schwarz:
For c9a846b, the behavior stayed the same as before. It seems something with the scope detection is not working properly. Might make sense to check this locally, while I run 44476ce.

```
Statistics:           3862 Files
  correct:            1233
    correct true:     1233
    correct false:       0
  incorrect:            12
    incorrect true:     12
    incorrect false:     0
  unknown:            2617
  Score:              2082 (max: 5857)
```

michael-schwarz:
With 44476ce it is

```
Statistics:           3862 Files
  correct:            1232
    correct true:     1232
    correct false:       0
  incorrect:            11
    incorrect true:     11
    incorrect false:     0
  unknown:            2619
  Score:              2112 (max: 5857)
```

michael-schwarz:
The statistics with 1211a9f are quite disappointing:

```
Statistics:           3862 Files
  correct:             130
    correct true:      130
    correct false:       0
  incorrect:             5
    incorrect true:      5
    incorrect false:     0
  unknown:            3727
  Score:               100 (max: 5857)
```

michael-schwarz:
These are the results. The run at 10:30am today is without the scope logic for comparison.

[with-scope.zip](https://github.com/goblint/analyzer/files/12802473/with-scope.zip)

michael-schwarz:
Doesn't look like 1eb7309 is it either, back to:

```
Statistics:           3862 Files
  correct:            1232
    correct true:     1232
    correct false:       0
  incorrect:            11
    incorrect true:     11
    incorrect false:     0
  unknown:            2619
  Score:              2112 (max: 5857)
```

src/goblint.ml
```diff
@@ -56,6 +56,8 @@ let main () =
         else
           None
       in
+      if AutoTune.isActivated "specification" && get_string "ana.specification" <> "" then
+        AutoTune.focusOnMemSafetySpecification ();
```
michael-schwarz, line 60: This is too late! As soon as we have a file, CIL has already run!

michael-schwarz:
After making `cil.addNestedScopeAttr` true unconditionally, the results are as follows:

```
Statistics:           3862 Files
  correct:            1231
    correct true:     1231
    correct false:       0
  incorrect:             7
    incorrect true:      7
    incorrect false:     0
  unknown:            2624
  Score:              2238 (max: 5857)
```

@mrstanb: You should investigate the remaining 7 unsoundnesses and supply an actually working patch for making `cil.addNestedScopeAttr` conditional again.

[with-scope-unconditional.zip](https://github.com/goblint/analyzer/files/12804748/with-scope-unconditional.zip)

michael-schwarz:
Remaining unsound tasks:

- [ ] memsafety-ext3/scopes4-2.yml
- [ ] memsafety-ext3/getNumbers4-2.yml
- [ ] memsafety-ext3/getNumbers2.yml
- [ ] ldv-memsafety-bitfields/test-bitfields-2.1-2.yml
- [ ] ldv-memsafety-bitfields/test-bitfields-2-1.yml
- [ ] ldv-memsafety-bitfields/test-bitfields-1-2.yml
- [ ] list-ext-properties/test-0158_1-3.yml

mrstanb:
> After making `cil.addNestedScopeAttr` true unconditionally, the results are as follows:
> 
> ```
> Statistics:           3862 Files
>   correct:            1231
>     correct true:     1231
>     correct false:       0
>   incorrect:             7
>     incorrect true:      7
>     incorrect false:     0
>   unknown:            2624
>   Score:              2238 (max: 5857)
> ```
> 
> @mrstanb: You should investigate the remaining 7 unsoundnesses and supply an actually working patch for making `cil.addNestedScopeAttr` conditional again.
> 
> [with-scope-unconditional.zip](https://github.com/goblint/analyzer/files/12804748/with-scope-unconditional.zip)

Thanks! I think I found a solution and I'll commit it later once I'm on my laptop again.

src/goblint.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1201/commits/4a4b6abc6edd18d1487084131a4f3bc038fe526f

mrstanb:
`list-ext-properties/test-0158_1-3.yml` is an invalid free caused by passing an invalid address to `free`. Should be resolved and detected soundly after https://github.com/goblint/analyzer/pull/1201/commits/395c30db266591bc2e20b663274628278bb3b6f0.

I also added the file itself as the `75/10` regr. test case. Seems to run fine.

mrstanb:
The cases `memsafety-ext3/scopes4-2.yml`, `memsafety-ext3/getNumbers4-2.yml`, `memsafety-ext3/getNumbers2.yml` are invalid-derefs due to pointer being derefed containing non-local and non-global vars in its points-to set.

All 3 should be now handled soundly in https://github.com/goblint/analyzer/pull/1201/commits/5cb10f61925e55932475e859569d496a630fee6a

I also added the 3 of them as regr. tests (cf. https://github.com/goblint/analyzer/pull/1201/commits/ea4410d8d0939d93f5effa56fd51c9549c717641)

---

As a sidenote: we might want to also warn for [CWE 5620](https://cwe.mitre.org/data/definitions/562.html) whenever a stack address is returned from a function. It's not strictly a direct memory safety issue, since there's no deref occurring, but it's still a bad coding practice which we could detect.

mrstanb:
` ldv-memsafety-bitfields/test-bitfields-2-1.yml` is a `memcpy` OOB access, but on `src` rather than on `dest`. Should be fixed by https://github.com/goblint/analyzer/pull/1201/commits/6745d799de882a056df9dd0c695668592a5f31eb and is also added as a regr. test case in https://github.com/goblint/analyzer/pull/1201/commits/136bec0232107c61a42a3ed369a03aee90fbf2b8

michael-schwarz:
With 3a2fe3f, we get

```
Statistics:           3862 Files
  correct:            1220
    correct true:     1220
    correct false:       0
  incorrect:             6
    incorrect true:      6
    incorrect false:     0
  unknown:            2636
  Score:              2248 (max: 5857)
````

with the following unsound results:

- [ ] ldv-memsafety-bitfields/test-bitfields-1-2.yml
- [ ] ldv-memsafety-bitfields/test-bitfields-2.1-2.yml
- [ ] memsafety-ext3/derefInLoop1.yml
- [ ] memsafety-ext3/scopes1.yml
- [ ] memsafety-ext3/scopes3.yml
- [ ] memsafety-ext3/scopes5.yml

Given most were already solved previously, I suspect some of them once more are an instance of the flag not being set correctly.

**Comment by michael-schwarz** [comment]
As suspected, when hardcoding it, the number of unsound results goes down to 2:

```
Statistics:           3862 Files
  correct:            1219
    correct true:     1219
    correct false:       0
  incorrect:             2
    incorrect true:      2
    incorrect false:     0
  unknown:            2641
  Score:              2374 (max: 5857)
```

**Comment by mrstanb** [comment]
` ldv-memsafety-bitfields/test-bitfields-1-2.yml` should be now fixed in https://github.com/goblint/analyzer/pull/1201/commits/753b5c1661009bbfa6542fd64ca3d6de62231a34 and added as test case in https://github.com/goblint/analyzer/pull/1201/commits/d68328163e07f228f44f50c0b9b27200ea631284

**Comment by mrstanb** [comment]
`ldv-memsafety-bitfields/test-bitfields-2.1-2.yml` is also fixed after comparing `memcpy`'s `dest` size to `n` (cf. https://github.com/goblint/analyzer/pull/1201/commits/9b728d352169f0eea1e4a21609233413f2ae1c79). Also added this case as a regr. test in https://github.com/goblint/analyzer/pull/1201/commits/48f2cfedda0a61517a69e012dfc07a08fe3646cb

**Comment by michael-schwarz** [comment]
It now behaves as expected, and is sound:

```
Statistics:           3862 Files
  correct:            1209
    correct true:     1209
    correct false:       0
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:            2653
  Score:              2418 (max: 5857)
```

File: src/analyses/base.ml [hunk 11]
```diff
@@ -2050,13 +2131,36 @@ struct
...
+          let ds_eq_n =
+            begin try ID.eq casted_ds casted_n
+              with IntDomain.ArithmeticOnIntegerBot _ -> ID.bot_of @@ Cilfacade.ptrdiff_ikind ()
+            end
+          in
+          begin match ID.to_bool ds_eq_n with
+            | Some b -> b
+            | None -> false
+          end
+        | _, _ -> false
```
michael-schwarz, line 2156:
```suggestion
        match dest_size, n_intdom with
        | `Lifted ds, `Lifted n ->
          let casted_ds = ID.cast_to (Cilfacade.ptrdiff_ikind ()) ds in
          let casted_n = ID.cast_to (Cilfacade.ptrdiff_ikind ()) n in
          let ds_eq_n =
            begin try ID.eq casted_ds casted_n
              with IntDomain.ArithmeticOnIntegerBot _ -> ID.top_of @@ Cilfacade.ptrdiff_ikind ()
            end
          in
          Option.value ~default:false ID.to_bool ds_eq_n
        | _ -> false
```

Both being `Top` is not enough to assume that the copying will work.

src/analyses/base.ml
```diff
@@ -2050,13 +2131,36 @@ struct
     let st: store = ctx.local in
     let gs = ctx.global in
     let desc = LF.find f in
-    let memory_copying dst src =
+    let memory_copying dst src n =
+      let dest_size = get_size_of_ptr_target ctx dst in
+      let n_intdom = match n with
+        | Some exp -> ctx.ask (Queries.EvalInt exp)
+        | None -> `Bot
+      in
```
michael-schwarz, line 2139:
```suggestion
      let n_intdom = Option.map_default (fun exp -> ctx.ask (Queries.EvalInt exp)) `Bot n in
```

src/analyses/base.ml
```diff
@@ -1043,10 +1043,27 @@ struct
...
+            else if AD.may_be_null adr then (
+              AnalysisStateUtil.set_mem_safety_flag InvalidDeref;
+              M.warn ~category:M.Category.Behavior.Undefined.nullpointer_dereference ~tags:[CWE 476] "May dereference NULL pointer"
+            );
+            (* Warn if any of the addresses contains a non-local variable *)
+            AD.iter (function
+                | AD.Addr.Addr (v,o) ->
+                  if not (CPA.mem v st.cpa) && not (is_global a v) then (
+                    (* TODO: Not the smartest move to set the global flag within an iter *)
+                    (* TODO: We can resort to using AD.exists instead *)
```
michael-schwarz, line 1060: Yes, that would be a good idea!

src/analyses/base.ml
```diff
@@ -2023,14 +2040,78 @@ struct
...
+      )
+    | _ ->
+      AnalysisStateUtil.set_mem_safety_flag InvalidFree;
+      M.warn ~category:(Behavior (Undefined InvalidMemoryDeallocation)) ~tags:[CWE 590] "Pointer %a in function %s doesn't evaluate to a valid address. Invalid memory deallocation may occur" d_exp ptr special_fn.vname
+
+  let points_to_heap_only ctx ptr =
+    match ctx.ask (Queries.MayPointTo ptr) with
+    | a when not (Queries.AD.is_top a)->
+      Queries.AD.for_all (function
+          | Addr (v, o) -> ctx.ask (Queries.IsHeapVar v)
```
michael-schwarz, line 2061:
```suggestion
          | Addr (v, _) -> ctx.ask (Queries.IsHeapVar v)
```

src/util/cilfacade.ml
```diff
@@ -40,7 +40,9 @@ let is_first_field x = match x.fcomp.cfields with
 let init_options () =
   Mergecil.merge_inlines := get_bool "cil.merge.inlines";
   Cil.cstd := Cil.cstd_of_string (get_string "cil.cstd");
-  Cil.gnu89inline := get_bool "cil.gnu89inline"
+  Cil.gnu89inline := get_bool "cil.gnu89inline";
+  if get_bool "cil.addNestedScopeAttr" then
+    Cabs2cil.addNestedScopeAttr := true
```
michael-schwarz, line 45:
```suggestion
  Cabs2cil.addNestedScopeAttr := get_bool "cil.addNestedScopeAttr";
```

to be robust against changes of the default in future.

michael-schwarz (approved):
After the comments are addressed, I think we can move ahead and merge this.

src/analyses/base.ml
mrstanb: Changed to `AD.exists` in https://github.com/goblint/analyzer/pull/1201/commits/de0220baf762f69ba6f0da19299ce568b243bed4

goblint.opam.locked
```diff
@@ -122,6 +122,9 @@ build: [
 ]
 dev-repo: "git+https://github.com/goblint/analyzer.git"
 available: os-distribution != "alpine" & arch != "arm64"
+pin-depends: [
+  [ "goblint-cil.2.0.2" "git+https://github.com/goblint/cil.git#c7ffc37ad83216a84d90fdbf427cc02a68ea5331" ]
+]
```
sim642, line 127: There's already a `pin-depends` section below.
```suggestion
```

src/analyses/memOutOfBounds.ml
```diff
@@ -159,6 +165,59 @@ struct
...
+            | `Bot -> ID.bot_of @@ Cilfacade.ptrdiff_ikind ()
+            | `Lifted eval_x ->
+              let typ_size_in_bytes = size_of_type_in_bytes typ in
+              let casted_eval_x = ID.cast_to (Cilfacade.ptrdiff_ikind ()) eval_x in
+              let bytes_offset = ID.mul typ_size_in_bytes casted_eval_x in
+              let remaining_offset = cil_offs_to_idx ctx typ o in
+              ID.add bytes_offset remaining_offset
+          end
+        with IntDomain.ArithmeticOnIntegerBot _ -> ID.bot_of @@ Cilfacade.ptrdiff_ikind ()
+      end
```
sim642, line 193: This looks like a copy-paste of `to_index` in `Offset`.

src/analyses/useAfterFree.ml
```diff
@@ -179,17 +175,15 @@ struct
...
   let return ctx (exp:exp option) (f:fundec) : D.t =
     Option.iter (fun x -> warn_exp_might_contain_freed "return" ctx x) exp;
     ctx.local
 
   let enter ctx (lval:lval option) (f:fundec) (args:exp list) : (D.t * D.t) list =
     let caller_state = ctx.local in
     List.iter (fun arg -> warn_exp_might_contain_freed "enter" ctx arg) args;
-    if AllocaVars.is_empty (fst caller_state) && HeapVars.is_empty (snd caller_state) then
+    [caller_state, (AllocaVars.empty (), snd caller_state)]
+  (* if AllocaVars.is_empty (fst caller_state) && HeapVars.is_empty (snd caller_state) then
```
sim642, line 186: What's this commented-out code now?

src/util/options.schema.json
```diff
@@ -288,6 +288,12 @@
           "type": "boolean",
           "description": "Indicates whether gnu89 semantic should be used for inline functions.",
           "default": false
+        },
+        "addNestedScopeAttr": {
+          "title": "cil.addNestedScopeAttr",
+          "type": "boolean",
+          "description": "Indicates whether variables that CIL pulls out of their scope should be marked.",
+          "default": false
```
sim642, line 296: Why does this need to be an option at all? There's no harm in having the attribute.

tests/regression/74-use_after_free/15-juliet-uaf-global-var.c
```diff
@@ -0,0 +1,24 @@
+//PARAM: --set ana.activated[+] useAfterFree --set ana.activated[+] threadJoins
+#include <stdlib.h>
+#include <stdio.h>
+#include <pthread.h>
```
sim642, line 4:
```suggestion
//PARAM: --set ana.activated[+] useAfterFree
#include <stdlib.h>
```

tests/regression/78-invalid-deref-scopes/01-scopes-no-static.c
sim642, line 1: Isn't there any existing group to put these? We can't afford the three-test groups because we'll soon run out of group numbers and there's no GobExec for regression testing yet.

src/util/options.schema.json
michael-schwarz: There is when you do `justci`l, as it will cause many compiler warnings and pollute files. If we want to keep using Goblint to create merged files, we should keep this option imo.

sim642: Seems to me that it'd be better to make CIL's printer not just print these special attributes like it already does with some others.
That would avoid this annotation from ever making it to the output, even when you want to both enable it (for memory analyses) and also print the merged file. Currently those annotations would be in the merged file explicitly, so reanalyzing the merged file would incorrectly have some of these attributes on variables which are all pulled up to the function level. If the attribute wasn't printed, there wouldn't be a spurious attribute which would be incorrectly placed onto pulled-up variables.

michael-schwarz: I thought about it, but decided against it in the end: the reasoning was that I want some way to inspect where these attributes were put.

src/analyses/useAfterFree.ml
mrstanb: This is due to something that is left for future work in the UAF analysis:
* We used to consider only the caller state, reachable from the callee parameters
* However, we also need to consider everything from the caller state, reachable from the global program variables
* Problem is that we need a way to get a hold of the global vars
* Until then, we decided to simply keep the entire caller state as the callee state

I'll remove the commented-out portion and I'll add a TODO for the future work

src/analyses/memOutOfBounds.ml
michael-schwarz: We refactored this. Some duplication remains, but it's not immediately clear how to get rid of it.

mrstanb: Should be now fixed in https://github.com/goblint/analyzer/pull/1201/commits/f018ea37bc1e2cfb66237283b44a8400fc5cf161.

We now kind of reuse `convert_offset` from `base.ml`, but we'd need some better way of making this logic reusable in the future.

tests/regression/78-invalid-deref-scopes/01-scopes-no-static.c
mrstanb: I merged all UAF + invalid deref scope + memory OOB access test cases into an `invalid_deref` group and the double free cases as part of the `invalid_free` group. Should be hopefully a bit more concise now

sim642:
Due to the module moves in #1206 GitHub shows conflicts that are "too complex" to solve in the web editor. Actually you can locally just merge master into this branch and there shouldn't be any conflicts. GitHub's merge thing seems to have some move detection disabled that git should do by default.

src/util/options.schema.json
michael-schwarz: I added a corresponding case to `check_arguments` now.

sim642:
I thought this was supposed to be sound, but apparently it's not. For example
```console
./goblint --conf conf/svcomp.json --set ana.specification ../sv-benchmarks/c/properties/valid-memsafety.prp ../sv-benchmarks/c/loops/invert_string-2.c
```
prints `SV-COMP result: true` but the expected verdict if `false(valid-deref)`.

michael-schwarz:
It was sound when I last checked it, is this task new?

sim642:
No, it's a random one I just picked to check. It was present in some BenchExec table that was attached here before and there it correctly returned `unknown`.

michael-schwarz:
Maybe the autotuner is just not active in the setting you have given where `svcomp.enabled` is not set? I think it's something along these lines, we didn't do any fundamental changes since the last run.

sim642:
`conf/svcomp.json` enables all of that though.

michael-schwarz:
[svcomp23-with-mem-safety.json](https://github.com/goblint/analyzer/files/13216266/svcomp23-with-mem-safety.json)

That was the config we used.

michael-schwarz:
Maybe it's `arrayoob` that isn't set there?

sim642:
> Maybe it's `arrayoob` that isn't set there?

That seems to be it. I made the autotuner enable it automatically like it activates the analyses. 

Although I'm surprised that the `memOutOfBounds` analysis doesn't check these bounds.

PR #1202, opened 2023-10-03
Remove spurious `long double` constant warning

michael-schwarz:
This removes spurious warnings when analyzing code that contains floating point constants that are larger that OCaml floats. (c.f. https://github.com/goblint/cil/issues/136)

As suggested by @sim642, we `assert false` in Goblint when we encounter such a value without a string representation -  with an exception for the constant zero as it can be represented correctly regardless of type, and CIL generates it for zero-initialized data.


tests/regression/00-sanity/37-long-double.t
```diff
@@ -0,0 +1,6 @@
+Testing that there is warning about treating long double as double constant.
```
sim642, line 1: You mean that there _isn't_ a warning?

sim642 (approved):

PR #1204, opened 2023-10-04
[Model Checking w/ Local Traces] Witnessing Violations of Safety Properties for Concurrent Programs

sofiiadan:
References local trace semantics implementation: #1035 .

**Witnessing Violations of Safety Properties for Concurrent Programs**:
(by Sofiia Danylchenko : https://github.com/sofiiadan )

Implementation of Violation Witness for Local Traces

- implemented violation witness for local traces for safety violations "no-overflow" and "unreach-error"
- added unreach-error recognition on localTraceAnalysis
- extended the scope of C language for local traces by adding unary and binary operations, unsigned and signed types and overflow management, pointers

TODOs:

- add more tests for witness creation in specific scenarios
- extend local traces scope for C language for pointers
- add more safety properties violation recognition (for example, data races)
- add createThread as attributes in violation witness GraphML format (https://github.com/sosy-lab/sv-witnesses/blob/main/README-GraphML.md)

**Bounded Model Checking with Local Traces #1035**:
(by Olga Fadeenkov : https://github.com/faddeenkov)

Implementation of Local Trace Semantics via an Abstract Interpretation Analysis

- implemented suitable edge effects that construct local traces
- implemented multi-threading with pthread_create and pthread_join. Latter one can merge traces when suitable
- support mutex locking and unlocking which merges local traces when suitable
- implement detection error in the form of termination whenever one faulty trace is detected

TODOs:

- implement proper solver. WL-solver utilizes sets for worklists, this is not reliable.
- Fix concept of dependency edges for mutexes such that a dependency edge goes to the 'locked' node (for now it goes to the locking node)


michael-schwarz:
I ran some benchmarks over the weekend, and form the results it seems that completeness of this approach does not hold yet: For many programs we claim to have found a violation despite the task being tagged as correct.

**For unreach-call**:

unreach-call


**For no-overflow**:

no_overflow


These would need to be resolved before it makes sense to compete in SV-COMP with this approach.

michael-schwarz:
Thank you again for your contribution! We decided to not pursue this concrete line of research for now, and I am thus closing this PR.

Issue #1126, opened 2023-08-03
Regressions from string literals domain


sim642:
On our pthread bench programs, I've already found via `git bisect` two regressions caused by #1048:

1. e94b0cce9a15ebbd5a20f063dcf08ba8a17289da causes aget_comb evals to go from 7138 to 7932.

2. 52fea5d72beabc85fb6c522b957965e9f0e1d24e causes fixpoint not reached errors on smtprc_comb.

PR #1208, opened 2023-10-06
Add `ana.base.strings.domain` option and unit string domain

sim642:
In https://github.com/goblint/bench/issues/62#issuecomment-1750439256 I noted:

One possible reason for so many contexts with `ana.base.context.non-ptr` disabled could be string literals (`StrPtr`) in address sets, which would be preserved in contexts. `ana.base.limit-string-addresses` (true by default) doesn't do enough limitation. When true, string literals are handled like a flat lattice (a la constant propagation), which still yields separate contexts for all different calls with string literal arguments. When false, string literals are handled as a set.

It would be possible to do even more limiting: not track any string literals, i.e. use a unit domain for them. I hacked it by making `of_string` always return `StrPtr None`. This at least got rid of many contexts for some functions, but not all.

This PR implements that hack more cleanly by extracting a `StringDomain` module and introducing the option `ana.base.strings.domain` with three possible values:
1. `disjoint` corresponds to `ana.base.limit-string-addresses` false.
2. `flat` corresponds to `ana.base.limit-string-addresses` true.
3. `unit` corresponds to the new even more imprecise domain with no information at all.


michael-schwarz:
I would strategically propose to delay this until @nathanschmidt can integrate his PR #1076 and potentially also #1118, as these will significantly extend the string domains, and the interface that is needed will hopefully be more clear after this (though I am not sure how significant the overlap is between these two issues here anyway).

@nathanschmidt could you comment with an estimated timeline for addressing the comments on the PRs? Iirc you said no earlier than this week?

sim642:
Neither of those has to do with the `StrPtr`s in `AddressDomain` though: one modifies array domains and the other adds a new relational analysis. So they're orthogonal and there's no conflict.

Right now, the big problem for us analyzing real-world programs is performance, not imprecision. Already given #1126, it's beneficial to have the ability turn off all these new things. We're shooting ourselves in the foot on real-world programs if the analyses keep getting more expensive without a choice.

nathanschmidt:
> @nathanschmidt could you comment with an estimated timeline for addressing the comments on the PRs? Iirc you said no earlier than this week?

@michael-schwarz I currently have quite some time constraints, but I plan to have a look next week.

src/cdomains/stringDomain.ml
```diff
@@ -0,0 +1,97 @@
+include Printable.StdLeaf
+
+let name () = "string"
+
+type t = string option [@@deriving eq, ord, hash]
+
+let hash x =
+  if GobConfig.get_string "ana.base.strings.domain" = "disjoint" then
```
jerhard, line 8: It may make sense to cache the value for `"ana.base.strings.domain"` in some Variant typed `ref`.

jerhard (approved):

src/cdomains/stringDomain.ml
jerhard: I now added a `ref` caching the value of this configuration.

michael-schwarz:
As I have been delayed with #1076 and #1118, I think it might make sense to merge this before those. Given that this does not change the behavior for SV-COMP, I'd be open to merging it even before Nov 22.

Issue #1190, opened 2023-09-26
Final errors about unsound results


sim642:
Recent experiences have shown that it's too easy to miss key messages about unsoundness on real-world projects with lots of output. This is made worse by confs that disable most output, making such problems impossible to notice.

The solution would be to print additional errors after all other messages when some of the following have occurred:
- [x] Fixpoint not reached.
- [x] Function definition missing.
- [x] Both branches dead.
- [x] No suitable function to call.
- [x] ASM ignored.
- [ ] ...?

By making these errors last in the output, they will be visible on the terminal after such long analysis finishes and be more likely to be noticed.
Also they should not be suppressed by options to make sure we always notice invalid results before using them in papers.

PR #1210, opened 2023-10-09
Fix longjmp crash on Uninitialized

sim642:
On concrat/pigz the longjmp analysis crashes because it takes a `jmp_buf` out from a `pthread_getspecific` which currently gives top results.
This PR replaces the crash with a message and a top value.

### DIFF
--- a/tests/regression/68-longjmp/56-longjmp-top.c
+++ b/tests/regression/68-longjmp/56-longjmp-top.c
@@ -12,7 +12,7 @@ int main() {
   if (!setjmp(buf)) {
     jmp_buf *buf_ptr;
     buf_ptr = pthread_getspecific(buf_key);
-    longjmp(*buf_ptr, 1); // TODO NO CRASH: problem?!
+    longjmp(*buf_ptr, 1); // NO CRASH: problem?!
   }
   else {
     __goblint_check(1); // reachable
--- a/tests/regression/68-longjmp/56-longjmp-top.c
+++ b/tests/regression/68-longjmp/56-longjmp-top.c
@@ -15,7 +15,7 @@ int main() {
     longjmp(*buf_ptr, 1); // NO CRASH: problem?!
   }
   else {
-    __goblint_check(1); // reachable
+    __goblint_check(1); // TODO reachable: https://github.com/goblint/analyzer/pull/1210#discussion_r1350021903
   }
   return 0;
 }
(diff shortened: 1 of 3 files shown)


tests/regression/68-longjmp/56-longjmp-top.c
```diff
@@ -0,0 +1,21 @@
...
+  jmp_buf buf;
+  pthread_setspecific(buf_key, &buf);
+
+  if (!setjmp(buf)) {
+    jmp_buf *buf_ptr;
+    buf_ptr = pthread_getspecific(buf_key);
+    longjmp(*buf_ptr, 1); // NO CRASH: problem?!
+  }
+  else {
+    __goblint_check(1); // reachable
```
sim642, line 18: The crash on `longjmp` is fixed but apparently this doesn't still become reachable?

michael-schwarz: Returning top here doesn't make code that is otherwise unreachable reachable iirc... I think this is a fundamental limitation that I am not sure how to address best.

sim642: I thought a top jmpbuf would just jump back to every `setjmp` without filtering by node/context, but apparently all we do is warn:
https://github.com/goblint/analyzer/blob/5cc481148d4e079327e6f395c02e28e99cdaa414/src/framework/constraints.ml#L1664-L1666
Maybe we should have an option to do that?

sim642: > The problem is a bit where one wants to propagate things to... Just up the call stack, assuming the invocation was not UB? Also down the callstack because someone might be trying to implement co-routines? Also to completely unrelated places?

To the same places we normally do: in the same function and up the call stack. It could just be that all `setjmp`s incorporate the unknown jump target instead of the same node and context check that normally happens. So unknown jump buffer would mean "all jumpbuffers that we normally could jump to".


> The other problem is that quite often the local state propagated somewhere because of an unknown longjmp will not be suitable for incorporation by join, as it is incorporated into an unsuitable function. This might lead to a huge precision loss.

Of course it would be very imprecise, but I'm not sure if it'd cause anything incompatible per se. Ambiguous longjmps to multiple possible targets should be similar. The compatibility is ensured by each upwards propagation of the jumps doing the appropriate `return`s and `combine_env`s.

> My idea would be to leave this as is, but make this type of warning one of the severe top-level warnings that is reported separately at the end of the analysis (in the sense of #1190 ).

I added that, so let's leave it at that.

PR #1213, opened 2023-10-09
Check duplicate library functions

sim642:
This adds duplicate library function checks, both within one library and across libraries. It revealed a handful of duplicates to remove. Previously the duplicates were silently ignored by `BatHashtbl.of_list`.
Also simplifies the old invalidate actions table construction because there's no point in manually-implemented laziness if a duplication check runs eagerly anyway.


karoliineh (approved):
Works wonderfully.

PR #1215, opened 2023-10-12
Improve GobView build

stilscher:
This PR attempts to simplify the build process and usage of GobView:

- new make target that copies the executable to the top-level directory (as done with the Goblint executable)
- copy the transpiled JavaScript files during the build already instead of doing it during the first run of Goblint (this allows to later extend GobView to start the first run from within the UI)
- update documentation

- [x] bump Gobview after https://github.com/goblint/gobview/pull/33 has been merged


src/maingoblint.ml
```diff
@@ -577,11 +577,7 @@ let do_html_output () =
 
 let do_gobview cilfile =
   let gobview = GobConfig.get_bool "gobview" in
-  let goblint_root = GobFpath.cwd_append (fst (Fpath.split_base (Fpath.v Sys.argv.(0)))) in
-  let dist_dir = Fpath.(goblint_root // (Fpath.v "_build/default/gobview/dist")) in
-  let js_file = Fpath.(dist_dir / "main.js") in
   if gobview then (
-    if Sys.file_exists (Fpath.to_string js_file) then (
       let save_run = GobConfig.get_string "save_run" in
       let run_dir = Fpath.v(if save_run <> "" then save_run else "run") in
```
sim642, line 582: `make view` copies things into `run/` but this output can be configured to go into other directories, where the files would be missing. Not sure what we should do about that.

stilscher: Good point. I think it does not really need to be configurable, but we should maybe choose a different name then for this folder.

scripts/creduce/warning.sh
sim642, line 1: Some unrelated scripts seem to have been committed here.

src/maingoblint.ml
sim642: It might even be better if the compiled Gobview (of which there is just one) didn't go to the same directory as each Goblint marshalled output (of which there could be multiple with `save_run`). From `./run` it'd be possible to refer to `../gobview-dist/` or whatnot. But maybe modern web browsers/servers forbid such up-going relative path in some cases, not sure.

stilscher: In https://github.com/goblint/gobview/pull/33 I made a suggestion on how one can change the goblint-http server to serve from both directories (with priority on gobview-dist).

michael-schwarz:
What is blocking the merge here?

stilscher:
I think we do want to have a separate distribution directory for GobView, but that requires some additional changes to the goblint-http server that need to be realized in the GobView repository first.

.gitignore
```diff
@@ -10,6 +10,7 @@ goblint
 goblint.byte
 goblint.json
 goblint.domaintest
+goblint-http
```
sim642, line 13: Isn't it named `goblint-http.exe` now? Could also make the rule `goblint-http*` I guess.

scripts/test-gobview.py
```diff
@@ -24,8 +24,8 @@ def cleanup(browser, thread):
 # serve GobView in different thread so it does not block the testing
 def serve():
   global p
-  goblint_http_path = '_build/default/gobview/goblint-http-server/goblint_http.exe'
-  p = subprocess.Popen(['./' + goblint_http_path,
+  goblint_http_path = './goblint-http'
```
sim642, line 27: Should this also be `./goblint-http.exe`?

scripts/test-gobview.py
```diff
@@ -24,8 +24,8 @@ def cleanup(browser, thread):
...
+  p = subprocess.Popen([goblint_http_path,
                   '-with-goblint', '../analyzer/goblint',
                   '-goblint', '--set', 'files[+]', '"../analyzer/tests/regression/00-sanity/01-assert.c"'])
```
sim642, line 30: These `-with-goblint`, `-goblint` and `--set` arguments are no longer needed? At least according to the updated inspecting documentation.

sim642 (commented):
Sorry for forgetting about this again!
I guess this should all work once conflicts with master are resolved and the submodule is updated to https://github.com/goblint/gobview/pull/33.

src/maingoblint.ml
```diff
@@ -621,7 +621,9 @@ let do_gobview cilfile =
       (* marshal timing statistics *)
       let stats = Fpath.(run_dir / "stats.marshalled") in
       Serialize.marshal (Timing.Default.root, Gc.quick_stat ()) stats;
-  )
+    )
+    else
+      Logs.error "Warning: Cannot locate GobView."
```
sim642, line 626: It's a bit confusing to have an error say it's a warning:
```suggestion
      Logs.error "Cannot locate GobView."
```
I might have caused this on master, but this is a good opportunity to fix it.

sim642 (approved):

Issue #392, opened 2021-10-13
Add unknown thread ID


sim642:
As revealed in #388, it would be very good to have an unknown thread ID, just like the unknown address in address sets. This would allow keeping known thread IDs while invalidating/joining with incompatible/etc.

michael-schwarz:
Anything left to be done here now that #1224 is merged?

sim642:
I don't think so. Looks like the issue just wasn't fully linked with the PR.

PR #1224, opened 2023-10-27
Add unknown thread ID

karoliineh:
Relates to #392.
- Adds unknown thread ID to `ThreadIdDomain.Thread`
- Uses `SetDomain` with `UnknownThread` element as top instead of `ToppedSet` for `ThreadSet` in  `concDomain`
- Adds `UnknownThread` element to `ThreadSet` when invalidating. Fixes an unsoundness issue and adds a test case.

However, I am not too sure what having an `UnknownThread` element and preserving other thread IDs in the set is good for. I could not find a case where this would improve our precision. Maybe others have ideas?


src/analyses/apron/relationPriv.apron.ml
```diff
@@ -1011,17 +1011,17 @@ struct
...
-      | _ ->
-        (* To match the paper more closely, one would have to join in the non-definite case too *)
-        (* Given how we handle lmust (for initialization), doing this might actually be beneficial given that it grows lmust *)
-        st
-      | exception SetDomain.Unsupported _ ->
-        (* elements throws if the thread set is top *)
-        st
+      if ConcDomain.ThreadSet.is_top tids
+        then st
+        else
```
michael-schwarz, line 1016: Then indentation is a bit wacky here.

michael-schwarz (approved):

sim642:
> It is not an analysis we currently have, but e.g. for an analysis trying to detect the case where thread may be joined multiple times by the same thread, one could imagine tracking may-joined threads

It's a slippery slope to do things for hypothetical analyses because it has a performance impact _right now_. A set of known threads in addition to unknown can increase, creating even more contexts that we all analyze identically with existing analyses. We already have issues with contexts exploding on real-world programs (including on not-so-large Concrat ones), so we're just shooting ourselves in the foot performance-wise with absolutely no benefit right now.

The soundness fixes here also work with the current `ToppedSet` and can be made independently from the performance-impacting "unknown thread with known threads set".

michael-schwarz:
> A set of known threads in addition to unknown can increase, creating even more contexts that we all analyze identically with existing analyses. 

Yes, this is certainly true. However, I think having a clean solution, that, e.g., coincides with the solution we have for pointers is preferable to these potential gains here.

Also, if we encounter such contexts being problematic in the future, we could still always opt to construct contexts in a different way that do not admit this blowup. This is the beauty of having completely decoupled contexts from local values by giving up on simulating the side-effect free formulation of function-calls.

michael-schwarz:
As soon as the conflict with master is resolved, I think we can merge this.

src/analyses/base.ml
```diff
@@ -2372,6 +2372,7 @@ struct
         | Int n when GobOption.exists (BI.equal BI.zero) (ID.to_int n) -> st
         | Address ret_a ->
           begin match eval_rv (Analyses.ask_of_ctx ctx) gs st id with
+            | Thread a when ValueDomain.Threads.is_top a -> invalidate ~ctx (Analyses.ask_of_ctx ctx) gs st [ret_var]
```
sim642, line 2375: In light of #1264, this and the other `invalidate`s for `ThreadJoin` should also be shallow I think. `pthread_join` writes the return value, but doesn't dereference its previous value for writing (it even can't because it's `void*`).

michael-schwarz:
CI failure is unrelated.

PR #1230, opened 2023-11-01
`threadAnalysis`: Only add to set of must-joined threads if argument to `pthread_join` evaluates to a singleton

### LINKED ISSUE #1223: Old `thread` analysis is unsound w.r.t. joins
The old thread analysis is unsound w.r.t. to joins as evidenced by the following program which is incorrectly reported as not racy.

~~~C
// PARAM: --set ana.activated[+] thread
#include <pthread.h>

int data = 0;
pthread_mutex_t data_mutex;

void *thread(void *arg) {
  pthread_mutex_lock(&data_mutex);
  data = 3; // RACE!
  pthread_mutex_unlock(&data_mutex);
  return NULL;
}

int main() {
  pthread_t tids[2];

  pthread_create(&tids[0], NULL, &thread, NULL);
  pthread_create(&tids[1], NULL, &thread, NULL);

  pthread_join(tids[0], NULL);

  data = 1; //RACE!

  return 1;
}
~~~

-------------------------

**Old description for reference:**

Running Goblint on `thread-join-array-const-race` from https://github.com/goblint/bench/pull/63 reveals that our unrolling is unsound w.r.t. thread joins. The program is
```c
// Thread pool constant array sequential joining.
// Extracted from concrat/level-ip.
#include <pthread.h>

int data = 0;
pthread_mutex_t data_mutex = PTHREAD_MUTEX_INITIALIZER;

void *thread(void *arg) {
  pthread_mutex_lock(&data_mutex);
  data = __VERIFIER_nondet_int(); // RACE!
  pthread_mutex_unlock(&data_mutex);
  return NULL;
}

int main() {
  int threads_total = 4;
  pthread_t tids[4];

  // create threads
  for (int i = 0; i < threads_total; i++) {
    pthread_create(&tids[i], NULL, &thread, NULL); // may fail but doesn't matter
  }

  // join threads
  for (int i = 0; i < threads_total - 1; i++) {
    pthread_join(tids[i], NULL);
  }

  return data; // RACE! (tids[3] not joined)
}
```
With the `svcomp` conf, we report no race, which is unsound. If `loopUnrollHeuristic` is disabled, then we become sound again. Therefore, thread joining is unsound when combined with unrolling.

michael-schwarz:
The analysis added all unique threads in the points-to-set of the argument to `pthread_join` to the set of must-joined threads.
This is wrong, as only one of those is joined at runtime, so only if the set is singleton and the thread is unique can it be added.

Closes #1223

### DIFF
--- a/src/analyses/threadAnalysis.ml
+++ b/src/analyses/threadAnalysis.ml
@@ -61,7 +61,8 @@ struct
            s
        in
        match TS.elements (ctx.ask (Queries.EvalThread id)) with
-       | threads -> List.fold_left join_thread ctx.local threads
+       | [t] -> join_thread ctx.local t (* single thread *)
+       | _ -> ctx.local (* if several possible threads are may-joined, none are must-joined *)
        | exception SetDomain.Unsupported _ -> ctx.local)
     | _ -> ctx.local
 
--- a/tests/regression/10-synch/28-join-array.c
+++ b/tests/regression/10-synch/28-join-array.c
@@ -1,0 +1,25 @@
+// PARAM: --set ana.activated[+] thread
+#include <pthread.h>
+
+int data = 0;
+pthread_mutex_t data_mutex;
+
+void *thread(void *arg) {
+  pthread_mutex_lock(&data_mutex);
+  data = 3; // RACE!
+  pthread_mutex_unlock(&data_mutex);
+  return NULL;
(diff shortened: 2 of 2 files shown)


michael-schwarz:
The failure of `semgrep` is unrelated.

PR #1233, opened 2023-11-02
Fix both branches dead from bot address in array

### LINKED ISSUE #1188: Both branches dead in zstd
In https://github.com/goblint/analyzer/commit/810fab57274821d600b5d841a8a1be977311a3b2 I added a case extracted from zstd where we go unsound because we think both branches are dead.

It has a `malloc`ed blob inside a `calloc`ed blob, so something weird is happening with those and the condition evaluates to `bot`.

### LINKED ISSUE #1232: Both branches dead unsoundness on Concrat
We have `[Error][Analyzer][Unsound] Both branches dead` on two Concrat benchmarks:
- [x] kona:

    ```
    [Error][Analyzer][Unsound] both branches over condition '! s' are dead (/mnt/goblint-svcomp/goblint-bench/bench/concrat/kona/main.c:26433:7-26433:10)
    ```
- [x] Remotery

    ```
   [Error][Analyzer][Unsound] both branches over condition '! ((unsigned long )tree != (unsigned long )((void *)0))' are dead (/mnt/goblint-svcomp/goblint-bench/bench/concrat/Remotery/main.c:4782:7-4782:62)
   [Error][Analyzer][Unsound] both branches over condition '! ((unsigned long )tree != (unsigned long )((void *)0))' are dead (/mnt/goblint-svcomp/goblint-bench/bench/concrat/Remotery/main.c:4622:7-4622:62)
   [Error][Analyzer][Unsound] both branches over condition '! ((unsigned long )tree != (unsigned long )((void *)0))' are dead (/mnt/goblint-svcomp/goblint-bench/bench/concrat/Remotery/main.c:4662:7-4662:62)
   [Error][Analyzer][Unsound] both branches over condition '! ((unsigned long )tree != (unsigned long )((void *)0))' are dead (/mnt/goblint-svcomp/goblint-bench/bench/concrat/Remotery/main.c:4700:7-4700:62)
   ```

sim642:
Closes #1232. Closes #1188.

The fix is from https://github.com/goblint/analyzer/issues/1188#issuecomment-1735060169. I'm not sure if this is the right fix, but it is _a_ fix.


michael-schwarz (changes requested):
Giving the value inside the array a super-bot value instead of a bottom value of correct type might fix this example, but is not a principled solution one would want to merge.

Such a solution should address what is read.

sim642:
> Such a solution should address what is read.

The points-to set is empty, there is nothing to read!

The morally right thing would be to have the value that's actually there: a top address set. But some array domains don't work that way and there's nothing nicer we can do about it before the PLDI deadline.

michael-schwarz:
Maybe you can use the partitioned array domain and not this one for which it is dubious what it does for programs with undefined behavior such as this example anyway?
Or have a separate branch for your PLDI drive, instead of merging this into the master?

sim642:
> Maybe you can use the partitioned array domain and not this one for which it is dubious what it does for programs with undefined behavior such as this example anyway?

There's nothing undefined about this. It can arise from every unknown pointer and that's simply what happens in all the programs from the issues. It's our as-sound-as-possible handling of unknown pointers that's causing us to be unsound in this case.

sim642:
> Then maybe what we have to do for invalidate calls is to assign a `top_of_type` value

My change is in that `top_value`. Not putting bottoms would mean eliminating the weird `can_recover_from_top` logic.

> The non-partitioning array-domain is extremely dubious for code that has library functions that write into arrays for this reason, and should not be used for any programs that contain such calls to library functions.

Partitioned arrays are a single joined value just like trivial arrays if there's no partitioning. If that case of partitioned arrays is somehow more sound, we should just make trivial arrays behave the same way. Invalidation should kill any partitioning anyway, so it's not the partitioning that's making the result more sound here.

michael-schwarz:
> My change is in that `top_value`. Not putting bottoms would mean eliminating the weird `can_recover_from_top` logic.

Yes, we just have to make sure it's not at the expense of killing any precision for the non-partitioned array domain in other places where `top_value` is used but putting an initial bottom is ok (such as for initial values, as discussed before).
So doing this would require a careful review  of where `top_value` is used, and replacing any uses where initializing with bottom is ok.
Then `can_recover_from_top` can be limited to computing these initial values. I would still keep it there, because otherwise we might just as well not track arrays at all, which would be bad. on the other hand, initializing those array domains that can recover from top to top allows being sound for programs with UB, so it is an important distinction to keep.

src/cdomains/valueDomain.ml
```diff
@@ -201,7 +201,7 @@ struct
       let typAttr = typeAttrs ai in
       let can_recover_from_top = ArrayDomain.can_recover_from_top (ArrayDomain.get_domain ~varAttr ~typAttr) in
       let len = array_length_idx (IndexDomain.top ()) length in
-      Array (CArrays.make ~varAttr ~typAttr len (if can_recover_from_top then (top_value ai) else (bot_value ai)))
+      Array (CArrays.make ~varAttr ~typAttr len (if can_recover_from_top then (top_value ai) else Bot))
```
jerhard, line 204: At least for the new example, it may be appropriate to actually put the top value of the appropriate type here.

sim642: As suggested in https://github.com/goblint/analyzer/pull/1233#issuecomment-1792074791, I made this change, eliminating the `can_recover_from_top` logic here altogether. As far as we test, no precision is lost.

michael-schwarz:
I'll review this in detail either tomorrow or Saturday.

michael-schwarz (commented):
It seems at first glance that `top_value` is indeed only called in places where it is not an initial value. I would still feel more comfortable if we could do a run on SV-COMP to confirm it has no precision impact before merging this.

michael-schwarz (approved):

PR #1234, opened 2023-11-03
Experiment with path- and context-sensitivity for memLeak analysis

michael-schwarz:
I'll use this issue to document some of the steps towards coming up with a good config for sv-comp.

<details>
  <summary>Machine Config</summary>
  
```
parallel runs:           40
resource limits:
- memory:                3000.0 MB
- time:                  1800 s
- cpu cores:             1
hardware requirements:
- cpu cores:             1
- memory:                3000.0 MB
------------------------------------------------------------

   SYSTEM INFORMATION
host:                    goblint2
os:                      Linux-5.15.0-84-generic-x86_64-with-glibc2.29
cpu:                     Intel Xeon Platinum 8260 CPU @ 2.40GHz
- cores:                 96
- max frequency:         3900.0 MHz
- turbo boost enabled:   True
ram:                     540649.373696 MB
------------------------------------------------------------
```
</details>

**Results without any additional path- & context-sensitivity:**

```
Statistics:           3862 Files
  correct:            1082
    correct true:     1082
    correct false:       0
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:            2780
  Score:              2164 (max: 5857)
  ```
  
  **Results with full path- & context-sensitivity:**
  
  ```
Statistics:           3862 Files
  correct:            1104
    correct true:     1104
    correct false:       0
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:            2758
  Score:              2208 (max: 5857)
  ```
  
  **Results with full path- & context-sensitivity & unique count of 5:**
  
  ```
  Statistics:           3862 Files
  correct:            1131
    correct true:     1131
    correct false:       0
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:            2731
  Score:              2262 (max: 5857)
```

  **Results with full path- & context-sensitivity & unique count of 5 & fix for #1235 :**
 
```
Statistics:           3862 Files
  correct:            1158
    correct true:     1158
    correct false:       0
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:            2704
  Score:              2316 (max: 5857)
```

PR #1241, opened 2023-11-07
Account for allocated heap memory which is unreachable from globals

mrstanb:
This PR:
* Adds a refinement to the reporting for `valid-memtrack`
* We now warn whenever there's allocated heap memory at program exit which is not reachable from the global program vars

Some notes:
* The SV-COMP rule for `valid-memtrack` states: `All allocated memory is tracked, i.e., pointed to or deallocated`. Hence, for now I've kept the setting of the global SV-COMP flag for `InvalidMemTrack` also for the case of not calling `free` on allocated heap memory (not necessarily only referenced by global vars). I'm open for a discussion here if you think we could improve on that as well
* There are some false positives for heap memory, reachable through local variables in case it's not freed (I also described this in test `76/09`):
```c
    g = malloc(sizeof(int));
    g = NULL;
    int *p = malloc(sizeof(int));

    return 0; //WARN
}
```

In the above example, `p`'s memory is leaked (missing `free`), but it's also reported as unreachable (because it's not reachable via the globals, but only via a local var). Should we try to improve on that (maybe at least in the error reporting for the unreachable-from-globals-memory case) or are we ok with it as is?

Sidenote:
* Added `76/08` and `76/09` as new test cases

TODOs:
* [ ] If `earlyglobs` is activated, then abort the analysis


michael-schwarz:
Thank you for the PR!

> In the above example, p's memory is leaked (missing free), but it's also reported as unreachable (because it's not reachable via the globals, but only via a local var).  Should we try to improve on that (maybe at least in the error reporting for the unreachable-from-globals-memory case) or are we ok with it as is?

Confusingly this is exactly the definition of `valid-memtrack` in SV-COMP (I was also surprised by this). See e.g.: https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1464
So we cannot really improve here and actually need to report this as a violation of `valid-memtrack`.

tests/regression/76-memleak/09-unreachable-with-local-var.c
```diff
@@ -0,0 +1,17 @@
...
+
+int *g;
+
+int main(int argc, char const *argv[]) {
+    g = malloc(sizeof(int));
+    // Reference to g's heap contents is lost here
+    g = NULL;
+    // We get a false positive for p's memory being unreachable
+    // It's still leaked, but due to free() being commented out
+    // TODO: Should we only improve the error reporting for unreachable memory in this case?
```
michael-schwarz, line 12: At least for `valid-memtrack` this is not actually a false-positive.

src/analyses/memLeak.ml
```diff
@@ -19,6 +19,22 @@ struct
...
 
   (* HELPER FUNCTIONS *)
+  let get_global_vars () =
+    (* Filtering by GVar seems to account for declarations, as well as definitions of global vars *)
+    List.filter_map (function GVar (v, _, _) -> Some v | _ -> None) !Cilfacade.current_file.globals
+
+  let get_reachable_mem_from_globals (global_vars:varinfo list) ctx =
+    global_vars
+    |> List.map (fun v -> Lval (Var v, NoOffset))
+    |> List.filter_map (fun exp ->
```
michael-schwarz, line 29: This can be improved a little: If we have global structs for example, we should follow all the pointer fields of the struct.

src/analyses/memLeak.ml
```diff
@@ -27,17 +43,25 @@ struct
...
+      (* Check and warn if there's unreachable allocated memory at program exit *)
+      let allocated_and_unreachable_mem = D.diff allocated_mem reachable_mem in
+      if not (D.is_empty allocated_and_unreachable_mem) then (
+        set_mem_safety_flag InvalidMemTrack;
+        M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "There is unreachable allocated heap memory at program exit. A memory leak might occur for the alloc vars %a\n" (Pretty.d_list ", " CilType.Varinfo.pretty) (D.elements allocated_and_unreachable_mem)
+      );
+      (* Check and warn if some of the allocated memory is not deallocated at program exit *)
       match assert_exp_imprecise, exp with
       | true, Some exp ->
         set_mem_safety_flag InvalidMemTrack;
```
michael-schwarz, line 58: We are already setting that flag above now.

```suggestion
```

src/analyses/memLeak.ml
```diff
@@ -27,17 +43,25 @@ struct
...
         set_mem_safety_flag InvalidMemcleanup;
-        M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "assert expression %a is unknown. Memory leak might possibly occur for heap variables: %a" d_exp exp D.pretty state
+        M.warn ~category:(Behavior (Undefined MemoryLeak)) ~tags:[CWE 401] "Assert expression %a is unknown. Memory leak might possibly occur for heap variables: %a" d_exp exp D.pretty allocated_mem
       | _ ->
         set_mem_safety_flag InvalidMemTrack;
```
michael-schwarz, line 62: Same here.

```suggestion
```

michael-schwarz:
I will start a run with the two suggestions from above applied to see what the results are.

michael-schwarz:
With these two suggestions on top of #1234, we succeed on an additional 17 tasks:

```
Statistics:           3862 Files
  correct:            1175
    correct true:     1175
    correct false:       0
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:            2687
  Score:              2350 (max: 5857)
```

michael-schwarz:
Tables: [results.2023-11-08_10-02-03.zip](https://github.com/goblint/analyzer/files/13294538/results.2023-11-08_10-02-03.zip)

tests/regression/76-memleak/09-unreachable-with-local-var.c
mrstanb: Ah, yes. I saw from the discussion in https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1464 that if the memory is reachable from locals then it's a correct reporting of a `valid-memtrack` violation.

I'll remove the `// TODO` in this case.

src/analyses/memLeak.ml
mrstanb: Thanks for the idea! I quickly tried out a few regr. tests with global struct variables (both pointer and non-pointer ones, as well as with nested structs; cf. `76/{10, 11, 12}`)

It seems that without changing anything in the implementation, Goblint can already detect the `valid-memtrack` violations in these test cases.

Nonetheless, if you think it makes sense to add anything further to the implementation, I'm more than happy to do that.

michael-schwarz: > It seems that without changing anything in the implementation, Goblint can already detect the `valid-memtrack` violations in these test cases.

Great! I think the other interesting case would be the cases where `valid-memtrack` holds, because something is still reachable via a global struct member. Have you also tried that direction?

sim642: If it means reachable, then it should use `Queries.ReachableFrom`.

michael-schwarz: No, it is **must**-reachable, while the query is **may**-reachable.

sim642: The current `MayPointTo` query is also a may-query. The must-ness comes from the singleton check of the may result.

Maybe the non-transitive aspect of `MayPointTo` is fine here and must-reachability via pointers would just make us less precise? In that case the function should also be named accordingly to avoid the inconsistency.

michael-schwarz: Must-reachability by pointers would make us more precise: One would consider all the things that the globals (or it's struct members) must point to (i.e. singleton points-to-set), and then can consider these as roots as well to include such things that must-be-reachable from these.
Doing a MayReachable query and then only considering the set if it is a singleton set would be less precise than considering the MayPointTo set if it is singleton.

mrstanb:
I added some changes regarding the check of whether some memory is still transitively reachable via the points-to set of a global struct variable (both of pointer and non-pointer type), as well as its fields and the contents of the points-to sets.

I also added two new regression test cases to check if the changes make sense (cf. `76/{13, 14}`).

src/analyses/memLeak.ml
mrstanb: > Great! I think the other interesting case would be the cases where valid-memtrack holds, because something is still (edit: must-)reachable via a global struct member. Have you also tried that direction?

I added some changes with respect to this suggestion. Please check out my comment here: https://github.com/goblint/analyzer/pull/1241#issuecomment-1813364110

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,105 @@
...
+    global_struct_non_ptr_vars
+    (* Filter out global struct vars that don't have pointer fields *)
+    |> List.filter_map (fun v ->
+        match ctx.ask (Queries.EvalValue (Lval (Var v, NoOffset))) with
+        | a when not (Queries.VD.is_top a) ->
+          begin match a with
+            | Queries.VD.Struct s ->
+              let struct_fields = ValueDomain.Structs.keys s in
+              let ptr_struct_fields = List.filter (fun f -> isPointerType f.ftype) struct_fields in
+              if List.length ptr_struct_fields = 0 then None else Some (s, ptr_struct_fields)
```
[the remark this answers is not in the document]
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1241/commits/845910410bbda2f9964ccd727d63a6e7b811ac7f

mrstanb:
Maybe we should add a check that if `earlyglobs` is activated, then we should abort the analysis?

Iirc, we discussed with @jerhard that `earlyglobs` would stop us from obtaining meaningful information about what global vars can reference.

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,105 @@ struct
   let context _ d = d
 
   (* HELPER FUNCTIONS *)
+  let get_global_vars () =
+    (* Filtering by GVar seems to account for declarations, as well as definitions of global vars *)
+    List.filter_map (function GVar (v, _, _) -> Some v | _ -> None) !Cilfacade.current_file.globals
```
sim642, line 25: Does CIL really guarantee that? I think in other places we don't assume it does and also handle `GVarDecl`.

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,105 @@ struct
...
+
+  let get_global_struct_ptr_vars () =
+    List.filter_map (function GVar (v, _, _) -> Some v | _ -> None) !Cilfacade.current_file.globals
+    |> List.filter (fun v -> match v.vtype with TPtr (TComp _, _) | TPtr ((TNamed ({ttype = TComp _; _}, _)), _) -> true | _ -> false)
+
+  let get_global_struct_non_ptr_vars () =
+    List.filter_map (function GVar (v, _, _) -> Some v | _ -> None) !Cilfacade.current_file.globals
+    |> List.filter (fun v -> match v.vtype with TComp (_, _) | (TNamed ({ttype = TComp _; _}, _)) -> true | _ -> false)
```
sim642, line 33: Wouldn't `List.partition` give both in a single pass?

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,105 @@ struct
   let context _ d = d
 
   (* HELPER FUNCTIONS *)
+  let get_global_vars () =
+    (* Filtering by GVar seems to account for declarations, as well as definitions of global vars *)
+    List.filter_map (function GVar (v, _, _) -> Some v | _ -> None) !Cilfacade.current_file.globals
+
+  let get_global_struct_ptr_vars () =
+    List.filter_map (function GVar (v, _, _) -> Some v | _ -> None) !Cilfacade.current_file.globals
+    |> List.filter (fun v -> match v.vtype with TPtr (TComp _, _) | TPtr ((TNamed ({ttype = TComp _; _}, _)), _) -> true | _ -> false)
```
sim642, line 29: This isn't very reliable, there may be more than one `typedef`. To do it properly, matching should be on `Cil.unrollType`.

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,105 @@ struct
...
+            List.fold_left (fun acc f ->
+                if isPointerType f.ftype then
+                  begin match ValueDomain.Structs.get s f with
+                    | Queries.VD.Address a ->
+                      let reachable_from_addr_set =
+                        List.fold_left (fun acc addr ->
+                            match addr with
+                            | Queries.AD.Addr.Addr (v, _) -> List.append acc (v :: get_reachable_mem_from_str_ptr_globals [v] ctx)
+                            | _ -> acc
+                          ) [] (Queries.AD.elements a)
```
sim642, line 62: No need for `List`, just `Queries.AD.fold`.

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,105 @@ struct
...
+        begin match a with
+          | Struct s ->
+            List.fold_left (fun acc f ->
+                if isPointerType f.ftype then
+                  begin match ValueDomain.Structs.get s f with
+                    | Queries.VD.Address a ->
+                      let reachable_from_addr_set =
+                        List.fold_left (fun acc addr ->
+                            match addr with
+                            | Queries.AD.Addr.Addr (v, _) -> List.append acc (v :: get_reachable_mem_from_str_ptr_globals [v] ctx)
```
sim642, line 60: Appending to the end of `acc` isn't very efficient. It's better to prepend new things.

This occurs elsewhere as well.

michael-schwarz:
> Maybe we should add a check that if earlyglobs is activated, then we should abort the analysis?

I don't think we use `earlyglobs` in sv-comp, so this doesn't seem too pressing.

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,105 @@ struct
...
+      match ctx.ask (Queries.EvalValue (Lval (Var heap_var, NoOffset))) with
+      | a when not (Queries.VD.is_top a) ->
+        begin match a with
+          | Struct s ->
+            List.fold_left (fun acc f ->
+                if isPointerType f.ftype then
+                  begin match ValueDomain.Structs.get s f with
+                    | Queries.VD.Address a ->
+                      let reachable_from_addr_set =
+                        List.fold_left (fun acc addr ->
```
michael-schwarz, line 58: I think we would need to check that this is singleton here once again. If it **may** point to several things, none of them are **must**-reachable.

src/analyses/memLeak.ml
mrstanb: I did some manual inspection to check for that. Iirc, `GVarDecl` was giving me a bunch of stuff from the header files as well which seemed a bit too much for me. Anyhow, I can add `GVarDecl` in there as well if we don't have a CIL guarantee regarding `GVar`.

src/analyses/memLeak.ml
mrstanb: I'll try it out, thanks for the tip!

mrstanb: I'm only not sure about the predicate formulation for `List.partition`, namely: we'd have the `true` case for the `TPtr` structs and the `false` one for the structs without a `TPtr` type. In this case what should we do with the rest of the types (otherwise, the match wouldn't be complete)?

src/analyses/memLeak.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1241/commits/6cc01b5177b6bc31899c290b6ae65660bf4aa805

src/analyses/memLeak.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1241/commits/6cc01b5177b6bc31899c290b6ae65660bf4aa805

src/analyses/memLeak.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1241/commits/6cc01b5177b6bc31899c290b6ae65660bf4aa805

src/analyses/memLeak.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1241/commits/6cc01b5177b6bc31899c290b6ae65660bf4aa805

src/analyses/memLeak.ml
mrstanb: Should be fixed in https://github.com/goblint/analyzer/pull/1241/commits/80492ccd1dcad21e49a949a989d1cee42bbb6585. I also added an `is_top` check that was missing for the address set `a`.

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,111 @@ struct
...
+        match unrollType v.vtype with
+        | TPtr (TComp _, _)
+        | TPtr ((TNamed ({ttype = TComp _; _}, _)), _) -> true
+        | TComp (_, _)
+        | (TNamed ({ttype = TComp _; _}, _)) -> false
+        | _ -> false)
+
+  let get_global_struct_non_ptr_vars () =
+    get_global_vars ()
+    |> List.filter (fun v -> match v.vtype with TComp (_, _) | (TNamed ({ttype = TComp _; _}, _)) -> true | _ -> false)
```
michael-schwarz, line 38: Probably we should also call `unrollType` here?

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,111 @@ struct
...
+          | Queries.AD.Addr.Addr (v, _) when not (ctx.ask (Queries.IsAllocVar v)) && isPointerType v.vtype -> get_reachable_mem_from_str_ptr_globals [v] ctx
+          | _ -> []
+        end
+      | _ -> []
+    in
+    global_struct_ptr_vars
+    |> List.fold_left (fun acc var ->
+        if ctx.ask (Queries.IsHeapVar var) then eval_value_of_heap_var var
+        else if not (ctx.ask (Queries.IsAllocVar var)) && isPointerType var.vtype then get_pts_of_non_heap_ptr_var var
+        else acc
```
michael-schwarz, line 92: Is it intended that we drop `acc` here in the first two cases?

src/analyses/memLeak.ml
michael-schwarz: Yes, then it is not quite a good usecase for `partition`. We could probably come up with an efficient thing that splits it into two lists and discards the rest in linear time, but I don't think this is crucial for this year's sv-comp.

src/analyses/memLeak.ml
mrstanb: Seems that I've forgotten to add it here. Should be fixed now: https://github.com/goblint/analyzer/pull/1241/commits/f2ca6d146e1271709259be24d5b20f9e61aab7dd

src/analyses/memLeak.ml
mrstanb: No, I've simply forgotten to prepend the returned `varinfo list`s in the first two cases to `acc`. Should be fixed in https://github.com/goblint/analyzer/pull/1241/commits/0e09d099d5facf7354014319805eb53e186119d2

michael-schwarz (approved):

src/analyses/memLeak.ml
```diff
@@ -20,6 +20,115 @@ struct
...
+        let reachable_from_fields =
+          List.fold_left (fun acc_field field ->
+              match ValueDomain.Structs.get s field with
+              | Queries.VD.Address a ->
+                let reachable_from_addr_set =
+                  List.fold_left (fun acc_addr addr ->
+                      match addr with
+                      | Queries.AD.Addr.Addr (v, _) -> (v :: get_reachable_mem_from_str_ptr_globals [v] ctx) @ acc_addr
+                      | _ -> acc_addr
+                    ) [] (Queries.AD.elements a)
```
sim642, line 124: This still involves an unnecessary intermediate list.

mrstanb: Sorry, should be now refactored correctly in https://github.com/goblint/analyzer/pull/1241/commits/9153eb3becd5904b99aa8250e50b2cd22f74a128

mrstanb:
> I may have missed this, but in what sense are global structs special? As in, why aren't we also considering the contents of global arrays and whatnot?
Or are we just less precise for global array contents then?

I'm not really sure about arrays, because I haven't worked a lot with the array domain of Goblint and so I can't express an opinion on that.

I've thought about global `union`s, however I'm unsure about what should happen in a situation like the following. Consider:
```c
typedef union {
    int i;
    char *str;
} un;

un *glob_var;

int main() {
    ...
}
```

What happens if the program first accesses `str` of `glob_var` and then subsequently accesses `i` of `glob_var`? Wouldn't `str` be "lost" so to speak? I mean the memory is there, but `i` is the thing that's now occupying the place of `str`, unless I'm thinking in the wrong direction.

michael-schwarz:
Yeah, unions do not seem to be a good candidate here. We discussed the issue of arrays at our Gobcon yesterday, and decided that due to the abstraction we use for arrays, it is unlikely that things are must-reachable from arrays and have thus decided against implementing dedicated support for it for now.

michael-schwarz:
I think we can now merge this into master as well? Wdyt @sim642?

PR #1246, opened 2023-11-16
Improve multi-threaded `valid-memcleanup`

mrstanb:
This PR is an attempt to improve precision for `valid-memcleanup` in multi-threaded programs

Main idea:
* Keep a global state of threads to allocated heap vars
* On `*alloc()` call:
	* Side-effect current thread and alloced heap var
* On `free()` call:
	* Remove the freed heap vars from the global state value, associated with the current thread
* On thread return/exit:
	* Check if the global state value for the returning/exiting thread is empty -> if no, then WARN

Test case `76/08` is still not working as expected, but I wanted to open the PR, so that you can also take a look at it


sim642:
> * Keep a global state of threads to allocated heap vars
> 
>  * On `*alloc()` call:
>       
>    * Side-effect current thread and alloced heap var
> 
>  * On `free()` call:
>       
>    * Remove the freed heap vars from the global state value, associated with the current thread

How can `free` remove anything from the global invariant? Side-effects just accumulate.

mrstanb:
> How can free remove anything from the global invariant? Side-effects just accumulate.

Fair point, thanks for the heads up! I think we could just keep track of allocated and freed memory as part of the global invariant in that case. That way we should be able to check for memory leaks at thread return/exit points.

michael-schwarz:
Does not suffice to just keep track locally? If one of the functions that just terminates execution altogether, we warn. Otherwise, each thread checks for itself if it did a cleanup.
If all threads cleaned up their own stuff and the program is excited in the regular manner, all should be good?

mrstanb:
So, I now tried again by removing the global state completely and using the local state instead:
* Local state is now a tuple of a map and a boolean
* The boolean is a simple flag that signifies whether `malloc` (or a similar function) has been called
* The map maps thread IDs to sets of heap variables
* The intuition is that each TID is mapped to its allocated memory
* Upon `*alloc()` calls, we add heap vars to the current thread's key in the map
* Upon `free()` calls, we remove heap vars from the current thread's key in the map
* Upon thread return/exit, we check if there's at least one value for the key of the current TID in the map -> if yes, we WARN

Note: for some reason, the check for the points-to set cardinality being one breaks the multi-threaded test cases and it fails to detect `free` calls in threads other than the main thread. I haven't looked into this issue and wanted to have it documented here as well.

P.S.: Thanks for your tips in the previous comments :)

mrstanb:
I added the two Goblint options `warn.memleak.memcleanup` and `warn.memleak.memtrack`. The idea is to be able to produce warnings only for the one or the other type of memory leak violation.

Since the `memLeak.ml` code in this PR seems to always produce warnings for both types of SV-COMP properties, I'd wait to have the PR #1241 merged and then add the checks for these properties in the `memLeak` analysis (since the other PR actually has some spots in the analysis where a warning is produced only for the one or the other type of SV-COMP property)

michael-schwarz (approved):

sim642:
> Note: for some reason, the check for the points-to set cardinality being one breaks the multi-threaded test cases and it fails to detect `free` calls in threads other than the main thread. I haven't looked into this issue and wanted to have it documented here as well.

Where exactly is this check missing and what does it mean if we don't check for singleton points-to there?

Also, I'm a bit puzzled how this can work with just local state (except for the global boolean now). That essentially means, leaks can be checked by knowing nothing about any other threads. But how is that possible without even observing what other threads do?

michael-schwarz:
> Where exactly is this check missing and what does it mean if we don't check for singleton points-to there?

I refactored it since then, this no longer applies.

michael-schwarz:
> Also, I'm a bit puzzled how this can work with just local state (except for the global boolean now). That essentially means, leaks can be checked by knowing nothing about any other threads. But how is that possible without even observing what other threads do?

An **overapproximation** can be checked in this manner, provided that the only reason for program termination is that all threads terminate. If each thread cleans up everything that it has allocated, there is no memory leak.

sim642:
> provided that the only reason for program termination is that _all threads terminate_.

But we don't check that here at all because threads don't observe each other at all.

For example, here we find no memory leak because the leaky thread never terminates:
```c
//PARAM: --set ana.malloc.unique_address_count 1 --set ana.activated[+] memLeak
#include <stdlib.h>
#include <pthread.h>

int *m1;

void *f1(void *arg) {
    m1 = malloc(sizeof(int));
    while (1);
}

    pthread_t t1;
    pthread_create(&t1, NULL, f1, NULL);
    return 0; //NOWARN
}
```

michael-schwarz:
> For example, here we find no memory leak because the leaky thread never terminates:

I don't think this is a memory leak (or at least not in the sense used in sv-comp). The program will not terminate as `f1` is an in infinite loop.

sim642:
It terminates immediately if you compile and run it: https://godbolt.org/z/3sjshTj3h.

It's not specific to the main thread returning even. Any other thread may call `exit` and the process terminates. And when it does, it doesn't wait for threads to finish.

michael-schwarz:
Crap, you're right of course. We take care of calls to `exit` but I seems I didn't consider the case of `main` implicitly calling `exit`. Actually, for main one would need some of type of all-threads-must-have-been-joined property.

michael-schwarz:
I got confused with the threading model of the JVM.

michael-schwarz:
We should add an is-single-thread query for returning from main then, that should make it sound again.

jerhard:
>We should add an is-single-thread query for returning from main then, that should make it sound again.

I will have a look at this.

jerhard:
I added a check that when the main thread reaches its return, it is definitely single-threaded in case ever any memory was allocated. 

I wondered whether a `pthread` exit in `main` should be handled the same way, but `pthread_exit` for the `main` thread behaves [differently](https://man7.org/linux/man-pages/man3/pthread_exit.3.html):

>To allow other threads to continue execution, the main thread should terminate by calling pthread_exit() rather than [exit(3)](https://man7.org/linux/man-pages/man3/exit.3.html).

I.e., if the `main` thread calls `pthread_exit`, and some other thread is running in an infinite loop, the program runs forever. If a infinitely-looping thread had allocated some memory before, this should not considered a memory leak, IMHO.

jerhard:
The analysis now requires the `thread` analysis for a precise handling in the multi-threaded case if there is any memory allocation, to be able to recover from the multithreaded mode. As the `thread` analysis is already in the `svcomp.json`, this should not be a problem, however.

jerhard:
Handling a `pthread_exit` like a  `return` in the thread analysis ([645b03cf](https://github.com/goblint/analyzer/pull/1246/commits/645b03cfa678acbf3fdc585d4ee0a7d71e8b9688)) fixes the issue with test case `76/18`.

michael-schwarz:
> Handling a pthread_exit like a return in the thread analysis ([645b03cf](https://github.com/goblint/analyzer/pull/1246/commits/645b03cfa678acbf3fdc585d4ee0a7d71e8b9688)) fixes the issue with test case 76/18 that was described in the previous post.

I looked at it, and this seems to indeed be what we want here.

tests/regression/76-memleak/08-invalid-memcleanup-multi-threaded.c
```diff
@@ -7,6 +7,7 @@ int *m1;
 
 void *f1(void *arg) {
     m1 = malloc(sizeof(int));
+    free(m1);
     // Thread t1 leaks m1 here
     pthread_exit(NULL); //WARN
```
sim642: The addition of this `free` changed the test from what it originally was.

tests/regression/76-memleak/16-no-mem-leak-thread-exit-main.c
```diff
@@ -0,0 +1,22 @@
...
+    return NULL;
+}
+
+int main(int argc, char const *argv[]) {
+    pthread_t t1;
+    pthread_create(&t1, NULL, f1, NULL);
+
+    pthread_exit(NULL);
+
+    // A pthread_join called in main will wait for other threads to finish
```
sim642, line 19: Did you mean?
```suggestion
    // A pthread_exit called in main will wait for other threads to finish
```

tests/regression/76-memleak/16-no-mem-leak-thread-exit-main.c
```diff
@@ -0,0 +1,22 @@
...
+    // Therefore, no memory leak here
+    return 0; // NOWARN
```
sim642, line 21: This is dead code, so of course it's `NOWARN`. But shouldn't the `NOWARN` be on the `pthread_exit` before?

src/analyses/threadAnalysis.ml
```diff
@@ -22,7 +22,7 @@ struct
   module P = IdentityP (D)
 
   (* transfer functions *)
-  let return ctx (exp:exp option) (f:fundec) : D.t =
+  let return ctx (exp:exp option) _ : D.t =
     let tid = ThreadId.get_current (Analyses.ask_of_ctx ctx) in
     begin match tid with
       | `Lifted tid -> ctx.sideg tid (false, TS.bot (), not (D.is_empty ctx.local))
```
sim642, line 28: Shouldn't this be conditioned on whether the `return` is actually a thread return? Currently any nested function call in a thread would make the entire thread be marked as dirtily exited.

tests/regression/76-memleak/08-invalid-memcleanup-multi-threaded.c
jerhard: I removed the call to `free` now.

src/analyses/threadAnalysis.ml
jerhard: The `ThreadAnalysis.return` now queries whether the current return is a thread-return. I also added a test case for the memleak analysis that fails without this check.

michael-schwarz:
I think we can now merge this into master as well? Wdyt @sim642?

sim642:
I'll merge it via #1257 because it already contains the merge conflict resolution between the latest versions of #1241 and #1246.

PR #1247, opened 2023-11-17
Fix region escaping in per-thread-array-init-race

sim642:
This is probably a very ad-hoc fix to one region escaping issue. In #989 something different was fixed but that doesn't handle this case at least.


michael-schwarz (approved):

michael-schwarz:
This one was the only unsoundness revealed by the TUM pre-runs.

PR #1252, opened 2023-11-17
Be more precise for `<<` of Intervals

### LINKED ISSUE #1250: Be more precise for overflow warnings for shifts
For the following program we produce a warning that we might have a signed integer over-and underflow. This seems unnecessarily imprecise.

~~~C
int main()
{
    int r;
    unsigned char c;
    r = c << 1U;
    return 0;
}
~~~

michael-schwarz:
Closes #1250

### DIFF
--- a/src/cdomains/intDomain.ml
+++ b/src/cdomains/intDomain.ml
@@ -610,6 +610,11 @@ module IntervalArith(Ints_t : IntOps.IntOps) = struct
     let x2y2 = (Ints_t.mul x2 y2) in
     (min4 x1y1 x1y2 x2y1 x2y2, max4 x1y1 x1y2 x2y1 x2y2)
 
+  let shiftleft (x1,x2) (y1,y2) =
+    let y1p = Ints_t.shift_left Ints_t.one y1 in
+    let y2p = Ints_t.shift_left Ints_t.one y2 in
+    mul (x1, x2) (y1p, y2p)
+
   let div (x1, x2) (y1, y2) =
     let x1y1n = (Ints_t.div x1 y1) in
     let x1y2n = (Ints_t.div x1 y2) in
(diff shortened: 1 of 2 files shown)


michael-schwarz:
This also means we succeed on the relevant SV-COMP task.

michael-schwarz:
Looks .like there's a few more SV-COMP No-Overflow tasks where this helps :)

michael-schwarz:
Fails on some LDV tasks e.g.

>./goblint --conf conf/svcomp24.json --sets ana.specification /home/goblint2/svcomp/sv-benchmarks/c/properties/unreach-call.prp --sets exp.architecture 64bit /home/goblint2/svcomp/sv-benchmarks/c/ldv-linux-3.0/usb_urb-drivers-video-arkfb.ko.cil.out.i

with

> Fatal error: exception Invalid_argument("Z.shift_left: count argument must be positive")

src/cdomains/intDomain.ml
```diff
@@ -610,6 +610,11 @@ module IntervalArith(Ints_t : IntOps.IntOps) = struct
     let x2y2 = (Ints_t.mul x2 y2) in
     (min4 x1y1 x1y2 x2y1 x2y2, max4 x1y1 x1y2 x2y1 x2y2)
 
+  let shiftleft (x1,x2) (y1,y2) =
```
sim642, line 613: Everywhere else this is named `shift_left`.

michael-schwarz:
Since this has an overall positive effect, I suggest to merge.

tests/regression/39-signed-overflows/06-shiftleft.c
```diff
@@ -0,0 +1,26 @@
...
+
+    if(top) { zero_or_one = 1; }
+
+    r = 8 << zero_or_one;
+
+    __goblint_check(r >= 8);
+    __goblint_check(r <= 16);
+
+    int regval;
+    unsigned long bla = (unsigned long )((1 << ((int )regval >> 6)) << 20);  //WARN
```
jerhard, line 23: Doesn't the subexpression `(1 << ((int )regval >> 6)` already overflow potentially? If so, one may add a line for that.

sim642 (approved):

tests/regression/39-signed-overflows/06-shiftleft.c
sim642: Due to #1258, this might only be warning due to the cast right now?

jerhard: The subexpression actually already potentially overflows and Goblint detects it; I extended the test case accordingly.

jerhard (approved):

PR #1254, opened 2023-11-18
Handle `sqrt` & Some Bodged Solution for computing through `abs`

michael-schwarz:
This contains @stilscher's useful changes for computing `sqrt`s for floats, plus some bodged together solution for refining through `abs` library functions we need for SV-COMP.

I propose to merge this for now and then replace it with @stilscher's cleaner solution once she finishes it.

References #1244 (and duplicates the working part of #1253).


tests/regression/39-signed-overflows/06-abs.c
```diff
@@ -0,0 +1,22 @@
+//PARAM: --enable ana.int.interval --set ana.activated[+] tmpSpecial
```
sim642, line 1: For this to be of any use, we should activate `tmpSpecial` in our svcomp conf?

src/analyses/libraryFunctions.ml
```diff
@@ -130,7 +130,7 @@ let c_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
     ("wcstombs", unknown ~attrs:[ThreadUnsafe] [drop "dst" [w]; drop "src" [r]; drop "size" []]);
     ("wcsrtombs", unknown ~attrs:[ThreadUnsafe] [drop "dst" [w]; drop "src" [r_deep; w]; drop "size" []; drop "ps" [r_deep; w_deep]]);
     ("mbstowcs", unknown [drop "dest" [w]; drop "src" [r]; drop "n" []]);
-    ("abs", unknown [drop "j" []]);
+    ("abs", special [__ "j" []] @@ fun j -> Math { fun_args = (Abs j) });
```
sim642, line 133: Could we also support `labs` and `llabs`?

And what about the floating-point `fabs` (and `fabsf` and `fabsl`)?

src/analyses/base.ml
```diff
@@ -2328,6 +2353,28 @@ struct
...
+      let apply_abs ik x =
+        let eval_x = eval_rv (Analyses.ask_of_ctx ctx) gs st x in
+        begin match eval_x with
+          | Int int_x ->
+            let xcast = ID.cast_to ik int_x in
+            (* the absolute value of the most-negative value is out of range for 2'complement types *)
+            (match (ID.minimal xcast), (ID.minimal (ID.top_of ik)) with
+             | _, None
+             | None, _ -> ID.top_of ik
+             | Some mx, Some mm when Z.equal mx mm -> ID.top_of ik
```
sim642, line 2365: I think this should also check the signedness because for unsigned types the minimal value is `0`, which is completely fine.

Also, this seems a bit roundabout way to do `ID.equal_to`. That would be better because it could benefit from the non-interval domain as well. For example, if it cannot be `INT_MIN` by congruence.

src/analyses/base.ml
```diff
@@ -2328,6 +2353,28 @@ struct
...
+             | _, _ ->
+               match ID.le xcast (ID.of_int ik Z.zero) with
+               | d when d = ID.of_int ik Z.zero -> xcast (* x positive *)
+               | d when d = ID.of_int ik Z.one -> ID.neg xcast (* x negative *)
+               | _ -> (* both possible *)
+                 let x1 = ID.neg (ID.meet (ID.ending ik Z.zero) xcast) in
+                 let x2 = ID.meet (ID.starting ik Z.zero) xcast in
+                 ID.join x1 x2
+            )
```
sim642, line 2374: This seems overly complex. Doesn't the general case already cover the special cases?

src/analyses/base.ml
```diff
@@ -1753,15 +1753,40 @@ struct
...
+        (match ctx.ask (Queries.TmpSpecial (v, Offset.Exp.of_cil NoOffset)) with 
+        | `Lifted (Abs arg) -> 
+          refineAbs op (CastE (t, arg)) e
+        | _ -> refine0)
+      | BinOp ((Lt|Le) as op, Lval (Var v, NoOffset), e, _) when tv ->
+        (match ctx.ask (Queries.TmpSpecial (v, Offset.Exp.of_cil NoOffset)) with 
+        | `Lifted (Abs arg) -> 
+          refineAbs op arg e
+        | _ -> refine0)
+      | _ -> refine0
```
sim642, line 1789: Why is this here, not in `BaseInvariant`? The whole point is to not clutter this already-massive file. `BaseInvariant` has access to `ctx`, so there should be nothing preventing from doing it there.

src/analyses/base.ml
```diff
@@ -1753,15 +1753,40 @@ struct
...
+      let refine0 =
+        let res = invariant ctx (Analyses.ask_of_ctx ctx) ctx.global ctx.local exp tv in
+        if M.tracing then M.tracec "branch" "EqualSet result for expression %a is %a\n" d_exp exp Queries.ES.pretty (ctx.ask (Queries.EqualSet exp));
+        if M.tracing then M.tracec "branch" "CondVars result for expression %a is %a\n" d_exp exp Queries.ES.pretty (ctx.ask (Queries.CondVars exp));
+        if M.tracing then M.traceu "branch" "Invariant enforced!\n";
+        match ctx.ask (Queries.CondVars exp) with
+        | s when Queries.ES.cardinal s = 1 ->
+          let e = Queries.ES.choose s in
+          invariant ctx (Analyses.ask_of_ctx ctx) ctx.global res e tv
+        | _ -> res
```
sim642, line 1765: Since this is a value, it's evaluated even when we're handling the `abs` special cases. Isn't that excessive?

tests/regression/39-signed-overflows/06-abs.c
stilscher: Yes, I think that would need to be activated.

michael-schwarz: Julian has done so in the https://github.com/goblint/analyzer/tree/benchmark_sv_comp branch.

src/analyses/libraryFunctions.ml
stilscher: `fabs` (and `fabsf` and `fabsl`) are specified as math library functions already and are also handled by the float domain. There is just no refinement for their parameters yet. Support for `labs` and `llabs` still needs to be added.

src/analyses/base.ml
michael-schwarz: Sure, but we could not get it work reliably when integrated into `BaseInvariant` (#1253). That's why it's a bodge

src/analyses/libraryFunctions.ml
sim642: My bad `fabs` is indeed there. `labs` and `llabs` should just be matter of adding them here I suppose.

src/analyses/base.ml
sim642: I just moved it there as-is to apply at the top level only, like here. I'll have to see what the issue with doing it at all levels in #1253 is.

src/analyses/base.ml
sim642: `refine0` is used in all cases, even in `refineAbs`, so this isn't immediately excessive at least. Harder to say whether it's of any use in the `abs` case though, but we can keep it.

src/analyses/base.ml
stilscher: You are right, the general case should be sufficient. I removed the other two.

src/analyses/libraryFunctions.ml
stilscher: Do we need to consider the different types for the evaluation as we do for the floating-point versions?

src/analyses/base.ml
michael-schwarz: Can the type ever be unsigned here? I thought it is cast to `int` by CIL?

sim642: Actually, this function is currently only called with `IInt` anyway.

src/analyses/libraryFunctions.ml
sim642: I guess yes, because currently in `base` it uses `IInt` explicitly.

src/analyses/base.ml
jerhard: Now with @stilscher's changes the function may be called with `ILong` and `ILongLong` as well.

stilscher: `ILong` and `ILongLong` should be signed as well, right? So that should not make a difference.

src/analyses/libraryFunctions.ml
jerhard: There are now versions for `labs` and  `llabs`, with the corresponding `ikind`.

src/analyses/base.ml
jerhard: Yes, I think one then does not need to handle the case of having an unsigned type, if all variations of `abs` functions take signed types only.

src/common/util/options.schema.json
```diff
@@ -544,7 +544,7 @@
               "type": "array",
               "items": { "type": "string" },
               "default": [
-                "congruence", "singleThreaded", "specification", "mallocWrappers", "noRecursiveIntervals", "enums", "loopUnrollHeuristic", "arrayDomain", "octagon", "wideningThresholds", "memsafetySpecification"
+                "congruence", "singleThreaded", "specification", "mallocWrappers", "noRecursiveIntervals", "enums", "loopUnrollHeuristic", "arrayDomain", "octagon", "wideningThresholds", "memsafetySpecification", "tmpSpecialAnalysis"
```
sim642, line 547: This should also be added to svcomp conf, if we're planning on using it there.

tests/regression/39-signed-overflows/06-abs.c
jerhard: Yes, there I added `tmpSpecial` it to the activated analyses in `svcomp24.json`. Since @stilscher added an autotuner, having `tmpSpecialAnalysis` in the autotuner setting should be preferable. 
Here, I therefore added `tmpSpecialAnalysis` to `ana.autotune.activated`.

src/common/util/options.schema.json
jerhard: I added `tmpSpecialAnalysis` to `ana.autotune.activated` in the `svcomp.json`.

sim642 (approved):
Conflicts need to be resolved, but otherwise I suppose this seems ready.

stilscher:
Based on the latest run, I wanted to take another look at some test cases that use `imaxabs`. We cannot handle them yet, but simply adding `imaxabs` to the library functions does also not seem to work since there is a debug warning about an invariant failing.

sim642:
Since `intmax_t` is determined by some headers, maybe we cannot so statically know its ikind (if we can at all). It may be different from `long long`: https://stackoverflow.com/questions/20459513/is-intmax-t-the-same-as-long-long-int.

PR #1255, opened 2023-11-19
Autotune termination spec before preprocessing, but not others.

jerhard:
Via the autotuner, in case the specification is Termination, activate Termination analysis before preprocessing, but do not activate other analysis in case the spec is different.

### DIFF
--- a/src/goblint.ml
+++ b/src/goblint.ml
@@ -39,7 +39,7 @@ let main () =
       print_endline GobSys.command_line;
     );
     (* When analyzing a termination specification, activate the termination analysis before pre-processing. *)
-    if get_bool "ana.autotune.enabled" && AutoTune.specificationIsActivated () then AutoTune.focusOnTermination ();
+    if get_bool "ana.autotune.enabled" && AutoTune.specificationTerminationIsActivated () then AutoTune.focusOnTermination ();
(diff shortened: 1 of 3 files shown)


src/goblint.ml
```diff
@@ -39,7 +39,7 @@ let main () =
       print_endline GobSys.command_line;
     );
     (* When analyzing a termination specification, activate the termination analysis before pre-processing. *)
-    if get_bool "ana.autotune.enabled" && AutoTune.specificationIsActivated () then AutoTune.focusOnSpecification ();
+    if get_bool "ana.autotune.enabled" && AutoTune.specificationIsActivated () then AutoTune.focusOnTermination ();
```
michael-schwarz, line 42: Should `Autotune.specificationIsActivated ()` be replaced with some sort of `Autotune.specificationTerminationIsActivated ()`?

sim642:
The early activation of the termination autotuner makes me wonder: how does it interact with loop unrolling? Both manipulate the AST.

jerhard:
> how does it interact with loop unrolling? Both manipulate the AST.

The termination preprocessing is done before the loop unrolling. The loop counter is incremented and bounds checked also in the unrolled loops. Running the termination regression tests with loop-unrolling (factor 2) does not yield any errors.

Issue #1243, opened 2023-11-08
Add support for the detection of `CWE-562`


mrstanb:
[CWE-562](https://cwe.mitre.org/data/definitions/562.html) is concerned with the issue of returning an address of a stack variable.

Here's a minimal example of a program that has a `CWE-562` weakness:
```c
char *cwe562() {
  char *str;
  return str;
}
```

I'm not sure if there are any SV-COMP Memory Safety cases for this CWE number (most likely not, at least from what I recall). Nonetheless, I think it'd be useful to have a detection mechanism in Goblint for this type of weakness. In addition, it could potentially be even brought into SV-COMP's Memory Safety category if it'd make sense.

I'll make sure to open a separate PR for this soon.

sim642:
In terms of SV-COMP, I think the act of returning such address wouldn't be considered a violation, but if it is then dereferenced, it would be an invalid dereference.

mrstanb:
> In terms of SV-COMP, I think the act of returning such address wouldn't be considered a violation, but if it is then dereferenced, it would be an invalid dereference.

Yes, absolutely. I think in relation to this, we could try to flag such returned memory in some way, so that whenever it's used we could directly flag the use as an invalid one. (Although, thinking about it, we might not need an extra flag, because we could just check if the memory has an address that is not pointing to something local or global and be done with it)

jerhard:
Following a dicussion at Gobcon, I closed the PR #1256. We might look again at implementing this when having a look at the escaping of variables, for this see https://github.com/goblint/analyzer/issues/1544, https://github.com/goblint/analyzer/issues/1491.

PR #1256, opened 2023-11-20
Add support for CWE-562

mrstanb:
This PR is a first quick draft of adding support for the detection of CWE-562. It's still WIP and since it's probably not directly relevant for SV-COMP, I'd say it has a lower prio.

There are a few things here and there that I need to fix before it's done.

Should close issue #1243 upon completion.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1820,6 +1820,41 @@ struct
       match exp with
       | None -> nst
       | Some exp ->
+        let is_globally_reachable_heap_var heap_var =
+          let global_vars =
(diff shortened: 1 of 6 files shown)


jerhard:
Close this. We might look again at implementing this when we have a look at related issues on escaping of variables see #1544, #1491.

mrstanb:
> Close this. We might look again at implementing this when we have a look at related issues on escaping of variables see #1544, #1491.

Thanks for the heads up! It seems that I've overlooked this PR and forgotten about it after working on the other PRs for SV-COMP back in November. Could you please let me know once a new PR should be opened for CWE-562, as I'd like to give it a go myself again, if possible :)
Thanks!

PR #1258, opened 2023-11-20
Do not set SV-COMP overflow flag on casts

michael-schwarz:
This restores the previous behavior on casts, and hopefully gives us ~250 successful no-overflow cases back that we lost due to some changes we made since last year.

### DIFF
--- a/tests/regression/29-svcomp/32-no-ov.c
+++ b/tests/regression/29-svcomp/32-no-ov.c
@@ -1,0 +1,7 @@
+// PARAM: --enable ana.int.interval --enable ana.sv-comp.enabled --enable ana.sv-comp.functions --set ana.specification "CHECK( init(main()), LTL(G ! overflow) )"
+
+int main(){
+	// This is not an overflow, just implementation defined behavior on a cast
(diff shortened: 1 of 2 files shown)


michael-schwarz:
A run of this is currently running on server2.

tests/regression/29-svcomp/32-no-ov.c
sim642, line 1: As a normal regression test, this isn't checking anything automatically. Adding a cram test would make it possible to check the SV-COMP result.

jerhard: I added a cram test that checks whether the sv-comp verdict  is `true`.

michael-schwarz:
We now get `33807` correct verdicts, which is strictly better than the `2937` we had before.

sim642:
> We now get `33807` correct verdicts, which is strictly better than the `2937` we had before.

Do you mean 3807?

jerhard:
Yes, it's 3807.

sim642 (approved):
If our prerun indicates we're still all sound, then this should be good to go.

PR #1261, opened 2023-11-21
Fix no-overflow cram test expected output

jerhard:
Fixes the expected output of a test that for sv-comp property no-overflow.


michael-schwarz (approved):

jerhard:
Done in [9c808c9](https://github.com/goblint/analyzer/commit/9c808c98eb55baf8e0f5b4839ca5cd99ae68747b)

PR #1264, opened 2023-11-21
Do not invalidate special function lvals recursively

karoliineh:


src/analyses/base.ml
```diff
@@ -2328,7 +2328,7 @@ struct
           | _ -> failwith ("non-floating-point argument in call to function "^f.vname)
         end
       in
-      let apply_abs ik x =
+let apply_abs ik x =
```
sim642, line 2331: Something happened with the indentation here.

michael-schwarz (approved):

sim642 (approved):

PR #1265, opened 2023-11-22
Test for `mutex-meet-tid` for ValidDeref & Move `Afterconfig.run` to ensure privatizations can potentially be auto-tuned

michael-schwarz:
For tasks such as the example one, we fail to establish Valid-Deref for the reason that our information about globals is too imprecise. With this PR we enable a more involved privatization when checking ValidDeref.

I am currently running this on the server with a reduced timeout of 4mins to see the potential benefits.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1056,7 +1056,13 @@ struct
             );
             (* Warn if any of the addresses contains a non-local and non-global variable *)
             if AD.exists (function
-                | AD.Addr.Addr (v, _) -> not (CPA.mem v st.cpa) && not (is_global a v)
+                | AD.Addr.Addr (v, _) ->
+                  (M.tracel "wtf" "checking for %a\n" CilType.Varinfo.pretty v;
+                   if v.vglob then
+                     (* this is OK *)
+                     false
+                   else
+                     (not (CPA.mem v st.cpa)) || WeakUpdates.mem v st.weak)
                 | _ -> false
               ) adr then (
               AnalysisStateUtil.set_mem_safety_flag InvalidDeref;
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1056,13 +1056,7 @@ struct
             );
             (* Warn if any of the addresses contains a non-local and non-global variable *)
             if AD.exists (function
-                | AD.Addr.Addr (v, _) ->
-                  (M.tracel "wtf" "checking for %a\n" CilType.Varinfo.pretty v;
-                   if v.vglob then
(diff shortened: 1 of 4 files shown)


michael-schwarz:
For some reason, the setting is not applied correctly.

> ./goblint --conf conf/svcomp.json --sets ana.specification /home/michael/Documents/sv-comp/sv-benchmarks/c/properties/valid-memsafety.prp --sets exp.architecture 32bit 2.c

fails

but

> ./goblint --conf conf/svcomp.json --sets ana.specification /home/michael/Documents/sv-comp/sv-benchmarks/c/properties/valid-memsafety.prp --sets exp.architecture 32bit 2.c --set ana.base.privatization mutex-meet-tid

works.

michael-schwarz:
It seems the autotuner needs to run earlier.

src/analyses/base.ml
```diff
@@ -1056,7 +1056,13 @@ struct
             );
             (* Warn if any of the addresses contains a non-local and non-global variable *)
             if AD.exists (function
-                | AD.Addr.Addr (v, _) -> not (CPA.mem v st.cpa) && not (is_global a v)
+                | AD.Addr.Addr (v, _) ->
+                  (M.tracel "wtf" "checking for %a\n" CilType.Varinfo.pretty v;
```
jerhard, line 1060: This tracing may be removed :)

michael-schwarz: Thanks! Sorry about that!

michael-schwarz:
Enabling the option does not yield a large improvement, and takes a long time. I'll thus remove the activation from this PR.

[diff.zip](https://github.com/goblint/analyzer/files/13439137/diff.zip)

tests/regression/74-invalid_deref/31-multithreaded.c
```diff
@@ -0,0 +1,21 @@
+//PARAM: --set ana.activated[+] useAfterFree --set ana.activated[+] threadJoins --set ana.path_sens[+] threadflag  --set ana.activated[+] memOutOfBounds --enable ana.int.interval --set ana.base.arrays.domain partitioned --set ana.base.privatization mutex-meet-tid
```
sim642, line 1: Is all of this necessary? The program doesn't contain any arrays for example.

Also, what is being tested? That Goblint doesn't crash?

michael-schwarz: That no warning is produced on line 18. I added the corresponding annotation.

sim642 (approved):

PR #1270, opened 2023-11-23
Fix ordering of queries by deduplicating indices.

jerhard:
Due to merging, there were some types of queries had the same integer index. This PR deduplicates the indices of these query types.

### DIFF
--- a/src/domains/queries.ml
+++ b/src/domains/queries.ml
@@ -331,8 +331,8 @@ struct
     | Any (MustTermLoop _) -> 53
     | Any MustTermAllLoops -> 54
     | Any IsEverMultiThreaded -> 55
-    | Any (TmpSpecial _) -> 53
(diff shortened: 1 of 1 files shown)


sim642 (approved):

michael-schwarz:
It seems error-prone to do this ourselves: Could we maybe have a deriver for this, similar to what `enum` does here: https://github.com/ocaml-ppx/ppx_deriving

sim642:
This `order` is the smaller of the worries there though. Duplicate indices just mean less short-circuiting in `compare`/`equal` and more collisions in `hash`.
The bigger problem is `compare` itself, where missing cases can cause real problems. The syntactic nature of PPXs means that it isn't really possible to derive `compare` for this type because the matching must happen on both type constructors (`Any` and the query) simultaneously but they're defined in different type declarations (because that's how constructors can only be defined).

PR #1271, opened 2023-11-23
Fix `Not_found` exception in autotuner with congruences and termination.

jerhard:
Previously, there a `Not_found` exception was raised by the autotuner when it tried to activate the `congruence` domain for the `__goblint_bounded` function used in the termination analysis, as the `fundec` for this function could not be found. 

This PR fixes this, by catching the exception, and not trying to enable the `congruence` domain for functions for which there are no declarations.


michael-schwarz:
I started a run (termination only) of this on `server02`.

michael-schwarz (approved):

michael-schwarz:
With this PR we now succeed on 618 termination tasks. It should have no influence on other tasks.

PR #1272, opened 2023-11-23
Improve trigonometric abstractions

stilscher:
This PR improves the abstsraction of the `cos` function if the argument is neither nan nor inf by

- storing also the inverted relationship between arguments of math functions and (assigned variable, math function) in the `tmpSpecial` analysis
- being more precise in the abstraction of `cos(x)` in case one can assure that variables representing `isInf(x)` and `isNaN` are definitely false


michael-schwarz:
This yields 4 more successful tasks, I'll start it on the server.

src/cdomains/floatDomain.ml
```diff
@@ -907,8 +929,16 @@ module FloatIntervalImplLifted = struct
   let acos = lift (F1.acos, F2.acos)
   let asin = lift (F1.asin, F2.asin)
   let atan = lift (F1.atan, F2.atan)
-  let cos = lift (F1.cos, F2.cos)
-  let sin = lift (F1.sin, F2.sin)
+  let cos ?(asPreciseAsConcrete=BoolDomain.MustBool.top ()) ?(notInf_notNaN=BoolDomain.MustBool.top ()) = function
```
michael-schwarz, line 932:
```suggestion
  let cos ?(notInf_notNaN=false) = function
```

`asPreciseAsConcrete` is unused, and `BoolDomain.MustBool.top ()` is `false`.

src/cdomains/floatDomain.ml
```diff
@@ -907,8 +929,16 @@ module FloatIntervalImplLifted = struct
...
+    | F32 a -> F32 (F1.cos ~asPreciseAsConcrete:true ~notInf_notNaN a)
+    | F64 a -> F64 (F2.cos ~asPreciseAsConcrete:true ~notInf_notNaN a)
+    | FLong a -> FLong (F2.cos ~notInf_notNaN a)
+    | FFloat128 a -> FFloat128 (F2.cos ~notInf_notNaN a)
+  let sin ?(asPreciseAsConcrete=BoolDomain.MustBool.top ()) ?(notInf_notNaN=BoolDomain.MustBool.top ()) = function
```
michael-schwarz, line 937:
```suggestion
  let sin ?(notInf_notNaN=true) = function
```

src/cdomains/floatDomain.ml
```diff
@@ -1165,10 +1195,10 @@ module FloatDomTupleImpl = struct
     map { f1= (fun (type a) (module F : FloatDomain with type t = a) -> F.asin); }
   let atan =
     map { f1= (fun (type a) (module F : FloatDomain with type t = a) -> F.atan); }
-  let cos =
-    map { f1= (fun (type a) (module F : FloatDomain with type t = a) -> F.cos); }
-  let sin =
-    map { f1= (fun (type a) (module F : FloatDomain with type t = a) -> F.sin); }
+  let cos ?(asPreciseAsConcrete=BoolDomain.MustBool.top ()) ?(notInf_notNaN=BoolDomain.MustBool.top ()) =
```
michael-schwarz, line 1198:
```suggestion
  let cos ?(asPreciseAsConcrete=false) ?(notInf_notNaN=false) =
```

src/cdomains/floatDomain.ml
```diff
@@ -1165,10 +1195,10 @@ module FloatDomTupleImpl = struct
...
+    map { f1= (fun (type a) (module F : FloatDomain with type t = a) -> F.cos ~asPreciseAsConcrete ~notInf_notNaN); }
+  let sin ?(asPreciseAsConcrete=BoolDomain.MustBool.top ()) ?(notInf_notNaN=BoolDomain.MustBool.top ()) =
```
michael-schwarz, line 1200:
```suggestion
  let sin ?(asPreciseAsConcrete=false) ?(notInf_notNaN=false) =
```

src/cdomains/floatDomain.ml
```diff
@@ -791,8 +791,30 @@ module FloatIntervalImpl(Float_t : CFloatType) = struct
   let acos = eval_unop eval_acos
   let asin = eval_unop eval_asin
   let atan = eval_unop eval_atan
-  let cos = eval_unop eval_cos
-  let sin = eval_unop eval_sin
+  let cos ?(asPreciseAsConcrete=false) ?(notInf_notNaN=false) op =
```
michael-schwarz, line 794: Seems like we should be able to get some reuse here for `sin` and `cos`.

src/analyses/base.ml
```diff
@@ -2355,6 +2355,40 @@ struct
...
+                          | v when Queries.ID.is_bot v -> false
+                          | v ->
+                            match Queries.ID.to_bool v with
+                            | Some b -> not b
+                            | None -> false)
+                      | _ -> false
+                    in
+                    let isNotNan = Q.MLInv.exists (fun (y, ml) -> matchnan ml && isFalse y) s in
+                    let isNotInf = Q.MLInv.exists (fun (y, ml) -> matchinf ml && isFalse y) s in
+                    isNotInf && isNotNan)
```
michael-schwarz, line 2385: Might make sense to restructure so the lookup for `isNotInf` doesn't always happen.

src/analyses/base.ml
```diff
@@ -2355,6 +2355,40 @@ struct
...
+                      | `Lifted LibraryDesc.Isnan _ -> true
+                      | _ -> false in
+                    let isFalse = function
+                      | `Lifted expy -> (
+                          match ctx.ask (Q.EvalInt expy) with
+                          | v when Queries.ID.is_bot v -> false
+                          | v ->
+                            match Queries.ID.to_bool v with
+                            | Some b -> not b
+                            | None -> false)
```
michael-schwarz, line 2380:
```suggestion
                            Queries.ID.to_ bool v = Some false)
```

src/analyses/base.ml
```diff
@@ -2355,6 +2355,40 @@ struct
           | _ -> failwith ("non-integer argument in call to function "^f.vname)
         end
       in
+      let apply_better_trig fk (float_fun : ?asPreciseAsConcrete:bool -> ?notInf_notNaN:bool -> FD.t -> FD.t) x =
```
michael-schwarz, line 2358: Is the type annotation needed, it looks a bit clunky.

src/analyses/tmpSpecial.ml
```diff
@@ -70,8 +86,30 @@ struct
         (* actually it would be necessary to check here, if one of the arguments is written by the call. However this is not the case for any of the math functions and no other functions are covered so far *)
         if List.exists (fun arg -> VarEq.may_change ask (mkAddrOf lv) arg) arglist then
           d
-        else
-          D.add (v, Offset.Exp.of_cil offs) ((ML.lift fun_args, Deps.of_list ((Lval lv)::arglist))) d
+        else (
+          if M.tracing then M.trace "tmpSpecialInv" "build tmpSpecial Entry for Lval %s\n" v.vname;
+          let d1 = MM.add (v, Offset.Exp.of_cil offs) ((ML.lift fun_args, Deps.of_list ((Lval lv)::arglist))) (fst d) in
```
michael-schwarz, line 91: I think some parens are unneeded here...

src/analyses/tmpSpecial.ml
```diff
@@ -82,9 +120,22 @@ struct
...
+        if M.tracing then M.trace "tmpSpecialInv" "MInv.find lv: %s\n" (fst lv).vname;
+        match MInv.find lv (snd ctx.local) with
+        | `Top -> Queries.Result.top q
+        | es ->
+          let mlv = MInvEntrySet.fold (fun (v, ml, deps) acc -> Queries.MLInv.add (v,ml) acc) es (Queries.MLInv.empty ())
+                    |> Queries.MLInv.filter (fun (v,ml) ->
+                        match ml, v with
+                        | `Bot, _ -> false
+                        | _, `Lifted exp -> true
+                        | _ -> false) in
```
michael-schwarz, line 136: This way of doing looks as if it may be in a figure of speech potentially trying to achieve the goal in a way that might be described as roundabout. You might want to do the filtering right in place. :wink:

src/analyses/tmpSpecial.ml
```diff
@@ -13,13 +13,29 @@ struct
...
+  module Deps = SetDomain.ToppedSet (CilType.Exp) (struct let topname = "All" end)
   module MLDeps = Lattice.Prod (ML) (Deps)
-  module D = MapDomain.MapBot (Mval.Exp) (MLDeps)
+
+  (* x maps to (mathf(arg), deps) arg if x = mathf(arg) *)
+  module MM = MapDomain.MapBot_LiftTop (Mval.Exp) (MLDeps)
+
+  (* arg maps to (x, mathf(arg), deps) if x = mathf(arg) *)
+  module ExpFlat = Lattice.Flat (CilType.Exp) (Printable.DefaultNames)
+  module MInvEntry = Lattice.Prod3 (ExpFlat) (ML) (Deps)
```
michael-schwarz, line 24: Have you tried this type:

~~~OCaml
module MInvEntry = Printable.Prod3 (CilType.Exp) (LibraryDesc.MathPrintable) (SetDomain.Make(CilType.Exp))
~~~

Might make it easier on the inside to give up on the lattice structure that is unused anyway.

src/analyses/tmpSpecial.ml
michael-schwarz: Also `Deps.of_list ((Lval lv)::arglist))` is constructed twice here.

sim642:
Since this is still in need of refactoring/cleanup, it's too late to include for SV-COMP 2024. Four additional tasks out of ~32000 is not worth the risk for a possible regression.

sim642:
It's not just about SV-COMP though. It will be _the_ representative version of Goblint for the entire year to come. If something doesn't yet meet our own standards, then it's not ready for such release. This is just too last minute, considering the freezing deadline we set for ourselves.

michael-schwarz:
Ok. I think it would've been nice for Sarah's work this past week to be reflected in that version, but I won't push too hard.

sim642:
Why is the `TmpSpecialInv` query needed? `TmpSpecial` itself already is a certain inverse (for refining the argument based on the return value), so `TmpSpecialInv` is some kind of inverse of the inverse. Is that not just normal forward evaluation? The new query is also used in `eval` not `invariant`.

src/analyses/base.ml
sim642: Possibly, because the type inference of optional arguments is not principal: https://dev.realworldocaml.org/variables-and-functions.html#inference-of-labeled-and-optional-arguments.

src/analyses/base.ml
```diff
@@ -2355,6 +2355,40 @@ struct
...
+            let notInf_notNaN = match x with
+              | Lval (Var xvar, offs) -> (
+                  match ctx.ask (Q.TmpSpecialInv (xvar, Offset.Exp.of_cil offs)) with
+                  | s when Q.MLInv.is_top s -> false
+                  | s ->
+                    let matchnan = function
+                      | `Lifted LibraryDesc.Isnan _ -> true
+                      | _ -> false in
+                    let matchinf = function
+                      | `Lifted LibraryDesc.Isnan _ -> true
```
sim642, line 2371:
```suggestion
                    let matchinf = function
                      | `Lifted LibraryDesc.Isinf _ -> true
```

stilscher:
As this PR only shows improvements on four SV-Comp tasks but introduces quite some additional code, we will close this PR.

PR #1274, opened 2023-11-27
Generalize and clean up `abs` invariant in base analysis

sim642:
This generalizes the `abs` handling from #1253 and #1254 to happen at any level of the expression, just like other tmpSpecials. This change is in the first commit.

The rest are some refactoring of the surrounding code because the indentation is already 14 tabs deep. I wanted to clean it up more, but the implementation of `__builtin_isgreater`, etc is implemented hackily to convert `Not{0}` to `1` to `1.0` to represent it as the expected value for float domains.

### DIFF
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -709,35 +709,31 @@ struct
                 | _ -> Int c
               in
               (* handle special calls *)
-              begin match t with
-                | TInt (ik, _) ->
-                  begin match x with
-                    | ((Var v), offs) ->
-                      if M.tracing then M.trace "invSpecial" "qry Result: %a\n" Queries.ML.pretty (ctx.ask (Queries.TmpSpecial (v, Offset.Exp.of_cil offs)));
-                      begin match ctx.ask (Queries.TmpSpecial (v, Offset.Exp.of_cil offs)) with
-                        | `Lifted (Abs (_ik, xInt)) ->
-                          inv_exp (Int (ID.join c (ID.neg c))) xInt st (* TODO: deduplicate *)
-                        | tmpSpecial ->
-                          let tv_opt = ID.to_bool c in (* TODO: simplify *)
-                          begin match tv_opt with
-                            | Some tv ->
-                              begin match tmpSpecial with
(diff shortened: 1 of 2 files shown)


michael-schwarz:
If it really is a generalization, it would be nice to have a new regression test that passes now.

src/analyses/baseInvariant.ml
```diff
@@ -744,22 +747,19 @@ struct
...
+                  let tmpSpecial = ctx.ask (Queries.TmpSpecial (v, Offset.Exp.of_cil offs)) in
+                  if M.tracing then M.trace "invSpecial" "qry Result: %a\n" Queries.ML.pretty tmpSpecial;
+                  begin match tmpSpecial with
+                    | `Lifted (Ceil (ret_fk, xFloat)) -> inv_exp (Float (FD.inv_ceil (FD.cast_to ret_fk c))) xFloat st
+                    | `Lifted (Floor (ret_fk, xFloat)) -> inv_exp (Float (FD.inv_floor (FD.cast_to ret_fk c))) xFloat st
+                    | `Lifted (Fabs (ret_fk, xFloat)) ->
+                      let inv = FD.inv_fabs (FD.cast_to ret_fk c) in
+                      if FD.is_bot inv then
+                        raise Analyses.Deadcode
+                      else
```
michael-schwarz, line 761: I know this is not new, but while we're looking at it: Why is this case distinction needed here, but no case distinction is required for any of the other float cases? (or the int case for that matter?)

michael-schwarz (approved):
Once my comment above is addressed, I'd be happy to merge this.

src/analyses/baseInvariant.ml
sim642: I'm not entirely sure why it's that way for `fabs`. Maybe because that delegates to `FD.inv_fabs` which indicates some contradiction using a bottom (because domains themselves shouldn't raise `Deadcode`). For `abs` there's no delegation to the domain, but just a nested `inv_exp` call where contradictions raise this already.

michael-schwarz: Hmm... Even then, it is strange why this is then not not needed for `Ceil` or `Floor`...

PR #1277, opened 2023-11-29
More precise abstractions of trigonometric functions using c-stubs

FelixKrayer:
These changes allow for a more precise abstraction of the trigonometric functions on monotonic intervals.
 Similar to #1253, c-stubs are used for the calculation of the resulting interval, once an interval is identified to be monotonic.
The option to enable/disable this option can be removed, if we trust, that the implementation of the trigonometric functions used in the analyzer is the same as the one used in the program to be analyzed.

You can find a rough explanation in this PDF: 
[Explanation_draft.pdf](https://github.com/goblint/analyzer/files/13505756/Explanation_draft.pdf)

- [x] add more detailed explanation


FelixKrayer:
I have added a first draft of an explanation that hopefully helps to understand the computations in the floatDomain.

stilscher:
Thank you for this addition! I took a look at the conceptual description and think it is a nice idea. Since I am quite busy right now, I might have to postpone the review for another two weeks.

stilscher (commented):
I had a few questions regarding the rounding modes. Apart from that I think it is a nice and generic idea to get a good abstraction for trigonometric computations in case the interval of the input variable is not too imprecise. I think a few more comments with some selected information from the pdf draft could improve the clarity of the code.

FelixKrayer:
I addressed the comments, fixed some rounding modes and added some comments to the code.
Do I have to worry about the Semgrep findings `semgrep.big_int_z` ? I am not sure how to understand and approach these.
Also do you have any idea how to approach the arbitrary constants for approximating pi? Do we just have to leave them like this for the lack of a better approach?

stilscher:
For approximating pi, @michael-schwarz and me discussed, that one could write a C-stub for `M_PI` and round it for a lower and upper bound in case it is not exactly representable.

src/cdomain/value/cdomains/floatDomain.ml
```diff
@@ -618,8 +618,87 @@ module FloatIntervalImpl(Float_t : CFloatType) = struct
...
+      Interval (min (Float_t.cos Down l) (Float_t.cos Down h), Float_t.of_float Up 1.)
+    else
+      of_interval (-. 1., 1.)
+
+  let eval_sin_cfun l h =
+    let (dist, l'', h'') = project_and_compress l h 2. in
+    if Messages.tracing then Messages.trace "CstubsTrig" "cos: dist %s; l'' %s; h'' %s\n" (Float_t.to_string dist) (Float_t.to_string l'') (Float_t.to_string h'');
+    (* phase shift -0.25 -> same phase as cos *)
+    let l'' = if l'' <= Float_t.of_float Down 0.25 then Float_t.add Down l'' (Float_t.of_float Down 0.75) else Float_t.sub Down l'' (Float_t.of_float Up 0.25) in
+    let h'' = if h'' <= Float_t.of_float Down 0.25 then Float_t.add Up h'' (Float_t.of_float Up 0.75) else Float_t.sub Up h'' (Float_t.of_float Down 0.25) in
```
stilscher, line 678: I think we should avoid this shift (and the duplication of the case distinction below). We could instead shift the argument and compute `sin(x)` as `cos(x - 0.25)` (with the respective rounding modes for lower and upper bound).

FelixKrayer:
Thanks for fixing the semgrep issues. I have removed the phase shift and changed the case checks instead

stilscher:
And there now seems to be an issue with GobView because there is no implementation for the stub we added for pi.

FelixKrayer:
> And there now seems to be an issue with GobView because there is no JavaScript implementation for the stub I added for pi.

Would it work to do something like 
`#ifndef M_PI`
`#define M_PI -1.`
`#endif`
in the c-stubs file and then check if `Float_t.pi <= 0.` and use the arbitrary under/overapprox constants in this case?

src/cdomain/value/cdomains/floatDomain.ml
```diff
@@ -618,8 +618,71 @@ module FloatIntervalImpl(Float_t : CFloatType) = struct
     | (l, _) when l > Float_t.zero -> false_zero_IInt ()
     | _ -> unknown_IInt () (**any interval containing zero has to fall in this case, because we do not distinguish between 0. and -0. *)
 
-  (**This Constant overapproximates pi to use as bounds for the return values of trigonometric functions *)
+  (**These constants over-/underpproximate pi *)
   let overapprox_pi = 3.1416
+  let underapprox_pi = 3.1415
```
michael-schwarz, line 623: Is this still needed after adding the stub?

src/cdomain/value/cdomains/floatDomain.ml
```diff
@@ -618,8 +618,71 @@ module FloatIntervalImpl(Float_t : CFloatType) = struct
...
+      else
+        Float_t.sub Up h' (Float_t.of_float Down (Z.to_float (Z.pred (Float_t.to_big_int h'))))
+    in
+    (dist, l'', h'')
+
+  let eval_cos_cfun l h =
+    let (dist, l'', h'') = project_and_compress l h 2. in
+    if Messages.tracing then Messages.trace "CstubsTrig" "cos: dist %s; l'' %s; h'' %s\n" (Float_t.to_string dist) (Float_t.to_string l'') (Float_t.to_string h'');
+    if (dist <= Float_t.of_float Down 0.5) && (h'' <= Float_t.of_float Down 0.5) && (l'' <= h'') then
+      (** case: monotonic decreasing interval*)
```
michael-schwarz, line 659: If I see it correctly, these are not doc comments attached to anything, so it should probably be single `*`.

src/common/cdomains/floatOps/stubs.c
```diff
@@ -101,3 +114,8 @@ CAMLprim value smallest_float(value unit)
 {
     return caml_copy_double(FLT_MIN);
 }
+
+CAMLprim value pi(value unit)
+{
+    return caml_copy_double(M_PI);
```
michael-schwarz, line 120: Why do we need only the double version of `pi` and not a float version like we have for the unions?

src/cdomain/value/cdomains/floatDomain.ml
michael-schwarz: Thanks for the quick changes! I was wondering if the same doesn't hold for `overapprox_pi`? Can we just use the value from the stub and round up?

src/common/cdomains/floatOps/stubs.c
stilscher: We found that their is implicit conversion from float to double and these stubs are only used for the `CFloat` module. The `CDouble` module can instead use the respective functions from the OCaml Float module (since OCaml Floats correspond to C doubles and we do not need to set the rounding mode here).

src/cdomain/value/cdomains/floatDomain.ml
```diff
@@ -708,14 +705,14 @@ module FloatIntervalImpl(Float_t : CFloatType) = struct
...
-      of_interval (0., (overapprox_pi))
+      Interval (Float_t.of_float Down 0., Float_t.succ Float_t.pi)
     | (l, h) -> norm @@ Interval (Float_t.acos Down h, Float_t.acos Up l) (* acos is monotonic decreasing in [-1, 1]*)
 
   let eval_asin = function
     | (l, h) when l = h && l = Float_t.zero -> of_const 0. (*asin(0) = 0*)
     | (l, h) when l < (Float_t.of_float Down (-.1.)) || h > (Float_t.of_float Up 1.) ->
       Messages.warn ~category:Messages.Category.Float "Domain error might occur: asin argument might be outside of [-1., 1.]";
-      div (of_interval ((-. overapprox_pi), overapprox_pi)) (of_const 2.)
+      div (Interval (Float_t.pred (Float_t.neg Float_t.pi), Float_t.succ Float_t.pi)) (of_const 2.)
```
stilscher, line 715: I think no rounding is necessary since the pi float from `math.h` should be exactly representable and negation is exact since IEEE 754 floats have a sign bit.

stilscher:
In order to debug the CI test that fails for mac OS, I ran it on a MacBook, but there it worked. This could be due to the MacBook having an ARM architecture whereas the macos-latest runner from github probably runs on an Intel architecture. So I wanted to try it on an old iMac as well, but the system is so old that I have not managed to install Goblint so far.

vesalvojdani:
I could reproduce this. It suffices as a test case:
```c
int main()
{
    cos(-1.0);
}
```

Seems that on Intel Macs, this `norm @@ Interval (Float_t.cos Down x, Float_t.cos Up x)` can fail for some values of `x` --- the show function prints them as equal, but somehow the result of the DOWN mode is strictly greater than the result of the UP mode.

FelixKrayer:
I changed the Float_t.to_string function so that all floats (and doubles) should be differentiable when printing now. This should help with debugging

stilscher:
So the problem seems to be that the rounding mode only specifies how floating point numbers are rounded _per instruction_. `cos` is however usually approximated with different polynomials. The implementation can vary depending on the system. The rounding mode only influences how the single instructions within this approximation are rounded. It is not safe to assume, that rounding `cos(x)` with rounding mode down is necessarily smaller than `cos x` with rounding mode up.

sim642:
I might have mentioned it somewhere at some point, but there's also crlibm. Maybe we'll have to turn to something like that.

michael-schwarz:
@sim642 How would crlibm help us here? It correctly rounds w.r.t. to the mathematical definition and not w.r.t. some arbitrarily (mis-)behaving implementation.

Am I missing something here?

FelixKrayer:
If we can assume that for any operation `op(<implementation>, <rounding mode>, <argument>)`, it holds that `op(CRlibm, up, x)` >= `op(<any impl>, <any RM>, x)` (and vice versa for down), we could use CRlibm for a safe upper (& lower) bound.
However, looking at the weirdly behaving `cos()`, I am not sure if that assumption holds here.

stilscher:
> A proposal of @michael-schwarz and me would be to make this feature configurable and compute the cos for the upper bound with both, Up and Down as rounding mode and take the greater of both (and similarly for the lower bound). This is of course not completely safe as there is no guarantee that the cos implementation is monotonic in between, but would allow to retain precision for the float values and it probably is a proper approximation in most cases.

At our meeting today, we discussed that we would like to implement the above idea, make this feature optional (turned off by default) and comment/warn that it can be unsound.

FelixKrayer:
> At our meeting today, we discussed that we would like to implement the above idea, make this feature optional (turned off by default) and comment/warn that it can be unsound.

I implemented the option `math_fun_eval_cstub` to toggle this functionality. Also all previous function applications of `fun down x` are now guarded by `min (fun up x) (fun down x)` and analogously applications of `fun up x` are guarded by a `max`. This is done for all calls to trigonometric functions as well as for `sqrt`

However, thinking about it, I am wondering if we should include `fun nearest x` into the min/max. Nearest can round up or down depending on what is closer to the actual outcome. Assuming a function calculates `(a op b) / (y op x)`, the result of this calculated with `nearest` can be greater than the max of `up` and `down`. This can happen if the numerator is rounded up and the denominator is rounded down, while for `up` and `down` both are rounded the same. 
An example for this would be `a = y = 1.; b = x = 0.000000001;` in` (a - b) / (y + x);`, where for `up` and `down` a result less than 1 is calculated while for `nearest` the result is exactly 1 (reproduce with [roundtest.c](https://github.com/goblint/analyzer/files/14454167/roundtest.c.txt)).

I have not yet thought about if there is a similar issue for rounding mode `ToZero`

stilscher (approved):
The failing indentation check is unrelated.

sim642 (changes requested):

PR #1281, opened 2023-12-01
RFC: Remove `spec` & `file` analyses

michael-schwarz:
This removes the `spec` and `file` analyses that stem from Ralf's Master's thesis.

We have sort of kept them compiling but not really tested or used them over the past few years. We have also found cases where they are unsound (do not consider all sub-expressions for dereference e.g.).

I would thus propose to remove these and point potential users of this feature to one of our past releases.


docs/developer-guide/messaging.md
```diff
@@ -47,16 +47,3 @@ The `~loc` argument is optional and defaults to the current location, but allows
...
-
-Warnings inside `.spec` files are converted to warnings.
-They parsed from string warnings: the first space-delimited substring determines the category and the rest determines the text.
-
-For example:
-```
-w1 "behavior.undefined.use_after_free"
-w2 "integer.overflow"
-w3 "unknown my message"
-w4 "integer.overflow some text describing the warning"
```
sim642, line 61: I think the `from_string_list` functions in `MessageCategory` were just added to parse these and should now be removed as well.

src/analyses/spec.ml
sim642, line 1: There's also `SpecDomain`, `FileDomain` and `MvalMapDomain` that all can go.

docs/developer-guide/messaging.md
michael-schwarz: They are also used in `of_yojson` for the messages. Do you recall why we needed this functionality?

src/analyses/spec.ml
michael-schwarz: Removed.

PR #1288, opened 2023-12-07
Organize more common modules into dune libraries

sim642:
This continues #1206.
My moving a couple of functions to better-suited modules, some odd module dependencies can be eliminated, allowing more things to be decomposed.

This PR introduces new dune libraries:
1. `goblint.config` for the JSON configuration system.
2. `goblint.domain` for general-purpose domain (`Lattice`) functors (e.g. `SetDomain`).
3. `goblint.library` for library function specifications.
4. `goblint.incremental` for incremental comparison and updating.
5. `goblint.tracing` for tracing.

### TODO
- [x] Fix Gobview (https://github.com/goblint/gobview/pull/35).


src/domains/mapDomain.ml
```diff
@@ -263,7 +263,7 @@ module HConsed (M: S) : S with
   type key = M.key and
   type value = M.value =
 struct
-  include Lattice.HConsed (M)
+  include Lattice.HConsed (M) (struct let assume_idempotent = false end)
```
michael-schwarz, line 266: Is this used by the BaseAnalysis? If it is , the this counters the effect of having the short-circuiting, as int values will often live in these.

sim642: It's not used at all, base uses `HashCached`.

src/cdomains/offset.ml
```diff
@@ -22,7 +22,7 @@ struct
     include CilType.Exp
     let name () = "exp index"
 
-    let any = CastE (TInt (Cilfacade.ptrdiff_ikind (), []), mkString "any_index")
+    let any = Cilfacade.any_index_exp
     let all = CastE (TInt (Cilfacade.ptrdiff_ikind (), []), mkString "all_index")
```
michael-schwarz, line 26: Why not also do this for `all` while we're at it?

src/framework/constraints.ml
```diff
@@ -12,12 +12,17 @@ module M = Messages
 
 (** Lifts a [Spec] so that the domain is [Hashcons]d *)
 module HashconsLifter (S:Spec)
-  : Spec with module D = Lattice.HConsed (S.D)
-          and module G = S.G
+  : Spec with module G = S.G
           and module C = S.C
```
michael-schwarz, line 16: Are we not exposing D here anymore on purpose?

sim642: Well yeah, it's clearly not needed (and probably the other constraints aren't either). I wouldn't be eager about removing all such constraints but there also isn't a reason to insist on keeping them since it causes duplication of these module definitions which can be quite long with multiple functor applications.

In fact, it cannot even be done here in the signature:
```
File "src/framework/constraints.ml", line 15, characters 48-54:
15 |   : Spec with module D = Lattice.HConsed (S.D) (struct let assume_idempotent = GobConfig.get_string "ana.int.refinement" = "never"
                                                     ^^^^^^
Error: Syntax error: module path expected.
```
Due to https://github.com/goblint/analyzer/pull/1186#discussion_r1363866695, the argument module cannot be defined anywhere before either without changing the behavior.

Issue #1260, opened 2023-11-21
Suppress thread-unsafe library function calls in single-threaded mode


sim642:
We don't record accesses in single-threaded mode for race detection at all (they don't even show with `allglobs`) to avoid overhead on single-threaded programs. However, thread-unsafe library function calls do get recorded in single-threaded mode, but there's no need to do that either.

```c
#include <stdlib.h>

int main() {
  rand();
  return 0;
}
```

```console
$ ./goblint bar.c --enable allglobs 
[Info][Deadcode] Logical lines of code (LLoC) summary:
  live: 3
  dead: 0
  total lines: 3
[Success][Race] Memory location rand (safe): (/usr/include/stdlib.h:454:12-454:24)
  call with [mhp:{tid=[main]}, multi:false, thread:[main]] (conf. 110)  (exp: rand) (bar.c:4:3-4:9)
[Info][Race] Memory locations race summary:
  safe: 1
  vulnerable: 0
  unsafe: 0
  total memory locations: 1
```

There should be no race summary printed for this program at all.

PR #1307, opened 2023-12-27
Suppress thread-unsafe library function calls in single-threaded mode

karoliineh:
Solves #1260

### DIFF
--- a/src/analyses/raceAnalysis.ml
+++ b/src/analyses/raceAnalysis.ml
@@ -369,7 +369,7 @@ struct
   let special ctx (lvalOpt: lval option) (f:varinfo) (arglist:exp list) : D.t =
     (* perform shallow and deep invalidate according to Library descriptors *)
     let desc = LibraryFunctions.find f in
-    if List.mem LibraryDesc.ThreadUnsafe desc.attrs then (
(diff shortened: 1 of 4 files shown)


src/analyses/raceAnalysis.ml
```diff
@@ -369,7 +369,7 @@ struct
   let special ctx (lvalOpt: lval option) (f:varinfo) (arglist:exp list) : D.t =
     (* perform shallow and deep invalidate according to Library descriptors *)
     let desc = LibraryFunctions.find f in
-    if List.mem LibraryDesc.ThreadUnsafe desc.attrs then (
+    if List.mem LibraryDesc.ThreadUnsafe desc.attrs && not (ctx.ask (Queries.MustBeSingleThreaded {since_start=true})) then (
```
sim642, line 372: The same check as for normal accesses would be even better:
https://github.com/goblint/analyzer/blob/ea1bf23148b3babf04d4d2d844f4738b2c7e64e5/src/analyses/raceAnalysis.ml#L307
It also suppresses the accesses when the program has gone back single-threaded by joining other threads.

sim642 (approved):

Issue #1266, opened 2023-11-22
Remove duplicate `ctx`-derived arguments in base analysis


sim642:
Base analysis is full of code like
```ocaml
set_many ~ctx (Analyses.ask_of_ctx ctx) gs st [...]
```
where `Analyses.ask_of_ctx ctx`, `gs` and `st` are directly taken from `ctx` itself. There's no point in having those three arguments because everything could be accessed via the `ctx` that's already passed around.

This is the case for _a lot_ of code in base, so this would be a major cleanup.

sim642:
Reopening because many `st` arguments still remain after #1308 due to possible mismatch between `ctx.local` and `st` in the middle of certain operations: https://github.com/goblint/analyzer/pull/1308#discussion_r1445953004.

PR #1308, opened 2023-12-27
Remove duplicate ctx-derived arguments in base analysis

karoliineh:
Solves #1266 

### TODO
- [x] `reachable_from_value`
- [x] `reachable_from_address`
- [x] `reachable_vars`
- [x] Before and after sv-benchmarks run. @sim642

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1678,7 +1678,7 @@ struct
   let invariant = Invariant.invariant
 
 
-  let set_savetop ~ctx ?lval_raw ?rval_raw ask (gs:glob_fun) st adr lval_t v : store =
+  let set_savetop ~ctx ?lval_raw ?rval_raw st adr lval_t v : store =
     if M.tracing then M.tracel "set" "savetop %a %a %a\n" AD.pretty adr d_type lval_t VD.pretty v;
     match v with
     | Top -> set ~ctx st adr lval_t (VD.top_value (AD.type_of adr)) ?lval_raw ?rval_raw
@@ -1769,15 +1769,15 @@ struct
               let iv = VD.bot_value ~varAttr:v.vattr t in (* correct bottom value for top level variable *)
               if M.tracing then M.tracel "set" "init bot value: %a\n" VD.pretty iv;
               let nv = VD.update_offset (Queries.to_value_domain_ask (Analyses.ask_of_ctx ctx)) iv offs rval_val (Some  (Lval lval)) lval t in (* do desired update to value *)
-              set_savetop ~ctx (Analyses.ask_of_ctx ctx) ctx.global ctx.local (AD.of_var v) lval_t nv ~lval_raw:lval ~rval_raw:rval (* set top-level variable to updated value *)
+              set_savetop ~ctx  ctx.local (AD.of_var v) lval_t nv ~lval_raw:lval ~rval_raw:rval (* set top-level variable to updated value *)
             | None ->
-              set_savetop ~ctx (Analyses.ask_of_ctx ctx) ctx.global ctx.local lval_val lval_t rval_val ~lval_raw:lval ~rval_raw:rval
+              set_savetop ~ctx ctx.local lval_val lval_t rval_val ~lval_raw:lval ~rval_raw:rval
           end
         | _ ->
-          set_savetop ~ctx (Analyses.ask_of_ctx ctx) ctx.global ctx.local lval_val lval_t rval_val ~lval_raw:lval ~rval_raw:rval
+          set_savetop ~ctx ctx.local lval_val lval_t rval_val ~lval_raw:lval ~rval_raw:rval
       end
     | _ ->
-      set_savetop ~ctx (Analyses.ask_of_ctx ctx) ctx.global ctx.local lval_val lval_t rval_val ~lval_raw:lval ~rval_raw:rval
+      set_savetop ~ctx ctx.local lval_val lval_t rval_val ~lval_raw:lval ~rval_raw:rval
 
 
   let branch ctx (exp:exp) (tv:bool) : store =
@@ -2666,7 +2666,7 @@ struct
             | _, _ -> begin
                 let new_val = get ask ctx.global fun_st address None in
                 if M.tracing then M.trace "taintPC" "update val: %a\n\n" VD.pretty new_val;
-                let st' = set_savetop ~ctx ask ctx.global st address lval_type new_val in
+                let st' = set_savetop ~ctx st address lval_type new_val in
                 let partDep = Dep.find_opt v fun_st.deps in
                 match partDep with
                 | None -> st'
@@ -2753,7 +2753,7 @@ struct
 
       match lval with
       | None      -> st
-      | Some lval -> set_savetop ~ctx (Analyses.ask_of_ctx ctx) ctx.global st (eval_lv ~ctx lval) (Cilfacade.typeOfLval lval) return_val
+      | Some lval -> set_savetop ~ctx st (eval_lv ~ctx lval) (Cilfacade.typeOfLval lval) return_val
     in
     combine_one ctx.local after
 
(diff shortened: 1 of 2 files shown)


src/analyses/base.ml
```diff
@@ -540,17 +540,17 @@ struct
   (* Get the list of addresses accessable immediately from a given address, thus
    * all pointers within a structure should be considered, but we don't follow
    * pointers. We return a flattend representation, thus simply an address (set). *)
-  let reachable_from_address (ask: Q.ask) (gs:glob_fun) st (adr: address): address =
+  let reachable_from_address ~ctx (adr: address): address =
     if M.tracing then M.tracei "reachability" "Checking for %a\n" AD.pretty adr;
-    let res = reachable_from_value ask gs st (get ask gs st adr None) (AD.type_of adr) (AD.show adr) in
+    let res = reachable_from_value ~ctx (get (Analyses.ask_of_ctx ctx) ctx.global ctx.local adr None) (AD.type_of adr) (AD.show adr) in
```
sim642, line 545: Could `get` not be changed to also take `ctx`?

src/analyses/base.ml
```diff
@@ -695,14 +695,15 @@ struct
       Base itself also answers EvalInt, so recursion goes indirectly through queries.
       This allows every subexpression to also meet more precise value from other analyses.
       Non-integer expression just delegate to next eval_rv function. *)
-  and eval_rv_ask_evalint a gs st exp =
-    let eval_next () = eval_rv_no_ask_evalint a gs st exp in
+  and eval_rv_ask_evalint ~ctx exp =
+    let ask = Analyses.ask_of_ctx ctx in
```
sim642, line 699: This seems unnecessary because you can just use `ctx.ask` below.

src/analyses/base.ml
```diff
@@ -898,10 +902,10 @@ struct
     if M.tracing then M.traceu "evalint" "base eval_rv_base %a -> %a\n" d_exp exp VD.pretty r;
     r
 
-  and eval_rv_base_lval ~eval_lv (a: Q.ask) (gs:glob_fun) (st: store) (exp: exp) (lv: lval): value =
+  and eval_rv_base_lval ~eval_lv ~ctx (a: Q.ask) (gs:glob_fun) (st: store) (exp: exp) (lv: lval): value =
```
sim642, line 905: Even if the `st` cannot be easily removed, maybe `a` and `gs` could be?

src/analyses/base.ml
```diff
@@ -1105,19 +1110,32 @@ struct
       if Queries.Set.mem anyq asked then
         Queries.Result.top q (* query cycle *)
       else (
-        let asked' = Queries.Set.add anyq asked in
         match q with
-        | EvalInt e -> query_evalint (ask asked') gs st e (* mimic EvalInt query since eval_rv needs it *)
+        | EvalInt e -> query_evalint ~ctx e (* mimic EvalInt query since eval_rv needs it *)
```
sim642, line 1114: This no longer detects query cycles because the `asked` set will always be empty. I'll see if there's an example of doing this somewhere else.

src/analyses/base.ml
```diff
@@ -1430,11 +1449,12 @@ struct
     (* Updating a single varinfo*offset pair. NB! This function's type does
      * not include the flag. *)
     let update_one_addr (x, offs) (st: store): store =
+      let ask = Analyses.ask_of_ctx ctx in
```
sim642, line 1452: Should be able to just use `ctx.ask` below.

src/analyses/base.ml
```diff
@@ -2627,13 +2651,13 @@ struct
...
-    let ask = (Analyses.ask_of_ctx ctx) in
-    AD.fold (fun addr st ->
+    let ask = Analyses.ask_of_ctx ctx in
+    AD.fold (fun addr (st: store) ->
         match addr with
         | Addr.Addr (v,o) ->
           if CPA.mem v fun_st.cpa then
             let lval = Addr.Mval.to_cil (v,o) in
-            let address = eval_lv ask ctx.global st lval in
+            let address = eval_lv ~ctx lval in
```
sim642, line 2660: This is now semantically different, it used to evaluate in `st` which changes during the fold, but now it's evaluated in `ctx.local`. We'll have to check if this is still the right thing to do or not.

src/analyses/baseInvariant.ml
```diff
@@ -64,7 +64,7 @@ struct
 
   let refine_lv_fallback ctx a gs st lval value tv =
     if M.tracing then M.tracec "invariant" "Restricting %a with %a\n" d_lval lval VD.pretty value;
-    let addr = eval_lv a gs st lval in
+    let addr = eval_lv ~ctx lval in
```
sim642, line 67: This is also semantically different now because maybe the `st` argument it previously used is something that keeps changing during refinement?

Other places below beg the same question.

src/analyses/base.ml
karoliineh: Should I just add the `st` parameter back here and for the cases below as of now, or are we planning on looking into those more deeply?

src/analyses/baseInvariant.ml
```diff
@@ -78,8 +78,9 @@ struct
         else
           old_val
       in
-      let state_with_excluded = set a gs st addr t_lval value ~ctx in
-      let value =  get a gs state_with_excluded addr None in
+      let ask = Analyses.ask_of_ctx ctx in
+      let state_with_excluded = set ask ctx.global st addr t_lval value ~ctx in
```
sim642, line 82: Seems like this `set` could also do without `ask` and `ctx.global`?

src/analyses/base.ml
sim642: This can be bypassed entirely because some stupid conversions are happening here:
`addr` (`Addr.t`) → `(v, o)` (`Addr.Mval.t`) → `lval` (`lval`) → `address` (`AD.t`).
Instead this could be directly done with
```ocaml
let address = AD.singleton addr in
```

karoliineh:
Although the regression tests did pass with the `st` argument removed, it is not clear how it affects the refinement. Thus, as it is not a priority right now, I have just put the `st` argument back everywhere. However, the cleanup from replacing the `ask` and `gs` arguments with `ctx` is still significant enough.

I have still removed the `st` argument from functions that did not even use it (e.g. [`evalbinop_base`](https://github.com/goblint/analyzer/pull/1308/commits/5f18ee7f2100f0ce8557c50662b18b9416ef7154) and `reachable_from_value`), and the functions that just passed the argument `st` to the functions that did not use it.

src/analyses/base.ml
```diff
@@ -1809,88 +1824,89 @@ struct
...
    **************************************************************************)
 
   (** From a list of expressions, collect a list of addresses that they might point to, or contain pointers to. *)
-  let collect_funargs ask ?(warn=false) (gs:glob_fun) (st:store) (exps: exp list) =
+  let collect_funargs ~ctx ?(warn=false) (st:store) (exps: exp list) =
     let do_exp e =
-      let immediately_reachable = reachable_from_value ask gs st (eval_rv ask gs st e) (Cilfacade.typeOf e) (CilType.Exp.show e) in
-      reachable_vars ask [immediately_reachable] gs st
+      let immediately_reachable = reachable_from_value ~ctx (eval_rv ~ctx st e) (Cilfacade.typeOf e) (CilType.Exp.show e) in
+      reachable_vars ~ctx [immediately_reachable]
```
michael-schwarz, line 1880: I'm not sure I quite get why is this guaranteed to be the same? To me, it looks like we are evaluating some things on `st` and others on `ctx.local`?

sim642: This PR currently leaves `st` as separate from `ctx.local` because of this, but there's a broader conceptual problem with it even before this PR.
Namely, even if a different `st` is passed and used for some local lookups, `ask` isn't (and cannot be) modified, so `EvalInt` query would still run on the original local state. So having a separate local state is overall questionable and should maybe be addressed separately in the future somehow.

src/analyses/base.ml
sim642: Here's an example of how to do it correctly: https://github.com/goblint/analyzer/blob/4f09c289cb2509b9e2e07bdb120ac7f34a2bcd59/src/analyses/base.ml#L1534-L1550.
Basically, `ctx` needs to be a function with the `asked` argument.

sim642 (approved):
Verdicts and performance on sv-benchmarks are essentially unchanged, which is good.

michael-schwarz (approved):

PR #1310, opened 2023-12-28
Unbox some types

sim642:
This unboxes some types:
1. records with a single field (often used for second-order polymorphism reasons),
2. variants with a single constructor with a single argument.

The idea is to optimize these by removing the extra layer of indirection. Moreover, it might be beneficial to avoid allocating all these `Analyses.ask_of_ctx` blocks.

Moreover, `ask_of_ctx` is eta-reduced to avoid additional closure allocation even with it unboxed.

### TODO
- [x] Benchmark on sv-benchmarks.
- [x] Check Gobview compatibility.


michael-schwarz:
Nice! I didn't know that OCaml already has this support for unboxed types in limited circumstances, I was only aware of the more general ideas and work on this at JaneStreet. 
(e.g. https://github.com/ocaml/RFCs/blob/unboxed-types/rfcs/unboxed-types.md and https://icfp22.sigplan.org/details/mlfamilyworkshop-2022-papers/13/Unboxed-types-for-OCaml).

I was initially worried about what this means for our JavaScript build and marshaling, but given this feature has been there for a long time, I guess all the kinks should be ironed out by now.

sim642:
> I was initially worried about what this means for our JavaScript build and marshaling, but given this feature has been there for a long time, I guess all the kinks should be ironed out by now.

Since these attributes are just for optimization, I suppose it'd be fine even if jsoo didn't optimize like this (although maybe it shares enough of the OCaml compilation pipeline to still do, not sure).
In such case the only potential is marshaled data, so I'll have to see if Gobview still works. Most of these shouldn't even be part of marshaled data, but the `Inc` thing in `IntDomain` might be.

sim642:
sv-benchmarks runs show ~4% CPU time improvement from these changes:
image

Most of it seems to come from https://github.com/goblint/analyzer/pull/1310/commits/be69a349aabf14a87980aa5415b36a81f36bd193, though. Or at least the benefit of unboxing only shows together with that change.
The reason being: unboxing removes one memory block indirection, eta-reduction removes a second one. As a result, `ask_of_ctx` is truly just a field lookup with no allocation.

michael-schwarz (approved):

PR #1312, opened 2023-12-28
Small output readability improvements

### LINKED ISSUE #1192: Small output readability improvements
Our messages contain various redundant information which makes the messages long and harder to read. Based on some option we could try to produce simpler output for some things:
- [x] `(0,[0,0])` → `0` — constant integers don't need the constant repeated in every domain.
- [ ] `(Unknown int([0,8]),[0,255])` → `[0,255]` — intervals are clearer than equivalent def_exc ranges.
- [x] `(Unknown int([-63,63]),[-9223372036854775808,9223372036854775807])` → `⊤` — top intervals are easier recognized as such.
- [x] `(alloc@sid:82773@tid:main(#top))` → `(alloc@sid:82773@tid:main)` — disabled malloc uniqueness counter is excessive.
- [x] `(alloc@sid:488@tid:Top Threads(#top))` → `(alloc@sid:488)` — top thread ID is excessive.
- [x] `(alloc@sid:82843@tid:main(#top))[def_exc:0; intervals:[0,0]]` → `(alloc@sid:82843@tid:main)[0]` — covered by above.
- [x] `(alloc@sid:82832@tid:main(#top))[def_exc:Unknown int([-63,63]); intervals:[-9223372036854775808,9223372036854775807]]` → `(alloc@sid:82832@tid:main)[⊤]` — covered by above.
- [ ] `mhp:{tid=POOL_thread; created=All Threads}` → `mhp:{tid=POOL_thread}` — top created threads set cannot exclude anything anyway?
- [x] `mhp:{tid=main; created=All Threads}, thread:main` → `thread:main` — thread ID repeated by `threadid` and `mhp` analyses.

karoliineh:
Resolves #1192


src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -131,8 +131,19 @@ module MallocWrapper : MCPSpec = struct
...
+          else
+            Format.dprintf "(#%s)" (UniqueCount.show c)
+        in
+        let tid =
+          if ThreadLifted.is_top t then
+            Format.dprintf ""
+          else
+            Format.dprintf "@tid:%s%t" (ThreadLifted.show t) uniq_count
+        in
+        Format.asprintf "(alloc@sid:%s%t)" (Node.show_id node) tid
```
sim642, line 146: The statement IDs (`sid`) were added into alloc variable names for incremental analysis at some point. Are they actually required, @jerhard / @stilscher?

The one remaining problem in this PR is that cram tests fail on OSX because `sid`s are different. I would just remove them from the variable names altogether (races on those variables are reported for the right node locations anyway), but maybe some incremental logic requires them to match variables by name?

Without it, alloc variable names in the same thread would all have the same name (but different `vid` and location). For the user looking at Goblint race warnings, these numbers are meaningless clutter in the output. Even for us, they are only meaningful when looking at the HTML at the same time.

src/cdomains/intDomain.ml
```diff
@@ -407,8 +407,16 @@ struct
...
+      "⊤"
+    else
+      I.show x.v  (* TODO add ikind to output *)
+  let pretty () x =
+    if I.is_top_of x.ikind x.v then
+      Pretty.text "⊤"
+    else
+      I.pretty () x.v (* TODO add ikind to output *)
   let pretty_diff () (x, y) = I.pretty_diff () (x.v, y.v) (* TODO check ikinds, add them to output *)
   let printXml o x = I.printXml o x.v (* TODO add ikind to output *)
```
sim642, line 421: Maybe for consistency the top-specialization should also be in `printXml` so it's consistent in HTML output. It's not automatically testable, but still.

src/cdomains/intDomain.ml
```diff
@@ -3688,18 +3714,16 @@ module IntDomTupleImpl = struct
...
-      None
-    )
-  let to_int = same BI.to_string % mapp2 { fp2 = fun (type a) (module I:SOverflow with type t = a and type int_t = int_t) -> I.to_int }
   let to_bool = same string_of_bool % mapp { fp = fun (type a) (module I:SOverflow with type t = a) -> I.to_bool }
   let minimal = flat (List.max ~cmp:BI.compare) % mapp2 { fp2 = fun (type a) (module I:SOverflow with type t = a and type int_t = int_t) -> I.minimal }
   let maximal = flat (List.min ~cmp:BI.compare) % mapp2 { fp2 = fun (type a) (module I:SOverflow with type t = a and type int_t = int_t) -> I.maximal }
   (* others *)
-  let show = String.concat "; " % to_list % mapp { fp = fun (type a) (module I:SOverflow with type t = a) x -> I.name () ^ ":" ^ (I.show x) }
+  let show x =
+    match to_int x with
```
sim642, line 3722: Same here, this could also be in `printXml`.

src/analyses/wrapperFunctionAnalysis.ml
stilscher: Julian and me think that it should be possible to remove the `sid`  from the variable names. It was added when removing the location from the name, because locations can change in different program versions even without actual semantic changes. Since the malloc variable names are not part of the `NodeVarinfoMap` we can at least not think of a possible problem when removing the `sid` from the name.

michael-schwarz: > Without it, alloc variable names in the same thread would all have the same name (but different vid and location). 

Do we print `vid`s by default everywhere we print names? Otherwise removing the `sid` is confusing because it will be harder to track the evolution of a variable across program states if both copies have the same name (I'm not sure if we can always rely on the order in such casses).

src/analyses/mHPAnalysis.ml
```diff
@@ -12,7 +12,9 @@ struct
     include MHP
     let name () = "mhp"
     let may_race = MHP.may_happen_in_parallel
-    let should_print _ = true
+    let should_print {tid; created; must_joined} =
+      not (ConcDomain.ThreadSet.is_empty created)
+      || not (ConcDomain.ThreadSet.is_empty must_joined)
```
michael-schwarz, line 17: How is this an improvement? All this information is needed to understand why accesses are considered racy.

src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -131,8 +131,19 @@ module MallocWrapper : MCPSpec = struct
       CilType.Location.show loc
 
     let name_varinfo (t, node, c) =
-      Format.asprintf "(alloc@sid:%s@tid:%s(#%s))" (Node.show_id node) (ThreadLifted.show t) (UniqueCount.show c)
-
+        let uniq_count =
+          if UniqueCount.is_top c then
+            Format.dprintf ""
```
michael-schwarz, line 136: If the uniqueness counter is activated and some things are kept apart, would it not be useful to know that this is

`alloc@...#T` (which means it's not `alloc@...#0`). If nothing is kept apart, I agree that this is indeed superfluous.

src/analyses/wrapperFunctionAnalysis.ml
```diff
@@ -131,8 +131,19 @@ module MallocWrapper : MCPSpec = struct
...
+          else
+            Format.dprintf "(#%s)" (UniqueCount.show c)
+        in
+        let tid =
+          if ThreadLifted.is_top t then
```
michael-schwarz, line 141: I'd argue the same here: If we are attempting to compute useful thread ids, I'd still make the top explicit.

src/cdomains/mHP.ml
```diff
@@ -24,7 +24,6 @@ let current (ask:Queries.ask) =
   }
 
 let pretty () {tid; created; must_joined} =
-  let tid_doc = Some (Pretty.dprintf "tid=%a" ThreadIdDomain.ThreadLifted.pretty tid) in
```
michael-schwarz, line 27: Whys is the tid of the thread not interesting here?

src/cdomains/threadIdDomain.ml
```diff
@@ -56,10 +56,13 @@ struct
         (struct let name = "no node" end))
 
   let show = function
```
michael-schwarz, line 58: I'd make the same argument here as for the malloc counters.

src/cdomains/mHP.ml
sim642: See the example in #1192. MHP takes `tid` directly by querying the thread ID analysis, but that analysis already shows this exact information. So thread IDs are always duplicated in race output.

src/analyses/mHPAnalysis.ml
michael-schwarz: Ok, for must-joined I guess this holds.
An empty created set is very strong information though (no information would be top).

src/analyses/wrapperFunctionAnalysis.ml
sim642: Useful to know what is?

This is completely consistent whether or not the counter limit is greater than zero: the variable without `#i` is the non-unique one, those with `#i` are the unique ones. We don't need two different representations for the same thing.

We don't need Goblint to tell the user in every warning what analysis the user activated. They activated those things themselves in the configuration they wrote.

src/analyses/mHPAnalysis.ml
sim642: If the created set is empty, we're already single-threaded and that's already reflected by the thread flag.

src/analyses/wrapperFunctionAnalysis.ml
michael-schwarz: I would want to follow the principle of least surprise. If I enable an option to keep `n` allocs apart, I want the system to tell me whether it was able to assign a unique counter to a given heap variable. Otherwise, I have to do some sort of inferring what it means for the system to leave out the `#i`.

Imagine for example I have a huge program and I want to inspect the warning at one program point. I turn on the uniqueness counter, and nothing changes about the warning. Was it that I specified the option in the wrong file? Did I load an old result? If I now see `#top` I know that my option had some effect.

src/analyses/wrapperFunctionAnalysis.ml
michael-schwarz: This then is an argument that we should keep `sid`s regardless.

karoliineh: The issue seems to be caused by the function `create_var` in `cilfacade.ml` because vars with the same name get the same hash and thus also the same `vid`. @sim642 has already questioned this function in https://github.com/goblint/analyzer/pull/1092#discussion_r1447083592.

src/analyses/wrapperFunctionAnalysis.ml
sim642: > Was it that I specified the option in the wrong file? Did I load an old result? If I now see `#top` I know that my option had some effect.

That logic doesn't apply to most of our options though. Making analyses path-sensitive, enabling int domain refinement, enabling widening thresholds, adding malloc or thread wrappers, etc.
The goal of this PR is not to fix the new issue you bring up. Nor does keeping the `#top` fix users accidentally looking at some wrong path for configuration or results.

michael-schwarz: Maybe we should discuss this at the GobCon with everyone. I feel the proposed change here makes it harder to understand and interpret the result rather than easier.

src/analyses/wrapperFunctionAnalysis.ml
jerhard: The string that is shown for a malloced-block does not necessarily have to be the same `vname` as the name; one could adapt the printing in `BaseType.Variables.show`. 

For HTML-output I would also be in favor of having at least one of the line or the sid visible.

src/cdomain/value/cdomains/threadIdDomain.ml
```diff
@@ -55,19 +55,23 @@ struct
...
-    | (f, None) -> f.vname
+  let show (f, ni_opt) = 
+    let vname = f.vname in
+    match ni_opt with
+    | None -> vname
+    | Some (n, i_opt) ->
+      let vname_loc = vname ^ "@" ^ CilType.Location.show (UpdateCil.getLoc n) in
+      match i_opt with
+      | Some i -> vname_loc ^ "#" ^ string_of_int i
+      | None when GobConfig.get_bool "dbg.full-output" -> vname_loc ^ "#top"
```
sim642, line 66: Maybe use `⊤` here as well, similarly to some other places now.

sim642 (approved):
Looks good by me. With `dbg.full-output` all the old output is retained now.

The splitting up of `mhp:{...}` in accesses I would leave for a separate PR because it's more than just `show`/`pretty`/`printXml` changes. Unless we want it to be like `thread:{tid=...; created={...}}`, which would be the case since the created thread set is also in the `threadid` analysis domain. Allowing one analysis to nicely output two components into accesses probably requires some `MCPAccess` changes.

michael-schwarz (approved):
LGTM now! Sorry I was unaware that this was still waiting for me review!

PR #1323, opened 2024-01-11
Implement `combine_assign` in `threadEscape`

karoliineh:


src/analyses/threadEscape.ml
```diff
@@ -133,17 +132,26 @@ struct
     let ask = Analyses.ask_of_ctx ctx in
     let vs = mpt ask (AddrOf lval) in
     if D.exists (fun v -> v.vglob || has_escaped ask v) vs then (
-      let escaped = escape_rval ctx rval in
+      let escaped = escape_rval ctx (Analyses.ask_of_ctx ctx) rval in
```
sim642, line 135:
```suggestion
      let escaped = escape_rval ctx ask rval in
```

src/analyses/threadEscape.ml
```diff
@@ -133,17 +132,26 @@ struct
...
     ) else begin
       ctx.local
     end
 
+  let combine_assign ctx (lval:lval option) (fexp:exp) f args fc au f_ask : D.t =
+    let ask = Analyses.ask_of_ctx ctx in
+    match lval with
+    | Some lval when D.exists (fun v -> v.vglob || has_escaped ask v) (mpt ask (AddrOf lval)) ->
+      let rval = Lval (ReturnUtil.return_lval ()) in
+      let escaped = escape_rval ctx f_ask rval in
```
michael-schwarz, line 146: Could you elaborate on why `f_ask` is the one we want here? Maybe also leave a comment on this?

karoliineh: Because `ask` will answer in the context of the outer function, but we are interested in escaping the return value of the function that is called that is not yet reachable in the outer context.

michael-schwarz (approved):

PR #1326, opened 2024-01-11
Support for inline assembler & `goto` instructions in inline assembler

WernerDrasche:
Depends on https://github.com/goblint/cil/pull/161.

---

The tests with annotations will come in very few days.
We are sorry for this (right now still) unpolished state.

Changes in goblint-cil:
- asm is now statement
- goto asm is parsed

Changes in analyzer: all changes are switched off by default and can be turned on by --disable asm_is_nop
- CFG now has edges to labels referenced in goto asm label list
- base emits new Invalidate {lvals} event (with the lvals of the out-clobbers)
- We tried to the best of our abilities to handle this event and/or add asm transfer functions in the analyses where necessary


michael-schwarz:
Depends on https://github.com/goblint/cil/pull/161.

michael-schwarz:
Before we can continue reviewing, can you please merge master into this and temporarily pin `goblint-cil` in the opam file such that the CI can run?

WernerDrasche:
I don't know how to tell this to use our goblint-cil repo (the pinning and downgrading procedure as described on your website only works locally I think)

michael-schwarz:
> Edit: I think it should work now.

The pin should be repeated in the `goblint.opam.locked` file (for the locked jobs that run on every commit), other than that you found the right way to do it.

WernerDrasche:
- The main.ml changes were unintentionally forgotten testing utilities.
- asmlabelslst_ne is not used in first step because we semantically distinguish between:
  - labels field was provided (Some [...]) 
  - labels field was omitted (None)
- the check asm function ensures that goto attribute <=> labels field (otherwise we would need state in the parser)
- as cil cfg is only used to make the goblint cfg, we don't do anything special for the cil one
- (just noticed that I didn't see the label printing todo in the email => will fix immediately)
- `ASM a` does not work for some reason (replying to the comment in email)
- I don't know what to do for the dummy instruction (do we need to introduce a new instruction type for that?)

src/analyses/mallocFresh.ml
```diff
@@ -29,6 +29,23 @@ struct
   let assign ctx lval rval =
     assign_lval (Analyses.ask_of_ctx ctx) lval ctx.local
 
+    let is_relevant_variable v =
+      (* Check if the variable is a global variable *)
+      v.vglob
+    
+    let event ctx e octx =
+        match e with
+        | Events.Invalidate {lvals} ->
```
jerhard, line 38: The `lvals` here are only from the `out` set. Could the value of a pointer variable (referencing an allocated block) in the `in` set be read and copied somewhere, e.g. to a global variable, by the assembly?

src/analyses/malloc_null.ml
```diff
@@ -16,6 +17,12 @@ struct
   module C = ValueDomain.AddrSetDomain
   module P = IdentityP (D)
 
+  let ignore_asm = ref true
+
+  (*TODO: why is this init not called??? *)
+  let init _ =
+    ignore_asm := get_bool "asm_is_nop"
```
jerhard, line 24: The `init` of the analysis is only called when an incremental analysis is to be performed and the analysis should be initialized with some data from the previous analysis run.

src/analyses/uninit.ml
```diff
@@ -29,6 +30,11 @@ struct
   let threadspawn ctx ~multiple lval f args fctx = ctx.local
   let exitstate  v : D.t = D.empty ()
 
+  let ignore_asm = ref true
+
+  let init _ = 
+    ignore_asm := get_bool "asm_is_nop"
```
jerhard, line 36: Only called for incremental runs.

src/analyses/useAfterFree.ml
```diff
@@ -24,9 +25,12 @@ struct
   module G = ThreadIdToJoinedThreadsMap
   module V = VarinfoV
 
-  let context _ _ = ()
+  let ignore_asm = ref true
 
+  let init _ =
+    ignore_asm := get_bool "asm_is_nop"
```
jerhard, line 31: Only executed for incremental runs.

src/cdomain/value/cdomains/valueDomain.ml
```diff
@@ -725,7 +725,7 @@ struct
       Array (CArrays.set ask n (array_idx_top) v)
     |                 t , Blob n       -> Blob (Blobs.invalidate_value ask t n)
     |                 _ , Thread tid   -> Thread (Threads.join (Threads.top ()) tid)
-    |                 _ , JmpBuf _     -> state (* TODO: no top jmpbuf *)
+    |                 _ , JmpBuf (b, c, _)     -> JmpBuf(b, c, true) (* TODO: no top jmpbuf *)
```
jerhard, line 728: I guess the TODO can be removed now.

src/common/framework/cfgTools.ml
```diff
@@ -141,6 +141,8 @@ end
 
 module CfgEdgeH = BatHashtbl.Make (CfgEdge)
 
+let ignore_asm = get_bool "asm_is_nop"
```
jerhard, line 144: This is executed right at the start, before the Configuration is loaded.

src/common/framework/edge.ml
```diff
@@ -21,9 +21,9 @@ type t =
     * transferred to the function node! *)
   | Test of CilType.Exp.t * bool
   (** The true-branch or false-branch of a conditional exp *)
-  | ASM of string list * asm_out * asm_in
+  | ASM of string list * asm_out * asm_in * bool
   (** Inline assembly statements, and the annotations for output and input
-    * variables. *)
+    * variables. The last field is a flag that indicates if this is a duplicated edge. *)
```
jerhard, line 26: You may add a few words to explain this in some more detail for future readers unaware of our discussions: Why are they duplicated and in what sense are they duplicated?

src/config/options.schema.json
```diff
@@ -17,6 +17,12 @@
       "type": "string",
       "default": ""
     },
+    "asm_is_nop": {
+      "title": "asm_is_nop",
+      "description": "treat asm statements as nop",
```
jerhard, line 22:
```suggestion
      "description": "Treat assembly statements as NOOP.",
```

src/framework/constraints.ml
```diff
@@ -781,7 +781,7 @@ struct
       | Entry f        -> tf_entry var edge prev_node f
       | Ret (r,fd)     -> tf_ret var edge prev_node r fd
       | Test (p,b)     -> tf_test var edge prev_node p b
-      | ASM (_, _, _)  -> tf_asm var edge prev_node (* TODO: use ASM fields for something? *)
+      | ASM (_, _, _,_)  -> tf_asm var edge prev_node (* TODO: use ASM fields for something? *)
```
jerhard, line 784:
```suggestion
      | ASM (_, _, _, _)  -> tf_asm var edge prev_node (* TODO: use ASM fields for something? *)
```

src/maingoblint.ml
```diff
@@ -570,7 +573,7 @@ let do_analyze change_info merged_AST =
         (* Cilfacade.current_file := ast'; *)
     in
 
-    Timing.wrap "analysis" (control_analyze merged_AST) funs
+    Timing.wrap "analysis" (control_analyze merged_AST) funs
```
jerhard, line 576: Please revert this change introducing a trailing white-space.

jerhard (commented):
Thank you for your pull request! I added a first round of comments that you may have a look at :)
As you already indicated, regression tests (that are run with `make test`) would be good.

michael-schwarz:
This message from github about the semgrep jobs confuses me: I only approved the run as usual. Not sure what this is about...

WernerDrasche:
What shall we do in cfgTools.ml if we can't access GobConfig?

src/analyses/malloc_null.ml
jerhard: I have to correct myself here: The `init` is actually called, even in a non-incremental run, but with `None` as a paramter. I got confused by some comment in the PR. Apologies for that.

sim642:
> What shall we do in cfgTools.ml if we can't access GobConfig?

What does that mean? `CfgTools` opens `GobConfig` and uses `get_bool` in a bunch of places just fine.

WernerDrasche:
`9 test(s) failed: ["00/41 both_branches-2", "02/71 pthread-once", "03/31 zstd-cctxpool-blobs", "29/30 nondets", "33/05 fun_ptranal", "41/01 qsort", "41/02 bsearch", "74/03 nested-ptr-uaf", "74/16 uaf-packed-struct"]`

These tests fail. Can this have anything to do with our changes?

src/analyses/expsplit.ml
```diff
@@ -96,6 +96,14 @@ struct
       D.add exp value ctx.local
     | Longjmped _ ->
       emit_splits_ctx ctx
+    | Invalidate {lvals} ->
+      let handle_lval local lval =
+        let exp = Lval lval in
+        if D.mem exp local then begin
+          D.remove exp local
+        end else local
```
michael-schwarz, line 104:
```suggestion
        if D.mem exp local then
          D.remove exp local
        else
          local
```

src/analyses/memLeak.ml
```diff
@@ -216,9 +216,8 @@ struct
         | ad when (not (Queries.AD.is_top ad)) && Queries.AD.cardinal ad = 1 ->
           (* Note: Need to always set "ana.malloc.unique_address_count" to a value > 0 *)
           begin match Queries.AD.choose ad with
-            | Queries.AD.Addr.Addr (v,_) when ctx.ask (Queries.IsAllocVar v) && ctx.ask (Queries.IsHeapVar v) && not @@ ctx.ask (Queries.IsMultiple v) ->
-              ToppedVarInfoSet.remove v ctx.local
-            | _ -> ctx.local
+            | Queries.AD.Addr.Addr (v,_) when ctx.ask (Queries.IsAllocVar v) && ctx.ask (Queries.IsHeapVar v) && not @@ ctx.ask (Queries.IsMultiple v) -> ToppedVarInfoSet.remove v state
+            | _ -> state
           end
```
michael-schwarz, line 221: Is there any semantic change to this file? It seems like only a rename happened here?

src/cdomain/value/cdomains/valueDomain.ml
```diff
@@ -724,7 +724,7 @@ struct
       Array (CArrays.set ask n (array_idx_top) v)
     |                 t , Blob n       -> Blob (Blobs.invalidate_value ask t n)
     |                 _ , Thread tid   -> Thread (Threads.join (Threads.top ()) tid)
-    |                 _ , JmpBuf _     -> state (* TODO: no top jmpbuf *)
+    |                 _ , JmpBuf (b, c, _)     -> JmpBuf(b, c, true)
```
michael-schwarz, line 727: The comment should probably still remain, as we normally would have to do more here but just haven't gotten around to it yet.

src/framework/constraints.ml
```diff
@@ -775,14 +775,14 @@ struct
...
-      | Test (p,b)     -> tf_test var edge prev_node p b
-      | ASM (_, _, _)  -> tf_asm var edge prev_node (* TODO: use ASM fields for something? *)
-      | Skip           -> tf_skip var edge prev_node
+      | Assign (lv,rv)  -> tf_assign var edge prev_node lv rv
+      | VDecl (v)       -> tf_vdecl var edge prev_node v
+      | Proc (r,f,ars)  -> tf_proc var edge prev_node r f ars
+      | Entry f         -> tf_entry var edge prev_node f
+      | Ret (r,fd)      -> tf_ret var edge prev_node r fd
+      | Test (p,b)      -> tf_test var edge prev_node p b
+      | ASM (_, _, _,_) -> tf_asm var edge prev_node (* TODO: use ASM fields for something? *)
```
michael-schwarz, line 784: Why not provide the arguments `asm_in` and `asm_out`  to `tf_asm` here? Then you would not need to call `asm_extract_ins_outs`

src/analyses/memLeak.ml
WernerDrasche: This was from merge with master.

src/framework/constraints.ml
WernerDrasche: Do you mean to add ins, outs as arguments to the asm transfer function?

WernerDrasche: I did that now.

src/analyses/accessAnalysis.ml
```diff
@@ -73,6 +75,17 @@ struct
       ctx.local
     end
 
+  let asm ctx outs ins =
+    if not !ignore_asm then begin
```
jerhard, line 79: Should the `asm_is_top` flag impact all analyses? If this check be moved into the `mCP`, so that it only needs to be done once?

sim642: I think it probably should be at the `FromSpec` level even. And the top-level `asm_is_nop` option should probably be replaced with something more organized like `sem.asm.nop`.

sim642:
The `asm goto` part has been redone in #2067. The invalidation part is interesting but probably too out of date to still try to use, it's not too hard to redo.

PR #1297, opened 2023-12-17
Linear Two-Variable Equalities Analysis

alina-weber:
## Linear Two-Variable Equalities Analysis

This pull request introduces an analysis for linear two variable equalities (ref: [A. Flexeder, M. Petter, and H. Seidl Fast Interprocedural Linear Two-Variable Equalities](http://doi.acm.org/10.1145/2049706.2049710)). The structure of the analysis implementation is mostly based on the affine equality analysis (`affeq`).

The supported equalities are of the form `x = y + c`, where `x` is a variable and `c` is a constant. We also perform partial evaluation in some cases, e.g. when there is an assignment of the form `x_1 = x_2 + x_3 - x_4` and we know that `x_3 == x_2` and `x_4 == x_2`, then it follows that `x_1 == x_2`.

To enable the analysis, the parameter `--set ana.activated[+] lin2vareq` needs to be added. 

## Domain

Abstract states in this domain are represented by structs containing an optional array and an apron environment. 
The bottom element is represented by a struct with "None" instead of the array.

The arrays are modeled as proposed in the paper: Each variable is assigned to an index and each array element represents a linear relationship that must hold at the corresponding program point.
The apron environment is used to organize the order of columns and variables.

The length of the array always corresponds to the number of variables in the environment.
If for example in the array at index `j` we store the element `(Some i, k)`, this means that our analysis found out that `x_i = x_j + k`. If the array entry at index `j` is `(None, k)`, it means that `x_j = k`, where `k` is a constant and `x_i` and `x_j` are variables.

## Overflow handling

The overflow handling in `affeq` was spread over multiple files and wasn't very precise. We rewrote it in a more readable way and we exploited the interval analysis to check if there is an overflow in a certain expression. 

The query `MaySignedOverflow` was added to the base analysis to handle possible overflow of signed integer expressions. Each subexpression is analyzed to see if an overflow happened. For each operator in the expression, we use the query EvalInt to approximate the bounds of each operand and we compute if in the worst case there could be an overflow.

This query is used by the domains `affeq` and `lin2vareq` in order to find expressions that possibly contain an overflow. 

## Casting

We found a bug in the analysis `affeq`, where casts where not properly handled. We fixed the bug for `affeq` and for our domain.

## Refactoring of `affeq`

In order to have less code duplication, we moved some functions from affineEqualityDomain to sharedFunctions, such that affineEqualityDomain and our linearTwoVarEqualityDomain can both use them.

PR #1327, opened 2024-01-12
`affeq`: Fix array OOB in `invariant`

michael-schwarz:
During benchmarking for #1297, it became apparent that the `invariant` function in `affeq` accesses arrays out-of-bounds due to an off-by-one error.

This fixes the issue and adds a test that has witness generation enabled to spot regressions.

### DIFF
--- a/tests/regression/63-affeq/19-witness.c
+++ b/tests/regression/63-affeq/19-witness.c
@@ -15,4 +15,20 @@ void main(void) {
         j = j + 3;
     }
     __goblint_check(3 * i - j + k == 1);
+
+    // Represented with fractional coefficients and thus not put into witness yet
+
+    int a = 0;
+    int b = 0;
+    int z = 0;
+
+    while(z < 100) {
+        a++;
+        b += 2;
+        z++;
+
+        __goblint_check(2*z - b == 0);
+        // b == 2*z is put into the witness
+    }
+
 }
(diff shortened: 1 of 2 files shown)


sim642 (approved):
Seems right, although the entire `invariant` is very roundabout: `Matrix` → `earray` → `array` → `Enum` → `list`. A single `List.init` should do it.

tests/regression/63-affeq/19-witness.c
```diff
@@ -0,0 +1,18 @@
...
+
+void main(void) {
+    int i;
+    int j;
+    int k;
+    i = 2;
+    j = k + 5;
+
+    while (i < 100) {
+        __goblint_check(3 * i - j + k == 1);
```
sim642, line 13: This invariant doesn't appear in `witness.yml` though.

michael-schwarz: That is another unrelated issue. The invariant is given as `i#318-1/3j#319+1/3k#320+4/3=0` which then likely does not survive conversion to a CIL expression. I suggest we open a new issue, as this is a general problem of `relationAnalysis`, which I won't have time to address any time in the next two weeks.

No longer crashing here is a step in right direction in the meanwhile.

jerhard (approved):

tests/regression/63-affeq/19-witness.c
michael-schwarz: I created a new issue for this.

Issue #1317, opened 2024-01-10
Use `Z` module directly


sim642:
Many places are very verbose and use `IntOps.BigIntOps` or an abbreviated `BI` module to work with big integers from Zarith.
Their type is fixed to be `Z.t`, so we should just use the `Z` module directly whereever possible. It's shorter and more consistent.

PR #1329, opened 2024-01-15
Use Z module directly

karoliineh:
Solves #1317


src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -3142,9 +3139,9 @@ struct
     | None, None -> bot ()
     | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
     | _, x when leq zero x -> top ()
-    | Some(c1, m1), Some(c2, m2) when not no_ov && m2 =: Ints_t.zero && c2 =: Ints_t.neg Ints_t.one -> top ()
-    | Some(c1, m1), Some(c2, m2) when m1 =: Ints_t.zero && m2 =: Ints_t.zero -> Some(c1 /: c2, Ints_t.zero)
-    | Some(c1, m1), Some(c2, m2) when m2 =: Ints_t.zero ->  if (c2 |: m1) && (c2 |: c1) then Some(c1 /: c2, m1 /: c2) else top ()
+    | Some(c1, m1), Some(c2, m2) when not no_ov && m2 =: Z.zero && c2 =: Z.neg Z.one -> top ()
+    | Some(c1, m1), Some(c2, m2) when m1 =: Z.zero && m2 =: Z.zero -> Some (c1 /: c2, Z.zero)
+    | Some(c1, m1), Some(c2, m2) when m2 =: Z.zero && c2 |: m1 && c2 |: c1 -> Some (c1 /: c2, m1 /: c2)
     | _, _ -> top ()
```
karoliineh, line 3145: Is this first case (returning `top ()` as the default case) redundant or is it necessary for clearness?

michael-schwarz: Are you sure that this case is redundant? It seems one might otherwise go into the second of the highlighted cases here?

karoliineh:
There now remain 5 module definitions in `IntDomain` where `IntOps.BigIntOps` is used in the whole project.

src/analyses/baseInvariant.ml
```diff
@@ -140,7 +139,7 @@ struct
               match ID.to_int n with
               | Some n ->
                 (* When x != n, we can return a singleton exclusion set *)
-                if M.tracing then M.tracec "invariant" "Yes, %a is not %s\n" d_lval x (BI.to_string n);
+                if M.tracing then M.tracec "invariant" "Yes, %a is not %s\n" d_lval x (Z.to_string n);
```
sim642, line 142: We should use the opportunity to add `GobZ.pretty` and use it here (and some other places). It would avoid unnecessary string conversions when output is not needed.

src/analyses/baseInvariant.ml
```diff
@@ -381,8 +380,8 @@ struct
           | _, _ -> a, b
         end
       | Lt | Le | Ge | Gt as op ->
-        let pred x = BI.sub x BI.one in
-        let succ x = BI.add x BI.one in
+        let pred x = Z.pred x in
+        let succ x = Z.succ x in
```
sim642, line 384: Should just inline these now.

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -1776,45 +1774,42 @@ struct
 end
 
 module BigInt = struct
-  include  BI
+  include IntOps.BigIntOps
```
sim642, line 1777: Why can't we get rid of this?

src/cdomain/value/cdomains/intDomain.mli
```diff
@@ -411,17 +407,17 @@ module Interval32 :Y with (* type t = (IntOps.Int64Ops.t * IntOps.Int64Ops.t) op
 
 module BigInt:
   sig
-    include Printable.S with type t = Z.t (* TODO: why doesn't this have a more useful signature like IntOps.BigIntOps? *)
```
sim642, line 414: This comment should maybe remain if we need to keep `BigInt` (see previous comment).

src/common/util/intOps.ml
```diff
@@ -241,7 +241,7 @@ struct
 
   let shift_left = Z.shift_left
   let shift_right = Z.shift_right
-  let bitnot x = sub (neg x) one
+  let bitnot = Z.lognot
   let bitand = Z.logand
   let bitor = Z.logor
   let bitxor = Z.logxor
```
sim642, line 247: Is this `bitnot` definition equivalent to the old one? Maybe there's some catch why it wasn't so before.

Also, we need to figure out what to do about this `bit*` vs `log*` confusion, where our naming differs from the standard library one (that Zarith also uses). I think a bunch of places now directly use `Z.logand` instead of `BI.bitand`, which may cause some confusion again.

One possibility would be to use `logand` for the bitwise operations to be consistent with OCaml. And then rename `bitand` to `c_logand` to emphasize the C nature of it.

unittest/cdomains/intDomainTest.ml
```diff
@@ -110,7 +110,6 @@ end
 
 module Ikind = struct let ikind () = Cil.ILong end
 module A = IntTest (IntDomain.Integers (IntOps.BigIntOps))
-module B = IntTest (IntDomain.FlattenedBI)
```
sim642, line 113: Could still keep the same test with
```ocaml
module B = IntTest (IntDomain.Flat (IntDomain.Integers (IntOps.BigIntOps)))
```

src/common/util/intOps.ml
karoliineh: I was wondering the same, but then I checked the Zarith documentation, and there stands:
```
val lognot: t -> t
(** Bitwise logical negation. 
    The identity [lognot a]=[-a-1] always hold.
 *)
```
which at least by the definition seems to be the exact same operation.

src/cdomain/value/cdomains/intDomain.ml
karoliineh: Because `BigInt.logand` and `BigInt.logor` (from `IntOpsDecorator`) are different from `Z.logand` and `Z.logor`

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -3574,12 +3575,14 @@ module IntDomTupleImpl = struct
...
     if List.mem `Neq xs then `Neq else
       `Top
-  let same show x = let xs = to_list_some x in let us = List.unique xs in let n = List.length us in
-    if n = 1 then Some (List.hd xs)
-    else (
-      if n>1 then Messages.info ~category:Unsound "Inconsistent state! %a" (Pretty.docList ~sep:(Pretty.text ",") (Pretty.text % show)) us; (* do not want to abort *)
+  let same show x =
+    let us = List.unique (to_list_some x) in
+    match us with
+    | [x] -> Some (List.hd us)
```
sim642, line 3581:
```suggestion
    match us with
    | [x] -> Some x
```

src/common/util/intOps.ml
michael-schwarz: Renaming is fine by me 👍

src/domains/intDomainProperties.ml
```diff
@@ -145,9 +145,9 @@ struct
   let valid_shift_left = make_valid2 ~name:"shift_left" ~cond:shift_cond CD.shift_left AD.shift_left
   let valid_shift_right = make_valid2 ~name:"shift_right" ~cond:shift_cond CD.shift_right AD.shift_right
 
-  let valid_lognot = make_valid1 ~name:"lognot" ~cond:not_bot CD.lognot AD.lognot
-  let valid_logand = make_valid2 ~name:"logand" ~cond:none_bot CD.logand AD.logand
-  let valid_logor = make_valid2 ~name:"logor" ~cond:none_bot CD.logor AD.logor
+  let valid_lognot = make_valid1 ~name:"lognot" ~cond:not_bot CD.c_lognot AD.c_lognot
+  let valid_logand = make_valid2 ~name:"logand" ~cond:none_bot CD.c_logand AD.c_logand
+  let valid_logor = make_valid2 ~name:"logor" ~cond:none_bot CD.c_logor AD.c_logor
```
sim642, line 150:
```suggestion
  let valid_c_lognot = make_valid1 ~name:"c_lognot" ~cond:not_bot CD.c_lognot AD.c_lognot
  let valid_c_logand = make_valid2 ~name:"c_logand" ~cond:none_bot CD.c_logand AD.c_logand
  let valid_c_logor = make_valid2 ~name:"c_logor" ~cond:none_bot CD.c_logor AD.c_logor
```

These also need to be added to `tests` below (which currently has duplicates).

michael-schwarz:
I think some comments also still refer to bitwise operations where they should now refer to logical operations to be consistent.

src/cdomain/value/cdomains/intDomain.mli
```diff
@@ -77,13 +77,13 @@ sig
...
 
-  val logand: t -> t -> t
+  val c_logand: t -> t -> t
   (** Logical and: [x && y] *)
 
-  val logor : t -> t -> t
+  val c_logor : t -> t -> t
   (** Logical or: [x || y] *)
 
 end
```
michael-schwarz, line 89: This is one of the places where comments and function names should be made to match.

sim642: This matches though. The `c_` prefix means it's logical in the C sense, which is what the comment also says.

sim642: Confusing how? They still correctly refer to what kind of C operations they're for and include example C expressions.

The confusing thing is OCaml using `log` prefix for bitwise operations and calls them "bitwise logical" in standard library documentation. All we can do is be consistent with it.

sim642 (approved):

michael-schwarz (approved):

PR #1334, opened 2024-01-24
Exclude Goblint stubs from YAML witnesses

sim642:
This fixes the issue identified by Freiburg.

The other "issue" with `(unsigned long )_ == 0UL` or really any `(unsigned long )arg == 0UL` has no easy improvement in Goblint. I thought it came from relation analysis because of the cast, but apparently it comes from base. Passing `0` into an `void*` argument performs no cast and simply puts the integer abstraction into the state for the pointer variable. The outputted invariant comes from `IntDomain` instead, which correctly inserts the cast.

I thought about using `Cilfacade.makeBinOp` which hopefully would be smarter about inserting casts for binary operators, but actually it's worse and turns it into `(unsigned long )arg == (unsigned long )((void *)0)` because of https://github.com/goblint/cil/issues/162. If that is fixed, then we might have a neater way.

### DIFF
--- a/src/witness/yamlWitness.ml
+++ b/src/witness/yamlWitness.ml
@@ -193,13 +193,23 @@ struct
     in
     let task = Entry.task ~input_files ~data_model ~specification in
 
+    let is_stub_node n =
+      let fundec = Node.find_fundec n in
+      Cil.hasAttribute "goblint_stub" fundec.svar.vattr
+    in
+
     let is_invariant_node (n : Node.t) =
       let loc = Node.location n in
       match n with
-      | Statement _ when not loc.synthetic && WitnessInvariant.is_invariant_node n -> true
+      | Statement _ when not loc.synthetic && WitnessInvariant.is_invariant_node n ->
+        not (is_stub_node n)
       | _ ->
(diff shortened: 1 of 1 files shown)


src/witness/yamlWitness.ml
```diff
@@ -193,15 +193,25 @@ struct
...
+      let fundec = Node.find_fundec n in
+      Cil.hasAttribute "goblint_stub" fundec.svar.vattr
+    in
+
     let is_invariant_node (n : Node.t) =
       let loc = Node.location n in
       match n with
-      | Statement _ when not loc.synthetic && WitnessInvariant.is_invariant_node n -> true
+      | Statement _ when not loc.synthetic && WitnessInvariant.is_invariant_node n ->
+        not (is_stub_node n)
```
michael-schwarz, line 205: Any particular reason why this is not added to the `when` clause above?

michael-schwarz (approved):
Other than the one remark on style. LGTM!

src/witness/yamlWitness.ml
sim642: I probably did it for line length reasons, but I now refactored it to just have all the conditions in the case body.

PR #1335, opened 2024-01-24
Fix `NullPtr` assignment crash in relation analysis

### LINKED ISSUE #1249: `exception Apron.Manager.Error  {  exn = Exc_invalid_argument; funid = Funid_substitute_texpr_array;   msg = unknown variable d#311#arg in the environment; }`
The test `ldv-linux-3.16-rc1/205_9a_array_safes_linux-3.16-rc1.tar.xz-205_9a-drivers--net--wireless--ti--wlcore--wlcore_spi.ko-entry_point.cil.out.i` crashes in our sv-comp runs with error

```
Fatal error: exception Apron.Manager.Error
 {  exn = Exc_invalid_argument; funid = Funid_substitute_texpr_array;
  msg = unknown variable d#311#arg in the environment; }
  ```

An extracted minimal example looks like this: Of course, this program here has UB because `a` is a null-pointer, but the analyses should nevertheless run and report an error and not just crash.

~~~C
// SKIP PARAM: --set ana.activated[+] apron
int *a;
int b;
void c(int d) {
    // *a is a null pointer here, so we should warn but maybe not crash
    *a = d;
}
int main() {
    c(b);
}
~~~

This is not highest priority, but might be worth having a look at, since it is a strange bug.

### PR DESCRIPTION
**Pull request by sim642**
Closes #1249.

If assigning through a pointer which isn't unknown, but also doesn't have any known pointers (so is NULL or maybe some string), then previously `D.bot ()` was returned.
This has an empty environment which broke `combine` in the particular test because `combine` expects to find the `#arg` variables added by `enter`.

If there's such a crash in SV-COMP, then likely it has to be due to a NULL pointer as well.

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -158,13 +158,14 @@ struct
         {st' with rel = rel''}
       )
     | (Mem v, NoOffset) ->
-      begin match ask.f (Queries.MayPointTo v) with
-        | ad when Queries.AD.is_top ad -> st
-        | ad ->
-          let mvals = Queries.AD.to_mval ad in
-          let ass' = List.map (fun mval -> assign_to_global_wrapper ask getg sideg st (ValueDomain.Addr.Mval.to_cil mval) f) mvals in
-          List.fold_right D.join ass' (D.bot ())
-      end
+      let ad = ask.f (Queries.MayPointTo v) in
+      Queries.AD.fold (fun addr acc ->
+          match addr with
+          | ValueDomain.Addr.Addr mval ->
+            D.join acc (assign_to_global_wrapper ask getg sideg st (ValueDomain.Addr.Mval.to_cil mval) f)
+          | UnknownPtr | NullPtr | StrPtr _ ->
+            D.join acc st (* Ignore assign *)
+        ) ad (D.bot ())
     (* Ignoring all other assigns *)
     | _ ->
       st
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/analyses/apron/relationAnalysis.apron.ml [hunk 1]
```diff
@@ -158,13 +158,14 @@ struct
...
-          let ass' = List.map (fun mval -> assign_to_global_wrapper ask getg sideg st (ValueDomain.Addr.Mval.to_cil mval) f) mvals in
-          List.fold_right D.join ass' (D.bot ())
-      end
+      let ad = ask.f (Queries.MayPointTo v) in
+      Queries.AD.fold (fun addr acc ->
+          match addr with
+          | ValueDomain.Addr.Addr mval ->
+            D.join acc (assign_to_global_wrapper ask getg sideg st (ValueDomain.Addr.Mval.to_cil mval) f)
+          | UnknownPtr | NullPtr | StrPtr _ ->
+            D.join acc st (* Ignore assign *)
```
michael-schwarz, line 167: Is simply ignoring the assignment in this case consistent with what base does?

sim642: As far as I can tell, yes (in `set`):
https://github.com/goblint/analyzer/blob/71e21b27973c78ef58c2567e314e128cfbb10564/src/analyses/base.ml#L1579-L1582
Just a bit more explicit here that we're ignoring unknown and NULL pointers.

michael-schwarz (approved):

Issue #1302, opened 2023-12-19
Unsoundness with context-insensitive analyses: sv-comp tests "list_search"


SchiJoha:
If all analyses are context insensitive, the sv-comp testcases `list-properties/list_search-1.i` and `list-properties/list_search-2.i` fail with the `svcomp24.json` configuration. (Similar to issue #1298)

During execution the error:  `[Error][Analyzer][Unsound] Both branches dead` is thrown. This comes from the fact, that any traversal through the list is detected to be infinite. Hence the whole remaining code is declared dead.

To set all analyses context insensitive I changed the `mCP.ml` file in line 83 from `cont_inse := map' find_id @@ get_string_list "ana.ctx_insens";` to `cont_inse := xs;`.
After compiling and executing the sv-comp testcase, Goblint incorrectly reports valid memory cleanup. The SV-COMP result is true, but false is expected.
I used those commands to run the testcases:
list_search-1: 
```
./goblint --conf conf/svcomp24.json --sets ana.specification "CHECK( init(main()), LTL(G valid-memcleanup) )" ../../sv-benchmarks/c/list-properties/list_search-1.i 
```
list_search-2: 
```
./goblint --conf conf/svcomp24.json --sets ana.specification "CHECK( init(main()), LTL(G valid-memcleanup) )" ../../sv-benchmarks/c/list-properties/list_search-2.i 
```
If only one of the analyses `base`, `mallocWrapper`, `var_eq`, `apron`, `memLeak` is analyzed context sensitively (and the remaining context insensitively), both tests produce the correct result (so the SV-COMP result is unknown). An overview can be found in the following picture. 

![auswertung](https://github.com/goblint/analyzer/assets/50140250/a7df67ba-baed-4021-aa77-fead010256f0)

jerhard:
@SchiJoha and I found after some debugging that this issue seems to be due to a erroneous `join` in in the relational `DHetero` where the result of some `bot_env` and `top_env` values results in some `bot_env` value.

PR #1340, opened 2024-01-26
master thesis: Taming Recursion with Three Context-Sensitive Analyses (Callstring, LoopfreeCallstring, Context Gas)

SchiJoha:
Implementation of three context-sensitive approaches to tame recursion:
- call string approach in the Sharir and Pnueli sense: a call stack tracking the last k function calls is simulated and used as context. As elements on the call stack, the function definitions (classical call string) or the statements (similar to call site) are used. We also enable the choice of `k = infinity`. 
- Loopfree Callstring: similar to call string but with (1) the callee stored in the call stack and (2) loops of the call stack stored in a set. Consequently, the stack elements are function definitions and sets containing function definitions. Overall no function is stored twice in the stack. 
- Context Gas: analyzes the first `k` calls on the call stack context-sensitive. The remaining calls receive no context information.

For the call-sequence `main -> f -> g -> f -> g` the approaches would have the following context (and gas):
- Callstring: `(main)`-> `(main, f)` -> `(main, f, g)` -> `(main, f, g, f)` -> `(main, f, g, f, g)`
- Loopfree: `(main)`-> `(main, f)` -> `(main, f, g)` -> `(main, {f, g})` -> `(main, {f, g})`
- Context Gas (gas = 2): `(Some (main-context))` gas 2 -> `(Some (f context))` gas 1 -> `(None)` gas 0  -> `(None)` gas 0  -> `(None)` gas 0


scripts/test-gobview.py
```diff
@@ -26,8 +26,8 @@ def serve():
   global p
   goblint_http_path = '_build/default/gobview/goblint-http-server/goblint_http.exe'
   p = subprocess.Popen(['./' + goblint_http_path,
-                  '-with-goblint', '../analyzer/goblint',
-                  '-goblint', '--set', 'files[+]', '"../analyzer/tests/regression/00-sanity/01-assert.c"'])
+                  '-with-goblint', '../GobContextGas/goblint',
+                  '-goblint', '--set', 'files[+]', '"../GobContextGas/tests/regression/00-sanity/01-assert.c"'])
 
 print("serving at port", PORT)
 thread = Thread(target=serve, args=())
```
michael-schwarz, line 33: This change should probably not be included here?

src/common/util/queueImmut.ml
```diff
@@ -0,0 +1,126 @@
+(* A Queue (FIFO, imutable) implementation, using two Lists for efficient push and pop handling.
```
michael-schwarz, line 1: Is this your own implementation or is it borrowed somewhere? If it is, we should credit the place (and verify licenses are compatible).

SchiJoha: this is my own implementation

src/common/util/queueImmut.ml
```diff
@@ -0,0 +1,126 @@
...
+
+   Example Queue: remove here, 1, 2, 3, 4, add here
+   Representation as Queue: Queue([1,2], [4,3]) or Queue([1], [4,3,2]) or Queue([1,2,3,4], []) or ...
+*)
+
+exception Empty
+
+type 'a t = Queue of 'a list * 'a list
```
michael-schwarz, line 9: Might be worth considering adding an `*.mli` here, and not exposing the type to the outside, then in here, you can remove the `Queue` constructor and just go with having the type be two lists. That would mean it's a bit less code in the end.

src/common/util/queueImmut.ml
michael-schwarz: Ok, nice! Did you compare performance with just using the normale Queue from OCaml? Was a difference notable?

src/common/util/queueImmut.ml
```diff
@@ -0,0 +1,126 @@
...
+  match q1, q2 with
+  | Queue ([], []), Queue ([], []) -> 0
+  | Queue ([], []), Queue ([], x) | Queue ([], []), Queue (x, []) -> -1
+  | Queue (x, []), Queue ([], []) | Queue ([], x), Queue ([], []) -> 1 
+  | Queue ([], q1_last), Queue (q2_first, q2_last) -> compare cmp (Queue (List.rev q1_last, [])) (Queue (q2_first, q2_last))
+  | Queue (q1_first, q1_last), Queue ([], q2_last) -> compare cmp (Queue (q1_first, q1_last)) (Queue (List.rev q2_last, []))
+  | Queue (x1::q1_first, q1_last), Queue (x2::q2_first, q2_last) -> 
+    let c = cmp x1 x2 in
+    if c <> 0 then c 
+    else compare cmp (Queue (q1_first, q1_last)) (Queue (q2_first, q2_last))
```
michael-schwarz, line 56: Is there some possibility for code reuse between `equal` and `compare`? It seems like most of the code is the same?

tests/regression/83-callstring_Stmt/09-thread_creation.c
```diff
@@ -0,0 +1,88 @@
...
+  __goblint_check(result == 10); // UNKNOWN
+  return NULL;
+}
+
+int main()
+{
+  pthread_t id;
+  pthread_t id2;
+
+  // Creat the thread
```
michael-schwarz, line 82:
```suggestion
  // Create the thread
```

also in a few other places :wink:

src/common/util/queueImmut.ml
SchiJoha: I didn't test the performance (the "efficienct" was related to the fact that two lists are used, which should be more efficient than using one list).
The reason for the new Queue is that the Ocaml Queue is mutable. When I edit the Queue in the `enter `transfer function, I need the old queue for the caller and the new queue for the callee. With the Ocaml Queue, I would be forced to copy the queue, and then edit the copied queue. Therefore I implemented an immutable Queue (I just recognized the typo in the comment :))), with which I don't need to copy the whole queue before editing.

sim642: Could just use `BatDeque`, which is the same thing I think.

tests/regression/83-callstring_Stmt/09-thread_creation.c
SchiJoha: The testcases are not perfect yet :) But I will improve them in a later commit

michael-schwarz: Sure, just wanted to remark because such things are missed easily 😉

src/config/options.schema.json
```diff
@@ -963,6 +963,24 @@
...
+            },
+            "ctx_gas_value": {
+              "title": "ana.context.ctx_gas_value",
+              "description": "Number of calls that are analyzed context-sensitive. Used for the ContextGasLifter, which is enabled with ana.context.ctx_gas. Must be >= 0!",
+              "type": "integer",
+              "default": 10
+            },
+            "callStack_height": {
+              "title": "ana.context.callStack_height",
+              "description": "Height of the CallStack that should be used as context for the callsting analysis. Must be >= 0!",
```
michael-schwarz, line 981:
```suggestion
              "description": "Height of the CallStack that should be used as context for the callstring analysis. Must be >= 0!",
```

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+
+(* Specifies the type of the callstack elements for the callstring analysis*)
+module type Callstack_Type =
+sig
+  include CilType.S
+  val stackTypeName: string
+  val pushElem: fundec -> ('d,'g,'c,'v) ctx -> t option (* returns an elements that should be pushed to the Callstack *)
+end
+
+(** Analysis with the k-callsting approach, which uses the last k callstack elements as context.
```
michael-schwarz, line 13:
```suggestion
(** Analysis with the k-callstring approach, which uses the last k callstack elements as context.
```

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
+open Analyses
+open GoblintCil
+open GobConfig
+
+(* Specifies the type of the callstack elements for the callstring analysis*)
+module type Callstack_Type =
+sig
+  include CilType.S
+  val stackTypeName: string
+  val pushElem: fundec -> ('d,'g,'c,'v) ctx -> t option (* returns an elements that should be pushed to the Callstack *)
```
michael-schwarz, line 10:
```suggestion
  val pushElem: fundec -> ('d,'g,'c,'v) ctx -> t option (* returns an element that should be pushed to the Callstack *)
```

I'm not sure if `pushElem` is the best name for this... Maybe we should think if we have a better idea.

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+(** Analysis with the k-callsting approach, which uses the last k callstack elements as context.
+    With the CT argument it is possible to specify the type of the callstack elements*)
+module Spec (CT:Callstack_Type) : MCPSpec = 
+struct
+  include Analyses.IdentitySpec
+
+  (* simulates a call stack of depth k*)
+  module CallStack = struct
+    include Printable.PQueue (CT)
+    let depth = get_int "ana.context.callStack_height" (* must be >= 0 *)
```
michael-schwarz, line 22: Does this not cause issues when the callStack height is changed from it's default value as this is evaluated when the module is constructed?

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+      | Some e -> 
+        let new_stack = BatDeque.cons e stack in (* pushes new element to stack*)         
+        (* removes element from stack, if stack was filled with k elements*)
+        match (BatDeque.size new_stack - depth) with
+        | x when x <= 0 -> new_stack
+        | 1 -> fst @@ Option.get (BatDeque.rear new_stack)
+        | _ -> failwith "Callstack Error: It shouldn't happen that more than one element must be deleted to maintain the correct height!"
+  end
+
+  module D = Lattice.Flat (CallStack)
```
michael-schwarz, line 36: We're being a bit cheeky here by abusing local states to in fact do contexts (and we're forced into it by the framework in sense). 
We probably should add a comment here what we are doing so people not deep into Goblint have an idea what's happening.

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+  module C = CallStack
+  module V = EmptyV
+  module G = Lattice.Unit
+
+  let name () = "callstring_"^ CT.stackTypeName
+  let startstate v = `Lifted (BatDeque.empty)
+  let exitstate v =  `Lifted (BatDeque.empty) (*TODO: should I use startstate here? Does this make a difference*)
```
michael-schwarz, line 43: @vesalvojdani Maybe you can help out here? I constantly forget what exactly `exitstate` is supposed to do...

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+
+  let context fd x = match x with 
+    | `Lifted x -> x
+    | _ -> failwith "Callstring: Context error! The context cannot be derived from Top or Bottom!" (* TODO: is there a possibility???*)
```
michael-schwarz, line 47: No, and this should also never happen.

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+
+  let callee_state ctx f = 
+    let elem = CT.pushElem f ctx in
+    let new_stack = CallStack.push (context f ctx.local) elem in (*TODO: is it ok to use context here??? *)
```
michael-schwarz, line 51: Sure, though in effect all it does here is unlift the value.

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+    `Lifted new_stack
+
+  let enter ctx r f args = [ctx.local, callee_state ctx f]
+
+  let combine_env ctx lval fexp f args fc au f_ask = ctx.local
+
+  let threadenter ctx ~multiple lval v args = [callee_state ctx (Cilfacade.find_varinfo_fundec v)]
```
michael-schwarz, line 58: is this right here? Should a new thread not start with an empty callstack?

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+  let stackTypeName = "loc"
+  let pushElem f ctx =
+    Some !Goblint_tracing.current_loc
+end
+
+(* Lifters for the Callstring approach with different Callstack element types*)
+let _ =
+  MCP.register_analysis (module Spec (Fundec) : MCPSpec); (* name: callstring_fundec*)
+  MCP.register_analysis (module Spec (Stmt) : MCPSpec); (* name: callstring_stmt*)
+  MCP.register_analysis (module Spec (Location) : MCPSpec) (* name: callstring_loc*)
```
michael-schwarz, line 87: Is having three different analyses the best way to expose is? Maybe one analysis and options for how the callstring ought to be abstracted is nicer?

src/analyses/callstringTracking.ml
```diff
@@ -0,0 +1,36 @@
...
+struct
+  include Analyses.IdentitySpec
+
+  let name () = "callstringTracking"
+
+  module FundecSet = SetDomain.Make (CilType.Fundec)
+  module FundecList = Lattice.Flat(Printable.Liszt(CilType.Fundec))
+  module D = Lattice.Prod (FundecList) (FundecSet)
+  module C = D
+  module G = D
```
michael-schwarz, line 13: Why do we need to set `G` to `D` here?

src/analyses/callstringTracking.ml
```diff
@@ -0,0 +1,36 @@
+open Analyses
+
+module Spec : MCPSpec =
+struct
+  include Analyses.IdentitySpec
+
+  let name () = "callstringTracking"
+
+  module FundecSet = SetDomain.Make (CilType.Fundec)
+  module FundecList = Lattice.Flat(Printable.Liszt(CilType.Fundec))
```
michael-schwarz, line 10: Would it be possible to reuse the modules from `callstring` here to make this configurable?

src/analyses/callstringTracking.ml
```diff
@@ -0,0 +1,36 @@
+open Analyses
+
+module Spec : MCPSpec =
+struct
+  include Analyses.IdentitySpec
+
+  let name () = "callstringTracking"
```
michael-schwarz, line 7: Why `callstringTracking`? Maybe `loopfreeCallstring` or something like this is better?

src/analyses/callstringTracking.ml
```diff
@@ -0,0 +1,36 @@
...
+  let startstate v = D.bot ()
+  let exitstate  v = D.bot ()
+
+  let get_list l = match l with
+    | `Lifted l -> l
+    | `Bot -> []
+    | `Top -> failwith "callstringTracking Error: the value of the Flat List shouldn't be top!"
+
+  let callee_state f ctx = 
+    let (list, set) = ctx.local in
```
michael-schwarz, line 24: `list` as a name is a bit suboptimal, it confuses the syntax highlighting.

src/analyses/callstringTracking.ml
```diff
@@ -0,0 +1,36 @@
...
+    if (not(FundecSet.is_empty set)) || List.mem f (get_list list)
+    then (list, FundecSet.add f set)
+    else (`Lifted (f::(get_list list)), set)
+
+  let enter ctx r f args = [ctx.local, callee_state f ctx]
+
+  let threadenter ctx ~multiple lval v args = [callee_state (Cilfacade.find_varinfo_fundec v) ctx]
```
michael-schwarz, line 31: Same question as in the other case: Is this a good callstring for new threads?

src/analyses/callstringTracking.ml
```diff
@@ -0,0 +1,36 @@
+open Analyses
```
michael-schwarz, line 1: Might be good to put a reference to S. et al "Clustered Relational Thread-Modular Abstract Interpretation with Local Traces" where this idea came from (at least to us), or some other paper that proposed this.

conf/callstringFundec.json
```diff
@@ -0,0 +1,146 @@
...
+      "escape",
+      "expRelation",
+      "mhp",
+      "assert",
+      "var_eq",
+      "symb_locks",
+      "region",
+      "thread",
+      "threadJoins",
+      "callstring_fundec"
```
michael-schwarz, line 34: Might be interesting if the analysis from #952 can help make things precise without being too expensive?

src/analyses/mCP.ml
```diff
@@ -65,6 +65,12 @@ struct
     let deps (x,_) = iter (check_dep x) @@ (find_spec x).dep in
     iter deps xs
 
+  let specific_cont_sens xs =
+    (*most analysis must be set to context insensitive, because we only want to analyse context sensitively for a specific analysis*)
```
michael-schwarz, line 69:
```suggestion
    (* most analyses must be set to context insensitive, because we only want to analyse context sensitively for a specific analysis *)
```

src/analyses/mCP.ml
```diff
@@ -76,11 +82,16 @@ struct
...
     base_id := find_id "base";
     activated := map (fun s -> s, find_spec s) xs;
-    path_sens := map' find_id @@ get_string_list "ana.path_sens";
-    cont_inse := map' find_id @@ get_string_list "ana.ctx_insens";
+    (*checks if an analysis is enabled with special handled context sensitivity*)
+    if Option.is_some special_inse
+    then (cont_inse := map' find_id (Option.get special_inse);
+          path_sens := [])
+    else (cont_inse := map' find_id @@ get_string_list "ana.ctx_insens";
+          path_sens := map' find_id @@ get_string_list "ana.path_sens");
```
michael-schwarz, line 94: If you rewrite this as a `match` that will probably be nicer. Also, why does this set all analyses to not be path-sensitive?

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,87 @@
...
+
+(** Analysis with the k-callsting approach, which uses the last k callstack elements as context.
+    With the CT argument it is possible to specify the type of the callstack elements*)
+module Spec (CT:Callstack_Type) : MCPSpec = 
+struct
+  include Analyses.IdentitySpec
+
+  (* simulates a call stack of depth k*)
+  module CallStack = struct
+    include Printable.PQueue (CT)
```
michael-schwarz, line 21: Probably one could have an even more efficient data structure here given the size is known. I was thinking along the lines of a ring buffer:

https://ocaml.org/p/containers/2.8.1/doc/CCRingBuffer/module-type-S/index.html

(This basically boils down to an array & a pointer, but this is likely not the most pressing thing here)

src/framework/constraints.ml
```diff
@@ -497,6 +497,90 @@ struct
...
+                 ; split = (fun d es -> ctx.split (d, cg_val ctx) es)
+                 ; context = (fun () -> Option.get (fst (ctx.context ())))}
+
+  let enter ctx r f args =
+    let liftmap_tup = List.map (fun (x,y) -> (x, cg_val ctx), (y, max 0 (cg_val ctx - 1))) in
+    liftmap_tup (S.enter (conv ctx) r f args) 
+
+  let threadenter ctx ~multiple lval f args = 
+    let liftmap f = List.map (fun (x) -> (x, max 0 (cg_val ctx - 1))) f in
+    liftmap (S.threadenter (conv ctx) ~multiple lval f args)
```
michael-schwarz, line 562: Why for threads?

src/framework/constraints.ml
```diff
@@ -497,6 +497,90 @@ struct
...
+module NoContext = struct let name = "no context" end
+module IntConf = 
+struct
+  let n () = get_int "ana.context.ctx_gas_value" + 1
+  let names x = Format.asprintf "%d" x
+end
+
+(** Lifts a [Spec] with the context gas variable. For every function call the context gas is reduced. 
+    If the context gas is 0, the remaining function calls are analyzed context insensitively (before the analysis is context sensitive)*)
+module ContextGasLifter (S:Spec)
```
michael-schwarz, line 509: The way I understand the current implementation, the context gas is global here in the sense that it limits the general depth of the callstack? is that understanding correct?
Have you considered having a per-function context gas? This would tame recursion without losing precision for non-recursive programs...

src/analyses/callstringAnalysis.ml
SchiJoha: Sure, maybe something like "newEle",  "newStackEle", "stackEle". Or do you have a suggestion?

src/framework/constraints.ml
SchiJoha: Yes, that should be correct. The context gas is always decreased when entering a function, but inside a function, the value stays the same. 
For example, if `main `has context gas `5` and calls function `f`, and function `g`, both functions are called with a context gas value of `4`. 

I also thought of that. Would this per-function context gas be globally, so across different paths? I could imagine that the accuracy is less balanced across paths... But I will discuss this approach with Julian in our meeting today :)

src/analyses/callstringTracking.ml
SchiJoha: We don't

src/analyses/callstringTracking.ml
SchiJoha: is `stack `ok?

src/analyses/callstringAnalysis.ml
michael-schwarz: Discussed this with @jerhard yesterday, seems like there's an argument to be made for either choice.

src/analyses/callstringTracking.ml
michael-schwarz: Sure, it's not a keyword.

src/analyses/callstringTracking.ml
michael-schwarz: As above.

src/framework/constraints.ml
michael-schwarz: Same as above.

src/analyses/mCP.ml
SchiJoha: This comes from a recent unsound issue we discovered (see Issue #1302). When analyzing a function that is called with 2 different values, 2 different paths and the same context, the values are merged and can lead to "bottom" in the apron analysis. 
Disabeling the path sensitivity was a "workaround" for this problem, but @jerhard and I are currently trying to fix this.

src/analyses/callstringAnalysis.ml
jerhard: As far as I can see, the `exitstate` is the start state of with which the `exitfuns` are analyzed, i.e. functions that are assumed to run at the end of the analyzed program. For many analyses, this is the `top` of the corresponding domain, since nothing is really known about the state when those functions are executed. Here however, putting the empty callstring seems to be the right thing.

michael-schwarz: Thanks! Maybe we can document this in `spec.ml` now?

src/analyses/mCP.ml
SchiJoha: We could detect the issue, and revert this change (so now the call-string analysis and the loopfree-callstring analysis are path-sensitive).
See  #1354

michael-schwarz:
We probably should tag a branch with the version used for the experiments and make a pre-release akin to what we did for `longjmp`[1] and a lightweight artifact as well. Not sure if we one cut a pre-release from a branch on a fork.

If not, we might have to pull this over to a branch on this repo.

1: https://github.com/goblint/analyzer/releases/tag/v2.1.0-longjmp

src/config/options.schema.json
```diff
@@ -963,6 +963,30 @@
...
+              "description": "If enabled, the first x recursions of each path are analyzed context-sensitive, the remaining context-insensitive. x denotes the context gas, which is defined in ana.context.ctx_gas_value.",
+              "type": "boolean",
+              "default": false
+            },
+            "ctx_gas_value": {
+              "title": "ana.context.ctx_gas_value",
+              "description": "Number of calls that are analyzed context-sensitive. Used for the ContextGasLifter, which is enabled with ana.context.ctx_gas. Must be >= 0!",
+              "type": "integer",
+              "default": 10
+            },
```
sim642, line 978: These should be one option like `ana.context.gas` with the numeric value (by default 0), where 0 would bean it's deactivated.
We have other options that work similarly.

src/config/options.schema.json
```diff
@@ -963,6 +963,30 @@
...
+              "title": "ana.context.inf_callStack",
+              "description": "If true, the call-stack of the call.string analysis is infinite.",
+              "type": "boolean",
+              "default": false
+            },
+            "callStack_height": {
+              "title": "ana.context.callStack_height",
+              "description": "Height of the call-stack that should be used as context for the call-string analysis. Must be >= 0!",
+              "type": "integer",
+              "default": 10
```
sim642, line 989: Similarly here, these could just be one option.

src/analyses/mCP.ml
```diff
@@ -65,6 +65,12 @@ struct
     let deps (x,_) = iter (check_dep x) @@ (find_spec x).dep in
     iter deps xs
 
+  let specific_cont_sens xs =
+    (* most analysis must be set to context-insensitive, because we only want to analyse context-sensitive for one specific analysis *)
+    let sens_ana = ["callstring_fundec"; "callstring_stmt"; "callstring_loc"; "loopfree_callstring"] in
+    let enabled = List.fold_left (fun acc x -> acc || (mem x xs)) false sens_ana in
+    (* returns the edited list of insensitive analyses if enabled *)
+    if enabled then Some(filter (fun x -> not (mem x sens_ana)) xs) else None
```
sim642, line 73: I don't think we need such magic behavior where activating some analysis will quietly change our usual context-sensitivity configuration. If the user wants only callstring contexts without anything else, then they can configure that with existing options.

src/analyses/loopfreeCallstring.ml
```diff
@@ -0,0 +1,88 @@
...
+    let printXml f = function
+      | `Lifted x -> 
+        begin
+          match x with 
+          | `Left x -> BatPrintf.fprintf f "<value>stack:\n%a</value>" FundecList.printXml x
+          | `Right x -> BatPrintf.fprintf f "<value>set:\n%a</value>" FundecSet.printXml x
+        end;
+      | _ -> failwith "Error loopfreeCallstring (printXml): the Flat Lattice containing Either shouldn't be `Top or `Bottom!"
+  end
+  module D = Lattice.Liszt (Either)
```
sim642, line 26: This reimplements the prefix-set data structure we already have in `ThreadIdDomain.History`. We should extract a common functor which can be reused for both to minimize the duplication.

Actually, the structure here is different: it's a list of eithers which could contain multiple lists and multiple sets. Is that intentional? If so, then what's the meaning of that? It wasn't so in the ESOP paper. If not, then this should just use the `ThreadIdDomain` structure whose type enforces that there aren't multiple lists or sets mixed.

Is there a reason this cannot use anything other than `fundec`s?
Conceptually, it's yet another stack abstraction like k-bounded stacks, etc. I think the prefix-set abstraction should fit into the other callstring analysis being proposed here.
That's what has been already done with the old analyses in `StackDomain`.

src/analyses/callstringAnalysis.ml
```diff
@@ -0,0 +1,89 @@
...
+  let stackTypeName = "loc"
+  let stackEle f ctx =
+    Some !Goblint_tracing.current_loc
+end
+
+(* Lifters for the call-string approach with different call-stack element types *)
+let _ =
+  MCP.register_analysis (module Spec (Fundec) : MCPSpec); (* name: callstring_fundec *)
+  MCP.register_analysis (module Spec (Stmt) : MCPSpec); (* name: callstring_stmt *)
+  MCP.register_analysis (module Spec (Location) : MCPSpec) (* name: callstring_loc *)
```
sim642, line 89: How do these compare to some similar ancient analyses in `stackTrace.ml`?
If they're subsumed by the new ones, we shouldn't keep the old ones around.

src/framework/constraints.ml
sim642: > If the context gas is 0, the remaining function calls are analyzed context insensitively (before the analysis is context sensitive)

This also seems surprising: when context gas is enabled, all other context-sensitivity configuration is silently ignored.
I would expect it to degenerate to whatever contexts `S.C` specifies. One can get this behavior by just making everything in `S.C` context-insensitive, but would be more general.

This also begs the question, why is this a `Spec` lifter, while the callstrings are actual analyses? It seems to me that this is also a mere stack abstraction, where the stack is abstracted just to its size, bounded by gas (which would collapse all deeper calls together when gas runs out). This is basically `Lattice.Chain`.

michael-schwarz:
Thanks for the reviews @sim642, we are currently in a bit of a time crunch to make it for SOAP, but we're factoring in some places where the intent of our approach was unclear, and will address the comments in detail some time after the deadline :smile: 

(And when I say we will answer I mostly mean @SchiJoha probably :stuck_out_tongue:)

tests/regression/81-loopfree_callstring/01-loop_calling_sens.c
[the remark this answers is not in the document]
SchiJoha: I agree

src/analyses/loopfreeCallstring.ml
SchiJoha: Yes it is intentional that sets and lists (now I updated it to elements) are mixed in one list. The Loopfree callstring approach is similar to the call string approach and stores the call stack containing the prior function calls. Our idea was to improve this approach by storing any loop occurring in this call stack by a set. So in case of the call-sequence `main -> f -> f` the first call to `f` has the context as for the call string approach `(main, f)`, and the second call to `f` has the context `(main, {f})`, where the set represents the recursive call. In case the call-sequence has more `f`s in it, the context of all those comming `f` is still `(main, {f})` and consequently, the number of contexts is reduced. And this is basically the goal, to reduce the number of contexts during analyzation to improve runtime and hopefully not loose to much precision. 
Also for bigger loops this would hold: `main -> f -> g -> h -> f` would behave similar to the call string approach till the second function call of `f`. For this second call, the context `(main, {f, g, h})` would be the context.

src/analyses/mCP.ml
jerhard: >If the user wants only callstring contexts without anything else, then they can configure that with existing options.

This is true, however, it is not really user-friendly to configure, since one has to add all the other analyses into the set of context-insensitive analyses. This is even more cumbersome to do when one also uses the auto-tuner, which may itself activate some analyses.

I think there should at least be an easy option for a user to let the implementation behave as here.

src/analyses/loopfreeCallstring.ml
SchiJoha: The current idea was to make it similar to the call string approach of Sharir and Pnueli, therefore only the `fundecs` are used as elements. Maybe one can think about using the statements as alternative input and use it as a second approach? @jerhard do you think this would make sense?

sim642: The examples you gave now are all pairs of a prefix and a set, which is why I would expect the type to also be such a pair. My point is that the current type also contains values like `[[main], {f, g, h}, [foo, bar], [baz], {ping, pong}]` because it's a list of either prefix or set. Are such things actually used?

SchiJoha: > Conceptually, it's yet another stack abstraction like k-bounded stacks, etc. I think the prefix-set abstraction should fit into the other callstring analysis being proposed here.
> That's what has been already done with the old analyses in StackDomain.
I am currently doing some changes on loopfree, and can check if it would be possible to reuse the `StackDomain`

SchiJoha: Yes they are all used. For example for the call-sequence: `main -> f -> f -> g -> g` we would compute for the last `g` the context `[main, {f}, {g}]`. Or if we don't have any recursion `main -> f -> g` it would result in `[main, f, g]`.

SchiJoha: In the current code the stack only contains `Fundecs` and sets of `Fundecs`, since this makes the implementation simpler and more efficient. So there are no lists anymore inside the stack

src/config/options.schema.json
SchiJoha: This would take away the option to configure a gas value of `0`, which is equivalent to an context-insensitve analysis. If one don't want this "fast configureable"  context-insensitive option for context gas, I can configure it as you mentioned.

src/analyses/mCP.ml
sim642: I agree `ana.ctx_insens` is a bit backwards and inconvenient for this. Maybe we could have two complementary options:
1. `ana.ctx_insens` (like a blacklist) as it is right now.
2. `ana.ctx_sens`/`ana.ctx_only` (like a whitelist) for a list of _only_ analyses that should be context-sensitive.

Both cannot be used at the same time, so the logic could be that if `ana.ctx_sens` is empty (as by default), then `ana.ctx_insens` is used, otherwise only `ana.ctx_sens` is considered.

src/analyses/loopfreeCallstring.ml
sim642: Oh, so this is a generalization of the history-based thread IDs we used in the ESOP paper. Even so, both variations should fit the same abstraction of stacks. That would allow using this generalization also for thread IDs.

If we've come up and implemented all these different stack abstractions, it would be a shame to needlessly hard-code each one to a different element type or use case.

src/config/options.schema.json
sim642: Sounds reasonable. The special disabled value could just be -1 then. Actually, that's what `incremental.restart.sided.fuel` even does, which is also fuel/gas for another purpose.

src/analyses/loopfreeCallstring.ml
sim642: > the tids do not just abstract the creation stack, but also have this C component of encountered create edges

Those are just part of the `threadid` analysis domain. `ThreadIdDomain.History`, which is the stack abstraction has nothing about edges.

michael-schwarz: > which is the stack abstraction has nothing about edges

It does not directly track this information but the logic is in `ThreadidDomain.History` nevertheless (but not in `compose`). 

~~~OCaml
  let threadenter ~multiple ((p, _ ) as current, (cs,_)) (n: Node.t) i v =
    let ni = Base.threadenter ~multiple n i v in
    let ((p', s') as composed) = compose current ni in
    if is_unique composed && (S.mem ni cs || multiple) then
      [(p, S.singleton ni); composed] (* also respawn unique version of the thread to keep it reachable while thread ID sets refer to it *)
    else
      [composed]
~~~

All I am saying is that we have to see how much re-use makes sense without the code getting overly complicated. The creation reasoning for tids for example is already quite involved, so I am not sure if generalizing that is a good idea.

src/config/options.schema.json
SchiJoha: That sounds good!  
I update it in the code

src/analyses/mCP.ml
SchiJoha: I am currently at the realization of this, but with those two options it's not possible to configure a context-insensitive analysis easily. Because if I set the `ana.ctx_sens` to an empty array, the `ana.ctx_insens` analysis is used and would need to contain all activated analyses. Currently it's possible to make the analysis context-insensitive with the context gas analysis and a gas of 0. 
Should I include an separate option or a string for the array (for example `all` in the `ana.ctx_insens` analysis to set all activated analyses insensitive), which allows it to make the analysis context-insensitive?

sim642: That's a good point. I would say it's fine because we never had a way to make a fully context-insensitive analysis anyway (other than adding everything to `ana.ctx_insens`, which will still work) and context gas 0 is another way.

In the future, we could make the schema more flexible (e.g. allow `null` values for a special meaning) or use `anyOf`s, if we really want a clean way to do that.

For a fully context-insensitive analysis, one could even imagine defining an optimized constraint system whose local unknowns are just CFG nodes, not pairs where the second context component will be effectively degraded to unit.

src/analyses/callstringAnalysis.ml
SchiJoha: They are not the same, mainly because the types are different. We use `Statements` and `Fundecs`, and they use `Locations` and `Variables`. If I understand it correctly, the `[stack_loc]` could be recreated with our callstring configuration by just adding a new instance of the `Callstack_Type`. The `[stack_trace_set]` uses a set instead of a set and therefore is not reconstructable with our approach. I don't fully understand the `[stack_trace]` and why all those functions are needed, but I think it would be possible to fit it into our implementation.

src/framework/constraints.ml
SchiJoha: The idea behind the context gas approach is to use the normal context-sensitive analyzer of Goblint for all function calls that have a gas value greater zero. In case the gas value is zero, the function does not receive any context information (we realize this by passing `NULL` as context). It may happen that multiple function calls to the same function receive no context and are merged for the analysis. Consequently, the context-sensitivity configuration is not ignored if the gas is greater zero. And since we need to switch from passing the sensitive contexts or passing `NONE` depending on the gas value, it is necessary to implement is as lifter since we need to overwrite the context information of the analyses. 
In case the context gas value never reaches zero, the context gas analysis is equal to the standard configuration of Goblint. 

For example, we have the call-sequence `main -> f(2) -> f(1) -> f(0)` and a gas of `3`. In this case, `main` would be analyzed with gas `2´ and therefore context-sensitive. `f(2)` receives gas `1`, and is also analyzed as usual. For `f(1)` the gas reaches zero, and consequently, this call is analyzed without context information. The same happens for `f(0)`, and consequently, they have the same context and are analyzed together. 

I hope this answers all parts of the question??

michael-schwarz: c.f. https://github.com/tum-cit-pl/taming-contexts for a formal description

src/analyses/loopfreeCallstring.ml
sim642: That's true, history thread IDs require a few more operations than `compose`/`add` for the stack abstraction: `is_unique`, `is_must_parent`, `may_create`, but these can probably be recast as generic stack properties.
There's no need to do that here though. I now understand how this abstraction is different from the history thread IDs one.

src/analyses/mCP.ml
```diff
@@ -120,7 +115,7 @@ struct
   let context fd x =
     let x = spec_list x in
     filter_map (fun (n,(module S:MCPSpec),d) ->
-        if mem n !cont_inse then
+        if (mem n !cont_inse) || not @@ (List.is_empty !cont_sens || mem n !cont_sens) then (* TODO: do I need to check if it is contained in activated ana?*)
```
sim642, line 118: Can't we just check `activated_ctx_sens` here?

src/analyses/callstringAnalysis.ml
jerhard: Whether one tracks functions or statements  influences the type of the analysis domain,  so formulating the analysis as a functor and instantiating it with the appropriate module needed for the domain seems  reasonable. 

One could wrap the analyses modules as they are defined here into one analysis that then delegates all transfer functions to one of the modules, depending on a configuration. The advantage would be that for the configuration one would only have one analysis and an option to configure the domain, as you suggest. Do you think this would be cleaner?

src/analyses/callstringAnalysis.ml
jerhard: >uses a set instead of a set

Typo?

SchiJoha: Yes. I meant a set instead of a list :) Thanks

michael-schwarz:
Is this ready for a new review (and merging) or are there still things you're working on>

SchiJoha:
> Is this ready for a new review (and merging) or are there still things you're working on?

Yes! It should be ready for review :)

gobview
michael-schwarz, line 1: Why is there a change to Gobview that is part of this PR? Did you have to change things in there?

src/config/options.schema.json
```diff
@@ -970,6 +977,18 @@
...
+              "title": "ana.context.gas_value",
+              "description": "x denotes the gas value for the ContextGasLifter. Negative value means deactivated, zero means context-insensitve analysis. If enabled, the first x recursive calls of the call stack are analyzed context-sensitive, the remaining with no context.",
+              "type": "integer",
+              "default": -1
+            },
+            "callString_length": {
+              "title": "ana.context.callString_length",
+              "description": "Length of the call string that should be used as context for the Call String analysis. In case the value is zero, the analysis is context-insensitive. For a negative value an infinite call string is used!",
+              "type": "integer",
+              "default": 2
```
michael-schwarz, line 991: Does this change our default behavior? or does some other analysis need to be activated for this to take effect? If yes, it should be mentioned here.

gobview
sim642: Nothing in GobView changed, but I think this is fine: currently Goblint `master` has GobView at the end of https://github.com/goblint/gobview/pull/44 branch (before merge), this just updates it to its merge commit.

src/analyses/loopfreeCallstring.ml
```diff
@@ -1,6 +1,13 @@
...
-
+(** 
+   Loopfree Callstring analysis [loopfree_callstring] that reduces the call string length of the classical Call String approach for recursions
+   The idea is to improve the Call String analysis by representing all detected call cycle of the call string in a set 
+   In case no call cycles appears, the call string is identical to the call string of the Call String approach
+   For example: 
+   - call string [main, a, b, c, a] is represented as [main, {a, b, c}]
+   - call string [main, a, a, b, b, b] is represented as [main, {a}, {b}]
+
+   This approach is inspired by @see <https://arxiv.org/abs/2301.06439> Schwarz, M., Saan, S., Seidl, H., Erhard, J., Vojdani, V. Clustered Relational Thread-Modular Abstract Interpretation with Local Traces. Appendix F.
```
sim642, line 9:
```suggestion
   This approach is inspired by
   @see <https://arxiv.org/abs/2301.06439> Schwarz, M., Saan, S., Seidl, H., Erhard, J., Vojdani, V. Clustered Relational Thread-Modular Abstract Interpretation with Local Traces. Appendix F. 
```

sim642 (approved):

michael-schwarz (approved):

michael-schwarz:
@jerhard @DrMichaelPetter do you still want to review this or should we merge it?

src/config/options.schema.json
```diff
@@ -356,11 +356,18 @@
...
           "title": "ana.ctx_insens",
-          "description": "List of context-insensitive analyses",
+          "description": "List of context-insensitive analyses, which is ignored if `ctx_sens` is not empty.",
           "type": "array",
           "items": { "type": "string" },
           "default": [ "stack_loc", "stack_trace_set" ]
         },
+        "ctx_sens": {
+          "title": "ana.ctx_sens",
+          "description": "List of context-sensitive analyses. In case of an empty array `ctx_insens` is used, otherwise `ctx_insens` is ignored.",
```
jerhard, line 366:
```suggestion
          "description": "List of context-sensitive analyses. In case this list is empty, `ana.ctx_insens` will be used to determine the set of context insensitive analyses.",
```

src/config/options.schema.json
```diff
@@ -970,6 +977,18 @@
                 "Do widening on contexts. Keeps a map of function to call state; enter will then return the widened local state for recursive calls.",
               "type": "boolean",
               "default": false
+            },
+            "gas_value": {
+              "title": "ana.context.gas_value",
+              "description": "x denotes the gas value for the ContextGasLifter. Negative value means deactivated, zero means context-insensitve analysis. If enabled, the first x recursive calls of the call stack are analyzed context-sensitive, the remaining with no context.",
```
jerhard, line 983:
```suggestion
              "description": "Denotes the gas value x for the ContextGasLifter. Negative value deactivate the context gas, zero yields a context-insensitve analysis. If enabled, the first x recursive calls of the call stack are analyzed context-sensitively. Any calls deeper in the call stack are analyzed with the same context.",
```

src/config/options.schema.json
```diff
@@ -970,6 +977,18 @@
...
+              "type": "integer",
+              "default": -1
+            },
+            "callString_length": {
+              "title": "ana.context.callString_length",
+              "description": "Length of the call string that should be used as context for the call_string and/or call_site analyses. In case the value is zero, the analysis is context-insensitive. For a negative value, an infinite call string is used! For this option to be effective, one analysis of the `callstring.ml` file must be activated.",
```
jerhard, line 989:
```suggestion
              "description": "Length of the call string that should be used as context for the call_string and/or call_site analyses. In case the value is zero, the analysis is context-insensitive. For a negative value, an infinite call string is used! For this option to have an effect, one of the analyses in `callstring.ml` must be activated.",
```

jerhard (commented):
Looks good! I only have some cosmetical suggestions for the code.

PR #1343, opened 2024-01-27
Refine Points-To set on locking a mutex

### LINKED ISSUE #1287: Refine Points-to-Sets in Base when Splitting Paths for Mutex Analysis
As proposed  by @hseidl and @sim642 yesterday at Gobcon, we should also refine the pointer information in cases where we perform a state splitting for the mutex analysis.

This would allow us to not produce warnings such as

> [Warning][Unknown] unlocking mutex (m1) which may not be held
> [Warning][Unknown] unlocking mutex (m2) which may not be held

~~~C
#include<pthread.h>

pthread_mutex_t m1;
pthread_mutex_t m2;

int main(int argc, char const *argv[])
{
    int top;
    pthread_mutex_t* ptr;
    ptr = &m1;

    if(top) {
        ptr = &m2;
    }

    pthread_mutex_lock(ptr);
    pthread_mutex_unlock(ptr); //NOWARN

    return 0;
}
~~~

This will be helpful not just for this category of warnings, but will also allow shrinking may-locksets back down, which may be crucial for the deadlock analysis.

michael-schwarz:
Make use of the fact that we split upon acquiring mutexes anyway and refine points-to-sets for mutexes. This means getting rid of spurious warnings about releasing mutexes that are not held.

Closes #1287.


src/analyses/mutexEventsAnalysis.ml
```diff
@@ -24,15 +24,15 @@ struct
     match lv_opt with
     | None ->
       Queries.AD.iter (fun e ->
-          ctx.split () [Events.Lock (e, rw)]
+          ctx.split () [Events.Lock (e, rw); Events.RefinePointerExp {exp = arg; ad = Queries.AD.singleton e}]
```
sim642, line 27: Why is a new special-purpose refinement event needed? Doesn't the general `SplitBranch` one work? Basically to `branch` on `arg == e` (or `arg == &e` maybe?). It's a singleton set, so it should work.

tests/regression/79-mutex2/01-split.c
sim642, line 1: 04-mutex/96 and 97 are still unused, so we could squeeze them in there.

sim642 (approved):

PR #1344, opened 2024-01-28
Handle `LNot` for float in forward evaluation and refinement on guards

### LINKED ISSUE #1211: `LNot` on floating-point values
## Refine
Extracted from concrat/axel, Goblint crashes on this:
```c
int main() {
  double x;
  if (!x) // TODO NO CRASH: Invalid_argument("Cilfacade.get_ikind: non-integer type double ")
    return 1;
  return 0;
}
```
The exception comes from here:
https://github.com/goblint/analyzer/blob/8468a5a676fae48a82e0284ea13170d0cefa935c/src/analyses/baseInvariant.ml#L570-L581

The problem is that even though the result type of the `LNot` expression is always integer, `e` underneath might be a float.

## Eval
The forward evaluation of `LNot` on floats is also incorrect:
https://github.com/goblint/analyzer/blob/8468a5a676fae48a82e0284ea13170d0cefa935c/src/analyses/base.ml#L186-L194
It returns a `Float` abstraction even though for `LNot` it should return an `Int` abstraction. This mismatch can cause a similar issue as #1209 when a `Float` is stored for an integer variable.

michael-schwarz:
Fixes the issues identified by @sim642 where 
- the type of the abstract values for expressions involving `LNot` was wrong
- having `!x` for float variables `x` in a guard lead to crashes

Closes #1211.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1645,6 +1645,9 @@ struct
     module V = V
     module G = G
 
+    let unop_ID = unop_ID
+    let unop_FD = unop_FD
+
     let eval_rv = eval_rv
     let eval_rv_address = eval_rv_address
     let eval_lv = eval_lv
@@ -2841,6 +2844,9 @@ struct
           module G = G
 
           let ost = octx.local
+
+          let unop_ID = unop_ID
+          let unop_FD = unop_FD
 
           (* all evals happen in octx with non-top values *)
           let eval_rv ~ctx st e = eval_rv ~ctx:octx ost e
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -179,8 +179,8 @@ struct
   let unop_FD = function
     | Neg  -> (fun v -> (Float (FD.neg v):value))
     | LNot -> (fun c -> Int (FD.eq c (FD.of_const (FD.get_fkind c) 0.)))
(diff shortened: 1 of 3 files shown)


src/analyses/base.ml
```diff
@@ -178,14 +178,15 @@ struct
     | LNot -> ID.lognot
 
   let unop_FD = function
```
sim642, line 180: There's also a copy in `BaseInvariant` apparently. Maybe these could be deduplicated in `BaseUtil`, `ValueDomain` or something?

src/analyses/base.ml
```diff
@@ -178,14 +178,15 @@ struct
...
-    | Neg  -> FD.neg
+    | Neg  -> (fun v -> (Float (FD.neg v):value))
+    | LNot -> (fun c -> Int (FD.eq c (FD.of_const (FD.get_fkind c) 0.)))
     (* other unary operators are not implemented on float values *)
-    | _ -> (fun c -> FD.top_of (FD.get_fkind c))
+    | _ -> (fun c -> Float (FD.top_of (FD.get_fkind c)))
```
sim642, line 184: Since `BNot` is the only remaining option, we should just explicitly match that and say it isn't type-valid anyway on a float.

sim642 (approved):

PR #1349, opened 2024-02-01
Remove intermediate module definition in `AffineEqualityAnalysis`

jerhard:
Clean up some relational code by making `AffineEqualityDomain.D2` satisfy the `RD` signature directly. This allows removing a separate module definition in `AffineEqualityAnalysis`.

### DIFF
--- a/src/analyses/apron/affineEqualityAnalysis.apron.ml
+++ b/src/analyses/apron/affineEqualityAnalysis.apron.ml
@@ -9,16 +9,10 @@ include RelationAnalysis
 let spec_module: (module MCPSpec) Lazy.t =
   lazy (
     let module AD = AffineEqualityDomain.D2 (VectorMatrix.ArrayVector) (VectorMatrix.ArrayMatrix) in
-    let module RD: RelationDomain.RD =
-    struct
-      module V = AffineEqualityDomain.V
-      include AD
-    end
-    in
     let module Priv = (val RelationPriv.get_priv ()) in
     let module Spec =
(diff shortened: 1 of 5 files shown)


src/analyses/apron/affineEqualityAnalysis.apron.ml
```diff
@@ -9,16 +9,10 @@ include RelationAnalysis
 let spec_module: (module MCPSpec) Lazy.t =
   lazy (
     let module AD = AffineEqualityDomain.D2 (VectorMatrix.ArrayVector) (VectorMatrix.ArrayMatrix) in
-    let module RD: RelationDomain.RD =
-    struct
-      module V = AffineEqualityDomain.V
-      include AD
-    end
-    in
```
sim642, line 17: `ApronAnalysis` contains a similar thing. Is the same simplification also possible there then?

jerhard: I  now did a similar thing for the `ApronAnalysis` although there the signatures there are a bit more convoluted, in that `ApronAnalysis.S2` and `S3` are not sub-signatures in either direction.

sim642 (approved):
I suppose this might cause some small conflicts for #1297 and #1330?

michael-schwarz (approved):

michael-schwarz:
> I suppose this might cause some small conflicts for #1297 and #1330?

I guess so, yes. However hopefully these should be very easy to solve.

PR #847, opened 2022-10-06
Automate manual tests

sim642:
Now that we have cram from #787, it's a good time to use it to automate previously manual tests.

### Changes
1. ~~Add tests for YAML witness generation. The `yamlWitnessStrip` executable is used to remove non-deterministic metadata parts from the output.~~ (#1357)
2. ~~Fix `ana.apron.invariant.one-var` not filtering enough.~~ (Cherry-picked before)
3. Make `update_suite.rb` print test names that have no automatic checks. Add `NOCRASH`, `FIXPOINT`, `NOTIMEOUT` and `CRAM` annotations to suppress that by indicating which implicit aspect is being tested.
4. Add such suppressing annotations to tests where the reason for no explicit checks is apparent from comments or git history.

### TODO
- [x] #855
- [x] ~~Add missing annotations:~~ (existing ones suppressed with `NOCHECK`)

   <details>

    ```
    No automatic checks in 01/51 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 02/57 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 03/03 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 03/22 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 13/62 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 13/65 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 18/22 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 18/23 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 18/24 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 19/01 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 20/01 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 20/02 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
    No automatic checks in 20/03 (maybe NOCRASH/FIXPOINT/NOTIMEOUT/CRAM?)
[...]

PR #1360, opened 2024-02-09
Add cram tests for CFGs

sim642:
This cherry-picks some commits from #847, but uses `graph-easy` to visualize the `.dot` output instead of testing the Graphviz code directly.

For end-to-end YAML witness location testing, #1357 is what we finally want, it's not super convenient and general to have to check for node locations through YAML witnesses. This is slightly more direct by both testing the CFG structure (which we've had problems about) and all node locations simultaneously.

### TODO
- [x] Get rid of `__goblint_dummy_init`, at least in these tests.
- [x] Go back to per-function CFGs in these tests because 00-sanity/21-empty-loops output is massive only due to the big switch in `main`, which is orthogonal to the test.
- [x] Fix GobView (https://github.com/goblint/gobview/pull/41).


.github/workflows/coverage.yml
```diff
@@ -38,6 +38,11 @@ jobs:
         uses: ocaml/setup-ocaml@v2
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
+          opam-depext-flags: --with-test # doesn't work
```
michael-schwarz, line 41: If it doesn't work why still pass it then?

tests/regression/cfg/util/cfgDot.ml
```diff
@@ -0,0 +1,73 @@
...
+        List.map (fun one_inst -> GoblintCil.mkStmtOneInstr one_inst) il in
+      let block = GoblintCil.mkBlock list_of_stmts in
+      ChangeDoChildrenPost(s, (fun _ -> s.skind <- Block(block); s))
+    | _ -> DoChildren
+
+  method! vvdec _ = SkipChildren
+  method! vexpr _ = SkipChildren
+  method! vlval _ = SkipChildren
+  method! vtype _ = SkipChildren
+end
```
michael-schwarz, line 17: Any particular reason this cannot be reused from `CilCfg`?

Such duplication is a bit bad because it may cause the test program and our actual program to go out of sync, which would mean tests still passing even if things are broken.

src/common/util/cilLocation.ml
```diff
@@ -0,0 +1,52 @@
...
+  eloc: location;
+}
+
+let get_labelLoc = function
+  | Label (_, loc, _) -> {loc; eloc = locUnknown}
+  | Case (_, loc, eloc) -> {loc; eloc}
+  | CaseRange (_, _, loc, eloc) -> {loc; eloc}
+  | Default (loc, eloc) -> {loc; eloc}
+
+(* let rec get_labelsLoc = function
```
michael-schwarz, line 14: What's up with this commented out snippet?

.github/workflows/coverage.yml
sim642: It's an opam bug that eventually should be fixed: https://github.com/ocaml/opam/issues/5836. The additional `apt install` step in our CI should really be redundant since there's also a `depext` for it.

tests/regression/cfg/util/cfgDot.ml
sim642: `CilCfg` has some things that depend on `Autotune` which is (and has to be) in `goblint.lib`. I'll see if I can do anything about it.

sim642: I extracted `CilCfg0` for now.

PR #1362, opened 2024-02-10
Add option `witness.invariant.all-locals` to only print locals definitely in scope

### LINKED ISSUE #1361: Witnesses: Filter invariants about variables possibly not in scope
@schuessf from the Ultimate team at the University of Freiburg noticed that our witnesses often contain invariants about local variables that are not in scope.
We should add an option to filter these out using the attributes we introduced to CIL in https://github.com/goblint/cil/pull/155.

michael-schwarz:
Closes #1361.


sim642:
It would be good to have tests for this, but I'll have to get #1357 ready for that.

Or a quick temporary alternative is to have a cram test just for Goblint output (not the witness itself) to test for the printed number of generated witness entries changing due to this option.

michael-schwarz:
Maybe this is worth merging ahead of the tests, so we can produce new witnesses for our colleagues in Freiburg to experiment with?

sim642 (approved):

PR #1365, opened 2024-02-15
Add `context_to_yojson` in messages

karoliineh:
so that the big `context` objects would not crowd the JSON and cause `java.lang.OutOfMemoryError: Java heap space`


sim642 (approved):
Even if this isn't enough to fix the GobPie memory error, it brings JSON messages output closer to the printed one, allowing GobPie to match it.

PR #1367, opened 2024-02-16
Move all tests to `dune runtest`

sim642:
Closes #1359.

### TODO
- [x] Add fine-grained dune aliases for subsets of tests.

Issue #1359, opened 2024-02-09
Define a `make` target for running relational tests


michael-schwarz:
Currently, one has to run a ~5 sets separately after making changes to something related to apron. This is inconvenient.

sim642:
I've wanted to do something even broader for a while: set up dune rules such that `dune runtest` would run all the tests we have. `make test` could just be an alias for it.
Dune does dependency tracking, so it does the minimum amount of work required to rerun everything. Our `make` doesn't because it just delegates to a shell script which is oblivious and would always redo everything.
Also `dune runtest` guarantees that you don't forget to rebuild Goblint before running the tests.

PR #1368, opened 2024-02-19
Clean up old `make` targets

sim642:
While working on #1359 and #1367, I was once again reminded of all the old broken or unused `make` targets (or more precisely their `bash` imitations).

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -35,10 +35,6 @@ rule() {
     ;; view)
       eval $(opam config env)
       dune build gobview
-    # alternatives to .exe: .bc (bytecode), .bc.js (js_of_ocaml), see https://dune.readthedocs.io/en/stable/dune-files.html#executable
-    ;; js) # https://dune.readthedocs.io/en/stable/jsoo.html
-      dune build $TARGET.bc.js &&
-      node _build/default/$TARGET.bc.js
     ;; watch)
       eval $(opam config env)
       # dune build -w $TARGET.exe
--- a/make.sh
+++ b/make.sh
@@ -109,8 +109,6 @@ rule() {
     ;; setup_gobview )
       [[ -f gobview/gobview.opam ]] || git submodule update --init gobview
       opam install --deps-only --locked gobview/
-    # ;; watch)
-    #   fswatch --event Updated -e $TARGET.ml src/ | xargs -n1 -I{} make
     ;; install)
       eval $(opam config env)
       dune build @install
--- a/make.sh
+++ b/make.sh
@@ -59,8 +59,6 @@ rule() {
       dune build goblint.byte &&
       rm -f goblint.byte &&
       cp _build/default/goblint.byte goblint.byte
-    # ;; tag*)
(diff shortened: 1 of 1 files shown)


make.sh
```diff
@@ -17,7 +17,6 @@ rule() {
     clean)
       git clean -X -f
       dune clean
-    ;; gen) gen
```
michael-schwarz, line 20: What was this even supposed to do?

make.sh
```diff
@@ -101,8 +94,6 @@ rule() {
       tar xf master.tar.gz && rm master.tar.gz
       rm -rf linux-headers && mv linux-headers-master linux-headers
       for n in $(compgen -c gcc- | sed 's/gcc-//'); do if [ $n != 5 ]; then cp -n linux-headers/include/linux/compiler-gcc{5,$n}.h; fi; done
-    ;; lock)
-      opam lock
```
michael-schwarz, line 105: Why is this outdated?

make.sh
```diff
@@ -151,16 +140,6 @@ rule() {
...
-    ;; travis) # run a travis docker container with the files tracked by git - intended to debug setup problems on travis-ci.com
-      echo "run ./scripts/travis-ci.sh to setup ocaml"
-      # echo "bind-mount cwd: beware that cwd of host can be modified and IO is very slow!"
-      # docker run -it -u travis -v $(pwd):$(pwd):delegated -w $(pwd) travisci/ci-garnet:packer-1515445631-7dfb2e1 bash
-      echo "copy cwd w/o git-ignored files: changes in container won't affect host's cwd."
-      # cp cwd (with .git, _opam, _build): 1m51s, cp ls-files: 0.5s
-      docker run -it -u travis -v `pwd`:/analyzer:ro,delegated -w /home/travis travisci/ci-garnet:packer-1515445631-7dfb2e1 bash -c 'cd /analyzer; mkdir ~/a; cp --parents $(git ls-files) ~/a; cd ~/a; bash'
-    ;; server)
-      rsync -avz --delete --exclude='/.git' --exclude='server.sh' --exclude-from="$(git ls-files --exclude-standard -oi --directory > /tmp/excludes; echo /tmp/excludes)" . serverseidl6.informatik.tu-muenchen.de:~/analyzer2
-      ssh serverseidl6.informatik.tu-muenchen.de 'cd ~/analyzer2; make nat && make test'
```
michael-schwarz, line 163: That's a super weird relic to survive all these years :rofl:

make.sh
sim642: There used to be a `gen` bash function which it called, but that was removed way back in #455.

make.sh
sim642: Because it doesn't work and requires manual attention. Documentation describes how to do it properly: https://github.com/goblint/analyzer/blob/196e82f5a942ede12a6fac00dd9eb5365386c038/docs/developer-guide/releasing.md#L18-L21

michael-schwarz (approved):

Issue #1225, opened 2023-10-30
Witness invariants for unrolled loops are incorrect


sim642:
Our CIL AST based loop unrolling duplicates nodes for the same program point (in the literal sense). Thus we end up generating witness invariants for each node but the same location, e.g. `i == 0` and `i == 15`, which are contradictory:
```yml
- entry_type: location_invariant
  metadata:
    format_version: "0.1"
    uuid: dcd2d1a7-ae43-46a4-9ac9-528fc3df8507
    creation_time: 2023-10-30T08:34:16Z
    producer:
      name: Goblint
      version: heads/pldi-bench-0-gec49852db
      command_line: '''./goblint'' ''--conf'' ''conf/svcomp.json'' ''--enable'' ''witness.yaml.enabled'' ''--sets'' ''ana.specification'' ''/mnt/goblint-svcomp/benchexec/sv-benchmarks/c/properties/unreach-call.prp'' ''--sets'' ''exp.architecture'' ''32bit'' ''/mnt/goblint-svcomp/benchexec/sv-benchmarks/c/loop-acceleration/array_3-1.i'''
    task:
      input_files:
        - /mnt/goblint-svcomp/benchexec/sv-benchmarks/c/loop-acceleration/array_3-1.i
      input_file_hashes:
        /mnt/goblint-svcomp/benchexec/sv-benchmarks/c/loop-acceleration/array_3-1.i: 9c5d8dd6c87f471ee77fd3b765c8ecabfaf01dd976e127275ea7c589f724f472
      data_model: ILP32
      language: C
      specification: CHECK( init(main()), LTL(G ! call(reach_error())) )
  location:
    file_name: /mnt/goblint-svcomp/benchexec/sv-benchmarks/c/loop-acceleration/array_3-1.i
    file_hash: 9c5d8dd6c87f471ee77fd3b765c8ecabfaf01dd976e127275ea7c589f724f472
    line: 29
    column: 8
    function: main
  location_invariant:
    string: i == 0
    type: assertion
    format: C
- entry_type: location_invariant
  metadata:
    format_version: "0.1"
    uuid: 84342cda-192f-4411-a241-5436848150c9
    creation_time: 2023-10-30T08:34:16Z
    producer:
      name: Goblint
      version: heads/pldi-bench-0-gec49852db
[...]

PR #1369, opened 2024-02-21
Fix and refactor syntactic loop unrolling

sim642:
While working on fixing loop heads in YAML witnesses, I wanted to finally implement a cleaner fix to #1225 than #1248 by recording somewhere which statements came from loop unrolling.
To do this, I wanted to refactor things a bit and added cram tests to make sure I don't accidentally break anything. But I also discovered that `continue`s are mishandled in loop unrolling from #563.

### Changes
1. Add cram tests for loop unrolling.
2. Mark unrolled loop head labels as not originating from input code but transformations, like CIL does.
3. Fix `continue` in loop unrolling. Due to (double) `ref`s, there was undesirable sharing which caused all unrolled iterations to jump directly to the unrolled part, not to the next unrolled iteration as intended. Imperative programming is again dangerous.
4. Add option `dbg.justcil-printer` to allow cram testing `justcil` output without clutter from line directives and builtin declarations.
5. Remove excessive blocks from loop unrolling output.
6. Remove `goblint_array_domain` attribute from non-array variables.

### DIFF
--- a/src/common/util/cilfacade.ml
+++ b/src/common/util/cilfacade.ml
@@ -66,8 +66,27 @@ let parse fileName =
     E.s (E.error "There were parsing errors in %s" fileName_str);
   file
 
+class myCilPrinter =
+object
+  inherit defaultCilPrinterClass as super
+
+  method! pLineDirective ?(forcefile=false) l =
+    Pretty.nil
+
+  method! pGlobal () (g: global) =
+    match g with
+    | GVarDecl (vi, l) when Hashtbl.mem builtinFunctions vi.vname -> Pretty.nil
+    | _ -> super#pGlobal () g
+end
+
+let dumpFile (pp: cilPrinter) (out : out_channel) (outfile: string) file =
+  Pretty.printDepth := 99999;
+  Pretty.fastMode := true;
+  iterGlobals file (fun g -> dumpGlobal pp out g);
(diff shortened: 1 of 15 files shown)


michael-schwarz:
Nice catch! Do you think you can also craft an example where the better handling of `continue` means that an assertion that was previously failing now succeeds?

sim642:
There's now a test that fails on `master`.

src/common/util/cilfacade.ml
```diff
@@ -66,8 +66,32 @@ let parse fileName =
     E.s (E.error "There were parsing errors in %s" fileName_str);
   file
 
+class cleanCilPrinterClass =
```
michael-schwarz, line 69: I think it would be good to have an explanation here what the printer does differently than the normal one? I guess not print line numbers and not print `builtins`?

src/config/options.schema.json
```diff
@@ -1860,6 +1860,13 @@
           "type": "string",
           "default": ""
         },
+        "justcil-printer": {
+          "title": "dbg.justcil-printer",
+          "description": "Printer to use for justcil",
```
michael-schwarz, line 1865: It may be good to have a longer description here.

michael-schwarz (approved):

PR #1372, opened 2024-02-27
Location fixes for YAML witness generation/validation

### LINKED ISSUE #1355: `for` loop invariant location in YAML witness
As pointed out by Freiburg, our YAML witnesses put `for` loop invariants _after_ the `for` keyword, but before the `for` loop initializer.
A more conventional location would be before the `for` keyword.

sim642:
Closes #1355.

### Changes
1. Adapt compilation to location fixes from https://github.com/goblint/cil/pull/167.
2. Deduplicate witness node identification logic to `WitnessUtil`. This prevents various copies going out of sync and makes them easier to test.
3. Extend the CFG cram test output with information about witness node identification. This makes it easy to see which CFG nodes can have which witness invariants.
4. Fix loop invariant locations. Loop heads for witnesses are not identified by back edges in CFG, but by checking for the `Loop` statement in the corresponding CIL AST. `Loop` statements are not part of our CFG nodes, so this lookup happens through `skippedByEdge`. This has two benefits:
   1. Loop invariants are no longer emitted for `goto`-loops. This matches YAML witness specification.
   2. Locations for loop heads now correctly point to the beginning of the structural loop keyword. This also works for `do`-`while` loops because the `Loop` node contains the location for the `do`.
5. Fix location for first initializer assignment. Previously initializer assignments in declarations used expression locations in the middle of the statement. Now the first assignment has its location before the entire declaration, which is a valid point before any initializers execute.
6. Make `precondition_loop_invariant` match loop nodes not arbitrary location nodes. This is our thing and not really used for anything, but at least the name matches the behavior now. This is from the time when `loop_invariant` was confused with location invariants in YAML witness specification.
7. ~~Mark YAML witnesses incompatible with loop unrolling for now.~~

### TODO
- [x] Merge https://github.com/goblint/cil/pull/167.
- [x] Fix GobView semantic search test.
- [x] Cannot have `location_invariant` right before loop because we have no non-loop-head node there. This meant our existing witness tests had to be changed to `loop_invariant`s which make more sense. But our inability to have location invariants before loops could also turn out to be problematic.

   ```c
   int i = 0;
   // we have no node here to attach i == 0 to
   while (i < 10)
     i++;
   ```
   Delegated to #1391.

- [x] Adapt to #1403.
- [x] Merge https://github.com/goblint/gobview/pull/44.


src/common/framework/myCFG.ml
```diff
@@ -42,19 +42,34 @@ sig
...
+module CfgEdge = struct
+  type t = Node.t * edges * Node.t [@@deriving eq, hash]
+end
+
+module CfgEdgeH = BatHashtbl.Make (CfgEdge)
+
+module type CfgBidirSkip =
+sig
+  include CfgBidir
+  val skippedByEdge: node -> edges -> node -> stmt list
```
michael-schwarz, line 56: We should probably comment what this is good for here.

michael-schwarz: And probably also make clear that this is in the forward direction, as the module here is a `CfgBidir`.

src/framework/control.ml
```diff
@@ -625,13 +625,16 @@ struct
...
+           iter (fun (node, v) ->
+              let loc = match node with
+                | Statement s -> Cil.get_stmtLoc s.skind (* nosemgrep: cilfacade *) (* Must use CIL's because syntactic search is in CIL. *)
+                | FunctionEntry _ | Function _ -> Node.location node
+              in
+              (* join values once for the same location and once for the same node *)
+              let join = Option.some % function None -> v | Some v' -> Spec.D.join v v' in
+              Hashtbl.modify_opt loc join by_loc;
+              NodeH.modify_opt node join by_node;
+            ) node_values;
```
michael-schwarz, line 637: Does this still behave correctly for the transformations? Could you elaborate on why this change was needed?

src/framework/control.ml
```diff
@@ -793,22 +799,22 @@ end
...
         to do a more precise analysis next time. *)
     (* TODO: do some more incremental refinement and reuse parts of solution *)
-    analyze_loop (module CFG) file fs change_info skippedByEdge
+    analyze_loop (module CFG) file fs change_info
 
 (** The main function to perform the selected analyses. *)
 let analyze change_info (file: file) fs =
   Logs.debug "Generating the control flow graph.";
-  let (module CFG), skippedByEdge = CfgTools.compute_cfg_skips file in
+  let (module CFG) = CfgTools.compute_cfg file in
```
michael-schwarz, line 818:
```suggestion
  let module CFG = CfgTools.compute_cfg file in
```

should work too, right?

src/maingoblint.ml
```diff
@@ -145,6 +145,8 @@ let check_arguments () =
   if get_bool "ana.autotune.enabled" && get_bool "incremental.load" then (set_bool "ana.autotune.enabled" false; warn "ana.autotune.enabled implicitly disabled by incremental.load");
   if get_bool "exp.basic-blocks" && not (get_bool "justcil") && List.mem "assert" @@ get_string_list "trans.activated" then (set_bool "exp.basic-blocks" false; warn "The option exp.basic-blocks implicitely disabled by activating the \"assert\" tranformation.");
   if (not @@ get_bool "witness.invariant.all-locals") && (not @@ get_bool "cil.addNestedScopeAttr") then (set_bool "cil.addNestedScopeAttr" true; warn "Disabling witness.invariant.all-locals implicitly enables cil.addNestedScopeAttr.");
+  if (get_int "exp.unrolling-factor" > 0 || AutoTune0.isActivated "loopUnrollHeuristic") && (get_bool "witness.yaml.enabled" || get_string "witness.yaml.validate" <> "" || get_string "witness.yaml.unassume" <> "") then
+    fail "YAML witnesses are incompatible with syntactic loop unrolling (https://github.com/goblint/analyzer/pull/1370).";
```
michael-schwarz, line 149: I think this is not a smart choice to make for now. At the very least we should extensively benchmark that impacts this has. I fear disabling loop unrolling will massively harm us in SV-COMP. This should be extensively benchmarked before changing it for our SV-COMP runs. (Ideally in a separate PR where this can be discussed extensively).

src/util/server.ml
```diff
@@ -121,6 +121,10 @@ let serve serv =
   |> Seq.map Packet.t_of_yojson
   |> Seq.iter (handle_packet serv)
 
+let is_server_node cfgnode =
```
michael-schwarz, line 124: What does this mean? When is a node a server node?

src/witness/witnessUtil.ml
```diff
@@ -79,7 +79,61 @@ struct
...
+
+  let is_stub_node n =
+    let fundec = Node.find_fundec n in
+    Cil.hasAttribute "goblint_stub" fundec.svar.vattr
+
+  let location_location (n : Node.t) = (* great naming... *)
+    match n with
+    | Statement s ->
+      let {loc; _}: CilLocation.locs = CilLocation.get_stmtLoc s in
+      if not loc.synthetic && is_invariant_node n && not (is_stub_node n) then (* TODO: remove is_invariant_node? *)
```
michael-schwarz, line 107: What's up with the TODO?

src/witness/witnessUtil.ml
```diff
@@ -79,7 +79,61 @@ struct
...
+    |> BatOption.filter (fun _loc -> not (is_stub_node n))
+
+  let is_loop_head_node n = Option.is_some (loop_location n)
+end
+
+module YamlInvariantValidate (FileCfg: MyCFG.FileCfg) =
+struct
+  include Invariant (FileCfg)
+
+  (* TODO: filter synthetic?
```
michael-schwarz, line 128: What's up with the TODO?

src/witness/witnessUtil.ml
```diff
@@ -79,7 +79,61 @@ struct
...
+
+     Almost all loops are transformed by CIL, so the loop constructs all get synthetic locations. Filtering them from the locator could give some odd behavior: if the location is right before the loop and all the synthetic loop head stuff is filtered, then the first non-synthetic node is already inside the loop, not outside where the location actually was.
+     Similarly, if synthetic locations are then filtered, witness.invariant.loop-head becomes essentially useless.
+     I guess at some point during testing and benchmarking I achieved better results with the filtering removed. *)
+
+  let is_loop_head_node = NH.mem loop_heads
 end
+[@@deprecated]
```
michael-schwarz, line 136: Where is this still used?

src/witness/yamlWitness.ml
```diff
@@ -287,7 +275,7 @@ struct
     let entries =
       if entry_type_enabled YamlWitnessType.LoopInvariant.entry_type then (
         LH.fold (fun loc ns acc ->
-            if WitnessInvariant.emit_loop_head && List.exists is_loop_head_node ns then (
+            if WitnessInvariant.emit_loop_head then ( (* TODO: remove double condition? *)
```
michael-schwarz, line 278: What's up with the TODO?

src/witness/yamlWitness.ml
```diff
@@ -340,11 +328,11 @@ struct
         entries
     in
 
-    (* Generate precondition invariants.
+    (* Generate precondition loop invariants.
```
michael-schwarz, line 331: @jerhard How does this change affect our planned usage for precondition invariants for the project with LMU? We need general precondition invariants there iirc?

tests/regression/cfg/util/dune
```diff
@@ -4,6 +4,7 @@
    goblint-cil
    goblint_logs
    goblint_common
+   goblint_lib ; TODO: avoid
```
michael-schwarz, line 7: What's up with the TODO?

src/witness/yamlWitness.ml
sim642: Having `precondition_location_invariant` entries is just a matter of defining the corresponding types and doing what we already have for `location_invariant` vs `loop_invariant`: use a different predicate for filtering nodes.

tests/regression/cfg/util/dune
sim642: Previously the `cfgDot` testing utility didn't depend on the entirety of goblint.lib because it's really just concerned with CFGs and thus less unnecessary work by `dune runtest`.

This is now needed because the witness node predicates are in goblint.lib. I'll add a comment here.

src/analyses/unassumeAnalysis.ml
```diff
@@ -41,26 +41,26 @@ struct
...
-    let module WitnessInvariant = WitnessUtil.Invariant (FileCfg) in
+    let module WitnessInvariant = WitnessUtil.YamlInvariant (FileCfg) in
 
     (* DFS, copied from CfgTools.find_backwards_reachable *)
     let reachable = NH.create 100 in
     let rec iter_node node =
       if not (NH.mem reachable node) then begin
         NH.replace reachable node ();
-        (* TODO: filter synthetic?
-           See YamlWitness. *)
```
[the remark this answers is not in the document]
sim642: I think previously the unassume analysis and YAML witness generation/validation had slightly different notions of acceptable nodes/locations. Now that this is extracted to `WitnessUtil` to be reused for both, there isn't such discrepancy any more.

src/framework/control.ml
sim642: Nope, this is a pattern for unpacking a first-class module (where the right-hand side is an expression): https://v2.ocaml.org/manual/firstclassmodules.html. `let module` is for local module definitions (where the right-hand side is a module): https://v2.ocaml.org/manual/expr.html#sss:expr-local-modules. The parenthesis are exactly what disambiguate the grammar.

src/framework/control.ml
sim642: 02bba9350b61590653b87a89e79b68c7fd905ba1 was needed to fix the GobView CI for syntactic/semantic search.

Really, it always needed to be that: since the syntactic search and its location matching are implemented in CIL, then it uses `Cil.get_stmtLoc` as its locations. Goblint's `Node.location` prefers expression locations instead, so the two don't necessarily match.
I don't really know why the communication between CIL and Goblint for syntactic search uses locations, not `sid`-s or something.

src/witness/witnessUtil.ml
sim642: It's not anymore, but I don't want to immediately remove it because there's a long comment explaining some old behavior with possible practical considerations. Now validation uses the same predicates as generation, which seems the reasonable thing to do, but without tests for YAML witness validation yet, I'm not completely sure whether this should be removed or there's some truth to this and the distinction needs to be reintroduced.

This PR has been driven by making Goblint's generated witnesses easier to consume by Ultimate, but does not consider the effect on Goblint Validator validating witnesses from all the other tools. There could be regressions in that regard so I don't want to completely forget this point.

src/witness/yamlWitness.ml
sim642: Stemming from the `loop_invariant`/`location_invariant` confusion times, we have an option `witness.invariant.loop-head` in addition to `witness.yaml.entry-types`. Currently both need to be enabled to get these invariants, but maybe the former option has been unnecessary for a while. This is something to be considered separately.

src/witness/witnessUtil.ml
sim642: This is another side of https://github.com/goblint/analyzer/pull/1372#discussion_r1528301516. I'll expand the comment on it.

src/maingoblint.ml
sim642: We already discussed this at GobCon and decided to go this route to be able to make any progress with Freiburg. SV-COMP 2025 is further away, so #1370 is less urgent at this point.

It is impossible to separate these location fixes from this incompatibility: the fixed loop identification is based on `Loop` statements, which the syntactic unrolling doesn't have for the unrolled iterations, otherwise our witnesses are simply unsound (#1225).

For our own sv-benchmarks runs with only verification, no validation, we could flip the choice, but that is inherently unrepresentative of actual SV-COMP with validation. If we output a trivially falsifiable loop invariant `i == 0 && i == 1 && ...` for a loop that doesn't even need unrolling (for which `0 <= i` would be valid), then we've just shot ourselves in the foot.

There's no way around #1370. The question is simply whether we block this PR and the Freiburg collaboration behind #1370 or not (and do it later by the time for SV-COMP 2025).

sim642:
> It might make sense to pull out the loop unrolling and sv-comp default config discussion and discuss that separately?

It's now extracted to #1402.

conf/svcomp.json
```diff
@@ -84,7 +84,6 @@
         "congruence",
         "octagon",
         "wideningThresholds",
-        "loopUnrollHeuristic",
```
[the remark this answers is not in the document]
michael-schwarz: I thought these were to be pulled out to #1402 so we can merge this and then handle the loop unrolling separately?

src/maingoblint.ml
sim642: This is now reverted after #1403. That seems to work for now.

PR #1377, opened 2024-03-01
Fix `evalbinop_base` for `==` on `Int` and `Address` 

### LINKED ISSUE #1376: Evaluation of `BinOp` unsound for arguments of type `Int` and `Address`
For this program

~~~C
#include <goblint.h>
int main ()
{
  int x;
  int three = 3;

  int y = (three == &x);

  __goblint_check(y == 1); //UNKNOWN!

  return 0;
}
~~~

we claim that `y` must have value `1`, which amounts  to the address of `x` equaling `3` which is exceedingly unlikely.

The offending code is this:

https://github.com/goblint/analyzer/blob/9c2c167bfb2d683a6b0e170ce52e6f40cdb3d1eb/src/analyses/base.ml#L322-L326

It gets called for `Int 3` for which `ID.to_bool` yields `Some true`

https://github.com/goblint/analyzer/blob/9c2c167bfb2d683a6b0e170ce52e6f40cdb3d1eb/src/cdomain/value/cdomains/intDomain.ml#L2070-L2074

and `Address {&x}` for which `AD.to_bool` also yields `Some true`

https://github.com/goblint/analyzer/blob/9c2c167bfb2d683a6b0e170ce52e6f40cdb3d1eb/src/cdomain/value/cdomains/addressDomain.ml#L222

Thus the returned value is `1` which is incorrect.

This bug seems to date back almost 10 years to commit  dc5452b (which was the last time that handling was touched, maybe even further).

Commit 25407b52a9cca1a98452957a9be561b4f6aff83 on #1297 fixes this issue. I will still make a separate PR from `master` such that this soundness fix is not intermingled with the rest of this huge PR.

The bug report is due to @DrMichaelPetter digging into the internals of Goblint for a week, he was able to uncover this bug in #1297.
I only contributed the small example to reproduce the issue that does not need apron or partitioned arrays.

michael-schwarz:
Closes #1376

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -321,8 +321,23 @@ struct
     | Address p, Int n
     | Int n, Address p when op=Eq || op=Ne ->
       let ik = Cilfacade.get_ikind t in
-      Int (match ID.to_bool n, AD.to_bool p with
-          | Some a, Some b -> ID.of_bool ik (op=Eq && a=b || op=Ne && a<>b)
+      Int (match ID.to_int n, ID.to_int (AD.to_int p)  with
+          | Some a, Some b ->
+            (* both are integer constants *)
+            ID.of_bool ik (op=Eq && Z.equal a b || op=Ne && not @@ Z.equal a b)
+          | Some a, _ ->
+            (* The integer has a definite value, we can learn from this if we know about nullness of the pointer *)
+            if AD.is_null p then
+              (* address is null-pointer we can definitely answer the query *)
+              if op = Eq then
+                ID.of_bool ik (Z.equal a Z.zero)
+              else
+                ID.of_bool ik (not @@ Z.equal a Z.zero)
+            else if AD.is_not_null p && Z.equal a Z.zero then
+              (* address is definitely not null, we can answer if Z is zero *)
+              ID.of_bool ik (op <> Eq)
+            else
+              bool_top ik
           | _ -> bool_top ik)
     | Address p, Int n  ->
       addToAddrOp p n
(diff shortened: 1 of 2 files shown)


src/analyses/base.ml
```diff
@@ -321,9 +321,26 @@ struct
...
+          | Some false ->
+            (* n is zero, p is NULL*)
+            ID.of_bool ik (op = Eq)
+          | _ -> bool_top ik
+        else if AD.is_not_null p then
+          match (ID.to_int n) with
+          | Some i when Z.equal i Z.zero ->
+            (* n is zero, p is not NULL *)
+            ID.of_bool ik (op = Ne)
+          | _ -> bool_top ik
```
sim642, line 339: Why does the `is_not_null` case use `ID.to_int` and `Z.equal` to check the same thing that `is_null` case does with `ID.to_bool`?

Also, `is_null` and `is_not_null` just give the two cases of `AD.to_bool` but with duplicate checking, no?
Seems to me that the original match over both `ID.to_bool` and `AD.to_bool` would still be a more concise way to do this double matching. The problem is just that it should exclude the case where both are non-null, where we know nothing about their equality.

michael-schwarz: > Why does the is_not_null case use ID.to_int and Z.equal to check the same thing that is_null case does with ID.to_bool?

I think like this the intent is more clearly visible.


> the original match over both `ID.to_boo`l and `AD.to_bool` would still be more concise

More concise probably yes but I think the current version is much easier to read and understand and does not require reasoning about what `AD.to_bool` does. The fact that this bug was hidden for so long I think is a good argument to be as literal here as possible.

sim642: Matching `ID.equal_to 0` would then make the intent the most clear.

sim642 (approved):

PR #1387, opened 2024-03-11
Fix test for relational unassume with strengthening

### LINKED ISSUE #1373: Relational unassume with strengthening is broken
On `master` (at fd0cb184315e3ca7b0ba8c9dcba2d3a331447503), 56-witness/25-apron-unassume-strengthening no longer passes.

sim642:
Closes #1373.

The previously-working behavior without type bounds was actually somewhat odd (as explained by the added comments), so it doesn't need to work anymore.
The actual update is just to the example which uses `<=` instead of `<` in the paper. That avoids the oddity.

This also includes a small `Lincons1` fix in strengthening to make the behavior consistent between octagons and polyhedra.

### DIFF
--- a/src/cdomains/apron/apronDomain.apron.ml
+++ b/src/cdomains/apron/apronDomain.apron.ml
@@ -639,13 +639,20 @@ struct
       (* Whether [con1] contains a var in [env]. *)
       let env_exists_mem_con1 env con1 =
         try
-          Lincons1.iter (fun _ var ->
-              if Environment.mem_var env var then
+          Lincons1.iter (fun coeff var ->
+              (* Lincons1 from polyhedra may contain variable with zero coefficient.
+                 These are silently not printed! *)
+              if not (Coeff.is_zero coeff) && Environment.mem_var env var then
                 raise Not_found
             ) con1;
           false
         with Not_found ->
           true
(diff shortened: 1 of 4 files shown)


src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -639,14 +639,21 @@ struct
...
-              if Environment.mem_var env var then
+          Lincons1.iter (fun coeff var ->
+              (* Lincons1 from polyhedra may contain variable with zero coefficient.
+                 These are silently not printed! *)
+              if not (Coeff.is_zero coeff) && Environment.mem_var env var then
                 raise Not_found
             ) con1;
           false
         with Not_found ->
           true
```
michael-schwarz, line 650: I was confused by `Not_found` here. Should we use a custom exception here such as `Found` to make it easier to understand what's happening?

michael-schwarz (approved):

src/cdomains/apron/apronDomain.apron.ml
sim642: You're right, it slightly confused me too. I'll use `Stdlib.Exit` here as we already do in other places.

PR #1395, opened 2024-03-20
Use ppx_deriving_hash 0.1.2

sim642:
This gets rid of some ppx_deriving_hash related TODOs by using a version of it which includes their support.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -30,7 +30,7 @@
     (yojson (>= 2.0.0))
     (qcheck-core (>= 0.19))
     ppx_deriving
-    ppx_deriving_hash
+    (ppx_deriving_hash (>= 0.1.2))
     (ppx_deriving_yojson (>= 3.7.0))
     (ounit2 :with-test)
     (qcheck-ounit :with-test)
--- a/goblint.opam
+++ b/goblint.opam
@@ -28,7 +28,7 @@ depends: [
   "yojson" {>= "2.0.0"}
   "qcheck-core" {>= "0.19"}
   "ppx_deriving"
-  "ppx_deriving_hash"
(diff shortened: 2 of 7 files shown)


michael-schwarz:
LGTM!

---

 Side remark: Have you discussed with @stilscher the issues that `ppx_deriving_hash` breaking the invariants ensured by the `Stdlib.hash` (hashes being portable between 32 and 64 bit machines) causes for gobview? The search being broken seems to, e.g., be a consequence of this.

michael-schwarz (approved):

sim642:
> Side remark: Have you discussed with @stilscher the issues that `ppx_deriving_hash` breaking the invariants ensured by the `Stdlib.hash` (hashes being portable between 32 and 64 bit machines) causes for gobview? The search being broken seems to, e.g., be a consequence of this.

Briefly, yes. ppx_deriving_hash is hardly the culprit here: Goblint still contains many manual `hash` implementations that also don't guarantee this. Also `hash`es from dependencies might not, so it's hard to guarantee that.

I vaguely remember this point coming up before but I don't know where and why. Our `Printable.relift` exists precisely to fix this by rehashing marshaled data. Maybe the issue at the time was just fixed by properly recursing `relift` into all the necessary places.

What search is broken right now? During some recent changes I actually tested GobView's syntactic/semantic search and it seemed to work. GobView CI also tests this.

michael-schwarz:
> GobView's semantic search

Last I heard from Sarah that was still broken.

sim642:
Odd, because in #1372 I broke the GobView CI and also fixed it, so it seems to still work there.

PR #1396, opened 2024-03-21
Avoid base invariant fallback not understood message on reflexive pointer literal equality

### LINKED ISSUE #1374: Reflexive pointer literal equality not understood
Running our simplest regression test now produces some odd
```
expression "& mutex1 == & mutex1" not understood
```
output:
```console
$ ./regtest.sh 04 01
./goblint --enable warn.debug --enable dbg.regression --html   tests/regression/04-mutex/01-simple_rc.c
[Debug][Analyzer] Invariant failed: expression "& mutex1 == & mutex1" not understood. (tests/regression/04-mutex/01-simple_rc.c:9:3-9:30)
[Debug][Analyzer] Invariant failed: expression "& mutex2 == & mutex2" not understood. (tests/regression/04-mutex/01-simple_rc.c:18:3-18:30)
  live: 12
  dead: 0
  total lines: 12
[Warning][Race] Memory location myglobal (race with conf. 110): (tests/regression/04-mutex/01-simple_rc.c:4:5-4:13)
  write with [lock:{mutex1}, thread:[main, t_fun@tests/regression/04-mutex/01-simple_rc.c:17:3-17:40]] (conf. 110)  (exp: & myglobal) (tests/regression/04-mutex/01-simple_rc.c:10:3-10:22)
  write with [mhp:{created={[main, t_fun@tests/regression/04-mutex/01-simple_rc.c:17:3-17:40]}}, lock:{mutex2}, thread:[main]] (conf. 110)  (exp: & myglobal) (tests/regression/04-mutex/01-simple_rc.c:19:3-19:22)
  read with [lock:{mutex1}, thread:[main, t_fun@tests/regression/04-mutex/01-simple_rc.c:17:3-17:40]] (conf. 110)  (exp: & myglobal) (tests/regression/04-mutex/01-simple_rc.c:10:3-10:22)
  read with [mhp:{created={[main, t_fun@tests/regression/04-mutex/01-simple_rc.c:17:3-17:40]}}, lock:{mutex2}, thread:[main]] (conf. 110)  (exp: & myglobal) (tests/regression/04-mutex/01-simple_rc.c:19:3-19:22)
[Info][Race] Memory locations race summary:
  safe: 0
  vulnerable: 0
  unsafe: 1
  total memory locations: 1
[Info] Writing xml to temp. file: /tmp/ocaml4ffc79tmp
Time needed: 296 ms
See result/index.xml
```

These are probably the result of #1343 which emits such refinements. The fact that they are emitted is not the problem, rather that base analysis claims to not understand trivial refinements which cannot refine anything anyway.

sim642:
Closes #1374.

The fix is uglier than I imagined it to be because address set refinement happens in `invariant_fallback`, not in the general `inv_exp`, so it doesn't simply fall into the existing
https://github.com/goblint/analyzer/blob/da91765f030b3583606e50ad21b0d0b336c52fc7/src/analyses/baseInvariant.ml#L775

A less ugly fix would be to make `inv_exp` also do addresses with HC4-revise, but that requires carefully porting all the ancient fallback logic with countless edge cases.

### DIFF
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -237,7 +237,7 @@ struct
     | `Refine (lval, value) ->
       refine_lv_fallback ctx st lval value tv
     | `NothingToRefine ->
-      if M.tracing then M.traceu "invariant" "Doing to refine.\n";
+      if M.tracing then M.traceu "invariant" "Nothing to refine.\n";
       st
     | `NotUnderstood ->
       if M.tracing then M.traceu "invariant" "Doing nothing.\n";
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -115,7 +115,7 @@ struct
   let invariant_fallback ctx st exp tv =
     (* We use a recursive helper function so that x != 0 is false can be handled
      * as x == 0 is true etc *)
-    let rec helper (op: binop) (lval: lval) (value: VD.t) (tv: bool): (lval * VD.t) option =
+    let rec helper (op: binop) (lval: lval) (value: VD.t) (tv: bool): [`Refine of lval * VD.t | `NotUnderstood] =
       match (op, lval, value, tv) with
       (* The true-branch where x == value: *)
       | Eq, x, value, true ->
@@ -123,8 +123,8 @@ struct
         (match value with
          | Int n ->
            let ikind = Cilfacade.get_ikind_exp (Lval lval) in
-           Some (x, Int (ID.cast_to ikind n))
-         | _ -> Some(x, value))
+           `Refine (x, Int (ID.cast_to ikind n))
+         | _ -> `Refine (x, value))
       (* The false-branch for x == value: *)
       | Eq, x, value, false -> begin
           match value with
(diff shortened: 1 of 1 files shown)


src/analyses/baseInvariant.ml
```diff
@@ -228,12 +231,15 @@ struct
...
-        None
+        `NotUnderstood
     in
     match derived_invariant exp tv with
-    | Some (lval, value) ->
+    | `Refine (lval, value) ->
       refine_lv_fallback ctx st lval value tv
-    | None ->
+    | `NothingToRefine ->
+      if M.tracing then M.traceu "invariant" "Doing to refine.\n";
```
michael-schwarz, line 240:
```suggestion
      if M.tracing then M.traceu "invariant" "Nothing to refine.\n";
```

michael-schwarz:
Maybe the fix should also attempt to forgo emitting these: Tasking the `event` system with what we know to be a fool's errand, might have a performance impact. While it will most likely be very small, it's probably still worth doing as it'll be very little extra work to add this check.

Also, I am not entirely convinced by the usage of polymorphic variants here yet. I guess the only difference to normal variants here is that `NothingToRefine` cannot be emitted by `helper`. Is this really an important invariant that we want to enforce (and are worth paying the penalty for less efficient code for it, provided this is still the case after `flambda`).

sim642:
> Also, I am not entirely convinced by the usage of polymorphic variants here yet. I guess the only difference to normal variants here is that `NothingToRefine` cannot be emitted by `helper`. Is this really an important invariant that we want to enforce (and are worth paying the penalty for less efficient code for it, provided this is still the case after `flambda`)?

The use of polymorphic variants doesn't actually change the memory representations here: argument-less polymorphic variants are unboxed integers like normal variants and a polymorphic variant with two arguments is really a block with a reference to a pair block, but that was also the case before with a pair `option`.

michael-schwarz:
Besides the point but still interesting:

> The use of polymorphic variants doesn't actually change the memory representations here: [...] a polymorphic variant with two arguments is really a block with a reference to a pair block, but that was also the case before with a pair `option`.

Interesting, do you have a reference for this?  RealWorldOCaml seems to disagree:
> Polymorphic variants use more memory space than normal variants when parameters are included in the data type constructors. Normal variants use the tag byte to encode the variant value and save the fields for the contents, but this single byte is insufficient to encode the hashed value for polymorphic variants. They must allocate a new block (with tag 0) and store the value in there instead. Polymorphic variants with constructors thus use one word of memory more than normal variant constructors.


https://dev.realworldocaml.org/runtime-memory-layout.html

I would read this as: 
- `(x * y ) option` 
      - 1 block for the pointer to the tuple, and `Some` vs `None` expressed in the tag
      - 1 block for the tuple itself
      - + memory for x and y

sim642:
> Polymorphic variants with constructors thus use one word of memory more than normal variant constructors.

Normal variants flatten their arguments instead of storing a tuple, so what this is saying is that `` `Refine of lval * value`` uses one block more than `Refine of lval * value` would. But also `Refine of (lval * value)` would use one block more for the pair which it explicitly contains. And that by abstraction is the same as `(lval * value) option`.

michael-schwarz (approved):
I added the code that prevents emitting refines of form `&a==&a` in the first place and tested it by checking that the error messages disappear on master (which doesn't have the rest of this PR) with this change.

michael-schwarz:
The atomic privatizations added in the course of https://github.com/goblint/analyzer/pull/1216 seem to depend on the refine event being emitted for `&a == &a`. That seems very strange.

sim642:
I extracted the optimization to #1407 to get approved thing merged.

PR #1397, opened 2024-03-23
Introduce `Analyses.IdentityUnitContextsSpec` & Replace `Lattice.Unit` contexts with `Printable.Unit`

### LINKED ISSUE #1378: Replace `Lattice.Unit` contexts in analyses with `Printable.Unit`
`Lattice.Unit` contexts come from the days where contexts had to be a `Lattice.t`. This requirement is now gone.
While it'll have no performance impact, it might be nice to have this fact reflected in the analyses to not confuse newcomers.

michael-schwarz:
Closes #1378.

### DIFF
--- a/src/analyses/abortUnless.ml
+++ b/src/analyses/abortUnless.ml
@@ -11,7 +11,7 @@ struct
 
   let name () = "abortUnless"
   module D = BoolDomain.MustBool
-  module C = Lattice.Unit
+  module C = Printable.Unit
 
   let context _ _ = ()
 
--- a/src/analyses/mutexTypeAnalysis.ml
+++ b/src/analyses/mutexTypeAnalysis.ml
@@ -6,13 +6,13 @@ open Analyses
 module MAttr = ValueDomain.MutexAttr
 module LF = LibraryFunctions
 
-module Spec : Analyses.MCPSpec with module D = Lattice.Unit and module C = Lattice.Unit =
+module Spec : Analyses.MCPSpec with module D = Lattice.Unit and module C = Printable.Unit =
 struct
   include Analyses.IdentitySpec
 
   let name () = "pthreadMutexType"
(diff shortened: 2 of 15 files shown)


src/analyses/abortUnless.ml
```diff
@@ -11,7 +11,7 @@ struct
 
   let name () = "abortUnless"
   module D = BoolDomain.MustBool
-  module C = Lattice.Unit
+  module C = Printable.Unit
 
   let context _ _ = ()
```
sim642, line 16: Couldn't we have `include Analyses.UnitContextsSpec` to include just these two definitions without also being forced to include `IdentitySpec`?

src/analyses/mutexTypeAnalysis.ml
```diff
@@ -6,13 +6,12 @@ open Analyses
...
 module LF = LibraryFunctions
 
-module Spec : Analyses.MCPSpec with module D = Lattice.Unit and module C = Lattice.Unit =
+module Spec : Analyses.MCPSpec with module D = Lattice.Unit and module C = Printable.Unit =
 struct
-  include Analyses.IdentitySpec
+  include Analyses.IdentityUnitContextsSpec
 
   let name () = "pthreadMutexType"
   module D = Lattice.Unit
```
sim642, line 14: Couldn't this even `include UnitAnalysis.Spec`?

src/analyses/abortUnless.ml
michael-schwarz: Yes, would be possible of course. The downside is that it would make many analyses a bit more verbose.

michael-schwarz: I would be happy to make this change, in case we think the places where we can then use it outweighs the other places where it would make use more verbose.

sim642 (approved):

PR #1399, opened 2024-03-28
Ego-Lane-Derived-Digests for Privatizations: `Lock- ` & `Write-Centered` Privatizations

michael-schwarz:
Add `tid` variants of the remaining privatizations to be evaluated for my thesis.

This attempts to make the construction from #1398 a bit more general, by replacing unknowns `[α]` with unknowns `[α,A]` for `A` from some ego-lane derived digest.
Implementation-wise this is the same map construction as outlined in the PR for `Protection-TID`.

TODO:

- [x] Write-Centered
- [x] Combined

Follow-up to #1398.


src/config/options.schema.json
```diff
@@ -752,9 +752,9 @@
             "privatization": {
               "title": "ana.base.privatization",
               "description":
-                "Which privatization to use? none/mutex-oplus/mutex-meet/mutex-meet-tid/protection/protection-read/mine/mine-nothread/mine-W/mine-W-noinit/lock/write/write+lock",
+                "Which privatization to use? none/mutex-oplus/mutex-meet/mutex-meet-tid/protection/protection-read/mine/mine-nothread/mine-W/mine-W-noinit/lock/lock-tid/write/write-tid/write+lock",
```
sim642, line 755:
```suggestion
                "Which privatization to use? none/mutex-oplus/mutex-meet/mutex-meet-tid/protection/protection-read/mine/mine-nothread/mine-W/mine-W-noinit/lock/lock-tid/write/write-tid/write+lock/write+lock-tid",
```

Although maybe we should just not list the possible values in the description. There's an `enum` right below.

sim642 (approved):

PR #1402, opened 2024-04-02
Disable SV-COMP syntactic loop unrolling

sim642:
This is cherry-picked from #1372 in preparation for it to get merged: https://github.com/goblint/analyzer/pull/1372#discussion_r1526572811.

I benchmarked the equivalent change at be3ee1d80: https://goblint.cs.ut.ee/results/158-rm-loop-unroll-after/table-generator-cmp.diff.html. This costs us ~100 tasks, mostly from nla-digbench-scaling/*_unwindbound and product-lines/elevator_spec*.
The actual delta is that we lose ~130 true verdicts, but also gain ~30 true verdicts (from cases where unrolling causes excessive resource usage).
The nightly BenchExec run will give a more up-to-date comparison, but I don't expect it to be different since nothing notable has been changed in SV-COMP analyses in this time.

To get this back, we'll have to finalize #1370 for SV-COMP 2025.

### DIFF
--- a/conf/svcomp-validate.json
+++ b/conf/svcomp-validate.json
@@ -1,0 +1,139 @@
+{
+  "ana": {
+    "sv-comp": {
+      "enabled": true,
+      "functions": true
+    },
+    "int": {
+      "def_exc": true,
+      "enums": false,
+      "interval": true
+    },
+    "float": {
+      "interval": true
+    },
+    "activated": [
+      "base",
+      "threadid",
+      "threadflag",
+      "threadreturn",
+      "mallocWrapper",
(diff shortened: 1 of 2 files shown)


michael-schwarz:
> To get this back, we'll have to finalize https://github.com/goblint/analyzer/pull/1370 for SV-COMP 2025.

This is under the assumption that the same behavior can be achieved there, which is not a given, e.g., vis-a-vis automatic detection of widening points or the termination analysis.

sim642:
Closing since #1403 made it redundant for #1372.

Issue #1296, opened 2023-12-15
Comparison of signed integer with unsigned integer and a constant: Invariant failed: expression not understood. 


FungOliver:
When comparing a signed integer with an unsigned integer and a constant in a loop. Goblint fails to parse those expressions:  

In the following example:
```c
//PARAM:  --enable ana.int.interval   --set ana.activated[+] apron   
#include <stdio.h>
#include <stdlib.h>
#include <time.h>


int main() {
   unsigned int length = 5;
   for ( int i = 0; i < length; i++) {
     if ( i < length + 3){
        assert( i < length+1);
     }
   }
   return 0;
}
```
I get the following error:
```
./goblint --enable warn.debug --enable dbg.regression --html  --enable ana.int.interval   --set ana.activated[+] apron    --html tests/regression/98-relational-array-oob/01-random-simple.c
[Warning][Integer > Overflow][CWE-191] Unsigned integer underflow (tests/regression/98-relational-array-oob/01-random-simple.c:9:8-9:36)
[Warning][Integer > Overflow][CWE-191] Unsigned integer underflow (tests/regression/98-relational-array-oob/01-random-simple.c:10:10-10:24)
[Debug][Analyzer] **Invariant failed: expression (unsigned int )i < length + 3U" not understood. (tests/regression/98-relational-array-oob/01-random-simple.c:10:10-10:24)
[Warning][Integer > Overflow][CWE-191] Unsigned integer underflow (tests/regression/98-relational-array-oob/01-random-simple.c:11:9-11:30)
[Debug][Analyzer] Invariant failed: expression "(unsigned int )i < length + 1U" not understood. (tests/regression/98-relational-array-oob/01-random-simple.c:11:9-11:30)
[Warning][Assert] Assertion " i < length+1" is unknown. Expected: SUCCESS -> failed (tests/regression/98-relational-array-oob/01-random-simple.c:11:9-11:30)
[Warning][Integer > Overflow][CWE-190] Signed integer overflow (tests/regression/98-relational-array-oob/01-random-simple.c:9:14-9:19)
[Info][Deadcode] Logical lines of code (LLoC) summary:
  live: 6
  dead: 0
  total lines: 6
Writing xml to temp. file: /tmp/ocaml3807a8tmp
Time needed: 505 ms
See result/index.xml
```

michael-schwarz:
~~~C

michael-schwarz:
The original example from this issue works without any hitch now.

This example though still 
~~~C
[...]

PR #1408, opened 2024-04-04
Improve `invariant` for unsigned types

michael-schwarz:
Currently, we check in invariant that the result of casting to the new `ik` is `ID.leq` the uncast value.

As we do not do a reduced product, this causes some issues when one of the domain is imprecise:
E.g., because of DefExc, component-wise `leq` might not hold, even if it still holds for interval domain.

```
 CastE: Lval(Var(i, NoOffset)) evaluates to (Unknown int([-31,31]),[0,2147483647]) 
 which is bigger than the type it is cast to which is unsigned int 
```

This replaced the checks with checks on intervals. This leads to the assertion in the example now passing and no overflow messages being emitted any more

References #1296


src/analyses/baseInvariant.ml
```diff
@@ -786,8 +786,14 @@ struct
...
-               match unrollType (Cilfacade.typeOf e) with
+             let cinterval = ID.cast_to ik i in
+             let cl, cu = ID.minimal cinterval, ID.maximal cinterval in
+             let l, u = ID.minimal i, ID.maximal i in
+             (match cl,cu,l,u with
+              | Some cl, Some cu, Some l, Some u when Z.leq cl l && Z.leq u cu ->
+                (* Checking that the maximal and minimal values still fit instead of checking leq which may fail if one domain is *)
+                (* imprecise, e.g., DefExc *)
+                (match unrollType (Cilfacade.typeOf e) with
                | TInt(ik_e, _)
```
sim642, line 797: Not sure why the indentation CI job gets skipped, but it looks screwed up now.

michael-schwarz: I'll have a look. Actually, I am trying to pull this out and replace `VD.is_safe_cast: typ -> typ-> bool` with 

- `VD.is_staticlaly_safe_cast` (same as before)
- `VD.is_dynamically_safe_cast: typ -> typ -> VD.t -> bool` taking the current values into account.

I think there's probably more places where this would be the right thing to do, e.g., here:

https://github.com/goblint/analyzer/blob/092eed37b8803f0f1886d0cd83ada4f204c150a4/src/analyses/baseInvariant.ml#L210-L240

michael-schwarz: The comment there seems suspicious, as `ID.cast_to_ik v = v` does not imply `VD.is_safe_cast t1 (Cilfacade.typeOfLval x)` and the reverse is also not true.

src/cdomain/value/cdomains/valueDomain.ml
```diff
@@ -353,6 +354,22 @@ struct
...
+    if is_statically_safe_cast t2 t1 then
+      true
+    else
+      match t2, t1, v with
+      | (TInt (ik2,_) | TEnum ({ekind=ik2; _},_)) , (TInt (ik1,_) | TEnum ({ekind=ik1; _},_)), Int v ->
+        let cinterval = ID.cast_to ik2 v in
+        let cl, cu = ID.minimal cinterval, ID.maximal cinterval in
+        let l, u = ID.minimal v, ID.maximal v in
+        (match cl,cu,l,u with
+         | Some cl, Some cu, Some l, Some u when Z.leq cl l && Z.leq u cu -> true
```
sim642, line 368: Doesn't `ID.cast_to` sometimes go to top when the target type doesn't fit the current values?
Because then the condition would trivially hold even though all values didn't fit.

sim642: > I think this should be safe, and it would even be safe if we chose the bounds of `ik2` for `cl` and `cu` as far as I can see.

But then I don't see the need to check against the casted range of `v` as opposed to `ik2` bounds. Checking whether `(l,u)` is included in the casted range (if it's not `ik2` bounds), then this check is just unnecessarily stricter, if the bounds are already sound.

For example, interval domain overflow check does just that. So we could skip the redundant `ID.cast_to` computation here.

michael-schwarz: That is of course true! I didn't know that `Size` was accessible from outside the `IntDomain` module.

PR #1411, opened 2024-04-05
Add flag `AnalysisState.executing_speculative_computations` to avoid flagging overflows during `invariant`

michael-schwarz:
As discussed in #1296


michael-schwarz:
The question is whether we should undo #1406 in light of this more general solution.

src/analyses/baseInvariant.ml
```diff
@@ -830,4 +830,12 @@ struct
             FD.top_of fk
         in
         inv_exp (Float ftv) exp st
+
+  let invariant ctx st exp tv =
+    (* The computations that happen here are not computations that happen in the programs *)
+    (* Any overflow during the forward evaluation will already have been flagged here *)
+    AnalysisState.executing_speculative_computations := true;
+    let res = invariant ctx st exp tv in
+    AnalysisState.executing_speculative_computations := false;
```
sim642, line 839: This should use `GobRef.wrap` at least. `invariant` can raise `Deadcode`.

sim642:
The dune promotion stuff may have been due to trailing whitespace removal which should be disabled for `.t` files.

src/common/framework/analysisState.ml
```diff
@@ -4,6 +4,12 @@
     This is set to true in control.ml before we verify the result (or already before solving if warn = 'early') *)
 let should_warn = ref false
 
+(** If this is true, any overflows happening in IntDomains will not lead to warnings being produced or
+    {!svcomp_may_overflow} being set to true. This is useful when, e.g., {!BaseInvariant.refine} executes computations that
```
sim642, line 8:
```suggestion
    {!svcomp_may_overflow} being set to true. This is useful when, e.g., {!BaseInvariant.Make.invariant} executes computations that
```

sim642 (approved):

PR #1412, opened 2024-04-12
Lin2var relations fine-tuning - switching from array to map

DrMichaelPetter:
This swaps out the basic datastructure in lin2var analysis from an array to a hashtable. Since we expect very sparse information kept in the relational part anyway, we expect a sizeable amount of runtime improvements vs. the conventional array implementation.


src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,33 +16,75 @@ open VectorMatrix
...
-  type t = Equality.t Array.t [@@deriving eq, ord]
+module EqualitiesConjunction = struct
+  module IntHash = struct
+    type t = int
+    let equal = (=)
+    let hash i = i land max_int
+  end
+  module IntHashtbl = BatHashtbl.Make(IntHash)
+
+  let equal i j = (fst i) = (fst j) && IntHashtbl.fold (fun lh rh acc -> acc && IntHashtbl.mem (snd j) lh && IntHashtbl.find (snd j) lh = rh) (snd i) true
```
michael-schwarz, line 37: Would it not be nicer to directly unwrap the tuple in the argument?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,33 +16,75 @@ open VectorMatrix
...
+  let hash : t -> int = fun x -> IntHashtbl.fold (fun k value acc -> 31 * 31 * acc + 31 * k + Rhs.hash value) (snd x) 0 (* TODO: derive *)
+
+  let empty () = (0, IntHashtbl.create 0)
 
-  let show m =
-    Array.fold_right (fun k result -> Equality.show k ^ "\n" ^ result) m ""
+  let make_empty_conj len = (len, IntHashtbl.create len)
 
-  let hash : t -> int = Array.fold_left (fun acc a -> 31 * acc + Equality.hash a) 0 (* TODO: derive *)
+  let get_dim econ = fst econ
```
michael-schwarz, line 65: I think directly unwrapping the argument here is also more intuitive,

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,33 +16,75 @@ open VectorMatrix
...
 
-  let empty () = [||]
+  let nontrivial econ lhs = IntHashtbl.mem (snd econ) lhs
 
-  let make_empty_array len = Array.init len (fun i -> (Some i, Z.zero))
+  (** sparse implementation of get rhs for lhs, but will default to no mapping for sparse entries *)
+  let get_rhs econ lhs = IntHashtbl.find_default (snd econ) lhs (Rhs.var_zero lhs)
+
+  (** inplace update on the EqualitiesConjunction, repling the rhs for lhs with a new one if meaningful, otherwise just drop the entry*)
```
michael-schwarz, line 72:
```suggestion
  (** inplace update on the EqualitiesConjunction, replacing the rhs for lhs with a new one if meaningful, otherwise just drop the entry*)
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
michael-schwarz: There's a couple more places like this, but in the end what one prefers probably comes down to taste.

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -391,10 +435,16 @@ struct
 
   let join a b =
     let join_d ad bd =
-      (*use copy of ad because result is later saved in there*)
-      let ad = Array.copy ad in
       (*This is the table which is later grouped*)
-      let table = BatList.map2i (fun i (ai, aj) (bi,bj) -> (i, Z.(aj - bj), (ai, aj), (bi,bj))) (Array.to_list ad) (Array.to_list bd) in
+      let joinfunction = fun lhs rhs1 rhs2 -> match rhs1, rhs2 with
```
michael-schwarz, line 439: Any particular reason for `fun` instead of having the arguments on the left side?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -673,7 +723,7 @@ struct
         let ri = Environment.var_of_dim t.env r in
         of_coeff xi [(Coeff.s_of_int (-1), xi); (Coeff.s_of_int 1, ri)] o :: acc
     in
-    BatOption.map_default (Array.fold_lefti get_const []) [] t.d
+    BatOption.get t.d |> fun x -> EConj.IntHashtbl.fold (fun lhs rhs list -> get_const list lhs rhs) (snd @@ x) []
```
michael-schwarz, line 726: Does the `@@` do anything here?

michael-schwarz:
Already very curious how the performance is gonna be!

sim642:
Depends on the benchmark results, but I'm wondering if maybe immutable `Map`s would be the better choice even. You get the same sparsity benefit, but `Map`s also can share unmodified subtrees while `Hashtbl`s cannot (due to the mutability).
Moreover, `copy` (which we do a lot) for a `Map` would essentially be identity as opposed to copying of `Hashtbl`s. The only complication is that we sometimes want to exploit mutable `_with` operations on relational domains, which wouldn't be possible with `Map`. A generic lifting via `ref` (to a `Map`) like I already have in #1339 easily provides the necessary conversion.

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: The reason is me being a fool in ocaml

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: Not at all, gonna remove that

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: Still, valid comment. I have not critically looked at this, since I do not yet know, whether the switch to hashtables is really a big gain.

DrMichaelPetter:
> Depends on the benchmark results, but I'm wondering if maybe immutable `Map`s would be the better choice even. You get the same sparsity benefit, but `Map`s also can share unmodified subtrees while `Hashtbl`s cannot (due to the mutability). Moreover, `copy` (which we do a lot) for a `Map` would essentially be identity as opposed to copying of `Hashtbl`s. The only complication is that we sometimes want to exploit mutable `_with` operations on relational domains, which wouldn't be possible with `Map`. A generic lifting via `ref` (to a `Map`) like I already have in #1339 easily provides the necessary conversion.

DrMichaelPetter:
> Depends on the benchmark results, but I'm wondering if maybe immutable `Map`s would be the better choice even. You get the same sparsity benefit, but `Map`s also can share unmodified subtrees while `Hashtbl`s cannot (due to the mutability). Moreover, `copy` (which we do a lot) for a `Map` would essentially be identity as opposed to copying of `Hashtbl`s. The only complication is that we sometimes want to exploit mutable `_with` operations on relational domains, which wouldn't be possible with `Map`. A generic lifting via `ref` (to a `Map`) like I already have in #1339 easily provides the necessary conversion.

This sounds reasonable. Thank you for this in-depth insight into Ocaml data-structures. This is where my lack of experience in Ocaml shines through :disappointed:

DrMichaelPetter:
So, I now have the results of the testrun with the correct hash and equal functions; finally, this eliminated the notorious "fixpoint not reached" errors. Performance wise, it looks as if there is a small runtime gain, but it is shadowed by the rest of the runtime. I may have to repeat the benchmarking with a more thinned out conf, in order to get a real insight into the gains.

DrMichaelPetter:
In the current version, I have switched to maps - this eliminated the copy overhead. Still, some work had to be put into dim_add, which during runtime created a full-blown array of size of the dimension of the state. It's replaced by some compromise between runtime and space now. Runtime benchmarks are pending, depending on the availability of our servers.

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -391,10 +452,17 @@ struct
...
       (*This is the table which is later grouped*)
-      let table = BatList.map2i (fun i (ai, aj) (bi,bj) -> (i, Z.(aj - bj), (ai, aj), (bi,bj))) (Array.to_list ad) (Array.to_list bd) in
+      let joinfunction lhs rhs1 rhs2 = match rhs1, rhs2 with 
+        | Some (ai,aj),Some (bi,bj) -> Some (Z.(aj - bj),(ai,aj),(bi,bj))
+        | None, Some (bi,bj)        -> Some (Z.neg bj   ,Rhs.var_zero lhs,(bi,bj))
+        | Some (ai,aj), None        -> Some (aj         ,(ai,aj),Rhs.var_zero lhs)
+        | _,_                       -> None
+      in
+      let flatten (a,(b1,b2,b3)) = (a,b1,b2,b3) in
+      let table = List.map (flatten) @@ EConj.IntMap.bindings @@ EConj.IntMap.merge joinfunction !(snd ad) !(snd bd) in
```
michael-schwarz, line 463:
```suggestion
      let table = List.map flatten @@ EConj.IntMap.bindings @@ EConj.IntMap.merge joinfunction !(snd ad) !(snd bd) in
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,96 +16,152 @@ open VectorMatrix
...
+  module IntMap = BatMap.Make(Int)
 
-  let show m =
-    Array.fold_right (fun k result -> Equality.show k ^ "\n" ^ result) m ""
+  type t = int * ( Rhs.t IntMap.t ref)
 
-  let hash : t -> int = Array.fold_left (fun acc a -> 31 * acc + Equality.hash a) 0 (* TODO: derive *)
+  let equal (idim,imap) (jdim,jmap) = idim = jdim (* compare the dimensions i.e. max indices *)
+                                      && IntMap.cardinal !imap = IntMap.cardinal !jmap  (* compare the actual number of bindings *)
+                                      && IntMap.fold (fun lh rh acc -> acc && IntMap.mem lh !jmap && IntMap.find lh !jmap = rh) !imap true (* check all bindings *)
```
michael-schwarz, line 36: Any reason why `BatMap.equal` doesn't cut it here? Is it slower?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,96 +16,152 @@ open VectorMatrix
...
+        | Some v, None-> Some (v,Rhs.var_zero k)
+        | None, Some v -> Some (Rhs.var_zero k,v)
+        | None, None -> None) !(snd i) !(snd j) in
+      let lst= IntMap.bindings merg in (* bindings are luckily already sorted by key *)
+      let rec compare_list li = match li with
+        | [] -> 0
+        | (_,(a,b))::rest -> let c = compare a b in
+          if c = 0 then compare_list rest
+          else c
+      in compare_list lst
```
michael-schwarz, line 51: Maybe a comment how this is different from `BatMap.compare`?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,96 +16,152 @@ open VectorMatrix
...
 
-  let make_empty_array len = Array.init len (fun i -> (Some i, Z.zero))
+  let show econ = let show_rhs i (v, o) = Printf.sprintf "var_%d=%s " i (Rhs.show (v, o)) in
+    IntMap.fold (fun lhs rhs acc -> acc ^ show_rhs lhs rhs) econ ""
+
+  let hash : t -> int = fun (dim,x) ->  dim + 13* IntMap.fold (fun k value acc -> 13 * 13 * acc + 31 * k + Rhs.hash value) !x 0 (* TODO: derive *)
```
michael-schwarz, line 56:
```suggestion
  let hash (dim,x) =dim + 13* IntMap.fold (fun k value acc -> 13 * 13 * acc + 31 * k + Rhs.hash value) !x 0 (* TODO: derive *)
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,96 +16,152 @@ open VectorMatrix
...
+
+  let maxentry (_,map) = IntMap.fold (fun lhs (_,_) acc -> max acc lhs) !map 0
+
+  let copy (dim,map) = let res = make_empty_conj dim in
+    snd res := !map;
+    res
+
+  let copy  (dim,map) = timing_wrap "copy" (copy) (dim,map)
+
+  let copy  (dim,map) = timing_wrap "copy" (copy) (dim,map)
```
michael-schwarz, line 84: Duplicated

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -338,22 +394,27 @@ struct
...
-      let res_d = Array.copy d in
+      let res_d = EConj.copy d in
       try
         meet_with_one_conj_with res_d i e;
         {d = Some res_d; env = t.env}
       with Contradiction ->
         {d = None; env = t.env}
 
+  let meet_with_one_conj t i e =
+    let res= meet_with_one_conj t i e in
```
michael-schwarz, line 405:
```suggestion
    let res = meet_with_one_conj t i e in
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -313,20 +369,20 @@ struct
...
     in
-    let (var1, b1) = ts.(i) in
+    let (var1, b1) = EConj.get_rhs ts i in
     (match var, var1 with
      | None, None -> if not @@ Z.equal b b1 then raise Contradiction
      | None, Some h1 -> subst_var ts h1 (None, Z.(b - b1))
      | Some j, None -> subst_var ts j (None, Z.(b1 - b))
      | Some j, Some h1 ->
-       (match ts.(j) with
+       (match (EConj.get_rhs ts j) with
```
michael-schwarz, line 385:
```suggestion
       (match EConj.get_rhs ts j with
```

michael-schwarz (commented):
Nice! I had a quick look and left some comments. I hope I'll find time to look at it in detail next week.

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,96 +16,113 @@ open VectorMatrix
...
 
-module Equality = struct
+module Rhs = struct
   (* (Some i, k) represents a sum of a variable with index i and the number k.
      (None, k) represents the number k. *)
   type t = (int option * GobZ.t) [@@deriving eq, ord, hash, show]
   let zero = (None, Z.zero)
   let var_zero i = (Some i, Z.zero)
-  let to_int x = Z.to_int @@ snd x
+  let to_int (_,x) = Z.to_int x
```
michael-schwarz, line 25: Of course this was already there before, but still: Maybe `offset_to_int` is a better name for this as it only considers the second component?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,96 +16,113 @@ open VectorMatrix
...
 
-module EqualitiesArray = struct
-  include Array
-  type t = Equality.t Array.t [@@deriving eq, ord]
+module EqualitiesConjunction = struct
+  module IntMap = BatMap.Make(Int)
 
-  let show m =
-    Array.fold_right (fun k result -> Equality.show k ^ "\n" ^ result) m ""
+  type t = int * ( Rhs.t IntMap.t ref)
```
michael-schwarz, line 32: I'm a bit confused what purpose it serves to have a `ref` here?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,96 +16,113 @@ open VectorMatrix
...
+        | (tbl,delta,list) (* k<head or list=[] *) -> (IntMap.add (op k delta) (BatOption.map (memobumpvar) refvar, offset) tbl, delta, list)
+      in
+      let (a,_,_) = IntMap.fold (bumpentry) !(snd m) (IntMap.empty,0,offsetlist) in (* Build new map during fold with bumped key/vals *)
+      (op (get_dim m) (Array.length indexes), ref a)
+
+  let modify_variables_in_domain m cols op = let res = modify_variables_in_domain m cols op in if M.tracing then
+      M.tracel "modify_dims" "dimarray bumping with (fun x -> x + %d) at positions [%s] in { %s } -> { %s }"
+        (op 0 1)
+        (Array.fold_right (fun i str -> (string_of_int i) ^ ", " ^ str) cols "")
+        (show @@ !( snd m))
```
michael-schwarz, line 99: Are the `@@`s needed here?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -391,10 +425,17 @@ struct
...
   let join a b =
     let join_d ad bd =
-      (*use copy of ad because result is later saved in there*)
-      let ad = Array.copy ad in
       (*This is the table which is later grouped*)
-      let table = BatList.map2i (fun i (ai, aj) (bi,bj) -> (i, Z.(aj - bj), (ai, aj), (bi,bj))) (Array.to_list ad) (Array.to_list bd) in
+      let joinfunction lhs rhs1 rhs2 = match rhs1, rhs2 with 
+        | Some (ai,aj),Some (bi,bj) -> Some (Z.(aj - bj),(ai,aj),(bi,bj))
+        | None, Some (bi,bj)        -> Some (Z.neg bj   ,Rhs.var_zero lhs,(bi,bj))
+        | Some (ai,aj), None        -> Some (aj         ,(ai,aj),Rhs.var_zero lhs)
```
michael-schwarz, line 432: Could you elaborate again why this is not symmetric in the one is None case? Why does it matter which of the arguments the RHS comes from?

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -345,7 +345,14 @@ struct
       match t.d with
       | None -> bot_env
       | Some m ->
-        let dim_change =
+        let dim_change = (* using Environment.dimchange with swapped parameters is not the originally inteded way for APRON dimchanges if keeping to the spec;
+                            instead, we should make use of Environment.dimchange2, which explicitely provides different format for diff arrays for adding and removing.
+                            We here however produce add_dimenion format in both cases.
```
michael-schwarz, line 350:
```suggestion
                            We here however produce add_dimension format in both cases.
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: I added specifics in the comments for this function now. If you still feel more need for explanation, you may reopen this issue of course :smile:

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: No, initially a brain fart, and then copy-pasted all over the file

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: Since it is not even used, due to us usually deconstructing the tuple anyway, I have removed this strangely named artifact.

I am starting to think, whether modelling the base data as an enumerated data type of either a tuple (int * Z.t) or a single int would have been more appropriate then (int option * Z.t)

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: I start to be more and more confused about that myself. I guess the main reason is historic, due to  the former array having been mutable, and me not daring enough to change that character. Also, I think I need re-education wrt. when to use _with , and what that really demands semantically. Lets's do that in person.

DrMichaelPetter: I have removed ref now

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
-  let empty () = [||]
+  let compare (idim,imap) (jdim,jmap) = let c = compare idim jdim in if c <> 0 then
+      c
+    else IntMap.compare Rhs.compare imap jmap
 
-  let make_empty_array len = Array.init len (fun i -> (Some i, Z.zero))
 
-  (** add new variables to domain with particular indices; translates old indices to keep consistency
+  let show_formatted formatter econ =
+    match (IntMap.bindings econ) with
```
michael-schwarz, line 45:
```suggestion
    match IntMap.bindings econ with 
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
-  let make_empty_array len = Array.init len (fun i -> (Some i, Z.zero))
 
-  (** add new variables to domain with particular indices; translates old indices to keep consistency
+  let show_formatted formatter econ =
+    match (IntMap.bindings econ) with 
+    | [] -> "{}"
+    | [(i,(refvar,off))] -> Printf.sprintf "{ %s=%s }" (formatter i) (Rhs.show_formatted (formatter) (refvar,off))
+    | (i,(refvar,off))::rest -> 
+      let str = List.fold 
+          (fun acc (j,rhs2)-> (Printf.sprintf "%s=%s ∧ %s" (formatter j) (Rhs.show_formatted (formatter) (rhs2)) acc) )
```
michael-schwarz, line 50: Are all these parens needed here?

```suggestion
          (fun acc (j, rhs2)-> Printf.sprintf "%s=%s ∧ %s" (formatter j) (Rhs.show_formatted formatter rhs2) acc) 
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
+          (Printf.sprintf "%s=%s" (formatter i) (Rhs.show_formatted (formatter) (refvar,off))) 
+          rest in
+      "{" ^  str ^ "}"
+
+  let show econ = show_formatted (Printf.sprintf "var_%d") econ
+
+  let hash (dim,x) =dim + 13* IntMap.fold (fun k value acc -> 13 * 13 * acc + 31 * k + Rhs.hash value) x 0 (* TODO: derive *)
```
michael-schwarz, line 57:
```suggestion
  let hash (dim,x) = dim + 13* IntMap.fold (fun k value acc -> 13 * 13 * acc + 31 * k + Rhs.hash value) x 0 (* TODO: derive *)
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
+  let make_empty_conj len = (len, IntMap.empty)
+
+  let get_dim (dim,_) = dim
+
+  let nontrivial (_,econmap) lhs = IntMap.mem lhs econmap
+
+  (** sparse implementation of get rhs for lhs, but will default to no mapping for sparse entries *)
+  let get_rhs (_,econmap) lhs = IntMap.find_default (Rhs.var_zero lhs) lhs econmap
+
+  (* set_rhs, staying loyal to immutable, sparse map underneath *)
```
michael-schwarz, line 70:
```suggestion
  (** set_rhs, staying loyal to immutable, sparse map underneath *)
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
+  let set_rhs (dim,map) lhs rhs = (dim,
+                                   if Rhs.equal rhs Rhs.(var_zero lhs) then
+                                     IntMap.remove lhs map
+                                   else
+                                     IntMap.add lhs rhs map
+                                  )
+  let copy = identity
+
+  let maxentry (_,map) = IntMap.fold (fun lhs (_,_) acc -> max acc lhs) map 0
```
michael-schwarz, line 79:
```suggestion
  let maxentry (_,map) = IntMap.fold (fun lhs _ acc -> max acc lhs) map 0
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
-        Option.map (fun var_index -> var_index - offset_map.(var_index)) var, offs in
-      Array.(filteri (fun i _ -> not @@ Array.mem i indexes) m (* filter out removed variables*)
-             |> map remove_offset_from_array_entry) (* adjust variable indexes *)
-
-  let remove_variables_from_domain m cols = timing_wrap "del_cols" (remove_variables_from_domain m) cols
-
-  let is_empty m = length m = 0
+      let rec bumpentry k (refvar,offset) = function (* directly bumps lhs-variable during a run through indexes, bumping refvar explicitely with a new lookup in indexes *)
+        | (tbl,delta,head::rest) when k>=head            -> bumpentry k (refvar,offset) (tbl,delta+1,rest) (* rec call even when =, in order to correctly interpret double bumps *)
+        | (tbl,delta,list) (* k<head or list=[] *) -> (IntMap.add (op k delta) (BatOption.map (memobumpvar) refvar, offset) tbl, delta, list)
```
michael-schwarz, line 102: `list` as a variable name confuses the syntax highlighters etc. (as seen here)

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
+      in
+      let (a,_,_) = IntMap.fold (bumpentry) (snd m) (IntMap.empty,0,offsetlist) in (* Build new map during fold with bumped key/vals *)
```
michael-schwarz, line 104:
```suggestion
      let (a,_,_) = IntMap.fold bumpentry (snd m) (IntMap.empty,0,offsetlist) in (* Build new map during fold with bumped key/vals *)
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
+       | Some ref_var when ref_var = var ->
+         (* var is the reference variable of its connected component *)
+         (let cluster = IntMap.fold
+              (fun i (ref, offset) l -> if ref = ref_var_opt then i::l else l) (snd d) [] in
+          (* obtain cluster with common reference variable ref_var*)
+          match cluster with (* new ref_var is taken from head of the cluster *)
+          | head :: tail ->
+            let headconst = snd (get_rhs d head) in (* take offset between old and new reference variable *)
+            List.fold (fun map i -> set_rhs map i Z.(Some head, snd (get_rhs d i) - headconst)) d cluster (* shift offset to match new reference variable *)
+          | _ -> d) (* empty cluster means no work for us *)
```
michael-schwarz, line 135:
```suggestion
          | [] -> d) (* empty cluster means no work for us *)
```

this will be the only thing matching here, right?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -454,16 +493,17 @@ struct
...
     if is_bot_env t || is_top t || List.is_empty vars then
       t
     else
-      let m = EArray.copy @@ Option.get t.d in
-      List.iter
-        (fun var ->
-           EArray.forget_variable_with m (Environment.dim_of_var t.env var))
-        vars;
-      {d = Some m; env = t.env}
+      let newm=(List.fold (fun map i -> EConj.forget_variable map (Environment.dim_of_var t.env i)) (Option.get t.d) vars) in
```
michael-schwarz, line 505:
```suggestion
      let newm = List.fold (fun map i -> EConj.forget_variable map (Environment.dim_of_var t.env i)) (Option.get t.d) vars in
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -403,16 +441,17 @@ struct
...
       in
-      let iterate l =
+      let iterate map l =
         match l with
-        | (idx_h, _, (_, b_h), _) :: t ->  List.iter (modify idx_h b_h) l
+        | (idx_h, _, (_, b_h), _) :: t ->  List.fold (fun map' e -> modify map' idx_h b_h e) map l
         | [] -> let exception EmptyComponent in raise EmptyComponent
       in
-      List.iter iterate new_components; Some ad
+      Some (List.fold (iterate) (EConj.make_empty_conj @@ fst ad) new_components)
```
michael-schwarz, line 453:
```suggestion
      Some (List.fold iterate (EConj.make_empty_conj @@ fst ad) new_components)
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -372,15 +400,15 @@ struct
...
+      | None -> tuple_cmp (var, b) (EConj.get_rhs ts i)
+      | Some j -> tuple_cmp (EConj.get_rhs ts i) @@ Tuple2.map2 (Z.add b) (EConj.get_rhs ts j)
     in
     if env_comp = -2 || env_comp > 0 then false else
     if is_bot_env t1 || is_top t2 then true else
     if is_bot_env t2 || is_top t1 then false else
       let m1, m2 = Option.get t1.d, Option.get t2.d in
       let m1' = if env_comp = 0 then m1 else VarManagement.dim_add (Environment.dimchange t1.env t2.env) m1 in
-      GobArray.for_alli (fun i t -> implies m1' t i) m2
+      EConj.IntMap.fold (fun lhs rhs acc -> acc && implies m1' rhs lhs) (snd m2) true (* even on sparse m2, it suffices to check the non-trivial equalities, still present in sparse m2 *)
```
michael-schwarz, line 411: `for_all` (https://ocaml.org/manual/5.1/api/Map.S.html)?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,189 @@ open VectorMatrix
...
+      c
+    else IntMap.compare Rhs.compare imap jmap
 
-  let make_empty_array len = Array.init len (fun i -> (Some i, Z.zero))
 
-  (** add new variables to domain with particular indices; translates old indices to keep consistency
+  let show_formatted formatter econ =
+    match (IntMap.bindings econ) with 
+    | [] -> "{}"
+    | [(i,(refvar,off))] -> Printf.sprintf "{ %s=%s }" (formatter i) (Rhs.show_formatted (formatter) (refvar,off))
```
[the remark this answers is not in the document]
michael-schwarz: But if `rest = []`, fold will return the accumulator unchanged, i.e., `(Printf.sprintf "%s=%s" (formatter i) (Rhs.show_formatted (formatter) (refvar,off)))` which (once put into `{ }` below) is the same as this? Or am I missing something?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
michael-schwarz: Now the order of arguments is flipped, was that on purpose?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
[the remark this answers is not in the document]
DrMichaelPetter: Oh, I didn't account for the initial accumulator to survive that.... that should do the trick, yes

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: `IntMap.for_all` wants `( int -> Rhs.t -> bool )`, and `(implies m1')` was  `(Rhs.t -> int -> bool)` for some reason, so I turned that around to fit together

michael-schwarz (approved):

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -391,10 +412,20 @@ struct
...
+         - rhs2 
+           however, we have to account for the sparseity of EConj maps by manually patching holes with default values *)
+      let joinfunction lhs rhs1 rhs2 = match rhs1, rhs2 with 
+        | Some (ai,aj),Some (bi,bj) -> Some (Z.(aj - bj),(ai,aj),(bi,bj))           (* this is explicitely what we want *)  
+        | None, Some (bi,bj)        -> Some (Z.neg bj   ,Rhs.var_zero lhs,(bi,bj))  (* account for the sparseity of binding 1 *)
+        | Some (ai,aj), None        -> Some (aj         ,(ai,aj),Rhs.var_zero lhs)  (* account for the sparseity of binding 2 *)
+        | _,_                       -> None  (* no binding for lhs in both maps is replicated implicitely in a sparse result map *)
+      in
+      let flatten (a,(b1,b2,b3)) = (a,b1,b2,b3) in
+      let table = List.map flatten @@ EConj.IntMap.bindings @@ EConj.IntMap.merge joinfunction (snd ad) (snd bd) in
```
sim642, line 428: Seems like `joinfunction` could already do the `flatten` logic in the same go.

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -309,8 +309,10 @@ sig
   val equal : t -> t -> bool
   val compare : t -> t -> int
   val hash : t -> int
+  (** empty creates a domain representation of dimension 0 *)
   val empty : unit -> t
```
sim642, line 313: Documentation comments need the empty line to be unambiguous.
```suggestion

  (** empty creates a domain representation of dimension 0 *)
  val empty : unit -> t
```

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -309,8 +309,10 @@ sig
...
   val copy : t -> t
+  (** is_empty is true, if the domain representation has a dimension size of zero *)
   val is_empty : t -> bool
```
sim642, line 316: Documentation comments need the empty line to be unambiguous.
```suggestion

  (** is_empty is true, if the domain representation has a dimension size of zero *)
  val is_empty : t -> bool
```

tests/regression/77-lin2vareq/33-dimarray.c
```diff
@@ -0,0 +1,12 @@
...
+//  special input triggering add_dimarray of [0,0] at line 10; visible via --trace modify_dims
+//      both occurances of variables need to be bumped/shifted by 2 indices
+//  in our case, d = c still needs to hold
+int a;
+b() {}
+void main() {
+  int c;
+  int d = c;
+  b();
+  __goblint_check(d == c); //SUCCESS
```
sim642, line 11: There should be `#include <goblint.h>` (after the top comment).

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,182 @@ open VectorMatrix
...
+  (** set_rhs, staying loyal to immutable, sparse map underneath *)
+  let set_rhs (dim,map) lhs rhs = (dim,
+                                   if Rhs.equal rhs Rhs.(var_zero lhs) then
+                                     IntMap.remove lhs map
+                                   else
+                                     IntMap.add lhs rhs map
+                                  )
+  let copy = identity
+
+  let maxentry (_,map) = IntMap.fold (fun lhs _ acc -> max acc lhs) map 0
```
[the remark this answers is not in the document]
DrMichaelPetter: This even was a relic that did not serve any purpose, so I removed it

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
DrMichaelPetter: I do not really see how I can do that, due to the flatten working on the tuple that gets as snd element the triple produced by the joinfunction, but maybe I missed something here?

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,182 @@ open VectorMatrix
...
+  let dim_remove (ch: Apron.Dim.change) m =
     if Array.length ch.dim = 0 || is_empty m then
       m
     else (
-      Array.modifyi (+) ch.dim;
-      let m' = Array.fold_left (fun y x -> forget_variable_with y x; y) (copy m) ch.dim in
-      remove_variables_from_domain m' ch.dim)
+      let cpy = Array.copy ch.dim in
+      Array.modifyi (+) cpy; (* this is a hack to restore the original https://antoinemine.github.io/Apron/doc/api/ocaml/Dim.html remove_dimensions semantics for dim_remove *)
+      let m' = Array.fold_lefti (fun y i x -> forget_variable y (x)) m cpy in  (* clear m' from relations concerning ch.dim *)
```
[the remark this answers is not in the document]
DrMichaelPetter: The copy is a defensive one, but probably not necessary - as it worked before without that, one could do the modify directly on ch.dim. Maybe I misunderstood this, but having only the fold do the addition would not carry the side-effect to modify_variables_in_domain afaics.

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
```diff
@@ -16,130 +16,182 @@ open VectorMatrix
...
-          (fun offsetcontext index _ ->
-             let newoffset, newlist = shift offsetcontext index in
-             offset_map.(index) <- newoffset;
-             (newoffset, newlist))
-          (0, Array.to_list indexes) offset_map
+      let memobumpvar = (* Memoized version of bumpvar *)
+        if (Array.length indexes < 10) then fun x -> bumpvar 0 x offsetlist else (* only do memoization, if dimchange is significant *)
+          (let h = Hashtbl.create @@ IntMap.cardinal map in (* #of bindings is a tight upper bound on the number of CCs and thus on the number of different lookups *)
+           fun x -> (* standard memoization wrapper *)
+             try Hashtbl.find h x with Not_found ->
```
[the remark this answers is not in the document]
sim642: For `int` keys a polymorphic hashtable is actually fine. It's just nice to have it then explicit in a comment. But turning it into a typed hashtable is also fine.

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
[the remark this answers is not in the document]
sim642: I meant, doesn't something like this still work?
```suggestion
      (* this is a hack to restore the original https://antoinemine.github.io/Apron/doc/api/ocaml/Dim.html remove_dimensions semantics for dim_remove *)
      let m' = Array.fold_lefti (fun y i x -> forget_variable y (i + x)) m ch.dim in  (* clear m' from relations concerning ch.dim *)
```

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
sim642: I mean, the first case of `joinfunction` could return
```ocaml
Some (lhs, Z.(aj - bj),(ai,aj),(bi,bj))
```
Just put the key (`lhs`) already into the result quadruple, just like the `flatten` on `bindings` would.

Although maybe this won't make much of a difference if `table` still needs to be a list. Then bindings (or at least values) still need to be transformed into a list.

DrMichaelPetter: Yeah, table needs to be a list for `.group` to happen later on in the code... still, I like your idea, getting rid of flatten.

src/cdomains/apron/linearTwoVarEqualityDomain.apron.ml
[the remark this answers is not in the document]
sim642: Sorry, I didn't notice that. Then indeed my suggestion doesn't really help.

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -309,11 +309,14 @@ sig
...
   val hash : t -> int
+  val copy : t -> t
+
   (** empty creates a domain representation of dimension 0 *)
   val empty : unit -> t
-  val copy : t -> t
+
   (** is_empty is true, if the domain representation has a dimension size of zero *)
   val is_empty : t -> bool
+  val is_empty : t -> bool
```
sim642, line 319:
```suggestion
  val is_empty : t -> bool
```
Somehow it's doubled now.

sim642 (approved):

PR #1413, opened 2024-04-12
`DefExc`,`Enums`: More precise `of_interval` and `starting` and `ending`

### LINKED ISSUE #1159: `DefExc.of_interval` ignores interval
> I don't think it's problematic, inside this domain the `ikind` is not stored and only used to compute the ranges:
>
> https://github.com/goblint/analyzer/blob/4d18300af5f6ae1b04bdbf29b29775af07fbc96c/src/cdomains/intDomain.ml#L2157
>
> https://github.com/goblint/analyzer/blob/4d18300af5f6ae1b04bdbf29b29775af07fbc96c/src/cdomains/intDomain.ml#L1989

_Originally posted by @michael-schwarz in https://github.com/goblint/analyzer/pull/1154#discussion_r1321208784_

---

> Hmm, that makes it quite misleading though. The result is based on `IBool`, not `(BigInt.zero, BigInt.one)` at all. So right now this is the only way anyway.
>
> But we should open an issue about fixing `DefExc.of_interval` to actually derive a bit range from the actual given interval and intersect it with the ikind's range. I didn't check but `starting` and `ending` should work similarly. It might make a difference because base analysis itself uses `of_interval`/`starting`/`ending`.

_Originally posted by @sim642 in https://github.com/goblint/analyzer/pull/1154#discussion_r1321423977_

michael-schwarz:
Closes #1159

### DIFF
--- a/src/cdomain/value/cdomains/intDomain.ml
+++ b/src/cdomain/value/cdomains/intDomain.ml
@@ -2081,7 +2081,7 @@ struct
       let a, b = Size.min_range_sign_agnostic x, Size.min_range_sign_agnostic y in
       let r = R.join (R.of_interval range_ikind a) (R.of_interval range_ikind b) in
       let ex = if Z.gt x Z.zero || Z.lt y Z.zero then S.singleton Z.zero else  S.empty () in
-      (`Excluded (ex, r))
+      norm ik @@ (`Excluded (ex, r))
 
   let starting ?(suppress_ovwarn=false) ikind x = if Z.compare x Z.zero > 0 then not_zero ikind else top_of ikind
   let ending ?(suppress_ovwarn=false) ikind x = if Z.compare x Z.zero < 0 then not_zero ikind else top_of ikind
--- a/src/cdomain/value/cdomains/intDomain.ml
+++ b/src/cdomain/value/cdomains/intDomain.ml
@@ -2508,7 +2508,7 @@ module Enums : S with type int_t = Z.t = struct
       of_int ik x
     else
       let a, b = Size.min_range_sign_agnostic x, Size.min_range_sign_agnostic y in
-      let r = R.join (R.of_interval ~suppress_ovwarn range_ikind a) (R.of_interval range_ikind b) in
+      let r = R.join (R.of_interval ~suppress_ovwarn range_ikind a) (R.of_interval ~suppress_ovwarn range_ikind b) in
       let ex = if Z.gt x Z.zero || Z.lt y Z.zero then BISet.singleton Z.zero else BISet.empty () in
       norm ik @@ (Exc (ex, r))
 
--- a/src/cdomain/value/cdomains/intDomain.ml
+++ b/src/cdomain/value/cdomains/intDomain.ml
@@ -2080,7 +2080,7 @@ struct
     else
       let a, b = Size.min_range_sign_agnostic x, Size.min_range_sign_agnostic y in
       let r = R.join (R.of_interval range_ikind a) (R.of_interval range_ikind b) in
-      let ex = if Z.geq x Z.zero || Z.leq y Z.zero then S.singleton Z.zero else S.empty () in
+      let ex = if Z.gt x Z.zero || Z.lt y Z.zero then S.singleton Z.zero else  S.empty () in
       (`Excluded (ex, r))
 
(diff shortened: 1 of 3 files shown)


tests/regression/01-cpa/75-refine-def-exc.c
```diff
@@ -0,0 +1,16 @@
+//PARAM:
```
sim642, line 1: This should be more explicit that we're testing def_exc and not intervals.
```suggestion
//PARAM: --enable ana.int.def_exc --disable ana.int.interval
```

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -2074,10 +2074,22 @@ struct
     | _ -> None
   let top_bool = `Excluded (S.empty (), R.of_interval range_ikind (0L, 1L))
 
-  let of_interval ?(suppress_ovwarn=false) ik (x,y) = if Z.compare x y = 0 then of_int ik x else top_of ik
+  let of_interval ?(suppress_ovwarn=false) ik (x,y) =
+    if Z.compare x y = 0 then
+      of_int ik x
+    else
+      let a, b = Size.min_range_sign_agnostic x, Size.min_range_sign_agnostic y in
+      let r = R.join (R.of_interval ~suppress_ovwarn range_ikind a) (R.of_interval range_ikind b) in
```
sim642, line 2082: Should `suppress_ovwarn` be passed to both?
```suggestion
      let r = R.join (R.of_interval ~suppress_ovwarn range_ikind a) (R.of_interval ~suppress_ovwarn range_ikind b) in
```

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -2491,7 +2503,14 @@ module Enums : S with type int_t = Z.t = struct
 
   let of_int ikind x = cast_to ikind (Inc (BISet.singleton x))
 
-  let of_interval ?(suppress_ovwarn=false) ik (x,y) = if x = y then of_int ik x else top_of ik
+  let of_interval ?(suppress_ovwarn=false) ik (x, y) =
+    if Z.compare x y = 0 then
+      of_int ik x
+    else
+      let a, b = Size.min_range_sign_agnostic x, Size.min_range_sign_agnostic y in
+      let r = R.join (R.of_interval ~suppress_ovwarn range_ikind a) (R.of_interval range_ikind b) in
```
sim642, line 2511: Same question here:
```suggestion
      let r = R.join (R.of_interval ~suppress_ovwarn range_ikind a) (R.of_interval ~suppress_ovwarn range_ikind b) in
```

sim642 (approved):

tests/regression/cfg/foo.t/run.t
```diff
@@ -125,6 +125,17 @@
       string: 1 <= a
       type: assertion
       format: C
+  - entry_type: location_invariant
+    location:
+      file_name: foo.c
+      file_hash: $FILE_HASH
+      line: 7
+      column: 2
```
sim642, line 133: Needs to be adapted to #1400 for merge.

tests/regression/cfg/foo.t/run.t
```diff
@@ -213,3 +224,14 @@
       string: 1 <= a
       type: assertion
       format: C
+  - entry_type: location_invariant
+    location:
+      file_name: foo.c
+      file_hash: $FILE_HASH
+      line: 4
+      column: 4
```
sim642, line 232: Needs to be adapted to #1400 for merge.

michael-schwarz:
Interestingly enough, this causes a regression where the increased precision leads to a lower number of contexts, which in turn causes us to lose 5 tasks on termination.

sim642:
This is quite counter-intuitive: how do fewer contexts lead to us failing to prove termination?

michael-schwarz:
Because the Context-sensitive callgraph becomes recursive: Earlier the partial(!) contexts happened to distinguish things by (happy) coincidence. This is no longer the case now.

sim642:
In that case something from #1340 should be able to distinguish the sufficient number of non-recursive cases in a less roundabout way.

michael-schwarz:
Yes, either something from there or one of the ideas from #1422.

PR #1415, opened 2024-04-17
Document implementation-defined behavior following GCC

michael-schwarz:
As a follow-up of #1414 this documents one further instance where we make assumptions (implementation-defined behavior follows GCC).

### DIFF
--- a/docs/user-guide/assumptions.md
+++ b/docs/user-guide/assumptions.md
@@ -17,3 +17,10 @@ _NB! This list is likely incomplete._
 
     See [PR #1414](https://github.com/goblint/analyzer/pull/1414).
 
+2. Implementation-defined behavior usually follows GCC.
+
+    For example, for the behavior on integers, Goblint assumes the compiler behaves as documented by [GCC](https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html). This is also in line with the behavior of Clang in this case.
+
+    > > The result of, or the signal raised by, converting an integer to a signed integer type when the value cannot be represented in an object of that type (C90 6.2.1.2, C99 and C11 6.3.1.3).
+    >
+    > For conversion to a type of width N, the value is reduced modulo 2^N to be within range of the type; no signal is raised.


docs/user-guide/assumptions.md
```diff
@@ -17,3 +17,10 @@ _NB! This list is likely incomplete._
 
     See [PR #1414](https://github.com/goblint/analyzer/pull/1414).
 
+2. Implementation-defined behavior usually follows GCC.
+
+    For example, for the behavior on integers, Goblint assumes the compiler behaves as documented by [GCC](https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html). This is also in line with the behavior of Clang in this case.
+
+    > > The result of, or the signal raised by, converting an integer to a signed integer type when the value cannot be represented in an object of that type (C90 6.2.1.2, C99 and C11 6.3.1.3).
+    >
+    > For conversion to a type of width N, the value is reduced modulo 2^N to be within range of the type; no signal is raised.
```
sim642, line 26: But this is only true with `sem.int.signed_overflow` being `assume_wraparound`, no?

sim642 (changes requested):

docs/user-guide/assumptions.md
sim642: Nothing here is saying that it's conversion _from unsigned_ though. The quote just says "converting an integer".
This is exactly why we need to make the assumptions very explicit.

michael-schwarz: Iirc unsigned types are _integers_ and signed integers are called _signed integers_ in the standard parlance.

sim642: What standard parlance? The C standard very explicitly talks about "signed integers" and "unsigned integers". I couldn't find anywhere in the standard that defines "integer" to mean "natural number".
Standard mathematical terminology also has "integer" including negative numbers.

Having the extra word "unsigned" explicitly in our assumptions only makes things clearer and less ambiguous to everyone. One shouldn't need to be fluent in C standardese to understand our assumptions.
If the wording is not clear/precise enough for us the developers, then it definitely won't be to anyone else (users, students, etc).

michael-schwarz:
Closing for now. To be useful this would need an explanation of a lot of additional details, which we probably don't want to write up right now.

Issue #1005, opened 2023-03-01
Int domain refinement causes fixpoint error


sim642:
Given the following program:
```c
// PARAM: --set ana.int.refinement fixpoint --enable ana.int.interval
// FIXPOINT
#include<assert.h>

int g = 0;

void main()
{
   int i = 0;
   while (1) {
      i++;
      for (int j=0; j < 10; j++) {
         if (i > 100) g = 1;
      }
      if (i>9) i=0;
   }
   return;
}
```
The verify phase fails with the fixpoint error due to
```
Map: g =
 (Unknown int([0,1]),[0,2147483647]) instead of (Unknown int([0,1]),[0,1]).
```

Although we implement interval refinement from exclusion range, this isn't applied here, probably because int domain refinement is disabled for `join` and `widen`. However, some refinement is required here because the two abstractions have the same concretization, so verification should succeed as it is sound.

There is possible need for int domain refinement in precise incremental benchmarks. Otherwise there are spurious precision comparisons due to similar reasons (exclusion range being more precise than interval).

sim642:
Surprisingly, letting `join` not refine, but making `widen` refine passes all our tests, including this one.

https://link.springer.com/chapter/10.1007/978-3-540-77505-8_23 mention an obscure concept of _pre-widening_ as a version of `join` which doesn't refine/reduce. Seems like we might want that in some places (like widening iterations), but a refining `join` in most others (joining over all incoming edges). The non-refinement in `join` and `widen` is only to make widening correct and terminating, elsewhere it should be fine to have that extra precision.

sim642:
Turns out there are other angles to this issue as well. With `--disable ana.opt.hashcons` everything is fine. Hashconsing assumes that lattice operations are idempotent and skips the operation if arguments are equal.

However, int domains violate this assumption, for example:
```
(Unknown int([0,1]),[0,2147483647]) narrow (Unknown int([0,1]),[0,2147483647]) = (Unknown int([0,1]),[0,1])
```

sim642:
Reopening because the root cause isn't fixed by #1186:
> #1005 also mentions a different but related issue that refinement could be applied more often still, but I think this we can save for a separate PR, this PR fixes the immediate issue and makes refinement usable together with hash consing).

michael-schwarz:
New example of this problem not fixed by #1186:

On master (5cd8650a2872341152f48a7d1cd20ac9fa0de994)

```
[...]

PR #1417, opened 2024-04-17
Tracking Benchmark Changes for Thesis

michael-schwarz:
The branch I used to benchmark for my thesis diverges at d8be117080e1f6d0bdc9e8331e888f2d3ac2cf7c.

This draft MR is not intended to be merged but should allow me to keep an overview of the changes and see what should be maybe back-ported.

PR for `master` exists:

- [ ] #1641 addressing #1542 and #1547 
- [ ] #1468 
- [ ] #1642
- [ ] #1643
- [ ] #1644
- [ ] #1645
- [ ] #1646

Probably not universally an improvement, maybe add later on a per-case basis.

- Way to supply suffix to name of `privPrecCompare` output
- Do not colorize `privPrecCompare` output to stderr (in case it is saved to somewhere)
- `exp.priv-prec-dump-proj` option to project precision of dumps to enums, intervals, and def_exc for comparison between different activated domains
- Warn instead of crash if an argument to a string function does not evaluate to an address.
- Fix to force refinement to only happen when dumping precision comparisons (to avoid #1005)

--------------
## Done

Back-ported:

- Typos (c28ca496b78dcab8e1d7191b16b0398896022ea8, e192c8e8d0489535fe5bfa91c7e28da5fc4ab12a, 5073759d37e158441e78dc48abcd80f48933632f, c5d2b3619b6a8b127273ec8418340ab2d9cfd861, 3ad6b396681e0cb936eba09c4d7b6ebed882a0a2)
- Traces: Enable def_exc_widen_by_join for reproducibility for SAS '21 results 3e13e04c021923485553471cb12437aa8f20d7b7
- https://github.com/goblint/analyzer/commit/3a30b0cb9bf532920fbf211408a439fc39717d87 (check whether cast to long long is injective before emitting invariants)

PR landed on `master` concurrently with MS proceeding on this branch (nothing to do):

- #1458 
- #1435
- #1471
- #1481
- #1444
- #1492
- #1430
- #1500 
- #1475 
- #1508

Hotfixes:

- Hotfix for https://github.com/goblint/analyzer/issues/1513 (should be solved by addressing https://github.com/goblint/cil/issues/169 instead)

PR with principled fix exists:

- Quick fix for #1421


michael-schwarz:
```
[Debug] cava.c:1507:9-1507:27 (alloc@sid:459@tid:[main]): write less precise than mine-W-interval
  write: (value:⊤, size:1024, no zeroinit:true)
  less precise than
  mine-W-interval: (value:⊤, size:1024, no zeroinit:true)
  diff: 1024 instead of 1024
 ```
 
 ```
 [Debug] cava.c:3552:5-3553:47 (alloc@sid:2113@tid:[main, input_fifo@cava.c:1931:5-1932:83]): protection more precise than write
  protection: Uninitialized
  more precise than
  write: (trivial arrays:Array: (value:Uninitialized, size:4096, no zeroinit:false), IntDomLifter(intdomtuple):1)
  reverse diff: compound: (trivial arrays:Array: (value:Uninitialized, size:4096, no zeroinit:false), IntDomLifter(intdomtuple):1) not same type as Uninitialized
 ````
 
 both seem fishy.

sim642:
cava has a lot of missing library functions though, so how exactly are you intending to analyze it without losing all precision from invalidations? I'm not sure how correctly those invalidations work with the more involved privatizations.

michael-schwarz:
I now added the same handling we have for the non-relational privatizations upon threadenter with unknown function also to the relational privatization.

While that may be of questionable nature, handling this differently in different privatizations is even more so.

michael-schwarz:
Everything that should be extracted now lives on its own PR. We can thus close this one.

PR #1419, opened 2024-04-18
Missing division by zero treatment in MaySignedOverflow

DrMichaelPetter:
During the treatment of expressions within the MaySignedOverflow query treatment in base.ml, we compute possible value ranges for expressions. We do so for division expressions too, but do not catch a possible division by zero within this treatment.

The first solution proposal to this would is in the case of such an exception just do the safe thing and return true.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1261,11 +1261,11 @@ struct
   *)
   (* TODO: deduplicate https://github.com/goblint/analyzer/pull/1297#discussion_r1477804502 *)
   let rec exp_may_signed_overflow ctx exp =
-    let binop = (GobOption.map2 Z.div )in
     let res = match Cilfacade.get_ikind_exp exp with
       | exception _ -> BoolDomain.MayBool.top ()
       | ik ->
         let checkDiv e1 e2 =
+          let binop = (GobOption.map2 Z.div )in
           match ctx.ask (EvalInt e1), ctx.ask (EvalInt e2) with
           | `Bot, _ -> false
           | _, `Bot -> false
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1261,9 +1261,29 @@ struct
   *)
   (* TODO: deduplicate https://github.com/goblint/analyzer/pull/1297#discussion_r1477804502 *)
   let rec exp_may_signed_overflow ctx exp =
+    let binop = (GobOption.map2 Z.div )in
     let res = match Cilfacade.get_ikind_exp exp with
       | exception _ -> BoolDomain.MayBool.top ()
       | ik ->
+        let checkDiv e1 e2 =
+          match ctx.ask (EvalInt e1), ctx.ask (EvalInt e2) with
+          | `Bot, _ -> false
+          | _, `Bot -> false
+          | `Lifted i1, `Lifted i2 ->
+            ( let divisor_contains_zero = (ID.meet i2 (ID.of_int ik Z.zero) = ID.of_int ik Z.zero)  in
+              if divisor_contains_zero then true else 
+                ( let (min_ik, max_ik) = IntDomain.Size.range ik in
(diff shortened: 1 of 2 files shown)


src/analyses/base.ml
```diff
@@ -1317,7 +1317,7 @@ struct
               | PlusA|PlusPI|IndexPI -> checkBinop e1 e2 (GobOption.map2 Z.(+))
               | MinusA|MinusPI|MinusPP -> checkBinop e1 e2 (GobOption.map2 Z.(-))
               | Mult -> checkBinop e1 e2 (GobOption.map2 Z.mul)
-              | Div -> checkBinop e1 e2 (GobOption.map2 Z.div)
+              | Div -> (try checkBinop e1 e2 (GobOption.map2 Z.div) with Division_by_zero -> true)
```
reb-ddm, line 1320: It seems to me that this only returns `true` if one of the bounds of the interval for `e2` is equal to zero. Should we also return `true` if the interval for `e2` contains zero?

DrMichaelPetter: This is a valid point, since this may have an influence on the monotonicity of other functions, would it not? Is there another way to achieve this without specializing ```checkBinop```?

src/analyses/base.ml
```diff
@@ -1261,9 +1261,29 @@ struct
...
+    let binop = (GobOption.map2 Z.div )in
     let res = match Cilfacade.get_ikind_exp exp with
       | exception _ -> BoolDomain.MayBool.top ()
       | ik ->
+        let checkDiv e1 e2 =
+          match ctx.ask (EvalInt e1), ctx.ask (EvalInt e2) with
+          | `Bot, _ -> false
+          | _, `Bot -> false
+          | `Lifted i1, `Lifted i2 ->
+            ( let divisor_contains_zero = (ID.meet i2 (ID.of_int ik Z.zero) = ID.of_int ik Z.zero)  in
```
michael-schwarz, line 1273: This looks like it should work, but might be fragile because of some domain weirdness (refinements not taking place etc).
I think the safer check is to check whether `ID.is_bot` holds for the meet.

michael-schwarz:
Can you also maybe commit the example where this used to cause an exception and now works?

src/analyses/base.ml
DrMichaelPetter: ... helps when you know the API. Thank you

tests/regression/77-lin2vareq/32-divbzero-in-overflow.c
```diff
@@ -0,0 +1,18 @@
...
+ * 
+ * Fixed in #1419
+ */
+int a;
+char b;
+int c() { 0 == a && 1 / b; }
+void d(int (*func)()) { func(); }
+int main()
+{
+    d(c);
```
michael-schwarz, line 15: Does the example still break before this fix if you inline the body of `c`?
If yes, it might make sense to do this to make the example easier.

michael-schwarz (approved):
LGTM!

tests/regression/77-lin2vareq/32-divbzero-in-overflow.c
DrMichaelPetter: yes, it helps. I overestimated creduce's smartness.

src/analyses/base.ml
```diff
@@ -1261,9 +1261,29 @@ struct
   *)
   (* TODO: deduplicate https://github.com/goblint/analyzer/pull/1297#discussion_r1477804502 *)
   let rec exp_may_signed_overflow ctx exp =
+    let binop = (GobOption.map2 Z.div )in
```
sim642, line 1264: What's this?

sim642: Right now this `binop` is in the entire scope of `exp_may_signed_overflow` and it's specifically for `Z.div` (but the name makes no mention of this.

Maybe this should be pushed down into `checkDiv` for which it was extracted.

DrMichaelPetter: You are completely right about this - moved it deeper into checkDiv now

sim642 (approved):

PR #1423, opened 2024-04-22
Type-safe global query system

sim642:
Queries like `WarnGlobal` and `InvariantGlobal` are only used globally in dummy `ctx`. Moreover, they have `Obj.t` arguments which are unsafe.
Every analysis lifter which adds its own global unknowns must have special cases for converting such arguments for `WarnGlobal` and `InvariantGlobal`. If that is forgotten, then Goblint simply segfaults.

This PR introduces **global queries** using a separate `global_query` "transfer" function and `gctx` global "context" for the queries, which doesn't require all the dummy data.

### TODO
- [ ] Convert `InvariantGlobal` query.
- [ ] Use this for #1394.

### DIFF
--- a/src/analyses/loopTermination.ml
+++ b/src/analyses/loopTermination.ml
@@ -16,11 +16,6 @@ let check_bounded ctx varinfo =
   | `Lifted v -> not (is_top_of (ikind v) v)
   | `Bot -> failwith "Loop counter variable is Bot."
 
-(** We want to record termination information of loops and use the loop
- * statements for that. We use this lifting because we need to have a
- * lattice. *)
-module Statements = Lattice.Flat (CilType.Stmt)
-
 (** The termination analysis considering loops and gotos *)
 module Spec : Analyses.MCPSpec =
 struct
@@ -35,7 +30,7 @@ struct
     include UnitV
     let is_write_only _ = true
   end
-  module G = MapDomain.MapBot (Statements) (BoolDomain.MustBool)
+  module G = BoolDomain.MustBool
 
   let startstate _ = ()
   let exitstate = startstate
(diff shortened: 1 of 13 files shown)


src/analyses/loopTermination.ml
```diff
@@ -52,7 +47,7 @@ struct
         (try
            let loop_statement = find_loop ~loop_counter in
            let is_bounded = check_bounded ctx loop_counter in
-           ctx.sideg () (G.add (`Lifted loop_statement) is_bounded (ctx.global ()));
+           ctx.sideg () is_bounded;
```
michael-schwarz, line 50: Why is the more precise per-loop information removed now?

michael-schwarz (commented):
Nice! I think we should add this to the documentation. 

Also, why is it necessary to stop recording per loop termination information here? I think such information can be very helpful, e.g., when debugging or potentially in the future, when we may want to slice out a program containing only the part for which we could not show termination.

src/analyses/loopTermination.ml
sim642: Because it already exists in the local states where the warnings are emitted per-loop.
The `MustTermLoop` query is completely unused and just extra maintenance burden. Also, the used `MustTermAllLoops` query becomes more efficient because it becomes just a constant-time lookup.

michael-schwarz: Ok, as the side-effect only happens only in postsolving there also isn't any precision advantage to keeping it per loop (which I initially thought we had there).

michael-schwarz:
Could we maybe think of better names for these things? I think `ctx` and `query` are things we are overusing already.

Also, `WarnGlobal` does not really fit into the category `query` anyway, as it returns unit. Maybe for it, a separate transfer function that does not get a `ctx` but only an `ask` and is called per analysis at the end of the analysis would be the more principled thing?

sim642:
Just `ask` isn't enough for these kinds of things though, they also need `global`. For example, the deadlock analysis needs to traverse the entire cycle starting from a global.

michael-schwarz:
True, maybe give it `ask` and `getg` then?

michael-schwarz:
This also looks like it's 95% done?

PR #1427, opened 2024-04-23
Pass `ctx` to context & Simplify callstring-based approaches

### LINKED ISSUE #1379: First-Class Support for Callstring-derived Contexts
After #1340 has landed, we might want to generalize  our constraint system to support Callstring-like contexts in a more natural manner.

I think the only change would be to change

~~~OCaml
  val context : fundec -> D.t -> C.t
~~~

to

~~~OCaml
  val context : fundec -> fundec -> C.t -> D.t -> C.t
~~~

where the additonal fundec would be the caller fundec.

That should be sufficiently general to allow call-string style contexts without keeping the callstring in a flat lattice inside the domain.

michael-schwarz:
- [x] Pass `ctx` as additional argument to `context`
- [x] Simplify callstring-based approaches to not track callstring in local state
- [x] Simplify loopfree callstring
- [x] Investigate how this works with `context` failing for `startvars` 

Closes #1379.

### DIFF
--- a/src/analyses/callstring.ml
+++ b/src/analyses/callstring.ml
@@ -48,13 +48,13 @@ struct
   let name () = "call_"^ CT.ana_name
 
   let context ctx fd _ =
-    let curr_ctx =
+    let curr_context =
       try
         ctx.context ()
       with Enter_func_has_no_context -> CallString.empty
     in
     let elem = CT.new_ele fd ctx in (* receive element that should be added to call string *)
-    CallString.push curr_ctx elem
+    CallString.push curr_context elem
 end
 
 (* implementations of CallstringTypes*)
--- a/src/analyses/callstring.ml
+++ b/src/analyses/callstring.ml
@@ -1,7 +1,7 @@
 (**
-   Call String analysis [call_string] and/or Call Site analysis [call_site]. 
-   The call string limitation for both approaches can be adjusted with the "callString_length" option. 
-   By adding new implementations of the CallstringType, additional analyses can be added. 
+   Call String analysis [call_string] and/or Call Site analysis [call_site].
+   The call string limitation for both approaches can be adjusted with the "callString_length" option.
+   By adding new implementations of the CallstringType, additional analyses can be added.
 *)
 
 open Analyses
@@ -18,7 +18,7 @@ end
 
 (** Analysis with infinite call string or with limited call string (k-CFA, tracks the last k call stack elements).
     With the CT argument it is possible to specify the type of the call string elements *)
-module Spec (CT:CallstringType) : MCPSpec = 
+module Spec (CT:CallstringType) : MCPSpec =
 struct
   include Analyses.IdentitySpec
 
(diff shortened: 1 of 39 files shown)


michael-schwarz:
This changes the behavior for function calls somewhat, in the sense that new threads start with an empty callstring.

src/framework/analyses.ml
```diff
@@ -171,8 +171,11 @@ type ('d,'g,'c,'v) ctx =
 
 exception Ctx_failure of string
 (** Failure from ctx, e.g. global initializer *)
+exception Enter_func_has_no_context
+(** Tried to call ctx() on enter_func. Caught by callstring-based approaches and turned into an initial context *)
```
sim642, line 175: Why does this need a new exception? The point of `Ctx_failure` is that there's just one exception to catch in case `ctx` cannot provide something and it's precisely used for that. This is already well-documented behavior: https://github.com/goblint/analyzer/blob/9c785b129aea4338a0d0258bd6bb88b14a331397/src/framework/analyses.ml#L162-L163

michael-schwarz: I think it more cleanly separates the things that could go wrong. For example, if the selective context activation functor breaks at some point, we will get useful error messages here if `context` is called with `ctx.context` missing from somewhere other than the initial constraints.

src/analyses/loopfreeCallstring.ml
```diff
@@ -1,60 +1,54 @@
...
+        then (`Right new_set)::rem_list (* combine all elements of the call cycle in a set *)
+        else append f new_set (e::prev_list) rem_list
+    in
+    append fd (FundecSet.empty ()) [] current
+
+  let context ctx fd x =
+    let curr_context =
+      try
+        ctx.context ()
+      with Enter_func_has_no_context -> []
```
jerhard, line 49: When one would (hypothetically) configure this analysis to be context-insensitive, would it crash with a `ctx_failure` exception? If so, that may even be desired behavior, but may be confusing to a user if the error message is unclear.

src/analyses/callstring.ml
```diff
@@ -41,46 +43,35 @@ struct
...
-  let enter ctx r f args = [ctx.local, callee_state ctx f]
-
-  let combine_env ctx lval fexp f args fc au f_ask = ctx.local
 
-  let threadenter ctx ~multiple lval v args = [callee_state ctx (Cilfacade.find_varinfo_fundec v)]
+  let context ctx fd _ =
+    let curr_context =
+      try
+        ctx.context ()
+      with Enter_func_has_no_context -> CallString.empty
```
jerhard, line 54: Same comment as for the `LoopFreeCallstring`.

michael-schwarz: (see other comment)

jerhard (approved):

src/framework/analyses.ml
michael-schwarz: @sim642 Is this a deal breaker for you? Otherwise, I would propose to merge so it doesn't become stale.

sim642: I'm now wondering if it'd maybe make more sense to just have a `startcontext` (like `startstate`) that defines the initial context (e.g. as an empty call string). Then there'd be no need to abuse an exception (separate or not) for a situation which isn't exceptional at all: entering the `main` function.

src/analyses/tutorials/taint.ml
```diff
@@ -28,7 +28,12 @@ struct
   module C = Printable.Unit
 
   (* We are context insensitive in this analysis *)
-  let context _ _ = ()
+  let context ctx _ _ = ()
+  let startcontext () = ()
+
+  (* Queries *)
+
+  (** Determines whether a variable [v] is tainted, given a [state]. *)
```
sim642, line 36: These comments appeared randomly now.
```suggestion
```

sim642 (approved):

Issue #1425, opened 2024-04-22
Investigate all "Excellent: ignored check"-s


sim642:
Extracted from #1392:
> Remove `TODO` _only if_ passing thanks to analysis improvement. Otherwise improvement is unsound from a bug.

#### update_suite incremental ast
- [ ] Excellent: ignored check on ../11-restart/13-changed_start_state2.c:10 is now passing!
- [ ] Excellent: ignored check on ../11-restart/13-changed_start_state2.c:11 is now passing!

#### update_suite
- [ ] Excellent: ignored check on tests/regression/04-mutex/58-pthread-lock-return.c:73 is now passing!
- [ ] Excellent: ignored check on tests/regression/04-mutex/58-pthread-lock-return.c:82 is now passing!
- [ ] Excellent: ignored check on tests/regression/04-mutex/58-pthread-lock-return.c:87 is now passing!
- [ ] Excellent: ignored check on tests/regression/57-floats/15-more-library.c:42 is now passing!
- [ ] Excellent: ignored check on tests/regression/57-floats/15-more-library.c:56 is now passing!
- [ ] Excellent: ignored check on tests/regression/57-floats/17-other.c:18 is now passing!

#### update_suite group apron
- [ ] Excellent: ignored check on tests/regression/36-apron/21-traces-cluster-based.c:48 is now passing!
- [ ] Excellent: ignored check on tests/regression/36-apron/21-traces-cluster-based.c:66 is now passing!
- [ ] Excellent: ignored check on tests/regression/36-apron/21-traces-cluster-based.c:69 is now passing!
- [ ] Excellent: ignored check on tests/regression/36-apron/22-traces-write-centered-vs-meet-mutex.c:25 is now passing!
- [ ] Excellent: ignored check on tests/regression/36-apron/34-large-bigint.c:18 is now passing!
- [ ] Excellent: ignored check on tests/regression/36-apron/38-branch-global.c:13 is now passing!
- [ ] Excellent: ignored check on tests/regression/36-apron/42-threadenter-arg.c:6 is now passing!
- [ ] Excellent: ignored check on tests/regression/36-apron/91-mine14-5b-no-threshhold.c:49 is now passing!

#### update_suite group apron2
- [ ] Excellent: ignored check on tests/regression/46-apron2/75-mutex_with_ghosts.c:56 is now passing!

#### update_suite group termination
[...]

PR #1428, opened 2024-04-23
Investigate apron, apron2 and termination "Excellent: ignored check"-s

sim642:
See #1425.

I went over all the apron, apron2 and termination instances from the issue and adapted the tests in one of the ways:
1. Remove TODO because we are soundly more precise now.
2. Remove TODO because we changed defaults to the passing one.
3. Adapt test to check originally intended aspect which we haven't solved but we became more precise due to orthogonal reasons.

### DIFF
--- a/tests/regression/36-apron/21-traces-cluster-based.c
+++ b/tests/regression/36-apron/21-traces-cluster-based.c
@@ -1,6 +1,6 @@
-// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.sv-comp.functions
+// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.sv-comp.functions --set ana.relation.privatization mutex-meet
 extern int __VERIFIER_nondet_int();
-
+// TODO: rename? this doesn't require clusters
 #include <pthread.h>
 #include <goblint.h>
 
@@ -45,7 +45,7 @@ void *t3_fun(void *arg) {
   y = h;
   pthread_mutex_unlock(&A);
   pthread_mutex_unlock(&B);
-  __goblint_check(y == x); // TODO (mutex-meet succeeds, protection unknown)
+  __goblint_check(y == x); // mutex-meet succeeds, protection unknown
   return NULL;
 }
 
@@ -63,9 +63,9 @@ int main(void) {
   x = g;
   y = h;
   pthread_mutex_lock(&B);
-  __goblint_check(y == x); // TODO (mutex-meet succeeds, protection unknown)
+  __goblint_check(y == x); // mutex-meet succeeds, protection unknown
   pthread_mutex_unlock(&B);
   pthread_mutex_unlock(&A);
-  __goblint_check(y == x); // TODO (mutex-meet succeeds, protection unknown)
+  __goblint_check(y == x); // mutex-meet succeeds, protection unknown
   return 0;
 }
--- a/tests/regression/36-apron/22-traces-write-centered-vs-meet-mutex.c
+++ b/tests/regression/36-apron/22-traces-write-centered-vs-meet-mutex.c
@@ -1,4 +1,4 @@
-// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.sv-comp.functions
+// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.sv-comp.functions --set ana.relation.privatization mutex-meet
 extern int __VERIFIER_nondet_int();
 
 #include <pthread.h>
@@ -22,11 +22,11 @@ void *t_fun(void *arg) {
     y = h;
     __goblint_check(y <= x);
     pthread_mutex_lock(&B);
-    __goblint_check(x == y); // TODO (mutex-meet succeeds, write unknown)
+    __goblint_check(x == y); // mutex-meet succeeds, (now-defunct) write unknown
     pthread_mutex_unlock(&B);
     i = x + 31;
     z = i;
-    __goblint_check(z >= x); // TODO (write succeeds, mutex-meet unknown)
+    __goblint_check(z >= x); // TODO ((now-defunct) write succeeds, mutex-meet unknown)
     pthread_mutex_unlock(&A);
     pthread_mutex_unlock(&C);
   }
--- a/tests/regression/36-apron/34-large-bigint.c
+++ b/tests/regression/36-apron/34-large-bigint.c
@@ -1,4 +1,4 @@
-// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag
+// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --set ana.activated[-] expRelation
 #include <goblint.h>
 
 void main() {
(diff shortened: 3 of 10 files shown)


tests/regression/04-mutex/58-pthread-lock-return.c
```diff
@@ -79,14 +80,25 @@ void *t_fun(void *arg) {
...
+  // fake values so test passes on OSX
+  reach1 = 1;
+  int r;
+  if (r)
+    reach2 = 1;
+  else
+    reach3 = 1;
 #endif
+  __goblint_check(reach1); // always reached
+  __goblint_check(reach2); // UNKNOWN (sometimes reached)
```
michael-schwarz, line 100: Should these not be `//UNKNOWN!` then?

tests/regression/36-apron/22-traces-write-centered-vs-meet-mutex.c
```diff
@@ -1,4 +1,4 @@
-// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.sv-comp.functions
+// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.sv-comp.functions --set ana.relation.privatization mutex-meet
```
michael-schwarz, line 1: This is the default anyway, right?

tests/regression/36-apron/34-large-bigint.c
```diff
@@ -9,14 +9,15 @@ void main() {
...
-      __goblint_check(y <= 18446744073709551614ull); // TODO (unknown with D, success with MPQ)
-      __goblint_check(18446744073709551614ull <= z); // TODO (unknown with D, success with MPQ)
-      __goblint_check(z <= 18446744073709551615ull); // TODO (unknown with D, success with MPQ)
+    if (18446744073709551611ull <= x && z <= 18446744073709551614ull) {
+      __goblint_check(18446744073709551611ull <= x); // TODO (unknown with D, success with MPQ)
+      __goblint_check(x <= 18446744073709551612ull); // TODO (unknown with D, success with MPQ)
+      __goblint_check(18446744073709551612ull <= y); // TODO (unknown with D, success with MPQ)
+      __goblint_check(y <= 18446744073709551613ull); // TODO (unknown with D, success with MPQ)
+      __goblint_check(18446744073709551613ull <= z); // TODO (unknown with D, success with MPQ)
+      __goblint_check(z <= 18446744073709551614ull); // TODO (unknown with D, success with MPQ)
```
michael-schwarz, line 18: Why were the constants changed here?

tests/regression/36-apron/22-traces-write-centered-vs-meet-mutex.c
```diff
@@ -22,11 +22,11 @@ void *t_fun(void *arg) {
     y = h;
     __goblint_check(y <= x);
     pthread_mutex_lock(&B);
-    __goblint_check(x == y); // TODO (mutex-meet succeeds, write unknown)
+    __goblint_check(x == y); // mutex-meet succeeds, (now-defunct) write unknown
     pthread_mutex_unlock(&B);
     i = x + 31;
     z = i;
-    __goblint_check(z >= x); // TODO (write succeeds, mutex-meet unknown)
+    __goblint_check(z >= x); // TODO ((now-defunct) write succeeds, mutex-meet unknown)
```
michael-schwarz, line 29: Do we want to remove `write` or make an issue for it? I think `write` and `protection` are both a bit dubious.

tests/regression/36-apron/90-mine14-5b.c
```diff
@@ -46,6 +46,8 @@ int main(void) {
   pthread_create(&id2, NULL, t_fun2, NULL);
   pthread_mutex_lock(&mutex);
   __goblint_check(x==y);
+  __goblint_check(0 <= x);
+  __goblint_check(x <= 10);
```
michael-schwarz, line 50: Are these present in Mine14b, example 5? If not, we should add a remark that we added them.

tests/regression/36-apron/91-mine14-5b-no-threshhold.c
```diff
@@ -46,7 +46,9 @@ int main(void) {
   pthread_create(&id, NULL, t_fun, NULL);
   pthread_create(&id2, NULL, t_fun2, NULL);
   pthread_mutex_lock(&mutex);
-  __goblint_check(x==y); //TODO
+  __goblint_check(x==y);
```
michael-schwarz, line 49: Why does this now pass all of a sudden?

michael-schwarz (approved):
Once the comments are addressed, this should be good to go!

tests/regression/36-apron/21-traces-cluster-based.c
```diff
@@ -1,6 +1,6 @@
-// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.sv-comp.functions
+// SKIP PARAM: --set ana.activated[+] apron --set ana.path_sens[+] threadflag --enable ana.sv-comp.functions --set ana.relation.privatization mutex-meet
 extern int __VERIFIER_nondet_int();
-
+// TODO: rename? this doesn't require clusters
```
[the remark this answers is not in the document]
sim642: Do you know what meaningful aspect this is actually checking? Would be good to have a more meaningful name than traces-even-even-more-rpb.

tests/regression/36-apron/22-traces-write-centered-vs-meet-mutex.c
sim642: `write` doesn't even exist anymore for relational, hence "now-defunct".

I wouldn't call `protection` "dubious". It's just a very straightforward (and inefficient) attempt, that's not wrong per-se (except for the documented unsoundness w.r.t escapes, which is just an implementation limitation). It could be quite interesting with ELINA, though, if it automatically decomposes the massive relation.

tests/regression/36-apron/91-mine14-5b-no-threshhold.c
sim642: Is there any reason this shouldn't have worked all along? `x == y` has nothing to do with widening thresholds. d0364dbbdc8dcb3ace1755729b696ec9ecaaa626 introduces the additional bounds checks where thresholds matter at all.

tests/regression/36-apron/90-mine14-5b.c
sim642: `x == y` isn't in his example either. It's only described in text

> Example 6. Consider the program in Fig. 5.(b) that models an abstract producer/consumer, where X and Y denote the amount of resources. The nonrelational interference analysis is sufficient to prove that X is bounded thanks to the explicit tests on X, but it cannot find any bound on Y . Using the invariant interference abstraction parameterized with the octagon domain [17], it is possible to infer that X = Y is a lock invariant. This information automatically infers a bound on Y from the bound on X.

According to this description, the bounds are the entire goal, so we haven't even been checking the full example so far.

tests/regression/36-apron/21-traces-cluster-based.c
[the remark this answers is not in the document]
michael-schwarz: Not, not off the top of my head, it has been a while since I wrote these.

tests/regression/36-apron/34-large-bigint.c
sim642: They're all reduced by 1 because `z <= 18446744073709551615ull` simply checks the `unsigned long long` type bound, which passes thanks to base's def_exc already. So it wouldn't be testing the intended analysis. This is supposed to be about the sound handling of such numbers which don't fit `double`-based Apron octagons.

PR #1430, opened 2024-04-23
Refactor must-locksets to use definite mvals instead of addresses

sim642:
Currently must-locksets are (reversed) sets of addresses, which can be the usual addresses (as mvals), but the type also allows `UnknownPtr`, `NullPtr` and `StrPtr`. None of these should be in must-locksets, but currently this is enforced by the combination of three different checks in three different files.

The goal of this PR is to refine the type of must-lockset elements from addresses to mvals with definite indices, whereby this correctness property is enforced by the type system.
**This immediately paid off:** it revealed _two cases_ of unsoundness in recursive mutex handling because definite indices were only checked for the must-lockset, but not for the recursive locking counter map (which must duplicate all the set logic to avoid becoming unsound). See #1432.

Hopefully this also allows some `Addr` wrapping and unwrapping to be avoided, which should be less verbose.

### TODO
- [x] Enforce definite indices in must-locks by type.
- [x] Rename `Simple`.
- [x] Add `all` and `is_all` operations to `Simple`.
- [x] Fix `NULL` warning in cram test.

### DIFF
--- a/src/analyses/mutexAnalysis.ml
+++ b/src/analyses/mutexAnalysis.ml
@@ -188,7 +188,6 @@ struct
     let ls, m = ctx.local in
     (* get the set of mutexes protecting the variable v in the given mode *)
     let protecting ~write mode v = GProtecting.get ~write mode (G.protecting (ctx.global (V.protecting v))) in
-    let non_overlapping locks1 locks2 = MustLockset.is_empty @@ MustLockset.inter locks1 locks2 in
     match q with
     | Queries.MayBePublic _ when MustLocksetRW.is_all ls -> false
     | Queries.MayBePublic {global=v; write; protection} ->
@@ -198,7 +197,7 @@ struct
       (* if Mutexes.mem verifier_atomic (Lockset.export_locks ctx.local) then
         false
       else *)
-      non_overlapping held_locks protecting
+      MustLockset.disjoint held_locks protecting
     | Queries.MayBePublicWithout _ when MustLocksetRW.is_all ls -> false
     | Queries.MayBePublicWithout {global=v; write; without_mutex; protection} ->
       let held_locks = MustLocksetRW.export_locks @@ fst @@ Arg.remove' ctx ~warn:false without_mutex in
@@ -207,7 +206,7 @@ struct
       (* if Mutexes.mem verifier_atomic (Lockset.export_locks (Lockset.remove (without_mutex, true) ctx.local)) then
         false
       else *)
-      non_overlapping held_locks protecting
+      MustLockset.disjoint held_locks protecting
     | Queries.MustBeProtectedBy {mutex = Addr mutex; global=v; write; protection} -> (* TODO: non-Addr? *)
       let mutex_lockset = MustLocksetRW.export_locks @@ MustLocksetRW.singleton (LockDomain.MustLock.of_mval mutex, true) in (* TODO: what if non-definite? *)
       let protecting = protecting ~write protection v in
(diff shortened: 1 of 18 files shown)


sim642:
This goes quite a long way to clean things up, but not as far as I would've liked. The oddity is that privatizations get the mutex as an `AddressDomain` address, which can still be any of the odd things (`UnknownPtr`, `NullPtr` or ambiguous index).

@michael-schwarz What are privatizations supposed to do with those? It seems to me that we completely forgot about the non-definite possibilities, which might make things even unsound. Privatizations may have to be adapted to behave more like the must-lockset analysis: only `lock` definite addresses and `unlock` should unlock all `semantic_equal` addresses in case of non-definite.

michael-schwarz:
> What are privatizations supposed to do with those? It seems to me that we completely forgot about the non-definite possibilities, which might make things even unsound.

I think I considered this issue when designing the privatizations. They should still be sound for the original setting according to the following arguments:

- For `lock` & `mine-W`: Incorporating too many values early makes the analysis imprecise, not unsound; All other operations are based on must-locksets, and for background locksets and minimal locksets only definite elements are kept 
- For `write`: lock is a no-op, other actions are based on must-lockset
- For `per-mutex` flavors incl. relational & protection-based: Protection by one of these indefinite addresses means they are always held when the global is accessed, so that is not harmful either

However, these are indeed a bit shaky in some places and may easily be subject to breakage. So in the long run, it does make sense to refactor this as well.

Also, with the new additions such as `atomic` and invariants about unprotected globals, I'm not sure if this still all holds.

src/analyses/mutexAnalysis.ml
```diff
@@ -146,38 +119,57 @@ struct
...
+            (* TODO: don't repeat the same semantic_equal checks *)
+            (* TODO: rmed per lockset element, not aggregated *)
+            (MustLocksetRW.remove_mval mv s, m')
+          else
+            (s, m')
+        )
         else
-          (s, m')
-      | _ -> (rm s, m)
+          (MustLocksetRW.remove_mval mv s, m) (* TODO: should decrement something if may be recursive? *)
```
michael-schwarz, line 162: No, not really, as nothing was incremented in the first place. Our `MutexTypeAnalysis` is flow-insensitive.
The only thing to consider is when that information changes during iteration. However, since the order is must here, no values where the multiplicity is > 0 for non-must-recursive mutexes should be present at fixpoint.

src/cdomain/value/cdomains/mval.ml
```diff
@@ -36,12 +36,13 @@ struct
   let to_cil_exp lv = Lval (to_cil lv)
 
   let is_definite (_, o) = Offs.is_definite o
-  let top_indices (x, o) = (x, Offs.top_indices o)
 end
 
 module MakeLattice (Offs: Offset.Lattice): Lattice with type idx = Offs.idx =
 struct
   include MakePrintable (Offs)
+
```
michael-schwarz, line 44: Why was this moved from Printable?

michael-schwarz (approved):

src/cdomain/value/cdomains/mval.ml
sim642: Because `Printable` indices in offsets aren't required to have a `top`, so `top_indices` cannot be defined for them. That's the case for the new `Z` offsets which are just `Printable`.
Everything else we had before (`IntDomain`, `Unit`, etc) just happened to have `top`.

PR #1431, opened 2024-04-24
More useful stats output by privPrecCompare

michael-schwarz:
Creates list of buckets and boils down comparisons to per-bucket comparisons.


sim642:
Can you explain in what way has the output changed over time and why the buckets are needed? I don't recall `privPrecCompare` being changed.

sim642:
This looks more human-readable indeed. It might be worth considering to port it into Goblint (possibly with slight generalization for all of our compare business).

Issue #1433, opened 2024-04-25
Dubious code duplication in `destabilize_vs`


Red-Panda64:
The destabilize_vs function is an almost verbatim copy of destabilize_normal, except that it is supposed to determine whether any called or start variables were destabilized. This is needed for the side_widen option "cycle".
```ocaml
(* Same as destabilize, but returns true if it destabilized a called var, or a var in vs which was stable. *)
let rec destabilize_vs x = (* TODO remove? Only used for side_widen cycle. *)
  if tracing then trace "sol2" "destabilize_vs %a" S.Var.pretty_trace x;
  let w = HM.find_default infl x VS.empty in
  HM.replace infl x VS.empty;
  VS.fold (fun y b ->
      let was_stable = HM.mem stable y in
      HM.remove stable y;
      HM.remove superstable y;
      HM.mem called y || destabilize_vs y || b || was_stable && List.mem_cmp S.Var.compare y vs
    ) w false
```

For comparison, here is the normally used `destabilize_normal` function:
```ocaml
let rec destabilize_normal x =
  if tracing then trace "sol2" "destabilize %a" S.Var.pretty_trace x;
  let w = HM.find_default infl x VS.empty in
  HM.replace infl x VS.empty;
  VS.iter (fun y ->
      if tracing then trace "sol2" "stable remove %a" S.Var.pretty_trace y;
      HM.remove stable y;
      HM.remove superstable y;
      Hooks.stable_remove y;
      if not (HM.mem called y) then destabilize_normal y
    ) w
```

It seems that `destabilize_vs` has a few peculiarities:
- `destabilize_vs` does not call the Hook (which is probably unintentional)
- `destabilize_vs` uses the short-circuiting of the `||` operator to influence control flow and control recursion
- depending on config options, `destabilize` might refer either to `destabilize_normal` or `destabilize_with_side`. However `destabilize_vs`, used by side-effects when the `cycle` strategy is active, will never behave like `destabilize_with_side`.

In a discussion with @michael-schwarz, we came to the conclusion that it is a bit dangerous to have this kind of code duplication, as it is likely that future updates in the destabilize variants will not be reflected in destabilize_vs. It might be best to drop `destabilize_vs` in favor of a more general `destabilize` function, which takes an optional callback to notify the caller about any destabilized variables. The caller could then perform the check done in `destabilize_vs` where it is needed.

sim642:
I think the reason we kept it so far is because there's a distinct check there
```ocaml
[...]

PR #1434, opened 2024-04-25
Clean up `destabilize_vs`

Red-Panda64:
### Problem
See #1433 

### Proposed Fix
Remove `destabilize_vs`. To replace its functionality, add an optional callback to all `destabilize` functions, allowing the caller to perform the check whether the global destabilized a stable start variable or a called variable when needed without a dedicated `destabilize` implementation. This way, the additional overhead of going through the start variables check only occurs in side with the "cycle" strategy, just as before.

### DIFF
--- a/src/solver/td3.ml
+++ b/src/solver/td3.ml
@@ -284,7 +284,11 @@ module Base =
             let was_stable = HM.mem stable y in
             HM.remove stable y;
             HM.remove superstable y;
-            HM.mem called y || destabilize_vs y || b || was_stable && List.mem_cmp S.Var.compare y vs
+            Hooks.stable_remove y;
+            if not (HM.mem called y) then
+              destabilize_vs y || b || was_stable && List.mem_cmp S.Var.compare y vs
+            else
+              true
           ) w false
       and solve ?reuse_eq x phase =
         if tracing then trace "sol2" "solve %a, phase: %s, called: %b, stable: %b, wpoint: %b" S.Var.pretty_trace x (show_phase phase) (HM.mem called x) (HM.mem stable x) (HM.mem wpoint x);


src/solver/td3.ml
```diff
@@ -496,16 +489,19 @@ module Base =
         (* solve x Widen *)
       in
 
-      let rec destabilize_normal x =
+      let rec destabilize_normal ?(notify_destabilized = fun _ _ _ -> ()) x =
```
michael-schwarz, line 492: Since the fun argument does not change between invocations it might make sense to have an inner recursive function that does not have this parameters but uses the one from the outside scope.

michael-schwarz (approved):
Nice! This is definitely an improvement over the state before where we risked things going out of sync frequently.

src/solver/td3.ml
```diff
@@ -272,21 +272,10 @@ module Base =
       in
       let add_sides y x = HM.replace sides y (VS.add x (try HM.find sides y with Not_found -> VS.empty)) in
 
-      let destabilize_ref: (S.v -> unit) ref = ref (fun _ -> failwith "no destabilize yet") in
-      let destabilize x = !destabilize_ref x in (* must be eta-expanded to use changed destabilize_ref *)
+      let destabilize_ref: (?notify_destabilized:(S.v -> bool -> bool -> unit) -> S.v -> unit) ref = ref (fun ?notify_destabilized _ -> failwith "no destabilize yet") in
+      let destabilize ?(notify_destabilized = (fun _ _ _-> ())) x = !destabilize_ref ~notify_destabilized x in (* must be eta-expanded to use changed destabilize_ref *)
 
-      (* Same as destabilize, but returns true if it destabilized a called var, or a var in vs which was stable. *)
-      let rec destabilize_vs x = (* TODO remove? Only used for side_widen cycle. *)
```
sim642, line 279: Maybe this TODO should be investigated. It would be a lot easier to just remove it than complicate the normal `destabilize` further.

src/solver/td3.ml
```diff
@@ -496,16 +489,19 @@ module Base =
...
         HM.replace infl x VS.empty;
         VS.iter (fun y ->
             if tracing then trace "sol2" "stable remove %a" S.Var.pretty_trace y;
+            let was_stable = HM.mem stable y in
+            let is_called = HM.mem called y in
             HM.remove stable y;
             HM.remove superstable y;
             Hooks.stable_remove y;
-            if not (HM.mem called y) then destabilize_normal y
+            notify_destabilized y is_called was_stable;
```
sim642, line 503: It seems quite wasteful to do additional hashtable lookups whose values are completely ignored in all our usual cases when `side_widen` is not `cycle`.
`Hashtbl.mem` might seem like a cheap operation, but pre-OCaml 5.1 (which we're still on) it's unnecessarily allocating: https://github.com/ocaml/ocaml/pull/11500. This was a problem for #609 at one point.
Moreover, `destabilize` is probably the hottest code in the solver, so these changes must be done very carefully.

src/solver/td3.ml
sim642: On the other hand, these used to be common before 20218e763d0ce5f4e9b0f4b878a1f277cf6764cb separated them. Do you think they were separated without good reason?

sim642 (changes requested):
This seems a bit premature to me. It's not clear whether the original reasons for separating this are no longer valid or not.

src/solver/td3.ml
michael-schwarz: I'm not sure if the removal is a good option at this point in time though, we're actively investigating further strategies to make widening at globals less common / add narrowing, so removing this seems like something we could do after we have investigated this and published our results.

src/solver/td3.ml
michael-schwarz: The problem ist that `stable` is modified right there -- so one can not make this operation lazy easily. The question is whether passing an additional argument whether this computation is required and then checking for that before calling `mem` is really cheaper...

sim642: Conceptually it would be possible to implement a version of `Hashtbl.remove` which returns a boolean (like hashtables do in some other languages) and does so in a single hashtable bucket lookup. But it would be horrible `BatHashtbl`-like `Obj.magic` business.

michael-schwarz: Yes, I went looking hoping BatHashtbl might already have this, but it doesn't 😞.

Red-Panda64: Maybe the `notify_destabilized` callback could also be called before removing y from those sets. Then, the lookup could be done in the callback as needed.

michael-schwarz:
We should add the hook & simplify the control flow and then we can merge it. The other change we do not want to merge for now.

Red-Panda64:
Apologies, I had forgotten this was still open. It is now reduced to unwrapping the obscured control flow around || and the hook call.

michael-schwarz (approved):

Issue #1420, opened 2024-04-18
`dbg.timeout`  somtimes not respected


michael-schwarz:
For benchmarking in my dissertation, I realized that `dbg.timeout` does not seem to be respected always. This was observed on the `phpsys` task from Concrat.

```
/home/goblint/michael-schwarz-dissertation/analyzer/goblint --conf /home/goblint/michael-schwarz-dissertation/analyzer/conf/traces.json --set dbg.timeout 900 --set ana.base.privatization mine-W phpspy.c --enable allglobs --enable dbg.timing.enabled --enable warn.debug -v --sets exp.priv-prec-dump /home/goblint/michael-schwarz-dissertation/bench/bench_result/phpspy.mine-W.prec
```

runs for north of 5h despite a 15min timeout being set.

It still regularly dumps statistics and the number of evals goes up so it should not be a case where some inner function is stuck forever

```
[Info] runtime: 05:25:49.543
[Info] vars: 273065, evals: 797399

[Debug] |rho|=273065
[Debug] |stable|=266244
[Debug] |infl|=273053
[Debug] |wpoint|=17
[Debug] |sides|=12676
[Debug] |side_dep|=0
[Debug] |side_infl|=0
[Debug] |var_messages|=0
[Debug] |rho_write|=0
[Debug] |dep|=249982
[Info] |called|=45
[Debug] Found 9818 contexts for 208 functions. Top 5 functions:
[Debug] 2431	contexts for L:entry state of copy_proc_mem (7233) on phpspy.c:7826:1-7861:1
[Debug] 543	contexts for L:entry state of callgrind_sort_callees (7380) on phpspy.c:19662:1-19670:1
[Debug] 272	contexts for L:entry state of event_handler_fout_snprintf (7371) on phpspy.c:18129:1-18167:1
[Debug] 240	contexts for L:entry state of log_error (1472) on phpspy.c:8044:1-8055:1
[Debug] 235	contexts for L:entry state of event_handler_callgrind (1466) on phpspy.c:18204:1-18251:1

Memory statistics: total=46955584.81MB, max=23249.31MB, minor=46946711.20MB, major=1017657.58MB, promoted=1008783.97MB
    minor collections=22387551  major collections=423 compactions=0


[Info] runtime: 05:25:59.610
[Info] vars: 273065, evals: 797407

[Debug] |rho|=273065
[Debug] |stable|=266245
[Debug] |infl|=273053
[Debug] |wpoint|=13
[...]

Issue #1221, opened 2023-10-20
Use SV-COMP time limit for soft exit


sim642:
SV-COMP and BenchExec can give the task time limit to the tool. Goblint doesn't use this information and just keeps running after the BenchExec soft time limit until it is killed at the hard time limit.

We should use this information to do a soft exit after the given time limit because continuing solving gives us nothing. It just wastes resources and makes benchmarks run longer.

sim642:
BenchExec should send `SIGTERM` on soft time limit and Goblint does react to it (and doesn't even override the signal handler), but somehow it doesn't work under BenchExec? SV-COMP 2023 results contain over 1416 instances of Goblint TIMEOUT with 960s CPU time.

sim642:
I couldn't reproduce this with `runexec` nor under BenchExec. Even with SV-COMP 2023 setup: BenchExec 3.16 and Goblint svcomp23.

michael-schwarz:
If it cannot be reproduced, should we close it?

sim642:
It still happens in Dirk's preruns though.

It might depend on what exactly Goblint is doing at the time it gets the signal. OCaml's signal handling isn't necessarily immediate. If a signal is received, the signal handler just marks it as received but doesn't do anything else. Only at certain points in the OCaml runtime is this signal flag checked and handlers executed.

sim642:
SV-COMP 2025 prerun results contain 1298 instances of Goblint TIMEOUT with 960s CPU time, so the issue still remains after the catch-all exception handler fixes.

PR #1435, opened 2024-04-26
Fix catch-all exception handlers

sim642:
This is an attempt to fix https://github.com/goblint/analyzer/issues/1420#issuecomment-2076884301.

@michael-schwarz You may want to try out of this helps with #1420.

### DIFF
--- a/src/cdomain/value/cdomains/addressDomain.ml
+++ b/src/cdomain/value/cdomains/addressDomain.ml
@@ -238,6 +238,7 @@ struct
     else
       ID.top_of ik
 
+  (** @raise Not_found if set is empty. *)
   let type_of xs =
     Addr.type_of (choose xs) (* TODO: what if ambiguous type? what if chooses NullPtr but also contains Addr with proper type? *)
 
--- a/src/cdomain/value/cdomains/addressDomain.ml
+++ b/src/cdomain/value/cdomains/addressDomain.ml
@@ -241,7 +241,7 @@ struct
   let type_of xs =
     try Addr.type_of (choose xs)
     with (* WTF? Returns TVoid when it is unknown and stuff??? *)
-    | _ -> voidType
+    | _ -> voidType (* TODO: do not catch all! *)
 
   let of_var x = singleton (Addr.of_var x)
   let of_mval x = singleton (Addr.of_mval x)
--- a/src/cdomain/value/cdomains/addressDomain.ml
+++ b/src/cdomain/value/cdomains/addressDomain.ml
@@ -239,9 +239,7 @@ struct
       ID.top_of ik
 
   let type_of xs =
-    try Addr.type_of (choose xs)
-    with (* WTF? Returns TVoid when it is unknown and stuff??? *)
-    | _ -> voidType (* TODO: do not catch all! *)
+    Addr.type_of (choose xs) (* TODO: what if ambiguous type? what if chooses NullPtr but also contains Addr with proper type? *)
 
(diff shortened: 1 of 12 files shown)


michael-schwarz:
Thanks, I'll give it a try today. I just wasted 19h of CPU time on our server because of this issue again.

michael-schwarz:
It worked at least on my machine, I'll give it another try on the server once it's done running benchmarks.

sim642:
I made my best to avoid catching `Timeout.Timeout` (for #1420) and `Sys.Break` (for #1221) anywhere, among some other exceptions that really shouldn't be swallowed like that.
Whether it remedies both issues is yet to be seen, but this is an improvement either way.

I also made the timeout timer more robust: it will keep resending the signal every second. So even if execution happens to be inside a catch-all exception handler at any point (in the future), then there will be more opportunities for the special exception. Currently, if it's swallowed once, it will never come again.

michael-schwarz:
The state as of [a35dc82](https://github.com/goblint/analyzer/pull/1435/commits/a35dc82ac7bcdbff4bc18f4bb20d051c20f3a9b0) does not fix the issue on our server. I'll try again with the new commits added.

src/cdomain/value/cdomains/addressDomain.ml
```diff
@@ -238,10 +238,9 @@ struct
     else
       ID.top_of ik
 
+  (** @raise Not_found if set is empty. *)
   let type_of xs =
-    try Addr.type_of (choose xs)
-    with (* WTF? Returns TVoid when it is unknown and stuff??? *)
-    | _ -> voidType
+    Addr.type_of (choose xs) (* TODO: what if ambiguous type? what if chooses NullPtr but also contains Addr with proper type? *)
```
michael-schwarz, line 243: Are we sure it's safe to remove the `void` fallback for `Not_found`?

michael-schwarz (commented):
It now works also on the server, the timeout is respected.

src/cdomain/value/cdomains/addressDomain.ml
sim642: I just moved it to two call sites in a35dc82ac7bcdbff4bc18f4bb20d051c20f3a9b0, where this otherwise lead to crashes on regression tests. And even those are kind of dubious: why should we ever try to get the type of an empty points-to set?
Even so, the worst that can now happen is that there's an uncaught `Not_found` (if that ever happens), whose stacktrace is easy to see and the fallback can easily be added to such call sites. I wouldn't do that _en masse_ because it itself would cover up problems where we try to operate on empty points-to set when they should've already been handled.

michael-schwarz: I am more worried about the `Addr.type_of` throwing then providing an empty points-to-set.

michael-schwarz: I think it may be crucial for some of our real-world programs. I think we should do a run for those after merging this.

sim642: Sure, I don't added call-site catching where it was needed for our test suite. If some `Not_found` or `Type_of_error` now slips through, it will be uncaught and crashes Goblint with a stacktrace, so it's quite easy to see which call of `AD.type_of` should have a fallback (or better decide what on the result).
So, yes, if they come up (on sv-benchmarks or our bench stuff), we can easily put in such specific catch.

michael-schwarz (approved):

PR #1439, opened 2024-04-30
Fix context gas in g2html

SchiJoha:
Updated code for html representation of the Context Gas to allow a similar representation as for Context Widening

### DIFF
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -518,7 +518,7 @@ struct
   struct 
     include Lattice.Prod (Base1) (Base2)
     let printXml f (x,y) =
-      BatPrintf.fprintf f "<value>\n<map>\n<key>\n%s\n</key>\n%a<key>\nContext Gas Value\n</key>\n%a</map>\n</value>\n" (XmlUtil.escape (Base1.name ())) Base1.printXml x Base2.printXml y
+      BatPrintf.fprintf f "\n%a<analysis name=\"context gas value\">\n%a\n</analysis>" Base1.printXml x Base2.printXml y
   end
   module D = Context_Gas_Prod (S.D) (Lattice.Chain (IntConf)) (* Product of S.D and an integer, tracking the context gas value *)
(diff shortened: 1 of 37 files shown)


src/framework/constraints.ml
```diff
@@ -518,7 +518,7 @@ struct
   struct 
     include Lattice.Prod (Base1) (Base2)
     let printXml f (x,y) =
-      BatPrintf.fprintf f "<value>\n<map>\n<key>\n%s\n</key>\n%a<key>\nContext Gas Value\n</key>\n%a</map>\n</value>\n" (XmlUtil.escape (Base1.name ())) Base1.printXml x Base2.printXml y
+      BatPrintf.fprintf f "\n%a<analysis name=\"context gas value\">\n%a\n</analysis>" Base1.printXml x Base2.printXml y
   end
   module D = Context_Gas_Prod (S.D) (Lattice.Chain (IntConf)) (* Product of S.D and an integer, tracking the context gas value *)
   module C = Printable.Option (S.C) (NoContext)
```
sim642, line 524: Isn't a similar fix needed for contexts, which are collapsed in g2html by default?

SchiJoha: The context is currently identically visualized, as in the default analysis, and in the case of context `None,` `no context` is displayed in the HTML. In this way it is similar to the visualization of the widening lifter. Do you prefer that I update the visualization to make the Option structure on top visible? 

(You mean with collapsed that the different contexts are not separated in "folders" but are all combined in one "folder"?)

sim642: I just mean that without `<analysis>` for `D`, nothing is shown at all. But for `C` everything shows up without such `<analysis>` wrapper?
I just expected them to work the same, but maybe g2html is weird and they don't.

SchiJoha: Yes, everything shows up

michael-schwarz:
@jerhard unless you see any issues here, I'd propose to merge?

jerhard (approved):

PR #1442, opened 2024-05-03
Widening gas

Red-Panda64:
# Description
This PR adds widening gas to the td3 solver. Widening gas permits an unknown to grow N times and still be combined with the regular join. Only once the unknown runs out of gas, the widening operator is used. Widening gas can be applied to regular unknowns and leaf unknowns.
Additionally, this PR extracts the side_widen strategies out of the main solve routine and wraps them in an interface.

## Config Options
The gas budget for regular unknowns is controlled via `solvers.td3.widen_gas`, whereas leaves are controlled via `solvers.td3.side_widen_gas`. Both config options are 0 by default, effectively disabling the feature.
When widening points are configured to be reset with the `solvers.td3.remove-wpoint` option, this too resets that wpoints gas.

## New side_widen Interface and Gas for Leaves
Widening gas co-exists with the other side-widening strategies.
Side widen strategies are meant to determine, when a leaf becomes a widening point. After a leaf has been selected as widening point, future updates to its value are applied with the widen operator. The widening gas can again be used to postpone this:
After a leaf has been marked as widening point, its gas counter is decreased for every increase to its value. Once it hits 0, widening is applied.

There is one exception, namely the side_widen strategy `sides-local`. It does not mark widening points, instead widening only if the incoming side-effect is larger than a previous side-effect that came from the same unknown. This has been adapted to work with widening gas like so: every time that the `sides-local` strategy would widen, the gas is decremented.

Functions of the new interface:
```ocaml
module type WPointSelect =
    sig
      type data

      (** Create data required by this widening point selection strategy.
          The parameters are not necessarily used by all strategies. 
      *)
      val create_data: (S.v -> bool) -> (S.v -> S.v -> unit) -> data
      (** Notifies this strategy that a side-effect has occured.
          This allows the strategy to adapt its internal data structure.
      *)
      val notify_side: data -> S.v option -> S.v -> unit
      (** Whether the destabilization of the side-effected var should record the destabilization
          of called variables and start variables. This information should be passed to [should_mark_wpoint].
      *)
      val record_destabilized_vs: bool
      (** This strategy can decide to prevent widening.
          Note that, if this strategy does not veto, this does not mean that widening
          will necessarily be performed. Nor does a call to this function imply that
          the value of the leaf has grown.
      *)
      val veto_widen: data -> unit HM.t -> VS.t -> S.v option -> S.v -> bool
      (** The value of the leaf has grown. Should it be marked a widening point?
          Even if this function is called, leaf y might already be a widening point
          from an earlier side-effect.
      *)
      val should_mark_wpoint: data -> unit HM.t -> VS.t -> S.v option -> S.v -> bool option -> bool
    end
```

Note that `notify_side` is needed for `unstable-self`, which records side-effects in the `infl` set. `veto_widen` is necessary to implement the above behavior for `sides-local`. `record_destabilized_vs` is currently only used by the `cycle` strategy. It indicates, whether the more expensive version of `destabilize` needs to be called, which also records whether a called or start variable was destabilized.

### DIFF
--- a/src/solver/sideWPointSelect.ml
+++ b/src/solver/sideWPointSelect.ml
@@ -12,7 +12,7 @@ module type S =
     type data
 
     (** Create data required by this widening point selection strategy.
-        The parameters are not necessarily used by all strategies. 
+        The parameters are not necessarily used by all strategies.
         @param is_stable This callback should return whether an unknown is stable.
         @param add_infl  Allows the strategy to record additional influences.
                           This is mainly intended for strategies like unstable-self,
@@ -187,7 +187,7 @@ module UnstableCalled : S =
   end
 
 (** Destabilized a called or start var. Problem: two partial context calls will be precise, but third call will widen the state.
-    If this side destabilized some of the initial unknowns vs, there may be a side-cycle between vs and we should make y a wpoint *)    
+    If this side destabilized some of the initial unknowns vs, there may be a side-cycle between vs and we should make y a wpoint *)
 module Cycle : S =
   functor (S:EqConstrSys) ->
   functor (HM:Hashtbl.S with type key = S.v) ->
--- a/src/solver/sideWPointSelect.ml
+++ b/src/solver/sideWPointSelect.ml
@@ -1,0 +1,222 @@
+(** Strategies for widening leaf unknowns *)
+
+open Batteries
+open ConstrSys
+open Messages
+
+module type S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  sig
+    type data
+
+    (** Create data required by this widening point selection strategy.
+        The parameters are not necessarily used by all strategies. 
+        @param is_stable This callback should return whether an unknown is stable.
+        @param add_infl  Allows the strategy to record additional influences.
+                          This is mainly intended for strategies like unstable-self,
+                          which records the influence of a side-effecting unknown x to the leaf y.
+    *)
+    val create_data: (S.v -> bool) -> (S.v -> S.v -> unit) -> data
+
+    (** Notifies this strategy that a side-effect has occured.
+        This allows the strategy to adapt its internal data structure.
+        @param data The internal state of this strategy
+        @param x    The optional source of the side-effect
+        @param y    The leaf receiving the side-effect
+    *)
+    val notify_side: data -> S.v option -> S.v -> unit
+
+    (** Whether the destabilization of the side-effected var should record the destabilization
+        of called variables and start variables. This information should be passed to [should_mark_wpoint]
+        by the solver.
+    *)
+    val record_destabilized_vs: bool
+
+    (** This strategy can decide to prevent widening.
+        Note that, if this strategy does not veto, this does not mean that widening
+        will necessarily be performed. Nor does a call to this function imply that
+        the value of the leaf has grown.
+        @param data      The internal state of this strategy
+        @param called    Set of called unknowns
+        @param old_sides Prior side-effects to leaf y
+        @param x         Optional source of the side-effect
+        @param y         Side-effected leaf y
+        @return [true]: widening will not be applied; [false]: widening might be applied
+    *)
+    val veto_widen: data -> unit HM.t -> VS.t -> S.v option -> S.v -> bool
+
+    (** The value of the leaf has grown. Should it be marked a widening point?
+        Widening points are widened when their value grows, unless vetoed.
+        Even if this function is called, leaf y might already be a widening point
+        from an earlier side-effect.
+        @param data            The internal state of this strategy
+        @param called          Set of called unknowns
+        @param old_sides       Prior side-effects to leaf y
+        @param x               Optional source of the side-effect
+        @param y               Side-effected leaf y
+        @param destabilized_vs Optional destabilization info, described in [record_destabilized_vs]
+        @return [true]: mark as widening point; [false]: do not mark as widening point
+
+    *)
+    val should_mark_wpoint: data -> unit HM.t -> VS.t -> S.v option -> S.v -> bool option -> bool
+  end
+
+(** Any side-effect after the first one will be widened which will unnecessarily lose precision. *)
+module Always : S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = unit
+
+    let create_data _ _ = ()
+    let notify_side _ _ _ = ()
+    let record_destabilized_vs = false
+    let veto_widen _ _ _ _ _ = false
+    let should_mark_wpoint _ _ _ _ _ _ = true
+  end
+
+(* On side-effect cycles, this should terminate via the outer `solver` loop. TODO check. *)
+(** Never widen side-effects. *)
+module Never : S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = unit
+
+    let create_data _ _ = ()
+    let notify_side _ _ _ = ()
+    let record_destabilized_vs = false
+    let veto_widen _ _ _ _ _ = false
+    let should_mark_wpoint _ _ _ _ _ _ = false
+  end
+
+(** Widening check happens by checking sides.
+    Only widen if value increases and there has already been a side-effect from the same source *)
+module SidesLocal : S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = unit
+
+    let create_data _ _ = ()
+    let notify_side _ _ _ = ()
+    let record_destabilized_vs = false
+    let veto_widen state called old_sides x y =
+      match x with
+      | None -> false
+      | Some x when VS.mem x old_sides -> false
+      | _ -> true
+    let should_mark_wpoint _ _ _ _ _ _ = true
+  end
+
+(** If there was already a `side x y d` from the same program point and now again, make y a widening point.
+    Different from `Sides` in that it will not distinguish between side-effects from different contexts,
+    only the program point matters. *)
+module SidesPP : S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = unit
+    let create_data _ _ = ()
+    let notify_side _ _ _ = ()
+    let record_destabilized_vs = false
+    let veto_widen state called old_sides x y = false
+    let should_mark_wpoint state called old_sides x y _ = match x with
+      | Some x ->
+        let n = S.Var.node x in
+        VS.exists (fun v -> Node.equal (S.Var.node v) n) old_sides
+      | None -> false
+      (* TODO: This is consistent with the previous implementation, but if multiple side-effects come in with x = None,
+          the leaf will never be widened. This is different from SidesLocal *)
+  end
+
+(** If there already was a `side x y d` that changed rho[y] and now again, we make y a wpoint.
+    x caused more than one update to y. >=3 partial context calls will be precise since sides come from different x. TODO this has 8 instead of 5 phases of `solver` for side_cycle.c *)
+module Sides : S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = unit
+
+    let create_data _ _ = ()
+    let notify_side _ _ _ = ()
+    let record_destabilized_vs = false
+    let veto_widen state called old_sides x y = false
+    let should_mark_wpoint state called old_sides x y _ = match x with | Some(x) -> VS.mem x old_sides | None -> true
+  end
+
+(* TODO: The following two don't check if a vs got destabilized which may be a problem. *)
+
+(* TODO test/remove. Check for which examples this is problematic! *)
+(** Side to y destabilized itself via some infl-cycle. Records influences from unknowns to globals *)
+module UnstableSelf : S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = { is_stable: S.v -> bool; add_infl: S.v -> S.v -> unit }
+
+    let create_data is_stable add_infl = { is_stable; add_infl }
+    let notify_side data x y = (match x with None -> () | Some x -> data.add_infl x y)
+    let record_destabilized_vs = false
+    let veto_widen _ _ _ _ _ = false
+    let should_mark_wpoint state called old_sides x y _ = not (state.is_stable y)
+  end
+
+(* TODO test/remove. *)
+(** Widen if any called var (not just y) is no longer stable. Expensive! *)
+module UnstableCalled : S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = { is_stable: S.v -> bool }
+
+    let create_data is_stable _ = { is_stable }
+    let notify_side _ _ _ = ()
+    let record_destabilized_vs = false
+    let veto_widen state called old_sides y x = false
+    let should_mark_wpoint state called old_sides y x _ = HM.exists (fun k _ -> not (state.is_stable k)) called (* this is very expensive since it iterates over called! see https://github.com/goblint/analyzer/issues/265#issuecomment-880748636 *)
+  end
+
+(** Destabilized a called or start var. Problem: two partial context calls will be precise, but third call will widen the state.
+    If this side destabilized some of the initial unknowns vs, there may be a side-cycle between vs and we should make y a wpoint *)    
+module Cycle : S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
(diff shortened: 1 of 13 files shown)


sim642:
Why does this need to be inside the solver? This sounds like _widening delay_, which can be completely solver-independent and can be implemented simply as a lifting of a domain (overriding its `widen`).

michael-schwarz:
@sim642 One could implement a similar feature as a domain lifter, yes. However, the effect would be subtly different: Here the delay (gas) is applied per unknown, whereas tracking it in the domain would mean that the gas value is propagated, e.g., between different loops, or that a complicated mechanism tracking not just one gas value but several inside a map would be required.
In the scope of this thesis, we investigate ideas for integrating a widening gas into the solver as a generalization of its automatic widening point detection which aims to be completely transparent to the writer of the analyses.
We are particularly interested in the values at globals here (but include it for locals for the sake of completeness). @Red-Panda64 will later on also implement a narrowing on globals in order to compare both approaches to having more precise globals.
The strategy of having delays tracked inside the domain is also interesting, and it may be worth comparing the impacts of both some time after the thesis is finished.

sim642:
> @sim642 One could implement a similar feature as a domain lifter, yes. However, the effect would be subtly different: Here the delay (gas) is applied per unknown, whereas tracking it in the domain would mean that the gas value is propagated, e.g., between different loops, or that a complicated mechanism tracking not just one gas value but several inside a map would be required.

I think the standard implementation is just domain lifting where transfer functions set the counter to 0, so none of such propagation occurs.

I'm again worried about the monolithic growth of the TD3 solver. For a long time we've wanted to undo that monolithic structure into a more modular one. While nobody has worked on decomposing the existing solver, at some point we decided to not worsen the problem and implement new things modularly. For example, `DepVals` is just a lifter for the solver and this could easily be as well.

Red-Panda64:
This seems like a policy discussion, so I cannot really say much.
However, as @michael-schwarz said, the widening gas will probably mainly be meaningful for globals, so it's good to control the gas of globals and locals independently. This would be difficult to do when baking it not the domain, right?

sim642:
> However, as @michael-schwarz said, the widening gas will probably mainly be meaningful for globals, so it's good to control the gas of globals and locals independently. This would be difficult to do when baking it not the domain, right?

I think it would even be easier: one can choose whether to wrap an analysis's `module D` or/and `module G` with such a lifter. Either at a global level or even on a per-analysis level.

sim642:
We discussed this during GobCon and here's the summary.

Currently, `module type WPointSelect` and all of its various implementations are somewhat deep in the `Td3.Base` module itself, but the logic that they implement is rather independent of the solver as a whole. They seem to be more-or-less functions on constraint system unknowns (`S.v`).
It would be good if they could be hoisted up (e.g. line 42 of `td3.ml`). It was pointed out that they still depend on `S.v` type and `S.Var` module for some operations, so they might need to be turned into functors with `S.Var` as the argument to make such hoisting possible. Hopefully this is then possible.

The change from `wpoint: unit HM.t` to `wpoint_gas: int HM.t` isn't much of a concern. Especially because `unit HM.t` stores unit values (basically `int`s 0) for values already, so this generalizes the data structure for all of the widenings but shouldn't incur any overhead to the cases when no gas is used.

Some small style improvements should also be done to make this merge-ready. It was mentioned that there are some parentheses which are unneeded (e.g. around `if .. then` conditions).

Red-Panda64:
I have extracted the modules out of td3.Base. For now, I have placed them in their own module in `src/solver`. Since this directory so far only contains solver implementations, I am not sure if it is a good place for them. However, I cannot really see any other place they would belong, because they exclusively affect the solving algorithm. Any opinions on this?

> but the logic that they implement is rather independent of the solver as a whole. 

So far, no other solver implements these strategies, so I have left the selection of the concrete implementation in td3. In their current form, I would not say that the strategies are really independent from td3. Some of the functions/strategies need td3-specific details, like an `add_infl` function, or data structures like the stable and called sets. I think it would be preferable to have them out of the td3.Base module but inside td3.

src/solver/td3.ml
```diff
@@ -252,6 +253,23 @@ module Base =
...
+      let choose_wpoint_select: string -> (module SideWPointSelect.S) = fun conf -> match conf with
+        | "always" -> (module SideWPointSelect.Always)
+        | "never" -> (module SideWPointSelect.Never)
+        | "sides-local" -> (module SideWPointSelect.SidesLocal)
+        | "sides" -> (module SideWPointSelect.Sides)
+        | "sides-pp" -> (module SideWPointSelect.SidesPP)
+        | "unstable-self" -> (module SideWPointSelect.UnstableSelf)
+        | "unstable-called" -> (module SideWPointSelect.UnstableCalled)
+        | "cycle" -> (module SideWPointSelect.Cycle)
+        | _ -> failwith ("Unknown value '" ^ conf ^ "' for option solvers.td3.side_widen!")
```
sim642, line 265: This switching could also be moved to `SideWPointSelect` where these are actually defined.

src/solver/td3.ml
```diff
@@ -315,17 +348,24 @@ module Base =
           in
           HM.remove called x;
           let old = HM.find rho x in (* d from older solve *) (* find old value after eq since wpoint restarting in eq/eval might have changed it meanwhile *)
+
+          (* if value has changed, reduce gas (only applies to marked widening points) *)
+          if not (term && phase == Narrow) && not (S.Dom.equal eqd old) then reduce_gas x;
```
sim642, line 353:
```suggestion
          if not (term && phase = Narrow) && not (S.Dom.equal eqd old) then reduce_gas x;
```

src/solver/td3.ml
```diff
@@ -335,7 +375,8 @@ module Base =
           if not (Timing.wrap "S.Dom.equal" (fun () -> S.Dom.equal old wpd) ()) then ( (* value changed *)
             if tracing then trace "sol" "Changed";
             (* if tracing && not (S.Dom.is_bot old) && HM.mem wpoint x then trace "solchange" "%a (wpx: %b): %a -> %a" S.Var.pretty_trace x (HM.mem wpoint x) S.Dom.pretty old S.Dom.pretty wpd; *)
-            if tracing && not (S.Dom.is_bot old) && HM.mem wpoint x then trace "solchange" "%a (wpx: %b): %a" S.Var.pretty_trace x (HM.mem wpoint x) S.Dom.pretty_diff (wpd, old);
+            (* is there a reason for the redundant widen check? *)
+            if tracing && not (S.Dom.is_bot old) && should_widen x then trace "solchange" "%a (wpx: %b): %a" S.Var.pretty_trace x (should_widen x) S.Dom.pretty_diff (wpd, old);
```
sim642, line 379: What is this comment about?

src/solver/td3.ml
```diff
@@ -346,16 +387,16 @@ module Base =
...
               (solve[@tailcall]) x Widen
             ) else (
-              if term && phase = Widen && HM.mem wpoint x then ( (* TODO: or use wp? *)
+              if term && phase = Widen && should_widen x then ( (* TODO: or use wp? *)
                 if tracing then trace "sol2" "solve switching to narrow %a" S.Var.pretty_trace x;
                 if tracing then trace "sol2" "stable remove %a" S.Var.pretty_trace x;
                 HM.remove stable x;
                 HM.remove superstable x;
                 Hooks.stable_remove x;
                 (solve[@tailcall]) ~reuse_eq:eqd x Narrow
```
sim642, line 396: Is it right to swap `HM.mem wpoint` for `should_widen` here?
This case is taken when in the widening phase the value has stabilized and thus should be switched to narrowing.
If I understand correctly, this now means that if all the widening has isn't used up at `x` (because `should_widen` requires it to be 0), then the solver doesn't switch to narrowing at all.

Previously the notions of "widening point" and "should apply widening now" more-or-less coincided, but with gas they no longer do. I think places like this that check `HM.mem wpoint` but don't actually call `widen` based on that should still check for "widening point", not "should apply widening now".

src/solver/td3.ml
```diff
@@ -395,90 +438,62 @@ module Base =
...
+
           (* make y a widening point if ... This will only matter for the next side _ y.  *)
-          let wpoint_if e =
-            if e then (
-              if tracing then trace "sol2" "side adding wpoint %a from %a" S.Var.pretty_trace y (Pretty.docOpt (S.Var.pretty_trace ())) x;
-              HM.replace wpoint y ()
-            )
-          in
-          match side_widen with
-          | "always" -> (* Any side-effect after the first one will be widened which will unnecessarily lose precision. *)
```
sim642, line 454: These comments should not be removed but moved to the corresponding extracted modules.

src/solver/sideWPointSelect.ml
```diff
@@ -0,0 +1,185 @@
...
+
+    let create_data _ _ = ()
+    let notify_side _ _ _ = ()
+    let record_destabilized_vs = false
+    let veto_widen state called old_sides x y =
+      match x with
+      | None -> false
+      | Some x when VS.mem x old_sides -> false
+      | _ -> true
+    let should_mark_wpoint _ _ _ _ _ _ = true
```
sim642, line 102: Why is this `true` if `sides-local` just had `()` in the old match?

src/solver/sideWPointSelect.ml
```diff
@@ -0,0 +1,185 @@
...
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = { is_stable: S.v -> bool; add_infl: S.v -> S.v -> unit }
+
+    let create_data is_stable add_infl = { is_stable; add_infl }
+    let notify_side data x y = (match x with None -> () | Some x -> data.add_infl x y)
+    let record_destabilized_vs = false
+    let veto_widen _ _ _ _ _ = false
+    (* TODO test/remove. Side to y destabilized itself via some infl-cycle. The above add_infl is only required for this option. Check for which examples this is problematic! *)
+    let should_mark_wpoint state called old_sides x y _ = not (state.is_stable y) (* this is very expensive since it folds over called! see https://github.com/goblint/analyzer/issues/265#issuecomment-880748636 *)
```
sim642, line 148: This was a "unstable_called" comment, but it's now in `UnstableSelf`. There's no `fold` here.

src/solver/sideWPointSelect.ml
```diff
@@ -0,0 +1,185 @@
...
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = { is_stable: S.v -> bool }
+
+    let create_data is_stable _ = { is_stable }
+    let notify_side _ _ _ = ()
+    let record_destabilized_vs = false
+    let veto_widen state called old_sides y x = false
+    (* TODO test/remove. Widen if any called var (not just y) is no longer stable. Expensive! *)
+    let should_mark_wpoint state called old_sides y x _ = HM.fold (fun k _ a -> a || not (state.is_stable k)) called false
```
sim642, line 162: Seems like this would be a bit more efficient (thanks to short-circuiting) with `HM.exists`.

src/solver/sideWPointSelect.ml
```diff
@@ -0,0 +1,185 @@
...
+    let veto_widen state called old_sides x y = false
+    (* destabilized a called or start var. Problem: two partial context calls will be precise, but third call will widen the state. *)
+    (* if this side destabilized some of the initial unknowns vs, there may be a side-cycle between vs and we should make y a wpoint *)
+    let should_mark_wpoint state called old_sides x y cycle =
+      match cycle with
+      | Some cycle ->
+        if tracing && cycle then trace "side_widen" "cycle: should mark wpoint %a" S.Var.pretty_trace y;
+        cycle
+      | None ->
+        Logs.warn "destabilize_vs information not provided to side_widen cycle strategy";
```
sim642, line 183: Why would this be the case now? If it can only happen when the implementation itself is buggy, then this warning is useless to the user. An `assert` (or `Option.get`) in this case would be appropriate.

src/solver/sideWPointSelect.ml
```diff
@@ -0,0 +1,185 @@
...
+                          which records the influence of a side-effecting unknown x to the leaf y.
+    *)
+    val create_data: (S.v -> bool) -> (S.v -> S.v -> unit) -> data
+    (** Notifies this strategy that a side-effect has occured.
+        This allows the strategy to adapt its internal data structure.
+        @param data The internal state of this strategy
+        @param x    The optional source of the side-effect
+        @param y    The leaf receiving the side-effect
+    *)
+    val notify_side: data -> S.v option -> S.v -> unit
```
sim642, line 28: There should be empty lines to separate unrelated declarations and documentation. Currently it's ambiguous which documentation belongs to which function (and odoc will complain).

src/solver/td3.ml
```diff
@@ -368,22 +409,24 @@ module Base =
...
-        (* if HM.mem called y then (init y; let y' = HM.find_default l y (S.Dom.bot ()) in HM.replace rho y y'; HM.remove l y; y') else *)
+          (* if HM.mem called y then (init y; let y' = HM.find_default l y (S.Dom.bot ()) in HM.replace rho y y'; HM.remove l y; y') else *)
         if cache && HM.mem l y then HM.find l y
         else (
           HM.replace called y ();
           let eqd = eq y (eval l x) (side ~x) in
           HM.remove called y;
-          if HM.mem wpoint y then (HM.remove l y; solve y Widen; HM.find rho y)
+          (* TODO: is should_widen correct here? *)
+          if should_widen y then (HM.remove l y; solve y Widen; HM.find rho y)
```
sim642, line 422: Same `HM.mem wpoint` thing here. This seems to be just starting a new widening phase.

src/solver/td3.ml
```diff
@@ -395,90 +438,62 @@ module Base =
...
+
         let widen a b =
           if M.tracing then M.traceli "sol2" "side widen %a %a" S.Dom.pretty a S.Dom.pretty b;
           let r = S.Dom.widen a (S.Dom.join a b) in
           if M.tracing then M.traceu "sol2" "-> %a" S.Dom.pretty r;
           r
         in
         let old_sides = HM.find_default sides y VS.empty in
-        let op a b = match side_widen with
-          | "sides-local" when not (S.Dom.leq b a) -> (
```
sim642, line 420: This `leq` check is not present in `SidesLocal.veto_widen`.

sim642:
> I have extracted the modules out of td3.Base. For now, I have placed them in their own module in `src/solver`. Since this directory so far only contains solver implementations, I am not sure if it is a good place for them. However, I cannot really see any other place they would belong, because they exclusively affect the solving algorithm. Any opinions on this?

It would've been fine to leave them in `td3.ml`, but a separate file might even be better. That directory already has other solver-specific helper modules anyway (e.g. `SolverBox`).

src/solver/td3.ml
Red-Panda64: I think it should be appropriate to use `should_widen` here. If x is a widening point with gas, that means it was never widened. If it was not widened, then narrowing would be pointless.

On the other hand, using `HM.mem wpoint` would also not be harmful either, nor would narrowing cause a large overhead if x does not change. I can change it to a membership check, if you prefer

src/solver/sideWPointSelect.ml
Red-Panda64: In the old implementation, sides-local would not use the `wpoint` set at all for side-effects. Instead, it would only widen if there is an incoming side-effect that increases the global and there had already been a side-effect from the same source. This is very different behavior to the other strategies, which just mark a global as `wpoint` and from then on widen on all increasing contributions.

To put these two behind the same interface, I decided to have the `veto_widen` function. Now, sides-local immediately marks any increasing global as a `wpoint`, but will veto widening whenever the source of the side-effect has not contributed before.
Alternatively, one could replace the `veto_widen` function with something like `widen_anyway`. Then sides-local would not need to mark anything in `wpoint`.

src/solver/td3.ml
Red-Panda64: That is true, but if b is less or equal to a, widening should have no effect, right?

src/solver/td3.ml
Red-Panda64: My comment is strangely worded, but I think it is odd that `should_widen` or previously `HM.mem wpoint x` is re-evaluated here.
Specifically, `let wp = should_widen x` before evaluating `eq` is used to determine, whether widening should be applied. The result of `should_widen` may have changed through `eq`, yet I would assume the value that is logged should reflect the value that is used for the actual widening decision.

But I should definitely be logging with `format_wpoint`, which will not only show whether x is a `wpoint`, but also how much gas it has.

src/solver/td3.ml
Red-Panda64: Ah, I guess you are right, narrowing could still have an effect due to some non-monotonicity.

src/solver/td3.ml
sim642: Reusing `wp` seems sensible here indeed.

src/solver/td3.ml
```diff
@@ -275,6 +281,21 @@ module Base =
       let destabilize_ref: (S.v -> unit) ref = ref (fun _ -> failwith "no destabilize yet") in
       let destabilize x = !destabilize_ref x in (* must be eta-expanded to use changed destabilize_ref *)
 
+      let format_wpoint x = Option.map_default (fun x -> Printf.sprintf "true (gas: %d)" x) "false" (HM.find_option wpoint_gas x) in
```
sim642, line 284: If this is just used for tracing, then something like the following would be nicer:
```suggestion
      let pretty_wpoint () x =
        match HM.find_option wpoint_gas x with
        | None -> Pretty.text "false"
        | Some x -> Pretty.dprintf "true (gas: %d)" x
      in
```
(Don't commit this directly, I have not typechecked this.)

That can be used in tracing with `%a`. It avoids the lookup and string construction if tracing is enabled but something other is being traced. Otherwise unnecessary garbage is created.

src/solver/td3.ml
```diff
@@ -319,17 +340,24 @@ module Base =
...
+                match phase with
+                | Widen -> S.Dom.widen old (S.Dom.join old eqd)
+                | Narrow when GobConfig.get_bool "exp.no-narrow" -> old (* no narrow *)
+                | Narrow ->
+                  (* assert S.Dom.(leq eqd old || not (leq old eqd)); (* https://github.com/goblint/analyzer/pull/490#discussion_r875554284 *) *)
+                  S.Dom.narrow old eqd
+              else (
+                box old eqd
+              )
+            )
```
sim642, line 360: This seems to be just changing the formatting which would be nice to not change unnecessarily.

michael-schwarz (approved):

Issue #1437, opened 2024-04-30
Switch from `BatEnum` to `Seq`


sim642:
In #975 some code was switched from eagerly-constructed lists to the standard `Seq.t`.
However, there's still many places where `BatEnum`/`Enum` is used, which offers similar lazy-sequence functionality.

We could try switching to the more standard `Seq.t` in those `Enum` instances, as long as there's no performance regression from doing so.

After #1443 and #1784, only a handful of instances remain:
- [ ] `BatFile.lines_of` and `BatIO.lines_of`. We'll have to find a sensible way of doing that.
- [ ] `BatCache`. We only use `manual_cache` which is just a thin wrapper around `Hashtbl`. We could easily just have a version in `Goblint_std`.
- [ ] `BatEnum.uniq`. There is no `Seq.uniq` or equivalent. We probably could just avoid that anyway by deduplicating smarter using sets.

PR #1443, opened 2024-05-03
Switch from `BatEnum` to `Seq`

karoliineh:
A try on switching to the more standard `Seq.t` in `Enum` instances where possible as proposed in #1437

Some things that I noticed:
- `Enum.cartesian_product` is called `product` in `Seq`
- `Enum.uncombine` is called `unzip` (or `split`) in `Seq`
- `Enum.fold` is called `fold_left` in `Seq`
- There is no corresponding function to `Enum.uniq` in `Seq`
- There is no `to_seq` in `BatHashtbl`
- There is no `to_seq_keys` in `BatHashtbl` as in the Ocaml standard library `Hashtbl`

### TODO
- [x] Fix Gobview.
- [x] Wait for https://github.com/ocaml-batteries-team/batteries-included/pull/1133 to release? https://github.com/ocaml-batteries-team/batteries-included/pull/1135
- [x] Bump OCaml lower bound to 4.14? Many `Seq` functions are that new. (#1448)
- [x] Merge https://github.com/goblint/gobview/pull/51.


sim642:
> What is still missing here?

A new batteries release would allot switching many `Hashtbl`-related places: https://github.com/ocaml-batteries-team/batteries-included/pull/1135.

sim642:
Batteries 3.9.0 has now been released, so that should allow porting some of the previously problematic places as well.

src/common/framework/cfgTools.ml
```diff
@@ -405,24 +405,24 @@ let createCFG (file: file) =
...
+                    |> List.of_seq
                   in
                   let targets = match targets with
                     | [] ->
                       let scc_node =
-                        NH.keys scc.nodes
-                        |> BatList.of_enum
+                        NH.to_seq_keys scc.nodes
+                        |> BatList.of_seq
                         |> BatList.min ~cmp:Node.compare (* use min for consistency for incremental CFG comparison *)
```
sim642, line 426: It's a bit silly to turn it into a list just to find the minimum, but there isn't `Seq.min` (or `BatSeq.min` with `~cmp`).
It's not worse than before, but leaving a TODO here would be good.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -295,27 +295,27 @@ struct
...
-      |> List.enum
-      |> Enum.map (Tuple2.map2 (fun e ->
+      |> List.to_seq
+      |> Seq.map (Tuple2.map2 (fun e ->
           match Convert.texpr1_of_cil_exp ask nd env e (Lazy.from_val no_ov) with
           | texpr1 -> Some texpr1
           | exception Convert.Unsupported_CilExp _ -> None
         ))
-      |> Enum.partition (fun (_, e_opt) -> Option.is_some e_opt)
+      |> Seq.partition (fun (_, e_opt) -> Option.is_some e_opt)
```
sim642, line 304: There maybe should be `Seq.memoize` before/after this stuff to avoid duplicating computations when working on `supported` and `unsupported` later on. The documentation for `Seq.partition` warns about this.

src/cdomains/apron/apronDomain.apron.ml
```diff
@@ -360,27 +360,27 @@ struct
...
-      |> List.enum
-      |> Enum.map (Tuple2.map2 (fun e ->
+      |> List.to_seq
+      |> Seq.map (Tuple2.map2 (fun e ->
           match Convert.texpr1_of_cil_exp ask nd env e no_ov with
           | texpr1 -> Some texpr1
           | exception Convert.Unsupported_CilExp _ -> None
         ))
-      |> Enum.partition (fun (_, e_opt) -> Option.is_some e_opt)
+      |> Seq.partition (fun (_, e_opt) -> Option.is_some e_opt)
```
sim642, line 369: Same `memoize` thing here.

src/analyses/threadId.ml
```diff
@@ -170,14 +170,14 @@ struct
...
+    let tids = Hashtbl.to_seq !tids in
+    let uniques = Seq.filter_map (fun (a,b) -> if Thread.is_unique a then Some a else None) tids in
+    let non_uniques = Seq.filter_map (fun (a,b) -> if not (Thread.is_unique a) then Some a else None) tids in
+    let uc = Seq.length uniques in
+    let nc = Seq.length non_uniques in
     M.debug_noloc ~category:Analyzer "Encountered number of thread IDs (unique): %i (%i)" (uc+nc) uc;
-    M.msg_group Debug ~category:Analyzer "Unique TIDs" (List.map (fun tid -> (Thread.pretty () tid, None)) uniques);
-    M.msg_group Debug ~category:Analyzer "Non-unique TIDs" (List.map (fun tid -> (Thread.pretty () tid, None)) non_uniques)
+    M.msg_group Debug ~category:Analyzer "Unique TIDs" (List.map (fun tid -> (Thread.pretty () tid, None)) @@ List.of_seq uniques);
+    M.msg_group Debug ~category:Analyzer "Non-unique TIDs" (List.map (fun tid -> (Thread.pretty () tid, None)) @@ List.of_seq non_uniques)
```
sim642, line 180: It might be better to turn both into lists and still use `List.length` on them to avoid duplicating the `Seq.filter`/`partition` computations.

src/analyses/extractPthread.ml
```diff
@@ -132,11 +132,9 @@ module Tbls = struct
       Hashtbl.find table k |> Option.default_delayed new_value_thunk
 
 
-    let get_key v =
-      table |> Hashtbl.filter (( = ) v) |> Hashtbl.keys |> Enum.get
+    let get_key v = table |> Hashtbl.to_seq |> Seq.find_map (fun (k,v') -> if v' = v then Some k else None)
```
sim642, line 135: Should leave a TODO here about the inefficiency of looking up by value from a `Hashtbl`.

src/analyses/extractPthread.ml
```diff
@@ -146,12 +144,10 @@ module Tbls = struct
 
     let get k = Hashtbl.find table k
 
-    let get_key v = table |> Hashtbl.enum |> List.of_enum |> List.assoc_inv v
+    let get_key v = table |> Hashtbl.bindings |> List.assoc_inv v
```
sim642, line 147: This looks like it could be similar to the `get_key` above to avoid constructing the intermediate list.

src/analyses/extractPthread.ml
```diff
@@ -146,12 +144,10 @@ module Tbls = struct
...
   end
 
-  let all_keys_count table =
-    table |> Hashtbl.keys |> List.of_enum |> List.length
-
+  let all_keys_count table = table |> Hashtbl.to_seq_keys |> Seq.length
```
sim642, line 150: Wouldn't this just be `Hashtbl.length` or something?

src/analyses/extractPthread.ml
```diff
@@ -171,9 +167,8 @@ module Tbls = struct
     let extend k v = Hashtbl.modify_def Set.empty k (Set.add v) table
 
     let get_fun_for_tid v =
-      Hashtbl.keys table
-      |> List.of_enum
-      |> List.find (fun k ->
+      Hashtbl.to_seq_keys table
+      |> Seq.find (fun k ->
           Option.get @@ Hashtbl.find table k |> Set.exists (( = ) v))
```
sim642, line 172: This also looks like some `get_key` thing. With `to_seq` one could avoid the `find` at least.

src/analyses/extractPthread.ml
```diff
@@ -334,14 +329,12 @@ end = struct
...
-      |> Hashtbl.values
-      |> List.of_enum
-      |> List.map Set.elements
-      |> List.concat
-    in
-    List.filter_map (f % action_of_edge) all_edges
+    table
+    |> Hashtbl.to_seq_values
+    |> Seq.map Set.to_seq
+    |> Seq.concat
```
sim642, line 335: This is `Seq.concat_map`.

src/analyses/extractPthread.ml
karoliineh: `Hashtbl.length` documentation says:
> `Hashtbl.length tbl` returns the number of bindings in `tbl`. It takes constant time. Multiple bindings are counted once each, so `Hashtbl.length` gives the number of times `Hashtbl.iter` calls its first argument.

But I am not entirely sure that the table here always has only one value per key.

sim642:
There's now a conflict with PR #1623 which added a component to `IntDomTuple`.

michael-schwarz:
@karoliineh: Is this now ready for review?

michael-schwarz:
Should we add semgrep rules to prevent adding new code that uses `BatEnum`?

sim642 (approved):

Issue #1440, opened 2024-04-30
Relational mutex-meet unsound on ConcurrencySafety-Main


sim642:
During #1394 I implemented ghost witness generation for relational mutex-meet. The benchmarks revealed unsoundness which is also present on master (2fa4f55e682da3ad937e394fe20449eeef4c9634) in:
- [x] ldv-linux-3.14-races/linux-3.14--drivers--net--irda--nsc-ircc.ko.cil (#1444)
- [x] ldv-linux-3.14-races/linux-3.14--drivers--net--irda--w83977af_ir.ko.cil (#1444)
- [ ] pthread-driver-races/char_generic_nvram_read_nvram_write_nvram
- [x] weaver/popl20-more-buffer-mult.wvr (does not use locks but only `__VERIFIER_atomic_*()`) (#1441)
- [x] weaver/popl20-more-buffer-series.wvr (does not use locks but only `__VERIFIER_atomic_*()`) (#1441)

The unsoundness only reveals itself when activating the Apron analysis unconditionally.

For example
```console
./goblint --conf conf/svcomp.json --set ana.specification ../sv-benchmarks/c/properties/unreach-call.prp ../sv-benchmarks/c/weaver/popl20-more-buffer-mult.wvr.c --set ana.activated[+] apron
```
has 60/100 lines dead.
Without unconditional apron, it has 1/104 lines dead. Oddly, if loop unrolling is disabled, things are also sound.

michael-schwarz:
For `pthread-driver-races/char_generic_nvram_read_nvram_write_nvram` we get:

```
[Error][Analyzer][Unsound] both branches over condition '*ppos >= (loff_t )nvram_len' are dead (../../sv-comp/sv-benchmarks/c/pthread-driver-races/char_generic_nvram_read_nvram_write_nvram.i:6715:6-6715:24)
[Error][Analyzer][Unsound] both branches over condition '*ppos >= (loff_t )nvram_len' are dead (../../sv-comp/sv-benchmarks/c/pthread-driver-races/char_generic_nvram_read_nvram_write_nvram.i:6732:6-6732:24)
SV-COMP result: true
```

michael-schwarz:
At first glance, it seems that for `pthread-driver-races/char_generic_nvram_read_nvram_write_nvram` our result is correct, as the parameter `ppos` points to `whoop_loff_t` for which a blob is allocated but never initialized. I'll try to do a detailed write-up tomorrow.

[...]

PR #1444, opened 2024-05-06
Fix relation analysis ignoring extern inits

sim642:
Fixes two instances from #1440.

For some reason the relation analysis simply ignored extern initializers, which meant that those variables evaluated to bottom in conditionals.

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -247,25 +247,21 @@ struct
   (* Basic transfer functions. *)
   let assign ctx (lv:lval) e =
     let st = ctx.local in
-    if !AnalysisState.global_initialization && e = MyCFG.unknown_exp then
-      st (* ignore extern inits because there's no body before assign, so env is empty... *)
-    else (
-      let simplified_e = replace_deref_exps ctx.ask e in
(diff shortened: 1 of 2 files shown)


src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -247,25 +247,21 @@ struct
   (* Basic transfer functions. *)
   let assign ctx (lv:lval) e =
     let st = ctx.local in
-    if !AnalysisState.global_initialization && e = MyCFG.unknown_exp then
-      st (* ignore extern inits because there's no body before assign, so env is empty... *)
```
michael-schwarz, line 251: It seems strange why this was a problem earlier, but isn't a problem now. Did you test that this does not cause us to have top everywhere, i.e., also for other variables that get a proper initialization later?

Also what is the `env` here now?

sim642: It's hard to git blame this due to some massive changes at the affine equalities time, but I suspect it might be some very old relic. Because global initializers are similar (there's no `body` to explicitly add them all) but there's no problem with those.
Nowadays, `write_global` just adds the assigned variable to the environment: https://github.com/goblint/analyzer/blob/2fa4f55e682da3ad937e394fe20449eeef4c9634/src/analyses/apron/relationAnalysis.apron.ml#L136-L139

This might be a remainder from the days when our Apron analysis didn't support globals at all.

michael-schwarz (approved):

Issue #1364, opened 2024-02-15
Make all usages of `Obj` explicit


sim642:
In 7edfff066099e9b96ceee39ea3250a62707727b1 of #1326 an `Obj`-related bug was found which wasn't caught by the OCaml typechecker due to the `Obj`.
Although there's no simple way to avoid the use of `Obj` in `MCP`, we should always be explicit about using `Obj`, instead of `open Obj` and bringing all the wildly unsafe functions into scope. This would make it clearer and searchable, where the type system is being bypassed.

- [x] Add Semgrep rule for `open Obj`.
- [x] Remove all `open Obj`.
- [x] Prefix all calls with `Obj.`.

PR #1445, opened 2024-05-06
Make all usages of `Obj` explicit

karoliineh:
Solves #1364

### DIFF
--- a/src/analyses/mCPRegistry.ml
+++ b/src/analyses/mCPRegistry.ml
@@ -139,7 +139,6 @@ struct
 
   open DLSpec
   open List
-  open Obj
 
   type t = (int * unknown) list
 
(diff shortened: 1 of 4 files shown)


src/analyses/mCPRegistry.ml
```diff
@@ -139,7 +139,6 @@ struct
 
   open DLSpec
   open List
-  open Obj
 
   type t = (int * unknown) list
```
sim642, line 143: I think we should get rid of `type unknown = Obj.t` while we're at it. This `unknown` shows up in all the MCP type signatures and obfuscates the fact that the unsafe `Obj` module is being used.

michael-schwarz (approved):
Good idea to also add it to the ~autotuner~ semgrep (sorry, brain fart)!

sim642 (approved):

PR #1448, opened 2024-05-07
Bump OCaml lower bound to 4.14

sim642:
### TODO
- [x] Merge https://github.com/goblint/gobview/pull/45.
- [x] Check unlocked CI job https://github.com/goblint/analyzer/actions/runs/8985410592.

### DIFF
--- a/.github/workflows/unlocked.yml
+++ b/.github/workflows/unlocked.yml
@@ -21,9 +21,6 @@ jobs:
           - 5.0.x
           - ocaml-variants.4.14.0+options,ocaml-option-flambda
           - 4.14.x
-          - 4.13.x
-          - 4.12.x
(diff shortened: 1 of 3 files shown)


michael-schwarz:
Should we also bump the OCaml version in the lockfile to `4.14.1`?

sim642:
> Should we also bump the OCaml version in the lockfile to `4.14.1`?

There's also 4.14.2 and I thought about it, but I wouldn't do it in this PR because that kind of forces everyone to reinstall their switches.

I think I'll do a separate PR at some point to upgrade all of our locked dependencies, not just the compiler. Then it's less annoying if the switch has to be recompiled just once.

michael-schwarz (approved):

PR #1450, opened 2024-05-09
Improve autotune for octagon

karoliineh:
I analyzed the SV-COMP NoOverflows tasks where Mopsa succeeded, but we failed, and uncovered some tasks where we succeeded when activating apron analysis but failed with autotune.

This PR proposes a fix that allows us to further succeed in 11 tasks (`nla-digbench-scaling/sqrt1-.*`) with autotune.

TODO:
- [x] Generalize the case handling operations inside `LNot`

### DIFF
--- a/src/autoTune.ml
+++ b/src/autoTune.ml
@@ -337,6 +337,7 @@ let isGoblintStub v = List.exists (fun (Attr(s,_)) -> s = "goblint_stub") v.vatt
 let rec extractVar = function
   | UnOp (Neg, e, _) -> extractVar e
   | Lval ((Var info),_) when not (isGoblintStub info) ->  Some info
+  | CastE (_, e) -> extractVar e
   | _ -> None
 
 let extractOctagonVars = function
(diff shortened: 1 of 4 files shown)


src/autoTune.ml
```diff
@@ -375,6 +376,15 @@ class octagonVariableVisitor(varMap, globals) = object
...
     | Lval ((Var info),_) when not (isGoblintStub info) ->  handle varMap 1 globals (Some (`Right info)) ; SkipChildren
+    | UnOp (LNot, BinOp (op, e1,e2, (TInt _)),_) when isComparison op -> (* TODO: Should be generalized *)
+      handle varMap 5 globals (
+        match extractVar e1, extractVar e2 with
+        | Some a, Some b -> Some (`Left (a,b))
+        | Some a, None
+        | None, Some a -> if isConstant e1 then Some (`Right a) else None
+        | _,_ -> None
+      );
+      DoChildren
```
DrMichaelPetter, line 387: Forgive my ignorance, this is the first time that I look into the autotuner. Is there a particular reason for 

- explicitely handling only one level of `LNot`? Does a recursive match make sense here?
- is there a particular reason to not reuse the `BinOp` case?

src/autoTune.ml
```diff
@@ -375,6 +376,15 @@ class octagonVariableVisitor(varMap, globals) = object
         DoChildren
       )
     | Lval ((Var info),_) when not (isGoblintStub info) ->  handle varMap 1 globals (Some (`Right info)) ; SkipChildren
+    | UnOp (LNot, BinOp (op, e1,e2, (TInt _)),_) when isComparison op -> (* TODO: Should be generalized *)
+      handle varMap 5 globals (
+        match extractVar e1, extractVar e2 with
+        | Some a, Some b -> Some (`Left (a,b))
+        | Some a, None
+        | None, Some a -> if isConstant e1 then Some (`Right a) else None
```
DrMichaelPetter, line 384: Ok, next try :smiley: : I know that is is done similarly in the `extractOctagonVars` part a little bit above, but why do we drop `a` when `e1` is constant? Should this not be symmetric, and `e2` be checked? Is the check providing anything more then extractVar already does anyway?

sim642: If I were to guess, it would make sense to be symmetric. The asymmetry could be an accidental result of trying to collapse both `match` cases into one right-hand-side.

DrMichaelPetter (approved):
LGTM

sim642 (approved):

PR #1458, opened 2024-05-13
`BaseAnalysis`: For non-definite AD, join over **all** components not just `cpa` in `set`

### LINKED ISSUE #1457: All Non-Relational Privatizations except Miné-W unsound
During #1456 I discovered that, in fact, all of our non-relational privatizations except the one casting Miné's analsysis in our framework are unsound.

~~~C
// PARAM: --set ana.base.privatization protection --enable ana.int.enums
// Like 82-how-could-we-have-been-so-blind.c, but with syntactic globals instead of escaping ones.
#include<pthread.h>
#include<stdlib.h>
struct a {
  int b;
};

int *ptr;
int *immer_da_oane;

int da_oane = 0;
int de_andre = 42;

pthread_mutex_t m;

void doit() {
  pthread_mutex_lock(&m);
  *ptr = 5;

  // Should be either 0 or 5, depending on which one ptr points to
  int fear = *immer_da_oane;
  __goblint_check(fear == 5); //UNKNOWN!

  pthread_mutex_unlock(&m);

  pthread_mutex_lock(&m);
  // This works
  int hope = *immer_da_oane;
  __goblint_check(hope == 5); //UNKNOWN!
  pthread_mutex_unlock(&m);

}

void* k(void *arg) {
  // Force MT
  return NULL;
}

int main() {
    int top;

    if(top) {
      ptr = &da_oane;
    } else {
      ptr = &de_andre;
    }

    immer_da_oane = &da_oane;

    pthread_t t1;
    pthread_create(&t1, 0, k, 0);

    doit();
    return 0;
}
~~~

The problem is that upon setting `*ptr = 5`, all targets in the may-point-to-set are updated, rather than joining together the result of updating the different targets. This is likely due to some bot/top confusion inside the local states of the base analysis where the non-present bindings are all assumed to be \bot.

michael-schwarz:
For non-definite addresses, `set` folds over the possible targets and updates all of them. The `cpa` component of the resulting new state `nst` is then joined with the `cpa` component of the old state `st`, resulting in all possible targets being may-updated.

Unfortunately, the resulting `priv` component is inaccurate then. For example, for the protection-based analysis all targets are treated as must-updated (and added to `P`), resulting in the bug outlined in #1457.

This PR now joins over all components of the local state, i.e., `cpa` and `priv` and additionally `dep` and `weak`. While for `dep` and `weak` no join is necessary to pass the tests, I still think it is the correct thing to do.

Closes #1457.


src/analyses/base.ml
```diff
@@ -1738,7 +1738,7 @@ struct
       (* if M.tracing then M.tracel "set" ~var:firstvar "new state1 %a" CPA.pretty nst; *)
       (* If the address was definite, then we just return it. If the address
        * was ambiguous, we have to join it with the initial state. *)
-      let nst = if AD.cardinal lval > 1 then { nst with cpa = CPA.join st.cpa nst.cpa } else nst in
+      let nst = if AD.cardinal lval > 1 then D.join st nst else nst in
```
sim642, line 1741: For ambiguous points-to sets, the join between old and new (with everything updated sequentially by the preceeding fold) is an optimization over executing each write separately and then joining all those new results (and not the old one).
But is that optimization even correct for (arbitrary) privatizations?

This certainly seems more sound than before, but I'm wondering if it's enough. Maybe we should just ditch the optimization and do the more naive thing that is obviously sound. All the potential writes are still done (either sequentially or in parallel), so this optimization only saves `AD.cardinal lval - 2` joins. Not sure if this is significant enough.

tests/regression/13-privatized/80-treachery-and-lies.c
sim642, line 1: The names of these tests should be meaningful. It will both help us in the future and when we want to submit them to SV-COMP.

src/analyses/base.ml
michael-schwarz: I had similar doubts and so tried to replace the entire line here with:

~~~C
AD.fold (fun addr acc -> D.join (update_one addr st) acc) lval (D.bot ())
~~~

but that lead to some strange test failures where I really have no idea what's going on:

```
4 test(s) failed: ["04/35 trylock_rc", "04/36 trylock_nr", "41/05 more"]
```

If we want to clean this up, we have to investigate what happens with these.

I think for the privatizations it should not be a problem though, as all the may and must sets behave as one would expect.

sim642: That's odd indeed. I'm fine with the current improvement then because it fixes the issue and doesn't break anything else.
The naive implementation failing some tests deserves an issue though. I had a quick tracing look and somehow there seemed to be `CPA` maps with many variables missing or `Uninitialized` (i.e. superbot) values, so something odd happens. Maybe when the address set is somehow empty, the new `D.bot ()` comes into play?

michael-schwarz: Moved to #1465.

sim642 (approved):

Issue #1459, opened 2024-05-14
Affine Equalities Mostly Contain Only Small Portions of Actual Information


DrMichaelPetter:
Setting up a post-analysis count of the share of elements in a matrix that are equal to 0, we come to the following histogram:

![sparseity](https://github.com/goblint/analyzer/assets/9989132/6e71e69e-6afd-4704-b1f8-4af8d32cca30)

- the x-axis is the percentage of 0-elements in relation to all matrix elements of an analysis
- the y-axis is the number of SV-Comp input files of which the average share of 0 entries

Since we explicitly represent those 0 entries in our current affine-equality array, representation a sparse representation would have some potential for runtime and memory consumption savings.

In former legacy implementations for older analyzers, sparse affine equality implementations where quite successful.

**Implementation**

Although there is an interface of abstract vectors and matrices are already present in the code, this was not designed with the idea of sparsely represented data in mind. All in all, this is an initial idea of the complications:

- Especially fold/map style interfaces need to be carefully either implemented with a artificially dense behaviour, or its call-sites need to be checked for compatibility with the sparsity of the representation.
- Liberal conversion to lists/arrays is now costly
- Representation must decide on a row-based or column-based representation; non representation-conforming accesses may be costly and are maybe candidates for caching?

PR #1460, opened 2024-05-14
Making Affine-equalities Sparse

DrMichaelPetter:
This is intended to become a proposal PR for #1459 , exploring if we can come up with performance improvements by switching to sparse representations for affine equalities.


sim642:
Can this be closed in favor of #1625? This only collects the stats anyway without really implementing it.

PR #1468, opened 2024-05-15
Make meet in AddressDomain more precise

### LINKED ISSUE #1467: Meet of string pointers in address domain imprecise
Currently the meet of an unknown string pointer and a known string pointer is the unknown string pointer.

```
{"string"} \sqcap {(unknown string)} = {(unknown string)}
```

The problem seems to be in the address domain, as the actual meet in the string domain which is defined to return the known pointer

https://github.com/goblint/analyzer/blob/bffc5e3cbb1f43bd5c9483a6545802cfba5d3d76/src/cdomain/value/cdomains/stringDomain.ml#L99-L108

is never actually called.

michael-schwarz:
Currently, in the address domain inside the buckets, for the meets we use `ProjectiveSetPairwiseMeet` 

https://github.com/goblint/analyzer/blob/bffc5e3cbb1f43bd5c9483a6545802cfba5d3d76/src/domain/disjointDomain.ml#L190-L205

where `B.may_be_equal` delegates to 

https://github.com/goblint/analyzer/blob/bffc5e3cbb1f43bd5c9483a6545802cfba5d3d76/src/cdomain/value/cdomains/addressDomain.ml#L180

`Addr.semantic_equal` given by

https://github.com/goblint/analyzer/blob/bffc5e3cbb1f43bd5c9483a6545802cfba5d3d76/src/cdomain/value/cdomains/addressDomain.ml#L102-L111

which calls `SD.sematic_equal`

https://github.com/goblint/analyzer/blob/bffc5e3cbb1f43bd5c9483a6545802cfba5d3d76/src/cdomain/value/cdomains/stringDomain.ml#L77-L81

this is needed to handle all sorts of different offsets that may be semantically equal, i.e., evaluate to the same physical address. In this case it keeps both elements.

After some discussion with @jerhard and @hseidl, we reached the following insights:

This behavior is overly conservative, as the mess of the address lattice contains some sub-lattices where the `meet` operation is semantically correct.

This now checks via a new predicate `amenable_to_meet` whether for the two addresses `a` and `b` this relationship between values and the meet holds.  If this is the case, a meet is performed and the element added to the resulting set only if it is distinct from `bot`.

Currently, we answer `amenable_to_meet`by true whenever both arguments are strings, or when the both are addresses that differ in the numeric(!) offsets only.

This PR also adds two tests where this yields additional precision.

Closes #1467


src/domain/disjointDomain.ml
```diff
@@ -182,16 +182,26 @@ module type MayEqualSetDomain =
 sig
   include SetDomain.S
   val may_be_equal: elt -> elt -> bool
+  val amenable_to_meet: elt -> elt -> bool
```
sim642, line 185: I'm wondering if the two predicates could be combined into one more meaningful thing. Because I guess now there's supposed to be some kind of relationship between them, given that we only check them in certain order.

Would this be the correct description?
```ocaml
val same_sublattice: elt -> elt -> bool option
(** @return [Some true] if definitely in same sublattice, [Some false] if definitely not, [None] if unknown. *)
```

src/cdomain/value/cdomains/addressDomain.ml
```diff
@@ -110,6 +110,11 @@ struct
     | StrPtr _, UnknownPtr -> None
     | _, _ -> Some false
 
+  let amenable_to_meet x y = match x,y with
+    | StrPtr _, StrPtr _ -> true
+    | Addr x, Addr y when Mval.equal (Mval.top_indices x) (Mval.top_indices y) -> true
```
sim642, line 115: Don't the address set buckets already guarantee this to only be called in such cases because `Mval.top_indices` are the representatives?

tests/regression/27-inv_invariants/22-meet-ptrs.c
```diff
@@ -0,0 +1,24 @@
...
+
+   int i = 2;
+   if(top) {
+      i = 8;
+   }
+
+   int* imprecise = &arr[i];
+
+   if(imprecise == &arr[2]) {
+      __goblint_check(imprecise == &arr[2]);
```
sim642, line 20: Maybe this didn't work before, but if I understand the improvement correctly, this is a rather pedestrian case of it.
We should now be able to refine `{&a[0..1]} == {&a[1..2]}` into `{&a[1]}` too, right?

michael-schwarz: I added a more intricate example, though refinement currently happens only in one direction, which we should probably investigate separately.

src/domain/disjointDomain.ml
michael-schwarz: I'm skeptical here: `may_be_equal` is a strange semantic property for which there is no real definition in terms of lattices.

So I think using lattice terminology here will be only confusing.  For example what is difference between `Some false` and `None` in your proposed definition above:
- Even if they are definitely not from the same sublattice, they may still be equal or not
- Same if I can't determine whether they are in the same sublattice.

So it once again boils down to the two properties above.

michael-schwarz:
~~~OCaml
  type eqinfo = Maybe | No | Meet
  val may_be_equal: elt -> elt -> eqinfo
~~~

One could do something like this but I don't think that is particularly nice.

src/cdomain/value/cdomains/addressDomain.ml
michael-schwarz: Even if the buckets guarantee this already, this function can be called with two arbitrary addresses, so we can not just return true.

src/domain/disjointDomain.ml
sim642: > I'm skeptical here: `may_be_equal` is a strange semantic property for which there is no real definition in terms of lattices.

I agree and that's why I'm wondering if there is a less ad-hoc way to view these things. `same_sublattice` might not be the right thing, but possibly something else.

There are now two predicates who have 4 possible combined results for a pair of arguments, but only three cases are distinguished where the predicates are used.
Suppose some addresses `x` and `y` had `may_be_equal x y = false` before. Then neither was added to the result.
Now if `amenable_to_meet x y = true`, then `meet x y` is added to the result, which makes the meet of the entire sets larger (i.e. less precise) than before.

This makes me wonder about the order of these checks. If the previous implementation was sound, then checking `may_be_equal` first (to add nothing) would still be sound and cannot become less precise due to `amenable_to_meet`, which could then be checked second to determine whether `meet x y` is added or both `x` and `y`. This would be more precise than the previous implementation by construction.

Or if the case `may_be_equal x y = false` and `amenable_to_meet x y = true` should be impossible, then why? That would indicate there being some implicit connection between the two predicates (a semantic and a lattice-theoretic) after all.

src/cdomain/value/cdomains/addressDomain.ml
sim642: That's true, although it could just be an implementation detail (like the buckets themselves are), which isn't exposed to the outside by the interface. Then such misuse would be impossible.

michael-schwarz:
Sorry for being MIA, I will take this up and hopefully get it mergeable before the holidays.

src/cdomain/value/cdomains/addressDomain.ml
michael-schwarz: I followed this assumption now to just assume that things placed into the same bucket are amenable to meet, which seems to be the case in our implementation and I think also makes sense semantically.

This largely simplifies the implementation and makes all sorts of extra considerations obsolete.

michael-schwarz:
@sim642 anything blocking the merge here? After your comments are integrated, it now is quite lightweight.

src/domain/disjointDomain.ml
```diff
@@ -184,14 +184,24 @@ sig
...
+module ProjectiveSetPairwiseMeet (E: Lattice.S) (B: MayEqualSetDomain with type elt = E.t) (R: Representative with type elt = E.t): SetDomain.S with type elt = E.t = struct
   include ProjectiveSet (E) (B) (R)
 
   let meet m1 m2 =
     let meet_buckets b1 b2 acc =
       B.fold (fun e1 acc ->
           B.fold (fun e2 acc ->
-              if B.may_be_equal e1 e2 then
+              (* If they have the same representative, we use the normal meet within this bucket *)
+              if R.equal (R.of_elt e1) (R.of_elt e2) then
```
sim642, line 195: Could extract `R.of_elt e1` out of the inner `fold`.

src/domain/disjointDomain.ml
```diff
@@ -184,14 +184,24 @@ sig
...
+                try
+                  let m = E.meet e1 e2 in
+                  if not (E.is_bot m) then
+                    add m acc
+                  else
+                    acc
+                with Lattice.Uncomparable ->
+                  failwith (GobPretty.sprintf "Elements %a and %a are in same bucket, but meet throws!" E.pretty e1 E.pretty e2)
```
sim642, line 203: Why rethrow a different exception at all. If that happens then `Uncomparable` could also just propagate to top level and crash. The stacktrace would go through there but also include the inner stacktrace that gets dropped by raising a new exception.
It's truly exceptional (and never supposed to happen), so the user won't be able to do anything with the text anyway.

tests/regression/13-privatized/89-write-lacking-precision.c
```diff
@@ -0,0 +1,30 @@
...
+struct a {
+  char* b;
+};
+
+struct a *c;
+struct a h = {""};
+struct a i = {"string"};
+
+void* d(void* args) {
+  struct a r;
```
sim642, line 12: This seems unused. Is it somehow necessary for the issue to trigger?

tests/regression/13-privatized/89-write-lacking-precision.c
```diff
@@ -0,0 +1,30 @@
...
+  if (c->b) {
+    __goblint_check(strlen(h.b) == 0); // Should also work for write!
```
sim642, line 14: I'm confused about what's happening here. It's checking if `c->b` is a non-NULL byte, so the true branch is taken if `c == &i`. But then the check is on `h` directly, not `c->b`.
The `if` seems completely irrelevant to the check which would also pass without it.

src/domain/disjointDomain.ml
michael-schwarz: This will point immediately at the problem and provide an explanation about what exact invariant is broken. I would consider this much more helpful than a stack trace that just says that some things were incomparable.

It's not a hill I am willing to die on though.

tests/regression/13-privatized/89-write-lacking-precision.c
michael-schwarz: I removed it.

sim642 (approved):

PR #1469, opened 2024-05-16
Separate soundness settings from precision improvements in the autotuner

karoliineh:
I was looking into the SV-COMP tasks and turning the autotuner on and off. The intuition for me was that "when turning on the autotuner, results should get better (more precise), and turning it off _should (only) result in being the same or less precise_". So I was quite surprised to learn that it is also responsible for turning on the analyses for soundness.

Thus, I tried to separate the autotuner functionality that turns on the analyses that ensure soundness from the functionalities that try to improve the precision.


src/autoTuneSvcompSpec.ml
```diff
@@ -0,0 +1,62 @@
+(** Autotuning of the activated analyses for soundness based on SV-COMP specification. *)
```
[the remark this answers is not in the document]
michael-schwarz: I see your point about not crashing when these analyses are not activated.

Still, we should change the name to be different from autotuner, just to make precise what is going on. Otherwise, people might get the impression that we are unsound by default, and then "tuning" things to avoid exposing unsoundnesses on sv-comp.
Probably the `longjmp` autotuner should also be renamed then.

This would leave us with "tuners" trying to optimize performance/ precision tradeoffs, and this other mechanism which ensures soundness.

src/autoSoundConfig.ml
```diff
@@ -0,0 +1,86 @@
...
+    Activating the analyses and enabling options that are
+    required for analyzing the property defined in the specification.
+    TODO: have only one function for matching all specifications and find a place where it can be called.
+*)
+let enableAnalysesForMemSafetySpecification (spec: Svcomp.Specification.t) =
+  match spec with
+  | ValidFree -> (* Enable the soundness analyses for ValidFree spec *)
+    let analyses = ["base"; "useAfterFree"] in
+    logEnablingAnalyses spec analyses;
+    enableAnalyses analyses
```
sim642, line 31: We should have some helper function to do both the logging and the enabling to reduce code duplication. For example, to write it like this:
```suggestion
    enableSpecAnalyses spec ["base"; "useAfterFree"]
```

src/autoSoundConfig.ml
sim642, line 1: Maybe move this to `src/util/`. `AutoTune` is one weird module that's outside of all subdirectories for no reason.

sim642 (approved):

PR #1470, opened 2024-05-16
Add ARG tests

sim642:
Like in #1462, I'm always afraid of doing any modifications in the ARG generation because it's very fragile and there have been no tests so far.

At some point I added https://github.com/goblint/analyzer/blob/bffc5e3cbb1f43bd5c9483a6545802cfba5d3d76/scripts/sv-comp/witness-isomorphism.py to semi-automatically check if two sets of GraphML witnesses are isomorphic. But that requires a set of reference witnesses (which we don't have in the repository) and the networkx Python library that implements graph isomorphism algorithms.

This PR adds an alternative means of fully automatic tests for ARGs. This is similar to the CFG tests: a GraphViz graph is rendered as ASCII using `graph-easy` and expected outputs are in the repository.
The tests are very similar to Cram tests: if something changes, the tests will fail with a diff that can be promoted if intended. However, they aren't Cram tests but just dune expectation tests using [rule generation](https://dune.readthedocs.io/en/stable/howto/rule-generation.html).

The mechanism is somewhat complex (at least without dune 3.14), but nobody other than me will probably ever have to deal with these anyway (because they should just keep passing).

### TODO
- [ ] Manually check that all current outputs are correct to begin with.


src/config/options.schema.json
```diff
@@ -1826,6 +1826,13 @@
           "description": "Output ARG as dot file.",
           "type": "boolean",
           "default": false
+        },
+        "argdotlabel": {
```
michael-schwarz, line 1830: Since we now have three ARG related options, should we maybe group them under `exp.arg.*`?

tests/sv-comp/gen/gen.ml
```diff
@@ -0,0 +1,33 @@
...
+   (sandbox always) ; must sandbox to prevent arg.dot-s from parallel runs from interfering
+   (package goblint) ; depend on entire goblint package for svcomp21 conf
+   (:c %s)
+   (:prop %%{project_root}/tests/sv-comp/unreach-call-__VERIFIER_error.prp))
+  (target %s.output)
+  (enabled_if %%{bin-available:graph-easy})
+  (action
+   (progn
+    (ignore-outputs
+     (run goblint --conf svcomp21.json --set ana.specification %%{prop} %%{c} --enable witness.graphml.uncil --enable exp.argdot --enable exp.arg --set exp.argdotlabel empty))
```
michael-schwarz, line 18: Could you elaborate why you use the `svcomp21` config for these? Would it not make sense to use a more recent config (maybe the most recent?) in order to catch bugs that only surface with new analyses / our new config?

tests/sv-comp/dune.inc
```diff
@@ -0,0 +1,902 @@
+
```
michael-schwarz, line 1: As far as I understand, this is auto-generated? Can you maybe in the generator print a comment into here to tell people not to manually do things here?
While people are usually familiar with `autoconf` and friends, this generation mechanism probably is new for most.

michael-schwarz (commented):
I looked it over and it seems to make sense at first glance. Maybe it would be a good idea for someone else who is familiar with the ARGs (maybe @karoliineh?) to also took a look at this since I'm out of my depth here?

tests/sv-comp/gen/gen.ml
sim642: I just used `svcomp21` because that's what the old witness generation scripts still reference and I had old witnesses for. I now changed it to `svcomp` (without autotuning to keep things somewhat under control).
This does mean that as we change our SV-COMP setup, these ARG tests might fail only due to precision changes. But they're rather simple, so this should rarely happen.

Issue #1456, opened 2024-05-12
Protection-Based Privatization more Precise than Write-Centered


michael-schwarz:
Starting from minimizing one of the larger programs, I have now arrived at this example program, where the Protection-Based Privatization is more precise than the Write-Centered (without intervals or other domains that require a widening).

~~~C
#include<pthread.h>
#include<stdlib.h>
struct a {
  struct a *b;
  struct a *c;
};

struct a i;

pthread_mutex_t m;

void doit() {
  void *newa = malloc(sizeof(struct a));

  pthread_mutex_lock(&m);
  i.c->b = newa;
  i.c = newa;
  pthread_mutex_unlock(&m);
}

void* k(void *arg) {
  doit();
  return NULL;
}

int main() {
    struct a other;
    i.c = &other;

    pthread_t t1;
    pthread_create(&t1, 0, k, 0);

    doit();
    return 0;
}
~~~

The program compiles with GCC and runs without segfaults.

Script to run comparison: `./goblint --conf conf/traces.json --set ana.base.privatization protection 2.c --enable allglobs --enable dbg.timing.enabled --enable warn.debug -v --sets exp.priv-prec-dump level-ip.protection.prec && ./goblint --conf conf/traces.json --set ana.base.privatization write 2.c --enable allglobs --enable dbg.timing.enabled --enable warn.debug -v --sets exp.priv-prec-dump level-ip.write.prec &&  ./privPrecCompare level-ip.protection.prec level-ip.write.prec &> out.txt`

The diff output is: 
```
[Debug] 2.c:17:3-17:13 (alloc@sid:10@tid:[main]): protection more precise than write
  protection: {
                b -> {&(alloc@sid:10@tid:[main])}
                c -> Uninitialized
              }
  more precise than
  write: (value:{
                  b -> {&(alloc@sid:10@tid:[main]), &(alloc@sid:10@tid:[main, k@2.c:31:5-31:33])}
                  c -> Uninitialized
                }, size:(16,[16,16],{16}), no zeroinit:true)
  reverse diff: compound: (value:{
                                   b -> {&(alloc@sid:10@tid:[main]), &(alloc@sid:10@tid:[main, k@2.c:31:5-31:33])}
                                   c -> Uninitialized
                                 }, size:(16,[16,16],{16}), no zeroinit:true) not same type as {
                                                                                                   b -> {&(alloc@sid:10@tid:[main])}
                                                                                                   c -> Uninitialized
                                                                                                 }

[...]

PR #1471, opened 2024-05-17
Union: `meet` in `Field`: Do not raise `Uncomparable` when one arg is `Top`

michael-schwarz:
This behavior leads to unnecessary imprecision in combination with the invariant refinement for privatizations at branches.

This was added  in https://github.com/goblint/analyzer/pull/1109 but it seems we were a bit overzealous in raising this exception even if one of the arguments is `Top`.

This fixes some of the issues described in #1456.

### DIFF
--- a/src/cdomain/value/cdomains/unionDomain.ml
+++ b/src/cdomain/value/cdomains/unionDomain.ml
@@ -26,7 +26,10 @@ module Field = struct
     if equal f g then
       f
     else
-      raise Lattice.Uncomparable
+      match f, g with
+      | `Top, f
+      | f, `Top -> f
+      | _ -> raise Lattice.Uncomparable
 end
 
 module Simple (Values: Arg) =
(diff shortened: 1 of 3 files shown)


src/cdomain/value/cdomains/unionDomain.ml
```diff
@@ -26,7 +26,10 @@ module Field = struct
     if equal f g then
       f
     else
-      raise Lattice.Uncomparable
+      match f, g with
+      | `Top, f
+      | f, `Top -> f
+      | _ -> raise Lattice.Uncomparable
```
sim642, line 32: This essentially duplicates the included `FlatConf` now. Maybe we could just reuse that implementation and turn tops into exceptions?
```ocaml
match meet f g with
| `Top -> raise Lattice.Uncomparable
| m -> m
```

michael-schwarz: It's `Bot` that would have to be turned into an exception here.

sim642: Sorry, yes. But with that flipped, the logic should more closely describe the intent.

sim642 (approved):

Issue #1473, opened 2024-05-20
Interval sets are unsound due to larger-than-type ranges


sim642:
In `c/pthread/fib_safe-5.i` from #1356, we produce an unsound invariant. Opening the HTML view to see globals reveals that they somehow have all integer domains enabled (#1472):
> base:priv:unprotected:cur → def_exc:Unknown int([-31,31]); intervals:[0,2147483647]; enums:not {} ([-31,31]); congruences:ℤ; interval_sets:[[-4294967296, 4294967296]]

With some additional tracing and CIL warnings, it becomes clear that interval sets are at fault:
```
%%% priv: (:-1)  invariant_global unprotected:cur
  %%% priv: getg unprotected:cur -> (Unknown int([-31,31]),[0,2147483647],not {} ([-31,31]),ℤ,[[-4294967296, 4294967296]])
  %%% priv: interval invariant [-4294967296,4294967296]
lib/libc/stub/src/stdlib.c:38: Warning: Truncating integer 4294967296 to 0
  %%% priv: -> exp:0 <= cur && cur <= 0
  %%% priv: interval invariant [0,2147483647]
  %%% priv: -> exp:0 <= cur
  %%% priv: -> exp:0 <= cur && (0 <= cur && cur <= 0)
```
Namely, interval sets somehow have an interval which is larger than the corresponding type. This itself breaks many internal assumptions.

The internal inconsistency is only revealed via witnesses, where conversion of `Z.t` into `Cil.exp` goes through some CIL functions that truncate to type. Because the interval is bigger than the type, it gets bit-truncated to 0 (because all 32 bits are zero).

karoliineh:
Debugging this revealed that the autotuning option `octagon` was manually switched off in the command used to generate the incorrect witness. By removing `--set 'ana.autotune.activated[-]' octagon`, the interval seems to be correctly calculated (`interval_sets:[[-2147483648, 2147483647]]`).

In the `svcomp-ghost.json`, the `ana.autotune.wideningThresholds` is still activated, making autotune set option `ana.int.interval_threshold_widening_constants` to `comparisons`, which is then used in `intDomain.IntervalSetFunctior.widen`:

https://github.com/goblint/analyzer/blob/ede09b88b8c0b16ab68045fe8e29e1dc8dddb474/src/cdomain/value/cdomains/intDomain.ml#L1397-L1401

where the list `ts` gets the value `[-281474976710656
[...]

PR #1476, opened 2024-05-21
Fix unsoundness bug in interval sets and refactor interval widen thresholds

karoliineh:
* Makes sure that if the next widen threshold is bigger than `max_ik` (or smaller than `man_ik`), `max_ik` (or `min_ik`) is picked instead. (solves #1473)
* Refactors the duplicate code of `widen.upper_threshold` and `widen.lower_threshold` of `IntervalFunctor` and `IntervalSetFunctor` to `IntervalArith` module, and uses this function for both.

### DIFF
--- a/src/cdomain/value/cdomains/intDomain.ml
+++ b/src/cdomain/value/cdomains/intDomain.ml
@@ -563,12 +563,14 @@ module IntervalArith (Ints_t : IntOps.IntOps) = struct
   let upper_threshold u max_ik =
     let ts = if get_interval_threshold_widening_constants () = "comparisons" then WideningThresholds.upper_thresholds () else ResettableLazy.force widening_thresholds in
     let u = Ints_t.to_bigint u in
-    let t = List.find_opt (fun x -> Z.compare u x <= 0 && Z.compare x (Ints_t.to_bigint max_ik) <= 0) ts in
+    let max_ik' = Ints_t.to_bigint max_ik in
+    let t = List.find_opt (fun x -> Z.compare u x <= 0 && Z.compare x max_ik' <= 0) ts in
     BatOption.map_default Ints_t.of_bigint max_ik t
   let lower_threshold l min_ik =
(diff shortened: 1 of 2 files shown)


src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -668,21 +679,9 @@ struct
...
-        let l = Ints_t.to_bigint l in
-        let t = List.find_opt (fun x -> Z.compare l x >= 0) ts in
-        BatOption.map_default Ints_t.of_bigint min_ik t
-      in
-      let lt = if threshold then lower_threshold l1 else min_ik in
+      let lt = if threshold then IArith.lower_threshold l1 min_ik else min_ik in
       let l2 = if Ints_t.compare l0 l1 = 0 then l0 else Ints_t.min l1 (Ints_t.max lt min_ik) in
-      let ut = if threshold then upper_threshold u1 else max_ik in
+      let ut = if threshold then IArith.upper_threshold u1 max_ik else max_ik in
       let u2 = if Ints_t.compare u0 u1 = 0 then u0 else Ints_t.max u1 (Ints_t.min ut max_ik) in
```
sim642, line 685: The reason non-set intervals didn't have this problem is because if the `Ints_t.min ut max_ik` here to make sure that if `ut` was chosen to be bigger than the type range, then `max_ik` is used instead. Interval sets probably didn't have such check.
But now it should be unnecessary because `IArith.upper_threshold` should ensure that it doesn't return anything bigger.

I'm also confused by the `Ints_t.max u1` here. It shouldn't be possible for the threshold/`max_ik` to be smaller than `u1` to begin with!?

On a different note, currently `ut` is computed even if `u2` doesn't need it, which is wasteful. We could just move `ut` into the else branch to avoid that.

sim642:
It would be nice to have tests for this. Instead of trying to construct programs with complex autotuning configurations, adding unit tests into `intDomainTest.ml` would be a nice way.

sim642 (approved):

michael-schwarz (approved):

Issue #1421, opened 2024-04-18
Crash on `pthread_mutex_lock` with complicated argument


michael-schwarz:
For the following program (extracted from the `axel` benchmark of Concrat), Goblint crashes with 

> Fatal error: exception Cilfacade.TypeOfError(typeOffset: Index on a non-array (0, struct a ))

~~~C
struct a {
  pthread_mutex_t b;
};
struct c {
  struct a *conn;
} d();

int main() {
  struct a str = {0};
  struct c axel = {0};
  axel.conn = &str;
  pthread_mutex_t* ptr = &((axel.conn + 0)->b);
  pthread_mutex_lock(ptr);
  pthread_mutex_unlock(ptr);
  pthread_mutex_lock(&((axel.conn + 0)->b));
}
~~~

whereas `gcc -Wall` accepts the program without any issues.

sim642:
The easy workaround would be to just allow constant 0 index offset on a struct as a no-op.
However, I think we might have a deeper issue here: `axel.conn + 0` is a perfectly valid addition of a constant to a pointer (to a struct). Somehow we forget about the pointer being there.

michael-schwarz:
> Somehow we forget about the pointer being there.

What do you mean by this? The points-to set for `ptr` is `{ str[0].b }` which seems ok!

sim642:
> The points-to set for `ptr` is `{ str[0].b }` which seems ok!

But it isn't OK because it's indexing a struct. That indexing should not be there at all, because really it just points to `str.b`.

The indexing would be valid if there was a pointer to the first element of an array:
```c
struct a {
  pthread_mutex_t b;
};
struct c {
  struct a *conn;
} d();

int main() {
  struct a str[1] = {0};
  struct c axel = {0};
  axel.conn = &str;
  pthread_mutex_t* ptr = &((axel.conn + 0)->b);
  pthread_mutex_lock(ptr);
  pthread_mutex_unlock(ptr);
  pthread_mutex_lock(&((axel.conn + 0)->b));
}
Here `str[0].b` would be right.
```

Somewhere in the pointer arithmetic we're going wrong and assuming an array where there isn't one. This probably stems from the fact that arrays are the same as their first element pointer in such code.

michael-schwarz:
In 811c183, I made a hotfix to be able to continue benchmarking that catches this error and errors and returns a top of the corresponding type.
Do we also want to have this hotfix on master?

karoliineh:
[...]

PR #1480, opened 2024-05-23
Handle operations on the addresses of non-array types

karoliineh:
Solves #1421

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -250,6 +250,7 @@ struct
      * the integer domain operations. *)
     let bool_top ik = ID.(join (of_int ik Z.zero) (of_int ik Z.one)) in
     (* An auxiliary function for ptr arithmetic on array values. *)
+    let exception UnknownPtr in
     let addToAddr n (addr:Addr.t) =
       let typeOffsetOpt o t =
         try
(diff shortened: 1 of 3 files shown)


src/analyses/base.ml
```diff
@@ -267,31 +262,32 @@ struct
...
-                | Some true -> `NoOffset
-                | _ -> `Field (f, `Index (n_offset, `NoOffset))
-              end
-            | None -> `Field (f, `Index (n_offset, `NoOffset))
+          let (f_offset_bits, _) = bitsOffset t (Field (f, NoOffset)) in
+          let f_offset = IdxDom.of_int (Cilfacade.ptrdiff_ikind ()) (Z.of_int (f_offset_bits / 8)) in
+          begin match IdxDom.(to_bool (eq f_offset (neg (iDtoIdx n)))) with
+            | Some true -> `NoOffset
+            | _ when isArrayType f.ftype -> `Field (f, `Index (n, `NoOffset))
+            | _ when GobOption.exists (Z.equal Z.zero) (ID.to_int n) -> `NoOffset (* adding (or subtracting) 0 to any type of pointer is ok *)
```
karoliineh, line 270: The two cases on lines 268-269 might actually be dead code?
And maybe the proper way to fix the issue would've been just to add a similar hack from line 267 to the case of `NoOffset`?

michael-schwarz, line 270: As it stands this means that the case of `n` being `-f_offset` is handled the same as the case where `n` is `0`?
That seems odd.

Would this not have to be `Field(f,NoOffset)` here?

PR #1485, opened 2024-05-24
[WIP] Thesis Weakly-Relational Pointer Analysis

reb-ddm:
This draft PR introduces a weakly-relational analysis between pointers (ref: [2-pointer Logic by Seidl et al.](https://link.springer.com/chapter/10.1007/978-3-031-56222-8_16])).
The analysis can infer equalities and disequalities between terms which are built from pointer variables, with the addition of constants and dereferencing.

Abstract states are represented by a congruence closure of terms over the uninterpreted function `*`.
Additionaly, a list of disequalities between terms is stored.
Each term stores the information about the CIL expression that was used to create the term.
This way, it is easier to reconvert a term to a CIL expression and to get information about the type of a term.

In order to calculate a normal form of an abstract state, we keep a map of the minimal representative term for each equivalence class in the congruence closure.

The disequalities are important during the assignment operation, because when we assign a value to an address on the heap, we need to forget the information we had about all terms thay may alias with this address.
Additionally to the disequalities inferred by this analysis, we exploit the May-Point-To analysis in order to derive additional disequalities between terms.

The implementation of "enter" duplicates the parameter variables of each function in order to infer in "combine" which terms are related to the initial parameters.
In order to use the May-Point-To analysis also on these duplicated variables, an additional analysis `startState` is added.
It remembers the May-Point-To information of the duplicated variables at the beginning of the function.


michael-schwarz:
As almost all files are new here, it should be easy to merge master and add `startcontext` (c.f. https://github.com/goblint/analyzer/pull/1427), so we can see if the CI passes then.

src/analyses/weaklyRelationalPointerAnalysis.ml
```diff
@@ -0,0 +1,186 @@
...
+          let valid_term (t,z) =
+            T.is_ptr_type (T.type_of_term t) && (T.get_var t).vid > 0 in
+          let equal_terms = List.filter valid_term comp in
+          let intersect_query_result res (term,z) =
+            let next_query = ask.f (MayPointTo (T.to_cil_sum z (T.to_cil term))) in
+            MayBeEqual.AD.meet res next_query in
+          List.fold intersect_query_result (MayBeEqual.AD.top()) equal_terms
+      end
+    | _ ->
+      MayBeEqual.AD.top()
```
michael-schwarz, line 54: Could we as alternative indicator of where the domain is helpful somehow measure how often this leads to an improvement?
We would have to somehow measure this at the top level though...

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1581 @@
...
+    | TComp _ -> true
+    | _ -> false
+
+  let is_struct_ptr_type = function
+    | TPtr(TComp _,_) -> true
+    | _ -> false
+
+  let is_ptr_type = function
+    | TPtr _ -> true
+    | _ -> false
```
michael-schwarz, line 143: Before calling these, you might want to unroll first to get rid of any `TNamed` (typedefs) that may surround them

src/analyses/weaklyRelationalPointerAnalysis.ml
```diff
@@ -0,0 +1,197 @@
...
+      f.sformals @ f.slocals @ List.map duplicated_variable f.sformals @ reachable_from_args ctx args
+    in
+    let new_state = D.remove_terms_not_containing_variables reachable_variables state_with_duplicated_vars in
+    if M.tracing then M.trace "wrpointer-function" "ENTER2: result: %s\n" (D.show new_state);
+    [ctx.local, new_state]
+
+  (*ctx caller, t callee, ask callee, t_context_opt context vom callee -> C.t
+     expr funktionsaufruf*)
+  let combine_env ctx var_opt expr f exprs t_context_opt t (ask: Queries.ask) =
+    ctx.local
```
michael-schwarz, line 161: This seems wrong, when combining env, all assigning etc should be done except for the assignment of the lvalue of the function call.
This is somewhat arbitrary, but should be done this way around, as the other parts of the framework assume this behavior.

src/analyses/weaklyRelationalPointerAnalysis.ml
```diff
@@ -0,0 +1,197 @@
...
+  let special ctx var_opt v exprs  =
+    let desc = LibraryFunctions.find v in
+    match desc.special exprs, v.vname with
+    | Assert { exp; refine; _ }, _ -> if not refine then
+        ctx.local
+      else
+        branch ctx exp true
+    | _, _ -> ctx.local
+
+  let duplicated_variable var = { var with vid = - var.vid - 4; vname = "wrpointer__" ^ var.vname ^ "'" }
```
michael-schwarz, line 138: This is fine for now, but we have more sophisticated facilities for maintaining copies of variables for such purposes (see the privatizations in apron).

src/analyses/weaklyRelationalPointerAnalysis.ml
```diff
@@ -0,0 +1,197 @@
...
+  let original_variable var = { var with vid = - (var.vid + 4); vname = String.lchop ~n:11 @@ String.rchop var.vname }
+
+  (*First all local variables of the function are duplicated (by negating their ID),
+    then we remember the value of each local variable at the beginning of the function
+    by using the analysis startState. This way we can infer the relations between the
+    local variables of the caller and the pointers that were modified by the function. *)
+  let enter ctx var_opt f args =
+    (* add duplicated variables, and set them equal to the original variables *)
```
michael-schwarz, line 146: Interesting! Is this some sort of anchoring?

src/analyses/weaklyRelationalPointerAnalysis.ml
```diff
@@ -0,0 +1,197 @@
...
+
+  let startstate v = D.top ()
+  let threadenter ctx ~multiple lval f args = [D.top ()]
+  let threadspawn ctx ~multiple lval f args fctx = D.top()
+  let exitstate  v = D.top ()
+
+end
+
+let _ =
+  MCP.register_analysis ~dep:["startState"; "taintPartialContexts"] (module Spec : MCPSpec)
```
michael-schwarz, line 197: Is `taintPartialContexts` actually used here?

src/analyses/weaklyRelationalPointerAnalysis.ml
```diff
@@ -0,0 +1,197 @@
...
+    if M.tracing then M.trace "wrpointer-function" "ENTER2: result: %s\n" (D.show new_state);
+    [ctx.local, new_state]
+
+  (*ctx caller, t callee, ask callee, t_context_opt context vom callee -> C.t
+     expr funktionsaufruf*)
+  let combine_env ctx var_opt expr f exprs t_context_opt t (ask: Queries.ask) =
+    ctx.local
+
+  (*ctx.local is after combine_env, t callee*)
+  let combine_assign ctx var_opt expr f args t_context_opt t (ask: Queries.ask) =
```
michael-schwarz, line 164: Can the assignment of a return lval (if there is one) not be expressed in terms of a normal assignment here? That would save some custom code...

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1601 @@
...
+
+  let bitsSizeOfPtr () = Z.of_int @@ bitsSizeOf (TPtr (TVoid [],[]))
+
+  (* equality of terms should not depend on the expression *)
+  let compare_exp _ _ = 0
+  let equal_exp _ _ = true
+  let hash_exp _ = 1
+
+
+  (* we store the varinfo and the Cil expression corresponding to thi term in the data type *)
```
michael-schwarz, line 29:
```suggestion
  (* we store the varinfo and the Cil expression corresponding to the term in the data type *)
```

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1601 @@
...
+  let equal_v_prop p1 p2 =
+    let equivalent_triple (t1,t2,o1) (t3,t4,o2) =
+      (equal t1 t3 && equal t2 t4 && Z.equal o1 o2) ||
+      (equal t1 t4 && equal t2 t3 && Z.(equal o1 (-o2)))
+    in match p1, p2 with
+    | Equal (a,b,c), Equal (a',b',c') -> equivalent_triple (a,b,c) (a',b',c')
+    | Nequal (a,b,c), Nequal (a',b',c') -> equivalent_triple (a,b,c) (a',b',c')
+    | _ -> false
+
+  let compare_v_prop p1 p2 =
```
michael-schwarz, line 52: Does this compare respect the rules for a compare?
Consider p1 and p2 that compare equal according to the `equivalent _triple` rule above, and some p' that compares smaller than p1.
Is it guaranteed to also compare smaller than p2?

If not, many data structures that rely on compare providing an order will silently provide incorrect results or have horrible run-time complexity.

src/analyses/weaklyRelationalPointerAnalysis.ml
reb-ddm: I use it in `combine_env` for the query `MayBeTainted`

src/analyses/weaklyRelationalPointerAnalysis.ml
reb-ddm: What do you mean exactly? The normal assignment just calls the function `assign_lval`, and in `combine_assign` I use the same function for the assignment of a return lval

src/cdomains/c2poDomain.ml
```diff
@@ -1281,13 +1281,16 @@ module D = struct
...
         | Some cc1, Some cc2 ->
-          (* add all terms to both elements *)
-          let terms = SSet.union (SSet.union cc1.set (BlDis.term_set cc1.bldis))
-              (SSet.union cc2.set (BlDis.term_set cc2.bldis)) in
-          let cc1, cc2 = Option.get (insert_set (Some cc1) terms), Option.get (insert_set (Some cc2) terms) in
-          equal_eq_classes cc1 cc2
-          && equal_diseqs cc1 cc2
-          && equal_bldis cc1 cc2
+          if cc1 == cc2 then
+            true
```
michael-schwarz, line 1198: You're right this is probably the better place for short-circuiting, as the wrappers outside may be different cells.

However, we can then remove the shortcut in 1277, pattern matching on an option is cheap!

src/cdomains/c2poDomain.ml
```diff
@@ -0,0 +1,1337 @@
...
+
+  let empty () = Some {uf = TUF.empty; set = SSet.empty; map = LMap.empty; diseq = Disequalities.empty; bldis = BlDis.empty}
+
+  let init () = init_congruence []
+
+  let bot () = None
+  let is_bot x = Option.is_none x
+  let top () = empty ()
+  let is_top = function None -> false
+                      | Some cc -> TUF.is_empty cc.uf
```
michael-schwarz, line 1218: Is this correct? Can I not have, e.g., a (block) disequality but still an empty TUF?

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1318 @@
...
+  (** Computes the initial map of minimal representatives.
+        It maps each element `e` in the set to `(e, 0)`. *)
+  let initial_minimal_representatives set =
+    List.fold_left (fun map element -> add element (element, Z.zero) map) empty (SSet.elements set)
+end
+
+module Lazy =
+struct
+  include Lazy
+  let hash x y = 0
```
sim642, line 535: `hash` with two arguments?! This seems unused.

src/analyses/singleThreadedLifter.ml
```diff
@@ -0,0 +1,59 @@
+(** This lifter transforms any analysis into a single threaded analysis by returning top when the code might be multi-threaded.
```
sim642, line 1: This sounds backwards to me: it takes an analysis which only works/is sound for single-threaded code and allows it to run on multi-threaded programs (by being top).
Or is the purpose to actually take a fine multi-threaded analysis and make it less precise?

src/config/options.schema.json
```diff
@@ -1144,6 +1144,31 @@
...
+              "title": "ana.c2po.askbase",
+              "description": "If true, the C-2PO Analysis uses the MayPointTo query to infer additional disequalities.",
+              "type": "boolean",
+              "default": true
+            },
+            "precise_join": {
+              "title": "ana.c2po.precise_join",
+              "description": "If true, the C-2PO Analysis uses a more precise version of the join algorithm, by using the automaton to compute the join.",
+              "type": "boolean",
+              "default": false
```
sim642, line 1161: If it's a choice between multiple algorithms, this maybe an enum option would make more sense, e.g. `join-algorithm` with two values. Then the description could mention the time-precision tradeoff or whatever.
Right now, it's like: why wouldn't I want precise?

src/config/options.schema.json
```diff
@@ -1144,6 +1144,31 @@
...
+            },
+            "normal_form": {
+              "title": "ana.c2po.normal_form",
+              "description": "If true, the C-2PO Analysis computes a normal form of the domain, using minimal representatives of the equivalence classes, in order to check for equailty between two domain elements. If false, it compares the equivalence classes of the two domain elements.",
+              "type": "boolean",
+              "default": false
```
sim642, line 1167: But what does this mean to a user? Is it a tradeoff between precision and efficiency? If so, which is better for which?

src/goblint_lib.ml
```diff
@@ -92,6 +92,7 @@ module Malloc_null = Malloc_null
 module MemLeak = MemLeak
 module UseAfterFree = UseAfterFree
 module MemOutOfBounds = MemOutOfBounds
+module C2poAnalysis = C2poAnalysis
```
sim642, line 95: This is better suited for the "Value" subsection where all other relational analyses are already. This analysis also handles pointers to locals/globals, no?

src/goblint_lib.ml
```diff
@@ -272,6 +275,10 @@ module MusteqDomain = MusteqDomain
 module RegionDomain = RegionDomain
 module StackDomain = StackDomain
 
+module CongruenceClosure = CongruenceClosure
+module UnionFind = UnionFind
+module C2poDomain = C2poDomain
```
sim642, line 280: These might deserve their own subsection like relational.

tests/regression/83-c2po/01-simple.c
sim642, line 1: These tests seem to be full of obscure pointer aliasing things. Do we plan to submit these to sv-benchmarks as well? Or is there undefined behavior in some?

It would be very interesting to know how other tools with their approaches cope with these. Or maybe the tests are quite deterministic/small that simple execution/model checking can just decide these?

sim642 (commented):
I haven't read the paper, so I cannot comment on the approach itself, but I looked over some of the code.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,743 @@
...
+  let bitsSizeOfPtr () = Z.of_int @@ bitsSizeOf (TPtr (TVoid [],[]))
+
+  (* equality of terms should not depend on the expression *)
+  let compare_exp _ _ = 0
+  let equal_exp _ _ = true
+  let hash_exp _ = 1
+
+  (* we store the varinfo and the Cil expression corresponding to the term in the data type *)
+  type t = (Var.t, exp[@compare.ignore][@eq.ignore][@hash.ignore]) term [@@deriving eq, ord, hash]
+  type v_prop = (Var.t, exp[@hash.ignore]) prop [@@deriving hash]
```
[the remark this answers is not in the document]
reb-ddm: Unfortunately the conversion back to expression is a bit complicated, because I convert arrays like `a[3]` to pointer terms `*(a+3)`. This way I can also express properties about arrays. And I do the same with structs, I calculate the offset of structs by using the Evalnt query on the cil offset. Also, there might be casts involved in the expressions, so in practice it's not always straightforward how to reconvert a term back to the cil expression. 
Given that we only analyze terms that are pointers or 64-bit integers, in my opinion it's not a problem to merge equivalent values with different expressions. For example `(int*)pointer1` and `(long*)pointer1` still have the same value, even if they don't have the same type. 
The difference in the analysis is that the dereferenced term of `(int*)pointer1` is ignored, as it is not 64-bits. While if we dereference `(long*)pointer1` it will be added to the domain.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,743 @@
...
+  let z_of_offset ask offs typ =
+    match IntDomain.IntDomTuple.to_int @@ cil_offs_to_idx ask offs typ with
+    | Some i -> i
+    | None
+    | exception (SizeOfError _) -> if M.tracing then M.trace "c2po-invalidate" "REASON: unknown offset";
+      raise (UnsupportedCilExpression "unknown offset")
+
+  let rec can_be_dereferenced = function
+    | TNamed (typinfo, _) -> can_be_dereferenced typinfo.ttype
+    | TPtr _| TArray _| TComp _ -> true
```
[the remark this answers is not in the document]
reb-ddm: I meant that the term representation of this type can be dereferenced. Arrays and structs are interpreted as pointers in this representation.

src/cdomains/duplicateVars.ml
```diff
@@ -0,0 +1,59 @@
...
+  (* the hash/compare values should not depend on the type.
+     But they have to be defined even though they are not used, for some reason.*)
+  let equal_typ _ _ = true
+  let hash_typ _ = 0
+  let compare_typ _ _ = 0
+
+  type t = AssignAux of (typ[@compare.ignore][@eq.ignore][@hash.ignore])
+         | ReturnAux of (typ[@compare.ignore][@eq.ignore][@hash.ignore])
+         | NormalVar of Varinfo.t
+         | DuplicVar of Varinfo.t [@@deriving eq,ord,hash]
```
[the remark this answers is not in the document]
reb-ddm: I thought that `RichVarinfo.BiVarinfoMap.Make` only needs the hash/compare functions for `t` and not for `typ`?

src/cdomains/c2poDomain.ml
```diff
@@ -0,0 +1,224 @@
...
+        if exactly_equal a b then
+          a'
+        else
+          match get_conjunction a with
+          | [] -> b'
+          | a_conj -> reset_normal_form (meet_conjs_opt a_conj b')
+    in if M.tracing then M.trace "c2po-meet" "MEET RESULT = %s" (show res);
+    res
+
+  let leq x y = equal (meet x y) x
```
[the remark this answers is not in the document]
reb-ddm: To directly decide this, we would need to look at all equivalence classes, which is basically the same as the algorithm for `equal`.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,743 @@
...
+open GoblintCil
+open DuplicateVars
+module M = Messages
+
+exception Unsat
+
+type ('v, 't) term = Addr of 'v | Aux of 'v * 't | Deref of ('v, 't) term * Z.t * 't [@@deriving eq, ord, hash]
+type ('v, 't) prop = Equal of ('v, 't) term * ('v, 't) term * Z.t | Nequal of ('v, 't) term * ('v, 't) term * Z.t
+                   | BlNequal of ('v, 't) term * ('v, 't) term
+[@@deriving eq, ord, hash]
```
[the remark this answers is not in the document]
reb-ddm: Not really, I can remove it

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,743 @@
...
+
+  let compare t1 t2 =
+    match t1,t2 with
+    | Addr v1, Addr v2
+    | Aux (v1,_), Aux (v2,_) -> Var.compare v1 v2
+    | Deref (t1,z1,_), Deref (t2,z2,_) -> let c = compare t1 t2 in
+      if c = 0 then Z.compare z1 z2 else c
+    | Addr _, _
+    | _, Deref _ -> -1
+    | _ -> 1
```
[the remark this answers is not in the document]
reb-ddm: Yes I forgot to remove it

src/cdomains/c2poDomain.ml
```diff
@@ -0,0 +1,224 @@
...
+  type t = domain [@@deriving ord, hash]
+
+  (** Convert to string *)
+  let show x = match x with
+    | None -> "⊥\n"
+    | Some x -> show_conj (get_conjunction x)
+
+  let show_all = function
+    | None -> "⊥\n"
+    | Some x -> show_all x
```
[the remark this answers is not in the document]
reb-ddm: Is it possible to return `Bot` in the `meet` function if I use `Lattice.LiftBot`?

src/cdomains/congruenceClosure.ml
reb-ddm: It's the same problem as before, where I define a type (in the following lines) that contains a `Lazy.t` type, and I say that it should ignore this lazy type for deriving compare, eq and hash. But it still expects a hash function (for some reason with two arguments), otherwise it doesn't compile.

src/cdomains/duplicateVars.ml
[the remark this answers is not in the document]
reb-ddm: Also, in my abstraction, there is only a single "AssignAux" variable, which can have different types for each program point, but there never exist two different "AssignAux" variables at the same time. Therefore, all "AssignAux" should be considered as the same variable, even when they have different types. The same holds for "ReturnAux".

tests/regression/83-c2po/01-simple.c
michael-schwarz: Pulled out into #1571.

michael-schwarz:
Any objections to merging @jerhard @sim642?

michael-schwarz (approved):

src/cdomains/unionFind.ml
```diff
@@ -10,10 +10,35 @@ module M = Messages
...
+let tuple3_equal p1 p2 = Tuple3.eq equal_term equal_term Z.equal (normal_form_tuple_3 p1) (normal_form_tuple_3 p2)
+let tuple3_cmp p1 p2 = Tuple3.comp compare_term compare_term Z.compare (normal_form_tuple_3 p1) (normal_form_tuple_3 p2)
+let tuple3_hash p1 = Hashtbl.hash (normal_form_tuple_3 p1)
+let tuple2_equal p1 p2 = Tuple2.eq equal_term equal_term (normal_form_tuple_2 p1) (normal_form_tuple_2 p2)
+let tuple2_cmp p1 p2 = Tuple2.comp compare_term compare_term (normal_form_tuple_2 p1) (normal_form_tuple_2 p2)
+let tuple2_hash p1 = Hashtbl.hash (normal_form_tuple_2 p1)
+
+type prop = Equal of ((term * term * Z.t) [@equal tuple3_equal][@compare tuple3_cmp][@hash tuple3_hash])
+          | Nequal of ((term * term * Z.t) [@equal tuple3_equal][@compare tuple3_cmp][@hash tuple3_hash])
+          | BlNequal of ((term * term) [@equal tuple2_equal][@compare tuple2_cmp][@hash tuple2_hash])
```
sim642, line 40: It might be worth giving (semantic) names to the types `term * term * Z.t` and `term * term` above and use them here. That would avoid all the extra attributes here.

src/cdomains/duplicateVars.ml
```diff
@@ -6,19 +6,20 @@ open CilType
...
-  let hash_typ _ = 0
-  let compare_typ _ _ = 0
+  let equal_typ a b = Stdlib.compare a b = 0
+  let hash_typ x = Hashtbl.hash x
+  let compare_typ a b = Stdlib.compare a b
 
-  type t = AssignAux of (typ[@compare.ignore][@eq.ignore][@hash.ignore])
-         | ReturnAux of (typ[@compare.ignore][@eq.ignore][@hash.ignore])
+  type t = AssignAux of typ
+         | ReturnAux of typ
```
sim642, line 22: Is this comment outdated? Because they are used now.

And as such, `CilType.Typ.t` should be used instead such that correct implementations are used. CIL types are cyclic, so `Stdlib.compare` may not terminate.

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+
+   - `diseq` represents the block disequalities. It is a map that maps each term to a set of terms that are definitely in a different block.
+
+   - `bldis` represents the disequalities. It is a map from a term t1 to a map from an integer z to a set of terms T, which represents the disequality t1 != z + t2 for each t2 in T.
+*)
+type t = {uf: TUF.t;
+          set: SSet.t;
+          map: LMap.t;
+          normal_form: T.v_prop list Lazy.t[@compare.ignore][@eq.ignore][@hash.ignore];
+          diseq: Disequalities.t;
```
sim642, line 562: ppx_deriving does not have any `[@compare.ignore]` or `[@eq.ignore]` attributes (ppx_compare does, but we don't use that).
So this uses the built-in behavior: always forces the lazy evaluation and compares that. This might've had an unintended and unexpected performance impact.

I only noticed this now due to the weird `Lazy.hash` definition above: ppx_deriving_hash doesn't have `[@hash.ignore]` but it also doesn't have built-in `Lazy.t` support (in the released version 0.1.2 at least), so that's why the compiler complains.

I think explicitly overriding all three for the entire type (in parenthesis!) should work:
```suggestion
          normal_form: (T.v_prop list Lazy.t[@compare fun _ _ -> 0][@eq fun _ _ -> true][@hash fun _ -> 0]);
```
The `Lazy` module extension above shouldn't be necessary then either.

src/common/util/richVarinfo.ml
```diff
@@ -17,10 +10,78 @@ sig
...
+     vinline=Option.default v.vinline update.vinline_;
+     vdecl=Option.default v.vdecl update.vdecl_;
+     vinit=Option.default v.vinit update.vinit_;
+     vid=v.vid;
+     vaddrof=Option.default v.vaddrof update.vaddrof_;
+     vreferenced=Option.default v.vreferenced update.vreferenced_;
+     vdescr=v.vdescr;
+     vdescrpure=v.vdescrpure;
+     vhasdeclinstruction=v.vhasdeclinstruction}
+end
```
sim642, line 69: This seems like a massive amount of boilerplate for what exactly? The fields of `varinfo` _are_ mutable, so the type could just be changed in-place with `v.vtype <- t` without introducing copies of the `varinfo` with the same `vid` but different data.

When `varinfo`s are correctly compared with `CilType.Varinfo`, then that only depends on `vid` anyway, so such mutations shouldn't break any data structures they are in either.

src/config/options.schema.json
```diff
@@ -1144,6 +1144,39 @@
...
+          "properties": {
+            "askbase": {
+              "title": "ana.c2po.askbase",
+              "description": "If true, the C-2PO Analysis uses the MayPointTo query to infer additional disequalities.",
+              "type": "boolean",
+              "default": true
+            },
+            "join_algorithm": {
+              "title": "ana.c2po.join_algorithm",
+              "description": "Which join algorithm 'c2po' should use. Values: 'partition' (default): the more efficient version, it uses only the partition to compute the join. Circular equalities are lost during the join; 'partition': a more precise version, uses the automaton to compute the join.",
```
FelixKrayer, line 1159: Just quickly skimmed over it since I listened to your presentation and it sounded interesting.
The only thing I saw was that in this description you describe the value `'partition'` twice. That is probably a typo and the second one should be `'precise'`.
Also, I am happy that the taintPartialContext analysis from my bachelor's thesis gets another use here

michael-schwarz:
@jerhard: You promised a review on this until 2024-11-19 in the Gobcon. Do you have an estimate when you will be able to provide a review?

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+    List.fold add_change
+
+  let filter_map_lhs f (diseq:t) =
+    Enum.fold
+      (fun diseq t -> match f t with
+         | None -> TMap.remove t diseq
+         | Some t2 ->
+           if not (T.equal t t2)
+           then TMap.add t2 (TMap.find t diseq) (TMap.remove t diseq) else
+             diseq) diseq (TMap.keys diseq)
```
jerhard, line 95: indentation.

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+
+  (** Find all elements that are in the same equivalence class as t. *)
+  let comp_t uf t =
+    let (t',z',uf) = TUF.find uf t in
+    fst(List.fold_left (fun (comp,uf) (v,((p,z),_)) ->
+        let v', z'',uf = TUF.find uf v in
+        if T.equal v' t' then (v, Z.(z'-z''))::comp,uf else comp,uf
+      )
+        ([],uf) (TMap.bindings uf))
+
```
jerhard, line 166: Indentation.

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+        if Z.equal r1 r2 then rest
+        else (* r1 <> r2 *)
+          let l1 = match map_find_opt (v,r1) arg
+            with None -> []
+               | Some list -> list in
+          (* just take the elements of set1 ? *)
+          let l2 = match map_find_opt (v,r2) arg
+            with None -> []
+               | Some list -> list in
+          fold_left2 (fun rest (v1,r'1) (v2,r'2) ->
```
jerhard, line 227: Indentation of `in`.

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+    List.fold_left (fun s (v,r,v') ->
+        s ^ "\t" ^ T.show v ^ ( if Z.equal r Z.zero then "" else if Z.leq r Z.zero then (Z.to_string r) else (" + " ^ Z.to_string r) )^ " = "
+        ^ T.show v' ^  "\n") "" clist
+
+  let show_arg arg =
+    let clist = bindings_args arg in
+    List.fold_left (fun s (v,z,v',r) ->
+        s ^ "\t" ^ T.show v' ^ ( if Z.equal r Z.zero then "" else if Z.leq r Z.zero then (Z.to_string r) else (" + " ^ Z.to_string r) )^ " --> "
+        ^ T.show v^ "+"^ Z.to_string z ^  "\n") "" clist
+
```
jerhard, line 344: In case such a construction turns out to be overly expensive, one may consider using `Buffer`s to build the strings more efficiently.

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+  (** Sometimes it's important to keep the dereferenced term,
+      even if it's not technically possible to dereference it from a point of view of the C types.
+      We still need the dereferenced term for the correctness of some algorithms,
+      and the resulting expression will never be used, so it doesn't need to be a
+      C expression that makes sense. *)
+  let deref_term_even_if_its_not_possible min_term z set =
+    match deref_term min_term z set with
+    | result -> result
+    | exception (T.UnsupportedCilExpression _) ->
```
jerhard, line 425: Rather use `try ... with` here.

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+(** Returns a list of all the transition that are present in the automata. *)
+let get_transitions (uf, map) =
+  List.concat_map (fun (t, zmap) ->
+      (List.map (fun (edge_z, res_t) ->
+           (edge_z, t, TUF.find_no_pc uf res_t)) @@
+       (LMap.zmap_bindings zmap)))
+    (LMap.bindings map)
+
+let exactly_equal cc1 cc2 =
+  cc1.uf == cc2.uf && cc1.map == cc2.map && cc1.diseq == cc2.diseq && cc1.bldis == cc2.bldis
```
jerhard, line 578: Why does one not check for structural equality here, as well? 
Or is it so that if these are not at the same address, one may as well check for semantic equality otherwise?

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+    all the terms that are present in those block disequalities.
+*)
+let rec add_normalized_bl_diseqs cc = function
+  | [] -> cc
+  | (t1,t2)::bl_conjs ->
+    let t1',_,uf = TUF.find cc.uf t1 in
+    let t2',_,uf = TUF.find uf t2 in
+    if T.equal t1' t2' then raise Unsat (*unsatisfiable*)
+    else let bldis = BlDis.add_block_diseq cc.bldis (t1',t2') in
+      add_normalized_bl_diseqs {cc with bldis;uf} bl_conjs
```
jerhard, line 801: Indentation.

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+  else
+    (* If the equality is *(t1' + z1) = *(t2' + z2), then we check if the two pointers are equal,
+       i.e. if t1' + z1 = t2' + z2.
+       This is useful when the dereferenced elements are not pointers and therefore not stored in our data strutcure.
+       But we still know that they are equal if the pointers are equal. *)
+  if Z.equal r Z.zero then
+    match t1,t2 with
+    | Deref (t1', z1, _),  Deref (t2', z2, _) ->
+      eq_query cc (t1', t2', Z.(z2 - z1))
+    | _ -> (false, cc)
```
jerhard, line 847: Indentation

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+      (*explicit dsequalities*)
+      Disequalities.map_set_mem (v2,Z.(r2-r1+r)) v1 cc.diseq
+
+(** Adds equalities to the data structure.
+    Throws "Unsat" if a contradiction is found.
+    Does not update disequalities. *)
+let meet_pos_conjs cc pos_conjs =
+  let res = let cc = insert_set cc (fst (SSet.subterms_of_conj pos_conjs)) in
+    closure cc pos_conjs
+  in if M.tracing then M.trace "c2po-meet" "MEET_CONJS RESULT: %s\n" (show_conj (get_conjunction res)); res
```
jerhard, line 877: Indetation.

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+let find_new_root new_reps uf v =
+  match TMap.find_opt v new_reps with
+  | None -> uf, None
+  | Some (new_t, z1) ->
+    let t_rep, z2, uf = TUF.find uf new_t in
+    uf, Some (t_rep, Z.(z2-z1))
+
+let remove_terms_from_diseq diseq new_reps cc =
+  let disequalities = Disequalities.get_disequalities diseq
+  in
```
jerhard, line 983:
```suggestion
  let disequalities = Disequalities.get_disequalities diseq in
```

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+  in congruence_neq {cc with uf} new_diseq
+
+let remove_terms_from_bldis bldis new_reps cc =
+  let uf_ref = ref cc.uf in
+  let find_new_root_term t =
+    let uf, new_root = find_new_root new_reps !uf_ref t in
+    uf_ref := uf;
+    Option.map fst new_root in
+  let bldis = BlDis.filter_map_lhs find_new_root_term bldis in
+  !uf_ref, BlDis.filter_map find_new_root_term bldis
```
jerhard, line 1006: What is the intention here? That `uf_ref` is first returned, and then the `filter_map find_new_root_term` modifies `uf_ref` on the side, but without effect?

src/cdomains/congruenceClosure.ml
```diff
@@ -0,0 +1,1275 @@
...
+    | Deref (t, z,_), Deref (v, z',_) ->
+      let (q', z1') = TUF.find_no_pc cc.uf v in
+      let (q, z1) = TUF.find_no_pc cc.uf t in
+      let s' = T.get_size t2 in
+      let diff = Z.(-z' - z1 + z1' + z) in
+      (* If they are in the same equivalence class and they overlap, then they are equal *)
+      (if T.equal q' q && there_is_an_overlap s s' diff then true
+       else
+         (* If we have a disequality, then they are not equal *)
+       if neq_query cc (t,v,Z.(z'-z)) then false else
```
jerhard, line 1252: indentation

tests/regression/83-c2po/28-return-value.c
```diff
@@ -0,0 +1,16 @@
...
+void *d(const *e) { return e + 200; }
+int *f() {}
+main() {
+  g(a, c, b);
+  if (0) {
+    __goblint_check(0); // NOWARN (unreachable)
+  }
+  __goblint_check(1); // reachable
+}
+g(int, struct h *, struct i *) {
```
jerhard, line 12: Add prototype definition of `g` before usage.

jerhard (changes requested):

src/analyses/startStateAnalysis.ml
```diff
@@ -0,0 +1,58 @@
...
+open Analyses
+open DuplicateVars.Var
+
+(**First all parameters (=formals) of the function are duplicated (by using DuplicateVars),
+    then we remember the value of each local variable at the beginning of the function
+  in this new duplicated variable. *)
+module Spec : Analyses.MCPSpec =
+struct
+  let name () = "startState"
+  module VD = BaseDomain.VD
```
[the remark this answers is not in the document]
jerhard: I removed the definition.

src/analyses/startStateAnalysis.ml
```diff
@@ -0,0 +1,56 @@
...
+    | _ -> AD.top ()
+
+  let startcontext () = D.empty ()
+  let startstate v = D.bot ()
+  let exitstate = startstate
+
+  let query ctx (type a) (q: a Queries.t): a Queries.result =
+    let open Queries in
+    match q with
+    | MayPointTo e -> eval (ask_of_ctx ctx) ctx.local e
```
jerhard, line 44: How can this answer `MayPointTo` queries?

jerhard: This seems to only answer `MayPointTo` queries about ghost variables that are meant to capture information about the start state of the function.

However, it would be better to introduce a new type of query specifically for those ghost variables, something like `StartStateMayPointTo` or `GhostMayPointTo`

jerhard:
@michael-schwarz and I had a look at the treatment of function calls in the c-2po analysis. We noticed the following:
- C-2PO does not keep equalities between parameters and actuals. E.g., when analyzing a function call `f(&g)`, the C-2PO analysis of `f` will not know that the parameter has the value `&g`. It may make sense to track this information in C-2PO.
- Currently, the `combine_env` and `combine_assign` function share a larger section of code. It may be possible to avoid this duplication if one were to keep ghost variables in the `combine_env`, thus avoiding reconstruction in `combine_assign`.

src/cdomains/congruenceClosure.ml
jerhard: I made the suggested change and removed the `Lazy` module extension.

src/analyses/startStateAnalysis.ml
jerhard: The C-2PO analysis poses queries about expressions that may involve a ghost or a  non-ghost variable using `MayPointTo`, inside `CongurenceClosure.MayBeEqual`. Introducing a separation between queries for ghost variables and non-ghost variables would require to visit the expression that is passed to the query to evaluate whether it contains a ghost variable; and depending on the outcome dispatch a different query. 
It seems that this makes things more convulted.

src/cdomains/duplicateVars.ml
jerhard: I removed the comment.

src/cdomains/congruenceClosure.ml
jerhard: Resolve for now. If there is more interest in the domain, one may try tuning it.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,722 @@
...
+
+  let term_of_varinfo vinfo =
+    if is_struct_type (Var.to_varinfo vinfo).vtype || (Var.to_varinfo vinfo).vaddrof then
+      Deref (Addr vinfo, Z.zero, Lval (Var (Var.to_varinfo vinfo), NoOffset))
+    else
+      aux_term_of_varinfo vinfo
+
+  (** Convert a Cil offset to an integer offset.
+      Copied from memOutOfBounds.ml. *)
+  let cil_offs_to_idx (ask: Queries.ask) offs typ =
```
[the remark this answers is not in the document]
jerhard: The duplicated code is modified and not an exact copy; it changes, e.g., the used `ikind`.  Thus, the reuse is not so obvious/ would require some refactoring.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,722 @@
...
+            end
+          | _ -> of_cil_neg ask false e, (None, Some Z.zero)
+        end
+      | _ -> of_cil_neg ask false e, (None, Some Z.zero)
+    in if neg then neg_t, pos_t else pos_t, neg_t
+
+  (** `prop_of_cil e pos` parses the expression `e` (or `not e` if `pos = false`) and
+      returns a list of length 1 with the parsed expresion or an empty list if
+        the expression can't be expressed with the type `prop`. *)
+  let rec prop_of_cil ask e pos =
```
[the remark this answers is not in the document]
jerhard: At the locations where this is used, a list is more convenient, since other operations such as `filter_valid_pointers` and `split` are defined on arguments of type `props list`.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,722 @@
...
+               let uf = modify_size parent_v uf (fun s -> s - size_v) in
+               let uf = modify_size v' uf ((+) size_v)
+               in Z.(r0+r''),uf) (Z.zero, uf) (v::list)
+           in v',r',uf
+         else search v' (v :: list)
+       in search v' [v])
+
+  (**
+     For a variable t it returns the reference variable v and the offset r.
+     This find DOES NOT perform path compression.
```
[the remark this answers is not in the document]
jerhard: The `t` is referring to a parameter `v`. I adapted the comment.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,722 @@
...
+    | Some (parent_t, _) -> T.equal v parent_t
+
+  (**
+     For a variable t it returns the reference variable v and the offset r.
+     This find performs path compression.
+     It returns als the updated union-find tree after the path compression.
+
+     Throws "Unknown value" if t is not present in the data structure.
+     Throws "Invalid Union Find" if it finds an element in the data structure that is a root but it has a non-zero distance to itself.
+  *)
```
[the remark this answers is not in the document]
jerhard: The `t` is referring to a parameter `v`. I adapted the comment.

src/cdomains/congruenceClosure.ml
jerhard: I adapted the code so that it is clearer the that `BlDis.filter_map` function in the last line is executed first.

src/common/util/richVarinfo.ml
jerhard: The idea here was that one could set all the values of the `varinfo`s that are generated (except the `id`)  by providing optional values for them.  Currently, in the `master` branch, one can only specify the `vname` and the `vtype`.

Previously, this PR looked up the values of some attributes of the rich `varinfo`s, such as whether their `vglob` or `vaddrof` attributes are `true`. However, due to some refactoring, this is no longer necessary.

jerhard (commented):
I adapted the PR such that the code is a bit easier to read. 
 
We discussed in [this comment](https://github.com/goblint/analyzer/pull/1485#issuecomment-2651432902) to change the handling of function calls. I did not perform these changes now, as currently we don't have big further plans with this domain. 
A more precise handling of function calls may still be interesting in the future.

jerhard (approved):

jerhard:
I don't quite see what the problem with the failing regression test is? It seems like the ordering of race warnings somehow has changed in one test case?

michael-schwarz:
It may be due to me enabling `warn.deterministic` on master for some tests - I merged that commit in now, so let's see!

michael-schwarz:
That seems to indeed have been it!

src/cdomains/duplicateVars.ml
```diff
@@ -0,0 +1,128 @@
...
+  let hash_typ x =
+    Hashtbl.hash x
+
+  let compare_typ a b =
+    (* TODO: Structural compare on (possibly cyclic) data type typ may not terminate in incremental analysis. *)
+    Stdlib.compare a b
+
+  type t =
+    | AssignAux of typ
+    | ReturnAux of typ
```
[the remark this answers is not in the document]
jerhard: The code now uses `CilType.Typ`  for `equal`, `hash`, and `compare`.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,1066 @@
...
+    match Cil.unrollType t with
+    | TPtr(typ, _) ->
+      is_struct_type typ
+    | _ ->
+      false
+
+  let is_ptr_type t =
+    match Cil.unrollType t with
+    | TPtr _ -> true
+    | _ -> false
```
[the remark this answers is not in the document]
jerhard: I removed the function definition, and use `GoblintCil.isPointerType` instead.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,1066 @@
...
+
+  let rec get_size_in_bits typ = match typ with
+    | TArray (typ, _, _) -> (* we treat arrays like pointers *)
+      get_size_in_bits (TPtr (typ,[]))
+    | _ ->
+      try Z.of_int (bitsSizeOf typ) with
+      | GoblintCil__Cil.SizeOfError (msg,_) when msg ="abstract type"->
+        Z.one
+      | GoblintCil__Cil.SizeOfError (msg,_) ->
+        raise (UnsupportedCilExpression msg)
```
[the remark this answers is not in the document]
jerhard: Thanks, I incorporated this.

src/cdomains/unionFind.ml
```diff
@@ -0,0 +1,1066 @@
...
+  let props_equal = List.equal equal_v_prop
+
+  let is_addr = function
+    | Addr _ -> true
+    | _ -> false
+
+  exception UnsupportedCilExpression of string
+
+  let rec get_size_in_bits typ = match typ with
+    | TArray (typ, _, _) -> (* we treat arrays like pointers *)
```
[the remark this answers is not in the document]
jerhard: I added `Cil.unrollType` in the cases I saw in `UnionFind` where a matching was done over a type.

sim642 (approved):

sim642:
Somehow every unlocked CI job on master after merging this is failing with
```
Fatal error: exception Sys_error("run/config.json: No such file or diretory")
```

PR #1492, opened 2024-05-27
Fix `mutex-meet` for malloc after thread creation

### LINKED ISSUE #1489: Apron: Both branches dead after malloc
In the epic saga of problems with both branches being dead in the relational analysis, we proudly present the next example:

~~~C
// SKIP PARAM: --set ana.activated[+] apron  --set ana.relation.privatization mutex-meet --sets ana.apron.domain interval
// Checks that assinging to malloc'ed memory does not cause both branches to be dead
#include <pthread.h>

void nop(void* arg) {
}

void main() {
  pthread_t thread;
  pthread_create(&thread, 0, &nop, 0);

  long *k = malloc(sizeof(long));
  *k = 5;
  if (1)
    ;
}

~~~

michael-schwarz:
The issue in #1489 was that the global invariant for the blob was still `bot`, making the `meet` deadcode.

TODO:
- [x] `mutex-meet-atomic` privatization still unsound
- [x] Verify if these are really all changes needed for `mutex-meet-atomic` (It seems weird that no handling of these cases is necessary in `get_mutex_global_g_with_mutex_inits_atomic` CC: @sim642)

Closes #1489


sim642:
`mutex:[__VERIFIER_atomic]` also starts out with `bottom (env: [||])`, so that's not the difference I suspect. Rather, the difference comes from this check: https://github.com/goblint/analyzer/blob/896f236c98ba829aee25e5f02cb08d4bb2bba9b1/src/analyses/apron/relationPriv.apron.ml#L500-L504
`get_mutex_global_g_with_mutex_inits_atomic` doesn't have such condition and always does the join (with non-bottom), which makes everything fine.
I wonder if also checking `RD.mem_var get_mutex_global_g g_var` before doing this would be a fix closer to the existing logic.

sim642 (approved):

PR #1493, opened 2024-05-28
make cil_exp_of_linexpr1 work with fractional expressions

### LINKED ISSUE #1328: `invariant` for `relationAnalysis`: Lincons with fractional coefficients not outputted
[...] The invariant is given as `i#318-1/3j#319+1/3k#320+4/3=0` which then likely does not survive conversion to a CIL expression. I suggest we open a new issue, as this is a general problem of `relationAnalysis` [...]

_Originally posted by @michael-schwarz in https://github.com/goblint/analyzer/pull/1327#discussion_r1450413546_

DrMichaelPetter:
Fixes #1328 , use Q instead of Z to extract coefficients, and then scale the coefficient with the lcm of their denominators.


src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -20,21 +20,21 @@ let int_of_scalar ?round (scalar: Scalar.t) =
...
-            | Some `Floor -> Some (Mpzf.fdiv_q n d) (* floor division *)
-            | Some `Ceil -> Some (Mpzf.cdiv_q n d) (* ceiling division *)
-            | None -> None
+            | Some `Floor -> Some (Mpzf.fdiv_q n d, Mpzf.of_int 1) (* floor division *)
+            | Some `Ceil -> Some (Mpzf.cdiv_q n d, Mpzf.of_int 1) (* ceiling division *)
+            | None -> Some (n,d)
           end
       in
-      Z_mlgmpidl.z_of_mpzf z
+      Q.make (Z_mlgmpidl.z_of_mpzf n) (Z_mlgmpidl.z_of_mpzf d)
```
sim642, line 37: This is `int_of_scalar`, which probably should remain `Z.t`-returning. A separate `frac_of_scalar` would be a better choice. Especially because it could largely avoid the rounding business and simply convert `Mpqf`-s into `Q.t`.

sim642: Apparently there's even `Z_mlgmpidl.q_of_mpqf`.

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -285,11 +287,38 @@ struct
...
+      match c with
+      | Scalar c -> BatOption.map (fun q -> Q.den q) (frac_of_scalar c)
+      | _ -> None
+    in
+    let lcm_denom = ref (BatOption.default Z.one (extract_den (Linexpr1.get_cst linexpr1) ())) in
+    let lcm_coeff (c:Coeff.union_5) v =
+      match (extract_den c v) with
+      | Some z -> lcm_denom := Z.lcm z !lcm_denom
+      | _      -> ()
+    in
```
michael-schwarz, line 311: Do we really need the `ref` here?

michael-schwarz (approved):
It would be good to add a (cram) test where we test that the information now actually appears in the witness. Maybe @sim642 can help setting that up?

src/cdomains/apron/sharedFunctions.apron.ml
michael-schwarz: I guess we do, the API doesn't really provide anything helpful here:

https://antoinemine.github.io/Apron/doc/api/ocaml/Linexpr1.html

DrMichaelPetter: I did not like the ref either, but did not immediately see a way to easily avoid it. At least it is contained in a small area

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -285,11 +287,38 @@ struct
...
 
 
+  let lcm_den linexpr1 =
+    let exception UnsupportedScalar 
+    in
+    let frac_of_scalar scalar =
+      if Scalar.is_infty scalar <> 0 then (* infinity means unbounded *)
+        None
+      else match scalar with
+        | Float f -> if Stdlib.Float.is_integer f then Some (Q.of_float f) else None
```
sim642, line 297: According to [Zarith documentation](https://antoinemine.github.io/Zarith/doc/latest/Q.html#VALof_float), `Q.of_float` is exact, so shouldn't this also work for non-integers?

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -285,11 +287,38 @@ struct
...
+        | Mpqf f -> Some (Z_mlgmpidl.q_of_mpqf f)
+        | _ -> raise UnsupportedScalar
+    in
+    let extract_den (c:Coeff.union_5) _ =
+      match c with
+      | Scalar c -> BatOption.map (fun q -> Q.den q) (frac_of_scalar c)
```
sim642, line 303:
```suggestion
      | Scalar c -> BatOption.map Q.den (frac_of_scalar c)
```

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -20,18 +21,19 @@ let int_of_scalar ?round (scalar: Scalar.t) =
...
-          Some n
+          Some (Mpzf.mul scale n)
         else
           begin match round with
-            | Some `Floor -> Some (Mpzf.fdiv_q n d) (* floor division *)
-            | Some `Ceil -> Some (Mpzf.cdiv_q n d) (* ceiling division *)
-            | None -> None
+            | Some `Floor -> Some (Mpzf.mul scale (Mpzf.fdiv_q n d)) (* floor division *)
+            | Some `Ceil ->  Some (Mpzf.mul scale (Mpzf.cdiv_q n d)) (* ceiling division *)
+            | None -> Some (Mpzf.divexact (Mpzf.mul scale n ) d) (* scale, preferably with common denominator *)
```
sim642, line 36: Previously this was `None` without rounding specified, but this seems to change the behavior of `int_of_scalar` (with the default 1 scaling factor). What happens if exact division isn't possible? This always returns `Some ...`.

DrMichaelPetter: Yeah, this deserves some attention as soon as I am back from Italy.

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -6,7 +6,8 @@ open GobApron
 
 module M = Messages
 
-let int_of_scalar ?round (scalar: Scalar.t) =
+
+let int_of_scalar ?(scalewith=Z.one) ?round (scalar: Scalar.t) =
```
sim642, line 10: Now looking at it again, it seems odd that `int_of_scalar` would have the `scalewith` argument. All it does is do an extra multiplication at the very end. That multiplication could just be done by whoever calls `int_of_scalar` themselves without overloading this function with even more complexity.

I would understand that need if the scaling happened before the rounding division or something, but it even doesn't.

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -285,11 +289,38 @@ struct
...
+    in
+    let frac_of_scalar scalar =
+      if Scalar.is_infty scalar <> 0 then (* infinity means unbounded *)
+        None
+      else match scalar with
+        | Float f -> if Stdlib.Float.is_integer f then Some (Q.of_float f) else None
+        | Mpqf f -> Some (Z_mlgmpidl.q_of_mpqf f)
+        | _ -> raise UnsupportedScalar
+    in
+    let extract_den (c:Coeff.union_5) _ =
```
sim642, line 303: What's the point of this wildcard argument?

src/cdomains/apron/sharedFunctions.apron.ml
DrMichaelPetter: The rounding division part is not the crucial point here, the scaling is only done there for consistency. The main focus is the part, where integer division is done - here, the scaling happens before the division. This is also, why we can not pull this out of int_of_scalar imho

src/cdomains/apron/sharedFunctions.apron.ml
DrMichaelPetter: That is a relic. Thanks for spotting.

src/cdomains/apron/sharedFunctions.apron.ml
sim642: Right, nevermind. I thought for a moment that if it's exact division, then it shouldn't matter, but it does when the factors of `d` are split between `scale` and `n`.

src/cdomains/apron/sharedFunctions.apron.ml
```diff
@@ -20,18 +21,21 @@ let int_of_scalar ?round (scalar: Scalar.t) =
...
+          Some (Mpzf.mul scale n)
         else
           begin match round with
-            | Some `Floor -> Some (Mpzf.fdiv_q n d) (* floor division *)
-            | Some `Ceil -> Some (Mpzf.cdiv_q n d) (* ceiling division *)
-            | None -> None
+            | Some `Floor -> Some (Mpzf.mul scale (Mpzf.fdiv_q n d)) (* floor division *)
+            | Some `Ceil ->  Some (Mpzf.mul scale (Mpzf.cdiv_q n d)) (* ceiling division *)
+            | None -> if Mpz.divisible_p (Mpzf.mul scale n ) d then
+                Some (Mpzf.divexact (Mpzf.mul scale n ) d) (* scale, preferably with common denominator *)
```
sim642, line 37: We should avoid computing `mul scale n` twice here.

sim642:
I should try this on our relational witnesses for Freiburg to see if this produces any new ones we couldn't before.

Issue #1451, opened 2024-05-10
Improve output for component of blob indicating whether it was created with `malloc` or `calloc`


michael-schwarz:
Currently the output seems to be:

- `(trivial arrays:Array: (value:Uninitialized, size:4096, no zeroinit:false), IntDomLifter(intdomtuple):1)` for calloc
- `(trivial arrays:Array: (value:Uninitialized, size:4096, no zeroinit:true), IntDomLifter(intdomtuple):1)` for malloc

sim642:
The "no zeroinit" is quite confusing to read indeed, at minimum it shouldn't be negated, so the values themselves could be flipped.
But also there are places in code that call it "origin" which is particularly confusing.

PR #1497, opened 2024-06-03
Improve output for component of blob indicating whether it was created with `malloc` or `calloc`

karoliineh:
Solves #1451


sim642 (approved):

michael-schwarz (approved):
Took me a while to figure out the old double negation stuff, but it is correct now! Definitely a step towards a more usable output!

PR #1500, opened 2024-06-04
Fix privatization unsoundness due to non-definite and unknown mutexes

sim642:
## Problem
As mentioned in https://github.com/goblint/analyzer/pull/1430#issuecomment-2075246339, privatizations don't do anything special with non-definite mutexes.

> > What are privatizations supposed to do with those? It seems to me that we completely forgot about the non-definite possibilities, which might make things even unsound.
> 
> I think I considered this issue when designing the privatizations. They should still be sound for the original setting according to the following arguments:
> 
> - For `lock` & `mine-W`: Incorporating too many values early makes the analysis imprecise, not unsound; All other operations are based on must-locksets, and for background locksets and minimal locksets only definite elements are kept 
> - For `write`: lock is a no-op, other actions are based on must-lockset
> - For `per-mutex` flavors incl. relational & protection-based: Protection by one of these indefinite addresses means they are always held when the global is accessed, so that is not harmful either
> 
> However, these are indeed a bit shaky in some places and may easily be subject to breakage. So in the long run, it does make sense to refactor this as well.
> 
> Also, with the new additions such as `atomic` and invariants about unprotected globals, I'm not sure if this still all holds.
> 
> _Originally posted by @michael-schwarz in https://github.com/goblint/analyzer/issues/1430#issuecomment-2132284679_

As the added test case shows, this is wrong and all of them are unsound.

## Solution
As discussed during GobCon, the fix is to only call `lock` and `unlock` of privatizations with definite mutexes. This PR adapts the recently-fixed-and-refactored (#1430) must lockset analysis logic to base and relation analyses for these events. Notably, `unlock` of indefinite mutex calls privatization's `unlock` for all may-aliasing held locks. This PR also fixes the long-standing unsoundness in privatizations of unlocking unknown mutexes: just by unlocking all held locks unconditionally.

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -685,13 +685,35 @@ struct
     let st = ctx.local in
     match e with
     | Events.Lock (addr, _) when ThreadFlag.has_ever_been_multi (Analyses.ask_of_ctx ctx) -> (* TODO: is this condition sound? *)
-      Priv.lock (Analyses.ask_of_ctx ctx) ctx.global st addr
+      begin match addr with
+        | UnknownPtr -> ctx.local
+        | Addr (v, _) when ctx.ask (IsMultiple v) -> ctx.local
+        | Addr (v, o) when LockDomain.Mval.is_definite (v, o) ->
+          Priv.lock (Analyses.ask_of_ctx ctx) ctx.global st addr
+        | _ -> ctx.local
+      end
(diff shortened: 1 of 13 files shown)


michael-schwarz (approved):
Nice, thank you for the quick fix!

michael-schwarz:
After cherry-picking this and #1430 onto #1417, the newly added tests seem to not terminate.

michael-schwarz:
I guess it is the same problem as

`// TODO: why nonterm without threadflag path-sens?` on master

michael-schwarz:
I would guess the reason it works on master may be https://github.com/goblint/analyzer/pull/1475

PR #1502, opened 2024-06-04
Narrowing from widening thresholds

Red-Panda64:
This PR allows narrowing of integer intervals when the bound being narrowed is a widening threshold. Beforehand, when an interval was widened to a threshold and the threshold fit (e.g. `widen [0, 0] [0,1] = [0, 10]` for an upper threshold 10), it could no longer be narrowed (even if []). This is because int intervals implement accelerated narrowing and will only narrow from from maximal upper or minimal lower bounds.
This PR will use the more precise bound, only when old bound is maximal/minimal or an upper/lower threshold. The narrowing must still terminate, since upper/lower bounds can only shrink/grow as often as there are thresholds.

Example:
```c
int main() {
    int i;
    for(i = 0; i < 10 && i < 20; i += 3);
    __goblint_check(i <= 12);
}
```
The `i < 20` condition is only in this example to be identified as an upper threshold.
Without this PR, the invariant cannot be shown when wideningThresholds are enabled: i is first widened to `[0, 10]`, then `[0, 20]` at the loop head. Then, the narrowing iteration cannot reduce i to the more precise interval `[0, 12]`. With these changes, the more precise interval is inferred.

### DIFF
--- a/src/cdomain/value/cdomains/intDomain.ml
+++ b/src/cdomain/value/cdomains/intDomain.ml
@@ -714,8 +714,8 @@ struct
     | Some (x1,x2), Some (y1,y2) ->
       let threshold = get_interval_threshold_widening () in
       let (min_ik, max_ik) = range ik in
-      let lr = if Ints_t.compare min_ik x1 = 0 || threshold && IArith.is_lower_threshold x1 then y1 else x1 in
-      let ur = if Ints_t.compare max_ik x2 = 0 || threshold && IArith.is_upper_threshold x2 then y2 else x2 in
+      let lr = if Ints_t.compare min_ik x1 = 0 || threshold && Ints_t.compare y1 x1 > 0 && IArith.is_lower_threshold x1 then y1 else x1 in
+      let ur = if Ints_t.compare max_ik x2 = 0 || threshold && Ints_t.compare y2 x2 < 0 && IArith.is_upper_threshold x2 then y2 else x2 in
       norm ik @@ Some (lr,ur) |> fst
 
 
--- a/src/cdomain/value/cdomains/intDomain.ml
+++ b/src/cdomain/value/cdomains/intDomain.ml
@@ -572,6 +572,16 @@ module IntervalArith (Ints_t : IntOps.IntOps) = struct
     let min_ik' = Ints_t.to_bigint min_ik in
     let t = List.find_opt (fun x -> Z.compare l x >= 0 && Z.compare x min_ik' >= 0) ts in
     BatOption.map_default Ints_t.of_bigint min_ik t
+  let is_upper_threshold u =
+    let ts = if get_interval_threshold_widening_constants () = "comparisons" then WideningThresholds.upper_thresholds () else ResettableLazy.force widening_thresholds in
+    let u = Ints_t.to_bigint u in
+    let t = List.find_opt (fun x -> Z.compare u x = 0) ts in
+    Option.is_some t
+  let is_lower_threshold l =
+    let ts = if get_interval_threshold_widening_constants () = "comparisons" then WideningThresholds.lower_thresholds () else ResettableLazy.force widening_thresholds_desc in
+    let l = Ints_t.to_bigint l in
+    let t = List.find_opt (fun x -> Z.compare l x = 0) ts in
+    Option.is_some t
 end
 
 module IntervalFunctor (Ints_t : IntOps.IntOps): SOverflow with type int_t = Ints_t.t and type t = (Ints_t.t * Ints_t.t) option =
@@ -702,9 +712,10 @@ struct
     match x, y with
     | _,None | None, _ -> None
     | Some (x1,x2), Some (y1,y2) ->
+      let threshold = get_interval_threshold_widening () in
       let (min_ik, max_ik) = range ik in
-      let lr = if Ints_t.compare min_ik x1 = 0 then y1 else x1 in
-      let ur = if Ints_t.compare max_ik x2 = 0 then y2 else x2 in
+      let lr = if Ints_t.compare min_ik x1 = 0 || threshold && IArith.is_lower_threshold x1 then y1 else x1 in
+      let ur = if Ints_t.compare max_ik x2 = 0 || threshold && IArith.is_upper_threshold x2 then y2 else x2 in
       norm ik @@ Some (lr,ur) |> fst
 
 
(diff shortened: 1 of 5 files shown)


Red-Panda64:
Good point. I've made that adjustment. Also, I have fixed an oversight I made: before, the narrow operator did not have to explicitly check whether the new bound is more precise, because it would only change the bound if the previous value was maximal/minimal. Now it does have to check.

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -572,6 +572,16 @@ module IntervalArith (Ints_t : IntOps.IntOps) = struct
     let min_ik' = Ints_t.to_bigint min_ik in
     let t = List.find_opt (fun x -> Z.compare l x >= 0 && Z.compare x min_ik' >= 0) ts in
     BatOption.map_default Ints_t.of_bigint min_ik t
+  let is_upper_threshold u =
+    let ts = if get_interval_threshold_widening_constants () = "comparisons" then WideningThresholds.upper_thresholds () else ResettableLazy.force widening_thresholds in
+    let u = Ints_t.to_bigint u in
+    let t = List.find_opt (fun x -> Z.compare u x = 0) ts in
+    Option.is_some t
```
sim642, line 579:
```suggestion
    List.exists (fun x -> Z.compare u x = 0) ts
```

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -572,6 +572,16 @@ module IntervalArith (Ints_t : IntOps.IntOps) = struct
...
+  let is_lower_threshold l =
+    let ts = if get_interval_threshold_widening_constants () = "comparisons" then WideningThresholds.lower_thresholds () else ResettableLazy.force widening_thresholds_desc in
+    let l = Ints_t.to_bigint l in
+    let t = List.find_opt (fun x -> Z.compare l x = 0) ts in
+    Option.is_some t
```
sim642, line 584:
```suggestion
    List.exists (fun x -> Z.compare l x = 0) ts
```

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -1382,8 +1393,9 @@ struct
       let min_ys = minimal ys |> Option.get in
       let max_ys = maximal ys |> Option.get in
       let min_range,max_range = range ik in
-      let min = if min_xs =. min_range then min_ys else min_xs in
-      let max = if max_xs =. max_range then max_ys else max_xs in
+      let threshold = get_interval_threshold_widening () in
+      let min = if min_xs =. min_range || threshold && min_ys <. min_xs && IArith.is_lower_threshold min_xs then min_ys else min_xs in
```
sim642, line 1397: I think they are flipped here
```suggestion
      let min = if min_xs =. min_range || threshold && min_xs <. min_ys && IArith.is_lower_threshold min_xs then min_ys else min_xs in
```

Red-Panda64: You're absolutely right. I've added a test for interval set that would have caught this mistake.

Red-Panda64:
I've also now added a unit test to make the modified narrow operator does indeed narrow from thresholds and does not permit growth. The test applies to intervals and interval sets, as well as upper and lower bounds.

sim642 (approved):

tests/unit/cdomains/intDomainTest.ml
```diff
@@ -218,9 +219,34 @@ struct
     assert_equal imax (I.widen ik i65536 i65537);
     assert_equal imax (I.widen ik i65536 imax)
 
+  let test_interval_narrow _ =
+    GobConfig.set_bool "ana.int.interval_threshold_widening" true;
+    GobConfig.set_string "ana.int.interval_threshold_widening_constants" "comparisons";
```
michael-schwarz, line 224: Is this list ever populated here?

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -572,6 +572,14 @@ module IntervalArith (Ints_t : IntOps.IntOps) = struct
     let min_ik' = Ints_t.to_bigint min_ik in
     let t = List.find_opt (fun x -> Z.compare l x >= 0 && Z.compare x min_ik' >= 0) ts in
     BatOption.map_default Ints_t.of_bigint min_ik t
+  let is_upper_threshold u =
+    let ts = if get_interval_threshold_widening_constants () = "comparisons" then WideningThresholds.upper_thresholds () else ResettableLazy.force widening_thresholds in
+    let u = Ints_t.to_bigint u in
+    List.exists (fun x -> Z.compare u x = 0) ts
```
michael-schwarz, line 578:
```suggestion
    List.exists (Z.equal u x) ts
```

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -572,6 +572,14 @@ module IntervalArith (Ints_t : IntOps.IntOps) = struct
     let min_ik' = Ints_t.to_bigint min_ik in
     let t = List.find_opt (fun x -> Z.compare l x >= 0 && Z.compare x min_ik' >= 0) ts in
     BatOption.map_default Ints_t.of_bigint min_ik t
+  let is_upper_threshold u =
+    let ts = if get_interval_threshold_widening_constants () = "comparisons" then WideningThresholds.upper_thresholds () else ResettableLazy.force widening_thresholds in
```
michael-schwarz, line 576: Why only in the comparisons configuration and not if they're manually supplied?

tests/unit/cdomains/intDomainTest.ml
sim642: It is and does work for the widening test above. The list includes some defaults (powers of 2) in addition to whatever may be collected from the program.

src/cdomain/value/cdomains/intDomain.ml
sim642: It would have to be just `(Z.equal u)` then.

src/cdomain/value/cdomains/intDomain.ml
sim642: This is the same logic as how widening thresholds are looked up in the first place. I think it works with both, but the `comparisons` vs `all` have been implemented very differently.

src/cdomain/value/cdomains/intDomain.ml
michael-schwarz: True!

Red-Panda64: Done!

PR #1504, opened 2024-06-05
Remove some `Option.is_some` usages

sim642:
This mainly switches `Option.is_some` followed by `Option.get` patterns with `GobOption.exists`, which is exactly for this purpose and avoids the double pattern match.

Two tiny refactorings at other `Option.is_some` usages are also included.

### DIFF
--- a/src/analyses/condVars.ml
+++ b/src/analyses/condVars.ml
@@ -116,7 +116,7 @@ struct
     match rval with
     | BinOp (op, _, _, _) when is_cmp op -> (* logical expression *)
       save_expr lval rval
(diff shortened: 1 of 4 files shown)


karoliineh (approved):
LGTM!

Is there a reason why you only considered `Option.is_some` usages? Looking more into some  `Option.get` usages, there are also a lot of patterns of
```
if ... = None then
  ... = Some (...)
Option.get(...)
```
, which could be simplified by using `Option.value`. There might be some other patterns as well, but I didn't look into it too thoroughly yet. Would you want me to look into and do those simplifications as well, and if so, should I do them on this branch or in another PR?

michael-schwarz (approved):

PR #1505, opened 2024-06-06
Simplify some `Option.get` usages

karoliineh:
- Removes the usages of `to_float` that was inherently just an identity function
- Simplifies some if statements with `Option.get` usages by using `Option.fold` instead

### DIFF
--- a/src/cdomain/value/cdomains/intDomain.ml
+++ b/src/cdomain/value/cdomains/intDomain.ml
@@ -2256,7 +2256,7 @@ struct
       shift_op a b
     in
     (* If one of the parameters of the shift is negative, the result is undefined *)
-    let is_negative = Stdlib.Option.fold ~none:true ~some:(fun x -> Z.compare x Z.zero < 0) in
+    let is_negative = Option.fold ~none:true ~some:(fun x -> Z.compare x Z.zero < 0) in
     if is_negative (minimal x) || is_negative (minimal y) then
       top_of ik
     else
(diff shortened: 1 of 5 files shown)


src/common/cdomains/floatOps/floatOps.mli
```diff
@@ -7,7 +7,7 @@ type round_mode =
   | Down
 
 module type CFloatType = sig
-  type t
+  type t = float
```
sim642, line 10: This should still be abstract, but `to_float` could just return `float` instead of `float option`.

sim642 (changes requested):

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -2256,9 +2256,8 @@ struct
       shift_op a b
     in
     (* If one of the parameters of the shift is negative, the result is undefined *)
-    let x_min = minimal x in
-    let y_min = minimal y in
-    if x_min = None || y_min = None || Z.compare (Option.get x_min) Z.zero < 0 || Z.compare (Option.get y_min) Z.zero < 0 then
+    let is_negative = Option.fold ~none:true ~some:(fun x -> Z.compare x Z.zero < 0) in
```
michael-schwarz, line 2259: Any reason not to use the dedicated comparison operators from Z? I think these would make the code easier to use and would be a nice improvement to do together with this :)

michael-schwarz (commented):
Minor nit-pick: This introduces our first usages of `Option.default`, the rest of the codebase uses `BatOption.map_default`.
I know we are very inconsistent with our usage of batteries vs. stdlib but this is one of the few places where we were consistent before.
Any reason to prefer `fold` over `map_default` here?

src/cdomain/value/cdomains/intDomain.ml
sim642: This fold could just be `GobOption.for_all` anyway.

sim642 (approved):

PR #1506, opened 2024-06-08
Fix Apron `compare` function usage

sim642:
https://github.com/antoinemine/apron/issues/99 is mitigated by https://github.com/antoinemine/apron/pull/108, which should be released as Apron v0.9.15 soon.

This PR aims to fix compatibility with this breaking change. It would be possible for us to implement a proper `GobApron.Environment.compare`, but I don't know if we really need it.
I'm more worried about our comparison of `Apron.Abstract1.t`-s via `Stdlib.compare`, which won't work now. It doesn't seem to break our tests, but it might still be possible in some case we end up needing to call such comparison and crashing (which will explicitly break instead of being implicitly wrong sometimes).

### TODO
- [x] Remove Apron from lock file.


sim642:
The new Apron is now published on opam (https://github.com/ocaml/opam-repository/pull/26174) and all versions of Goblint are incompatible (https://github.com/ocaml/opam-repository/pull/26199), so this fix is crucial for release.

This is probably easiest done after #1522.

sim642:
Since the release on opam-repository, all our nightly unlocked CI builds fail spectacularly, so I'll just merge these fixes before #1522.

A 1 minute sv-benchmarks run didn't reveal any new `EXCEPTION (Failure)` results, so we haven't been relying on the old problematic `compare` functions in a significant manner.

PR #1510, opened 2024-06-13
Enable `pthreadMutexType` analysis by default

sim642:
This enables `pthreadMutexType` analysis by default which can make some (by default) warnings about mutex misuse more precise.
However, I would not add it to SV-COMP confs because it has no use there right now. Maybe if we submit some tasks about races with recursive mutexes, there might be use.

### sv-benchmarks ConcurrencySafety
This has minimal impact on sv-benchmarks, where we don't need to use it anyway:
image


src/analyses/mutexTypeAnalysis.ml
```diff
@@ -15,8 +15,9 @@ struct
   (* Removing indexes here avoids complicated lookups and allows to have the LVals as vars here, at the price that different types of mutexes in arrays are not dinstinguished *)
   module O = Offset.Unit
```
michael-schwarz, line 16: Is this alias now used anywhere? Also, we should probably move the comment to where we now use `Mval.Unit`.

michael-schwarz (approved):

src/analyses/mutexTypeAnalysis.ml
sim642: It is in some conversions below, but I moved the comment around.

PR #1512, opened 2024-06-14
Add primitive verdict-based YAML violation witness rejection

### LINKED ISSUE #1301: Reject YAML violation witnesses
We can reject violation witnesses for programs that we can prove correct, assuming we are sound. This probably won't make us very good in the validation track though.

We could also consider actually interpreting the YAML violation witness and refining our analysis with that information. If there's no violation in the state space restricted by the violation witness and we can prove it, the witness is also invalid.

sim642:
Closes #1301.

Most of this is just adding YAML witness `violation_sequence` parsing support and adapting the parsing of locations in YAML witnesses to allow for fields which the updated 2.0 format allows to omit.

The actual rejection part is laughably trivial: do nothing!
1. If program is correct and we can prove it, we output `true`, which counts as refutation of violation witness.
2. If program is correct and we cannot prove it, we output `unknown`.
3. If program is incorrect, we output `unknown`.

Actually using any of the information in the violation sequence is a separate issue. We should be able to kill some paths or refine some values based on that, a la (linear) observer automaton.

### TODO
- [x] Add some tests.

### DIFF
--- a/conf/svcomp-validate.json
+++ b/conf/svcomp-validate.json
@@ -1,0 +1,142 @@
+{
+  "ana": {
+    "sv-comp": {
+      "enabled": true,
+      "functions": true
+    },
+    "int": {
+      "def_exc": true,
+      "enums": false,
+      "interval": true
+    },
+    "float": {
+      "interval": true
+    },
+    "activated": [
+      "base",
+      "threadid",
+      "threadflag",
+      "threadreturn",
+      "mallocWrapper",
+      "mutexEvents",
+      "mutex",
+      "access",
+      "race",
+      "escape",
+      "expRelation",
+      "mhp",
+      "assert",
+      "var_eq",
+      "symb_locks",
+      "region",
+      "thread",
+      "threadJoins",
+      "abortUnless",
+      "unassume"
+    ],
+    "path_sens": [
+      "mutex",
+      "malloc_null",
+      "uninit",
+      "expsplit",
+      "activeSetjmp",
+      "memLeak",
+      "threadflag"
+    ],
+    "context": {
+      "widen": false
+    },
+    "malloc": {
+      "wrappers": [
+        "kmalloc",
+        "__kmalloc",
+        "usb_alloc_urb",
+        "__builtin_alloca",
+        "kzalloc",
+
+        "ldv_malloc",
+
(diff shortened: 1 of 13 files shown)


sim642:
I tried running this on SV-COMP 2024 YAML violation witnesses (it's extremely difficult to even get them). If I did everything correctly, there are at most 160 (probably even much fewer) tasks with expected true verdict. Among them, only two of CPAchecker's we terminate with `unknown`. So we cannot reject anything.

Also doing some refinement with violation sequences might allow us to reject witnesses even for tasks with expected verdict false, if the error path cannot lead to the violation. But I'm not very hopeful about that.

tests/util/yamlWitnessStrip.ml
```diff
@@ -10,8 +10,13 @@ struct
 
   let strip_file_hashes {entry_type} =
     let stripped_file_hash = "$FILE_HASH" in
-    let location_strip_file_hash location: Location.t =
-      {location with file_hash = stripped_file_hash}
+    let location_strip_file_hash (location: Location.t): Location.t =
+      let file_hash =
+        match location.file_hash with
+        | Some _ -> Some stripped_file_hash (* TODO: or just map to None always? *)
```
michael-schwarz, line 16: Was this addressed?

conf/svcomp-validate.json
michael-schwarz, line 1: Is this config used somewhere?

michael-schwarz (approved):

conf/svcomp-validate.json
sim642: It's like `svcomp` conf: it includes changes since the previous SV-COMP submission to be included in the next one. In this case, it's the activation of `abortUnless` (which is already in `svcomp` conf) and `violation_sequence` witness entries.

tests/util/yamlWitnessStrip.ml
sim642: Version 2.0 of the format makes some fields (like `file_hash` for every location) optional/unnecessary. So instead of changing all of them to `$FILE_HASH` in cram tests, we could just now omit them there. I'd prefer to do that in a separate PR though: it means all cram tests about witnesses get changed.

PR #1517, opened 2024-06-19
Simplify `IntDomTuple` witness invariants

sim642:
Before this PR, `IntDomTuple` outputted witness invariants as conjunction of those from each int domain. Except when it was a definite integer, then only the single equality was returned.

This PR extends this logic to avoid obviously redundant and duplicate information in witness invariants, which can make them annoyingly large:
1. If it is a definite integer, output single equality.
2. If it is inclusion set, output conjunction of equalities (interval bounds and exclusions are redundant).
3. Otherwise, output interval bounds based on all integer domains (avoids duplicate bounds from def_exc exclusion bit ranges). Moreover, exclusion set is filtered based on that interval because often def_exc has `x != 0` which is pointless if interval has some strictly positive bounds. Congruence and interval set (latter not used in SV-COMP) domains are added unchanged, so all known information should still be there.

### TODO
- [x] Add option: `ana.base.invariant.int.simplify`.

### DIFF
--- a/src/cdomain/value/cdomains/intDomain.ml
+++ b/src/cdomain/value/cdomains/intDomain.ml
@@ -605,16 +605,28 @@ struct
           Invariant.(a || i) [@coverage off] (* bisect_ppx cannot handle redefined (||) *)
         ) (Invariant.bot ()) ps
 
-  let of_interval e ik (x1, x2) =
-    if Z.equal x1 x2 then
+  let of_interval_opt e ik = function
+    | (Some x1, Some x2) when Z.equal x1 x2 ->
       of_int e ik x1
-    else (
+    | x1_opt, x2_opt ->
       let (min_ik, max_ik) = Size.range ik in
       let inexact_type_bounds = get_bool "witness.invariant.inexact-type-bounds" in
-      let i1 = if inexact_type_bounds || Z.compare min_ik x1 <> 0 then Invariant.of_exp Cil.(BinOp (Le, kintegerCilint ik x1, e, intType)) else Invariant.none in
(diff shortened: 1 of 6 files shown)


michael-schwarz:
> Congruence ... (not used in SV-COMP)

I think congruences are enabled by the autotuner.

sim642:
> I think congruences are enabled by the autotuner.

Sorry, yes. I meant that just about interval sets.

sim642:
There's now an option to disable these simplifications. This includes the possibility to even disable the definite integer simplification.
These simplifications are the default, to give nicest output out of the box. If this turns out to not be precise enough and the abstract values are better, the option can be changed. Although it would be nicer to improve `all` simplifications for such cases as well, if possible.

src/cdomain/value/cdomains/intDomain.ml
```diff
@@ -3778,19 +3790,46 @@ module IntDomTupleImpl = struct
...
-    | Some v ->
-      if get_bool "witness.invariant.exact" then
-        (* If definite, output single equality instead of every subdomain repeating same equality *)
-        Invariant.of_exp Cil.(BinOp (Eq, e, kintegerCilint ik v, intType))
-      else
-        Invariant.top ()
-    | None ->
-      let is = to_list (mapp { fp = fun (type a) (module I:SOverflow with type t = a) -> I.invariant_ikind e ik } x)
-      in List.fold_left (fun a i ->
+  let invariant_ikind e ik ((_, _, _, x_cong, x_intset) as x) =
```
michael-schwarz, line 3793: What do you think of doing a single refine call before generating invariant? This would ensure things such as inclusion sets being more precise than intervals etc.

michael-schwarz (approved):
LGTM! It's nice to see that we now make more use of the possibilities for reuse.

PR #1518, opened 2024-06-19
Remove restriction that no loop unrolling is done when breaks are in the loop

karoliineh:
In SV-COMP no-overflow tasks there are cases similar to the following example:

```c
int counter = 0;
int main() {
    int  X = __VERIFIER_nondet_int();
    long long x = 0;

    while (counter++<10) {
        ...
        if (!(x <= X))
            break;
        ...
    }
    ...
```

Where we do not detect loops with fixed iterations due to the restriction:
> There is one single break in an if statement where the condition compares a local variable v to a constant e.

First, this PR removes this restriction, as reaching these breaks should make the loop iterations smaller anyway.

However, all these tasks also have the increment `counter++` in the loop condition, which we do not handle and which is very difficult to handle using only the semantics, as these statements are picked into pieces by CIL:
```c
int counter = 0;
int main() {
    int  X = __VERIFIER_nondet_int();
    long long x = 0;

    tmp = counter;
    counter = counter + 1;
    while (tmp<10) {
        ...
        if (!(x <= X))
            break;
        ...
        tmp = counter;
        counter = counter + 1;
    }
    ...
```

Thus, I opted for another assumption that when we find a fixed loop and its start but cannot detect the increment within the loop, we assume the increment is 1.

This solved the tasks with the bounds being 5 and 10, but for bounds 20 and 50, I also had to apply the hack introduced in #1516 to just use the fixed loop size that was found by the heuristics if there are less than 10 loops in the program (the only stat I could find/access to estimate the size of the program (being small)).

### DIFF
--- a/src/util/loopUnrolling.ml
+++ b/src/util/loopUnrolling.ml
@@ -45,20 +45,19 @@ class findBreakVisitor(compOption: exp option ref) = object
   method! vstmt stmt =
     match stmt.skind with
     | Block _ -> DoChildren
-    | Break _ -> raise WrongOrMultiple
+    | Break _ -> SkipChildren
     | If (cond, t, e, _, _) ->  (
-        checkNoBreakBlock t;
         match e.bstmts with
         | [s] -> (
             match s.skind with
             | Break _ -> (
                 match !compOption with
-                | Some _ -> raise WrongOrMultiple (*more than one loop break*)
+                | Some _ -> SkipChildren (*more than one loop break*)
                 | _ -> compOption := Some cond; SkipChildren
               )
-            | _ -> checkNoBreakStmt stmt; SkipChildren
+            | _ -> SkipChildren
           )
-        | _ -> checkNoBreakStmt stmt; SkipChildren
+        | _ -> SkipChildren
       )
     | _ ->  SkipChildren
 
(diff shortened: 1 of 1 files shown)


src/util/loopUnrolling.ml
```diff
@@ -31,34 +31,25 @@ class checkNoBreakVisitor = object
...
 
-let checkNoBreakStmt stmt =
-  let visitor = new checkNoBreakVisitor in
-  ignore @@ visitCilStmt visitor stmt
-
-let checkNoBreakBlock block =
-  let visitor = new checkNoBreakVisitor in
-  ignore @@ visitCilBlock visitor block
-
 class findBreakVisitor(compOption: exp option ref) = object
```
sim642, line 34: Looking at this visitor now, I understand why it didn't want multiple `break`s. It's main goal is to actually find the comparison for the loop, which will be an `if` with a `break` inside.
By only allowing one such `break` it guaranteed that the result is unambiguous: there couldn't be multiple such comparisons found.

Now, if there are multiple matching `if`s with `break`s inside, this will still find the first one because `compOption` is only assigned when it's `None`, so that can happen at most once.
Is the loop condition always in the first matching one though? If so, then why did the old version have a problem at all? Or maybe there are loops where the condition isn't actually the first one?

Issue #1513, opened 2024-06-15
Printing of `Apron` values very slow 


michael-schwarz:
With `verbose` mode, we observe terrible slowdowns that seem to be due to very slow pretty-printing.

~~Currently, they are not considered which is bad for long-running benchmarks where `privPrecCompare` may take a long time or even fail to terminate.~~

michael-schwarz:
Turns out almost all the time is spent constructing pretty_diffs:

~~~ocaml
      (if D.leq v1 v2 then nil else dprintf "diff: %a\n" D.pretty_diff (v1, v2))
      ++
      (if D.leq v2 v1 then nil else dprintf "reverse diff: %a\n" D.pretty_diff (v2, v1))
~~~

This takes several orders of magnitude more than just doing the comparisons.

michael-schwarz:
It seems to be somehow not the actual implementation inside apronDomain.ml of `pretty_diff` though: That can be commented out to return `Pretty.nil` and the slowdown still stays.

michael-schwarz:
The culprit seems to in fact be `D.pretty ()` calls...

michael-schwarz:
Which in turn come from Apron, so there must be something super slow happening in Apron.

michael-schwarz:
The flamegraph on the other hand hints at a majority of time (90%) being spent in `camlGoblintCil__Pretty__breakString_148` with the rest of the time being spent in `camlStdlib__Bytes__sub_341`


![flames](https://github.com/goblint/analyzer/assets/13812333/4a3bd26b-b081-48bf-abff-3a26358c7d21)


Perf data in Firefox format: https://gigamove.rwth-aachen.de/en/download/039038f988c7dfbc813e63447a35a0b1(accessible until 07.10).


This does not seem to be a blocker for my thesis, as I will generate missing reports manually in non-verbose mode, but we should look at it.

sim642:
Apparently the `Pretty.text` primitive directly uses that but does something involved under the hood to break the text up further (based on `\n`). It's just the massive strings of Apron constraints that it takes so long to split.

I suspect the Apron printer is using `Format` with some small output width and causing line breaks where we don't want them anyway. I think I've noticed this in tracing output before as well, but never looked into it.

michael-schwarz:
Maybe something can be done here by using `String.split_on_char`?

See e.g. the performance numbers reported here: https://gist.github.com/mooreryan/220b47feea6b253630dab09c4b6ed18c

[...]

PR #1527, opened 2024-06-27
Refactor Apron `Printable`s

sim642:
This implements a fix to my observation that Apron output is always line broken with 80 column margin from `Format` by default https://github.com/goblint/analyzer/issues/1513#issuecomment-2192286933.
This avoids completely arbitrary line breaks (which may still go over column 80 in indented output line tracing) which also make searching the tracing files more annoying. At the same time, it might bypass #1513 (of course we should still optimize it in CIL as well).

This also cleans up other `Printable`-like things in Apron domains by adding the functions into `GobApron`.


michael-schwarz (approved):
I have not checked whether this fixes the performance issues described in https://github.com/goblint/analyzer/issues/1513, but it seems reasonable to have it!

Issue #1465, opened 2024-05-15
`BaseAnalysis`: Investigate why `join` over alternatives for non-definite AD target does not work


michael-schwarz:
For #1458 I also attempted to replace this snippet 

https://github.com/goblint/analyzer/blob/5cd8650a2872341152f48a7d1cd20ac9fa0de994/src/analyses/base.ml#L1735-L1743

with

~~~OCaml
AD.fold (fun addr acc -> D.join (update_one addr st) acc) lval (D.bot ())
~~~

but that lead to some strange test failures where I really have no idea what's going on:

```
3 test(s) failed: ["04/35 trylock_rc", "04/36 trylock_nr", "41/05 more"]
```

This would be the cleaner solution, so we should investigate it.

_Originally posted by @michael-schwarz in https://github.com/goblint/analyzer/pull/1458#discussion_r1599661735_

michael-schwarz:
That's odd indeed. I'm fine with the current improvement then because it fixes the issue and doesn't break anything else.
The naive implementation failing some tests deserves an issue though. I had a quick tracing look and somehow there seemed to be `CPA` maps with many variables missing or `Uninitialized` (i.e. superbot) values, so something odd happens. Maybe when the address set is somehow empty, the new `D.bot ()` comes into play?

_Originally posted by @sim642 in https://github.com/goblint/analyzer/pull/1458#discussion_r1599943567_

karoliineh:
The tests pass when we replace it with:
```c
if AD.is_empty lval then
  st
else
  AD.fold (fun addr acc -> D.join (update_one addr st) acc) lval (D.bot ())
```

1. In "04/35 trylock_rc", and "04/36 trylock_nr", `lval` becomes empty because of the call `time (NULL)`, where in LibFuns we try to write to the argument (`("time", unknown [drop "arg" [w]]);`) that is then a `NullPointer`, which is removed by `collect_invalidate`.
2. In "41/05 more" the same happens because of the call `strtol("bla", 0, 8)`, where we try to write to the second argument `("strtol", unknown [drop "nptr" [r]; drop "endptr" [w]; drop "base" []]);` that is also a `NullPointer`.

I tried to catch all cases where `lval` might be empty with `assert (not @@ AD.is_empty lval);`. Then one more test ("05/21 race-null-type") crashes, but that also contains a call to `time (NULL)`.

Issue #1461, opened 2024-05-14
Reconsider not writing to known addresses if a pointer contains an unknown element


michael-schwarz:
We currently do completely ignore the possibility that any of the known pointers are written as soon as we have `?` in the points-to set. This seems like it is philosophically the wrong thing to do to me.

In my opinion, we should at least potentially update the targets which we have identified.

https://github.com/goblint/analyzer/blob/a84e07d8780d5575ef90ad38cae51afde3964791/src/analyses/base.ml#L1730-L1748

sim642:
I suspect the `SetDomain.Unsupported` code is some legacy thing that never even happens now. `AddressSet.top` isn't a lifted `Top` but a normal (foldable-over) set with `UnknownPtr` in it. And `update_one` there just ignores that one when doing `Addr.to_mval`.

PR #1531, opened 2024-07-01
Refactor writing to pointers

karoliineh:
* [Removes dead code](https://github.com/goblint/analyzer/commit/9485208d7c9898a4c61c9139ea5fe351ca03e809) that indicates as if we are not writing to known addresses if a pointer contains an unknown element, solves #1461
* Fixes issue #1465, and replaces [complicated set logic with simplified AD.fold](https://github.com/goblint/analyzer/commit/65e1ccc7b0a52940466223344a44d6cef4fc15fd), for that:
  * [Asserts that the set of `lval`-s is not empty when written to](https://github.com/goblint/analyzer/commit/c4d96e86c11ddeed260b4ea87c41be42fa4141fe)
  * [Does not remove `NullPtr`-s from address sets](https://github.com/goblint/analyzer/commit/c7199a04656119b03fa3facf80146495790defc6) in  `collect_invalidate` and `reachable_from_value` #1175
  * Fixes a regression test by [including `stdlib.h` for correctly importing strtol](https://github.com/goblint/analyzer/commit/7816bed8049b77290b60d4b50f38b5c80b875884)
* [does not remove `UnknownPtr` before calling `reachable_vars`](https://github.com/goblint/analyzer/commit/a30fbfc3add97d40d400bce360f9cb821439f35e) in `ReachableFrom` #1175

### TODO
- [ ] Test #1562.


karoliineh:
Decisions from GobCon on 09.07:

> What should happen when we write to a pointer that is either known or NullPtr?
> * a) Should we keep the NullPtr and join the known values
> * b) Or should we assume that if it is null, the program crashes anyways, and if it continues, the known value can only be the one that was written

- [ ] Add b) as default behavior, make it configurable (for now), with option called `sem.abort-on-null-deref`

sim642:
There was the matter of naming/describing the option but we discussed it and something seems to have been implemented. Not sure about the test failures though.

I think it makes sense to do this though.

src/analyses/base.ml
```diff
@@ -1731,25 +1727,20 @@ struct
...
-      M.info ~category:Unsound "Assignment to unknown address, assuming no write happened."; st
+    let update_one (x : Addr.t) store =
+      match x with
+      | Addr x -> update_one_addr x store
+      | NullPtr ->
+        begin match get_string "sem.null-pointer.dereference" with
+          | "assume_none" -> D.bot ()
+          | "assume_top" -> store
+          | _ -> assert false
+        end
```
sim642, line 1821: Regarding something I noticed in https://github.com/goblint/analyzer/pull/1777#discussion_r2190778428: this now makes writing to `NULL` more explicit, but somehow there's still no warning.
I'm now wondering if we're actually missing a warning here or we actually emit one from somewhere else already. Should probably figure this out while we're at it.

src/analyses/base.ml
```diff
@@ -1814,23 +1810,20 @@ struct
...
+      | UnknownPtr
+      | StrPtr _ -> store
+    in
+    assert (not @@ AD.is_empty lval);
+    AD.fold (fun addr acc -> D.join (update_one addr st) acc) lval (D.bot ())
```
sim642, line 1826: > [...] Not sure about the test failures though.

The old CI runs seem to be gone, but I now merged master into this and locally at least two malloc_null tests fail.
The path-sensitivity of malloc_null analysis causes a scenario where a path has the `malloc`-ed pointer as `{NULL}`. Writing into that pointer, the assert doesn't fail because the points-to set is non-empty, but due to `assume_none` we end up returning `D.bot ()`.
That bottom is a strange value: it appears live, but has lost all local variables state of base.

I suppose turning a `D.bot ()` result into `raise Deadcode` could work here.
It does reveal a certain asymmetry with `assume_none` on `NULL` pointer reading: the latter doesn't `raise Deadcode` I think,  but just assumes nothing changed.

src/analyses/base.ml
sim642: At least for normal assignments, the warning comes from `eval_lv` which evaluates the left-hand side to a points-to set.

Issue #1542, opened 2024-07-10
Apron: Invariants for variables of type `unsigned long` wrong


michael-schwarz:
Consider the following program extracted from `libaco` which is one of the programs I am considering for #1417:

~~~C
#include<pthread.h>
pthread_mutex_t mtx;
int g;

void *pmain(void *pthread_in_arg )
{
  unsigned long tmp_gl_ct ;

  pthread_mutex_lock(& mtx);
  tmp_gl_ct = g;
  pthread_mutex_unlock(& mtx);
  return ((void *)0);
}

int main()
{
  pthread_t t1;
  pthread_create(& t1, 0, pmain, 0);
  pthread_join(t1, 0);
  return 0;
}
~~~

Creating a YAML witness from it  (via `./goblint --conf conf/traces-rel.json --set ana.activated[+] apron --set ana.path_sens[+] threadflag --set ana.relation.privatization mutex-meet-tid-cluster12 --enable allglobs -v --enable witness.yaml.enabled notlibaco.c --set witness.yaml.path notlibaco.yml`) yields

~~~YAML
- entry_type: location_invariant
  metadata:
    format_version: "0.1"
    uuid: 2bcbb0e1-df8c-4b8d-8a00-ea9aa25ab8b1
    creation_time: 2024-07-10T12:42:46Z
    producer:
      name: Goblint
      version: heads/michael-schwarz-dissertation-0-g970a99913-dirty
      command_line: '''./goblint'' ''--conf'' ''conf/traces-rel.json'' ''--set'' ''ana.activated[+]''
        ''apron'' ''--set'' ''ana.path_sens[+]'' ''threadflag'' ''--set'' ''ana.relation.privatization''
        ''mutex-meet-tid-cluster12'' ''--enable'' ''allglobs'' ''-v'' ''--enable''
        ''witness.yaml.enabled'' ''notlibaco.c'' ''--set'' ''witness.yaml.path'' ''notlibaco.yml'''
    task:
      input_files:
      - notlibaco.c
      input_file_hashes:
        notlibaco.c: e36c8d4402977c57e85e6fe02221ae759457e969afc37d9852c8c500d836c19f
      data_model: LP64
      language: C
  location:
    file_name: notlibaco.c
    file_hash: e36c8d4402977c57e85e6fe02221ae759457e969afc37d9852c8c500d836c19f
    line: 12
    column: 3
    function: pmain
  location_invariant:
[...]

PR #1545, opened 2024-07-16
[Thesis MS] Fixes for Self-Validation Issues

michael-schwarz:
Should address #1542.

Contains:

- Fix for issue where `read_global` read in from a global variable which does not exist in single-threaded mode
- ...

Related but not part of the diff: 

- 3a30b0cb9bf532920fbf211408a439fc39717d87 (check whether cast to `long long` is injective before emitting invariants)


michael-schwarz:
For closing #1542 and #1547, we should backport this and https://github.com/goblint/analyzer/commit/3a30b0cb9bf532920fbf211408a439fc39717d87 to master.

Issue #810, opened 2022-08-02
Re-consider `exp.volatiles_are_top` as default


michael-schwarz:
We currently enable this everywhere, which leads to `volatile` and `extern` variables always having the value `T`.

While this of course safe and makes sense when analyzing e.g. drivers, it may be unnecessarily imprecise to consider all `volatile`s `T` in other settings. We should, e.g., investigate if this can be turned off for `sv-comp` at least.

sim642:
It would also be worth checking if sv-benchmarks even use these. There shouldn't be any `extern`s since the programs are supposed to be closed. There might be `volatile`s though.

The other thing to review is how well `exp.volatiles_are_top` is followed at all. AFAIK Apron analysis doesn't even consider that, but if we have such an option, then the support should be consistent.

michael-schwarz:
There are a few places in sv-comp where `volatile`s are used, though it does not seem like there are sufficient numbers to justify doing this for this year's svcomp.

sim642:
As noted in https://github.com/goblint/analyzer/pull/1552#discussion_r1891393590, only the `none` base privatization even considers the option (for locals). And for some reason during writing, not reading.

So there's a lot to reconsider here besides just the default. Other privatizations and other (relational) value analyses don't use this option and thus behave differently. The option probably makes sense for embedded/driver code, at least for `volatile` globals, but all value analyses should then actually respect it, otherwise they will still return more precise values.

PR #1552, opened 2024-08-02
Optimize `none` base privatization, add eager Vojdani privatization

sim642:
# `none` privatizations
The `none` privatization in base analysis is a sliced copy of the old unsound Vojdani privatization (which was removed in #736). Thus, it is still based on `sync`: `write_global` doesn't side-effect, but adds to local state and `sync` does all the side effects. This is an unnecessarily roundabout way to be (almost) as imprecise as possible.
What's perhaps worse is that it `sync`s unconditionally (so not just at join points) and each time iterates over the entire `cpa`. This could be a stupid hidden cost when wanting super-fast analysis without any privatization, e.g. our `large-program` example conf. This is quadratically bad in program size (nodes * variables) compared to `protection` privatization almost none of this would happen.

In this PR, I've added `NonePriv2`, which doesn't rely on `sync` (except for the inevitable branched thread creation) and surprisingly fails fewer tests when made the default privatization on our regression suite (it's more precise or more sound?!).
It's constructed according to some old traces related work and still has global variables in local states (but reads them joined with global unknowns).

Finally, I've added `NonePriv3`, which never puts any globals into local states (in multi-threaded mode), and has `read_global` and `write_global` directly using `getg` and `sideg`, respectively.
This is what probably should just replace `NonePriv`, but I've added them all right now to make investigating differences and benchmarking slightly easier.

**EDIT:** Now only `NonePriv3` remains in place of `NonePriv`.

# Eager Vojdani privatization
#736 removed the old unsound Vojdani privatization which triggered the whole traces research direction. Its unsoundness was due to an attempt at _lazy reading_ (which was only hinted at in some thesis/paper but I cannot find anymore where).

In this PR, I've added `VojdaniPriv`, which is the _sound_ Vojdani privatization as described in his thesis and various papers. This does _eager reading_, which should avoid the unsoundness. It also doesn't use `sync` because it handles `lock` and `unlock` itself, but that may also be a mistake when trying to represent the original analysis faithfully.

I don't know if we'd want to have this back, but it could also make for interesting benchmarking.


sim642:
> I don't know if we'd want to have this back, but it could also make for interesting benchmarking.

And indeed it did, on SV-COMP ConcurrencySafety at least (if I implemented everything correctly):
* `vojdani` with its eagerness is slower than the others most of the way, which is expected. However, at the high end it's faster than `protection`.
* `vojdani` is also more precise than `protection` on many tasks. Only `protection-read` makes it more precise than `vojdani` again.
* `protection-read` (and by a tiny margin `protection`) are faster than any of the `none` variants, which is quite surprising. I suppose this is due to a fewer side effects and such dependencies, but it could mean that our large-program example conf is recommending a less precise _and slower_ analysis than out of the box.

image

michael-schwarz:
I will hopefully get around to reviewing this within the next week or so!

src/analyses/basePriv.ml
```diff
@@ -117,63 +117,80 @@ struct
...
-      (* Do not impose invariant, will not hold without privatization *)
-      if M.tracing then M.tracel "set" ~var:x.vname "update_one_addr: BAD! effect = '%B', or else is private! " (not invariant);
-      st
-    )
-    else (
-      (* Here, an effect should be generated, but we add it to the local
-      * state, waiting for the sync function to publish it. *)
-      (* Copied from MainFunctor.update_variable *)
-      if ((get_bool "exp.volatiles_are_top") && (is_always_unknown x)) then
-        {st with cpa = CPA.add x (VD.top ()) st.cpa}
```
michael-schwarz, line 145: Why can these checks be removed?

src/framework/constraints.ml
```diff
@@ -81,8 +81,18 @@ struct
...
+            (* must still sync for side effects, e.g. old sync-based none privatization soundness in 02-base/51-spawn-special *)
+            let rec sync_ctx =
+              { ctx with
+                ask = (fun (type a) (q: a Queries.t) -> S.query sync_ctx q);
+                local = d;
+                prev_node = Function dummyFunDec;
+              }
+            in
+            (* TODO: more accurate ctx? *)
+            ignore (sync sync_ctx)
```
michael-schwarz, line 95: Is this now still necessary with only `None3` around?

src/analyses/basePriv.ml
sim642: That's a good question! None of the other privatizations actually do anything special for `exp.volatiles_are_top` either. It could be that the option has been broken for years now anyway.

If that's the case and we want to fix it (or remove it), then I'd say it's a separate matter to be dealt with. Either way, I don't think this is something that every privatization needs to handle itself, but that should be somewhere higher.

src/framework/constraints.ml
```diff
@@ -81,8 +81,18 @@ struct
             ignore (getl (Function fd, c))
           | exception Not_found ->
             (* unknown function *)
-            M.error ~category:Imprecise ~tags:[Category Unsound] "Created a thread from unknown function %s" f.vname
+            M.error ~category:Imprecise ~tags:[Category Unsound] "Created a thread from unknown function %s" f.vname;
             (* actual implementation (e.g. invalidation) is done by threadenter *)
+            (* must still sync for side effects, e.g. old sync-based none privatization soundness in 02-base/51-spawn-special *)
```
michael-schwarz, line 86:
```suggestion
            (* must still sync for side effects, e.g., old sync-based none privatization soundness in 02-base/51-spawn-special *)
```

src/analyses/basePriv.ml
michael-schwarz: Since we're touching this right now, it would be good to leave it in for now and maybe make an issue to check what the others do?

sim642: I added it back. I thought about opening a new issue,  but then found #810 which can be considered to also include these matters.

vesalvojdani (approved):
Good, finally the morally correct way to do things has been implemented.

PR #1508, opened 2024-06-12
Introduce sync reason `JoinCall` & Limit `sync` at function start nodes

michael-schwarz:
Helps avoid "sync" calls for function start nodes when threadflag is context-sensitive. Previously, it was only avoided when the flag was path-sensitive (c.f. #1475).

This has some immediate termination advantages, and can potentially also boost precision.

TODO:

- [x] Investigate what happens if `threadflag` is path-sensitive but not context-sensitive? (It seems sound, but this notion may require re-thinking in general #1509 )
- [x] Investigate interplay with https://github.com/goblint/analyzer/pull/1340 (Does the check have to be dynamic?)

Closes #1507

PR #1570, opened 2024-09-20
Context Gas per Function ⛽

### LINKED ISSUE #1569: Provide per-fundec context gas
For the extended version of the rabbit hole paper, we want to experiment with providing a lower context gas that is instead tracked per fundec.

In this way, only calls to functions for which the gas has been exhausted are analyzed without context, but any function called from such functions will be analyzed context-sensitively again to limit the precision loss in cases where it is not required for termination.

michael-schwarz:
This provides an alternative version of the context gas where it is tracked per function.

- The gas lifter now takes another module that specifies how the gas is handled (allows for further instances)
- The lattice for the gas is now the chain lattice of the actually desired height rather than always having infinite height

I only have two regression tests so far, it would be great if someone else could come up with some more tests before we merge this and run experiments.

Closes #1569


src/framework/constraints.ml
```diff
@@ -507,16 +507,54 @@ struct
...
+  let callee_gas f v = max 0 (v - 1)
+  let thread_gas f v = max 0 (v - 1)
+end
+
+module PerFunctionGas(GasVal:GasVal):Gas = struct
+  module G = Lattice.Chain (struct include GasVal let names x = Format.asprintf "%d" x end)
+  module M = MapDomain.MapTop_LiftBot(CilType.Fundec)(G)
+  let startgas () = M.empty ()
+  let is_exhausted f v = GobOption.exists (fun g -> g <= 0) (M.find_opt f v)  (* v <= 0 *)
+  let is_any_exhausted v = M.exists (fun _ g -> g <=0) v
```
jerhard, line 541: For programs with many functions, this may become a bit expensive if called often. How often will this be called, given that it is used in the query?

Maybe adding an entry to the data structure that maintains the information whether any of the entries is `0` could be usedul?

src/framework/control.ml
```diff
@@ -20,10 +20,16 @@ let spec_module: (module Spec) Lazy.t = lazy (
...
+    (* Chain lattice has elements [0,n-1], but we want [0,gas_value] *)
+    let n () = get_int "ana.context.gas_value" + 1
+  end
+  in
   let module S1 =
     (val
       (module MCP.MCP2 : Spec)
-      |> lift (get_int "ana.context.gas_value" >= 0) (module ContextGasLifter)
+      |> lift (get_int "ana.context.gas_value" >= 0 && get_string "ana.context.gas_scope" = "global") (module ContextGasLifter(GlobalGas(GasVal)))
+      |> lift (get_int "ana.context.gas_value" >= 0 && get_string "ana.context.gas_scope" = "function") (module ContextGasLifter(PerFunctionGas(GasVal)))
```
jerhard, line 32: One might be able to get rid of a bit of redundancy here by extracting a function that checks the `ana.context.gas_scope` configuration and returns the appropriate module.

jerhard (commented):
Looks mostly good, I left some small remarks.

src/framework/constraints.ml
```diff
@@ -549,37 +587,36 @@ struct
...
+    let liftmap d = List.map (fun (x) -> (x, Gas.thread_gas f (cg_val ctx))) d in
     liftmap (S.threadenter (conv ctx) ~multiple lval f args)
 
   let query ctx (type a) (q: a Queries.t):a Queries.result =
     match q with
     | Queries.GasExhausted ->
+      (* The query is only used in a way where overapproximating gas exhaustion is not harmful *)
       let (d,i) = ctx.local in
-      (i <= 0)
+      Gas.is_any_exhausted i
```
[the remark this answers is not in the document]
michael-schwarz: The need for this query is explained here: https://github.com/goblint/analyzer/pull/1508 and https://github.com/goblint/analyzer/pull/1475.

The idea is to avoid calls to `sync` for start nodes of functions when the `threadflag` is context-sensitive, i.e., the abstract value at a start point is never given as the join of things in ST and MT mode (which requires handling similar to entering multi-threaded mode to be sound for some privatizations). Normally, this can be decided just by whether `threadflag` is not configured to be context-insensitive, but the gas may make this analysis context-insensitive even if it is not in general, which is why this query is needed.

The place where this query is asked is here:

https://github.com/goblint/analyzer/blob/3c0ff8c927558a574c7454c6d08d9e8b445d1ebe/src/analyses/commonPriv.ml#L63-L80

So it is asked when the threadflag is context sensitive. The `Gas.is_any_exhausted` is only called when the lifter is actually active.

https://github.com/goblint/analyzer/blob/3c0ff8c927558a574c7454c6d08d9e8b445d1ebe/src/analyses/mCP.ml#L321-L327


This function is called by 3 relational and 3 non-relational privatizations when a sync edge with reason `JoinCall` is encountered to determine whether its handling of `branched_sync` needs to be called.

https://github.com/goblint/analyzer/blob/3c0ff8c927558a574c7454c6d08d9e8b445d1ebe/src/analyses/basePriv.ml#L433-L437


That call originates here:

https://github.com/goblint/analyzer/blob/3c0ff8c927558a574c7454c6d08d9e8b445d1ebe/src/framework/constraints.ml#L668-L674

It is thus called once per function call.

src/framework/constraints.ml
michael-schwarz: I left a longer comment below, but basically it is called once per function call, provided the gas is active, and 
threadflag is normally context-sensitive unless gas is exhausted.

The data structure to memoize this would need to be quite complicated as the join may replenish the gas for any of these functions, so I am not sure if that is worth it.

However, I just realized it may be possible to localize this by giving he `JoinCall` another argument.

src/framework/constraints.ml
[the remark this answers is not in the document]
michael-schwarz: The comments are still true, but by giving `JoinCall` the `fundec` as an argument and also giving the query the argument, no over-approximation is needed anymore!

src/framework/constraints.ml
michael-schwarz: I have removed this now.

src/framework/control.ml
michael-schwarz: I managed to pull it out, but it was a bit of a fight, mainly because I wanted to still avoid constructing the module eagerly.
I'm not sure if this is the most elegant way, I'm not intricately familiar with first-class modules, feel free to suggest a more concise way to do this.

CC also @sim642: I think out of the Goblint team, you're probably the most familiar with first-class modules.

src/framework/control.ml
```diff
@@ -18,12 +18,23 @@ let spec_module: (module Spec) Lazy.t = lazy (
   let arg_enabled = get_bool "witness.graphml.enabled" || get_bool "exp.arg.enabled" in
   let termination_enabled = List.mem "termination" (get_string_list "ana.activated") in (* check if loop termination analysis is enabled*)
   let open Batteries in
+  let gas () =
```
jerhard, line 21: Maybe this could be moved to some of the files defining the gas?

michael-schwarz: I now succeeded moving almost everything into `constraints.ml`. Is this cleaner now?

michael-schwarz: Actually I even tried having a new file for this now. Maybe this is helpful to not make `constraints.ml` too large?

sim642: Indeed, not dumping everything into `Constraints` is a good idea.

src/framework/contextGasLifter.ml
```diff
@@ -0,0 +1,141 @@
...
+struct
+  include S
+
+  module Context_Gas_Prod (Base1: Lattice.S) (Base2: Lattice.S) =
+  struct
+    include Lattice.Prod (Base1) (Base2)
+    let printXml f (x,y) =
+      BatPrintf.fprintf f "\n%a<analysis name=\"context gas value\">\n%a\n</analysis>" Base1.printXml x Base2.printXml y
+  end
+  module D = Context_Gas_Prod (S.D) (Gas.M) (* Product of S.D and an integer, tracking the context gas value *)
```
SchiJoha, line 37: The comment needs to be updated here

src/framework/contextGasLifter.ml
```diff
@@ -0,0 +1,141 @@
...
+  let skip ctx                                    = S.skip (conv ctx), cg_val ctx
+  let special ctx r f args                        = S.special (conv ctx) r f args, cg_val ctx
+  let combine_env ctx r fe f args fc es f_ask     = S.combine_env (conv ctx) r fe f args (Option.bind fc Fun.id) (fst es) f_ask, cg_val ctx
+  let combine_assign ctx r fe f args fc es f_ask  = S.combine_assign (conv ctx) r fe f args (Option.bind fc Fun.id) (fst es) f_ask, cg_val ctx
+  let paths_as_set ctx                            = List.map (fun (x) -> (x, cg_val ctx)) @@ S.paths_as_set (conv ctx)
+  let threadspawn ctx ~multiple lval f args fctx  = S.threadspawn (conv ctx) ~multiple lval f args (conv fctx), cg_val ctx
+  let event ctx e octx                            = S.event (conv ctx) e (conv octx), cg_val ctx
+end
+
+let get_gas_lifter () =
```
SchiJoha, line 104: Nice idea 👍🏼

SchiJoha (approved):
I looked through the changes from the Lifter, and everything looks good :)

jerhard (approved):

src/framework/contextGasLifter.ml
```diff
@@ -0,0 +1,141 @@
...
+          and module G = S.G
+=
+struct
+  include S
+
+  module Context_Gas_Prod (Base1: Lattice.S) (Base2: Lattice.S) =
+  struct
+    include Lattice.Prod (Base1) (Base2)
+    let printXml f (x,y) =
+      BatPrintf.fprintf f "\n%a<analysis name=\"context gas value\">\n%a\n</analysis>" Base1.printXml x Base2.printXml y
```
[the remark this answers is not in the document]
michael-schwarz: As this always shows the entire state, this should be easy to read off as in one case it is a map from function to gas and in the other case it is just a number.

src/framework/contextGasLifter.ml
michael-schwarz: Yes! Thanks for spotting it!

PR #1573, opened 2024-09-23
Improve use of `Lattice.BotValue` and `Lattice.TopValue`

### LINKED ISSUE #1572: Handle `Lattice.BotValue` in `Lattice.LiftBot` and similar
This was wondered in https://github.com/goblint/analyzer/pull/1485#discussion_r1714983146.
We already have special `BotValue` exception that can indicate such thing, but oddly this is not properly lifted by the `LiftBot` functor.

sim642:
Closes #1572.

Also removes `Lattice.Unsupported` by replacing usages with the more specific `BotValue` or `TopValue`.

There's also `Lattice.Uncomparable` which seems similar:
https://github.com/goblint/analyzer/blob/52817d651f1de47dee935925952ad55e277013db/src/domain/lattice.ml#L14-L15
But #1109 seems to use it with some slightly different meaning.

### DIFF
--- a/src/domain/lattice.ml
+++ b/src/domain/lattice.ml
@@ -93,10 +93,10 @@ struct
   include Base
   let leq = equal
   let join x y =
-    if equal x y then x else raise (Unsupported "fake join")
+    if equal x y then x else raise (Unsupported "fake join") (* TODO: TopValue? *)
   let widen = join
   let meet x y =
-    if equal x y then x else raise (Unsupported "fake meet")
+    if equal x y then x else raise (Unsupported "fake meet") (* TODO: BotValue? *)
   let narrow = meet
   include NoBotTop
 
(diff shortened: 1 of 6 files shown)


src/domain/lattice.ml
```diff
@@ -259,24 +256,36 @@ struct
     | (_, `Top) -> `Top
     | (`Bot, x) -> x
     | (x, `Bot) -> x
-    | (`Lifted x, `Lifted y) -> `Lifted (Base.join x y)
+    | (`Lifted x, `Lifted y) ->
+      try `Lifted (Base.join x y)
+      with TopValue -> `Top
```
michael-schwarz, line 261: Earlier we would have raised here. Is swallowing this here the right choice? Also applies everywhere below.

sim642: This shouldn't change any current behavior because currently we would have an uncaught `TopValue` exception, which I haven't seen anywhere.

The point of lifting a domain without a top to have a top is to be able to do joins without getting an exception. Otherwise every use case has to manually reimplement one of these domain lifters into the domain itself, like in https://github.com/goblint/analyzer/pull/1485#discussion_r1714983146.

michael-schwarz (approved):

PR #1575, opened 2024-09-24
Move `Spec` lifters out of `Constraints`

sim642:
They are moved to a new `lifters` directory.

The splitting is done such that `git blame` history is preserved for default git settings. Thus the git graph is a bit of a braid.


michael-schwarz:
Nice! I think this is a really good step towards making modules shorter and easier to follow!

michael-schwarz (approved):

PR #1579, opened 2024-09-27
Add "ERROR (both branches dead)" verdict for SV-COMP

### LINKED ISSUE #1576: `ERROR` verdict for both branches dead in SV-COMP
Apparently there are SV-COMP tasks where we output the "both branches dead" unsoundness warning, but still happily give a `true` SV-COMP result. We've been lucky that this hasn't caused unsound verdicts.
In these cases the verdict should instead be something like `ERROR (both branches dead)`.

We do this for fixpoint errors with `ERROR (fixpoint)`.

sim642:
Closes #1576.

### DIFF
--- a/src/common/framework/analysisState.ml
+++ b/src/common/framework/analysisState.ml
@@ -36,3 +36,8 @@ let postsolving = ref false
 
 (* None if verification is disabled, Some true if verification succeeded, Some false if verification failed *)
 let verified : bool option ref = ref None
+
+let unsound_both_branches_dead: bool option ref = ref None
+(** [Some true] if unsound both branches dead occurs in analysis results.
+    [Some false] if it doesn't occur.
+    [None] if [ana.dead-code.branches] option is disabled and this isn't checked. *)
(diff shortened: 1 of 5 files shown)


michael-schwarz:
Before merging this we should verify that it is not just because the condition transfer functions are quite strong that we end up with this warning because the predecessor is in fact unreachable.

sim642:
We should have this behavior either way independently of any particular task if we want to be serious about soundness.
If the both branches dead warning is triggered too generously, that's a separate precision problem for that.

michael-schwarz (commented):
LGTM! We should keep a keen lookout for the instances of this.

I'm still wondering if if we can devise some way to not raise this when it's simply to imprecision of domains (Maybe don't raise it for Polyhedra at all now?)

sim642:
I tried forcing Apron to somehow reduce integer constraints in #1585 to avoid the spurious cautiousness but none of it is particularly nice.
We should see from (our own) pre-runs how often this actually comes up on sv-benchmarks and then decide what to do about these.

PR #1580, opened 2024-09-27
Add termination analysis success messages for loop bounds

### LINKED ISSUE #1577: Success messages with bounds from termination analysis
Our race analysis can output "proofs" as success messages with `allglobs` enabled. This is useful for inspecting why Goblint believes the program to be correct.

We could do something similar for termination analysis: when we manage to prove a loop to be bounded and thus terminate, optionally output a success message with the bound (at least non-relationally, if not relationally). @karoliineh looked at some SV-COMP termination program where termination wasn't obvious because the loop could both increment (by 2) or decrement (by 1) the counter based on some condition.

sim642:
Closes #1577.

These are disabled by default, but can be enabled with `dbg.termination-bounds`.

Also cleans up the implementation of `LoopTermination.special` in general (pointless `else ()`, too wide `try` scope, annoying half-indentation).


sim642:
Although this doesn't really explain the case from #1577:
```console
./goblint --conf conf/svcomp.json --set ana.specification ../sv-benchmarks/c/properties/termination.prp ../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c --html --enable dbg.termination-bounds
```
outputs (among other things)
```
[Success][Termination] Loop terminates: bounded by 0 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 1 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 2 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 3 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 4 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 5 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 6 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 7 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 8 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by 9 iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
[Success][Termination] Loop terminates: bounded by (Not {0, 1}([0,32]),[10,2147483896],not {0} ([0,32]),ℤ) iteration(s) (../sv-benchmarks/c/termination-crafted-lit/KroeningSharyginaTsitovichWintersteiger-CAV2010-Fig1.c:17:9-17:16)
```

The maximum bound 2147483896 is one lower, but that's just because we activate congruence analysis we can rule out 2147483897 (which is is excluded because it is odd).
And we don't go to top because we assume no overflows, so 2147483896 + 2 doesn't overflow to top. But if it were to be a loop incrementing by one, this could go up to 2147483897 and we consider it unbounded. So I'm a bit confused by the termination checking logic with this.

jerhard:
Regarding our discussion at today's Gobcon, whether it suffices to exclude that the abstract value for the artificially introduced loop counter variable is top: As for an artificial loop counter `i` it is ensured that `i` is incremented by one in each iteration, if the loop is ounbounded, `i` will receive every possible value (even if any finite number of unrollings is performed). 

As a side note, the artificial loop counter should have an `unsigned` integer type, and thus should not be affected by the analyzer assuming that signed overflows do not happen.

hseidl:
… but it should never wrap around, though :-)     HelmutAm 29.10.2024 um 14:59 schrieb Julian Erhard ***@***.***>:﻿
Regarding our discussion at today's Gobcon, whether it suffices to exclude that the abstract value for the artificially introduced loop counter variable is top: As for an artificial loop counter i it is ensured that i is incremented by one in each iteration, if the loop is ounbounded, i will receive every possible value (even if any finite number of unrollings is performed).
As a side note, the artificial loop counter should have an unsigned integer type, and thus should not be affected by the analyzer assuming that signed overflows do not happen.

src/analyses/loopTermination.ml
```diff
@@ -8,12 +8,13 @@ open TerminationPreprocessing
...
+let ask_bound ctx varinfo =
   let open IntDomain.IntDomTuple in
   let exp = Lval (Var varinfo, NoOffset) in
   match ctx.ask (EvalInt exp) with
-  | `Top -> false
-  | `Lifted v -> not (is_top_of (ikind v) v)
+  | `Top -> `Top
+  | `Lifted v when is_top_of (ikind v) v -> `Top
+  | `Lifted v -> `Lifted v
   | `Bot -> failwith "Loop counter variable is Bot."
```
michael-schwarz, line 18: It looks like this could be moved inside the `special`?

michael-schwarz (approved):
That will hopefully make debugging easier in the future. Nice addition!

src/analyses/loopTermination.ml
sim642: It could be, but `special` already has very deep matching, which wouldn't make it nicer. Having it at the top level is like `ThreadEscape.has_escaped`, although currently no other analysis needs it.

PR #1582, opened 2024-09-27
Refactor loopUnrolling autotuner

karoliineh:
Refactor some code in `loopUnrolling.ml`, simplifies the following:
- Use Log instead of printing to stderr
- Improve debug messages and simplify their syntax
- Make sure the goal (bound) of the loop is found from the exact same loop statement comparison expression where the var was found
- Move functions defined in `fixedLoopSize` to top level to make function definitions' style consistent
- Use `let*` option monad syntax from `GobOption.Syntax` in LoopUnrolling


sim642 (approved):
The conflict with master now needs resolving too.

michael-schwarz (approved):

PR #1589, opened 2024-10-02
Exclude multithreaded main-main races by threadflag

sim642:
This is a bit silly but the threadflag analysis `may_race` did not exclude `main`-`main` races even though it has all the information to do so.
It almost never makes a difference because threadflag answers the `MustBeUniqueThread` query based on its flag still, and the threadid analysis uses that information in its `may_race`.

The more notable thing is the unexpected cram test (change). Usually accesses while single-threaded aren't even added to `allglobs`, but apparently the thread ID variable write in `pthread_create` is added although as if single-threaded. This is odd for two reasons:
1. As single-threaded it shouldn't even be in `allglobs`.
2. It's actually ambiguous whether it should be written before or after the thread is created. Maybe it should be done after, thus in multi-threaded mode altogether to over-approximate the possibility.

This was always in the cram test but went unnoticed until I played around with this.

### DIFF
--- a/src/analyses/threadFlag.ml
+++ b/src/analyses/threadFlag.ml
@@ -54,13 +54,15 @@ struct
 
   module A =
   struct
-    include BoolDomain.Bool
-    let name () = "multi"
-    let may_race m1 m2 = m1 && m2 (* kill access when single threaded *)
-    let should_print m = not m
+    include Flag
+    let name () = "flag"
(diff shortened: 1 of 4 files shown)


michael-schwarz:
Any reason this is still a draft? We should be able to merge to reduce the number of in-flight PRs, right?

PR #1590, opened 2024-10-02
Handle loop statement comparison between two variables in loop unrolling

karoliineh:
Previously, we only detected fixed-sized loops when the loop comparison was between a variable and a constant. However, we already found the last assignment to the variable being changed in the loop body, and thus, we have a function for finding the last assigned value.

This PR adds the functionality also to detect fixed loops when the loop condition compares variables by finding the last assignment for the other compared variable.

This PR was inspired by the following loop from the SV-COMP task `loops/invert_string-3.c`, which we were lucky to solve based on the old heuristics but cannot solve anymore due to the new, more conservative heuristic.

```c
unsigned int max = 5;
char str1[max], str2[max];
int i, j;

for (i=0; i<max; i++) {
    str1[i]=__VERIFIER_nondet_char();
}
```

This task has the case including a cast. I added another case that would handle similar conditions without a cast as a precaution.

### DIFF
--- a/src/util/loopUnrolling.ml
+++ b/src/util/loopUnrolling.ml
@@ -220,13 +220,21 @@ let findBreakComparison loopStatement =
   with WrongOrMultiple ->
     None
 
-let getLoopVar = function
+let getLoopVar loopStatement func = function
   | BinOp (op, (Const (CInt (goal, _, _) )), Lval ((Var varinfo), NoOffset), (TInt _)) when isCompare op && not varinfo.vglob ->
     (* TODO: define isCompare and flip in cilfacade and refactor to use instead of the many separately defined similar functions *)
     let flip = function | Lt -> Gt | Gt -> Lt | Ge -> Le | Le -> Ge | s -> s in
     Some (flip op, varinfo, goal)
   | BinOp (op, Lval ((Var varinfo), NoOffset), (Const (CInt (goal, _, _) )), (TInt _)) when isCompare op && not varinfo.vglob ->
     Some (op, varinfo, goal)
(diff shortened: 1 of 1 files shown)


src/util/loopUnrolling.ml
```diff
@@ -220,13 +220,21 @@ let findBreakComparison loopStatement =
...
-let getLoopVar = function
+let getLoopVar loopStatement func = function
   | BinOp (op, (Const (CInt (goal, _, _) )), Lval ((Var varinfo), NoOffset), (TInt _)) when isCompare op && not varinfo.vglob ->
     (* TODO: define isCompare and flip in cilfacade and refactor to use instead of the many separately defined similar functions *)
     let flip = function | Lt -> Gt | Gt -> Lt | Ge -> Le | Le -> Ge | s -> s in
     Some (flip op, varinfo, goal)
   | BinOp (op, Lval ((Var varinfo), NoOffset), (Const (CInt (goal, _, _) )), (TInt _)) when isCompare op && not varinfo.vglob ->
     Some (op, varinfo, goal)
+  (* When loop condition has a comparison between variables, we assume that the left one is the counter and right one is the bound.
+     TODO: can we do something more meaningful instead of this assumption? *)
```
sim642, line 231: Theoretically it would be possible to lift this from the `option` monad to the `list` monad for non-determinism and return both here. Then the following checks `getsPointedAt`, `constBefore`, etc would filter it down to the one that meets all the conditions (although I guess both could still pass all of them).
It would automatically achieve backtracking instead of what currently happens: if the first one doesn't pass `getPointedAt`, we don't retry everything with the second variable.

I don't know if it would be worth the trouble though.

sim642 (approved):

PR #1595, opened 2024-10-08
Restrict unrolling to loops that are nested within less than 4 loops

karoliineh:
This PR addresses the excessive resource usage introduced in [bb6f9aa](https://github.com/goblint/analyzer/compare/060004c...bb6f9aa).

As the amount of unrolling grows exponentially with the number of nested loops, we determined what would be a reasonable amount of nestings with a 5s run on the sv-benchmarks and found that unrolling less than 4 loops is a reasonable metric.

On the downside, this PR introduces an issue where some termination tasks do not reach a fixpoint. Unfortunately, I was not able to trace it to the cause, but we made a quick run, and it seems to not impact our overall scores (compared to SV-COMP 2023, we lost 2 termination tasks, but we also currently run out of resources for these).

So, (hopefully) this PR fixes more things than it breaks. *Edit*: we made a run and, indeed, this PR does not break anything else but fixes the excessive resource usage with 717 tasks going from errors back to unknowns.

### DIFF
--- a/src/util/loopUnrolling.ml
+++ b/src/util/loopUnrolling.ml
@@ -402,16 +402,19 @@ let copy_and_patch_labels break_target current_continue_target stmts =
   let patchLabelsVisitor = new patchLabelsGotosVisitor(StatementHashTable.find_opt gotos) in
   List.map (visitCilStmt patchLabelsVisitor) stmts'
 
-class loopUnrollingVisitor(func, totalLoops) = object
+class loopUnrollingVisitor (func, totalLoops) = object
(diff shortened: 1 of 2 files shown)


sim642 (approved):

sim642:
> On the downside, this PR introduces an issue where some termination tasks do not reach a fixpoint

For reference, in SV-COMP 2024 we also had a bunch of fixpoint errors: https://sv-comp.sosy-lab.org/2024/results/results-verified/goblint.results.SV-COMP24.table.html#/table?filter=0(0*status*(status(in(ERROR%20%28verify%29)))).
Most of them are also with the termination property, so there's probably some preexisting underlying issue that less unrolling also triggers.

PR #1598, opened 2024-10-14
⛽: Use minimum to combine incoming gas values in per-context gas

michael-schwarz:
As observed by @jerhard for our extended version of the context paper, the gas (if tracked per `fundec`) needs to be combined with the  minimum operation as otherwise it can be replenished unboundedly in the case of several mutually recursive functions.

This PR adds an example that was previously problematic, and fixes the issue.


jerhard (approved):
Looks good!

PR #1599, opened 2024-10-15
 Improve loop bound detection for loop unrolling by handling casts in loop statements

karoliineh:
Inspired by an `unreach-call` task `num_conversion_1.c` where we couldn't detect a bound for the following loop:

```c
unsigned char c;
c = 0;
while (c < (unsigned char) 8) {
  ...
}
```

Unrolling the loop 8 times here helps us solve the task. In addition to the case in this program, where the loop statement is in the form `CastE(TInt(int, ), Lval(Var(c, NoOffset))) < Const(Int(8,int,None))` I included all possible combinations of casting for both the left and right expressions in the loop comparison statement.

### DIFF
--- a/src/util/loopUnrolling.ml
+++ b/src/util/loopUnrolling.ml
@@ -243,28 +243,25 @@ let findBreakComparison loopStatement =
   with WrongOrMultiple ->
     None
 
-let getLoopVar loopStatement func = function
-  | BinOp (op, Const (CInt (goal, _, _)), Lval (Var varinfo, NoOffset), TInt _)
-  | BinOp (op, Const (CInt (goal, _, _)), CastE (TInt _, Lval (Var varinfo, NoOffset)), TInt _)
-  | BinOp (op, CastE (TInt _, Const (CInt (goal, _, _))), Lval (Var varinfo, NoOffset), TInt _)
-  | BinOp (op, CastE (TInt _, Const (CInt (goal, _, _))), CastE (TInt _, Lval (Var varinfo, NoOffset)), TInt _) when isCompare op && not varinfo.vglob ->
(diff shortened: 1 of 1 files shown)


src/util/loopUnrolling.ml
```diff
@@ -244,16 +244,23 @@ let findBreakComparison loopStatement =
     None
 
 let getLoopVar loopStatement func = function
-  | BinOp (op, (Const (CInt (goal, _, _) )), Lval ((Var varinfo), NoOffset), (TInt _)) when isCompare op && not varinfo.vglob ->
+  | BinOp (op, Const (CInt (goal, _, _)), Lval (Var varinfo, NoOffset), TInt _)
+  | BinOp (op, Const (CInt (goal, _, _)), CastE (TInt _, Lval (Var varinfo, NoOffset)), TInt _)
+  | BinOp (op, CastE (TInt _, Const (CInt (goal, _, _))), Lval (Var varinfo, NoOffset), TInt _)
+  | BinOp (op, CastE (TInt _, Const (CInt (goal, _, _))), CastE (TInt _, Lval (Var varinfo, NoOffset)), TInt _) when isCompare op && not varinfo.vglob ->
```
sim642, line 250: This is getting out of hand, so we should refactor this a bit. For example:
1. First match the `BinOp` (with comparison operator and integer type) to just get the two operand expressions.
2. In that case, then apply `Cil.stripCasts` to both of them.
3. In that case, then match on the pair of stripped operand expressions.

([ppx_viewpattern](https://github.com/sim642/ppx_viewpattern/) could simplify this, but I'm not sure if that's production-quality.)

sim642 (changes requested):

sim642 (approved):

PR #1605, opened 2024-10-25
Sanity test all solvers and fix `topdown_term`

sim642:
We've kept various solvers around and compiling, but they're untested. For example, it turns out that `topdown_term` is completely broken and only computes bottoms.

### Changes
1. Adds cram test for 00-sanity/01-assert to check minimal functioning of all solvers.
2. Fix start variables in `topdown_term` to restore basic functionality.
3. Fix invalid widen call in `topdown_term` for side effects.
4. Fix option name in `topdown_space_cache_term`.


src/solver/sLR.ml
```diff
@@ -66,7 +66,7 @@ module SLR3 =
           if tracing then trace "sol" "Contrib:%a" S.Dom.pretty tmp;
           let tmp =
             if wpx then
-              if HM.mem globals x then S.Dom.widen old tmp
+              if HM.mem globals x then S.Dom.widen old tmp (* TODO: no join in second argument, can call widen incorrectly? *)
```
michael-schwarz, line 69: Should we not fix this rather than produce TODOs?

src/solver/sLR.ml
```diff
@@ -527,7 +527,7 @@ let _ =
   Selector.add_solver ("widen2",  (module PostSolver.EqIncrSolverFromEqSolver (W2)));
   Selector.add_solver ("widen3",  (module PostSolver.EqIncrSolverFromEqSolver (W3)));
   let module S2 = TwoPhased (struct let ver = 1 end) in
-  Selector.add_solver ("two",  (module PostSolver.EqIncrSolverFromEqSolver (S2)));
+  Selector.add_solver ("two",  (module PostSolver.EqIncrSolverFromEqSolver (S2))); (* TODO: broken even on 00-sanity/01-assert *)
```
michael-schwarz, line 530: Broken how? if it doesn't work at all, we should remove it --- no point in keeping broken stuff around!

tests/regression/00-sanity/01-assert.t
```diff
@@ -9,3 +9,390 @@
...
+  [Warning][Deadcode] Function 'main' does not return
+  [Warning][Deadcode] Function 'main' has dead code:
+    on lines 13..14 (01-assert.c:13-14)
+  [Warning][Deadcode] Logical lines of code (LLoC) summary:
+    live: 7
+    dead: 2
+    total lines: 9
+
+  $ goblint --enable warn.deterministic --set solver two 01-assert.c
+  [Error] Fixpoint not reached at L:entry state of main (299) on 01-assert.c:4:1-15:1
```
michael-schwarz, line 143: This is an indication that we should remove `two` or fix it. Testing for it remaining broken in this way is not that useful...

src/solver/sLR.ml
sim642: I wasn't sure if SLR3 maybe ensures it somehow indirectly when handling side effects per origin. I now ran all tests with `slr3` and it triggered some invalid widening calls, so I made the fix now.

src/solver/sLR.ml
sim642: Broken as it computes bottom everywhere, like `topdown_term`. The latter I managed to easily fix here, but I don't know the SLR solvers well enough to tell if there's a similar silly issue somewhere.
It's the only SLR variation with such issue (and most share code), so there's not much to remove code-wise. I commented this `add_solver` out though, so Goblint cannot be executed with this solver.

michael-schwarz (approved):

PR #1606, opened 2024-10-25
Split `IntDomain`s into separate files

sim642:
Similarly to #1575 because `intDomain.ml` is the largest file we have.

They are moved to a new `int` subdirectory. There's also a `IntDomain0` with all the various interfaces and their adapters.
The splitting is done such that git blame history is preserved for default git settings. Thus the git graph is a bit of a braid.


michael-schwarz (approved):
Do we need to wait on anything here, or should we just go ahead and merge before it gets stale?

Adapting other PRs to this is probably easier than recreating this with all the history preservation things.

PR #1607, opened 2024-10-28
Add `pthread_self` support

sim642:
The added test is an obscure program that uses `pthread_self` to join the main thread from a secondary thread and avoid a data race.

By completing a TODO it partially even improves precision with plain thread IDs.

### DIFF
--- a/src/cdomain/value/cdomains/threadIdDomain.ml
+++ b/src/cdomain/value/cdomains/threadIdDomain.ml
@@ -83,7 +83,7 @@ struct
       (v, None)
 
   let is_main = function
-    | ({vname; _}, None) -> List.mem vname @@ GobConfig.get_string_list "mainfun"
+    | ({vname; _}, None) -> GobConfig.get_bool "ana.thread.include-node" && List.mem vname @@ GobConfig.get_string_list "mainfun"
     | _ -> false
 
   let is_unique = is_main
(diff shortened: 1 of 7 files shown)


src/cdomain/value/cdomains/threadIdDomain.ml
```diff
@@ -86,7 +86,7 @@ struct
     | ({vname; _}, None) -> List.mem vname @@ GobConfig.get_string_list "mainfun"
     | _ -> false
 
-  let is_unique _ = false (* TODO: should this consider main unique? *)
+  let is_unique = is_main
```
michael-schwarz, line 89: Why is this sound? What if multiple threads are created from the main function is created multiple times?

michael-schwarz (commented):
Nice added feature! I am still skeptical about considering `main` unique, but other than that it looks good!

src/cdomain/value/cdomains/threadIdDomain.ml
sim642: Oh, good point!

I now added an extra condition to the `is_main` function based on `ana.thread.include-node`. If nodes are included, then only the actual root thread has `None` (others have a node where they're created, even if it's the same `main` function). if nodes are not included, then we cannot tell them apart indeed and it should be non-unique.

michael-schwarz (approved):

Issue #1168, opened 2023-09-14
Re-evaluate defaults for `ana.malloc.unique_address_count` in SV-Comp


michael-schwarz:

sim642:
This was changed to 5 in #1234, so I suppose we can consider it done for now?

michael-schwarz:
This was only changed for MemSafety, one might want to re-evaluate it also for the other categories.

karoliineh:
@sim642 made a run with `--set ana.malloc.unique_address_count 1` on sv-benchmarks with no-overflow property. This gives 31 new correct results with a 60s timeout. However, it costs us a 14% slowdown as well.

So, for the best-case scenario, we should 
- [ ] try to autotune it so that we only set `ana.malloc.unique_address_count > 0` for programs with allocs that are not in loops and see if that helps with the loss in performance;
- [ ] try out different values to see which is the best default for no-overflows;
- [ ] inspect other categories in addition to MemSafety and no-overflows.

However, if we do not manage to do all that before SV-COMP 2025, we could at least enable it with `1` for no-overflows.

sim642:
> However, if we do not manage to do all that before SV-COMP 2025, we could at least enable it with `1` for no-overflows.

This should be a fairly simple change to do now, so feel free to open a PR. I think I still have the 60s no-overflows run diff here: https://goblint.cs.ut.ee/results/191-no-overflow-pr-1511-before-alloc-1/table-generator-cmp.diff.html#/.
We can postpone other additional improvements to SV-COMP 2026 then.

PR #1612, opened 2024-10-31
Autotune for `ana.malloc.unique_address_count` for NoOverflows in SV-Comp

karoliineh:
Relates to #1168

For noOverflows, I find if there are any allocations in the program that are not within loops, and for those programs set `ana.malloc.unique_address_count` as 1 by default.

To enable this, I also separated the previous "specification" option that was autotuning a bunch of stuff and was not enabled due to it not being helpful. Thus, I can enable only the relevant stuff for the noOverflows category.


src/autoTune.ml
```diff
@@ -45,20 +45,22 @@ class functionVisitor(calling, calledBy, argLists, dynamicallyCalled) = object
...
-class findAllocsNotInLoops = object
+class findAllocsInLoops = object
   inherit nopCilVisitor
 
+  val mutable inloop = false
+
   method! vstmt stmt =
     match stmt.skind with
-    | Loop _ -> SkipChildren
+    | Loop _ -> inloop <- true; DoChildren
```
sim642, line 55: `inloop` also needs to be reset after the children are processed (like how loop nesting counter is decremeted). Otherwise allocations after a loop will also have `inloop` `true`.

src/autoTune.ml
```diff
@@ -44,6 +44,33 @@ class functionVisitor(calling, calledBy, argLists, dynamicallyCalled) = object
...
+      | _ -> stmt
+    in
+    match stmt.skind with
+    | Loop _ -> inloop <- true; ChangeDoChildrenPost(stmt, outOfLoop)
+    | _ -> DoChildren
+
+  method! vinst = function
+    | Call (_, Lval (Var f, NoOffset), args,_,_) ->
+      let desc = LibraryFunctions.find f in
+      begin match desc.special args with
```
sim642, line 66: This gives some `exception LibraryDsl.Pattern.Expected("^::")` results on sv-benchmarksL https://goblint.cs.ut.ee/results/235-no-overflow-pr-1612-after-2/table-generator-cmp.diff.html#/table.
This needs to be adapted like #1560.

sim642: I also added a missing `is_special` check.

sim642 (approved):

PR #1617, opened 2024-11-03
Fix `thread` for non-unique spawns

### LINKED ISSUE #1615: `thread` analysis requires context-sensitive threadids
As discovered by @Red-Panda64, the old `thread` analysis assumes we have context-sensitive thread ids activated when doing a `threadenter` which we know to be mutliple (c.f. #1187)

https://github.com/goblint/analyzer/blob/750f1ee274a00d6614cd2141f302ce551b094d57/src/analyses/threadAnalysis.ml#L97-L101

Another call to `ThreadId.get_current_unlift` is in `thread_spawn` for the spawned thread. However, this may be correct as we would still have a non-top thread id  for the created thread at this point (?) 

https://github.com/goblint/analyzer/blame/750f1ee274a00d6614cd2141f302ce551b094d57/src/analyses/threadAnalysis.ml#L103-L113

michael-schwarz:
The handling of `threadenter` in the case of a thread being spawned non-uniquely was broken:

- It failed when thread ids where non-unique.
- It actually caused a side-effect to the thread id of the **creator** marking that as non-unique rather than the created thread.

This PR fixes this issue which hopefully means our experiments for PLDI can continue (CC @Red-Panda64: try if merging this into your branch helps), and also increases precision for cases where the non-history thread id is enabled.

Closes #1615

### DIFF
--- a/src/analyses/threadAnalysis.ml
+++ b/src/analyses/threadAnalysis.ml
@@ -95,9 +95,7 @@ struct
   let startstate v = D.bot ()
 
   let threadenter ctx ~multiple lval f args =
-    if multiple then
-      (let tid = ThreadId.get_current_unlift (Analyses.ask_of_ctx ctx) in
-       ctx.sideg tid (true, TS.bot (), false));
+    (* ctx is of creator, side-effects to denote non-uniqueness are performed in threadspawn *)
     [D.bot ()]
 
   let threadspawn ctx ~multiple lval f args fctx =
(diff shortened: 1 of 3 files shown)


sim642:
The tests seem to be about not crashing and precision, but isn't this also a matter of soundness?

michael-schwarz:
I thought so as well and tried to create a new test that would fail before but would work now. However, marking the creator as non-unique also leads to the created thread being marked as such (through the recursion in `is_not_unique`).
So it seems like it actually is only a precision issue vis à vis the old implementation. 

Nevertheless, the original soundness tests that were added in #1187 are still around and pass.

tests/regression/40-threadid/12-multiple-created-only.c
sim642, line 1: I think 10-synch would be a better place for these tests. 40-threadid has tests about history TIDs, while 10-synch has all kinds of tests for the old thread analysis what this is really about.

src/analyses/threadAnalysis.ml
```diff
@@ -106,9 +104,9 @@ struct
     let repeated = D.mem tid ctx.local in
     let eff =
       match creator with
-      | `Lifted ctid -> (repeated, TS.singleton ctid, false)
-      | `Top         -> (true,     TS.bot (),         false)
-      | `Bot         -> (false,    TS.bot (),         false)
+      | `Lifted ctid -> (repeated || multiple, TS.singleton ctid, false)
+      | `Top         -> (true, TS.bot (), false)
+      | `Bot         -> (false || multiple, TS.bot (), false)
```
sim642, line 109:
```suggestion
      | `Bot         -> (multiple, TS.bot (), false)
```

sim642 (approved):

Issue #1564, opened 2024-09-05
Parsing Error


pvouzis:
I am getting the following parsing error. Below is my configuration.

Any idea what might be going on?

This is installed on ubuntu 22.04

```
/home/ubuntu/goblint-2.4.0/goblint compile_commands.json 
/usr/include/glib-2.0/glib/gspawn.h[76:0-21] : syntax error
Parsing errorFatal error: exception Goblint_lib__Maingoblint.FrontendError("Frontc.ParseError: Parse error")
```

My setup

```
/home/ubuntu/goblint-2.4.0/goblint --version 
Goblint version: 2.4.0 (b129fab7de3ab5fd85d2f1340b2d9c5a4e4e2653)
Cil version:     2.0.4
Dune profile:    dev
OCaml version:   4.14.2
OCaml flambda:   true
Build time:      2024-09-05T14:17:42
```

```
cat compile_commands.json 
[
  {
    "arguments": [
      "/usr/bin/gcc",
      "-Wimplicit",
      "-Wparentheses",
      "-Wsequence-point",
      "-Wreturn-type",
      "-Wcast-qual",
      "-Wall",
      "-Wno-unknown-pragmas",
      "-Wno-long-long",
      "-Wall",
      "-pthread",
      "-I/usr/include/glib-2.0",
      "-I/usr/lib/x86_64-linux-gnu/glib-2.0/include",
      "-O3",
      "-c",
      "-o",
      "executeHTTP.o",
      "executeHTTP.c"
    ],
    "directory": "/home/ubuntu/nbclient/http",
    "file": "/home/ubuntu/nbclient/http/executeHTTP.c",
    "output": "/home/ubuntu/nbclient/http/executeHTTP.o"
  }
]
```

pvouzis:
Then I installed it again from master (not from the release) and I don't get the parsing error, butI get the following error:

```
Fatal error: exception LibraryDsl.Pattern.Expected("Library function is called with more arguments than expected.")
```

Before that there a bunch of warnings, if that helps (these are only the last few before the error):
```
 /home/ubuntu/analyzer-master/goblint compile_commands.json 
gcc: warning: nbagent_main.o: linker input file unused because linking not done
gcc: warning: communication/communication.o: linker input file unused because linking not done
gcc: warning: communication/tcp.o: linker input file unused because linking not done
gcc: warning: communication/websocket.o: linker input file unused because linking not done
[...]

PR #1619, opened 2024-11-04
Print innermost backtrace mark for uncaught exception even with backtrace printing off

### LINKED ISSUE #1616: Error message from library mechanism are unhelpful to users
For the --- admittedly incorrect --- program:

~~~C
// PARAM: --set ana.context.gas_value 0 --set ana.activated[+] thread --set ana.activated[+] threadid

#include <pthread.h>
#include <stdio.h>

void *t_fun(void *arg) {
  return NULL;
}

int main(void) {
  pthread_create(t_fun);
  return 0;
}
~~~

we produce the error message

```
./goblint --enable warn.debug --enable dbg.regression --html --set ana.context.gas_value 0 --set ana.activated[+] thread --set ana.activated[+] threadid  tests/regression/80-context_gas/26-thread.c
Fatal error: exception LibraryDsl.Pattern.Expected("^::")
See result/index.xml
```

We should instead maybe produce some warning detailing that we failed because the arguments of `pthread_create` are strange.

sim642:
Closes #1616.

The output for the problematic example is now
```
Fatal error: exception LibraryDsl.Pattern.Expected("^::")
Marked with function pthread_create
```

It's still not the best output like https://github.com/goblint/analyzer/issues/1616#issuecomment-2454143344. That's the intended output for such problem with the clearest error message, but if the problematic call is the only one in the program, then CIL doesn't seem to be doing some unification with the declaration that it otherwise would.

### DIFF
--- a/src/util/backtrace/goblint_backtrace.ml
+++ b/src/util/backtrace/goblint_backtrace.ml
@@ -76,12 +76,19 @@ let print_marktrace oc e =
       Printf.fprintf oc "Marked with %s\n" (mark_to_string m)
     ) ms
 
+let print_innermost_mark oc e =
+  match find_marks e with
+  | m :: _ -> Printf.fprintf oc "Marked with %s\n" (mark_to_string m)
+  | [] -> ()
+
 let () =
   Printexc.set_uncaught_exception_handler (fun e bt ->
       (* Copied & modified from Printexc.default_uncaught_exception_handler. *)
       Printf.eprintf "Fatal error: exception %s\n" (Printexc.to_string e); (* nosemgrep: print-not-logging *)
       if Printexc.backtrace_status () then
-        print_marktrace stderr e;
+        print_marktrace stderr e
+      else
+        print_innermost_mark stderr e;
       Printexc.print_raw_backtrace stderr bt;
       flush stderr
     )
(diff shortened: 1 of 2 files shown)


michael-schwarz:
Should the error message not also contain some human-readable information such as 
`Check that the input program compiles with GCC. If it does, the specification of the library function in Goblint is likely incorrect or incomplete. When appropriate, you could also try commenting out the offending call.`?

The current error message is only helpful to Goblint developers; such an error message would also give other users a way to bypass this message or understand what's happening.

sim642:
> Should the error message not also contain some human-readable information such as
> `Check that the input program compiles with GCC. If it does, the specification of the library function in Goblint is likely incorrect or incomplete. When appropriate, you could also try commenting out the offending call.`?

I'm not sure we should make exception strings such spoon-feedy text. If we want to be consistent about it, we should change every CIL lexing/parsing/typing/merging/etc error to also to print it. #1564 clearly shows it.
Or should we suggest the user delete their code to avoid Goblint erroring on it?

> such an error message would also give other users a way to bypass this message

Not really. The only they have to be Goblint developers to actually fix `LibraryFunctions`.

michael-schwarz:
Maybe we can find some compromise between my extremely wordy suggestion and the extremely terse message we currently have and then move forward and merge this?

michael-schwarz (approved):
Even if we can't agree on an error message, we should still merge it. It is clearly better than what we currently have.

PR #1623, opened 2024-11-05
Bitfield Domain

ManuelLerchner:
## Overview

This pull request enhances the existing `IntDomain` by introducing a new BitfieldDomain. The full `IntDomain` now consists of `(DefExc, Interval, Enums, Congruence, IntervalSet, Bitfield)`.

The new Bitfield Domain is based on the Paper [`Abstract Domains for Bit-Level Machine Integer and Floating-Point Operations
`](https://doi.org/10.29007/b63g), and keeps track of each bit in an integer individually (conceptually, this is similar to a boolean lattice for each bit).
This allows for very precise analysis of bitwise operations.

The new domain can be enabled with `--enable ana.int.bitfield`

## Example

For example, joining $of\\_int(8) \sqcup of\\_int(10) \equiv 0b1000 \sqcup 0b1010 = 0b10?0$ maintains most information about the bits; only the second bit differs in both numbers and therefore becomes unknown (unknown bits are shown using a question-mark symbol). 
In this case, no over-approximation happens and the join is exact.

Bitwise operators such as `&, |, <<, >>, ~, ...` can now deal with this information very precisely.

## Refinements

As the existing `IntDomain` allows for mutual refinements of the subdomains, this PR also adds all newly possible refinement directions to improve precision in the `IntDomain`.

We did, however, have some problems with non-termination when using refinements.
Some of these issues are however also present on the master branch, and indicate a more fundamental problem with refinements (See #1671).

--- 
> [!NOTE] 
> This implementation is part of the `Automated Bug Hunting and Beyond` practical course at TUM.


src/analyses/baseInvariant.ml
```diff
@@ -395,27 +395,55 @@ struct
...
+        if PrecisionUtil.get_bitfield () then
+          let a', b' = ID.meet a (ID.logand a c), ID.meet b (ID.logand b c) in 
+          let (cz, co) = ID.to_bitfield ikind c in 
+          let (az, ao) = ID.to_bitfield ikind a' in 
+          let (bz, bo) = ID.to_bitfield ikind b' in 
+          let cDef1 = Z.logand co (Z.lognot cz) in 
+          let aDef0 = Z.logand az (Z.lognot ao) in
+          let bDef0 = Z.logand bz (Z.lognot bo) in
+          let az = Z.logand az (Z.lognot (Z.logand bDef0 cDef1)) in 
+          let bz = Z.logand bz (Z.lognot (Z.logand aDef0 cDef1)) in
```
michael-schwarz, line 409: Should most of these computations not be moved inside your domain? (After getting the bitfields).

This works on the internal representation of these bitfield things, so maybe should be in there. Granted, the intervals also leak outside of their domain,but people intuitively understand how they work.

Also, please leave some comment here about what happens. One can puzzle it together, but it took @DrMichaelPetter and me about 10 mins yesterday to do so.

src/analyses/baseInvariant.ml
```diff
@@ -395,27 +395,55 @@ struct
...
         let a =
           match ID.to_int b with
           | Some x when Z.equal x Z.one ->
             (match ID.to_bool c with
              | Some true -> ID.meet a (ID.of_congruence ikind (Z.one, Z.of_int 2))
              | Some false -> ID.meet a (ID.of_congruence ikind (Z.zero, Z.of_int 2))
-             | None -> if M.tracing then M.tracel "inv" "Unhandled case for operator x %a 1 = %a" d_binop op ID.pretty c; a)
-          | _ -> if M.tracing then M.tracel "inv" "Unhandled case for operator x %a %a = %a" d_binop op ID.pretty b ID.pretty c; a
+             | None -> a)
+          | _ -> a
```
michael-schwarz, line 433: The tracing output should still be produced if bitfield is off.

src/analyses/baseInvariant.ml
```diff
@@ -395,27 +395,55 @@ struct
...
+          let bDef0 = Z.logand bz (Z.lognot bo) in
+          let az = Z.logand az (Z.lognot (Z.logand bDef0 cDef1)) in 
+          let bz = Z.logand bz (Z.lognot (Z.logand aDef0 cDef1)) in
+          ID.meet a' (ID.of_bitfield ikind (az, ao)), ID.meet b' (ID.of_bitfield ikind (bz, bo))
+        else a, b
+      | BXor ->
+        (* Be careful: inv_exp performs a meet on both arguments of the BOr / BXor. *)
+        if PrecisionUtil.get_bitfield () then
+          let a' = ID.meet a (ID.logxor c b)
+          in let b' = ID.meet b (ID.logxor a c)
```
michael-schwarz, line 416: Please change to use standard indentation.

~~~Ocaml
let ... = ... in
~~~

src/analyses/baseInvariant.ml
```diff
@@ -395,27 +395,55 @@ struct
...
+          let (cz, co) = ID.to_bitfield ikind c in 
+          let (az, ao) = ID.to_bitfield ikind a' in 
+          let (bz, bo) = ID.to_bitfield ikind b' in 
+          let cDef0 = Z.logand cz (Z.lognot co) in 
+          let aDef1 = Z.logand ao (Z.lognot az) in
+          let bDef1 = Z.logand bo (Z.lognot bz) in
+          let ao = Z.logand ao (Z.lognot (Z.logand bDef1 cDef0)) in 
+          let bo = Z.logand bo (Z.lognot (Z.logand aDef1 cDef0)) in
+          ID.meet a' (ID.of_bitfield ikind (az, ao)), ID.meet b' (ID.of_bitfield ikind (bz, bo))
+        else a, b
```
michael-schwarz, line 446: Same as above.

src/cdomain/value/cdomains/int/congruenceDomain.ml
```diff
@@ -139,6 +139,17 @@ struct
...
+  let to_bitfield ik x = 
+    let is_power_of_two x = (Z.logand x (x -: Z.one) = Z.zero) in
+    match x with None -> (Z.zero, Z.zero) | Some (c,m) ->
+      if m = Z.zero then (Z.lognot c, c)
+      else if is_power_of_two m then 
+        let mod_mask = m -: Z.one in 
+        let z = Z.lognot c in 
+        let o = Z.logor (Z.lognot mod_mask) c in 
+        (z,o)
+      else (Z.lognot Z.zero, Z.lognot Z.zero)
```
michael-schwarz, line 151: Does one need to consider the signedness here?

src/cdomain/value/cdomains/int/congruenceDomain.ml
```diff
@@ -489,6 +500,9 @@ struct
     refn
 
   let refine_with_congruence ik a b = meet ik a b
+  let refine_with_bitfield ik a (z,o) = 
+    if Z.lognot z = o then meet ik a (Some (o, Z.zero))
+    else a
```
michael-schwarz, line 505: Is this the only case where we can benefit? Naively, one thinks we should be able to benefit along the reasoning of the `to_bitfield` above?

src/cdomain/value/cdomains/int/defExcDomain.ml
```diff
@@ -299,6 +299,12 @@ struct
       let ex = if Z.gt x Z.zero || Z.lt y Z.zero then S.singleton Z.zero else  S.empty () in
       norm ik @@ (`Excluded (ex, r))
 
+  let to_bitfield ik x = 
+    match x with 
+      `Definite c -> (Z.lognot c, c) |
+      _ -> let one_mask = Z.lognot Z.zero 
+      in (one_mask, one_mask)
```
michael-schwarz, line 306: Indentation

src/cdomain/value/cdomains/int/defExcDomain.ml
```diff
@@ -530,6 +536,9 @@ struct
     ] (* S TODO: decide frequencies *)
 
   let refine_with_congruence ik a b = a
+  let refine_with_bitfield ik x (z,o) = 
+    if Z.lognot z = o then meet ik x (`Definite o)
+    else x
```
michael-schwarz, line 541: Do you not need to consider the ik here?

src/cdomain/value/cdomains/int/enumsDomain.ml
```diff
@@ -250,6 +250,16 @@ module Enums : S with type int_t = Z.t = struct
...
+  let to_bitfield ik x = 
+    match x with 
+      Inc i when BISet.is_empty i -> (Z.zero, Z.zero) |
+      Inc i when BISet.is_singleton i -> 
+      let o = BISet.choose i 
+      in (Z.lognot o, o) |
+      Inc i -> BISet.fold (fun o (az, ao) -> (Z.logor (Z.lognot o) az, Z.logor o ao)) i (Z.zero, Z.zero) |
+      _ -> let one_mask = Z.lognot Z.zero     
+      in (one_mask, one_mask)
+
```
michael-schwarz, line 262: With this indentation it is near impossible to review!

src/cdomain/value/cdomains/int/intervalDomain.ml
```diff
@@ -1,4 +1,5 @@
 open IntDomain0
+open GoblintCil
```
michael-schwarz, line 2: Where are you using this? If it's only very few places, one may not want to open it.

src/cdomain/value/cdomains/int/intervalSetDomain.ml
```diff
@@ -236,6 +236,28 @@ struct
...
+         if isSigned ik && isPositive then Ints_t.logand signMask o
+         else o 
+    in fst (norm_interval ik (min ik x, max ik x))
+
+  let to_bitfield ik x = 
+    let joinbf (z1,o1) (z2,o2) = (Ints_t.logor z1 z2, Ints_t.logor o1 o2) in 
+    let rec from_list is acc = match is with
+        [] -> acc |
+        j::js -> from_list js (joinbf acc (Interval.to_bitfield ik (Some j)))
+    in from_list x (Ints_t.zero, Ints_t.zero)
```
michael-schwarz, line 259: Is this not just a fold?

src/cdomain/value/cdomains/int/intervalDomain.ml
```diff
@@ -79,11 +80,77 @@ struct
...
+          let bigger_number = Ints_t.add without_remainder position in
+
+          let bit_status =
+            if Ints_t.compare bigger_number endv <= 0 then
+              `top
+            else 
+            if Ints_t.equal (Ints_t.logand (Ints_t.shift_right startv pos) Ints_t.one) Ints_t.one then
+              `one
+            else
+              `zero
```
michael-schwarz, line 138: where do these polymorphic variants now come from?

src/cdomain/value/util/precisionUtil.ml
```diff
@@ -51,17 +57,19 @@ let reset_lazy () =
   enums := None;
   congruence := None;
   interval_set := None;
-  annotation_int_enabled := None
+  annotation_int_enabled := None;
+  bitfield := None
```
michael-schwarz, line 61: Please swap

src/framework/control.ml
```diff
@@ -380,7 +380,7 @@ struct
     let test_domain (module D: Lattice.S): unit =
       let module DP = DomainProperties.All (D) in
       Logs.debug "domain testing...: %s" (D.name ());
-      let errcode = QCheck_base_runner.run_tests DP.tests in
+      let errcode = QCheck_base_runner.run_tests DP.tests  ~verbose:true in
```
michael-schwarz, line 383: Why verbose?

tests/regression/82-bitfield/00-simple-demo.c
```diff
@@ -0,0 +1,29 @@
+// PARAM: --enable ana.int.bitfield
```
michael-schwarz, line 1: In general it seems very few of your pograms tests actually test the join of your domain. That would be something that probably should be added.

src/cdomain/value/cdomains/int/intDomTuple.ml
```diff
@@ -88,25 +92,28 @@ module IntDomTupleImpl = struct
...
-    | (_, _, _, Some true, _)
-    | (_, _, _, _, Some true) ->
-      true
+    | (Some true, _, _, _, _,_)
+    | (_, Some true, _, _, _,_)
+    | (_, _, Some true, _, _,_)
+    | (_, _, _, Some true, _,_)
+    | (_, _, _, _, Some true,_) 
+    | (_, _, _, _, _, Some true) 
+      -> true
```
michael-schwarz, line 105: Please stick with the normal indentation.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+      let bf_bit = nth_bf_bit (z,o) 0 in
+      concretize (z >>. 1, o >>: 1) |>
+      if bf_bit = `Undetermined then
+        List.concat_map (fun c -> [c <<: 1; (c <<: 1) |: Ints_t.one])
+      else if bf_bit = `Invalid then
+        failwith "Should not have happened: Invalid bit during concretization of a bitfield."
+      else
+        List.map (fun c -> c <<: 1 |: bit)
+
+  let concretize bf = List.map Ints_t.to_int (concretize bf)
```
michael-schwarz, line 119: Where is this used?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+  type int_t = Ints_t.t
+  type t = (Ints_t.t * Ints_t.t) [@@deriving eq, ord, hash]
+
+  module BArith = BitfieldArith (Ints_t)
+
+  let top () = (BArith.one_mask, BArith.one_mask)
+  let bot () = (BArith.zero_mask, BArith.zero_mask)
+
+  let top_of ik = 
+    if isSigned ik then top ()
```
michael-schwarz, line 173: I don't think i understand how you handle `ikinds` in general here.

- Do you norm the upper bits?
- If not do you not risk having values that are semantically equal but syntactically not? 
- What happens for `ikind`s that are wider than the `BArith.one_mask`?

Why can this not work for signed types? And does this not mean you find out nothing as soon as any computations happen in the program (even for unsigned types)?

C99 6.3.1.

> If an int can represent all values of the original type, the value is converted to an int; otherwise, it is converted to an unsigned int. These are called the integer
promotions. All other types are unchanged by the integer promotions.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+    if isSigned ik then top () 
+    else (BArith.one_mask, Ints_t.of_bigint (snd (Size.range ik)))
+
+  let bot_of ik = bot ()
+
+  let to_pretty_bits (z,o) = 
+    let known_bitmask = ref (BArith.bits_known (z,o)) in
+    let invalid_bitmask = ref (BArith.bits_invalid (z,o)) in
+    let o_mask = ref o in
+    let z_mask = ref z in
```
michael-schwarz, line 182: Why `ref`?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+    if t = top () then "top" else
+      let (z,o) = t in
+      Format.sprintf "{%s, (zs:%s, os:%s)}" (to_pretty_bits t) (Ints_t.to_string z) (Ints_t.to_string o) 
+
+  include Std (struct type nonrec t = t let name = name let top_of = top_of let bot_of = bot_of let show = show let equal = equal end)
+
+  let range ik bf = (BArith.min ik bf, BArith.max ik bf)
+
+  let maximal (z,o) =     
+    if (z < Ints_t.zero) <> (o < Ints_t.zero) then Some o
```
michael-schwarz, line 218: Can you explain this?

michael-schwarz (commented):
Thanks for the PR! I left some first comments in some places, but have not worked through the entire thing.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
Draggon01: The 'concretize' method is used inside the shift operations (shift_left, shift_right) as it returns the values which needs to be shifted.

michael-schwarz: Fully concretizing an abstract value is not something that should be done lightly, as you have noted yourself. If you do a shift and know nothing about the operand (which can happen), this operation alone will completely kill the chance of ever terminating, already from constructing this list of length 2^64.

Does your domain not have an efficient way of computing $$| \gamma (x) |$$ by counting the number `n` of zero bits in `z xor x` (size 2^n)? In this case, you should only do something for small n (maybe up to 3) to stay tractable.

michael-schwarz: Ah, I see you actually bound the argument in all calls with the bitwidth, so it is not too problematic. Probably you should make a comment here as well then that the function is not a reasonable shift function for any arguments, but only for second arguments with small(!) concretizations.

Draggon01: Sure that makes sense, I will add comments that make this more obvious.

src/cdomain/value/cdomains/int/congruenceDomain.ml
AdrianKrauss: We benefit along the reasoning of the `to_bitfield` in the `refine_with_congruence` of the bitfield domain because we transform a value from the congruence into the bitfield domain. Unfortunately, I don't see how we can use similar tricks for the converse direction.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
Draggon01: It was not necessary to use references in this case, therefore the code was changed and the ref removed.

src/cdomain/value/cdomains/int/intervalDomain.ml
michael-schwarz: Ok, since it's only used three times (and only in one function), I think i would personally prefer getting rid of the `open` and qualifying  those references with `GoblintCil.isSigned` or using a local open in that function.

src/cdomain/value/cdomains/int/congruenceDomain.ml
AdrianKrauss: I'm unsure whether you mean the signedess of c and m or of the returned bitfield. But in both cases, we do not have to consider it explicitly. `normalize` always returns positive values for c and m if m is unequal to zero. Additionally, a possible sign bit is set to top in the returned bitfield representation.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
AdrianKrauss: We basically check whether the sign bit is clear, i.e., whether we only represent positive or negative numbers. Under such a condition, the maximal represented number is just the number with the maximal number of ones.

michael-schwarz: Hmm, that works if the `ikind` and the type of `Ints_t` coincide. But what if you have, e.g., the `ikind` of  `signed char`? There you would need to check the 7th bit right?

I wasn't there for the first couple of meetings, so maybe you'll need to enlighten us tomorrow how you deal with different `ikind`s in your domain in general :smiley:

src/cdomain/value/cdomains/int/intervalDomain.ml
```diff
@@ -79,11 +80,77 @@ struct
...
+
   let of_int ik (x: int_t) = of_interval ik (x,x)
   let zero = Some IArith.zero
   let one  = Some IArith.one
   let top_bool = Some IArith.top_bool
 
+  let to_bitfield ik z = 
+    match z with None -> (Ints_t.lognot Ints_t.zero, Ints_t.lognot Ints_t.zero) | Some (x,y) ->
+      let (min_ik, max_ik) = Size.range ik in
+      let startv = Ints_t.max x (Ints_t.of_bigint min_ik) in
```
jerhard, line 107: Can you make this function easier to read?
A comment about what is done here overall may help. Possibly one may also find a more telling name for the helper function `analyze_bits`?

src/cdomain/value/cdomains/int/intervalSetDomain.ml
```diff
@@ -236,6 +236,28 @@ struct
...
   let of_interval ?(suppress_ovwarn=false) ik (x,y) =  norm_interval  ~suppress_ovwarn ~cast:false ik (x,y)
 
+  let of_bitfield ik x = 
+    let min ik (z,o) = 
+      let signBit = Ints_t.shift_left Ints_t.one ((Size.bit ik) - 1) in 
+      let signMask = Ints_t.lognot (Ints_t.of_bigint (snd (Size.range ik))) in
+      let isNegative = Ints_t.logand signBit o <> Ints_t.zero in
+      if isSigned ik && isNegative then Ints_t.logor signMask (Ints_t.lognot z)
+      else Ints_t.lognot z
+    in let max ik (z,o) =
```
jerhard, line 246: Indentation.

src/cdomain/value/cdomains/int/intervalSetDomain.ml
```diff
@@ -236,6 +236,28 @@ struct
 
   let of_interval ?(suppress_ovwarn=false) ik (x,y) =  norm_interval  ~suppress_ovwarn ~cast:false ik (x,y)
 
+  let of_bitfield ik x = 
+    let min ik (z,o) = 
+      let signBit = Ints_t.shift_left Ints_t.one ((Size.bit ik) - 1) in 
+      let signMask = Ints_t.lognot (Ints_t.of_bigint (snd (Size.range ik))) in
```
jerhard, line 242: `signBit` and `signMask` only seem to depend on `ik'. Can one extract them from `min` and `max`, so they don't have to be defined twice?

tests/unit/cdomains/intDomainTest.ml
```diff
@@ -250,7 +250,696 @@ struct
...
+      assert_shift_right ik bot bot bot;
+
+      assert_shift_right ik (`I [0]) top (`I [0]);
+
+      if isSigned ik
+      then (
+        assert_shift_right ~rev_cond:true ik (`I [max_of ik]) top top; (* the sign bit shouldn't be set with right shifts if its unset *)
+
+        assert_shift_right ik (`I [2]) (`I [-1]) top; (* Negative shifts are undefined behavior *)
+        (*assert_shift_right ik (`I [min_of ik]) top top;*) (*TODO implementation-defined sign-bit handling *)
```
jerhard, line 639: What is this todo about?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+      Ints_t.sub top_bit Ints_t.one
+
+  let nth_bit p n = Ints_t.one &: (p >>: n) =: Ints_t.one
+
+  let nth_bf_bit (z,o) n =
+    match nth_bit z n, nth_bit o n with
+    | true, true -> `Undetermined
+    | false, false -> `Invalid
+    | true, false -> `Zero
+    | false, true -> `One
```
jerhard, line 90: Is there a specific reason you are using polymorphic variants? If not, it would be preferable to define a variant type. See also [here](https://ocaml.org/manual/5.1/polyvariant.html#s:polyvariant-weaknesses).

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+  include Std (struct type nonrec t = t let name = name let top_of = top_of let bot_of = bot_of let show = show let equal = equal end)
+
+  let range ik bf = (BArith.min ik bf, BArith.max ik bf)
+
+  let maximal (z,o) =     
+    if (z < Ints_t.zero) <> (o < Ints_t.zero) then Some o
+    else None
+
+  let minimal (z,o) =     
+    if (z < Ints_t.zero) <> (o < Ints_t.zero) then Some (!:z)
```
jerhard, line 222: Using the polymorphic compare `<` here probably works, but the compare on `Ints_t` is probably still the better choice?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+      let new_bitfield = wrap ik (z,o)
+      in
+      let overflow_info = if suppress_ovwarn then {underflow=false; overflow=false} else {underflow=underflow; overflow=overflow} in      
+      if not (underflow || overflow) then
+        ((z,o), overflow_info)
+      else if should_wrap ik then
+        (new_bitfield, overflow_info)
+      else if should_ignore_overflow ik then 
+        (* (M.warn ~category:M.Category.Integer.overflow "Bitfield: Value was outside of range, indicating overflow, but 'sem.int.signed_overflow' is 'assume_none' -> Returned Top"; *)
+        (* (bot (), overflow_info)) *)
```
jerhard, line 254: Commented out code should be removed in the final version.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+    else None
+
+  let equal_to i bf = 
+    if BArith.of_int i = bf then `Eq
+    else if leq (BArith.of_int i) bf then `Top
+    else `Neq
+
+  (* Conversions *)
+
+  let of_interval ?(suppress_ovwarn=false) ik (x,y) =
```
jerhard, line 284: Does this function share code with the `to_bitfield` function in the interval domain? If so, could this be deduplicated?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+        let bigger_number = Ints_t.add without_remainder position in
+
+        let bit_status =
+          if Ints_t.compare bigger_number endv <= 0 then
+            `top
+          else 
+          if Ints_t.equal (Ints_t.logand (Ints_t.shift_right startv pos) Ints_t.one) Ints_t.one then
+            `one
+          else
+            `zero
```
jerhard, line 306: Here, one could also define a variant type, instead of using a polymorphic variant type.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,623 @@
...
+      | None -> top_of ik
+      | Some ls -> 
+        List.fold_left (fun acc i -> BArith.join acc (BArith.of_int i)) (bot_of ik) ls
+    in
+    meet ik t joined
+
+
+  (* Unit Tests *)
+
+  let arbitrary ik =
```
jerhard, line 601: `arbitrary` is used for unit tests, indeed; still the comment may be a bit misleading, since there are no unit tests at following the comment.

jerhard (commented):
Thank you for the PR! I left some comments below.

src/framework/control.ml
sim642: I'm curious if you've actually used this and found anything with this. Domain tests for individual integer domains are part of unit tests where they are directly triggered. This attempts to test the full domain used by the configured analysis, but I suspect this doesn't have `arbitrary`-s defined all the way down to the integer domains through all the functors. For that reason I've even considered removing this feature altogether, but if it has actually been useful then let me know.

src/cdomain/value/cdomains/int/intervalDomain.ml
AdrianKrauss: done

src/cdomain/value/cdomains/int/bitfieldDomain.ml
iC4rl0s: No, they were naively chosen to be used for readability purposes here but not necessary in any way. This part will be removed.

tests/unit/cdomains/intDomainTest.ml
iC4rl0s: This test case could be removed if you agree with our design choice to handle right shifts as arithmetic ones.  
Given the C standard's definition of right shifts on signed integers, it might be better to leave the upper bits undetermined. In that case this test case should pass.

src/cdomain/value/cdomains/int/congruenceDomain.ml
```diff
@@ -49,11 +49,11 @@ struct
   let bot_of ik = bot ()
 
   let show = function ik -> match ik with
-    | None -> "⟂"
+    | None -> "bot"
```
jerhard, line 52: What is the reason changing the symbol for a bottom value?

src/cdomain/value/cdomains/int/congruenceDomain.ml
```diff
@@ -49,11 +49,11 @@ struct
...
     | Some (c, m) when (c, m) = (Z.zero, Z.zero) -> Z.to_string c
     | Some (c, m) ->
       let a = if c =: Z.zero then "" else Z.to_string c in
-      let b = if m =: Z.zero then "" else if m = Z.one then "ℤ" else Z.to_string m^"ℤ" in
+      let b = if m =: Z.zero then "" else if m = Z.one then "Z" else Z.to_string m^"Z" in
```
jerhard, line 56: What is the reason for changing the symbol for the whole numbers ℤ?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,709 @@
...
+       accm := snd(add_paper Ints_t.zero !accm Ints_t.zero !qm));
+
+      pv := !pv >>: 1;
+      pm := !pm >>: 1;
+      qv := !qv <<: 1;
+      qm := !qm <<: 1;
+    done; 
+    let (rv, rm) = add_paper !accv Ints_t.zero Ints_t.zero !accm in 
+    let o3 = ref(rv |: rm) in 
+    let z3 = ref(!:rv |: rm) in
```
jerhard, line 529: Can one avoid making these variables references?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,709 @@
...
+        List.fold_left (fun acc i -> BArith.join acc (BArith.of_int i)) (bot_of ik) ls
+    in
+    meet ik t joined
+
+
+  let refine_bor (az, ao) (bz, bo) (cz, co) =     
+    let cDef0 = cz &: (!: co) in 
+    let cDef1 = co &: (!: cz) in         
+    let aDef0 = az &: (!: ao) in   
+    let bDef0 = bz &: (!: bo) in
```
jerhard, line 655: Have you tested whether this works as intended to refine on  a bitwise level? As far as I can see, the `!:` is a _logical_ not operation, i.e., it is not a bitwise not.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,709 @@
...
+    (* example (with t for top): (tttt) & (t010) = (1011) *)
+    (* we can refine (tttt) to (1t11) *)
+    let az = az &: (!: cDef1) in 
+    let bz = bz &: (!: cDef1) in 
+    (* if a bit is definitely 1 in b and definitely 0 in c, the same bit must be definitely 0 in a *)
+    (* example (with t for top): (tttt) & (t110) = (1011) *)
+    (* we can refine (tttt) to (t0tt) *)
+    let ao = ao &: (!: (bDef1 &: cDef0)) in     
+    let bo = bo &: (!: (aDef1 &: cDef0)) in
+    ((az, ao), (bz, bo))
```
jerhard, line 683: Have you tested whether this works as intended to refine on  a bitwise level? As far as I can see, the `!:` is a _logical_ not operation, i.e., it is not a bitwise not.

tests/regression/82-bitfield/09-refine-intervalA.c
```diff
@@ -0,0 +1,22 @@
+// PARAM: --enable ana.int.bitfield --set ana.int.refinement fixpoint
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int main() {
+  int a = rand();
+
+  // 1110 in binary
+  int inv_mask = ~0xe;  // 1111...10001 in binary
```
jerhard, line 10: Are these two conflicting comments?

jerhard (commented):
I had another look and left some more comments below.

jerhard:
@ManuelLerchner  To document this PR better for people  who may look at it in the future , could you please add a few sentences in your top post to elaborate on the approach and what you have changed?

src/cdomain/value/cdomains/int/congruenceDomain.ml
AdrianKrauss: This change wasn't intentionally, my bad. I changed it because it was easier to read for me during debugging, but I already changed it back.

src/cdomain/value/cdomains/int/congruenceDomain.ml
AdrianKrauss: Same as above.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,709 @@
...
+  https://is.muni.cz/th/kasap/thesis.pdf
+*)
+
+(* Bitfield arithmetic, without any overflow handling etc. *)
+module BitfieldArith (Ints_t : IntOps.IntOps) = struct
+
+  include InfixIntOps (Ints_t)
+
+  let zero_mask = Ints_t.zero
+  let one_mask = !:zero_mask
```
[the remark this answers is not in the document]
AdrianKrauss: No, `!:` is a **bitwise** logical negation. However, it would be the same as minus one which is not provided by `Ints_t`.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
AdrianKrauss: According to the comment in the corresponding module, it is actually a bitwise operation. Nonetheless, the regression test 82 10 tests this behaviour. So, it works as intended.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
AdrianKrauss: Same as above.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,709 @@
...
+    let o_mask = o in
+    let z_mask = z in
+
+    let rec to_pretty_bits' o_mask z_mask known_bitmask invalid_bitmask acc =
+      let current_bit_known = (known_bitmask &: Ints_t.one) = Ints_t.one in
+      let current_bit_impossible = (invalid_bitmask &: Ints_t.one) = Ints_t.one in
+      let bit_value = o_mask &: Ints_t.one in
+      let bit =
+        if current_bit_impossible then "⊥"
+        else if not current_bit_known then "⊤"
```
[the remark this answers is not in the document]
ManuelLerchner: Since this is just a matter of style, we can of course change it.
We are however used to this notation and would keep it as it is, but in the end it is your decision.
Would you suggest changing it?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,709 @@
...
+    let invalid_bitmask = (BArith.bits_invalid (z,o)) in
+    let o_mask = o in
+    let z_mask = z in
+
+    let rec to_pretty_bits' o_mask z_mask known_bitmask invalid_bitmask acc =
+      let current_bit_known = (known_bitmask &: Ints_t.one) = Ints_t.one in
+      let current_bit_impossible = (invalid_bitmask &: Ints_t.one) = Ints_t.one in
+      let bit_value = o_mask &: Ints_t.one in
+      let bit =
+        if current_bit_impossible then "⊥"
```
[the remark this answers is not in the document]
ManuelLerchner: Yes, this can happen, but only for our `bot`-value. `bot` is defined to be the value where all bits are impossible.
(Due to the normalization, it can never happen, that just a single bit is impossible. If a single bit is impossible, we normalize it such that all (infinitely many) bits are impossible)

src/framework/control.ml
AdrianKrauss: We found it quite useful to find some basic bugs in our code. In my opinion, a great advantage is also that the domain tests use all the possible integer kinds which makes the testing for us easier, i.e. we have to write less test cases and only have to focus on some special edge cases.

sim642: Testing with all ikinds is part of unit tests (`dune runtest tests/unit`). This particular code just concerns the option `dbg.test.domain` when enabled during the analysis of a particular program. This option is not used by anything in our test suite, so unless you added any tests that explicitly enable this option, this wouldn't be used anyway.

src/analyses/baseInvariant.ml
jerhard: @AdrianKrauss The tracing output still seems to be missing?

tests/regression/82-bitfield/09-refine-intervalA.c
```diff
@@ -0,0 +1,21 @@
+// PARAM: --enable ana.int.bitfield --set ana.int.refinement fixpoint
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int main() {
+  int a = rand();
+
+  int inv_mask = ~0xe;  // 1111...10001 in binary
```
jerhard, line 9: This is a constant, why is it annotated with a range of values in the comment?

src/cdomain/value/cdomains/int/enumsDomain.ml
```diff
@@ -250,6 +250,19 @@ module Enums : S with type int_t = Z.t = struct
...
   let to_incl_list = function Inc s when not (BISet.is_empty s) -> Some (BISet.elements s) | _ -> None
 
+  let to_bitfield ik x = 
+    let ik_mask = snd (Size.range ik) in 
+    let one_mask = Z.lognot Z.zero in
+    match x with 
+    | Inc i when BISet.is_empty i -> (Z.zero, Z.zero) 
+    | Inc i when BISet.is_singleton i ->       
+      let o = BISet.choose i in 
+      let o = (if Cil.isSigned ik then o else Z.logand ik_mask o) in
```
jerhard, line 260:
```suggestion
      let o = if Cil.isSigned ik then o else Z.logand ik_mask o in 
```

tests/regression/82-bitfield/12-precision.c
```diff
@@ -0,0 +1,47 @@
...
+    switch (r) {
+        case 0:
+        state = 0; /* 0b0000 */
+         break;
+        case 1:
+        state = 8; /* 0b1000 */
+         break;
+        default:
+        state = 10; /* 0b1010 */
+         break;
```
jerhard, line 25: The identation here is strange.

tests/regression/82-bitfield/12-precision.c
```diff
@@ -0,0 +1,47 @@
...
```
jerhard, line 43: Indentation.

tests/regression/82-bitfield/00-simple-demo.c
jerhard: Did you add more tests w.r.t. that?

tests/regression/82-bitfield/09-refine-intervalA.c
ManuelLerchner: Ah, i see the source of confusion here. The comment should indicate that `inv_mask = 0b1111.1111.1111.1111.1111.1111.1111.0001` when represented in binary. The `...` is not meant as a range.
I'll further improve the comment.

src/analyses/baseInvariant.ml
```diff
@@ -395,27 +395,55 @@ struct
...
+        if PrecisionUtil.get_bitfield () then
+          (* from a ^ b = c follows a = b ^ c *)
+          let a' = ID.meet a (ID.logxor c b) in 
+          let b' = ID.meet b (ID.logxor a c) in
+          a', b'
+        else 
+          (if M.tracing then M.tracel "inv" "Unhandled operator %a" d_binop op;
+           (* Be careful: inv_exp performs a meet on both arguments of the BOr / BXor. *)
+           (a, b)
+          )
```
sim642, line 420: Is there a reason to really limit this to just bitfields?
Seems like it could be very similar to `PlusA`: `meet_com ID.logxor`.

src/analyses/baseInvariant.ml
```diff
@@ -395,27 +395,55 @@ struct
...
-        a, b
+        if PrecisionUtil.get_bitfield () then 
+          (* refinement based on the following idea: bit set to zero in c and set to one in b must be zero in a and bit set to one in c must be one in a too (analogously for b) *)
+          let ((az, ao), (bz, bo)) = BitfieldDomain.Bitfield.refine_band (ID.to_bitfield ikind a) (ID.to_bitfield ikind b) (ID.to_bitfield ikind c) in 
+          ID.meet a (ID.of_bitfield ikind (az, ao)), ID.meet b (ID.of_bitfield ikind (bz, bo))
+        else if b_int = None then 
+          (if M.tracing then M.tracel "inv" "Unhandled operator %a" d_binop op;
+           (a, b)
+          )
+        else a, b
```
sim642, line 446: The `b_int = None` case could just be the default like for `BOr` above.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,736 @@
...
+  let (<>:) = fun a b -> Ints_t.compare a b <> 0
+
+  let (+:) = Ints_t.add
+  let (-:) = Ints_t.sub
+  let ( *: ) = Ints_t.mul
+  let (/:) = Ints_t.div
+  let (%:) = Ints_t.rem
+
+  let (>>.) = fun a b -> a >>: b |: !:((Ints_t.one <<: b) -: Ints_t.one)
+end
```
sim642, line 22: Perhaps this should be extracted to `intOps.ml` itself and reused. For example `CongruenceDomain` defines many similar operators, but could just use this functor instead.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,736 @@
...
+    else
+      ((z >>: c) |: sign_mask, o >>: c)
+
+  let shift_right ik bf (z2, o2) = 
+    if is_const (z2, o2) 
+    then 
+      shift_right ik bf (Ints_t.to_int o2)
+    else
+      let defined_shifts = constrain_to_bit_width_of ik (z2, o2) in
+      let shift_amounts = concretize defined_shifts in (* O(2^(log n)) *)
```
sim642, line 152: If `log` is base 2, then this is just $$O(n)$$. Also, what is $$n$$ here?

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,736 @@
...
+        prefix ^ acc
+      else
+        create_pretty_bf_string (o_mask >>: 1) (z_mask >>: 1) (known_bitmask >>: 1) (invalid_bitmask >>: 1) (bit ^ acc)
+    in
+    "0b" ^ create_pretty_bf_string o z known_bitmask invalid_bitmask ""
+
+  let show t = 
+    if t = bot () then "⊥" else
+      let (z,o) = t in
+      Format.sprintf "{%s, (zs:%s, os:%s)}" (to_pretty_bits t) (Ints_t.to_string z) (Ints_t.to_string o)
```
sim642, line 241: It seems a bit too verbose because the pretty bits already contain all the information. If not remove the others altogether, then at least make them conditional on `dbg.full-output`.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,736 @@
...
+      | None -> (false, false) (* ik does not change *)
+      | Some (GoblintCil.Cil.TInt (old_ik, _)) ->         
+        let underflow = Z.compare (BArith.min old_ik (z,o)) min_ik < 0 in 
+        let overflow = Z.compare max_ik (BArith.max old_ik (z,o)) < 0 in 
+        (underflow, overflow)
+      | _ -> 
+        let isPos = z < Ints_t.zero in 
+        let isNeg = o < Ints_t.zero in
+        let underflow = if GoblintCil.isSigned ik then (((Ints_t.of_bigint min_ik) &: z) <> Ints_t.zero) && isNeg else isNeg in
+        let overflow = (((!:(Ints_t.of_bigint max_ik)) &: o) <> Ints_t.zero) && isPos in
```
sim642, line 287: These should probably not use polymorphic comparison:
```suggestion
        let isPos = z <: Ints_t.zero in 
        let isNeg = o <: Ints_t.zero in
        let underflow = if GoblintCil.isSigned ik then (((Ints_t.of_bigint min_ik) &: z) <>: Ints_t.zero) && isNeg else isNeg in
        let overflow = (((!:(Ints_t.of_bigint max_ik)) &: o) <>: Ints_t.zero) && isPos in
```

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,736 @@
...
+        (underflow, overflow)
+    in
+    let overflow_info = if suppress_ovwarn then {underflow=false; overflow=false} else {underflow=underflow; overflow=overflow} in      
+    (norm ~suppress_ovwarn:(suppress_ovwarn) ~ov:(underflow || overflow) ik (z,o), overflow_info)
+
+  let join ik b1 b2 = norm ik @@ (BArith.join b1 b2)
+
+  let meet ik x y = norm ik @@ (BArith.meet x y)
+
+  let leq (x:t) (y:t) = (BArith.join x y) = y
```
sim642, line 297: Could this be checked directly? That could avoid allocating in a predicate.
Also, the polymorphic equality probably isn't supposed to be here.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,736 @@
...
+    in      
+    let result = construct_bitfield (Size.bit ik - 1) (bot()) in
+    let casted = (Ints_t.of_bigint (Size.cast ik ((Ints_t.to_bigint (fst result)))), Ints_t.of_bigint (Size.cast ik ((Ints_t.to_bigint (snd result))))) 
+    in (wrap ik casted, {underflow=false; overflow=false})
+
+  let of_bool _ik = function true -> BArith.one | false -> BArith.zero
+
+  let to_bool d =
+    if not (leq BArith.zero d) then Some true
+    else if d = BArith.zero then Some false
```
sim642, line 369: Polymorphic equality.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,736 @@
...
+  let logor  ik i1 i2 = BArith.logor i1 i2 |> norm ik
+
+  let lognot ik i1 = BArith.lognot i1 |> norm ik
+
+  let get_arch_bitwidth () : int = if GobConfig.get_bool "ana.sv-comp.enabled" then (
+      match GobConfig.get_string "exp.architecture" with
+      | "32bit" -> 32
+      | "64bit" -> 64
+      | _ -> Sys.word_size
+    ) else Sys.word_size
```
sim642, line 420: This seems a bit suspicious. Shouldn't this be based on the ikind used? The architecture stuff determines what CIL uses as ikind bounds.
This somewhat duplicates that logic which is error-prone.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -0,0 +1,736 @@
...
+    ((az, ao), (bz, bo))
+
+  let arbitrary ik = 
+    let open QCheck.Iter in
+    let int_arb = QCheck.map ~rev:Ints_t.to_int64 Ints_t.of_int64 GobQCheck.Arbitrary.int64 in
+    let pair_arb = QCheck.pair int_arb int_arb in
+    let shrink (z, o) =
+      (GobQCheck.shrink pair_arb (z, o) 
+       >|= (fun (new_z, new_o) -> 
+           (* Randomly flip bits to be opposite *)
```
sim642, line 721: In what sense does this shrink/simplify the abstract value? What prevents a bit from being flipped back and forth indefinitely?
I would expect shrinking to flip 1s to 0s (in either mask) or something.

src/cdomain/value/cdomains/int/congruenceDomain.ml
```diff
@@ -343,7 +360,7 @@ struct
       see: http://www.es.mdh.se/pdf_publications/948.pdf *)
   let bit2 f ik x y = match x, y with
     | None, None -> None
-    | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
+    | None, _ | _, None -> raise (ArithmeticOnIntegerBot ((show x) ^ " op " ^ (show y)))
```
sim642, line 363: What is this supposed to change? Seems like nothing.

tests/regression/82-bitfield/00-simple-demo.c
ManuelLerchner: I just added some additional tests.

src/cdomain/value/cdomains/int/defExcDomain.ml
```diff
@@ -299,6 +299,15 @@ struct
...
       norm ik @@ (`Excluded (ex, r))
 
+  let to_bitfield ik x = 
+    match x with 
+    | `Definite c -> (Z.lognot c, c) 
+    | _ when Cil.isSigned ik -> let one_mask = Z.lognot Z.zero in 
+      (one_mask, one_mask)
+    | _ -> let one_mask = Z.lognot Z.zero in 
+      let ik_mask = snd (Size.range ik) in 
+      (one_mask, ik_mask)
```
sim642, line 309:
```suggestion
    | _ when Cil.isSigned ik ->
      let one_mask = Z.lognot Z.zero in 
      (one_mask, one_mask)
    | _ ->
      let one_mask = Z.lognot Z.zero in 
      let ik_mask = snd (Size.range ik) in 
      (one_mask, ik_mask)
```

src/cdomain/value/cdomains/intDomain0.ml
```diff
@@ -260,6 +264,20 @@ sig
   val shift_right : Cil.ikind -> t -> t -> t * overflow_info
 end
 
+module type Bitfield_SOverflow = 
+sig 
+
+  include SOverflow 
+
+  (* used for refinements in other domains *)
+  val is_const : t -> bool
```
sim642, line 273: Couldn't they just use `to_int`? Or are they really cases that don't then try to get the constant value itself?

src/util/std/gobTuple.ml
```diff
@@ -0,0 +1,37 @@
...
+let _ = Tuple6.fourth
+let _ = Tuple6.fifth
+let _ = Tuple6.sixth
+
+let _ = Tuple6.map1
+let _ = Tuple6.map2
+let _ = Tuple6.map3
+let _ = Tuple6.map4
+let _ = Tuple6.map5
+let _ = Tuple6.map6
```
sim642, line 37: What compile warnings?

tests/unit/cdomains/intDomainTest.ml
```diff
@@ -250,7 +250,687 @@ struct
...
+  let ik_lst = [Cil.IChar; Cil.IUChar; Cil.IShort; Cil.IUShort; ik; Cil.IUInt;]
+
+  let string_of_ik ik = match ik with
+    | Cil.IInt -> "int"
+    | Cil.IUInt -> "unsigned_int"
+    | Cil.IChar -> "char"
+    | Cil.IUChar -> "unsigned_char"
+    | Cil.IShort -> "short"
+    | Cil.IUShort -> "unsigned_short"
+    | _ -> "undefined C primitive type"
```
sim642, line 268: Essentially duplicates `CilType.Ikind.show` and the like.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
iC4rl0s: Yes. We'll change that. $n$ is the size of the ikind.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
iC4rl0s: This should be possible if the infix operators are constrained to arithmetic, logical and bitwise operations respectively. While the `IntervalSetDomain` uses a similar notation to ours, the operators in the `CongruenceDomain` have `Z.t` instead of `IntOps.IntOps.t`. The `Z` module uses the same name for the functions for which the operators are defined as `IntOps.IntOpsBase` so this should be no issue since we can use another signature for the functor argument. The `ArrayDomain` uses similar infix operators like the congruence domain too and there is `ConvenienceOps` from `VectorMatrix.ml`that does something similar to your proposal. The question is how far is it necessary to generalise these operators. Our domain makes heavy use of our infix operators which is why they exist in the first place.

src/cdomain/value/cdomains/int/congruenceDomain.ml
AdrianKrauss: Like above, I changed it for debugging purposes but forgot to revert it which is now done.

src/analyses/baseInvariant.ml
AdrianKrauss: The reasoning behind the `b_int = None` case was to only record the tracing if both refinements could not be performed, i.e. the refinement in lines 419 - 426 and the refinement with our bitfields.

src/framework/control.ml
AdrianKrauss: Sry, I confused those things. So, we actually did not use that feature for our debugging.

src/util/std/gobTuple.ml
ManuelLerchner: I think this was originally added to remove warnings about "unused functions". But apparently, it is no longer necessary. I removed it.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
ManuelLerchner: Having direct access to decimal representations can sometimes be very handy, so I would like to keep the option.
I think it makes sense to place them behind a `dbg.full-output` clause however.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
ManuelLerchner: We can't blindly flip bits to 0 in both masks, as this will create `impossible`-bits.
We must set bits to be "opposite" in both masks. This guarantees a single value.

Currently shrinking works by:
1. Identifying unknown bits
2. Setting some of those unknown bits to 0. (By setting the corresponding bit in the zero mask to 1 and forcing it to be 0 in the ones-mask)
3. This "simplifies" the abstract value, as it reduces the number of possible values and makes the decimal representation of the masks smaller.

Once a bit is flipped to 0, it stays 0 as only unknown bits get altered.

I think the implementation makes sense, but the comment is a bit confusing. 
I'll improve the description.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
iC4rl0s: Agreed. This was replaced with the width of `top_typ`'s `ik` from`Size`.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
sim642: My intuition is that using bitwise negation is cheaper than allocating and leaving garbage on the OCaml heap. In particular, `join` allocates a new pair as an extra OCaml block.
It would be possible to micro-benchmark the two versions (in very simplified setting), we have some things in `bench/` if you're willing to give it a try.

Otherwise, switching to `Ints_t.equal` would be the minimum thing to do here. And perhaps leave a TODO about trying a non-allocating version.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -291,7 +291,9 @@ module BitfieldFunctor (Ints_t : IntOps.IntOps): Bitfield_SOverflow with type in
 
   let meet ik x y = norm ik @@ (BArith.meet x y)
 
-  let leq (x:t) (y:t) = (BArith.join x y) = y
+  let equal_bf (z1,o1) (z2,o2) = Ints_t.equal z1 z2 && Ints_t.equal o1 o2
```
sim642, line 288: This should be unnecessary, the derived `equal` should be exactly this.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
sim642: We can leave it as is, a broader cleanup in Goblint would probably be good anyway.

tests/regression/82-bitfield/00-simple-demo.c
sim642, line 1: All the tests should be moved into 83-bitfield because there's now group 82-widening_gas on master.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
ManuelLerchner: I have not benchmarked it, but I think it should be safer to go with your intuition then ^^

src/cdomain/value/cdomains/int/enumsDomain.ml
```diff
@@ -362,6 +375,13 @@ module Enums : S with type int_t = Z.t = struct
     | Inc e, Some (c, m) -> Inc (BISet.filter (contains c m) e)
     | _ -> a
 
+  let refine_with_bitfield ik x (z,o) =
+    match BitfieldDomain.Bitfield.to_int (z,o) with 
+    | Some y ->     
+      meet ik x (Inc (BISet.singleton y))
+    | _ ->
+      x
+
```
michael-schwarz, line 384: In line with #1675 (and #1673) this should now be modified to ensure no new elements appear.

```suggestion
  let refine_with_bitfield ik x (z,o) =
    match x, BitfieldDomain.Bitfield.to_int (z,o) with 
    | Inc, Some y ->     
      meet ik x (Inc (BISet.singleton y))
    | _ ->
      x

```

michael-schwarz:
The old ones were almost all addressed, I closed them.

The open one is this one https://github.com/goblint/analyzer/pull/1623#discussion_r1924982598 and the one about `refine` I just made. Once these two are addressed, I think it is ready to be merged.

Great job everyone :100: :tada: :tada:

src/analyses/baseInvariant.ml
AdrianKrauss: The tracing is produced a few lines below because as far as I understood the tracing should only be produced if nothing could be refined. This is the case iff b is not a definite integer and the bitfield is turned off. That is the intention behind the else if clause with `b_int = None` below. If there is a better way to implement this behaviour or if I misunderstood the tracing output for this case, I'll adjust this.

michael-schwarz: Yeah, you're right. I got confused.

PR #1626, opened 2024-11-10
Introduce a simplified version of td3 with warrowing (see PR 1611)

FelixKrayer:
Copied from PR #1611 by @arkocal 
The simplified warrowing solver

As to which solver we should keep as the simplified version, this one:

1. Can be documented a little more for didactic purposes, but I find it to be very readable as is.
2. We should find the paper most closely matching this solver.
3. This one does not pass all tests due to narrowing.

I created this pullrequest so I can address comments and do changes.
Addressing comments by @sim642 :
1. removed `remove-wpoint` option. Having this option is not necessary for this solver. However, I changed it to always remove wpoints, without checking the config, as this is the default setting in the config.
2. solver data is no longer a record. There is no need for that
3. moved query/side out. There is no real reason they were inside iterate, it was just the way it turned out after deleting  
4. TODO: decide if phases should be included. This solver does not have them, since the parallel solvers do not use them and it was created for comparison with those. However, for generalizing this solver I see the use of implementing the phases.

### DIFF
--- a/src/solver/td_simplified.ml
+++ b/src/solver/td_simplified.ml
@@ -63,53 +63,50 @@ module Base : GenericEqSolver =
           ) w
       in
 
-      let rec iterate x = (* ~(inner) solve in td3*)
-        let query x y = (* ~eval in td3 *)
-          if tracing then trace "solver_query" "entering query for %a; stable %b; called %b" S.Var.pretty_trace y (HM.mem stable y) (HM.mem called y);
-          get_var_event y;
-          if not (HM.mem called y) then (
-            if S.system y = None then (
-              init y;
-              HM.replace stable y ()
-            ) else (
-              HM.replace called y ();
-              if tracing then trace "iter" "iterate called from query";
-              iterate y;
-              HM.remove called y)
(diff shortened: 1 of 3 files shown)


michael-schwarz:
@sim642 @arkocal: What is you take on phases or not?
I don't have a strong opinion here, but think we should move forward with merging this.

arkocal:
I am fine with either variant, so I would have a light preference to just proceed and merge. @sim642 had the point that the basic example should have a termination guarantee, which is reasonable, just not a deal breaker for me as I find warrowing a little more intuitive than phases.

tldr: I would keep it as is, but have no strong preference, so Simmo please feel free to override this if you do.

src/solver/td_simplified.ml
sim642, line 1: This module needs to be added to `Goblint_lib` to fix the CI failure.

Also, this PR should be adapted to #1605, i.e., the new solver added to the cram test along with all the others.

sim642 (changes requested):

sim642 (approved):

PR #1631, opened 2024-11-25
Replace readwrite protection with read protection

sim642:
The mutex analysis collects two kinds of protection information: readwrite and write. The former is a subset of the latter.

This PR changes it to collect just read protection instead of readwrite. When asked for readwrite protection, the intersection is still taken to find mutexes that protect both all reads and all writes.

I hoped the `protection-read` privatization could slightly be improved by using read-protection instead of readwrite-protection for its extra check, but that didn't work out. As-is this is more of a refactoring/cleanup: analysis results aren't really affected, although the read-protection sets can be seen in g2html.

While changing that I realized we currently use `write:bool` in two different ways:
1. In queries about protection, `write=true` means write-protection and `write=false` means readwrite-protection.
2. In mutex analysis itself, `write=true` means a global write and `write=false` means a global read (only!).

I don't think this causes any bugs but it's still rather confusing because there are no readwrite accesses to globals (separate read and write access events are emitted).
As-is this PR makes things more consistent and correct, although readwrite accesses are never emitted, so this could be encoded more precisely with polymorphic variants.


sim642:
Closing in favor of #1788.

PR #1635, opened 2024-11-29
Linear Two Variable Equality Domain Refining Value Domain

DrMichaelPetter:
In this PR, lin2vareq interprets inequality guards as opportunities to refine value domains beyond the syntactical obvious intersection with bounds. Check the following example:

```
int x, y, z;
x = 3*y + 1; // a
z = 5*x + 7; // b
if (x>0) {
  __goblint_check( x >  0 );
  __goblint_check( y > -1 ); // A
  __goblint_check( z >  7 ); // B
}
```
Invariants A and B can assumed to hold due to assignments a and b establishing relations. Both invariants can not be tracked via value analysis.


DrMichaelPetter:
Screenshot_20241218_112713
Preliminary results look already promising, even on a 2GB 5min run. I still have to sort out some bugs in the svcomp-runs, so maybe it will become even better.

PR #1636, opened 2024-12-02
Narrowing of globals in TD

Red-Panda64:
# Description

Adds the capability of narrowing globals (and function entry unknowns) to TD. Contributions to each global unknown are tracked per source unknown. When an unknown x triggers a side-effect to y, this side-effect is applied to x's contribution to y by warrowing. The overall value of a global unknown is computed as the join over all contributions to said unknown.

# Configuration Options

- `ana.base.priv.protection.changes-only`: protection-based privatization won't produce spurious side-effects to protected unknowns. Otherwise, abstract values for globals that are refined by a guard edge may be written back to said global without a true write-access ("pseudo-writes"). This should have no benefit when narrowing of globals is disabled, as the pseudo-writes are already subsumed by the respective global. Implemented for base privatization only.
- `solvers.td3.narrow-globs.enabled`: enables the global narrowing scheme.
- `solvers.td3.narrow-globs.conservative-widen`: side-effects are only applied to contributions by widening if applying them with a join would affect the overall value of a global.
- `solvers.td3.narrow-globs.immediate-growth`: irrespective of this option, side-effects from x to y are accumulated throughout an evaluation of x to ensure soundness. This guarantees that the final side-effect will subsume all prior ones from this evaluation. This option permits the immediate application of accumulated values during constraint evaluation, provided they are not subsumed by the current contribution of x to y. Otherwise, the accumulated value is applied after x is fully evaluated.
- `solvers.td3.narrow-globs.narrow-gas`: limit the number of switches between widening and narrowing per contribution. This is necessary to enforce termination.
- `solvers.td3.narrow-globs.eliminate-dead`: if an unknown x triggers no side-effect to y but did so previously, its contribution is narrowed with bottom. Should y become bottom and be a function entry unknown, the exit unknown will be queried, propagating bottom throughout the function body.

# Modifications to td3.ml
- new structure `divided_side_effects`: to track current values of contributions, last combination mode (widen or narrow), and counter for remaining mode switches
- new structure `narrow_globs_start_values`: explicitly track initial values for unknowns to ensure they are accounted for when the join over all contributions is computed
- new structure `prev_sides` maps unknowns to side-effected targets on the last evaluation

# Modifications to basePriv.ml
Optionally track "true writes" for protection based privatizations: in addition to CPA tracking the most precise available value per global variable, the new data structure tracks only values that were written. In particular, this excludes refined values that were obtained via guard edges (e.g. inside an `if (g < 10)` block), which are also stored in CPA for increased precision but do not correspond to real writes.

# Modifications to constraint system
Constraint system can provide a hint in the form of the "postmortem" function, that returns for some side-effect target unknown that has become bottom, which local unknowns might now be worth re-evaluating. The purpose of this is to allow the re-evaluation of function exit nodes, when their entry becomes bottom, in order to propagate this bottom value and remove superfluous side effects along the way.


Red-Panda64:
For now, this is only a draft because I will see if I can't extract the changes to td3.ml and put them behind a reusable module.

Red-Panda64:
Although it should be possible to extract this into a module, doing so is a bit of a hassle because of the way td3's current side-effect widening strategies are very closely tied to the internals of td3. In any case, it makes no sense to try and do this until #1442 has been accepted or rejected.

michael-schwarz (approved):

src/analyses/basePriv.ml
```diff
@@ -943,16 +979,22 @@ struct
...
-             If global is read-protected by multiple locks,
-             then inner unlock shouldn't yet publish. *)
-          if not Param.check_read_unprotected || is_unprotected_without ask ~write:false x m then
-            sideg (V.protected x) v;
-          if atomic then
-            sideg (V.unprotected x) v; (* Publish delayed unprotected write as if it were protected by the atomic section. *)
-
+          (* Only apply sides for values that were actually written to globals!
+             This excludes invariants inferred through guards. *)
+          begin match D.precise_side x v st.priv with
```
sim642, line 984: This seems like a separate improvement. I'm a bit unsure about it.

I first thought this should already be covered, but it's not because the implementation doesn't have the $$W$$ set that's mentioned in the traces paper. Instead we've abused $$\mathrm{dom}(\texttt{CPA})$$ for it, but that indeed doesn't cut it. An `invariant` write is supposed to add the variable to $$P$$, but not $$W$$, but the map domain based implementation makes it impossible.

Wouldn't it be simpler to just add the $$W$$ set into the implementation?
In that case I also don't see much reason to keep the current approach. I think protection is the only privatization which abuses the domain of `CPA` for some other purpose.

src/constraint/constrSys.ml
```diff
@@ -48,6 +48,8 @@ sig
 
   val sys_change: (v -> d) -> v sys_change_info
   (** Compute incremental constraint system change from old solution. *)
+
+  val postmortem: v -> v list
```
sim642, line 52: The description says:

> Constraint system can provide a hint in the form of the "postmortem" function, that returns for some side-effect target unknown that has become bottom, which local unknowns might now be worth re-evaluating. The purpose of this is to allow the re-evaluation of function exit nodes, when their entry becomes bottom, in order to propagate this bottom value and remove superfluous side effects along the way.

Doesn't the existing solver dependency information already have this? This essentially is a static definition of (a transitive subset of) `infl` which kind of goes against our whole dynamic dependency discovery approach.

src/solver/td3.ml
sim642, line 1: Regarding https://github.com/goblint/analyzer/pull/1636#issuecomment-2514734278:
PR #1442, which among other things refactors side effect widening out from the solver, is now merged. So most of this should also hopefully go there instead of being in the middle of TD3 like the other side widen strategies were.

src/solver/td3.ml
```diff
@@ -340,7 +377,34 @@ module Base =
...
+                let changed = HM.create 0 in
+                Fun.protect ~finally:(fun () -> (
+                      if narrow_globs_eliminate_dead then begin
+                        let prev_sides_x = HM.find_option prev_sides x in
+                        Option.may (VS.iter (fun y ->
+                            if not @@ HM.mem acc y then begin
+                              ignore @@ divided_side D_Narrow x y (S.Dom.bot ());
+                              if S.Dom.is_bot @@ HM.find rho y then
+                                let casualties = S.postmortem y in
+                                List.iter (fun x -> solve x Widen) casualties
```
sim642, line 391: This has similar connotations to #365: calling `solve` like this on an `x` which is influenced by the current `y` turns it into a forward-propagating solver (in part), as opposed to remaining fully demand-driven.
Like with #365, this seems like it could ask to solve things which maybe nothing depends on any more, just doing extra work.

tests/regression/84-widening_gas/01-side_parallel.c
sim642, line 1: This PR shouldn't move the tests of #1442, rather it should put its own new tests into 84-dividedsides or something then.

src/solver/td3.ml
michael-schwarz: This is precisely why that is there! I'll share the draft paper with you that explains this once we have a little more time next week.

src/constraint/constrSys.ml
michael-schwarz: It is indeed some subset of the transitive subset, but it allows limiting what is re-queried rather than calling solve for all last dependencies, which is potentially very expensive.

For those with access to our drafts at https://github.com/tum-cit-pl, this is discussed in https://github.com/tum-cit-pl/combatting-precision-loss Section 4.3.

In some sense this is a hint of the constraint system to the solver in the sense of "might now be worth re-evaluating", which is a subjective thing.

We're open to changing the name but a functionality like this is needed for our paper.

src/solver/td3.ml
michael-schwarz: This is closely tied to the remark above: The idea is that for those unknowns in `postmortem` we are willing to pay this additional penalty in order to collect dead side-effects and potentially improve the values at globals. Dead code will be pruned at the end by `reach` but that doesn't shrink side-effects.

For those with access to our drafts at https://github.com/tum-cit-pl, this is discussed in https://github.com/tum-cit-pl/combatting-precision-loss Section 4.3, where we also have the remark that forward propagating solvers exhibit this behavior out of the box.

src/solver/td3.ml
michael-schwarz: I'm not sure if this is where we want to pull it out to. It seems that the `SideWpointSelect` module currently does one thing, and the choice of **when** to widen is unrelated to whether globals should be narrowed again. Also narrowing on globals needs non-local changes to the solvers, such as things needing to be wrapped around eval etc.
I'm skeptical whether mashing everything together will give us something that is easier to understand than what we currently have.

src/analyses/basePriv.ml
michael-schwarz: The difference likely is that this only side-effects the actually written values whereas #1693 side-effects the joined value, so this here is somehow a cheap way to get what kind of amounts to path-sensitivity in `W`.

I think it may be worth paying this additional price, it was critical in the experimental evaluation here iirc.

michael-schwarz:
@sim642 Is there anything you'd still like to see done before merging this?

src/solver/td3.ml
sim642: The "where" isn't the difficult question: it could easily be a separate module if it doesn't fit the existing one.

But still, since the beginning (https://github.com/goblint/analyzer/pull/1636#issuecomment-2514734278) the plan was to have it reasonably separate, which was also (at least one) motivation behind #1442.

sim642:
I don't really like how this is going against our long-term goal of not having an uncomprehensively massive `td3.ml`. It was a big reason for reverting the aborting feature (#364), even though it had significant benefit.
Or if we want to allow `td3.ml` as a dumping ground for all sorts of new things without regard to modularization, then this is something we should re-decide collectively, because it will make it increasingly difficult to add new features if everything is super-coupled.

src/solver/td3.ml
michael-schwarz: I have now factored it out.

michael-schwarz:
@sim642: I have now pulled things out into a separate module. Let me know if you prefer this version!

michael-schwarz:
@sim642: Any further comments here? Otherwise I suggest merging now that the paper is accepted at PLDI.

tests/regression/84-dividedsides/21-dead-escape.c
```diff
@@ -0,0 +1,39 @@
...
+  }
+  i++;
+
+  // The unknowns for protected:i etc. exist.
+  // Also, i is in the global set of escapees.
+  // However, locally i is known not to have escaped,
+  // so none of these unknowns are queried and this check
+  // succeeds whether eliminate-dead is on or not.
+  __goblint_check(i == 1);
+  // Paradoxically, a != &i is not known without eliminate-dead.
```
sim642, line 25: So is there an independent issue that should be opened here?

src/analyses/basePriv.ml
sim642: I am a bit hesitant about having this tangentially-related change in the same PR, but this extra abstraction is clean enough that it's OK to have around.
I currently don't have the time to more deeply think about this and #1693.

tests/regression/84-dividedsides/14-widening-thresholds.c
```diff
@@ -0,0 +1,29 @@
+// PARAM: --enable ana.int.interval --enable ana.int.interval_threshold_widening --set ana.int.interval_threshold_widening_constants comparisons --enable solvers.td3.narrow-globs.enabled
+#include <pthread.h>
+
+//TODO: this tests that threshold widening also narrows correctly. This should not be in this folder
```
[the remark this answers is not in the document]
michael-schwarz: I think it does, so it makes sense to have it here.

tests/regression/84-dividedsides/21-dead-escape.c
michael-schwarz: I'm not sure how paradoxical it really is. The pointer has escaped at some point and was recorded at the global.
We don't filter points-to-sets read from globals to exclude things which are locally known to not have escaped. I don't think this is particularly common, and iterating and querying for each lval is probably just expensive.

src/solver/td3UpdateRule.ml
```diff
@@ -0,0 +1,242 @@
...
+(** Inactive *)
+
+module Inactive:S =
+  functor (S:EqConstrSys) ->
+  functor (HM:Hashtbl.S with type key = S.v) ->
+  functor (VS:Set.S with type elt = S.v) ->
+  struct
+    type data = unit
+    type eq_wrapper = S.v -> ((S.v -> S.d -> unit) -> S.d) -> S.d
+    let active = false
```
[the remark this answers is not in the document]
michael-schwarz: I forgot to remove it.

src/solver/td3UpdateRule.ml
```diff
@@ -0,0 +1,242 @@
...
+    let copy_marshal _ = ()
+    let relift_marshal _ = ()
+
+    let register_start _ _ _  = ()
+
+    let get_wrapper ~solve_widen ~init ~stable ~data ~sides ~add_sides ~rho ~destabilize ~(side:?x:S.v -> S.v -> S.d -> unit) ~assert_can_receive_side =
+      (fun x eqx -> eqx (side ~x))
+  end
+
+module Narrow:S =
```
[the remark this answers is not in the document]
michael-schwarz: The separation into separate update rules and abstract garbage collection coming on top  in the paper makes some idealized assumptions and compromises which are not met in Goblint:

- Only one side-effect to each unknown during evaluation of a RHS (violated by having multiple paths)
- All side effects are delayed until the RHS is evaluated (decreases analyzer performance e.g. for calls because first solve would always yield bottom as side-effect to startpoint has not yet happened)


I thought about giving one file per update rule, but since we do not want to maintain the [Apinis et al.,'12] version, this would leave us with one file with a trivial module (default rule) and one large file that is basically all of the current file.

I think for now, one file is good. We can then start to split in case we decide to add more update rules.

src/constraint/constrSys.ml
michael-schwarz: `iter_vars` seems to want to iterate through vars, but this is a much simpler query here, that just produces for the unknown corresponding to the procedure start point the endpoint. I added an explanation that applies generally to side-effecting constraint systems.

tests/regression/84-dividedsides/14-widening-thresholds.c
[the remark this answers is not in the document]
sim642: In that case the comment shouldn't be a TODO about moving to a different folder anymore.

src/constraint/constrSys.ml
sim642: I think we just formulated it as an `iter`-style thing at the time to avoid allocating massive lists of unknowns in the incremental setting. `iter_vars` could probably also return a `Seq.t` to achieve a similar effect.

I just had a look at `VarQuery.t` and realized that the query itself isn't polymorphic over the type of the unknown, only the returned iterator is. This probably can be generalized like `Queries.t`, but it's a bigger change than I expected, so this might have to do for now.

sim642 (approved):
Other than one comment in a test (https://github.com/goblint/analyzer/pull/1636#discussion_r2049247513), I think this is now in good enough shape.

Issue #1551, opened 2024-08-02
`Mutex-Meet-TID` without clusters more precise than approach with clusters


michael-schwarz:
For the following small program:

~~~C
#include<pthread.h>
pthread_mutex_t c;
int d, e, f;

void b(void* arg);

void main(int argc, char *argv) {
  pthread_t t;
  e = pthread_create(&t, 0, b, &f);
  pthread_mutex_lock(&c);
  d++;
  pthread_mutex_unlock(&c);
  pthread_mutex_lock(&c);
}
~~~

~~~bash
./goblint --conf conf/traces-rel.json --set ana.activated[+] threadJoins --set ana.activated[+] apron --set ana.path_sens[+] threadflag --set ana.relation.privatization mutex-meet-tid-cluster12 1.c --sets exp.relation.prec-dump cluster.prec
./goblint --conf conf/traces-rel.json --set ana.activated[+] threadJoins --set ana.activated[+] apron --set ana.path_sens[+] threadflag --set ana.relation.privatization mutex-meet-tid 1.c --sets exp.relation.prec-dump tid.prec
./apronPrecCompare cluster.prec tid.prec
~~~

we get

> Apron Octagon(domain: Apron Octagon, privatization: PerMutexMeetPrivTID(cluster12,join)) less precise than Apron Octagon(domain: Apron Octagon, privatization: PerMutexMeetPrivTID(no-clusters,join))    (equal: 6, more precise: 0, less precise: 1, incomparable: 0, total: 7)

While such effects may happen due to W/N, it seems odd that it can be observed on such small programs. `b` having no implementation available seems crucial for this to happen, so it may be invalidation related somehow...

<details><summary>Detailed comparison</summary>
<p>
```
  Apron Octagon(domain: Apron Octagon, privatization: PerMutexMeetPrivTID(no-clusters,join)): [|d-1.>=0; -d+1.>=0; -d+e+2147483649.>=0; d+e+2147483647.>=0;
    e+2147483648.>=0; -d-e+2147483648.>=0; d-e+2147483646.>=0;
    -e+2147483647.>=0; -d+f+2147483649.>=0; d+f+2147483647.>=0;
    -e+f+2147483648.>=0; e+f+2147483648.>=0; f+2147483648.>=0;
    -d-f+2147483648.>=0; d-f+2147483646.>=0; -e-f+2147483647.>=0;
    e-f+2147483648.>=0; -f+2147483647.>=0|] (env: [|0> __daylight:int;
[...]

PR #1643, opened 2024-12-17
Relational MM: Cluster LMust

michael-schwarz:
Re-creates #1555 but targeting master this time.

Currently, `LMust` is unclustered even when clusters are employed. This leads to precision loss compared to the unclustered variant when:

- Values are published to a mutex `a` for some clusters only
- `a` is added to `LMust`
- Upon `lock(a)`, `V.mutex_inits` is not consulted for **any** cluster, as `a` is in `LMust`

However, `V.mutex_inits` should still be consulted for clusters that have not been published yet.

This also clusters `LMust` and adds a regression test for this behavior.

c.f. #1551

### DIFF
--- a/src/analyses/apron/relationPriv.apron.ml
+++ b/src/analyses/apron/relationPriv.apron.ml
@@ -761,11 +761,17 @@ module type ClusterArg = functor (RD: RelationDomain.RD) ->
 sig
   module LRD: Lattice.S
 
+  module Cluster: sig
+    include Printable.S
+  end
+
   val keep_only_protected_globals: Q.ask -> LockDomain.MustLock.t -> LRD.t -> LRD.t
   val keep_global: varinfo -> LRD.t -> LRD.t
 
   val lock: RD.t -> LRD.t -> LRD.t -> RD.t
-  val unlock: W.t -> RD.t -> LRD.t
+  val unlock: W.t -> RD.t -> LRD.t * (Cluster.t list)
+
+  val filter_clusters: LRD.t -> (Cluster.t -> bool) -> LRD.t
 
   val name: unit -> string
 end
@@ -775,6 +781,7 @@ module NoCluster:ClusterArg = functor (RD: RelationDomain.RD) ->
 struct
   open CommonPerMutex(RD)
   module LRD = RD
+  module Cluster = Printable.Unit
 
   let keep_only_protected_globals = keep_only_protected_globals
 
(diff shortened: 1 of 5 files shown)


sim642:
> However, `V.mutex_inits` should still be consulted for clusters that have not been published yet.

This PR is marked as a precision improvement, but why isn't this a soundness matter?

src/analyses/apron/relationPriv.apron.ml
```diff
@@ -761,11 +761,17 @@ module type ClusterArg = functor (RD: RelationDomain.RD) ->
 sig
   module LRD: Lattice.S
 
+  module Cluster: sig
+    include Printable.S
+  end
```
sim642, line 766:
```suggestion
  module Cluster: Printable.S
```

src/analyses/apron/relationPriv.apron.ml
```diff
@@ -761,11 +761,17 @@ module type ClusterArg = functor (RD: RelationDomain.RD) ->
...
+
   val keep_only_protected_globals: Q.ask -> LockDomain.MustLock.t -> LRD.t -> LRD.t
   val keep_global: varinfo -> LRD.t -> LRD.t
 
   val lock: RD.t -> LRD.t -> LRD.t -> RD.t
-  val unlock: W.t -> RD.t -> LRD.t
+  val unlock: W.t -> RD.t -> LRD.t * (Cluster.t list)
+
+  val filter_clusters: LRD.t -> (Cluster.t -> bool) -> LRD.t
```
sim642, line 774: I think the more conventional way is having the predicate first.

src/analyses/commonPriv.ml
```diff
@@ -268,7 +268,7 @@ struct
 
   (** Mutexes / globals to which values have been published, i.e. for which the initializers need not be read **)
   module LMust = struct
-    include SetDomain.Reverse (SetDomain.ToppedSet (LLock) (struct let topname = "All locks" end))
+    include SetDomain.Reverse (SetDomain.ToppedSet (Printable.Prod(LLock)(Cluster)) (struct let topname = "All locks" end))
```
sim642, line 271: This comment should also mention clusters now.

michael-schwarz:
> This PR is marked as a precision improvement, but why isn't this a soundness matter?

It in fact is. I added another regression test for which master currently is unsound.

sim642 (approved):

Issue #1558, opened 2024-08-09
Mutex-Meet unsound (function pointers get lost)


michael-schwarz:
This is a follow-up to #1557, which is extracted as the issue has nothing to do with relational analysis.

~~~C
#include<pthread.h>
#include<stdlib.h>
#include<stdio.h>
struct a {
  int (*b)();
};

struct a* t;
int m() {
    printf("m\n");
    int x =5;
}

void* j(void* arg) {};

void munge(struct a** a) {
    struct a r = **a;
    r.b();
}


void main() {
    pthread_t tid;
    pthread_create(&tid, 0, j, NULL);
    t = calloc(1, sizeof(struct a));
    t->b = &m;
    munge(&t);
}
~~~

For this program, Goblint with the `mutex-meet` privatization for base determines `m` to be uncalled which does not hold in the concrete. 

We, however do produce a warning:

> [Warning][Imprecise][Call] Function pointer *(r.b) may contain unknown functions. (2.c:18:5-18:10)
> [Warning][Unsound][Call] No suitable function to be called at call site. Continuing with state before call. (2.c:18:5-18:10)

michael-schwarz:
The example can be further simplified to

~~~C
#include<pthread.h>
#include<stdlib.h>
#include<stdio.h>
struct a {
  void (*b)();
};

struct a* t;
void m() {
    printf("m\n");
}

void* j(void* arg) {};

void main() {
    pthread_t tid;
    pthread_create(&tid, 0, j, NULL);
    t = calloc(1, sizeof(struct a));
    t->b = &m;
    struct a r = *t;
    r.b();
}
~~~

PR #1644, opened 2024-12-17
Make `update_offset` idempotent for blobs

michael-schwarz:
**Recreates #1559 which at the time was only merged into the branch for my thesis.**

Otherwise, a publish lead to a Blob value being placed inside a zero-inited Blob, which lead to value Top inside the blob, which caused function calls to be lost.

See #1558 which shows how Goblint without this fix is unsound.

### DIFF
--- a/src/cdomain/value/cdomains/valueDomain.ml
+++ b/src/cdomain/value/cdomains/valueDomain.ml
@@ -1001,10 +1001,10 @@ struct
             else
               mu (Blob (join x (do_update_offset ask x offs value exp l' o' v t), s, zeroinit))
           end
-        | Blob (x,s,zeroinit), _ ->
+        | Blob (x,s,zeroinit), `NoOffset -> (* `NoOffset is only remaining possibility for Blob here *)
           begin
-            match offs, value with
-            | `NoOffset, Blob (x2, s2, zeroinit2) -> mu (Blob (join x x2, ID.join s s2, zeroinit))
+            match value with
+            | Blob (x2, s2, zeroinit2) -> mu (Blob (join x x2, ID.join s s2, zeroinit))
             | _ ->
               let l', o' = shift_one_over l o in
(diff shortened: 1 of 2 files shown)


src/cdomain/value/cdomains/valueDomain.ml
```diff
@@ -1003,26 +1003,29 @@ struct
...
-                    ) blob_size_opt
-                | _ -> false
-              end
-            in
-            if do_strong_update then
-              Blob ((do_update_offset ask x offs value exp l' o' v t), s, zeroinit)
-            else
-              mu (Blob (join x (do_update_offset ask x offs value exp l' o' v t), s, zeroinit))
+            match offs, value with
+            | `NoOffset, Blob (x2, s2, zeroinit2) -> mu (Blob (join x x2, ID.join s s2,ZeroInit.join zeroinit zeroinit2))
```
sim642, line 1007: The existing outer match has `offs` as the second value (which is ignored in this case). Before this case in the outer match, `Blob` with `Index` and `Field` are already handled, so it should be the case that `offs` will always be `NoOffset` here anyway.
That would get rid of my concern about `NoOffset` somehow being special here.

sim642, line 1007: `ZeroInit` is a fake lattice so this join will raise for different things. Other blob-related hackery here somehow picks one over the other, possibly because they should be same in such cases? If so, then it might be better to have an `assert` here (to make our assumption explicit) and not do the join at all.

michael-schwarz: That's true! I have simplified the code now!

sim642 (approved):

sim642:
Turns out this accidentally fixed a TODO in our regression suite as well: ff0e24c468f83ea95f20efbe49edf665ce1c8096.

PR #1645, opened 2024-12-17
PrivPrecCompare: Add `bot_in_blob_leq_bot` so `bot` and `Blob(bot)` are considered equal

michael-schwarz:
This was needed to not report spurious precision differences for the benchmarks in my thesis.
If we want this on the main branch, we can merge, otherwise we can close it.

### DIFF
--- a/src/cdomain/value/cdomains/valueDomain.ml
+++ b/src/cdomain/value/cdomains/valueDomain.ml
@@ -561,7 +561,13 @@ struct
     | (_, Top) -> true
     | (Top, _) -> false
     | (Bot, _) -> true
-    | (_, Bot) -> false
+    | (x, Bot) ->
+      if !AnalysisState.bot_in_blob_leq_bot then
+        match x with
+        | Blob (x,s,o) -> leq x Bot
+        | _ -> false
+      else
+        false
     | (Int x, Int y) -> ID.leq x y
     | (Float x, Float y) -> FD.leq x y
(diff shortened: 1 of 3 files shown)


sim642:
I haven't thought about it much or checked if this would completely break something, but I wonder if such thing could/should be non-optional. I think `ValueDomain.join` has cases for joining blobs with non-blobs by lifting the joining into the blob, etc. Perhaps something similar should be in `leq` unconditionally to maintain its algebraic consistency with `join`, etc.

I suspect domain quickchecking doesn't really work on `ValueDomain`, so such consistency might not be too easy to check that way.

michael-schwarz:
I am also not sure if one wants to have this in general, but I think having it in the compare scripts and having the option to dis/enable it, seems like it is generally a step in the right direction?

michael-schwarz:
Do we want to merge this or close it? I'm fine with either!

sim642:
I was wondering about these `join` cases:
https://github.com/goblint/analyzer/blob/99df9da1e812a1adecf631a90020251033d46678/src/cdomain/value/cdomains/valueDomain.ml#L603-L604
But they're actually after these:
https://github.com/goblint/analyzer/blob/99df9da1e812a1adecf631a90020251033d46678/src/cdomain/value/cdomains/valueDomain.ml#L589-L590

So the special cases don't even trigger for such things. Indeed, then this isn't what should normally happen, but I'm ok with having it for comparison.

sim642 (approved):

PR #1646, opened 2024-12-17
Relational: Use same invalidation strategy as base

### LINKED ISSUE #1535: `Relational Analyses`: Deep invalidation for library functions havocs all variables (also those that don't have their address taken)
After investigating why almost all relational asserts we have shown for the ESOP paper are now out of reach, it seems it has to do with our invalidation mechanism.

We sometimes specify that things need to be written "deeply", i.e., every piece of memory that is reachable from somewhere needs to be invalidated. This is, e.g., the case for the first argument of `fscanf`.

When this is an unknown argument, such as `stdin`, our relational analysis responds by invalidating **all** relational information it has about **any** variable.

https://github.com/goblint/analyzer/blob/502cf9f7b7ad704025270a6b8a6df1570927783a/src/analyses/apron/relationAnalysis.apron.ml#L483-L489

This includes also variables which do not have their address taken and can thus never be pointed to by such pointers.
(N.B. There is a separate option `InvalidateGlobals` for calls to unknown functions that we assume may modify globals).

~~~C
// SKIP PARAM: --set ana.activated[+] apron --set sem.int.signed_overflow assume_none
#include <pthread.h>
#include <goblint.h>
#include <stdio.h>

int debug;
int other;

int main() {
  int top;

  // Needed so Base & DefExc doesn't find this information because it invalidates less
  if(top) {
    debug = 3;
  }

  fscanf(stdin, "%d", &other);
 
   // Fails as debug is invalidated
  __goblint_check(debug <= 3);
  return 0;
}
~~~

Originally, these TopPointers were filtered out before they reached this analysis, but this has changed recently:
(c.f. https://github.com/goblint/analyzer/pull/1142 and https://github.com/goblint/analyzer/commit/83978e953dbbf7d6a8c920d071b380bbd1441637).

**As a consequence, our relational analyses are very prone to havoccing all information they track even upon very normal calls to library functions, that, e.g., print something to stdout.**

Adding a filter `|> List.filter (fun v -> v.vaddrof)` above already restores some of the precision, but we should still aim for a more precise version (at least for sensible things such as writing to `stdout` or reading from `stdin`).

Interestingly, Base seems to take a more measured approach here, and succeeds in showing these things.

On top of being crucial for my thesis, this might also be of concern for our witness exchange project with Freiburg.

michael-schwarz:
Closes #1535


src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -469,10 +469,9 @@ struct
       | None -> None
       | Some st ->
         let ad = ask.f (Queries.ReachableFrom e) in
-        if Queries.AD.is_top ad then
-          None
-        else
-          Some (Queries.AD.join ad st)
+        (* See https://github.com/goblint/analyzer/issues/1535 *)
+        let ad = Queries.AD.remove UnknownPtr ad in
+        Some (Queries.AD.join ad st)
```
sim642, line 474: Since the `None` case is now gone, this could be simplified to not have an `option` at all because it will always be `Some`.

Also, is it even necessary to remove `UnknownPtr` then? `forget_reachable` will fold over this, but only handle known addresses with `Addr` and ignore `UnknownPtr` anyway.
That's also what base does: the `UnknownPtr` is there (and warned about), but nothing is written.

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -674,21 +679,19 @@ struct
 
   let threadenter man ~multiple lval f args =
     let st = man.local in
+    (* TODO: HACK: Simulate enter_multithreaded for first entering thread to publish global inits before analyzing thread.
+       Otherwise thread is analyzed with no global inits, reading globals gives bot, which turns into top, which might get published...
+       sync `Thread doesn't help us here, it's not specific to entering multithreaded mode.
+       EnterMultithreaded events only execute after threadenter and threadspawn. *)
+    if not (ThreadFlag.has_ever_been_multi (Analyses.ask_of_man man)) then
+      ignore (Priv.enter_multithreaded (Analyses.ask_of_man man) man.global man.sideg st);
```
sim642, line 687: Why is this moved out? Base only does it for non-unknown thread functions.

tests/regression/46-apron2/98-invalidate-more.c
```diff
@@ -0,0 +1,22 @@
...
+
+  // Needed so Base & DefExc doesn't find this information because it invalidates less
+  if(top) {
+    debug = 3;
+  }
+
+  fscanf(stdin, "%d", &other);
+
+   // Fails as debug is invalidated
+  __goblint_check(debug <= 3);
```
sim642, line 20: The comment is outdated now?

src/analyses/apron/relationAnalysis.apron.ml
michael-schwarz: The logic of why it is needed is the same here as it is for the case where a known thread is created: One should not go into multi-threaded mode without notifying the privatization that this switch happened to ensure everything that should be published is published.

Maybe base has some reason this can be avoided, but I'm not really sure why that should be the case. We may want to make an issue to investigate this.

michael-schwarz: Also see the last few comments in https://github.com/goblint/analyzer/issues/1551

michael-schwarz: https://github.com/goblint/analyzer/issues/1551#issuecomment-2267519495

sim642: The one test in this PR doesn't cover this, does it?

I tried the very simple thing of creating a thread from unknown function pointer and that seems to work with base. That's because threadflag analysis emits the `EnterMultiThreaded` event which is handled by base (and relational).
The comment for this hack is about the globals values that get used when analyzing the created thread. For an unknown function pointer, there's no thread body to analyze anyway, so it's not clear to me when it would make a difference.

michael-schwarz: The test covers it in some sense. We mean to invalidate `e` and output 

> [Info][Imprecise] Invalidating expressions: & e (1.c:11:3-11:35)

But this invalidation does not take and we still claim to be able to prove things about `e`.

michael-schwarz: Edit: That case is listed in the linked issue.

sim642: Do you mean this?
```c
#include<pthread.h>
pthread_mutex_t c;
int d, e, f;

void b(void* arg);

void main(int argc, char *argv) {
  pthread_t t;
  e = pthread_create(&t, 0, b, &f);
  pthread_mutex_lock(&c);
  d++;
  pthread_mutex_unlock(&c);
  pthread_mutex_lock(&c);
}
```
Where or what are we claiming to prove about `e`? There aren't any assertions. Can you add it as a test?

I traced `sol` and `sol2` on it and what this changes is that we no longer have `-e+2147483647.>=0`, `e+2147483648.>=0` and all derived octagon constraints from that. But these are just type bounds, so they shouldn't be unsound to begin with.

The point of this PR is to unify logic between base and relation analyses, but this move introduces a new difference. If the old place is really wrong, then we should also fix it in base.
I want to avoid another issue like #1535 down the line where the inconsistency comes up and there's no understanding of why they're different.

michael-schwarz: Sure, I guess I should have linked to this comment: https://github.com/goblint/analyzer/issues/1551#issuecomment-2265205198

> Relying on YAML might make debugging easier. There we get
> 
> > ./goblint --conf conf/traces-rel.json --set ana.activated[+] threadJoins --set ana.activated[+] apron --set ana.path_sens[+] threadflag --set ana.relation.privatization mutex-meet-tid 1.c --sets exp.relation.prec-dump cluster.prec --html --enable witness.yaml.enabled --set witness.yaml.path 1.yml
> 
> > ./goblint --conf conf/traces-rel.json --set ana.activated[+] threadJoins --set ana.activated[+] apron --set ana.path_sens[+] threadflag --set ana.relation.privatization mutex-meet-tid-cluster12 1.c --set witness.yaml.validate 1.yml
> 
> ```
> [Warning][Witness] invariant unconfirmed: (2147483648LL + (long long )e) + (long long )f >= 0LL (1.c:14:1)
> [Warning][Witness] invariant unconfirmed: (2147483648LL - (long long )e) + (long long )f >= 0LL (1.c:14:1)
> [Warning][Witness] invariant unconfirmed: (2147483648LL + (long long )e) - (long long )f >= 0LL (1.c:14:1)
> [Warning][Witness] invariant unconfirmed: (2147483647LL - (long long )e) - (long long )f >= 0LL (1.c:14:1)
> ```
> 
> > [Info][Witness] witness validation summary:
> > confirmed: 32
> > unconfirmed: 4
> 
> The invariants provided here look slightly dubious to me as to whether they actually hold.

These invariants exploit that `e` is supposedly `0`, even though`e` and `f` have been invalidated.

michael-schwarz: I have now added a cram test for this & unified handling between base and relational.

michael-schwarz:
@sim642: This should now be ready for review.

The sporadic build fail is due to https://erratique.ch/ (where a package is loaded from) being a bit erratic (sorry :upside_down_face:) when it comes to being online.

PR #1648, opened 2024-12-18
Rename `ctx` -> `man`

### LINKED ISSUE #1438: Rename `ctx`
As brought up in GobCon, we might want to rename `ctx` because it's not a context in the analysis sense.

### Candidates
1. `man` – manager.
2. `gob` - Goblint (you are communicating with the system via these calls)

michael-schwarz:
This will conflict with absolutely everything. I would thus appreciate it if we can merge this speedily.

Closes #1438.

### DIFF
--- a/src/analyses/basePriv.ml
+++ b/src/analyses/basePriv.ml
@@ -57,7 +57,7 @@ end
 let old_threadenter (type d) ask (st: d BaseDomain.basecomponents_t) =
   (* Copy-paste from Base make_entry *)
   let globals = CPA.filter (fun k v -> is_global ask k) st.cpa in
-  (* let new_cpa = if !earlyglobs || ThreadFlag.is_multi ctx.ask then CPA.filter (fun k v -> is_private ctx.ask ctx.local k) globals else globals in *)
+  (* let new_cpa = if !earlyglobs || ThreadFlag.is_multi man.ask then CPA.filter (fun k v -> is_private man.ask man.local k) globals else globals in *)
   let new_cpa = globals in
   {st with cpa = new_cpa}
 
@@ -1802,7 +1802,7 @@ struct
   let write_global ?invariant ask getg sideg st x v = time "write_global" (Priv.write_global ?invariant ask getg sideg st x) v
   let lock ask getg cpa m = time "lock" (Priv.lock ask getg cpa) m
   let unlock ask getg sideg st m = time "unlock" (Priv.unlock ask getg sideg st) m
-  let sync reason ctx = time "sync" (Priv.sync reason) ctx
+  let sync reason man = time "sync" (Priv.sync reason) man
   let escape ask getg sideg st escaped = time "escape" (Priv.escape ask getg sideg st) escaped
   let enter_multithreaded ask getg sideg st = time "enter_multithreaded" (Priv.enter_multithreaded ask getg sideg) st
   let threadenter ask st = time "threadenter" (Priv.threadenter ask) st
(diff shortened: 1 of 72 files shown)


src/analyses/basePriv.ml
```diff
@@ -1802,7 +1802,7 @@ struct
   let write_global ?invariant ask getg sideg st x v = time "write_global" (Priv.write_global ?invariant ask getg sideg st x) v
   let lock ask getg cpa m = time "lock" (Priv.lock ask getg cpa) m
   let unlock ask getg sideg st m = time "unlock" (Priv.unlock ask getg sideg st) m
-  let sync reason ctx = time "sync" (Priv.sync reason) ctx
+  let sync reason man = time "sync" (Priv.sync reason) man
```
sim642, line 1805: Now looking at it, this seems wrong to begin with: all privatization functions only get a few things from `man` as separate arguments:
```suggestion
  let sync ask getg sideg st reason = time "sync" (Priv.sync ask getg sideg st) reason
```
The type was the same, but the timing was measured from partial application, so it would've always been close to 0 I think.

src/analyses/extractPthread.ml
```diff
@@ -233,22 +233,22 @@ let promela_main : fun_name = "mainfun"
...
+let fun_ctx man f =
+  let man_hash =
+    match PthreadDomain.Ctx.to_int man with
     | Some i ->
       i |> i64_to_int |> Tbls.CtxTbl.get |> string_of_int
     | None ->
       "TOP"
   in
-  f.vname ^ "_" ^ ctx_hash
+  f.vname ^ "_" ^ man_hash
```
sim642, line 244: I think here `ctx` was something different (`PthreadDomain.Ctx`), so this (and perhaps other uses of it in this file) shouldn't be renamed.

Curiously, it wasn't renamed in the function name. Did you use something other than plain search and replace?

src/analyses/region.ml
```diff
@@ -76,113 +76,113 @@ struct
...
-  let combine_assign ctx (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask: Queries.ask) : D.t =
+  let combine_assign man (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask: Queries.ask) : D.t =
     match au with
     | `Lifted reg -> begin
-      let old_regpart = ctx.global () in
+        let old_regpart = man.global () in
       let regpart, reg = match lval with
         | None -> (old_regpart, reg)
         | Some lval -> Reg.assign lval (AddrOf (ReturnUtil.return_lval ())) (old_regpart, reg)
       in
```
sim642, line 149: There seems to be something weird about this indentation.

src/framework/analyses.ml
```diff
@@ -143,23 +143,13 @@ struct
...
-(* Experiment to reduce the number of arguments on transfer functions and allow
-   sub-analyses. The list sub contains the current local states of analyses in
-   the same order as written in the dependencies list (in MCP).
-
-   The foreign states when calling special_fn or enter are joined if the foreign
-   analysis tries to be path-sensitive in these functions. First try to only
-   depend on simple analyses.
-
-   It is not clear if we need pre-states, post-states or both on foreign analyses.
-*)
```
sim642, line 155: This comment was hilariously outdated. There should still be some documentation comment `(** ... *)` saying it's a "manager" and perhaps something more.

src/witness/z3/violationZ3.z3.ml
```diff
@@ -16,7 +16,7 @@ struct
     (* ("smt.core.minimize", "true"); *)
     (* ("sat.core.minimize", "true"); *)
   ]
-  let ctx = mk_context cfg
+  let man = mk_context cfg
```
sim642, line 19: The ones in this file should remain `ctx` which is some Z3 context.

src/analyses/extractPthread.ml
michael-schwarz: Yes, I tried to keep things that just happened to be named `ctx`, but must have missed a few instances.

sim642 (approved):

PR #1652, opened 2024-12-24
Analysis of `pthread_barrier`s

### LINKED ISSUE #1651: Analysis of `pthread_barrier` for race detection
One of the features we could additionally have for checking for data races could be a support for `pthread_barrier`.

~~~C
#include <stdio.h>
#include <pthread.h>

#define THREAD_NUMS 4
pthread_barrier_t barrier;

void *t0(void *param)
{
    lock(a); g = 3; unlock(a);
    pthread_barrier_wait(&barrier);
    printf("t0 ready\n");
}

void *t1(void *param)
{
    lock(a); g = 3; unlock(a);
    pthread_barrier_wait(&barrier);
    printf("t1 ready\n");
}

void *t2(void *param)
{
     lock(a); g = 3; unlock(a);
    pthread_barrier_wait(&barrier);
    printf("t2 ready\n");
}

int main(void)
{
    pthread_t t[3];

    pthread_barrier_init(&barrier, NULL, THREAD_NUMS);

    pthread_create(&t[0], NULL, t0, NULL);
    pthread_create(&t[1], NULL, t1, NULL);
    pthread_create(&t[2], NULL, t2, NULL);

    pthread_barrier_wait(&barrier);
    g = 5; //NORACE

    pthread_barrier_destroy(&barrier);

}
~~~

If one removes any of the `pthread_barrier_wait` calls, all successors to `pthread_barrier_wait` would be dead.

To this end, one would track locally: In each thread a must $$\underline M$$ and a may set  $$\overline M$$ of x for which `pthread_barrier_wait(x)` has been called.

 
Globally, one would track at a global unknown [x]:
 
 - The capacity passed to `pthread_barrier_init`
 - A set of unique thread ids that call `pthread_barrier_wait` at most once. (If any of these is non-unique or one thread calls wait several times, we go to top)

Then: 

- If the number of threads that may call `pthread_barrier_wait` is below the capacity (and all call it at most once and are unique), all successors of `pthread_barrier_wait` are dead.
- If  at an access `pthread_barrier_wait(x)` is known to have returned $$x \in \underline M$$, the above condition is known to hold and the other access is by a thread that has to call `pthread_barrier_wait(x)` for it to return, but $$x$$ is not in $$\overline M$$ there, these two accesses cannot race.
 
 It should be possible to cast this into a may-race-predicate enhanced digest Julian and I have discussed a little :wink:

michael-schwarz:
Makes use of the $$||_{\\mathcal{A}}: \\mathcal{A} \to \\mathcal{A} \to \\{ \textsf{false},\top \\}$$ predicate we want to define for (abstract) digests to provide a sound basis for all of our MHP stuff for races and give a principled account of what I added in #518 .

Interestingly, it even does so outside of the context of data races --- which may be a cute insight on top of putting the race analysis on solid foundations with this predicate (and hopefully finding a way to get `pthread_once` to also fit (potentially in a reduced fashion)).

Technically, it accumulates the `capacity` and `MHP` information for all calls to `wait` for each barrier at a global unknown, and checks upon a call to `wait` that there are at least `min(capacity)-1` other accesses where $$||_{\\mathcal{A}}$$ is true pairwise among these, as well as with the caller to `wait`.

TODO:

- [x] Use TIDs and Joins
- [ ] Use further abstract digests (such as must-lockset digest)
- [ ] Use information also for excluding races (beyond unreachability)
- [ ] Come up with a less expensive algorithm for checking there are `min(capacity)-1` elements where $$||_{\\mathcal{A}}$$  holds pairwise

Closes #1651.

### DIFF
--- a/src/analyses/pthreadBarriers.ml
+++ b/src/analyses/pthreadBarriers.ml
@@ -53,7 +53,26 @@ struct
             let count = Waiters.cardinal relevant_waiters in
             match capacity with
             | `Lifted c -> 
-              Z.geq (Z.of_int count) (BatOption.default Z.zero (Capacity.I.minimal c)) 
+              (* Add 1 as the thread calling wait at the moment will not be MHP with itself *)
+              let min_cap = (BatOption.default Z.zero (Capacity.I.minimal c)) in
+              if Z.leq min_cap Z.one then
+                true
+              else if min_cap = Z.of_int 2 && count = 1 then
+                true
+              else if Z.geq (Z.of_int (count + 1)) min_cap then
+                (* This is quite a cute problem: Do (min_cap-1) elements exist in the set such that 
+                  MHP is pairwise true? This solution is a sledgehammer, there should be something much
+                  better algorithmically (beyond just laziness) *)
+                let waiters = Waiters.elements relevant_waiters in
+                let min_cap = Z.to_int min_cap in
+                let lists = List.init (min_cap - 1) (fun _ -> waiters) in
+                let candidates = BatList.n_cartesian_product lists in
+                List.exists (fun candidate ->
+                  let pairwise = BatList.cartesian_product candidate candidate in
+                  List.for_all (fun (a,b) -> MHP.may_happen_in_parallel a b) pairwise
+                ) candidates
+              else
+                false
             | _ -> true
         in
         if may_run then
--- a/src/analyses/pthreadBarriers.ml
+++ b/src/analyses/pthreadBarriers.ml
@@ -1,0 +1,43 @@
+(** Must have waited barriers for Pthread barriers ([pthreadBarriers]). *)
+
+open GoblintCil
+open Analyses
+module LF = LibraryFunctions
+
+module Spec =
+struct
+  module Barriers = SetDomain.ToppedSet (ValueDomain.Addr) (struct let topname = "All barriers" end)
+  module MustBarriers = Lattice.Reverse (Barriers)
+
+  include Analyses.IdentitySpec
+  module V = VarinfoV
+
+  let name () = "pthreadBarriers"
+  module D = Lattice.Prod (Barriers) (MustBarriers)
+
+  include Analyses.ValueContexts(D)
+
+  let possible_vinfos (a: Queries.ask) barrier =
+    Queries.AD.to_var_may (a.f (Queries.MayPointTo barrier))
+
+  let special man (lval: lval option) (f:varinfo) (arglist:exp list) : D.t =
+    let desc = LF.find f in
+    match desc.special arglist with
+    | BarrierWait barrier ->
+      let may, must = man.local in
+      let barriers = possible_vinfos (Analyses.ask_of_man man) barrier in
+      let may = List.fold_left (fun may a -> Barriers.add (ValueDomain.Addr.of_var a) may) may barriers in
+      let must = match barriers with
+      | [a] -> Barriers.add (ValueDomain.Addr.of_var a) must
+      | _ -> must
+      in
+      (may, must)
+    | _ -> man.local
+
+  let startstate v = (Barriers.empty (), Barriers.empty ())
+  let threadenter man ~multiple lval f args = [man.local]
+  let exitstate  v = (Barriers.empty (), Barriers.empty ())
+end
+
+let _ =
+  MCP.register_analysis (module Spec : MCPSpec)
--- a/src/analyses/pthreadBarriers.ml
+++ b/src/analyses/pthreadBarriers.ml
@@ -6,11 +6,21 @@ module LF = LibraryFunctions
 
 module Spec =
 struct
-  module Barriers = SetDomain.ToppedSet (ValueDomain.Addr) (struct let topname = "All barriers" end)
-  module MustBarriers = Lattice.Reverse (Barriers)
+  module Barriers = struct
+    include SetDomain.ToppedSet (ValueDomain.Addr) (struct let topname = "All barriers" end)
+    let name () = "mayBarriers"
+  end
+
+  module MustBarriers = struct
+    include Lattice.Reverse (Barriers)
+    let name () = "mustBarriers" 
+  end
 
   include Analyses.IdentitySpec
   module V = VarinfoV
+
+  module Waiters = SetDomain.ToppedSet (MHP) (struct let topname = "All MHP" end)
+  module G = Lattice.Prod (Queries.ID) (Waiters)
 
   let name () = "pthreadBarriers"
   module D = Lattice.Prod (Barriers) (MustBarriers)
@@ -32,6 +42,12 @@ struct
       | _ -> must
       in
       (may, must)
+    | BarrierInit { barrier; count } ->
+      let count = man.ask (Queries.EvalInt count) in
+      let publish_one b = man.sideg b (count, Waiters.bot ()) in 
+      let barriers = possible_vinfos (Analyses.ask_of_man man) barrier in
+      List.iter publish_one barriers;
+      man.local
     | _ -> man.local
 
   let startstate v = (Barriers.empty (), Barriers.empty ())
--- a/src/analyses/pthreadBarriers.ml
+++ b/src/analyses/pthreadBarriers.ml
@@ -16,11 +16,14 @@ struct
     let name () = "mustBarriers" 
   end
 
+  module Capacity = Queries.ID
+
   include Analyses.IdentitySpec
   module V = VarinfoV
 
-  module Waiters = SetDomain.ToppedSet (MHP) (struct let topname = "All MHP" end)
-  module G = Lattice.Prod (Queries.ID) (Waiters)
+  module TID = ThreadIdDomain.Thread
+  module Waiters = SetDomain.ToppedSet (TID) (struct let topname = "All MHP" end)
+  module G = Lattice.Prod (Capacity) (Waiters)
 
   let name () = "pthreadBarriers"
   module D = Lattice.Prod (Barriers) (MustBarriers)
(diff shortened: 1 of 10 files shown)


michael-schwarz:
OS X CI does OS X things, I'll just ignore that for now.

michael-schwarz:
>  Come up with a less expensive algorithm for checking there are `min(capacity)-1` elements where $$||_{\\mathcal{A}}$$  holds pairwise ?

Does anyone have an idea how this could be done? I'm not sure if one can do better complexity-wise, but it should be possible at least w.r.t. the constants.

michael-schwarz:
I talked it through with @DrMichaelPetter and we came up with some dynamic programming-like algorithm that hopefully performs better practically, but neither of us see how one can do better complexity wise.

michael-schwarz:
Apparently mathematicians call this problem the `Admissible Subset Selection` Problem. But looking around a bit, I have not found an algorithm that would help solve the problem.

michael-schwarz:
This now uses `Seq` to hopefully avoid constructing the full product every single time.

jprotopopov-ut:
> By the way, @jprotopopov-ut was doing some Linux kernel stubbing using barriers and I mentioned this PR, but I don't know if this ended up being used/useful there.

It did not work properly, but I did not spend any time to investigate the reasons and instead ended up implementing ad-hoc barriers using a set of atomic flags

michael-schwarz:
What do you mean it wasn't working properly? Was it unsound (would be worth investigating) or just not precise enough for your case (that is expected from time to time)?

.github/workflows/docs.yml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Install dependencies
         run: opam install . --deps-only --locked --with-doc
 
+      - name: Install os gem for operating system detection
+        run: sudo gem install os
+
```
sim642: The docs workflow shouldn't need this because it doesn't run the tests.

scripts/update_suite.rb
```diff
@@ -2,6 +2,7 @@
 
 # gobopt environment variable can be used to override goblint defaults and PARAMs
 
+require 'os'
```
sim642: If this needs to be separately installed, then we might have to do something about it for running the tests in opam-repository CI.
Maybe detect the presence of the `os` gem and skip the tests if it's missing (and output a warning). That's kind of what happens with the `parallel` gem: the script still works without.

tests/regression/86-barrier/01-simple.c
```diff
@@ -0,0 +1,26 @@
...
+    
+    pthread_barrier_init(&barrier, NULL, 2);
+    
+    if(top) {
+        pthread_barrier_wait(&barrier);
+        // Unreachable
+        i = 1;
+    }
+
+    __goblint_check(i == 0);
```
sim642, line 25: There should be `#include <goblint.h>` in such tests.

tests/regression/86-barrier/05-mhp-not-enough.c
```diff
@@ -0,0 +1,45 @@
...
+
+    return NULL;
+}
+
+int main(int argc, char const *argv[])
+{
+    int top;
+    int i = 0;
+
+    pthread_barrier_init(&barrier, NULL, 3);
```
sim642, line 26: In concrete, 2 would also be enough, right? It's just that the analysis is not precise enough to rule things out in that case.
If that's the case, then it might be worth mentioning.

src/analyses/pthreadBarriers.ml
```diff
@@ -0,0 +1,167 @@
...
+  module TID = ThreadIdDomain.ThreadLifted
+
+  module MHPplusLock = struct
+    module Locks = LockDomain.MustLockset
+
+    include Printable.Prod (MHP) (Locks)
+    let name () = "MHPplusLock"
+
+    let mhp (mhp1, l1) (mhp2, l2) =
+      MHP.may_happen_in_parallel mhp1 mhp2 && Locks.is_empty (Locks.inter l1 l2)
```
sim642: This looks a bit like it's duplicating the logic we also do for checking races/deadlocks via `may_race`.
The `may_race` name is a bit misleading, but perhaps this `MHPplusLock` could just be `MCPAccess.A` and this check is just a `MCPAccess.A.may_race` check. The deadlock analysis already works this way.

From the tests I already find it impressive how much locks and MHP play together with barriers, but generalizing to `may_race` might even allow going a step further: use one barrier to exclude MHP for another barrier check, i.e. to have them jointly and transitively benefit from each other.

src/analyses/pthreadBarriers.ml
```diff
@@ -0,0 +1,167 @@
...
+
+  let name () = "pthreadBarriers"
+
+  module MustObserved = MapDomain.MapTop_LiftBot (TID) (MustBarriers)
+  module D = Lattice.Prod (Barriers) (MustObserved)
+
+  include Analyses.ValueContexts(D)
+
+  let possible_vinfos (a: Queries.ask) barrier =
+    Queries.AD.to_var_may (a.f (Queries.MayPointTo barrier))
```
sim642, line 47: This is dropping offsets. Can't this cause issues?
I'm thinking `s.b1` and `s.b2` could be different barriers, but here they reduce down to `s` and would almost be treated as one barrier, no?

src/analyses/pthreadBarriers.ml
```diff
@@ -0,0 +1,167 @@
...
+              | `Lifted c ->
+                let count = Waiters.cardinal relevant_waiters in
+                (* Add 1 as the thread calling wait at the moment will not be MHP with itself *)
+                let min_cap = (BatOption.default Z.zero (Capacity.I.minimal c)) in
+                if Z.leq min_cap Z.one then
+                  (may, must)
+                else if Z.geq (Z.of_int (count + 1)) min_cap then
+                  (* This is quite a cute problem: Do (min_cap-1) elements exist in the set such that
+                      MHP is pairwise true? This solution is a sledgehammer, there should be something much
+                      better algorithmically (beyond just laziness) *)
```
sim642: This sounds like looking for a (`min_cap` - 1)-clique in the MHP graph with the set of nodes. That seems to be exponential in `min_cap` though: https://en.wikipedia.org/wiki/Clique_problem#Cliques_of_fixed_size.

src/analyses/pthreadBarriers.ml
```diff
@@ -0,0 +1,167 @@
...
+                  let must =
+                    let waiters = Waiters.elements relevant_waiters in
+                    let min_cap = Z.to_int min_cap in
+                    let can_proceed_pred seq =
+                      let rec doit seq = match Seq.uncons seq with
+                        | None -> true
+                        | Some (h, t) -> Seq.for_all (MHPplusLock.mhp h) t && doit t
```
sim642: Can you try adding a tail-recursion attribute here?
The `doit` call should be tail-recursive, at least when done using `if`, but perhaps the OCaml compiler is smart enough to also consider that.

src/cdomains/mHP.ml
```diff
@@ -92,3 +92,8 @@ let may_happen_in_parallel one two =
     else
       true
   | _ -> true
+
+let may_be_non_unique_thread mhp =
+  match mhp.tid with
+  | `Lifted tid -> not (TID.is_unique tid)
```
sim642: Why is this `may_be` when the underlying thread ID is directly known?

src/domain/disjointDomain.ml
```diff
@@ -117,6 +117,7 @@ struct
       add e acc
     ) (empty ()) es
   let elements m = fold List.cons m [] (* no intermediate per-bucket lists *)
+  let to_seq m = fold Seq.cons m Seq.empty
```
sim642: Might be worth leaving a comment here that the implementations here aren't efficient: they eagerly fold over the set to construct the entire sequence, as opposed to lazily constructing the sequence.

tests/regression/86-barrier/05-mhp-not-enough.c
michael-schwarz: No, with two the code marked as unreachable can be reached. One call to `wait` is by the thread created on line 29, the other one is by `main` (assuming `top` is not `0`).

src/analyses/pthreadBarriers.ml
michael-schwarz: I added a remark in case we try to look at this again.

src/cdomains/mHP.ml
michael-schwarz: Which name would you suggest?

src/analyses/pthreadBarriers.ml
michael-schwarz: Yes, nice observation! It indeed causes imprecision, but only in the direction that is fine. The problem if we wanted to have this more fine-grained is that we would need to have logic about all the different ways in which different offsets may be used to access the same data.

tests/regression/86-barrier/05-mhp-not-enough.c
sim642: You're right, I think I missed the thread creation before the if and only noticed the one in the else branch.

src/cdomains/mHP.ml
sim642: Just `is_non_unique_thread` I guess. Or maybe even avoiding the flip and `is_unique_thread`.

src/analyses/pthreadBarriers.ml
michael-schwarz: It does to some degree, yes. I was a bit unsure how to obtain the cop-out with non-unique threads here, but on second thought this corresponds to checking whether the things can self-race. I have now made the modification.

The only thing that is a bit ugly is that we now track thread ids twice,  as there is no thread-safe way to extract them from one of these tuples. But maybe that's also not such a huge deal...

tests/regression/86-barrier/13-barrierception.c
```diff
@@ -0,0 +1,45 @@
...
+
+int g;
+int h;
+
+pthread_barrier_t barrier;
+pthread_barrier_t barrier2;
+pthread_mutex_t mutex;
+
+void* f1(void* ptr) {
+    g = 2; //NORACE
```
sim642: This seems irrelevant here, there's no other access of `g` anyway. The test is about the checks.

Issue #1591, opened 2024-10-03
Use sets for widening thresholds instead of lists


sim642:
When widening thresholds are collected from the program, we do use a `Set` to avoid duplicates, but then it's just converted to a list with `elements`:
https://github.com/goblint/analyzer/blob/bb6f9aa1a2a4f38d022112e1ab840a4bdb382286/src/cdomain/value/util/wideningThresholds.ml#L103-L108
Later, `IntDomain` uses this list by linearly searching for the next threshold `x` for `u`:
https://github.com/goblint/analyzer/blob/bb6f9aa1a2a4f38d022112e1ab840a4bdb382286/src/cdomain/value/cdomains/intDomain.ml#L563-L568

This linear search is $$O(n)$$ while the same lookup in binary search trees would be $$O(\log n)$$. And actually [`Set.Make.find_first_opt`](https://ocaml.org/manual/5.2/api/Set.Make.html#VALfind_first_opt) from the OCaml standard library can be used exactly for this purpose (because that's what `Set`s are internally).
So we could avoid the intermediate lists and do the more efficient search on sets directly. Although I doubt it will make a notable difference to Goblint performance as a whole.

PR #1654, opened 2025-01-10
Use sets for widening thresholds instead of lists

karoliineh:
Solves #1591

### DIFF
--- a/src/cdomain/value/cdomains/intDomain0.ml
+++ b/src/cdomain/value/cdomains/intDomain0.ml
@@ -560,17 +560,19 @@ module IntervalArith (Ints_t : IntOps.IntOps) = struct
     ResettableLazy.force ts
 
   let upper_threshold u max_ik =
-    let ts = find_thresholds WideningThresholds.upper_thresholds in
     let u = Ints_t.to_bigint u in
     let max_ik' = Ints_t.to_bigint max_ik in
-    let t = WideningThresholds.Thresholds.find_first_opt (fun x -> Z.compare u x <= 0 && Z.compare x max_ik' <= 0) ts in
-    BatOption.map_default Ints_t.of_bigint max_ik t
+    find_thresholds WideningThresholds.upper_thresholds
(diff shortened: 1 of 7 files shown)


src/cdomain/value/cdomains/intDomain0.ml
```diff
@@ -560,26 +555,30 @@ module IntervalArith (Ints_t : IntOps.IntOps) = struct
...
+    let ts = if get_interval_threshold_widening_constants () = "comparisons" then lower_or_upper else WideningThresholds.thresholds in
+    ResettableLazy.force ts
+
   let upper_threshold u max_ik =
-    let ts = if get_interval_threshold_widening_constants () = "comparisons" then WideningThresholds.upper_thresholds () else ResettableLazy.force widening_thresholds in
+    let ts = find_thresholds WideningThresholds.upper_thresholds in
     let u = Ints_t.to_bigint u in
     let max_ik' = Ints_t.to_bigint max_ik in
-    let t = List.find_opt (fun x -> Z.compare u x <= 0 && Z.compare x max_ik' <= 0) ts in
+    let t = WideningThresholds.Thresholds.find_first_opt (fun x -> Z.compare u x <= 0 && Z.compare x max_ik' <= 0) ts in
```
sim642, line 566: Actually this entire function is not monotonic as it's supposed to be, but just the `Z.compare u x <= 0` part is. The `Z.compare x max_ik' <= 0` could be done using a follow-up `Option.filter` or something.

The second check comes from #1476 and the point is that if the smallest upper threshold is bigger than the type, then don't pick it. So that's something which can be checked after finding the first upper threshold first.
Even in the list-based version, a separate check would've been better because if `List.find` already went into values bigger than the maximum, the predicate just returned `false` and it kept iterating the sorted list, although nothing smaller could've ever come again.

It might be good to have more unit tests like #1476 that could catch this problem, although I'm not sure how easy it is to construct such case.

PR #1659, opened 2025-01-19
Consider pointees separately for refinement

### LINKED ISSUE #1658: Refine pointer sets upon assertion about pointee
I just realized that we do not refine pointers when they are dereferenced in an assertion. In some cases, the assertion can only hold for some pointees, and in this case the pointer set should also be refined.

~~~C

int two = 2;
int three = 3;

int main() {
	int* ptr;
	int top;

	if(top) {
		ptr = &two;
	} else {
		ptr = &three;
	}

	if(*ptr == 2) {
                // This can only be true if ptr is &two
                // The assertion however does not succeed
		__goblint_check(ptr == &two);
	}

  	return 0;
}
~~~

michael-schwarz:
Closes #1658

### DIFF
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -117,7 +117,7 @@ struct
           match a with
           | Addr (base, _) as orig ->
             let (a:VD.t) = Address (AD.singleton (AD.Addr.of_var base)) in
-            if M.tracing then M.tracel "invo" "Consider case of lval %a = %a" d_lval lv VD.pretty a;
+            if M.tracing then M.tracel "inv" "Consider case of lval %a = %a" d_lval lv VD.pretty a;
             let st = set' lv a st in
             let old_val = get ~man st (AD.singleton orig) None in
             let old_val = VD.cast (Cilfacade.typeOfLval x) old_val in (* needed as the type of this pointer may be different *)
@@ -128,7 +128,7 @@ struct
             if is_some_bot v then
               D.join acc (try contra st with Analyses.Deadcode -> D.bot ())
             else (
-              if M.tracing then M.tracel "invo" "improve lval %a from %a to %a (c = %a, c' = %a)" d_lval x VD.pretty old_val VD.pretty v pretty c VD.pretty c';
+              if M.tracing then M.tracel "inv" "improve lval %a from %a to %a (c = %a, c' = %a)" d_lval x VD.pretty old_val VD.pretty v pretty c VD.pretty c';
               D.join acc (set' x v st)
             )
           | _ ->
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -85,6 +85,17 @@ struct
 
   let refine_lv man st c x c' pretty exp =
     let set' lval v st = set st (eval_lv ~man st lval) (Cilfacade.typeOfLval lval) ~lval_raw:lval v ~man in
+    let default () =
+      (* For accesses via pointers in complicated case, no refinement yet *)
+      let old_val = eval_rv_lval_refine ~man st exp x in
+      let old_val = map_oldval old_val (Cilfacade.typeOfLval x) in
+      let v = apply_invariant ~old_val ~new_val:c' in
+      if is_some_bot v then contra st
+      else (
+        if M.tracing then M.tracel "inv" "improve lval %a from %a to %a (c = %a, c' = %a)" d_lval x VD.pretty old_val VD.pretty v pretty c VD.pretty c';
+        set' x v st
+      )
+    in
     match x with
     | Var var, o when refine_entire_var ->
       (* For variables, this is done at to the level of entire variables to benefit e.g. from disjunctive struct domains *)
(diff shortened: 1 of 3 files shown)


src/analyses/baseInvariant.ml
```diff
@@ -100,6 +100,25 @@ struct
         if M.tracing then M.tracel "inv" "st from %a to %a" D.pretty st D.pretty r;
         r
       )
+    | Mem (Lval lv), NoOffset ->
+      let lvals = eval_lv ~man st x in
+      let res = AD.fold (fun a acc ->
```
michael-schwarz, line 105: Does something special need to be done for the unknown pointer?

src/analyses/baseInvariant.ml
```diff
@@ -100,6 +100,25 @@ struct
         if M.tracing then M.tracel "inv" "st from %a to %a" D.pretty st D.pretty r;
         r
       )
+    | Mem (Lval lv), NoOffset ->
```
sim642, line 103: Why is it necessary to limit this to `NoOffset`? Seems like it shouldn't really matter whether the pointer in in a struct field or something.

src/analyses/baseInvariant.ml
```diff
@@ -100,6 +100,25 @@ struct
...
+          if M.tracing then M.tracel "inv" "Consider case of lval %a = %a" d_lval lv VD.pretty (Address (AD.singleton a));
+          let st = set' lv (Address (AD.singleton a)) st in
+          let old_val = get ~man st (AD.singleton a) None in
+          let old_val = VD.cast (Cilfacade.typeOfLval x) old_val in (* needed as the type of this pointer may be different *)
+          (* this what I would originally have liked to do, but eval_rv_lval_refine uses queries and thus stale values *)
+          (* let old_val = eval_rv_lval_refine ~man st exp x in *)
+          let old_val = map_oldval old_val (Cilfacade.typeOfLval x) in
+          let v = apply_invariant ~old_val ~new_val:c' in
+          if is_some_bot v then 
+            D.join acc (try contra st with Analyses.Deadcode -> D.bot ())
```
sim642, line 115: This now swallows all deadcode, but this might not be entirely right. In particular, if the invariant doesn't hold for any pointees, this will join together a bunch of `bot`s, but not raise `Deadcode`, which I think it would've done before.
We might have to have some extra logic here that if all cases were dead, then `Deadcode` is also raised after the fold. I think `MCP` does something similar (but other way around, for any instead of all).

sim642:
Or perhaps not be so ambitious and handle nested dereferences... It's surprisingly tricky to get right even in very simple theoretical setting (the unassume paper appendix on pointers). We spent over an hour with Vesal trying to get it right in a semi-general case (where lvalues are sequences of derefs ending with a variable, so no arbitrary expressions in dereference, nor offsets). The formal definition is also very inefficient in subtle ways because the AST of such lvalues is in the wrong direction compared to the order of dereferencing pointers (which happens from inside out).
It's worth revisiting separately to not block this PR.

src/analyses/baseInvariant.ml
michael-schwarz: I fixed this and added a new regression test.

src/analyses/baseInvariant.ml
michael-schwarz: It is now handled explicitly.

src/analyses/baseInvariant.ml
michael-schwarz: I have generalized this now.

michael-schwarz:
I have now somewhat generalized the handling here. Recursive pointers etc are not considered yet, we can do this in a separate PR if we want to be ambitious.

src/analyses/baseInvariant.ml
```diff
@@ -100,17 +111,38 @@ struct
...
+            let old_val = map_oldval old_val (Cilfacade.typeOfLval x) in
+            let v = apply_invariant ~old_val ~new_val:c' in
+            if is_some_bot v then
+              D.join acc (try contra st with Analyses.Deadcode -> D.bot ())
+            else (
+              if M.tracing then M.tracel "inv" "improve lval %a from %a to %a (c = %a, c' = %a)" d_lval x VD.pretty old_val VD.pretty v pretty c VD.pretty c';
+              D.join acc (set' x v st)
+            )
+          | _ ->
+            default ()
```
sim642, line 135: This is a bit odd: for every non-suitable pointer in the points-to set, the `default` fallback on the entire set is performed again. So with multiple such pointers, `default` is recomputed in exactly the same way (it doesn't depend on anything here).
Also, if there's a non-suitable pointer and we already do the imprecise default thing, then it may be followed by any number of suitable pointers for which we do the precise thing, but join it into the imprecise one.

sim642 (approved):

michael-schwarz:
This leads to many new cases of `BothBranchesDead` as well as to one new unsoundness:

~~~C
struct aws_al {
  unsigned long current_size;
  unsigned long length;
};

struct aws_pq {
    int pred;
    struct aws_al container;
};

unsigned long nondet_size_t() { return __VERIFIER_nondet_ulong(); }

int main() {

  struct aws_pq queue = { 0, { 0, 0}};
  struct aws_al *const list = &queue.container;

  if (list->current_size == 0UL) {
    if (! (list->length == 0UL)) {

    }
  }
}
~~~

is the reduced example for `BothBranchesDead`. Here, it is reported over `! (list->length == 0UL)`.

PR #1663, opened 2025-01-24
Support for `pthread_once`

### LINKED ISSUE #1662: Support for `pthread_once` for race detection
Add along the lines of tracking `(A,S)` in the draft in the other repo: https://github.com/tum-cit-pl/digest-this-data-race.

This first class support means we will be able to remove https://github.com/goblint/analyzer/pull/547.

michael-schwarz:
See https://github.com/tum-cit-pl/digest-this-data-race.

TODO:

- [x] Remove workaround for #547 
- [x] Provide more tests

Closes #1662.

### DIFF
--- a/tests/regression/87-once/14-opaque.c
+++ b/tests/regression/87-once/14-opaque.c
@@ -1,4 +1,4 @@
-// PARAM: --set ana.activated[+] pthreadOnce --enable ana.int.interval
+// NOMAC PARAM: --set ana.activated[+] pthreadOnce --enable ana.int.interval
 #include <pthread.h>
 #include <stdio.h>
 // #include <goblint.h>


michael-schwarz:
The failing Ci is due to a force push.

lib/libc/stub/src/pthread.c
sim642, line 1: We had this stub to be sound out-of-the box.
Do we now need to have the new pthreadOnce analysis activated to be sound?

michael-schwarz: No, the code (potentially) doing the call in `src/framework/constraints.ml` is enabled unconditionally.

Also, the old stub is actually not sound: It seems to assume the function is guaranteed to be called every time `pthread_once` is called. This is wrong. 

~~~
// PARAM:
#include <goblint.h>
#include <pthread.h>

pthread_once_t once;
int g = 0;

void r() {
  g++;
}

int main() {
  int x;
  pthread_once(&once, r);
  // The value of g is now 1, since r() was called exactly once.
  __goblint_check(g == 1);
  pthread_once(&once, r);
  __goblint_check(g == 1);

  return 0;
}
~~~

```
[Error][Assert] Assertion "g == 1" will fail. Expected: SUCCESS -> failed (tests/regression/99-tutorials/05-prob.c:18:3-18:26)
````

I guess we just never noticed this problem, as we are typically modify globals inside of `pthread_once`, and Goblint tends to be flow-insensitive for them anyway.

Comment (sim642) [reply lib/libc/stub/src/pthread.c:1]: > No, the code (potentially) doing the call in `src/framework/constraints.ml` is enabled unconditionally.

So the pthreadOnce analysis is just for extra precision?


> I guess we just never noticed this problem, as we are typically modify globals inside of `pthread_once`, and Goblint tends to be flow-insensitive for them anyway.

We probably never considered using it in single-threaded mode. The same might even be true for some other pthread functions, but that's beside this PR.

File: src/domains/events.ml [hunk 1]
```diff
@@ -47,3 +51,5 @@ let pretty () = function
   | Assert exp -> dprintf "Assert %a" d_exp exp
   | Unassume {exp; tokens} -> dprintf "Unassume {exp=%a; tokens=%a}" d_exp exp (d_list ", " WideningToken.pretty) tokens
   | Longjmped {lval} -> dprintf "Longjmped {lval=%a}" (docOpt (CilType.Lval.pretty ())) lval
+  | EnterOnce {once_control; tf} -> dprintf "todo"
+  | LeaveOnce {once_control} -> dprintf "todo"
```
sim642: There should be proper strings here.

src/domains/events.ml
```diff
@@ -16,6 +16,8 @@ type t =
   | Assert of exp
   | Unassume of {exp: CilType.Exp.t; tokens: WideningToken.t list}
   | Longjmped of {lval: CilType.Lval.t option}
+  | EnterOnce of {once_control: CilType.Exp.t; tf:bool}
```
sim642: What does this `tf` mean?

src/analyses/pthreadOnce.ml
```diff
@@ -0,0 +1,109 @@
...
+    man.local
+
+  let startstate v = (Onces.empty (), Onces.empty ())
+
+  let possible_vinfos (a: Queries.ask) barrier =
+    Queries.AD.to_var_may (a.f (Queries.MayPointTo barrier))
+
+  let event man (e: Events.t) oman : D.t =
+    match e with
+    | Events.EnterOnce { once_control; tf } when tf ->
```
sim642:
```suggestion
    | Events.EnterOnce { once_control; tf = true } ->
```

src/analyses/pthreadOnce.ml
```diff
@@ -0,0 +1,109 @@
...
+  let name () = "pthreadOnce"
+  module D = Lattice.Prod (ActiveOnces) (SeenOnces)
+  include Analyses.ValueContexts(D)
+
+  (* transfer functions *)
+  let assign man (lval:lval) (rval:exp) : D.t =
+    man.local
+
+  let branch man (exp:exp) (tv:bool) : D.t =
+    man.local
```
sim642: Couldn't `IdentitySpec` simplify this?

src/framework/constraints.ml
```diff
@@ -258,6 +258,34 @@ struct
         let ad = man.ask (Queries.EvalFunvar e) in
         Queries.AD.to_var_may ad (* TODO: don't convert, handle UnknownPtr below *)
     in
+    let once once_control init_routine =
```
sim642: This begs my expected question: could this also be a `Spec` lifter? I haven't looked at this deeply enough to know whether this should be actually doable.

lib/libc/stub/src/pthread.c
michael-schwarz: > So the pthreadOnce analysis is just for extra precision?

Yes!

src/analyses/pthreadOnce.ml
michael-schwarz: Indeed!

src/framework/constraints.ml
michael-schwarz: I don't really see this leading to cleaner code here, we would need to introduce some sort of `CallProcedure p` event since we don't have access to `tf_proc` inside a lifter. However, I took another look at the code and managed to cut it down to about 10 additional lines.

src/framework/constraints.ml
```diff
@@ -266,15 +287,22 @@ struct
...
-                tf_normal_call man lv e fd args getl sidel demandl getg sideg
-              | exception Not_found ->
-                tf_special_call man lv f args)
-          end
+          let d =
+            let is_once = LibraryFunctions.find ~nowarn:true f in
+            match is_once.special args with
+            | Once { once_control; init_routine } -> once once_control init_routine
+            | _
+            | exception LibraryDsl.Expected _-> (* propagate weirdness inside *)
```
jerhard: Under which conditions is this `Expected` exception thrown? This may also be commented here, or for the function that is called.

tests/regression/87-once/02-normal.c
```diff
@@ -0,0 +1,34 @@
...
+  g = 10; //NORACE
+  pthread_mutex_unlock(&mutex1);
+  return NULL;
+}
+
+int main(void) {
+  pthread_t id;
+  pthread_create(&id, NULL, t_fun, NULL);
+
+  pthread_once(&once, fun);
```
jerhard, line 27: I am confused. Why don't the two accesses to `g` in lines `10` and `18` race? The access in `fun` is not protected by a mutex.

michael-schwarz: We discussed this offline: To reach line `18`, the call to `fun` must already have happened.

michael-schwarz:
@sim642 @jerhard: What is missing until you can approve this PR?

src/framework/constraints.ml
```diff
@@ -243,9 +243,37 @@ struct
...
+          tf_proc var edge prev_node None init_routine [] getl sidel demandl getg sideg d'
+        in
+        let later_call = S.event man (Events.EnterOnce { once_control;  ran = true }) man in
+        D.join (leave_once first_call) (leave_once later_call)
+      in
+      let is_once = LibraryFunctions.find ~nowarn:true f in
+      match is_once.special args with
+      | Once { once_control; init_routine } -> once once_control init_routine
+      | _
+      | exception LibraryDsl.Expected _ -> (* If there is a mismatch between our library function prototype and the call here, this is raised. However, we catch this here to still do `tf_special_call` where this case is handeled *)
```
sim642: If our library function specification is wrong, we should fix it. So I'm not sure why we should be handling this here now.
Previously the exception wasn't even exposed, so it wasn't treated specially anywhere else and that wasn't an issue.
Also, if `LibraryFunctions` raises this exception on `f` here and we delegate to the analyses with `S.special`, many of them will also just make the same library function lookup and get the same exception, which isn't caught and crashes Goblint.

The comment doesn't make it any more clear to me. It says

> However, we catch this here to still do `tf_special_call` where this case is handeled

but this whole thing is in `tf_special_call` to begin with.

michael-schwarz: The idea was to delegate the call to `S.special` to get the same behavior and warning as before. An  exception being raised in the analysis where the huge pattern match over all possible library functions is, is arguably more useful to a developer then raising it here, leading them on a wild goose chase about `pthread_once` that has nothing to do with the problem.

michael-schwarz: But I don't have strong opinions, I would also be fine with raising on the top level.

sim642: I think I would then prefer not exposing the exception at all. We can just have a comment the code here to note that such exceptions raised from here are probably unrelated to `pthread_once`.

michael-schwarz:
Not sure what's up with the CI, seems unrelated.

tests/regression/87-once/01-sanity.c
```diff
@@ -0,0 +1,27 @@
...
+
+int main(void) {
+  pthread_t id;
+
+  pthread_once(&once, fun);
+  pthread_once(&once, fun);
+
+  __goblint_check(g < 2);
+
+  if(g = 1) {
```
sim642:
```suggestion
  if(g == 1) {
```

tests/regression/87-once/01-sanity.c
```diff
@@ -0,0 +1,27 @@
...
+}
+
+
+int main(void) {
+  pthread_t id;
+
+  pthread_once(&once, fun);
+  pthread_once(&once, fun);
+
+  __goblint_check(g < 2);
```
sim642, line 21: There should be `#include <goblint.h>` here and perhaps in some others.

tests/regression/87-once/03-unknown.c
```diff
@@ -0,0 +1,44 @@
+// PARAM: --set ana.activated[+] pthreadOnce
+#include <pthread.h>
+#include <stdio.h>
+
+int g;
+pthread_once_t once1 = PTHREAD_ONCE_INIT;
+pthread_once_t once2; // PTHREAD_ONCE_INIT is `0`
```
sim642: What is the relevance of `PTHREAD_ONCE_INIT` being 0? Is this an extra assumption we're now making?

tests/regression/87-once/04-thread.c
```diff
@@ -0,0 +1,58 @@
...
+
+void nesting() {
+  h = 5; //NORACE
+}
+
+void fun() {
+  // Even though this is only called inside the once, the accesses in the new thread and the accesses here can happen in parallel
+  // Checks that active is not passed to the created thread, but seen is passed
+  pthread_t tid = pthread_create(&tid, NULL, t_other, NULL);
+  g = 42; //RACE
```
sim642: If there's an actual race possible, not just our imprecision, then this should be `RACE!` here and on the other side.

tests/regression/87-once/04-thread.c
```diff
@@ -0,0 +1,58 @@
...
+pthread_once_t once = PTHREAD_ONCE_INIT;
+pthread_once_t i_once = PTHREAD_ONCE_INIT;
+pthread_mutex_t mut;
+
+void *t_other(void* arg) {
+  g = 17; //RACE
+
+  pthread_mutex_lock(&mut);
+  i = 7; //NORACE
+  pthread_mutex_unlock(&mut);
```
sim642, line 21: What is the mutex for? Against self-races?
If `fun` is run once, then `t_other` should be unique and not self-race.

Or is it that `fun` being run once from different threads causes two different unique thread IDs which would then race here? That seems like imprecision on our part that's worth a TODO.

tests/regression/87-once/05-unknown-tid.c
```diff
@@ -0,0 +1,28 @@
...
+void* thread(void* arg) {
+  pthread_once(&once, fun);
+  return NULL;
+}
+
+int main(void) {
+  pthread_t id;
+
+  for(int i=0; i < 100; i++) {
+    // Will receive unknown TID
```
sim642: Really unknown TID, not just non-unique?

tests/regression/87-once/10-pointer-once.c
```diff
@@ -0,0 +1,33 @@
...
+  return NULL;
+}
+
+int main(void) {
+  pthread_t id;
+  int top;
+
+  pthread_create(&id, NULL, thread, NULL);
+  if(top) { optr = &once1; }
+  pthread_once(optr, init0);
```
sim642, line 30: Is it intentional that if `!top`, then `optr == NULL` from global initialization? Or should it be `&once` then?

tests/regression/87-once/10-pointer-once.c
```diff
@@ -0,0 +1,33 @@
...
+int g;
+void init0();
+
+pthread_once_t* optr;
+pthread_once_t once = PTHREAD_ONCE_INIT;
+pthread_once_t once1 = PTHREAD_ONCE_INIT;
+pthread_mutex_t mutex1 = PTHREAD_MUTEX_INITIALIZER;
+
+void init0() {
+  g++; //RACE
```
sim642: Is this also `RACE!` then?

tests/regression/87-once/11-combination.c
```diff
@@ -0,0 +1,45 @@
...
+}
+
+
+void* thread(void* arg) {
+  pthread_once(&once, init0);
+
+  pthread_mutex_lock(&mtx);
+  if(g == NULL) {
+    // Will not happen!
+    // Also showing that g is not NULL here requires "Extended Support for pthread once" as explained in appendix C.
```
sim642, line 21: Why not `__goblint_check(0); // NOWARN (unreachable)` then? Or is this extended support not implemented?

src/analyses/pthreadOnce.ml
```diff
@@ -0,0 +1,83 @@
...
+    | _ -> man.local
+
+  let access man _ = man.local
+
+  module A =
+  struct
+    include D
+    let name () = "onces"
+    let may_race (a1, s1) (a2, s2) =
+      (Onces.is_empty (Onces.inter a1 (Onces.union a2 s2))) && (Onces.is_empty (Onces.inter a2 (Onces.union a1 s1)))
```
sim642: Can't one use `disjoint` instead of `is_empty` on `inter` here?

tests/regression/87-once/03-unknown.c
michael-schwarz: PTHREAD_ONCE_INIT corresponds to a struct with all values zero-initialized, so coincides with the default value for global variables and can be omitted.
It technically is an assumption, but we (and everyone else) makes the same assumption about `PTHREAD_MUTEX_INITIALIZER` as well, so I don't think it's worth documenting.

tests/regression/87-once/04-thread.c
michael-schwarz: It is indeed imprecision and I added a comment to this effect.

tests/regression/87-once/11-combination.c
michael-schwarz: The extended support is not implemented, and this branch simulates what we would additionally find after implementing it.

src/analyses/pthreadOnce.ml
michael-schwarz: It boils down to three disjointness checks.

tests/regression/87-once/03-unknown.c
sim642: Why does it matter for us at all? The new analyses never look at the contents, they only care about the address of these things.

I now looked up that on Linux it seems to be `int`, which might be a separate concern: base, apron, etc could track its value, which they probably shouldn't. We should probably make sure that they don't by adding this type to some ignorable things like mutexes and thread IDs. Although for the base analysis value domain this could also be annoying, because we have `Mutex` for the opaque thing, but introducing new constructors for each opaque thing is probably very inconvenient. It should be fine to abuse it for `pthread_once_t` to get the main functionality merged here, but perhaps `ValueDomain.Mutex` should be generalized to `ValueDomain.Opaque` or something less specific then.

src/analyses/pthreadOnce.ml
sim642: Nice! Also avoiding the unions is good for not allocating unnecessary garbage.

tests/regression/87-once/03-unknown.c
michael-schwarz: > Why does it matter for us at all?

It matters in that the program has defined behavior, and that's the only thing that matters.

I'm not sure if we need some opaque thing here, it seems to work without it.

src/analyses/pthreadOnce.ml
```diff
@@ -0,0 +1,85 @@
...
+  include Analyses.IdentitySpec
+
+  let name () = "pthreadOnce"
+  module D = Lattice.Prod (ActiveOnces) (SeenOnces)
+  include Analyses.ValueContexts(D)
+
+  let startstate v = (Onces.empty (), Onces.empty ())
+
+  let possible_vinfos (a: Queries.ask) barrier =
+    Queries.AD.to_var_may (a.f (Queries.MayPointTo barrier))
```
sim642: Here's a thought I just had: what if the `pthread_once_t` is dynamically allocated and thus potentially non-unique?
It seems that then different `pthread_once` calls might accidentally use different concrete instances which abstractly appear the same, but that wouldn't rule out anything w.r.t. scheduling.
So should here be some `IsMultiple` filtering as well?

The man page says

> The behavior of `pthread_once()` is undefined if `once_control` has automatic storage duration or is not initialized by `PTHREAD_ONCE_INIT`.

I think local variables have automatic storage duration, but not `malloc`-ed things, right?
I'm also not sure if manually initializing the dynamically allocated memory (or using `calloc` because it's 0 by default) is considered fine by this.
But either way, given that the flag is just `int`, it seems that it could just work perfectly fine when dynamically allocated and correctly initialized.

tests/regression/87-once/03-unknown.c
sim642: I mean, we probably will make unsound claims about `pthread_once(&once, fun); assert(once == 0);`. Currently base analysis would prove this, but I'm doubtful that this actually holds.

src/analyses/pthreadOnce.ml
michael-schwarz: Good catch, we're unsound here. In fact, something you remarked for the barriers hurts us here:

~~~C
// PARAM: --set ana.activated[+] pthreadOnce --enable ana.int.interval
#include <pthread.h>
#include <stdio.h>
#include <goblint.h>
#include <stdlib.h>

struct {
  pthread_once_t once1;
  pthread_once_t once2;
} onces = {PTHREAD_ONCE_INIT, PTHREAD_ONCE_INIT};

int g;
pthread_mutex_t mtx;

void increment() {
  g++;
}

int main(void) {

  pthread_once(&onces.once1, increment);
  pthread_once(&onces.once2, increment);

  __goblint_check(g == 2); //TODO
  printf("g = %d\n", g);

  return 0;
}
~~~

Here, we claim `g==2` will fail, as we do not consider the two onces separately.

michael-schwarz: I have now added code to take care of both cases. Interestingly, as we track must information here, two `mvals` that alias each other (through strange offset things) are no problematic here. We will consider them separately and only lose some precision.

tests/regression/87-once/03-unknown.c
michael-schwarz: I have now fixed this. I haven't renamed `ValueDomain.Mutex` though. We already use it for condition variables anyway, and I guess `pthread_once` is just as mutex-flavored as condition variables.

sim642 (approved):

michael-schwarz:
Not considering `Mutex` to be bot seems to also have fixed two instances of both branches dead on SV-COMP.

PR #1667, opened 2025-01-30
Disable race analyses for other ConcurrencySafety properties

karoliineh:
* Make autotuner disable always-enabled `race`, `region`, and `symb_locks` analyses for other ConcurrencySafety properties, in addition to its previous behavior of disabling all thread analyses for single-threaded programs
* Rename autotune option `singleThreaded` → `reduceAnalyses`
* Refactor logging info about enabling analyses into `enableAnalyses` function
* Use `List.mem` when looking for one specification in the list of specifications instead of iterating over the whole list


src/witness/svcompSpec.ml
```diff
@@ -97,3 +97,15 @@ let to_string spec =
...
+  match spec1, spec2 with
+  | UnreachCall f1, UnreachCall f2 -> String.equal f1 f2
+  | NoDataRace, NoDataRace
+  | NoOverflow, NoOverflow
+  | Termination, Termination
+  | ValidFree, ValidFree
+  | ValidDeref, ValidDeref
+  | ValidMemtrack, ValidMemtrack
+  | ValidMemcleanup, ValidMemcleanup -> true
+  | _, _ -> false
```
sim642, line 111: This should be derived (and thus called `equal`).

src/autoTune.ml
```diff
@@ -218,17 +218,21 @@ let enableAnalyses anas =
...
     match desc.special args with
     | LibraryDesc.ThreadCreate _ -> true
     | _ -> LibraryDesc.Accesses.find_kind desc.accs Spawn args <> []
   in
   let hasThreadCreate = hasFunction isThreadCreate in
-  if not @@ hasThreadCreate then (
-    Logs.info "no thread creation -> disabling thread analyses \"%s\"" (String.concat ", " notNeccessaryThreadAnalyses);
-    disableAnalyses notNeccessaryThreadAnalyses;
-  )
+  let hasDataRaceSpec = List.exists (SvcompSpec.equals SvcompSpec.NoDataRace) (Svcomp.Specification.of_option ()) in
```
sim642, line 229: This could be `List.mem` (`BatList.mem_cmp` to be more precise), but that needs `compare` instead of `equal` (could also be derived).

conf/svcomp26.json
sim642, line 1: There's no need to create this configuration yet, just changing `svcomp` conf is enough.

michael-schwarz (approved):

src/config/options.schema.json
```diff
@@ -534,7 +534,7 @@
                 "type": "string",
                 "enum": [
                   "congruence",
-                  "singleThreaded",
+                  "reduceAnalyses",
```
sim642, line 537: After some back-and-forth, this still just does the single threaded thing and the race analyses are instead part of `concurrencySafetySpecification`, right?

PR #1668, opened 2025-01-31
Add abstract interface to `Invariant`

sim642:
This adds the intended abstract interface to clarify that "Invariants for witnesses" are not just CIL expressions: https://github.com/goblint/analyzer/pull/1665#issuecomment-2618905460.

### TODO
- [ ] What to do about precondition loop invariants?

### DIFF
--- a/src/witness/yamlWitness.ml
+++ b/src/witness/yamlWitness.ml
@@ -293,8 +293,8 @@ struct
                 Invariant.(acc || R.ask_local_node n ~local (Invariant {Invariant.default_context with lvals})) [@coverage off] (* bisect_ppx cannot handle redefined (||) *)
               ) (Invariant.bot ()) ns
             in
-            match inv with
-            | `Lifted inv ->
+            match Invariant.to_exp inv with
+            | Some inv ->
               let fundec = Node.find_fundec (List.hd ns) in (* TODO: fix location hack *)
               let location_function = fundec.svar.vname in
               let location = Entry.location ~location:loc ~location_function in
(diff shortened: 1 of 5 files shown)


src/witness/yamlWitness.ml
```diff
@@ -450,11 +450,11 @@ struct
...
         FMap.iter (fun f con_invs ->
             List.iter (fun current_c ->
-                begin match current_c.invariant with
-                  | `Lifted c_inv ->
+                begin match Invariant.to_exp current_c.invariant with
+                  | Some c_inv ->
                     (* Collect all start states that may satisfy the invariant of current_c *)
                     List.iter (fun c ->
-                        let x = R.ask_local (c.node, c.context) ~local:c.state (Queries.EvalInt c_inv) in
+                        let x = R.ask_local (c.node, c.context) ~local:c.state (Queries.EvalInt c_inv) in (* TODO: illegal query *)
```
sim642, line 457: The generation of our proposed precondition loop invariants performs `EvalInt` queries with witness invariants, assuming they have implicit casts, etc like expressions from the program.
This would no longer be the case with #1665.

So we have to decide what to do here:
1. If we don't want to pursue precondition loop invariants and keep them, they could be removed, avoiding the issue.
2. Otherwise, `Invariant.t` might need to be turned into a pair of CIL expressions (with and without implicit casts) behind the abstraction. This might be desirable in the long run for #1635, although that uses separate expression construction from `invariant` anyway.

michael-schwarz: I think we should keep them, especially if we intend to work with LMU to extend the witness format.

sim642: This probably needs some separate discussion if this is still the kind of entry/invariant we want to propose. LMU+Freiburg have recently proposed function contracts as well: https://arxiv.org/abs/2501.12313.

michael-schwarz: Yeah! At first glance, I think we can basically encode these in the invariants as a conjunction of implications of the form `pre_i \implies post_i` where `pre_i` uses these `\at(_, Pre)` predicates (which would mean we need an ACSL parser, or some hacks such as functions `at` and a constant `Pre`).

PR #1669, opened 2025-01-31
Ability to enable/disable `may_race` checks per race diges

### LINKED ISSUE #1664: Ability to enable/disable `may_race` checks per race digest
Such options are needed to make experiments to see how relevant what digest is for our new project.

michael-schwarz:
Closes #1664.

### DIFF
--- a/src/analyses/mallocFresh.ml
+++ b/src/analyses/mallocFresh.ml
@@ -64,7 +64,9 @@ struct
   struct
     include BoolDomain.Bool
     let name () = "fresh"
-    let may_race f1 f2 = not (f1 || f2)
+    let may_race f1 f2 =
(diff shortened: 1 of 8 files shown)


sim642:
This was useful for experiments, but those are now done I guess.

I'm wondering how much sense it makes to keep this extra configurability for the future. What this allows is having some analysis active but not using the information it computes to exclude races — a feature that no user would ever want.
It also creates a huge opportunity for misunderstanding and misconfiguration: the options are enabled by default, which makes it seem like all of the features are being used when in fact they are not. Only some of us would know that they do nothing unless some corresponding analysis is also activated. Having the defaults flipped wouldn't be any better: one might then try to enable them without effect or try to activate some analysis without effect.

The race digest experiments are tagged anyway and these changes could easily be cherry-picked for future experiments if really necessary.

Issue #1671, opened 2025-02-06
Int domain refinement causes fixpoint error


AdrianKrauss:
A second program, which we believe has the same problem but terminates with a fixpoint error (which maybe makes the debugging easier), is the following:

```c
// PARAM:  --conf ./svcomp_bitfield.json --set ana.specification ./unreach-call.prp

reach_error() {}
a(b) {
  if (!b) {
    reach_error();
    abort();
  }
}

main() {
  int c = 0;
  unsigned d;
  unsigned e = 0;
  while (5) {
    c = c + 1;
    d = d + 2;
    e = e + 3;
    a(e == c + d);
  }
}
```

However, the fixpoint error of the second program only occured with activated refinements and the activated bitfield domain of our PR (see #1623 ). Without our bitfield domain but still with activated refinements, the second program terminates normally without any problems. Therefore, we need a slightly different config file to produce this error:
[svcomp_bitfield.json](https://github.com/user-attachments/files/18686491/svcomp_bitfield.json)

michael-schwarz:
Have you tried disabling witnesses? They often cause issues.

sim642:
> Have you tried disabling witnesses? They often cause issues.

If so, then it might be another example of #1322. I suspect the `join old new` has something to do with it by always joining in old path dependencies. PR #1484 has a similar case where `widen old new` would be really beneficial (and where the transition to https://link.springer.com/chapter/10.1007/978-981-19-9601-6_4 would be a good idea).

michael-schwarz:
The problem seems to be unrelated to witnesses.

~~~C
// PARAM: --set ana.int.refinement fixpoint --enable ana.int.enums  --enable ana.int.interval --set sem.int.signed_overflow assume_none
int main() {
   int count = 0;
   while (1) {
     count++;
     count++;
   }
 }
~~~

However, it appears to be connected to `assume_none` for `sem.int.signed_overflow`...

michael-schwarz:
And huge props to @AdrianKrauss and the rest of you, this minimized example has been essential in getting to the root cause!

michael-schwarz:
Minimal example where it misbehaves with not finding the fixpoint (decoupled from SV-COMP). Interestingly #1674 does not fix the issue, whereas #1675 does.

~~~C
[...]

PR #1672, opened 2025-02-11
Fix `fixpoint` option `ana.int.refinement` - loop was never executed

michael-schwarz:
While investigating #1671, @jerhard and i discovered that the option to refine int domains until a fixpoint is reached has been broken since the very beginning.

In the PR #260 adding refinement, commit https://github.com/goblint/analyzer/commit/1aee2eb26d8276cb41a920736ba086e3757e3dca added a special detection for bottom to ensure bottom values are normalized across all domains.

~~~Ocaml
if is_bot !dt then dt := bot_of ik; quit_loop := true;
~~~

There, parentheses are missing. Thus, the loop is always left after one iteration.

### DIFF
--- a/src/cdomain/value/cdomains/int/intDomTuple.ml
+++ b/src/cdomain/value/cdomains/int/intDomTuple.ml
@@ -265,7 +265,7 @@ module IntDomTupleImpl = struct
          let old_dt = !dt in
          List.iter (fun f -> dt := f !dt) (refine_functions ik);
          quit_loop := equal old_dt !dt;
-         if is_bot !dt then dt := bot_of ik; quit_loop := true;
+         if is_bot !dt then (dt := bot_of ik; quit_loop := true);
          if M.tracing then M.trace "cong-refine-loop" "old: %a, new: %a" pretty old_dt pretty !dt;
        done;
      | _ -> ()


sim642:
Well, this is a surprising find!

Could you add some 38=int-refinements test that actually requires more than one iteration?

sim642:
For reference Miné's tutorial (Section 6.2.4, 2017) says:

> In the reduction between intervals and congruences, only one application of each reduction is necessary, provided we first refine intervals from congruences, and then congruences from intervals.

If I remember correctly, our int domain refinement does all directions in a single pass rather than in some arbitrary order, no? In that case such example should exist (perhaps one in the tutorial?) where two iterations are needed.

michael-schwarz:
> In that case such example should exist (perhaps one in the tutorial?) where two iterations are needed.

Yes, the point was not about needing two rounds (there are such examples), but about there being an observable difference in the end result between `once` and `fixpoint`. 
For  `once`, the outer fixpoint will see values have changed after each step refinement, and cause a re-evaluation.
So I think the end result with both strategies will be the same, only that the iteration happens outside instead of locally.
The only way I could see there being a difference would be if more of these outer evaluations have an effect on whether widening is applied.

sim642:
Such outer fixpoint iteration only exists when such refinement is within a loop though. If it's outside a loop (just a chain of `if`s or something), then solver's fixpoint iteration doesn't come back to do the following iterations for the int refinement.

michael-schwarz:
We actually refine one after the other, and in the direction where iteration may be necessary

~~~Ocaml

sim642 (approved):

PR #1676, opened 2025-02-12
Recognize pointers wrapped in `TNamed` in `memOutOfBounds` and fix bit-to-byte comparison bug

karoliineh:
Looking through Goblint logs for the SV-COMP MemSafety tasks revealed several issues that are fixed in this PR:
1. Pointers wrapped in `TNamed` were not recognized and considered as pointers
2. Bytes were compared to bits
3. In some edge cases, equality between pointers wrapped in `TNamed` was not working

Additional fixes:
* The comparison `ID.lt casted_es casted_offs` requires subtracting 1 from `casted_es`, but the corresponding log message about the size of `casted_es` was misleading because of the subtraction. The log message now displays the actual size before the subtraction to prevent confusion.

### DIFF
--- a/src/analyses/memOutOfBounds.ml
+++ b/src/analyses/memOutOfBounds.ml
@@ -127,7 +127,7 @@ struct
          `Top)
 
   let get_ptr_deref_type ptr_typ =
-    match ptr_typ with
+    match Cil.unrollType ptr_typ with
     | TPtr (t, _) -> Some t
     | _ -> None
 
--- a/src/analyses/memOutOfBounds.ml
+++ b/src/analyses/memOutOfBounds.ml
@@ -279,10 +279,10 @@ struct
              M.warn "Size of lval dereference expression %a is bot. Out-of-bounds memory access may occur" d_exp e)
           | `Lifted es ->
             let casted_es = ID.cast_to (Cilfacade.ptrdiff_ikind ()) es in
-            let one = intdom_of_int 1 in
-            let casted_es = ID.sub casted_es one in
             let casted_offs = ID.cast_to (Cilfacade.ptrdiff_ikind ()) offs_intdom in
             let ptr_size_lt_offs =
+              let one = intdom_of_int 1 in
+              let casted_es = ID.sub casted_es one in
               begin try ID.lt casted_es casted_offs
                 with IntDomain.ArithmeticOnIntegerBot _ -> ID.bot_of @@ Cilfacade.ptrdiff_ikind ()
               end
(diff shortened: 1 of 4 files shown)


tests/regression/01-cpa/76-pointer-typedef.c
```diff
@@ -0,0 +1,31 @@
+// PARAM: --set ana.malloc.unique_address_count 2
+// Extracted (using creduce) from SV-COMP task list-simple/dll2c_remove_all.i
+#include <stdlib.h>
```
sim642, line 3:
```suggestion
#include <stdlib.h>
#include <goblint.h>
```

src/analyses/memOutOfBounds.ml
```diff
@@ -279,10 +279,10 @@ struct
              M.warn "Size of lval dereference expression %a is bot. Out-of-bounds memory access may occur" d_exp e)
           | `Lifted es ->
             let casted_es = ID.cast_to (Cilfacade.ptrdiff_ikind ()) es in
-            let one = intdom_of_int 1 in
-            let casted_es = ID.sub casted_es one in
-            let casted_offs = ID.cast_to (Cilfacade.ptrdiff_ikind ()) offs_intdom in
+            let casted_offs = ID.div (ID.cast_to (Cilfacade.ptrdiff_ikind ()) offs_intdom) (ID.of_int (Cilfacade.ptrdiff_ikind ()) (Z.of_int 8)) in
```
michael-schwarz, line 282: Nice catch, I guess this tended to kill us almost all of the time.

michael-schwarz (approved):

src/analyses/memOutOfBounds.ml
sim642: There's also a documentation problem (by me) that might've contributed to this:
https://github.com/goblint/analyzer/blob/153ce288ada8ecf6d8276cae512e5ad58ef65d88/src/cdomain/value/cdomains/offset_intf.ml#L92-L96
Contrary to the description, it returns the offset in bits (matching what CIL has). It was fine for `semantic_equal` because both would be in bits, but not for other uses (this is the only other one I think).

I'll have to open a follow-up PR to fix that. Changing the documentation would be easy, but it's not entirely the right thing: `ptrdiff_t` is large enough to fit offsets in bytes, but not necessarily in bits (in practice the difference might be irrelevant unless a program actually allocates more than 8th of the address space). Having `to_index` return bytes would be consistent with `ptrdiff_ikind`.

sim642 (approved):

PR #1677, opened 2025-02-12
Add missing type unrollings

### LINKED ISSUE #1600: Audit for missing type unrollings
https://github.com/goblint/analyzer/pull/1519#issuecomment-2417032186 highlights an issue in our codebase: we do a lot of pattern matching on `Cil.typ` to handle particular cases, but the type isn't unrolled beforehand with `Cil.unrollType`. This can cause some logic to fall back to imprecise behavior due to `TNamed` instead of `TInt`/etc, although the `TNamed` (or a nesting of them!) unrolls to one of the desired types.

I'm not sure how to best find all such places in the codebase, but pattern matching, etc should in most cases probably happen on unrolled types. That is, our analysis should be invariant under extracting/inlining `typedef`s in the program.

sim642:
Closes #1600. I once started working on it, but never completed it.

PR #1676 independently found two places with missing type unrolling (which are also among the ones found here) that have clear impact on precision. Thus, it would be a good idea to more systematically fix the issue as a preventative measure.

Currently this PR is mostly a scattering of TODOs that still need to be looked at.


sim642:
> Are there any circumstances under which unrolling might be the wrong thing to do?

Yes, some places actually want to match `TNamed` and use the `typedef` name. For example, mutex types are recognized like this (instead of looking at the internal `struct` that they happen to be).
So this cannot be done blindly.

src/analyses/base.ml
```diff
@@ -2030,10 +2031,11 @@ struct
...
       | Some exp ->
-        let t_override = match Cilfacade.fundec_return_type fundec with
-          | TVoid _ -> M.warn ~category:M.Category.Program "Returning a value from a void function"; assert false
-          | ret -> ret
-        in
+        let t_override = Cilfacade.fundec_return_type fundec in
+        if Cil.isVoidType t_override then (
+          M.warn ~category:M.Category.Program "Returning a value from a void function";
+          assert false
+        );
```
michael-schwarz, line 2038: This is unrelated to the main topic of this PR, but doesn't the `assert false` cause the program to crash, and the warning before to never be printed? Does not seem to make much sense.

src/analyses/baseInvariant.ml
```diff
@@ -581,8 +581,8 @@ struct
       else
         match exp, c_typed with
         | UnOp (LNot, e, _), Int c ->
-          (match Cilfacade.typeOf e with
-           | TInt  _ | TPtr _ ->
+          (match Cil.unrollType (Cilfacade.typeOf e) with
+           | TInt  _ | TPtr _ -> (* TODO: TEnum as well? *)
```
michael-schwarz, line 585: Yeah, `TEnum` is `TInt` for all intents and purposes, so we should add it here.

src/autoTune.ml
```diff
@@ -311,6 +311,7 @@ class addTypeAttributeVisitor = object
 
   (*Set arrays with important types for thread analysis to unroll*)
   method! vtype typ =
+    (* TODO: reuse predicates in Access module (also handles TNamed correctly) *)
```
michael-schwarz, line 314: Should we not do this rather than leave new TODOs in the code base?

src/autoTune0.ml
```diff
@@ -91,6 +91,7 @@ let collectFactors visitAction visitedObject =
   ignore (visitAction visitor visitedObject);
   factors
 
-let is_large_array = function
-  | TArray (_,Some (Const (CInt (i,_,_))),_) -> i > Z.of_int @@ 10 * get_int "ana.base.arrays.unrolling-factor"
+let is_large_array t =
+  match Cil.unrollType t with
+  | TArray (_,Some (Const (CInt (i,_,_))),_) -> i > Z.of_int @@ 10 * get_int "ana.base.arrays.unrolling-factor" (* TODO: Cil.lenOfArray but with Z.t? *)
```
michael-schwarz, line 96: Or use `Cil.lenofArray` and do it on `int`? Arrays of length greater than $2^63$ are not common anyway.

src/cdomain/value/cdomains/int/congruenceDomain.ml
```diff
@@ -139,21 +139,21 @@ struct
 
   let of_congruence ik (c,m) = normalize ik @@ Some(c,m)
 
-  let of_bitfield ik (z,o) = 
+  let of_bitfield ik (z,o) =
     match BitfieldDomain.Bitfield.to_int (z,o) with
     | Some x -> normalize ik (Some (x, Z.zero))
     | _ ->
       (* get posiiton of first top bit *)
```
michael-schwarz, line 146:
```suggestion
      (* get position of first top bit *)
```

src/cdomain/value/cdomains/int/congruenceDomain.ml
```diff
@@ -182,7 +182,7 @@ struct
         let (min_ikorg, max_ikorg) = range ikorg in
         ikorg = t || (max_t >=: max_ikorg && min_t <=: min_ikorg)
       in
-      match torg with
+      match Option.map Cil.unrollType torg with
       | Some (Cil.TInt (ikorg, _)) when p ikorg ->
```
michael-schwarz, line 186: Why not `TEnum` here?

michael-schwarz (changes requested):
Thanks for putting all this effort into this! I left some comments here and there.

src/analyses/baseInvariant.ml
sim642: I might also look through other uses of `TInt` to see if `TEnum` is handled consistently as well, if we're already in the process of doing these things.

src/analyses/base.ml
sim642: That might be true indeed if we aren't doing early warnings or such. I'll think about it and perhaps turn it into logging/`failwith`.

src/cdomain/value/cdomains/valueDomain.ml
```diff
@@ -507,8 +513,8 @@ struct
           Address (match v with
               | Int x when ID.to_int x = Some Z.zero -> AD.null_ptr
               | Int x -> AD.top_ptr
-              (* we ignore casts to void*! TODO report UB! *)
-              | Address x -> (match t with TVoid _ -> x | _ -> cast_addr t x)
+              (* we ignore casts to void* (above)! TODO report UB! *)
+              | Address x -> cast_addr t x
```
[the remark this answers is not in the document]
sim642: This case was actually unreachable, there's a separate match case above specially for `void*` things.

src/util/loopUnrolling.ml
```diff
@@ -261,7 +261,7 @@ let findLoopVarAndGoal loopStatement func (op, exp1, exp2) =
   | _ -> None
 
 let getLoopVar loopStatement func = function
-  | BinOp (op, exp1, exp2, TInt _) when isCompare op -> findLoopVarAndGoal loopStatement func (op, exp1, exp2)
+  | BinOp (op, exp1, exp2, _) when isCompare op -> findLoopVarAndGoal loopStatement func (op, exp1, exp2)
```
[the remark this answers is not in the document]
sim642: Not really, all _comparison_ operators (filtered by `isCompare`) always have `int` type by construction anyway. So `TInt` always matched.
Even if it's always the case, I think it might be less confusing to not match it at all because it appears like something special is being achieved when it's not.

src/common/util/cilfacade.ml
```diff
@@ -27,18 +27,28 @@ let isCharType t =
   | TInt ((IChar | ISChar | IUChar), _) -> true
   | _ -> false
 
+let isBoolType t =
+  match Cil.unrollType t with
+  | TInt (IBool, _) -> true
```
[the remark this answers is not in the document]
sim642: I'm not sure it's possible to have that at all. N1570 6.7.2.2.4 says:

> Each enumerated type shall be compatible with `char`, a signed integer type, or an unsigned integer type.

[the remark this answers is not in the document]
michael-schwarz: Is `IBool` not an `unsigned integer type`? (Not sure how central that question is though).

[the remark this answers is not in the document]
sim642: Actually it might be, but I'm not entirely sure if this function should consider it as such. For example, `isCharType` immediately above and others defined in CIL itself don't seem to.
They might be intended to exactly match the standard, where integer types and enumeration types aren't exactly the same thing. They're just convertible in various ways I think.

src/analyses/baseInvariant.ml
sim642: I've now done that in most places.

src/autoTune0.ml
sim642: I think I considered but it didn't seem much nicer. I changed it nevertheless because `Cil.lenOfArray` does constant folding which could be a good idea (for array lengths written as constant expressions via `#define`-s).
I'll leave the TODO though because it's silly on the CIL side: the computation is done with `Z.t` and downcast...

src/autoTune.ml
sim642: Right now the ones in `Access` aren't directly suitable for reuse here: what `Access` considers ignorable isn't exactly what unrolling considers important (e.g. `FILE`, `jmp_buf`, etc).

I think this needs a bit bigger refactoring. In particular, what I'd like to have is something like `LibraryTypes` (but initially at least less DSLy than `LibraryFunctions`). These are currently unconditional everywhere, but they should also be controlled by `lib.activated`.
I think recently @karoliineh found something in sv-benchmarks where our Linux kernel-specific overrides trigger on LDV tasks because they use the same names, but actually have stub implementations different from the kernel's. In the specific case we were perhaps imprecise because of it (?) but this is actually accidental task fingerprinting.

src/analyses/base.ml
sim642: I just tried triggering this case and couldn't even do it. Apparently CIL already covers it here: https://github.com/goblint/cil/blob/f5ee39bd344dc74e2a10e407d877e0ddf73c9c6f/src/frontc/cabs2cil.ml#L6840-L6843. I think we don't show the CIL warning but CIL removes the returned expression anyway, so Goblint never sees it.

I'll change this to a plain `assert` without a more specific message because this shouldn't even be triggerable by the user.

michael-schwarz (approved):

PR #1682, opened 2025-02-14
Ensure that `__goblint_check` does not affect memLeak analysis

### LINKED ISSUE #1678: memLeak analysis is affected by `__goblint_check`
@karoliineh found that our SV-COMP valid-memsafety verdict is impacted by inserting a `__goblint_check` into the program.
That is supposed to not impact the analysis at all (only output), so this behavior is wrong.

The offending code is here:
https://github.com/goblint/analyzer/blob/153ce288ada8ecf6d8276cae512e5ad58ef65d88/src/analyses/memLeak.ml#L230-L245
In particular, it should probably only be handled when `Assert` has `refine = true` or something.

karoliineh:
Fixes #1678

### DIFF
--- a/tests/regression/76-memleak/32-no-mem-leak-goblint-check.c
+++ b/tests/regression/76-memleak/32-no-mem-leak-goblint-check.c
@@ -1,8 +1,9 @@
 // PARAM: --set ana.activated[+] memLeak --set ana.malloc.unique_address_count 1
+#include <stdlib.h>
 #include <goblint.h>
 
 int main() {
     int *ptr = malloc(sizeof(int));
-    __goblint_check(ptr); // UNKNOWN
+    __goblint_check(ptr == 0); // FAIL
     free(ptr);
 }
--- a/tests/regression/76-memleak/32-no-mem-leak-goblint-check.c
+++ b/tests/regression/76-memleak/32-no-mem-leak-goblint-check.c
@@ -1,0 +1,8 @@
+// PARAM: --set ana.activated[+] memLeak --set ana.malloc.unique_address_count 1
+#include <goblint.h>
+
+int main() {
+    int *ptr = malloc(sizeof(int));
+    __goblint_check(ptr); // UNKNOWN
+    free(ptr);
+}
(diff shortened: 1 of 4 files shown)


tests/regression/76-memleak/32-no-mem-leak-goblint-check.c
```diff
@@ -0,0 +1,8 @@
+// PARAM: --set ana.activated[+] memLeak --set ana.malloc.unique_address_count 1
+#include <goblint.h>
+
+int main() {
+    int *ptr = malloc(sizeof(int));
+    __goblint_check(ptr); // UNKNOWN
```
michael-schwarz, line 6: Probably unrelated, but it seems weird that this one is unknown, it should succeed, as `ptr` is supposed to be non-null here...

michael-schwarz (approved):

tests/regression/76-memleak/32-no-mem-leak-goblint-check.c
sim642: Perhaps the missing `#include <stdlib.h>` causes `malloc` to have a weird implicit type and casts by default.

sim642: @karoliineh checked it's not that. I now traced a bunch and realized what strange things are happening.

`__goblint_check` has an argument of `int` type, so CIL inserts a cast: `(int)ptr`. However, when pointers are converted to integers, the integer has type `Cil.upointType`, which is larger than `int`. So there's an integer downcast that can wrap to 0.
If the argument had `_Bool` type, then the implicit cast would use a special case for booleans and correctly exclude 0. I can open a PR to change our header, but I don't yet know if it would behave weirdly pre-C11. (CIL's `Machdep` generation defaults to `typedef int bool;` if `stdbool.h` is missing for some reason).

It's pretty interesting and makes me wonder if such issues can actually happen in some real programs using `int` for booleans as well (for checking that pointers are non-`NULL`).
The standard `assert` cleverly avoids this problem by only being a macro (not a function) with a "_scalar_ `expression`" argument (not necessarily `int`). So it can bypass the cast to `int`.

But then again, we have special stubs to redefine the `assert.h` macro to `__goblint_assert` which again has `int` argument! (To avoid handling compiler-specific macro implementations.)

michael-schwarz: Thanks for looking into all these details! Maybe we can come up with a regression test here that does not depend on this oddity, and then merge this PR?

I think it's fundamentally a different issue than the one we fix here.

michael-schwarz (approved):

PR #1684, opened 2025-02-17
Replace `int` -> `_Bool` in goblint stubs arguments 

sim642:
Follow-up on https://github.com/goblint/analyzer/pull/1682#discussion_r1958121037.

The replacement is straightforward but many tests seem to have failed, so this change is not as easy as it seemed.

### TODO
- [x] ~~Are there pre-C11 compatibility issues?~~ Nevermind, `_Bool` exists since C99.
- [x] Fix all tests.
- [x] Less hacky removal of `_Bool` casts? Try in `LibraryFunctions`.
- [x] Use `eval_bool` everywhere.
- [x] Merge #1814.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -40,3 +40,9 @@ c3e2cc848479ae86de5542b6ab0e75a74e9cf8c9
 
 # Rename ctx -> man
 0c155e68607fede6fab17704a9a7aee38df5408e
+
+# Trim trailing whitespace in BitfieldDomain
+d4e2a5f84ed3b7fbff89e34b2f7833de975e0671
+
+# Fix BaseInvariant indentation
+15e7a7ebd34e9fabdd4129e97eb86830fea8395f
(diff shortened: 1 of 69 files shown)


sim642:
Doing the `_Bool` cast stripping in `LibraryFunctions` doesn't completely work: `EvalInt` query on non-integers (like pointers) returns `Bot` to indicate the type mismatch instead of forcing it to an integer.

Having the assert analysis query with the cast added back (but not printed) seems to fix the issue.

sim642:
I now merged in master and used `Queries.eval_bool` in place of the TODOs.

@michael-schwarz This reveals a problem in the bitfield domain. In particular, it doesn't handle casts to `_Bool` correctly:
```
[Warning][Integer > Overflow][CWE-190] Unsigned integer overflow (tests/regression/83-bitfield/08-refine-with-bitfield.c:89:5-89:31)
[Error][Assert] Assertion "a & 0x80" produces a bottom. What does that mean? (currently uninitialized arrays' content is bottom) (tests/regression/83-bitfield/08-refine-with-bitfield.c:89:5-89:31)
```
The overflow warning there is also spurious.

It has to do with the fact that we consider `_Bool` to be 1-bit, so according to the bitfield domain casting 128 to `_Bool` overflows and something breaks in the cast as well.

PR #1688, opened 2025-02-18
Audit usages of `fold_right`

michael-schwarz:
This PR replaces usages of  `fold_right` with usages of `fold_left` wherever there is obviously no difference between both. This also raised an instances where I am unsure if the code in Goblint is correct.

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -482,9 +482,8 @@ struct
       match reachables ask es with
       | None ->
         (* top reachable, so try to invalidate everything *)
-        RD.vars st.rel
-        |> List.filter_map RV.to_cil_varinfo
-        |> List.map Cil.var
+        let to_cil_lval x = Option.map Cil.var @@ RV.to_cil_varinfo x in
+        RD.vars st.rel |> List.filter_map to_cil_lval
       | Some ad ->
         let to_cil addr rs =
           match addr with
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -465,17 +465,15 @@ struct
 
   (* Give the set of reachables from argument. *)
   let reachables (ask: Queries.ask) es =
-    let reachable e st =
-      match st with
-      | None -> None
-      | Some st ->
+    let reachable acc e =
+      Option.bind acc (fun st ->
         let ad = ask.f (Queries.ReachableFrom e) in
         if Queries.AD.is_top ad then
           None
         else
-          Some (Queries.AD.join ad st)
+          Some (Queries.AD.join ad st))
     in
-    List.fold_right reachable es (Some (Queries.AD.empty ()))
+    List.fold_left reachable (Some (Queries.AD.empty ())) es
 
 
   let forget_reachable man st es =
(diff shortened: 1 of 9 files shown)


src/analyses/malloc_null.ml
```diff
@@ -94,7 +94,7 @@ struct
   (* Remove null values from state that are unreachable from exp.*)
   let remove_unreachable (ask: Queries.ask) (args: exp list) (st: D.t) : D.t =
     let reachable =
-      let do_exp e a =
+      let do_exp a e =
```
michael-schwarz, line 97: This seems very odd, the return value depends on the order of the elements in `args`: As soon as a top pointer is encountered, AD is reset to `{}` but accumulation continues as normal after that?

michael-schwarz: Intuitively, one would like to:

- return `a` instead of `AD.empty ()`
- return some special value indicating that everything is to be considered reachable.

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -465,17 +465,15 @@ struct
 
   (* Give the set of reachables from argument. *)
   let reachables (ask: Queries.ask) es =
-    let reachable e st =
-      match st with
-      | None -> None
-      | Some st ->
+    let reachable acc e =
+      Option.bind acc (fun st ->
```
sim642, line 469: This could even use [let-punning](https://ocaml.org/manual/5.2/bindingops.html#ss%3Aletops-punning) after `let open GobOption.Syntax in` or something.

src/analyses/malloc_null.ml
```diff
@@ -164,7 +164,7 @@ struct
   let return_addr () = !return_addr_
 
   let return man (exp:exp option) (f:fundec) : D.t =
-    let remove_var x v = List.fold_right D.remove (to_addrs v) x in
+    let remove_var x v = List.fold_left (Fun.flip @@ D.remove) x (to_addrs v) in
```
sim642, line 167: I think this should also just work:
```suggestion
    let remove_var x v = List.fold_left (Fun.flip D.remove) x (to_addrs v) in
```
If so, then the same goes for some following places as well.

src/cdomain/value/cdomains/int/defExcDomain.ml
sim642, line 1: This file contains some unrelated changes.

michael-schwarz: It's some whitespace leftovers from a recent PR, plus a TODO related to this PR. Not sure if it's worth pulling that out.

src/analyses/apron/relationAnalysis.apron.ml
michael-schwarz: I played around with it, but was unconvinced whether it is easier / more beautiful:

~~~Ocaml
    let reachable acc e =
      let open GobOption.Syntax in
      let* st = acc in
      let ad = ask.f (Queries.ReachableFrom e) in
      if Queries.AD.is_top ad then
        None
      else
        Some (Queries.AD.join ad st)
    in
~~~

michael-schwarz: @jerhard @DrMichaelPetter @karoliineh @vesalvojdani  Opinions on monadic syntax or not?

DrMichaelPetter: Me as a contributor more on the novice side would rather see the explicit binding version instead of the let-punning - call me a simpleton, but I like it better the explicit way.

sim642 (approved):

PR #1689, opened 2025-02-18
Cleanup some `option` related code

michael-schwarz:


src/cdomain/value/domains/valueDomainQueries.ml
```diff
@@ -61,9 +61,7 @@ let eval_int_binop (module Bool: Lattice.S with type t = bool) binop (eval_int:
   if ID.is_bot i || ID.is_bot_ikind i then
     Bool.top () (* base returns bot for non-int results, consider unknown *)
   else
-    match ID.to_bool i with
-    | Some b -> b
-    | None -> Bool.top ()
+    BatOption.default (Bool.top ()) (ID.to_bool i)
```
sim642, line 64: This might make no difference here, but previously `Bool.top` was only called when actually needed, now it's always called but possibly ignored.
There is `BatOption.default_delayed` for that purpose. Maybe there are other instances where that would be more needed.

src/common/framework/cfgTools.ml
```diff
@@ -432,9 +432,10 @@ let createCFG (file: file) =
                   List.iter (fun (fromNode, toNode) ->
                       addEdge_fromLoc fromNode (Test (one, false)) toNode;
                       added_connect := true;
-                      match NH.find_option node_scc toNode with
-                      | Some toNode_scc -> iter_scc toNode_scc (* continue to target scc as normally, to ensure they are also connected *)
-                      | None -> () (* pseudo return, wasn't in scc, but is fine *)
+                      Option.iter (fun toNode_scc ->
+                          iter_scc toNode_scc (* continue to target scc as normally, to ensure they are also connected *)
+                        ) (NH.find_option node_scc toNode)
```
sim642, line 437: I think this should work too:
```suggestion
                      Option.iter iter_scc (NH.find_option node_scc toNode) (* continue to target scc as normally, to ensure they are also connected *)
```

src/constraint/constrSys.ml
```diff
@@ -294,6 +292,7 @@ struct
               f get set
             )
         in
-        Some f'
+        f'
+        ) (S.system x)
```
sim642, line 296: I think usually we have the lambda body indented by one more level in calls like this.

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -291,16 +291,14 @@ struct
...
-  let pass_to_callee fundec any_local_reachable var =
-    (* TODO: currently, we pass all locals of the caller to the callee, provided one of them is reachbale to preserve relationality *)
+  let belongs_to_fundec fundec var =
+    (* TODO: currently, we pass all locals of the caller to the callee, provided one of them is reachable to preserve relationality *)
     (* there should be smarter ways to do this, e.g. by keeping track of which values are written etc. ... *)
     (* See, e.g, Beckschulze E, Kowalewski S, Brauer J (2012) Access-based localization for octagons. Electron Notes Theor Comput Sci 287:29–40 *)
     (* Also, a local *)
     let vname = GobApron.Var.show var in
-    let locals = fundec.sformals @ fundec.slocals in
-    match List.find_opt (fun v -> VM.var_name (Local v) = vname) locals with (* TODO: optimize *)
```
sim642, line 301: I suspect the `TODO: optimize` should be preserved. I guess it's about using repeated string conversions to find some variable.

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -438,20 +434,18 @@ struct
...
+      let unify_st' = Option.map_default (fun lv ->
           let ask = Analyses.ask_of_man man in
           assign_to_global_wrapper ask man.global man.sideg unify_st lv (fun st v ->
               let rel = RD.assign_var st.rel (RV.local v) RV.return in
               assert_type_bounds ask rel v (* TODO: should be done in return instead *)
             )
-        | None ->
-          unify_st
+        )
+          unify_st r
```
sim642, line 445:
```suggestion
        ) unify_st r
```

src/analyses/base.ml
```diff
@@ -2851,19 +2808,17 @@ struct
...
             | Some voidVal when Addr.type_of addr = voidType -> {st with cpa = CPA.add v voidVal st.cpa}
             | _ ->
-              begin
-                let address = AD.singleton addr in
-                let new_val = get ~man fun_st address None in
-                if M.tracing then M.trace "taintPC" "update val: %a" VD.pretty new_val;
-                let st' = set_savetop ~man st address lval_type new_val in
-                match Dep.find_opt v fun_st.deps with
-                | None -> st'
-                (* if a var partitions an array, all cpa-info for arrays it may partition are added from callee to caller *)
```
sim642, line 2861: Should the comment remain?

sim642 (commented):
This simplifies some uses but I'm unsure about others.
In particular, replacing some `match`es with large bodies with `Option.map_default` (or similar) sometimes moves what's being `match`ed much further down (after the lambda). This might hurt readability because the code doesn't read top-down as it executes.

I think it would be good for someone else to also have a look with this in mind to see if they also get this impression.

src/analyses/accessAnalysis.ml
arkocal, line 1: The changes look fine, but is there a reason we are using stdlib Option and not BatOption here?

src/cdomain/value/domains/valueDomainQueries.ml
michael-schwarz: I tried to use it where functions seemed expensive. Given it is a constant here, I would expect `flambda` to inline the boolean (but I may be wrong).

src/analyses/apron/relationAnalysis.apron.ml
michael-schwarz: I thought it was about the `@` which is completely unnecessary, but I added it back.

src/analyses/accessAnalysis.ml
michael-schwarz: `BatOption` is missing `iter`.

src/analyses/base.ml
```diff
@@ -2068,6 +2053,7 @@ struct
           man.sideg (V.thread tid) (G.create_thread rv);
           Priv.thread_return ask (priv_getg man.global) (priv_sideg man.sideg) tid st'
         | _ -> st'
+        ) nst exp
```
sim642: The indentation seems off here.

PR #1690, opened 2025-02-19
HashCachedContextLifter: Introduce the lifter and the option

arkocal:
Using hashcache on contexes makes a big difference in memory usage (churn) and a reasonable improvement on runtime when HashConsing is disabled. This is particularly important for the upcoming parallel solvers

### DIFF
--- a/src/common/domains/printable.ml
+++ b/src/common/domains/printable.ml
@@ -176,6 +176,7 @@ struct
 
   let lift m = {m; lazy_hash = LazyHash.make m}
   let unlift {m; _} = m
+  let relift x = x
 
   let lift_f f x = f (unlift x)
   let lift_f' f x = lift @@ lift_f f x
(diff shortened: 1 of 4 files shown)


src/common/domains/printable.ml
```diff
@@ -176,6 +176,7 @@ struct
 
   let lift m = {m; lazy_hash = LazyHash.make m}
   let unlift {m; _} = m
+  let relift x = x
```
michael-schwarz, line 179: Would I not need to call `relift` for `M` and reset cached hashes here?

arkocal: Why should the hashes change?

arkocal: For future reference: This is now included in the next commit to be sure.

michael-schwarz (approved):

sim642 (approved):
With hashconsing there are separate functors for applying it do `D` and `C`. This only applies to `C`.
Is there a reason that hashcaching `C` is particularly necessary while on `D` it doesn't really matter?

arkocal:
> With hashconsing there are separate functors for applying it do `D` and `C`. This only applies to `C`. Is there a reason that hashcaching `C` is particularly necessary while on `D` it doesn't really matter?

After analyzing with memtrace, I have found about 90% of memory usage that we can cut down with this comas from `C`. It can be used on `D` as well, but it is not as critical.

PR #1693, opened 2025-02-20
Add explicit W set to base protection privatization

sim642:
Regarding https://github.com/goblint/analyzer/pull/1636#discussion_r1963055883.

Surprisingly, there was even a TODO about it:
https://github.com/goblint/analyzer/blob/a9354564fbb6b3d86879a4949f9d93c04399584c/src/analyses/basePriv.ml#L940-L943

### DIFF
--- a/src/analyses/basePriv.ml
+++ b/src/analyses/basePriv.ml
@@ -899,18 +899,25 @@ struct
     let name () = "P"
   end
 
-  (* W is implicitly represented by CPA domain *)
-  module D = P
+  (** May written variables. *)
+  module W =
+  struct
+    include MayVars
+    let name () = "W"
+  end
+
+  module D = Lattice.Prod (P) (W)
 
   module Wrapper = Wrapper (VD)
   module G = Wrapper.G
   module V = ProtectionBasedV.V
 
-  let startstate () = P.empty ()
+  let startstate () = P.empty (), W.empty ()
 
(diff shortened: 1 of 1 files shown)


michael-schwarz (commented):
This looks reasonable! Could you maybe add a regression test where this make a difference?

I will try if this also behaves the intended way in the context of #1636. If it does and after adding a test, we should be good to go!

michael-schwarz (changes requested):
It falls short of what the snippet in #1636 achieves, likely because that one encodes a cheap path-sensitivity like behavior.

I pushed the merged verison to https://github.com/goblint/analyzer/tree/tmp_dn_1693 so we can play around with it.

sim642:
Hmm, I suppose keeping separate `CPA` maps for written and refined values could be more precise indeed.
Although that doesn't seem like it should to be specific to Protection-Based Reading, but potentially all/many privatizations?

michael-schwarz:
> Although that doesn't seem like it should to be specific to Protection-Based Reading, but potentially all/many privatizations?

Sure! I think for `base` we know that it pays off. Not sure whether it's worth adding it to all privatizations?

sim642:
To each one individually, probably not. I was thinking more along the lines of some kind of generic lifting with the dual `CPA`s or something. I don't have a specific idea though, so it might not really be possible so generally.

Issue #1696, opened 2025-02-27
Remaining verdicts `Error (both branches dead)` on SV-COMP


michael-schwarz:
#1587 helped fixed some of these, but unfortunately not all: Before there were 118 instances, after it, we still have 90 instances left. 

Most of them are still from the `hardness` suite proposed in the NFM paper by Berger et al., but some are from LDV too.

- [ ] `hardness-nfm22/hardness_codestructure_dependencies_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_normal_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-100_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-10_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-250_file-24.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-250_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-250_file-73.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-25_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-500_file-13.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-500_file-18.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-500_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-500_file-7.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-500_file-74.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-500_file-92.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-50_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodestructure_filler-pe-ci_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodestructure_filler-pe-cn_file-6.yml`
- [ ] `hardness-nfm22/hardness_fillercode_fillercodestructure_filler-pe-co_file-6.yml`
[...]

PR #1698, opened 2025-02-27
Relational: Assert type bounds after `branch`

### LINKED ISSUE #1697: Both branches dead for short circuiting `&&` and Octagons
~~~C
//PARAM: --enable ana.int.interval --set ana.activated[+] apron
int main() {
    int a;
    unsigned short b;
    b = a;
    1 && (b + b < b ? b ?: 1 : 1);
    return 0;
}
~~~

yields

> [Error][Analyzer][Unsound] both branches over condition 'tmp' are dead (tests/regression/25-vla/07-stuff.c:6:5-6:34)

This is minimized from `hardness-nfm22/hardness_variablewrapping_normal_file-0.yml`. I tried to keep simplifying the program, but that makes the problem go away.

michael-schwarz:
After branches, it pays off to check that computed constraints remain consistent with type-bounds. 

This closes #1697 where this issue manifested as a `both branches dead` message. The test is for an equivalent `arithmetic on integer bot` case.

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -272,6 +272,8 @@ struct
         RD.assert_inv ask rel' e' (not b) (no_overflow ask e)
       )
     in
+    let vars = Basetype.CilExp.get_vars e |> List.unique ~eq:CilType.Varinfo.equal |> List.filter RD.Tracked.varinfo_tracked in
+    let res = List.fold_left (assert_type_bounds ask) res vars in
     if RD.is_bot_env res then raise Deadcode;
     {st with rel = res}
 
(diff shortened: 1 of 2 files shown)


DrMichaelPetter (approved):

sim642 (changes requested):
The CI shows a lot of test failures with this.

What I'm wondering is why type bounds would be lost by `branch` to begin with?
I guess this assertion wasn't added there because intuitively type bounds would persist which would make this unnecessary, but seems like that's not the case?

michael-schwarz:
> The CI shows a lot of test failures with this.

Yeah, I used `e` instead of `e'` by mistake.

> What I'm wondering is why type bounds would be lost by `branch` to begin with?

This is avery good question I also have no answer to.

michael-schwarz:
The remaining regression test are ones where the termination analysis tries to build non-terminating loops over `int`, which does not work because that analysis set the `assume_none` flag.

sim642:
> The remaining regression test are ones where the termination analysis tries to build non-terminating loops over `int`, which does not work because that analysis set the `assume_none` flag.

This might be the place where I wondered about a https://github.com/goblint/analyzer/issues/1696#issuecomment-2687941671 -like situation before: what's supposed to happen if we want to assume that something, which definitely happens, does not happen?
It's the kind of nonsense where UB allows compilers to do weird things (delete the loop altogether or whatever), but I don't know what an analyzer should do about it.

michael-schwarz:
Closed in favor of #1699.

PR #1699, opened 2025-02-27
Relational: Assert type bounds after `assign` & Fix loop tests

michael-schwarz:
**This is an alternative fix to #1697 that we could consider instead of #1698 (I prefer it to #1698 in fact).**

Compared to that one, the difference is that, instead of after an `assert`, type bounds are asserted after an `assign`, which maybe is the more logical place to do it, as the assert should simply not lose the bounds but needn't establish new ones.

This precision improvement then raises the following pre-existing issue: Our termination analysis assumes signed overflows do not happen (this is, e.g., an assumption one can make for `SV-COMP` as programs there contain no UB).

However, this means it is trivial for our analysis to establish that any loop over an `int` which is increased in each iteration terminates.

~~~C
int i = 0; 
while(1) {
 i++;
}
~~~ 

is found to terminate as the inserted iteration counter is established to be equal to `i`, and `i` itself is "bounded" by `MAX_INT`. This issue is already there before this PR.
However, with this PR, most termination regression tests are found to be trivial, and some are incorrect. Thus, most of them are changed to iterate over `unsigned` types here.

### DIFF
--- a/tests/regression/78-termination/14-do-while-nonterminating.c
+++ b/tests/regression/78-termination/14-do-while-nonterminating.c
@@ -3,12 +3,14 @@
 
 int main()
 {
-  int i = 1;
+  unsigned int i = 1;
 
   do
   {
     printf("Inside the do-while loop\n");
     i++;
+
+    if(i ==0) {i = 2;}
   } while (i >= 2); // NONTERMLOOP termination analysis shall mark while as non-terminating loop
 
   printf("Exited the loop\n");
--- a/tests/regression/78-termination/25-leave-loop-goto-terminating.c
+++ b/tests/regression/78-termination/25-leave-loop-goto-terminating.c
@@ -1,9 +1,9 @@
-// SKIP TERM PARAM: --set "ana.activated[+]" termination --set ana.activated[+] apron --enable ana.int.interval --set ana.apron.domain polyhedra
+// SKIP TODO TERM PARAM: --set "ana.activated[+]" termination --set ana.activated[+] apron --enable ana.int.interval --set ana.apron.domain polyhedra
 #include <stdio.h>
 
 int main()
 {
-  int counter = 0;
+  unsigned int counter = 0;
 
   while (1)
   {
--- a/tests/regression/78-termination/29-do-while-continue-nonterminating.c
+++ b/tests/regression/78-termination/29-do-while-continue-nonterminating.c
@@ -3,12 +3,14 @@
 
 int main()
 {
-  int i = 1;
+  unsigned int i = 1;
 
   do
   {
     printf("Inside the do-while loop\n");
     i++;
+
+    if(i ==0) {i = 2;}
 
     if (i % 2)
     {
--- a/tests/regression/78-termination/34-nested-for-loop-nonterminating.c
+++ b/tests/regression/78-termination/34-nested-for-loop-nonterminating.c
@@ -3,11 +3,11 @@
 
 int main()
 {
-  int outerCount, innerCount;
+  unsigned int outerCount, innerCount;
 
   for (outerCount = 1; outerCount <= 3; outerCount++)
   {
-    for (innerCount = 1; innerCount > 0; innerCount++) // NONTERMLOOP termination analysis shall mark beginning of for as non-terminating loop
+    for (innerCount = 1; innerCount > 0; innerCount += 4) // NONTERMLOOP termination analysis shall mark beginning of for as non-terminating loop
     {
       printf("(%d, %d) ", outerCount, innerCount);
     }
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -256,8 +256,9 @@ struct
             if M.tracing then M.traceli "relation" "assign inner %a = %a (%a)" CilType.Varinfo.pretty v d_exp e' d_plainexp e';
             if M.tracing then M.trace "relation" "st: %a" RD.pretty apr';
             let r = RD.assign_exp ask apr' (RV.local v) e' (no_overflow ask simplified_e) in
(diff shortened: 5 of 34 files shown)


sim642:
This makes more sense to me than #1698. Although it's still not fully clear to me why precision would be lost by the assignment to begin with.

The relation analysis checks for overflows in the right-hand side. If there isn't any, then it should have already proven that the value fits into the type bounds and this extra assertion wouldn't be necessary.
Or is this only needed for the case where the right-hand side has a possible overflow and the assigned variable is topped? Re-asserting its type bounds then seems right.

The other thing I'm wondering about is the termination stuff. Why does using `unsigned int` over `int` make any difference?
An `unsigned int` variable still is bounded by the type (or are we for some reason not re-asserting its bounds somewhere?), which has smaller range than the top of `unsigned long long` used for the termination analysis ghost counters.

michael-schwarz:
> Although it's still not fully clear to me why precision would be lost by the assignment to begin with.
> Re-asserting its type bounds then seems right.

In the example, the type bounds were indeed lost after an assignment ` b = (unsigned short )a;`, so I guess it is at assignments between different types where we do not establish the bounds. Iirc, we don;'t have type information at the inner level where the assignment happens anyway.


> The other thing I'm wondering about is the termination stuff. Why does using unsigned int over int make any difference?

The issue arises completely independently of our implementation:
The problem is that any loop which increases an `int` in each iteration (or after finitely many iterations) is bounded by definition, as we know that overflows do not happen (our assumption), so proving a property about such loops is completely vacuous. Sadly, almost all tests were of this nature.

Loops that range over unsigned types, on the other hand, may be truly infinite without UB even if they increase in each iteration, so there is a proof obligation here.

> An unsigned int variable still is bounded by the type (or are we for some reason not re-asserting its bounds, which has smaller range than the top of unsigned long long used for the termination analysis ghost counters.

Yes, this indeed is true, but is not the point here. It is not the boundedness of any program variables that we are interested in, but whether our added variable (which is by construction incremented in each iteration and starts at zero) can always be shown to be different from `top` at the loop head.

While the loop

~~~C
    for(int i = 0;1;i++) {

    }
~~~

is bounded (we can establish `lc = i` because an increment past `MAX_INT` for `i` cannot happen and we can bound `i` by `MAX_INT`),

~~~C
    for(unsigned int i = 0;1;i++) {
    }
~~~

is not (`i` is bounded by the type bounds), but `lc` is not bounded by it, but only by its type bounds.

Thus, most of the tests were not testign anything and they now do.

michael-schwarz:
Our choice of `unsigned long long` implies that there are some loops (over `unsigned long long`) for which we can not show termination, even when they terminate, namely when they have `MAX_ULL` (+1 ?) iterations.

~~~C
unsigned long long i = 0;

while(i <= MAX_ULL) { i++; }
~~~

Here, we would have top for our variable `lc` and thus claim that the loop may not terminate, even when it in fact does.

tests/regression/78-termination/14-do-while-nonterminating.c
```diff
@@ -3,12 +3,14 @@
...
 {
-  int i = 1;
+  unsigned int i = 1;
 
   do
   {
     printf("Inside the do-while loop\n");
     i++;
+
+    if(i ==0) {i = 2;}
```
sim642, line 13: What's up with this?

tests/regression/78-termination/25-leave-loop-goto-terminating.c
```diff
@@ -1,9 +1,9 @@
-// SKIP TERM PARAM: --set "ana.activated[+]" termination --set ana.activated[+] apron --enable ana.int.interval --set ana.apron.domain polyhedra
+// SKIP TODO TERM PARAM: --set "ana.activated[+]" termination --set ana.activated[+] apron --enable ana.int.interval --set ana.apron.domain polyhedra
```
sim642, line 1: Why doesn't this work anymore?

tests/regression/78-termination/29-do-while-continue-nonterminating.c
```diff
@@ -3,13 +3,15 @@
...
 {
-  int i = 1;
+  unsigned int i = 1;
 
   do
   {
     printf("Inside the do-while loop\n");
     i++;
 
+    if(i ==0) {i = 2;}
```
sim642, line 13: What's up with this?

tests/regression/78-termination/34-nested-for-loop-nonterminating.c
```diff
@@ -3,11 +3,11 @@
 
 int main()
 {
-  int outerCount, innerCount;
+  unsigned int outerCount, innerCount;
 
   for (outerCount = 1; outerCount <= 3; outerCount++)
   {
-    for (innerCount = 1; innerCount > 0; innerCount++) // NONTERMLOOP termination analysis shall mark beginning of for as non-terminating loop
+    for (innerCount = 1; innerCount > 0; innerCount += 4) // NONTERMLOOP termination analysis shall mark beginning of for as non-terminating loop
```
sim642, line 10: What's up with this `+= 4`?

michael-schwarz: Unless the loop condition is modified, this inner loop would terminate once `innerCount` reaches `0`. I thus decided to have a little fun with it, and increment by `4` in each iteration.

tests/regression/78-termination/25-leave-loop-goto-terminating.c
michael-schwarz: I'm not even sure why this worked in the first place, we even have the comment 

```
// Apron is not able to detect this
```

in the code. I think before it only worked because all proof obligations over loops where a signed variable is incremented in each iteration are trivial.

sim642 (approved):

PR #1701, opened 2025-02-28
Fix refinement in `baseInvariant` for more complicated lvals

michael-schwarz:
This is a follow-up to #1659 which was incorrect as demonstrated by the nightly SV-COMP run.

The offset handling was slightly off there, leading to points-to sets becoming empty when they should not. The correct thing is to:

- From an `Mem (Lval lv), off` pair, take the `Lval lv` part (including any possible offsets it may have), and consider all the values of those separately  by
    - Refining the value for `Lval lv` in base
    - Evaluating `Mem (Lval lv), off` w.r.t. to this new state (for technical reasons computed by appending off to the value of `Lval lv` - queries use stale values).
    - Setting the result to the meet, resp. raise Deadcode if it is `bot`.
    
 I did not test all cases, but the unsoundness is fixed and all the `BothBranchesDead` ones I tested as well, so I would suggest to merge this right away, as the current state is broken.

### DIFF
--- a/tests/regression/03-practical/36-struct-ptr.c
+++ b/tests/regression/03-practical/36-struct-ptr.c
@@ -1,21 +1,20 @@
 #include <goblint.h>
 struct aws_al {
 	unsigned long current_size;
-  };
+};
 
-  struct aws_pq {
-	  int pred;
-	  struct aws_al container;
-  };
+struct aws_pq {
+	int pred;
+	struct aws_al container;
+};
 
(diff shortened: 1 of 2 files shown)


tests/regression/03-practical/36-struct-ptr.c
sim642, line 1: Indentation is weird in this file.

sim642 (approved):

PR #1703, opened 2025-03-04
Fix flaky cram tests where `vid`s leak into witnesses

### LINKED ISSUE #1647: Cram tests not robust enough
Locally, I get a bunch of errors of the type (for various things):

```
+++ b/_build/.sandbox/7396a923cc6398b0b3f839e1fce54e87/default/tests/regression/56-witness/66-ghost-alloc-lock.t.corrected
@@ -20,13 +20,13 @@
   - entry_type: ghost_instrumentation
     content:
       ghost_variables:
-      - name: alloc_m559918035_locked
+      - name: alloc_m135980175_locked
         scope: global
         type: int
         initial:
           value: "0"
           format: c_expression
-      - name: alloc_m861095507_locked
+      - name: alloc_m349172907_locked
         scope: global
         type: int
         initial:
@@ -46,7 +46,7 @@
           column: 3
           function: t_fun
         updates:
-        - variable: alloc_m559918035_locked
+        - variable: alloc_m135980175_locked
           value: "1"
           format: c_expression
```

However, those things work in the CI... 
 
 @sim642 suspects it has to do with `vid`s leaking into the names of ghost variables, which is not really stable across version.

michael-schwarz:
This works by canonizing the names such as `alloc_m559918035_locked` to `TEMP_VAR1` and `TEMP_VAR2`. The bash script comes from `deepseek`, so it might make sense if someone who knows bash has a look (CC @DrMichaelPetter). 

Closes #1647.


tests/regression/56-witness/66-ghost-alloc-lock.t
```diff
@@ -16,136 +16,6 @@
...
-      type: assertion
-      format: C
-  - entry_type: flow_insensitive_invariant
-    flow_insensitive_invariant:
-      string: '! multithreaded || (0 <= g1 && g1 <= 1)'
-      type: assertion
-      format: C
+  $ (yamlWitnessStrip < witness.yml) > new-stripped.yml
+  $ ./compare-ghost-alloc.sh 66-ghost-alloc-stripped.yml new-stripped.yml
+  The files are the same after renaming the variables.
```
sim642, line 21: This moves the diffing into the script rather than having dune do it. Also it hides the additional stripping being done which means that the additional stripping is assumed to be correct: if it was wrong, then it strips both incorrectly and we'd never see.

So I'd prefer a simpler solution: the additional script just does renamings in the produced (and YAML stripped) witness and that's outputted.
Then the final thing is clearly visible in the test and the diff should also be obvious: `alloc_*` variables are replaced with something deterministic.

DrMichaelPetter (approved):
LGTM

michael-schwarz:
After @DrMichaelPetter's suggestions, I also used `shellcheck` (https://www.shellcheck.net/) and it is happy with the script too.

sim642 (approved):

PR #1705, opened 2025-03-07
Add bash completion for array element addition and removal

sim642:
This allows bash completing things like `--set witness.yaml.entry-types[+]` to offer a list of valid things to add or remove (with `[-]`) from the list.


michael-schwarz (commented):
I personally don't have argument completion set up, so I didn't test it. However, it seems like an innocent local thing we should probably just merge.

arkocal:
Works fine on my system.

Issue #1550, opened 2024-08-01
Timing for multithreaded use


FelixKrayer:
While implementing parallel solvers for Goblint, it became apparent that the timing functions used in Goblint are not thread-safe. Thus, issues arise, when multiple domains running in parallel call the `Timing.wrap` function
The execution with more than one parallel Domain often fails because of an assertion error in the `exit` function in the module `Make` of `goblint_timing.ml` (line 134: `assert (tree.name = name)`). I assume the likely cause for that is that the `current` stack is not thread-safe.

Furthermore, it has to be decided how timing would even work with multiple parallel Domains.
When two Domains run in parallel and both provide a time value for a specific timed section, how would these time contributions be merged? Is it more reasonable to add these values or to use the largest value? Adding would more accurately show the work done, but would ignore the potential speedup through parallelization. Thus a final timing value generated by adding parallel times could potentially be longer than the whole run of Goblint.

sim642:
Indeed, the big question is how it's even supposed to work. The current implementation (improved from CIL's `Stats`) requires well-nested `enter` and `exit` calls globally, which `wrap` from multiple domains does not guarantee.

Our timing measures both CPU time and wall time. Summing CPU times across domains would make sense because that's the total amount of work done by the CPU. Although currently we just look it up by `Unix.times` which is about the entire process. I'm not sure if there's any way to even get CPU time used per-domain.

Measurement of wall time itself is easy, but aggregation is hard, neither sum nor max makes sense. Since parallelism is only to be used in a specific place (not parsing, etc), it might simply have to be that there's wall time for the entire `solving`, but not its sub-timings which have been aggregated.

We also measure allocated memory using `Gc.allocated_bytes`, which I also don't know how it works with multiple domains.

It would be worth trying to find others (not necessarily OCaml) that try to profile multi-threaded programs and see if they've come up with anything better.

How urgent is this? Can we get by by just disabling timing for our multicore attempts? The minimum fix might be to make the `current` stack in the timing implementation use `DLS`.

FelixKrayer:
For now, it is only important that errors are avoided in a temporary fix (as you mentioned through DLS or disabling of the timing). We can get timings for benchmarks in other ways. For example, we can measure the time of the whole solving process which should not be an issue.
However, I believe that we should find a suitable solution before merging the parallelized solvers into the master branch. There are other Issues to be solved before a merge as well, e.g. the performance issues from Ocaml 5 #1137 since this version is a prerequisite for the Domains.

PR #1708, opened 2025-03-11
Introduce backwards compatible infrastructure for parallelism

arkocal:
This contains the rather harmless parts. As discussed, all variations of modules have been implemented using `domain_shims` and the select stanza in dune. `ppx_optcomp` is not involved anymore.

The PR is rather large, but can be reviewed commit-by-commit.

### DIFF
--- a/goblint.opam
+++ b/goblint.opam
@@ -73,6 +73,7 @@ depends: [
 depopts: [
   "apron" {>= "v0.9.15"}
   "z3"
+  "domainslib"
 ]
 conflicts: [
   "result" {< "1.5"}


arkocal:
Turns out introducing domains_shims breaks GobView. The application than outputs 
```
Uncaught TypeError: runtime.caml_thread_self is not a function
```
on the console and hangs indefinitely.

michael-schwarz:
> Turns out introducing domain_shims breaks GobView. The application than outputs

Do you know where this call happens? Is it something we can safely stub to just return unit or somehow avoid these calls when in Gobview mode?

michael-schwarz:
I guess we either need to turn the gobview job into an unlocked one, or provide a different lockfile for the gobview job somehow?

src/config/options.schema.json
```diff
@@ -2517,6 +2517,12 @@
               "description": "Check TD3 data structure invariants",
               "type": "boolean",
               "default": false
+            },
+            "parallel_domains" :{
+              "title": "solvers.td3.parallel_domains",
+              "description": "Maximal number of Domains that the solver can use in parallel. Only applies, when a solver of the 'td_parallel_*' family is used.",
+              "type": "integer",
+              "default": 2
```
sim642, line 2525: Seems like the existing `jobs` option would also work for this.

src/common/util/messages.ml
```diff
@@ -195,10 +197,14 @@ let print ?(ppf= !formatter) (m: Message.t) =
     | Debug -> "white" (* non-bright white is actually some gray *)
     | Success -> "green"
   in
-  let pp_prefix = Format.dprintf "@{<%s>[%a]%a@}" severity_stag Severity.pp m.severity Tags.pp m.tags in
+  let pp_print_option ?(none = fun _ () -> ()) pp_v ppf = function
+    | None -> none ppf ()
+    | Some v -> pp_v ppf v
+  in
```
sim642, line 203: What's this about? Isn't it just a copy of what's in `Stdlib.Format`?

src/cdomain/value/cdomains/mutexAttrDomain.ml
```diff
@@ -21,7 +22,7 @@ end
 include Lattice.FlatConf (struct include Printable.DefaultConf let bot_name = "Uninitialized" let top_name = "Top" end) (MutexKind)
 
 (* Needed because OS X is weird and assigns different constants than normal systems... :( *)
-let recursive_int = lazy (
+let recursive_int = DomainsafeLazy.from_fun (fun () ->
```
sim642, line 25: Why is `DomainsafeLazy` only used here? There are other top-level `lazy` values like
* `Offset.Index.Exp.all`
* `Cilfacade.any_index_exp`
* `LibraryFunctions.intmax_t`

There's also `ResettableLazy` which is used all over. Are all of those safe then?

src/lifters/wideningTokenLifter.ml
```diff
@@ -12,13 +12,13 @@ module Token = WideningToken
 module TS = SetDomain.ToppedSet (Token) (struct let topname = "Top" end)
 
 (** Reference to current {!add} implementation. Maintained by {!Lifter}. *)
-let add_ref: (Token.t -> unit) ref = ref (fun _ ->
+let add_ref: (Token.t -> unit) Domain.DLS.key = Domain.DLS.new_key (fun () _ ->
```
sim642, line 15: Similar question here: we have a lot of top-level `ref`s around, but only this one is made domain-local. Are all others fine then?

src/util/timing/goblint_timing.ml
```diff
@@ -15,8 +15,8 @@ let next_tef_pid = ref 0
 
 module Make (Name: Name): S =
 struct
-  let enabled = ref false
-  let options = ref dummy_options
+  let enabled_dls = Domain.DLS.new_key (fun () -> false)
+  let options_dls = Domain.DLS.new_key (fun () -> dummy_options)
```
sim642, line 19: Is it necessary for these to be domain-local? These are more like global options than state.
#1550 mentions that maybe only `current` needs to be domain-local.

src/util/tracing/goblint_tracing.ml
```diff
@@ -38,9 +39,13 @@ let traceTag (sys : string) : Pretty.doc =
   let rec ind (i : int) : string = if (i <= 0) then "" else " " ^ (ind (i-1)) in
   (text ((ind !indent_level) ^ "%%% " ^ sys ^ ": "))
 
+let trace_mutex = GobMutex.create ()
```
sim642, line 42: Tracing isn't the only output from Goblint. Why don't others (like `Logs` and `Messages`) also need locks?

src/config/options.schema.json
michael-schwarz: I think it's worth having the concurrency settings for non-core stuff (g2html, preprocessing) separate from the solver concurrency, especially when comparing runtimes.

src/cdomain/value/cdomains/mutexAttrDomain.ml
michael-schwarz: Good point! Also what about `LazyEval`? Are we fine with force being called concurrently there?

src/util/tracing/goblint_tracing.ml
michael-schwarz: Those should happen at the end in postprocessing, which is not concurrent afaik.

sim642: Logging with `Logs` can happen at any time I think, although it might just be that the transfer functions rarely use it (and prefer `Messages`, for a good reason), but I haven't checked.
`Messages` could be used with `warn_at` being `early`, it's just not the default.

arkocal: I would leave this for the later PRs that contain the actual solver implementations. Traces here are included as an example for handling the kinds of situations where the concurrent implementation differs from the sequential. 

Practically: We need a shared mutex for all kinds of inputs (because logging  and tracing can interfere as well). However, this was not an issue during development.

src/common/util/messages.ml
arkocal: The module now uses `BatFormat`, which does not define the method.

src/util/timing/goblint_timing.ml
arkocal: Felix himself has introduced all of them in https://github.com/goblint/analyzer/commit/62422e9a14dedbb913c5fabc2bdb2cd2fa37c1b0 

As far as can see `enabled_dls` is rather tracking if it is active in a given moment (set via `start` and `stop`), and making it a reference makes the solver crash at the end. Options however are global options and can be made into refs.

9e1614d4f9b93b809d135a8d48671d1f367500b3 reflects this now.

src/cdomain/value/cdomains/mutexAttrDomain.ml
arkocal: These two were the ones discovered by Felix back then.

Currently it is quite out of scope to go through all analysis and decide for each if they are thread-safe. In order to keep this moving (this branch does not have the solvers and they are slowly diverging) we can:
a. remove this for now, continue with the introduction of `create` nodes and re-add this once the solvers are in. Then we can test more carefully to see if this really is necessary and why.
b.  Mark this with a comment, make an issue and make a more systematic analysis of the whole situation later.

sim642:
Looks like the Copilot reviews won't be too useful for us right now since it doesn't want to review .ml files.

src/util/timing/goblint_timing.ml
sim642: > As far as can see `enabled_dls` is rather tracking if it is active in a given moment (set via start and stop), and making it a reference makes the solver crash at the end.

But we only ever call `start` at the very beginning of Goblint and never even call `stop`. So I don't understand why there would be any issue. `start` also modifies `options` and that's fine.

src/common/util/messages.ml
sim642: But why `BatFormat`? The standard library version from OCaml 5 should be safe: it's been reimplemented to use DLS properly. Batteries as no domains-specific fixes.

src/cdomain/value/cdomains/mutexAttrDomain.ml
sim642: I would be OK with leaving this as the only case right now, but having a new issue to list the possible other places we've listed above.

src/util/parallel/threadpool.no-domainslib.ml
sim642, line 1: Shouldn't there be a non-parallel implementation here?

src/util/parallel/dune
```diff
@@ -0,0 +1,18 @@
...
+ (libraries
+   batteries
+   (select gobMutex.ml from
+      (domainslib -> gobMutex.domainslib.ml)
+      ( -> gobMutex.no-domainslib.ml)
+   )
+   (select threadpool.ml from 
+      (domainslib -> threadpool.domainslib.ml)
+      (-> threadpool.no-domainslib.ml)
+      )
```
sim642, line 15:
```suggestion
    )
```

src/util/parallel/gobMutex.domainslib.ml
```diff
@@ -0,0 +1,51 @@
+(* Simple Mutex Implementation using Domain-Local Await (https://github.com/ocaml-multicore/domain-local-await)
+   Copyright © 2023 Vesa Karvonen
+
+   Permission to use, copy, modify, and/or distribute this software for any purpose with or without fee is hereby granted,
+   provided that the above copyright notice and this permission notice appear in all copies.
+
+   THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES
+   OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY
+   DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+   ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. *)
```
sim642, line 10: The big question with this is whether this licence just allows us to swallow the code into MIT-licenced Goblint.

src/util/parallel/dune
```diff
@@ -0,0 +1,18 @@
...
+ (public_name goblint.parallel)
+ (libraries
+   batteries
+   (select gobMutex.ml from
+      (domainslib -> gobMutex.domainslib.ml)
+      ( -> gobMutex.no-domainslib.ml)
+   )
+   (select threadpool.ml from 
+      (domainslib -> threadpool.domainslib.ml)
+      (-> threadpool.no-domainslib.ml)
```
sim642, line 14: I think it's a good idea to have `.mli` files for these as well (not dependent on domainlib presence).

src/util/parallel/gobMutex.domainslib.ml
michael-schwarz: Ther ISC license is compatible in spirit with the MIT license. Maybe we just leave the license notice in this file as is and add at the top in the LICENSE.md that the MIT license applies to all files unless specified otherwise?

I think polluting the opam repo with a package containing only this file for trivial license stuff may be excessive.

sim642: Since domainslib explicitly integrates domain-local-await, but not picos primitives, it looks like this will have to stay for the time being. As to the licensing, it looks like this should just be fine.

src/util/parallel/threadpool.no-domainslib.ml
```diff
@@ -0,0 +1,17 @@
+module Thread_pool =
+struct
+
+  let run pool f = f ()
+
+  let create n = ()
+
+  let add_work pool f =
+    let promise = Domain.spawn f in
```
sim642, line 9: The dummy implementation is supposed to be OCaml 4 compatible, so it can't use `Domain` at all.

arkocal: Domain is provided by `domain_shims`, which is already included here. It acts as `Domain` on Ocaml5 and provides async parallelism on OCaml4.

src/util/timing/goblint_timing.ml
arkocal: Interesting, making them into refs did cause issues, so I will need to investigate this further.

src/util/parallel/threadpool.no-domainslib.ml
sim642: domain_shims implements `Domain` via `Thread` though, so it still allows interleaved execution (but not parallel execution). Even for that, output needs to be done under a real mutex, not a dummy one.
If the non-parallel mutex is a dummy, then the non-parallel thread pool probably should be even stupider: just run in the same thread entirely.

src/util/parallel/dune
```diff
@@ -0,0 +1,18 @@
+(include_subdirs no)
+
+(library
+ (name goblint_parallel)
+ (public_name goblint.parallel)
+ (libraries
+   batteries
```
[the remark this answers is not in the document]
arkocal: Not anymore, removed

src/util/parallel/dune
arkocal: Should be introduced now.

src/util/parallel/threadpool.no-domainslib.ml
arkocal: The disadvantage of using a real mutex in the Dummy would be, that we introduce that overhead in the single threaded case wherever we use the mutexes.

Since we do not plan to use this absence of true parallelism, running in the same thread entirely seems to be the better option. (We could also consider just throwing an error, but running in the same thread is OK IMO)

src/common/util/messages.ml
```diff
@@ -195,7 +195,7 @@ let print ?(ppf= !formatter) (m: Message.t) =
     | Debug -> "white" (* non-bright white is actually some gray *)
     | Success -> "green"
   in
-  let pp_prefix = Format.dprintf "@{<%s>[%a]%a@}" severity_stag Severity.pp m.severity Tags.pp m.tags in
+  let pp_prefix = (fun ppf -> Format.fprintf ppf "@{<%s>[%a]%a@}" severity_stag Severity.pp m.severity Tags.pp m.tags) in
```
sim642, line 198: There's an unnecessary diff here now.

src/goblint_lib.ml
```diff
@@ -465,6 +465,8 @@ module PrivPrecCompareUtil = PrivPrecCompareUtil
 module RelationPrecCompareUtil = RelationPrecCompareUtil
 module ApronPrecCompareUtil = ApronPrecCompareUtil
 
+(** {2 Debugging} *)
+
```
sim642, line 469: There's nothing here.

src/util/parallel/threadpool.mli
```diff
@@ -0,0 +1,11 @@
+module Thread_pool : sig
+  type t
+  type 'a promise
+
+  val create : int -> t
+  val run : t -> (unit -> 'a) -> 'a
+  val add_work : t -> (unit -> 'a) -> 'a promise
+  val await_all : t -> unit promise list -> unit
+  val finished_with : t -> unit
+end
```
sim642, line 10: The inner module seems redundant if everything is just inside it. Currently they would be `Threadpool.Thread_pool.create`, etc.

src/common/util/messages.ml
arkocal: Removed

src/goblint_lib.ml
arkocal: Also removed

src/util/parallel/threadpool.mli
arkocal: Flattened

michael-schwarz (approved):

src/util/parallel/domainsafeLazy.ml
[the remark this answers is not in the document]
sim642: @arkocal Just to confirm, did you check that the updated implementation actually works for the subsequent parallel solver stuff and doesn't break anything?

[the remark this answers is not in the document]
arkocal: Yes, I cannot see any additional tests failing due to this (we already had some tests failing as expected due to unsupported features)

src/util/timing/goblint_timing.ml
arkocal: When `enabled` is a regular ref, enabling timing results in `Fatal error: exception Stdlib.Stack.Empty` after printing the timings, so far I could not figure out why. Maybe you can give me some pointers at gobcon?

sim642: Oh, maybe `enabled` needs to be `Atomic` to avoid some race condition and out-of-date values between domains?

arkocal: With some help from @michael-schwarz we did figure out what went wrong.

The enter method requires a root frame, which is created only once at start. Since the stack is in DLS, every thread should create a root frame, which is now being handled in `split_from_parent` in DLS. 

The values recorded by the threads are not included in the final report, but should be correct.

dune-project
```diff
@@ -67,6 +67,8 @@ Goblint includes analyses for assertions, overflows, deadlocks, etc and can be e
     (conf-ruby :with-test)
     (benchmark :with-test) ; TODO: make this optional somehow, (optional) on bench executable doesn't work
     conf-gcc ; ensures opam-repository CI installs real gcc from homebrew on MacOS
+    domain-local-await
+    domain-shims
   )
   (depopts
```
sim642, line 73: I think `domainslib` should be under `depopts` since we're doing conditional compilation based on it.

arkocal: Makes sense, see 05a6869

sim642: The non-locked opam file also needs to be updated. But the locked one should stay as-is, otherwise Goblint won't install on OCaml 4 anymore.

arkocal: 57b2e21

sim642 (approved):

PR #1709, opened 2025-03-12
Use naïve unassume for Apron analysis when strengthening disabled

sim642:
Our Apron analysis is supposed to implement "strengthening-based dual-narrowing unassume".
However, we also have the option `ana.apron.strengthening` to disable the strengthening step in Apron `join` to avoid massive slowdown. The performance evaluation in the unassume paper and Goblint Validator at SV-COMP both disable it.

Today I discovered that this has surprisingly devastating consequences: disabling `ana.apron.strengthening` doesn't turn Apron analysis unassume into the naïve one, but in fact something much less precise. Specifically, it has the effect of additionally havocing all non-invariant variables. So strengthening is also needed to just keep any constraints on non-unassumed variables.

I don't yet know how bad the impact on Goblint Validator at SV-COMP is in practice. This might have really trashed our precision, although only the relational information (on tasks where the autotuner enables Apron analysis in the first place).

The fix is relatively simple though: when `ana.apron.strengthening` is disabled, then use naïve unassume in Apron analysis as well. This is achieved by just not restricting the unassumed state to the unassumed variables. In that case both `join` operands have full environment and are joined as normal: the join in homogeneous and the disabled strengthening is irrelevant.

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -740,7 +740,12 @@ struct
         in
         ({ f } : Queries.ask) in
       let rel = RD.assert_inv dummyask rel e false (no_overflow ask e_orig) in (* assume *)
-      let rel = RD.keep_vars rel (List.map RV.local vars) in (* restrict *)
+      let rel =
+        if GobConfig.get_bool "ana.apron.strengthening" then
+          RD.keep_vars rel (List.map RV.local vars) (* restrict *)
+        else
+          rel (* naive unassume: will be homogeneous join below *)
+      in
 
       (* TODO: parallel write_global? *)
       let st =
--- a/tests/regression/56-witness/48-apron-unassume-no-strengthening.yml
+++ b/tests/regression/56-witness/48-apron-unassume-no-strengthening.yml
@@ -1,0 +1,34 @@
+- entry_type: invariant_set
+  metadata:
+    format_version: "2.0"
+    uuid: 9b19e72a-376d-4c34-80d4-f46020de0c92
+    creation_time: 2025-03-12T12:53:06Z
+    producer:
+      name: Simmo Saan
+      version: n/a
+    task:
+      input_files:
+      - 48-apron-unassume-no-strengthening.c
+      input_file_hashes:
+        48-apron-unassume-no-strengthening.c: 57377f1e739e562d852b16206d5b1092e2188d3a4fe89bfca405a81bde0e1546
+      data_model: LP64
+      language: C
+  content:
+  - invariant:
+      type: location_invariant
+      location:
+        file_name: 48-apron-unassume-no-strengthening.c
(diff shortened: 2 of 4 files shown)


src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -740,7 +740,12 @@ struct
         in
         ({ f } : Queries.ask) in
       let rel = RD.assert_inv dummyask rel e false (no_overflow ask e_orig) in (* assume *)
-      let rel = RD.keep_vars rel (List.map RV.local vars) in (* restrict *)
+      let rel =
+        if GobConfig.get_bool "ana.apron.strengthening" then
```
sim642, line 744: I just realized we have a bit of an inconsistency with how our relation and Apron stuff are split. This analysis is also applied for our OCaml-implemented relational domains, but only `ApronDomain` implements the strengthening logic.
Given everything else (working with Apron variables and expressions), it might be possible to move the strengthening implementation up a bit such that it also works for others. Not that this would make a notable practical difference though.

michael-schwarz: It might, once @DrMichaelPetter's additional relational domains make a debut....

tests/regression/56-witness/48-apron-unassume-no-strengthening.yml
```diff
@@ -0,0 +1,34 @@
+- entry_type: invariant_set
+  metadata:
+    format_version: "2.0"
+    uuid: 9b19e72a-376d-4c34-80d4-f46020de0c92
+    creation_time: 2025-03-12T12:53:06Z
+    producer:
+      name: Simmo Saan
+      version: n/a
```
michael-schwarz, line 8: I still think you should provide versions of yourself, e.g., to prepare for the singularity. :wink:

michael-schwarz (approved):

src/analyses/apron/relationAnalysis.apron.ml
michael-schwarz: Maybe this can be extracted into an issue?

sim642:
> I don't yet know how bad the impact on Goblint Validator at SV-COMP is in practice. This might have really trashed our precision, although only the relational information (on tasks where the autotuner enables Apron analysis in the first place).

I did sv-benchmarks runs validating our own witnesses with 60s time limit: https://goblint.cs.ut.ee/results/249-all-validate-pr-1709-after/table-generator-cmp.diff.html#/table?filter=0(0*status*(status(in(ERROR%20%28witness%20error%29,EXCEPTION%20%28Stack%20overflow%29,unknown)),category(notIn()))),1(0*status*(status(in(ERROR%20%28witness%20error%29,true,unknown)),category(notIn()))).
There's 33 validations going from unknown to true, so there's a difference but luckily this hasn't cost us much.

PR #1714, opened 2025-03-19
Fix test 82 04

arkocal:
The old test had no widening at all, since there was only one side effect per variable.

    Also changed so that one variable has more side effects than gas.

### DIFF
--- a/tests/regression/82-widening_gas/04-side_simple_update.c
+++ b/tests/regression/82-widening_gas/04-side_simple_update.c
@@ -18,11 +18,10 @@ void *thread(void *arg) {
   pthread_mutex_lock(&A);
   b = 2;
   c = 2;
-  c = 3;
   pthread_mutex_unlock(&A);
 
   pthread_mutex_lock(&A);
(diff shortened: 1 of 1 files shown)


tests/regression/82-widening_gas/04-side_simple_update.c
```diff
@@ -28,7 +34,7 @@ int main(void) {
   pthread_mutex_lock(&A);
   __goblint_check(a <= 1);
   __goblint_check(b <= 2);
-  __goblint_check(c <= 3);
+  __goblint_check(c <= 4); // UNKNOWN!
```
michael-schwarz, line 37: The meaning of `//UNKNOWN!` in our regression suite is that the respective thing also does not hold in the concrete. I think the annotation to have here would be `//TODO`. Also, why 4?

michael-schwarz (changes requested):
Good catch! Once my small comment is addressed, this is ready to go!

michael-schwarz (approved):

PR #1717, opened 2025-03-25
Use LazyEval instead of Lazy in C2PODomain.

jerhard:
This PR fixes the [CI failures for marshalling tests](https://github.com/goblint/analyzer/actions/runs/14050445654/job/39339696261) that occur with C2PO in the ulocked pipeline.

The failure of the marshalling tests for C2PO is due to the normal form that appears as part of the domain of C2PO, being `Lazy.t`. Thus, these normal forms are of a functional value that cannot be marshaled. 

This PR changes this by using `LazyEval` in place of `Lazy`. Then, the input data for the computation of the normal form --- the other record entries of the domain --- are kept instead of a closure capturing these values. The `eval` that is defined then performs the evaluation of the normal form.

### DIFF
--- a/src/cdomains/congruenceClosure.ml
+++ b/src/cdomains/congruenceClosure.ml
@@ -599,30 +599,16 @@ module MRMap = struct
     List.fold_left add_element empty (SSet.elements set)
 end
 
-(**
-   This is the type for the abstract domain.
-
-   - `uf` is the union find tree
-
-   - `set` is the set of terms that are currently considered.
-     It is the set of terms that have a mapping in the `uf` tree.
-
-   - `map` maps reference variables v to a map that maps integers z to terms that are equivalent to *(v + z).
-     It represents the transitions of the quantitative finite automata.
-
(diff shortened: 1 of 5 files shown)


michael-schwarz (approved):

src/cdomains/congruenceClosure.ml
```diff
@@ -788,12 +813,13 @@ let split conj =
...
-   normal_form = lazy([]);
-   diseq = Disequalities.empty;
-   bldis = BlDis.empty}
+  let data = {uf = TUF.empty;
+              set = SSet.empty;
+              map = LMap.empty;
+              diseq = Disequalities.empty;
+              bldis = BlDis.empty}
+  in
+  data
```
sim642, line 822: The `let` is excessive.

sim642 (approved):

src/cdomains/congruenceClosure.ml
jerhard: Now the excessive `let` is avoided.

PR #1720, opened 2025-03-26
`IsMultiple` in `MallocWrapper`: Check uniqueness of allocating(!) thread

### LINKED ISSUE #1719: Goblint cannot handle even simple cases of malloc'ed mutexes
Looking at some tasks @vesalvojdani's CoOperRace could do but Goblint can't, I realized Goblint can't handle dynamically allocated mutexes even in the most simple of cases:

~~~C
#include <pthread.h>
pthread_mutex_t *data1Lock;
int g;

void *funcA(void *param) {
    pthread_mutex_lock(data1Lock);
    g = 1; //NORACE
    pthread_mutex_unlock(data1Lock);
}

int main(int argc, char *argv[]) {
    data1Lock = (pthread_mutex_t *) malloc(sizeof(pthread_mutex_t));
    pthread_mutex_init(data1Lock, ((void *)0));

    pthread_t one, two;

    pthread_create(&one, ((void *)0), &funcA, ((void *)0));
    pthread_create(&two, ((void *)0), &funcA, ((void *)0));
}
~~~

Here, we even do the initialization before even going multi-threaded, and all allocs happen from the main thread, outside any loops.
One may even say it's a bit embarrassing that this trivial program is enough to defeat Goblint.

michael-schwarz:
To determine uniqueness of a malloced object, Goblint is supposed to perform the following logic:

- A: Check that is assigned a counter value (numbering mallocs within a thread) other than `top`
- B: Check that the allocating thread is unique

Unfortunately, the logic for (B) is broken: It checks whether the thread **making a uniqueness query** for the blob is unique, instead of the thread which **allocated** the memory in the first place.

The added testcase is a precision issue, but it also is a soundness problem. 

If the querying thread happens to be unique but the allocating one is not, we give the wrong answer here.  Because of various internals (always spawning the unique copy of a thread in a loop too), getting Goblint to actually surface the unsoundness was beyond me, though.

SV-COMP: After this blunder is removed, we might benefit in MemSafety and it may also be interesting to re-check if one should enable the malloc uniqueness for races (CC @karoliineh): With ` --set ana.malloc.unique_address_count 1`, we now succeed on `sv-comp/sv-benchmarks/c/pthread/twostage_3.i`.

Closes #1719

### DIFF
--- a/src/analyses/wrapperFunctionAnalysis.ml
+++ b/src/analyses/wrapperFunctionAnalysis.ml
@@ -171,7 +171,10 @@ module MallocWrapper : MCPSpec = struct
       NodeVarinfoMap.mem_varinfo v
     | Q.IsMultiple v ->
       begin match NodeVarinfoMap.from_varinfo v with
-        | Some (_, _, c) -> UniqueCount.is_top c || not (man.ask Q.MustBeUniqueThread)
+        | Some (t, _, c) -> UniqueCount.is_top c ||
+                            (match t with
+                             | `Lifted tid -> not (Thread.is_unique tid)
+                             | _ -> true)
         | None -> false
       end
     | _ -> Queries.Result.top q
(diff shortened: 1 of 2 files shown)


sim642 (approved):
I have _deja vu_ that we've had a similar issue somewhere in the past, but couldn't easily find it.

PR #1721, opened 2025-03-28
Fallback for malloc uniqueness for when threading analyses are disabled

michael-schwarz:
Follow-up to #1719  and #1720.


michael-schwarz:
The SV-COMP runs shows that this mitigates the losses of #1720. We gain all lost tasks back.

sim642 (approved):

PR #1726, opened 2025-04-15
Document def_exc and enums exclusion bit range assumption

sim642:


src/cdomain/value/cdomains/int/enumsDomain.ml
```diff
@@ -11,7 +11,10 @@ module Enums : S with type int_t = Z.t = struct
   let range_ikind = Cil.IInt
   let size t = R.of_interval range_ikind (let a,b = Size.bits_i64 t in Int64.neg a,b)
 
-  type t = Inc of BISet.t | Exc of BISet.t * R.t [@@deriving eq, ord, hash] (* inclusion/exclusion set *)
+  type t =
+    | Inc of BISet.t (* Inclusion set. *)
+    | Exc of BISet.t * R.t (** Exclusion set. Bit range corresponds to {e some} {!ikind}, so always includes 0. *)
```
michael-schwarz, line 16: Is this guaranteed already? I remember seeing some  bounds corresponding to the meet of `ikind`s in SV-COMP at some point such as `[0,31]`. But maybe this is fixed?

michael-schwarz: Same for DefExc.

src/cdomain/value/cdomains/intDomain0.ml
```diff
@@ -464,6 +464,7 @@ module Size = struct (* size in bits as int, range as int64 *)
       if M.tracing then M.tracel "cast" "Cast %a to range [%a, %a] (%a) = %a (%s in int64)" GobZ.pretty x GobZ.pretty a GobZ.pretty b GobZ.pretty c GobZ.pretty y (if is_int64_big_int y then "fits" else "does not fit");
       y
 
+  (** @return Bit range corresponds to {e some} {!ikind}, so always includes 0. *)
```
michael-schwarz, line 467: Does this hold in the case of  Unsigned? It seems we try to fit it into a smaller type here: `a, upper_bound_less`?

sim642: Ugh, yes, it's not even true here, so I'll remove it.

This does point to bunch of old stuff though. Namely PR #100 which introduced this special behavior to fix some non-associativity problems.
The oddity is how the range doesn't depend on the ikind being used, but the smallest ikind that would fit the value and match the sign of the value. So even with `int` (with bit range [-31,31]) value 2e9, this otherwise would return [0,32] (for `unsigned int`). And the special casing turns it into [0,31], which actually is a weaker form of #1723 as well.
My hypothesis is that #1723 should also solve the old non-associativity issue in a less hacky way, but that's beside the point here.

michael-schwarz (approved):

PR #1727, opened 2025-04-15
Remove dummy `top` and `is_top` implementations from int domains

sim642:
While doing #1726 I noticed this beauty:
https://github.com/goblint/analyzer/blob/18489f370d2902350ee57de766f59af3b11e07d7/src/cdomain/value/cdomains/int/defExcDomain.ml#L84
Although it's supposed to be harmless.

Nevertheless, int domains have dummy `top` and `is_top` implementations: in many cases even just `failwith`s because they aren't actually needed. Instead, we only ever use `top_of` and `is_top_of`, as evidenced by the fact that we never crash with these dummy implementations.

Therefore, this is a slight cleanup of the (rather messy) int domain signature to not require `top` and `is_top` at all.

**EDIT:** Now also fixes unsoundness of def_exc for `__int128`.

### DIFF
--- a/src/cdomain/value/cdomains/int/defExcDomain.ml
+++ b/src/cdomain/value/cdomains/int/defExcDomain.ml
@@ -81,8 +81,8 @@ struct
   let name () = "def_exc"
 
 
-  let top_range = R.of_interval range_ikind (-999L, 999L) (* Since there is no top ikind we use a range that includes both IInt128 [-127,127] and IUInt128 [0,128]. Only needed for intermediate range computation on longs. Correct range is set by cast. *)
-  let top_overflow () = `Excluded (S.empty (), top_range)
+  let overflow_range = R.of_interval range_ikind (-999L, 999L) (* Since there is no top ikind we use a range that includes both IInt128 [-127,127] and IUInt128 [0,128]. Only needed for intermediate range computation on longs. Correct range is set by cast. *)
+  let top_overflow () = `Excluded (S.empty (), overflow_range)
   let bot () = `Bot
   let top_of ik = `Excluded (S.empty (), size ik)
   let bot_of ik = bot ()
--- a/src/cdomain/value/cdomains/int/defExcDomain.ml
+++ b/src/cdomain/value/cdomains/int/defExcDomain.ml
@@ -80,9 +80,6 @@ struct
   type int_t = Z.t
   let name () = "def_exc"
 
-
-  let top_range = R.of_interval range_ikind (-99L, 99L) (* Since there is no top ikind we use a range that includes both ILongLong [-63,63] and IULongLong [0,64]. Only needed for intermediate range computation on longs. Correct range is set by cast. *)
-  let top () = `Excluded (S.empty (), top_range)
   let bot () = `Bot
   let top_of ik = `Excluded (S.empty (), size ik)
   let bot_of ik = bot ()
@@ -116,8 +113,6 @@ struct
     else
       let upperb = Exclusion.max_of_range r in
       Z.compare i upperb <= 0
-
-  let is_top x = x = top ()
 
   let equal_to i = function
     | `Bot -> failwith "unsupported: equal_to with bottom"
@@ -351,10 +346,10 @@ struct
       (* We don't bother with exclusion sets: *)
       | `Excluded _, `Definite _
       | `Definite _, `Excluded _
-      | `Excluded _, `Excluded _ -> top ()
+      | `Excluded _, `Excluded _ -> top_of ik
       (* The good case: *)
       | `Definite x, `Definite y ->
-        (try `Definite (f x y) with | Division_by_zero -> top ())
+        (try `Definite (f x y) with | Division_by_zero -> top_of ik)
       | `Bot, `Bot -> `Bot
       | _ ->
         (* If only one of them is bottom, we raise an exception that eval_rv will catch *)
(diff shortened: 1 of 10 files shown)


src/cdomain/value/cdomains/int/defExcDomain.ml
michael-schwarz, line 1: It seems that `norm` relies on things being outside the range for some overflow-handling. Did you investigate whether returning something with the correct range has adverse effects in that some code that should be executed in here isn't any more?

This might be especially bad if we later try to detect overflows here...

src/cdomain/value/cdomains/intDomain0.ml
```diff
@@ -319,7 +321,7 @@ struct
   let is_bot x = I.is_bot x.v
   let top_of ikind = { v = I.top_of ikind; ikind}
   let top () = failwith "top () is not implemented for IntDomLifter."
-  let is_top x = I.is_top x.v
+  let is_top _ = false
```
michael-schwarz, line 324: Maybe this should raise now directly (as the inner implementations used to do)? It doesn't do what one would expect it to do, and any usage is bad.

src/cdomain/value/cdomains/int/defExcDomain.ml
sim642: Nothing about `norm` is specific to `top` or the [-99,99] range: it just restricts the result of some operation to the ikind's bit range and operations are still free to produce any range they want.
In fact, for most of the old `top ()` usages in operations, it's probably better for the potential overflow checking to not even use this definitely overflowing [-99,99] range, but the one for `ik` instead: all these `eq`/`ne`/`logand`, etc things can't even cause overflows. Them using `top ()` instead of `top_of ik` seems like a historical artifact.

Now looking at it a bit closer, I see why `IntDomTuple`'s overflow logic doesn't include def_exc: `DefExcDomain` implements the older-style signature where operations don't even return `overflow_info` (and `norm` doesn't do anything with overflows other than manipulating the range in the returned value).
So to incorporate def_exc for that, a lot needs to be updated to send out the (non-)overflow information for `IntDomTuple` to even use.

Although, looks like such `top_range` thing would still be (at least intermediately) needed to have `add` report overflow on two top values for the ikind: returning `top_of ik` wouldn't trigger overflow logic in `norm`.
So I'll keep a version of it for now to make the potential migration easier by keeping `top_overflow ()`.

Actually, the [-99,99] range is possibly broken as well because we've added support for 128bit integers, for which this range isn't even big enough!
I'll see if I can trigger some unsoundness and bump the `top_range` to be bigger. Although in the long run, with overflow handling in def_exc, it should become unnecessary as all operations should compute the resulting range themselves.

src/cdomain/value/cdomains/int/defExcDomain.ml
```diff
@@ -382,11 +380,11 @@ struct
...
   let eq ik x y = match x,y with
     (* Not much to do with two exclusion sets: *)
-    | `Excluded _, `Excluded _ -> top ()
+    | `Excluded _, `Excluded _ -> top_of ik
     (* Is x equal to an exclusion set, if it is a member then NO otherwise we
      * don't know: *)
-    | `Definite x, `Excluded (s,r) -> if S.mem x s then of_bool IInt false else top ()
-    | `Excluded (s,r), `Definite x -> if S.mem x s then of_bool IInt false else top ()
+    | `Definite x, `Excluded (s,r) -> if S.mem x s then of_bool IInt false else top_of ik
+    | `Excluded (s,r), `Definite x -> if S.mem x s then of_bool IInt false else top_of ik
```
michael-schwarz, line 387: Should this not be `top_of Iint` then? Just for symmetry with the other case?

sim642: That sounds right, so I've changed it in `eq` and `ne`.

Although these things look like a mess: other comparisons (e.g. `lt`) go as far as returning `top_bool`, regardless of `ik`.

src/domain/lattice.ml
```diff
@@ -27,16 +27,28 @@ sig
...
   val bot: unit -> t
   val is_bot: t -> bool
+end
+
+module type Top =
+sig
+  type t
   val top: unit -> t
   val is_top: t -> bool
 end
```
michael-schwarz, line 42: Could we go for `BoundedMeetSemiLattice` and ` BoundedJoinSemiLattice` here as meaningful mathematical objects? And use this opportunity to actually make our PO a PartialOrder and not some sort of lattice in disguise, i.e., remove `join`, `meet`, `narrow` and `widen` from it?

sim642: I could try that in #1728 to also weaken arguments of `LiftBot` and `LiftTop`.

Splitting the signatures into precise algebraic structures is nice, but it comes with a cost that we already have with `Printable` and `Lattice`: to make meaningful use of them, one needs functors for each one, e.g. `Printable.Prod`, `Lattice.POProd`, `Lattice.BoundedMeetSemiLatticeProd`, `Lattice.BoundedJoinSemiLatticeProd`, `Lattice.Prod`, etc.
Of course all code duplication can be avoided by `include`, but they still need to be separate functors with separate names.

I've thought about splitting `Printable.S` in the past, because it is even more bloated, but it just leads to a bigger mess.

michael-schwarz: The question is whether all these lifters are used then. If we don't have a `BoundedMeetSemiLatticeProd` anywhere, no need to provide the functor for it.

michael-schwarz: It's not completely orthogonal: I think introducing ill-defined module signatures such as `Top` (there is a special element intended to be greater than all others and a way to check whether some element is identical to it, but no way to compare elements) only makes it harder to understand what's going on.

michael-schwarz: I get what you mean, I just find the notion of having a `Top` signature that one can implement without even having a PO which would seem essential to being able to give such a thing non-helpful. But this is not a hill I am willing to die on!

michael-schwarz (approved):

Issue #1732, opened 2025-04-19
Disable GraphML witnesses for SV-COMP (and remove support?)


michael-schwarz:
In all categories where we produce witnesses, SV-COMP will remove support for GraphML-witnesses this year. We should also disable them in our SV-COMP setting already now as constructing them is costly (c.f. #1322) and issues with these may hide other unsoundnesses or issues in the nightly rides.

While we need ARG for abstract debugging etc, I personally also don't see any reason for keeping support for producing GraphML files. Maybe we should remove them altogether?

michael-schwarz:
I have now disabled the support in the SV-COMP config (#1733). This leaves the question of whether we want to completely remove support.

sim642:
PR #1001 added among other things Graphviz export of ARGs, so the GraphML witness export could indeed be removed without losing access to any information. I cannot even remember when I last looked at a GraphML witness.

I hope to do it by SV-COMP 2026 time. This probably needs a bit of refactoring of things because the SV-COMP mode is currently somewhat coupled to ARGs (as in, for every property, e.g., no-data-race, we also return an ARG, but a trivial one), which now wouldn't ever be looked at.

PR #1733, opened 2025-04-19
Disable `graphml` witnesses in `svcomp.json`

michael-schwarz:
This is the part of #1732 that can be done immediately. I'll leave it to someone more familiar with the ARG which parts of the GraphML facility need to stay.

### DIFF
--- a/conf/svcomp.json
+++ b/conf/svcomp.json
@@ -89,11 +89,6 @@
     }
   },
   "witness": {
-    "graphml": {
-      "enabled": true,
-      "id": "enumerate",
-      "unknown": false
-    },
     "yaml": {
       "enabled": true,
       "format-version": "2.0",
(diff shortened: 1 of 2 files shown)


conf/svcomp.json
```diff
@@ -89,11 +89,6 @@
     }
   },
   "witness": {
-    "graphml": {
-      "enabled": true,
```
sim642, line 93: If removing this part entirely, instead of switching to `false`, then one might as well remove it from svcomp-validate.json as well.

michael-schwarz: is there any advantage to keeping it around and switching it to `false`?

sim642: Not really. I just had it explicit in svcomp-validate.json to make it more obvious in the diff of the two confs.

sim642 (approved):

michael-schwarz:
scatter

The impact is quite nice!

PR #1738, opened 2025-05-02
Remove GraphML witness export

### LINKED ISSUE #1732: Disable GraphML witnesses for SV-COMP (and remove support?)

sim642:
Closes #1732.


michael-schwarz:
~~~ocaml
|> lift arg_enabled (module WitnessConstraints.PathSensitive3)
~~~

Should this module be renamed? It's connection to witnesses is only historical at this point.

michael-schwarz (approved):
Other than my comments above, this looks good to go!

sim642:
> ```ocaml
> |> lift arg_enabled (module WitnessConstraints.PathSensitive3)
> ```
> 
> Should this module be renamed? It's connection to witnesses is only historical at this point. Maybe one can make an `ARG` folder? Then things such as `ArgTools` and `MyARG` are also removed from the witness folder.

I now moved them around and renamed this module.
I also moved ARG things around in the documentation, which I thought about at one point but then forgot, so thanks for the reminder.

michael-schwarz (approved):

PR #1739, opened 2025-05-05
Bit operation additions

H-Innos:
Additions for my BSc thesis "Improving Bitwise Operation Analysis in Goblint", supervised by Simmo Saan. This includes making the logand, logor, logxor, lognot, and shift_right abstractions more precise in the interval, interval sets, defexc, and enums domains. Additionally, I added support for the analysis of bit-fields.


src/analyses/tutorials/signs.ml
michael-schwarz, line 1: Perhaps the fixed tutorial should not be committed?

src/cdomain/value/cdomains/valueDomain.ml
```diff
@@ -1049,12 +1049,17 @@ struct
                 Top
           end
         | _ ->
-          let result =
+          let result =              (* TODO: siin toimub*)
```
michael-schwarz, line 1052: keegi ei tea ;)

sim642:
I fixed the merge conflicts with #1623 and #1727, bringing this up to date with master.

src/cdomain/value/cdomains/int/defExcDomain.ml
```diff
@@ -469,9 +493,44 @@ struct
...
+          `Excluded (S.empty (), R.of_interval range_ikind (Int64.of_int @@ -b, Int64.of_int b))
+      end
+    | `Excluded (_, r1), `Excluded (_, r2) -> `Excluded (S.empty (), R.join r1 r2)
+    (* The good case: *)
+    | `Definite x, `Definite y ->
+      (try `Definite (Z.logxor x y) with | Division_by_zero -> top_of ik)
+    | `Bot, `Bot -> `Bot
+    | _ ->
+      (* If only one of them is bottom, we raise an exception that eval_rv will catch *)
+      raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y))))
```
michael-schwarz, line 533: This looks a little like copy pasta :spaghetti:

src/cdomain/value/cdomains/int/enumsDomain.ml
```diff
@@ -209,10 +216,83 @@ module Enums : S with type int_t = Z.t = struct
...
+        else
+          match R.minimal r, R.maximal r with 
+          | None, _ | _, None -> R.top ()
+          | Some r1, Some r2 -> let b = Int.max (Z.numbits i) (Int64.to_int(Int64.max (Int64.abs r1) (Int64.abs r2))) in
+            R.of_interval range_ikind (Int64.of_int @@ -b, Int64.of_int b)
+      in 
+      let r' = BISet.fold (fun i acc -> R.join (f i) acc) x (R.bot ()) in 
+      Exc (BISet.empty (), r')
+    | Exc (_, r1), Exc (_, r2) -> Exc (BISet.empty (), R.join r1 r2)
+    | _ -> top_of ikind)
```
michael-schwarz, line 295: These three functions look very similar?

sim642:
There seems to be a bug: the regression test 83/07 refine-with-interval crashes. It looks like `&` for intervals computes something unsound there.

I also did [a 60s sv-benchmarks run](https://goblint.cs.ut.ee/results/251-all-pr-1739-after/table-generator-cmp.diff.html#/), which revealed 10 unsound unreach-call results:

- [ ] hardware-verification-bv/btor2c-lazyMod.adding.1.prop1-func-interl.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.adding.2.prop1-func-interl.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.adding.3.prop1-func-interl.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.adding.4.prop1-func-interl.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.adding.5.prop1-func-interl.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.adding.6.prop1-func-interl.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.sw_sym_ex_v.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.unsafe_analog_estimation_convergence.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.unsafe_intersymbol_analog_estimation_convergence.yml
- [ ] hardware-verification-bv/btor2c-lazyMod.vis_arrays_palu.yml

sim642:
I did a new 60s sv-benchmarks run with the latest fixes. All of the unsoundness is now fixed!
What's more, with these changes we can solve 26 new tasks: https://goblint.cs.ut.ee/results/252-all-pr-1739-after2/table-generator-cmp.diff.html#/table?filter=0(0*status*(status(in(unknown)),category(notIn()))),1(0*status*(status(notIn()),category(notIn()))),2(0*status*(status(in(true)),category(notIn())))&hidden=1

michael-schwarz:
I saw the commit message 

> now correctness proven for all

and got curious. What exactly are you proving here and how are you going about it? Are these manual proofs or some sort of mechanized approach? If yes, can we use it for our other domains too?

sim642:
> and got curious. What exactly are you proving here and how are you going about it? Are these manual proofs or some sort of mechanized approach? If yes, can we use it for our other domains too?

Since the thesis will be in Estonian, I can't say "wait a week". Various cases of each bitwise operation are formulated as Z3 (instead of OCaml code) queries over the bitvector SMT theory by bitblasting or whatever Z3 decides to do.
The point being that it's very difficult to reason about bits in two's complement representations of negative numbers and how those correspond to interval (or bit range) bounds. So it's an extra layer of confidence, but for manual reencoding of the OCaml implementation as Z3.

One probably could do this more widely, but the manual re-encoding is still error prone, so the guarantees aren't super strong.
In a broader context, it would be much more valuable to try to extract some domain implementations from Rocq or whatnot.

Nevertheless, I'm wondering if we should have these Z3 proof scripts added into this repository for the sake of it?

michael-schwarz:
Interesting!

> Nevertheless, I'm wondering if we should have these Z3 proof scripts added into this repository for the sake of it?

Not sure if we want to add it to the repo as it may go out of sync, but I'm definitely curious what they look like!

H-Innos:
> Not sure if we want to add it to the repo as it may go out of sync, but I'm definitely curious what they look like!

All the proofs are in this repo: https://github.com/H-Innos/Z3

tests/regression/98-bitwise-operations/06-bitfield.c
```diff
@@ -0,0 +1,34 @@
...
+    s.a = 1; // NOWARN
+    s.b = 1; // NOWARN
+
+    __goblint_check(s.a <= 7);
+    __goblint_check(s.a >= -4);
+    s.a = 8; // WARN
+    s.a = -5; // WARN
+
+    __goblint_check(s.b <= 7);
+    __goblint_check(s.b >= 0);
```
sim642, line 30: All of the `__goblint_check`-s should be before the assignments. Currently the checks are checking the assigned values 1, not the uninitialized top values.
So the checks aren't actually checking the right thing and would even pass using Goblint that doesn't support bitfields.

src/cdomain/value/cdomains/int/defExcDomain.ml
sim642: They were at the time, but now there seem to be improvements to various cases of `logxor` which makes them quite different.

src/cdomain/value/cdomains/int/enumsDomain.ml
sim642: They were at the time, but now there seem to be improvements to various cases similar to def_exc, so they are now quite different.

PR #1741, opened 2025-05-07
Deduplicate `IntDomain` module types

sim642:
This applies the `_intf` trick, which we also have in a few other places, to `IntDomain` where it's particularly useful because there are many large module types that currently are duplicated in `intDomain.mli` and `intDomain0.ml`.

### DIFF
--- a/src/cdomain/value/cdomains/intDomain0.ml
+++ b/src/cdomain/value/cdomains/intDomain0.ml
@@ -3,6 +3,8 @@ open GoblintCil
 open Pretty
 open PrecisionUtil
 
+include IntDomain_intf
+
 module M = Messages
 
 let (%) = Batteries.(%)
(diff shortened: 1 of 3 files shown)


src/cdomain/value/cdomains/intDomain0.ml
```diff
@@ -69,7 +70,7 @@ let should_wrap ik = not (Cil.isSigned ik) || get_string "sem.int.signed_overflo
   * Always false for unsigned types, true for signed types if 'sem.int.signed_overflow' is 'assume_none'  *)
 let should_ignore_overflow ik = Cil.isSigned ik && get_string "sem.int.signed_overflow" = "assume_none"
 
-type overflow_info = { overflow: bool; underflow: bool;}
+type overflow_info = IntDomain_intf.overflow_info = { overflow: bool; underflow: bool;}
```
michael-schwarz, line 73: Is this strictly necessary? Isn't `IntDomain_intf.overflow_info` a non-abstract type?

michael-schwarz (approved):

PR #1743, opened 2025-05-08
Add weak dependencies to TD3

heigotornik:
Added a new function to the right-hand side of an unknown - **demand**.

The new constraint system functional form definition is _val system : v -> ((v -> d) -> (v -> d -> unit) -> (v -> unit) -> d) m_, where the added  _(v -> unit)_ represents **demand**.
Thread spawning now uses `demandl (Function fd, c);` inplace of `ignore (getl (Function fd, c))`.

TD3 implements _demand_ as `demand_thr`  and adds a new hashmap `weak_dep`.
TD3 with weak dependencies can be run in two "modes" - eager and lazy.

**Eager**. 
Can be toggled with `--enable solvers.td3.wk_deps.eager`
1. If `demand` is called the demanded unknown `y` is added to the hashmap `weak_dep` and solved immediately. 
2. If at the end of the iteration any unknown in `weak_dep` or in `vs` is unstable, they are solved again until all unknowns in both are stable.

**Lazy**:
Default behavior currently.
Same as eager, but without solving immediately when calling `demand`.

Part of the bachelor's thesis "Weak Dependencies for Side-Effecting Constraint Systems".

### DIFF
--- a/src/config/options.schema.json
+++ b/src/config/options.schema.json
@@ -2417,6 +2417,19 @@
               "description": "Check TD3 data structure invariants",
               "type": "boolean",
               "default": false
+            },
+            "wk_deps": {
+              "title": "solvers.td3.wk_deps",
+              "type": "object",
+              "properties": {
+                "eager": {
+                  "title": "solvers.td3.wk_deps.eager",
+                  "description": "Should weak dependencies be solved immediately when first added to wk_deps",
+                  "type": "boolean",
+                  "default": false
+                }
+              },
+              "additionalProperties": false
             }
           },
           "additionalProperties": false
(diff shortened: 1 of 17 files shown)


michael-schwarz:
Thanks for opening the PR! can you please provide a summary of the proposed changes in the description  above?

michael-schwarz:
@arkocal: How does this interact with your create nodes?  If it's a longer discussion, we can maybe have it at Gobcon on Tuesday?

arkocal:
This is sufficiently similar to Felix implementation of create nodes, that it is most likely more reasonable to not keep both. (it would unnecessarily complicate the constraint systems). I believe it would be possible to use this or that going forward.

One major difference, which would maybe also be nice to have here, is that the unaffected solvers can be lifted to be ignorant to this change. 

I think it makes sense to discuss the exact semantics of both to make sure this is the case. After that it becomes a question of timing when to merge.

michael-schwarz:
I guess with #1747 open, this can now be closed?

Issue #1730, opened 2025-04-17
Audit for domain-unsafe global code


sim642:
Initial list from #1708 reviews, but could be extended over time:

- [ ] Top-level `lazy` values:
   - [ ] `Offset.Index.Exp.all`
   - [ ] `Cilfacade.any_index_exp`
   - [ ] `LibraryFunctions.intmax_t`
- [ ] Top-level `ref` values.
- [ ] `ResettableLazy` and its usages.
- [ ] `LazyEval` and its usages.
- [ ] Output streams:
   - [ ] `Logs`
   - [ ] `Messages`

arkocal:
In addition, we can consider marking features/analysis as domain-safe or note.

AFAIK, not all analyses consist only of pure functions, and it might be tricky to change this everywhere. In such cases, it should be marked and we should check for inconsistent configs at startup.

PR #1744, opened 2025-05-14
Remove some usages of `MyCFG.current_node`

michael-schwarz:
`current_node` is a bit problematic as it is global mutable state. For the concurrent solvers, it would be nice to get rid of such things: DLS is only a solution where one Domain maps to one uninterrupted task.

Also, in general, it's ugly to have such mutable global state.

In many cases the usage was actually not truly needed. In some cases it is a bit more complicated, these remain to be done here.

### DIFF
--- a/src/lifters/specLifters.ml
+++ b/src/lifters/specLifters.ml
@@ -445,34 +445,6 @@ struct
     | q -> query' man q
 end
 
-
-(** Limits the number of widenings per node. *)
-module LimitLifter (S:Spec) =
-struct
-  include (S : module type of S with module D := S.D and type marshal = S.marshal)
-
-  let name () = S.name ()^" limited"
-
-  let limit = ref 0
-
-  let init marshal =
-    limit := get_int "dbg.limit.widen";
-    S.init marshal
-
-  module H = MyCFG.NodeH
-  let h = H.create 13
-  let incr k =
-    H.modify_def 1 k (fun v ->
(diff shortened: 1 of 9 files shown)


src/lifters/specLifters.ml
```diff
@@ -445,34 +445,6 @@ struct
...
-        if v >= !limit then failwith (GobPretty.sprintf "LimitLifter: Reached limit (%d) for node %a" !limit Node.pretty_plain_short (Option.get !MyCFG.current_node));
-        v+1
-      ) h;
-  module D = struct
-    include S.D
-    let widen x y = Option.may incr !MyCFG.current_node; widen x y (* when is this None? *)
-  end
-end
-
-
```
michael-schwarz, line 475: It seems unclear how to keep it. Given it almost(?) never used, I think it's OK to remove it.

michael-schwarz:
Related #1730.

src/lifters/specLifters.ml
sim642: Does it even work (anymore)?
It seems to me that solvers should be modifying `MyCFG.current_node` before calling `widen` (which happens outside of transfer functions), for it to work at all.

michael-schwarz: I'm not sure. All the more reason to remove it!

sim642: Having `widen` be aware of the unknown somehow would be useful for #1219, but that probably requires a fresh implementation anyway and it's not worth keeping this unused weirdness around, even if it could work then.

sim642 (approved):

arkocal:
Thanks for beginning the effort! We should solve the issue in precisionUtil and messages as well, but I am not opening issues yet, since the topics are on next weeks GobCon agenda.

PR #1745, opened 2025-05-15
Refactor `goblint.constraint` dune library

arkocal:
This PR should implement what is already in #1743 , while also providing lifters so that not all solvers must be modified. It should also make sure this is in line with the work on `create` nodes for concurrent solvers.

While going through the complete file to make sure everything is done at the right place, I have realized that some complexity has grown here and it would make sense to remove some complexity before introducing some, so I will keep this running as a draft, but feel free to drop comments if you disagree with any of the changes or want to discuss anything.

Every commit should be more or less self contained.

### DIFF
--- a/src/framework/analyses.ml
+++ b/src/framework/analyses.ml
@@ -1,4 +1,4 @@
-(** {{!Spec} Analysis specification} and {{!MonSystem} constraint system} signatures. *)
+(** {{!Spec} Analysis specification} signatures. *)
 
 open GoblintCil
 open Pretty
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -1,4 +1,4 @@
-(** Construction of a {{!Analyses.MonSystem} constraint system} from an {{!Analyses.Spec} analysis specification} and {{!MyCFG.CfgBackward} CFGs}.
+(** Construction of a {{!Goblint_constraint} constraint system} from an {{!Analyses.Spec} analysis specification} and {{!MyCFG.CfgBackward} CFGs}.
     Transformatons of analysis specifications as functors. *)
 
 open Batteries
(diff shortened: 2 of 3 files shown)


src/constraint/constrSys.ml
```diff
@@ -71,6 +71,16 @@ sig
   val postmortem: LVar.t -> LVar.t list
 end
 
+(** A solver is something that can translate a system into a solution (hash-table) *)
```
arkocal, line 74: Move this up as this is the simplest introductory case.

src/constraint/constrSys.ml
```diff
@@ -31,7 +31,7 @@ type 'v sys_change_info = {
 }
 
 (** A side-effecting system. *)
-module type MonSystem =
+module type MonadicSystem =
```
arkocal, line 34: Renamed so that it is obvious from the name that this does not relate to monotonicity.

src/constraint/constrSys.ml
```diff
@@ -55,7 +55,7 @@ sig
 end
 
 (** Any system of side-effecting equations over lattices. *)
-module type EqConstrSys = MonSystem with type 'a m := 'a option
+module type EqConstrSys = MonadicSystem with type 'a m := 'a option
```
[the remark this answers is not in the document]
sim642: Looks like PR #325 removed the other variation where it used to be a list. Since this extra abstraction over a monad is no longer used for a while, I think it's fine to just inline the option.

arkocal:
I have unwrapped the module, as it was marked as TODO in the dune file. Now I see there are also other modules that have this. Should we still proceed with that. I think it with submodules it becomes a bit more important to not pollute the global namespace.

If so, what is the right way to go about the documentation here?

ps. Since most things defined here are types anyway, I did not create `.mli` files.

sim642:
> If so, what is the right way to go about the documentation here?

I did some fixes to the wrapped `goblint.constraint` documentation now.

michael-schwarz (approved):

sim642 (approved):

PR #1746, opened 2025-05-16
Introduce demand in constraint system

arkocal:
This implements #1743 following #1745 .

The focus is on the constraint systems here. Another PR should follow for renaming and possibly simplifying the lifters. When td3 is able to handle `demand`, some lifters can be removed.

### DIFF
--- a/src/framework/compareConstraints.ml
+++ b/src/framework/compareConstraints.ml
@@ -158,7 +158,7 @@ struct
   include PrecCompare.MakeHashtbl (Var) (Dom) (VH)
 end
 
-module CompareEqSys (Sys: EqConstrSys) (VH: Hashtbl.S with type key = Sys.Var.t) =
+module CompareEqSys (Sys: DemandEqConstrSys) (VH: Hashtbl.S with type key = Sys.Var.t) =
 struct
   module Compare = CompareHashtbl (Sys.Var) (Sys.Dom) (VH)
 
--- a/src/framework/compareConstraints.ml
+++ b/src/framework/compareConstraints.ml
@@ -158,7 +158,7 @@ struct
   include PrecCompare.MakeHashtbl (Var) (Dom) (VH)
 end
 
-module CompareEqSys (Sys: DemandEqConstrSys) (VH: Hashtbl.S with type key = Sys.Var.t) =
+module CompareEqSys (Sys: EqConstrSys) (VH: Hashtbl.S with type key = Sys.Var.t) =
 struct
   module Compare = CompareHashtbl (Sys.Var) (Sys.Dom) (VH)
 
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -19,9 +19,9 @@ end
 module FromSpec (S:Spec) (Cfg:CfgBackward) (I: Increment)
   : sig
     include DemandGlobConstrSys with module LVar = VarF (S.C)
-                           and module GVar = GVarF (S.V)
-                           and module D = S.D
-                           and module G = GVarG (S.G) (S.C)
+                                 and module GVar = GVarF (S.V)
+                                 and module D = S.D
+                                 and module G = GVarG (S.G) (S.C)
   end
 =
 struct
--- a/src/framework/constraints.ml
+++ b/src/framework/constraints.ml
@@ -15,10 +15,10 @@ sig
 end
 
 
-(** The main point of this file---generating a [GlobConstrSys] from a [Spec]. *)
+(** The main point of this file---generating a [DemandGlobConstrSys] from a [Spec]. *)
 module FromSpec (S:Spec) (Cfg:CfgBackward) (I: Increment)
   : sig
-    include GlobConstrSys with module LVar = VarF (S.C)
+    include DemandGlobConstrSys with module LVar = VarF (S.C)
                            and module GVar = GVarF (S.V)
                            and module D = S.D
                            and module G = GVarG (S.G) (S.C)
@@ -50,7 +50,7 @@ struct
     if !AnalysisState.postsolving then
       sideg (GVar.contexts f) (G.create_contexts (G.CSet.singleton c))
 
-  let common_man var edge prev_node pval (getl:lv -> ld) sidel getg sideg : (D.t, S.G.t, S.C.t, S.V.t) man * D.t list ref * (lval option * varinfo * exp list * D.t * bool) list ref =
+  let common_man var edge prev_node pval (getl:lv -> ld) sidel demandl getg sideg : (D.t, S.G.t, S.C.t, S.V.t) man * D.t list ref * (lval option * varinfo * exp list * D.t * bool) list ref =
     let r = ref [] in
     let spawns = ref [] in
     (* now watch this ... *)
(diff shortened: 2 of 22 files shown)


src/framework/compareConstraints.ml
```diff
@@ -158,7 +158,7 @@ struct
   include PrecCompare.MakeHashtbl (Var) (Dom) (VH)
 end
 
-module CompareEqSys (Sys: EqConstrSys) (VH: Hashtbl.S with type key = Sys.Var.t) =
+module CompareEqSys (Sys: DemandEqConstrSys) (VH: Hashtbl.S with type key = Sys.Var.t) =
```
sim642, line 161: I'm only now thinking about this, but it seems like perhaps here we could avoid the demand: the comparison of solutions shouldn't need to invoke the right-hand sides at all.

src/framework/constraints.ml
```diff
@@ -15,10 +15,10 @@ sig
...
 
-(** The main point of this file---generating a [GlobConstrSys] from a [Spec]. *)
+(** The main point of this file---generating a [DemandGlobConstrSys] from a [Spec]. *)
 module FromSpec (S:Spec) (Cfg:CfgBackward) (I: Increment)
   : sig
-    include GlobConstrSys with module LVar = VarF (S.C)
+    include DemandGlobConstrSys with module LVar = VarF (S.C)
                            and module GVar = GVarF (S.V)
                            and module D = S.D
                            and module G = GVarG (S.G) (S.C)
```
michael-schwarz, line 24: Maybe we want the `and` to still align with `with`. Alternatively, we should go for  a more standard indentation.

src/framework/constraints.ml
```diff
@@ -78,7 +78,7 @@ struct
           | fd ->
             let c = S.context man fd d in
             sidel (FunctionEntry fd, c) d;
-            ignore (getl (Function fd, c))
+            demandl (Function fd, c)
```
michael-schwarz, line 81: This raises an interesting point: Like this, demand is something that can only be triggered by the top-level in constraints and not by individual analyses: For this it would need to be exposed via a mechanism similar to `sideg`. But I guess this is not needed?

Other point: `spawn` sounds a lot like `create` and somewhat like `demand`. Are we happy with the nomenclature here?

src/framework/control.ml
```diff
@@ -507,7 +508,7 @@ struct
           | d1::d2::[] -> (* the directories of the runs *)
             if d1 = d2 then Logs.warn "Beware that you are comparing a run with itself! There should be no differences.";
             (* instead of rewriting Compare for EqConstrSys, just transform unmarshaled EqConstrSys solutions to GlobConstrSys soltuions *)
```
michael-schwarz, line 510:
```suggestion
            (* instead of rewriting Compare for EqConstrSys, just transform unmarshaled EqConstrSys solutions to GlobConstrSys solutions *)
```

while we're at it ;)

src/solver/effectWConEq.ml
```diff
@@ -87,4 +87,4 @@ module Make =
   end
 
 let _ =
-  Selector.add_solver ("effectWConEq", (module PostSolver.EqIncrSolverFromEqSolver (Make)));
+  Selector.add_solver ("effectWConDemandEq", (module PostSolver.DemandEqIncrSolverFromEqSolver (Make)));
```
michael-schwarz, line 90: Do we want to rename the solver here?

michael-schwarz:
`tests/unit/solver/solverTest.ml` probably also needs to be adapted, it fails to compile.

src/solver/effectWConEq.ml
arkocal: What exactly does this solver do?

Since this is user facing, the name should relate to what the solver actually does rather than its formal signature. As this one does not really do anything interesting with the demands, it could be confusing to rename.

src/framework/constraints.ml
arkocal: For some reason, `ocp-indent` wants to align it with the `i` of `with`. (maybe two spaces after the end of the name?) Went with that since I cannot commit anything else anyway.

src/solver/postSolver.ml
```diff
@@ -332,3 +333,18 @@ module EqIncrSolverFromEqSolver (Sol: GenericEqSolver): GenericEqIncrSolver =
       Post.post xs vs vh;
       (vh, ())
   end
+
+
+module DemandEqIncrSolverFromEqSolver (Sol: GenericEqSolver): DemandEqIncrSolver =
+  functor (Arg: IncrSolverArg) (S: DemandEqConstrSys) (VH: Hashtbl.S with type key = S.v) ->
+  struct
+    module Sol' = Sol (EqConstrSysFromDemandConstrSys (S)) (VH)
+    module Post = MakeList (ListArgFromStdArg (EqConstrSysFromDemandConstrSys (S)) (VH) (Arg))
```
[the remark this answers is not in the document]
arkocal: f954e214e

src/solver/effectWConEq.ml
arkocal: Oh, I had renamed that. Reverted :sweat_smile:

arkocal:
> `tests/unit/solver/solverTest.ml` probably also needs to be adapted, it fails to compile.

3e642fc5d

src/framework/control.ml
```diff
@@ -507,7 +508,7 @@ struct
...
-            let module Splitter = GlobConstrSolFromEqConstrSol (EQSys) (LHT) (GHT) in
+            let module Splitter = GlobConstrSolFromEqConstrSol (EQSys: DemandGlobConstrSys) (LHT) (GHT) in
```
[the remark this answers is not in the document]
arkocal: I am not sure, the solution itself does not really depend on the existence of demand (even though the functor does)

Also the translator `EqConstrSysFromGlobConstrSys` is actually from DemandGlobConstrSys, but we agreed we should rename and simplify those. (Maybe after td3 is adapted, maybe right after this PR)

michael-schwarz (approved):

src/framework/compareConstraints.ml
arkocal: Oh I misunderstood your point, I thought you wanted to factor it out completely. 

This is of course simple 600c417c

src/framework/constraints.ml
sim642: It was also aligned like this before. I think it aligns the `module`-s.

src/framework/constraints.ml
sim642: > This raises an interesting point: Like this, demand is something that can only be triggered by the top-level in constraints and not by individual analyses: For this it would need to be exposed via a mechanism similar to `sideg`. But I guess this is not needed?

As far as we're aware, we have no other uses for `demand` currently.
Since individual analyses cannot define unknowns with right-hand sides, they don't have anything useful to really `demand`.

> Other point: `spawn` sounds a lot like `create` and somewhat like `demand`. Are we happy with the nomenclature here?

`man.spawn` actually corresponds to the analyzed program spawning a new thread. I think that naming is still fine (and actually avoids confusing it with `demand`/`create` of the constraint systems).

sim642 (approved):

Issue #1758, opened 2025-06-18
Printf arguments are not evaluated


Robotechnic:
If found this while manipulating goblint.
When running goblint with `ana.arrayoob` set to true I found out that this code:
```c
#include <stdio.h>

int table[10] = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};

int main() {
    for (int i = 0; i < sizeof(table); i++) {
        table[i] = 0;
    }
    printf("\n");
    return 0;
}
```
Causes a warning to appear which is normal. But this code:
```c
#include <stdio.h>

int table[10] = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};

int main() {
    for (int i = 0; i < sizeof(table); i++) {
        printf("%d ", table[i]);
    }
    printf("\n");
    return 0;
}
```
Doesn't trigger any warning. Turns out that it does the same with other types of warning. For instance, with `ana.int.interval`:
```c
#include <stdio.h>
#include <limits.h>

int main() {
    int a = INT_MAX + 1;
    return 0;
}
```
Causes a warning, while
```c
#include <stdio.h>
#include <limits.h>

int main() {
    printf("%d\n", INT_MAX + 1);
    return 0;
}
```
don't

PR #1761, opened 2025-06-19
Fix for #1758

Robotechnic:
This fixes issue #1758. It might need some benchmarks, but it works and passes all the tests.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2477,6 +2477,7 @@ struct
             set ~man st lv_a lv_typ (VD.top_value (unrollType lv_typ))
         end
     in
+    List.map (eval_rv ~man st) args |> ignore;
     let st = match desc.special args, f.vname with
     | Memset { dest; ch; count; }, _ ->
       (* TODO: check count *)
--- a/tests/regression/26-undefined_behavior/02-array-out-of-bounds.c
+++ b/tests/regression/26-undefined_behavior/02-array-out-of-bounds.c
@@ -11,6 +11,7 @@ int main()
     for (int i = 0; i < 10; ++i)
     {
         arr[i] = 5; //WARN
+        printf("%d\n", arr[i]); //WARN
     }
     return 0;
 }


sim642:
At least on sv-benchmarks this has no measurable impact OLS regression on CPU time gives $$y = 0.98x + 0.01$$ without outliars. It's also the case that sv-benchmarks uses library functions a lot less than real-world programs would.

Still, we need to do this to be sound, there's no way around it. If this turns out to negatively impact some realistic benchmarks, we might want to optimize something, but that's a separate matter.

michael-schwarz (commented):
Thank you for submitting this PR, it's quite a surprise that this stayed hidden for such a long time!

Here's some remarks:

- There are some library functions that explicitly do not evaluate their arguments (or at least not with the usual semantics (such as the sizeof operator which iirc we map to a library function). Do we need to account for those?
- This adds such evaluation to the base analysis. Do other analyses also need it?
- Is `eval_rv` always the right thing to do? Do arguments to library functions not sometimes act as `lvals`?
- We may want to add a comment where we call `eval_rv` purely for its side-effects.
- In #1758, you have more examples. Why not add them as tests?

I would also be interested to see the effect of this on a larger set of benchmarks, say the concrat suite...

sim642:
To comment on a few of these:

> * There are some library functions that explicitly do not evaluate their arguments (or at least not with the usual semantics (such as the sizeof operator which iirc we map to a library function). Do we need to account for those?

`SizeOfE` is actually a constructor in `Cil.exp`, so that shouldn't be the problem.

> * This adds such evaluation to the base analysis. Do other analyses also need it?

Only if they produce warnings from domains during expression evaluation, but I don't know if that is the case anywhere else.
The other obvious candidate is relation analysis, but it doesn't even do structural evaluation.
The access/race analysis already covers this, e.g. 04-mutex/20-stdfun_rc.



> * Is `eval_rv` always the right thing to do? Do arguments to library functions not sometimes act as `lvals`?

To act as an lval, the argument needs to be `AddrOf lval`, not just `lval`. And anyway we have to evaluate things in lvals as well, e.g. pointer arithmetic under `*` or indexing expressions.

michael-schwarz (approved):

PR #1764, opened 2025-06-20
Division by zero messages

Robotechnic:
I added error and warning messages for division-by-zero errors.


tests/regression/01-cpa/77-divbyzero.c
sim642, line 1: In #1756 I added another test as 01/77, so this should be renamed to 01/78 before merging to avoid a (non-git) conflict.

src/analyses/base.ml
```diff
@@ -200,8 +200,24 @@ struct
...
     | Mult -> ID.mul
-    | Div -> ID.div
-    | Mod -> ID.rem
+    | Div -> 
+      fun x y ->
+        (match ID.equal_to Z.zero y with
+         | `Eq ->
+           M.error ~category:M.Category.Integer.div_by_zero ~tags:[CWE 369] "Second argument of division is zero"
+         | `Top ->
+           M.warn ~category:M.Category.Integer.div_by_zero "Second argument of division might be zero"
```
sim642: I think the warning cases should also have the same CWE tag.

PR #1768, opened 2025-06-25
Fix pointer arithmethic on NULL

### LINKED ISSUE #1762: May-NULL plus may-zero yields non-NULL pointer
This is a minimization of an issue found by @Robotechnic:
```c
int main(void) {
    int *p;
    int i;
    int *q = p + i;
    *q = 42; // TODO WARN (may NULL dereference)
}
```
We currently do not output a NULL dereference warning from base, even though `p` may be `NULL` and `i` may be 0.
It looks like our pointer arithmetic misses this possibility, because we get that `q → {?}`, which means non-`NULL` only, but no `NULL`.

sim642:
Closes #1762.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -200,7 +200,7 @@ struct
     | PlusA -> ID.add
     | MinusA -> ID.sub
     | Mult -> ID.mul
-    | Div -> 
+    | Div ->
       fun x y ->
         (match ID.equal_to Z.zero y with
          | `Eq ->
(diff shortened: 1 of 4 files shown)


src/analyses/base.ml
```diff
@@ -303,24 +303,28 @@ struct
...
-      | None -> default addr
+      match addr with
+      | Addr (x, o) -> AD.of_mval (x, addToOffset n (Some x.vtype) o)
+      | NullPtr ->
+        begin match ID.equal_to Z.zero n with
+          | `Eq -> AD.null_ptr
+          | `Neq -> AD.unknown_ptr
+          | `Top -> AD.top_ptr
+        end
+      | UnknownPtr
```
michael-schwarz, line 314: Can the unknown pointer not become a null pointer by pointer arithmetic?

sim642: Indeed, a similar case split might have to be here.

michael-schwarz (approved):

PR #1769, opened 2025-06-26
Chages in the error notation

Robotechnic:
When using the VS Code terminal, it is possible to click on error messages to go directly to the error range involved. Turns out the error format is slightly different than the one used by Goblint.
Goblint does: `file : line : column - line : column` while VS Code recognise by default `file : line . column - line . column`.
This PR fixes that. 
I tried with a default VS Code configuration without any extensions installed, and it works.

### DIFF
--- a/src/common/util/cilType.ml
+++ b/src/common/util/cilType.ml
@@ -37,14 +37,14 @@ struct
     (* TODO: add special output for locUnknown *)
     x.file ^ ":" ^ string_of_int x.line ^ (
       if x.column >= 0 then
-        ":" ^ string_of_int x.column
+        "." ^ string_of_int x.column
       else
         ""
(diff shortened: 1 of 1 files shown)


sim642:
Looks like this causes all of our tests to fail because we're probably matching the old format in some scripts.
Also, all of the cram tests will get touched by this.

sim642:
I'm still wondering where this notation comes from. I managed to find some kind of default regexes in VSCode here: https://github.com/microsoft/vscode/blob/fad39b690188aa2a5e9ebe87c3d51ee1afdf9dfd/src/vs/workbench/contrib/tasks/common/problemMatcher.ts#L1491-L1603. But none of them matches the format with the `.` instead of `:`.

PR #1772, opened 2025-07-01
Add assumption about array index reads

### LINKED ISSUE #1702: Scanning for 0 in an array flags code dead
Analyzing a real-world C project, we could extract the following minimized example:
```c
int sprnames[3] = {
    1,2,3
};

int main(int argc, char const *argv[])
{
    int* namelist;
    namelist = sprnames;
    while (*namelist != 0)
	    namelist++; 
    
    return 0;
}
```
Of course, the program over-confidently relies on the 4th cell of the array to be zero. This is afaik not guaranteed. The pattern, as brittle as it is seems to be a common one of programs of this era. Goblint flags this as dead after the loop:
```
[Warning][Deadcode] Function 'main' does not return
[Warning][Deadcode] Function 'main' has dead code:
  on line 13 (evil.c:13-13)
[Warning][Deadcode] Logical lines of code (LLoC) summary:
  live: 4
  dead: 1
  total lines: 5
[Warning][Deadcode][CWE-571] condition '*namelist != 0' (possibly inserted by CIL) is always true (evil.c:10:12-10:26)
```
Would we not rather want to consider the loop condition unknown in the end, and continue with the rest alive?

sim642:
This documents the assumption from https://github.com/goblint/analyzer/issues/1702#issuecomment-2697596994.

With this we could close #1702 as not planned.

### DIFF
--- a/docs/user-guide/assumptions.md
+++ b/docs/user-guide/assumptions.md
@@ -49,5 +49,25 @@ _NB! This list is likely incomplete._
 
     This affects the `base` analysis.
 
+    See [issue #1770](https://github.com/goblint/analyzer/issues/1770).
+
+4.  Array index reads are in bounds.
+
+    [C11's N1570][n1570] at J.2 states that
+
+    > The behavior is undefined in the following circumstances:
+    >
+    > - [...]
+    > - An array subscript is out of range, even if an object is apparently accessible with the given subscript (as in the lvalue expression `a[1][7]` given the declaration `int a[4][5]`) (6.5.6).
+
+    after a long list of undefined behaviors.
+
+    Goblint does not by default report reading from out of bounds array indices.
+    Goblint reports out of bounds array index accesses when `ana.arrayoob` option is enabled.
+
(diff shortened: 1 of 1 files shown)


docs/user-guide/assumptions.md
```diff
@@ -49,5 +49,25 @@ _NB! This list is likely incomplete._
...
+
+    > The behavior is undefined in the following circumstances:
+    >
+    > - [...]
+    > - An array subscript is out of range, even if an object is apparently accessible with the given subscript (as in the lvalue expression `a[1][7]` given the declaration `int a[4][5]`) (6.5.6).
+
+    after a long list of undefined behaviors.
+
+    Goblint does not by default report reading from out of bounds array indices.
+    Goblint reports out of bounds array index accesses when `ana.arrayoob` option is enabled.
```
michael-schwarz:
```suggestion
    Goblint reports out of bounds array index accesses when the `ana.arrayoob` option is enabled.
```

michael-schwarz (approved):

PR #1775, opened 2025-07-02
Relational & VarEq: Invalidate LHS on `Longjmpd` event

### LINKED ISSUE #1774: Relational analysis unsound for `setjmp/longjmp`
~~~C
// PARAM: --enable ana.int.interval --set ana.activated[+] apron
#include <stdio.h>
#include <stdlib.h>
#include <setjmp.h>
#include <goblint.h>

jmp_buf env_buffer;

int fun() {
   longjmp(env_buffer, 2);
}


int main () {
   int val;

   val = setjmp(env_buffer);
   if(val) {
      __goblint_check(1); // Reachable
   } else {
      fun();
   }



   __goblint_check(1); // Reachable
   return(0);
}
~~~

The two program points that are supposed to be reachable here are found to be dead when enabling the relational analysis. Without it, everything works.

(Discovered by @leunam99)

michael-schwarz:
This is necessary as (via the query system) information about the return value of `setjmp` upon first call is propagated into the relational analysis, despite handling `setjmp` as an unknown function.
A similar problem arises for `VarEq` as well, which is also fixed in this PR.

@leunam99: Can you check if this fixes the issue you saw?

Closes #1774.

### DIFF
--- a/src/analyses/varEq.ml
+++ b/src/analyses/varEq.ml
@@ -587,6 +587,11 @@ struct
           remove ask (Cil.var v) st
         in
         List.fold_left remove_var man.local (EscapeDomain.EscapedVars.elements vars)
+    | Events.Longjmped {lval} ->
+      BatOption.map_default (fun lv ->
+          let ask = Analyses.ask_of_man man in
+          man.local|> remove_reachable ~deep:false ask [mkAddrOf lv])
+        man.local lval
     | _ ->
       man.local
 end
--- a/src/analyses/varEq.ml
+++ b/src/analyses/varEq.ml
@@ -590,7 +590,7 @@ struct
     | Events.Longjmped {lval} ->
       BatOption.map_default (fun lv ->
           let ask = Analyses.ask_of_man man in
-          man.local|> remove_reachable ~deep:false ask [mkAddrOf lv])
+          remove ask lv man.local)
         man.local lval
     | _ ->
       man.local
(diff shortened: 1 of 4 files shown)


michael-schwarz:
Failing CI job is due to force-push.

src/analyses/varEq.ml
```diff
@@ -587,6 +587,11 @@ struct
           remove ask (Cil.var v) st
         in
         List.fold_left remove_var man.local (EscapeDomain.EscapedVars.elements vars)
+    | Events.Longjmped {lval} ->
+      BatOption.map_default (fun lv ->
+          let ask = Analyses.ask_of_man man in
+          man.local|> remove_reachable ~deep:false ask [mkAddrOf lv])
```
sim642: This is unnecessarily roundabout: taking the address of `lv` and then `remove_reachable ~deep:false` will query for the points-to-set, which will just be `lv` itself.
I think something like this should do the same thing, but I haven't (type)-checked:
```suggestion
          remove ask lv man.local
```

michael-schwarz: I'm not sure if this is true: Consider e.g. an lv al of the shape `*x`? Or am I missing something?

leunam99:
> @leunam99: Can you check if this fixes the issue you saw?

Yes, it does.

src/analyses/varEq.ml
sim642: The rest of the var_eq analysis currently uses `remove` on lvals elsewhere:
1. `assign` calls `add_eq`, which in all cases starts by `remove` of the assigned lval.
2. `combine_assign` calls `remove` directly on the assigned lval.

If that's wrong, then there's a separate and much bigger issue with the core functionality of the analysis (all it does is keep equalities from `assign`).

I think it should still be doing the right thing, because `remove` is based on `may_change` which does the `MayPointTo` queries anyway.

PR #1776, opened 2025-07-03
sparseOctagons: a new sparse implementation of Miné's Octagons

DrMichaelPetter:
This PR is for experimenting with a re-implementation of [Miné's Octagons](https://www-apr.lip6.fr/~mine/publi/article-mine-HOSC06.pdf) in Ocaml with a focus on sparsity. For now we are just interested, how this re-implementation does perform in comparison to the 'standard' apron implementation.


sim642:
Just a note in case we ever want to merge this: it might need adapting to the #2081.

PR #1777, opened 2025-07-03
Consider option that `malloc(0)` may return `NULL`

### LINKED ISSUE #1418: Unsoundness for `malloc(0)`
The C standard says: 

> If the size of the space requested is zero, the behavior is implementation defined: either a null pointer is returned, or the behavior is as if the size were some nonzero value, except that the returned pointer shall not be used to access an object

-- C99 draft, 7.20.3 (1)

This is independent from the `sem.malloc.fail` option. GCC does not make a choice here and leaves the decision up to the standard library.

~~~C
int main(void){
  int* ptr = malloc(0);

  if(ptr == 0) {
    // Reachable
    __goblint_check(1);
  }
}
~~~

We report:

> [Warning][Deadcode][CWE-570] condition '(unsigned long )ptr == (unsigned long )((int *)0)' is always false (tests/regression/11-heap/17-malloc-zero-bytes.c:6:6-6:14)

michael-schwarz:
- Adds an option to configure behavior when allocating 0 bytes in line with `C99 draft, 7.20.3 (1)`
- Replace bool to indicate whether allocation is on heap or stack with variant type
- Introduce more sharing in base analysis between allocating functions in special.

Closes #1418

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2503,11 +2503,21 @@ struct
         end
     in
     let alloc loc size =
+      (* Whether malloc(0) is assumed to return the null pointer, a valid pointer, or both cases need to be considered. *)
+      let malloc_zero_null, malloc_zero_pointer =
+        match get_string "sem.malloc.zero" with
+        | "null" -> true, false
+        | "pointer" -> false, true
+        | "either" -> true, true
+        | _ -> failwith "Invalid value for sem.malloc.zero."
+      in
       let bytes = eval_int ~man st size in
-      let is_zero = ID.equal_to Z.zero bytes in
+      let cmp_bytes_with_zero = ID.equal_to Z.zero bytes in
+      let bytes_may_be_zero = cmp_bytes_with_zero <> `Neq in
+      let bytes_may_be_nonzero = cmp_bytes_with_zero <> `Eq in
       let heap_var =
-        let include_null = get_bool "sem.malloc.fail" || (is_zero <> `Neq && get_string "sem.malloc.zero" <> "pointer") in
-        let include_pointer = (is_zero <> `Eq || get_string "sem.malloc.zero" <> "null") in
+        let include_null = get_bool "sem.malloc.fail" || (bytes_may_be_zero && malloc_zero_null) in
+        let include_pointer = bytes_may_be_nonzero || malloc_zero_pointer in
         let res = if include_pointer then AD.of_var (alloced_var loc man) else AD.bot () in
         if include_null then
           AD.join res AD.null_ptr
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2745,6 +2745,7 @@ struct
           st
       end
     | Alloca size, _ -> begin
+        (* The behavior for alloc(0) is implementation defined. Here, we rely on the options specified for the malloc also applying to alloca. *)
         match lv with
         | Some lv ->
           let heap_var = alloc Queries.AllocationLocation.Stack size in
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -143,8 +143,8 @@ struct
    * Initializing my variables
    **************************************************************************)
 
-  let heap_var on_stack man =
-    let info = match (man.ask (Q.AllocVar {on_stack})) with
+  let alloced_var location man =
+    let info = match (man.ask (Q.AllocVar location)) with
       | `Lifted vinfo -> vinfo
       | _ -> failwith("Ran without a malloc analysis.") in
     info
@@ -2501,9 +2501,27 @@ struct
           | _ ->
             set ~man st lv_a lv_typ (VD.top_value (unrollType lv_typ))
         end
-    in 
+    in
+    let alloc ?(offset=`NoOffset) loc size =
+      let bytes = eval_int ~man st size in
+      let is_zero = ID.equal_to Z.zero bytes in
+      let heap_var =
+        let include_null = get_bool "sem.malloc.fail" || (is_zero <> `Neq && get_string "sem.malloc.zero" <> "pointer") in
+        let include_pointer = (is_zero <> `Eq || get_string "sem.malloc.zero" <> "null") in
+        let res = if include_pointer then
+            AD.of_mval (alloced_var loc man, offset)
+          else
+            AD.bot ()
+        in
+        if include_null then
+          AD.join res AD.null_ptr
+        else
+          res
+      in
+      heap_var
+    in
     (* Evaluate each functions arguments. `eval_rv` is only called for its side effects, we ignore the result. *)
-    List.iter (fun arg -> eval_rv ~man st arg |> ignore) args; 
+    List.iter (fun arg -> eval_rv ~man st arg |> ignore) args;
     let st = match desc.special args, f.vname with
     | Memset { dest; ch; count; }, _ ->
       (* TODO: check count *)
@@ -2723,7 +2741,7 @@ struct
     | Alloca size, _ -> begin
         match lv with
         | Some lv ->
-          let heap_var = AD.of_var (heap_var true man) in
+          let heap_var = alloc Queries.AllocationLocation.Stack size in
           (* ignore @@ printf "alloca will allocate %a bytes\n" ID.pretty (eval_int ~man size); *)
           set_many ~man st [(heap_var, TVoid [], Blob (VD.bot (), eval_int ~man st size, ZeroInit.malloc));
                             (eval_lv ~man st lv, (Cilfacade.typeOfLval lv), Address heap_var)]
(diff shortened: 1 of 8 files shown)


src/domains/queries.ml
```diff
@@ -101,7 +105,7 @@ type _ t =
   | IterVars: itervar -> Unit.t t
   | PathQuery: int * 'a t -> 'a t (** Query only one path under witness lifter. *)
   | DYojson: FlatYojson.t t (** Get local state Yojson of one path under [PathQuery]. *)
-  | AllocVar: {on_stack: bool} -> VI.t t
+  | AllocVar: AllocationLocation.t -> VI.t t
```
sim642, line 108: Polymorphic comparison will be fine for this right now, but to future proof things this query should be explicitly handled in equal and hash at the end.

tests/regression/11-heap/19-malloc-zero.c
sim642, line 1: Maybe we could also test the false case, e.g. by checking for an out of bounds warning for no offset.

src/analyses/base.ml
```diff
@@ -2501,9 +2501,27 @@ struct
           | _ ->
             set ~man st lv_a lv_typ (VD.top_value (unrollType lv_typ))
         end
-    in 
+    in
+    let alloc ?(offset=`NoOffset) loc size =
```
sim642: The offset argument is never used. Calloc case adds it itself.

src/analyses/base.ml
```diff
@@ -2501,9 +2501,27 @@ struct
...
+      let bytes = eval_int ~man st size in
+      let is_zero = ID.equal_to Z.zero bytes in
+      let heap_var =
+        let include_null = get_bool "sem.malloc.fail" || (is_zero <> `Neq && get_string "sem.malloc.zero" <> "pointer") in
```
sim642: Here sem.malloc.fail overrides the new option, but I'm not sure it should do that. The standard doesn't describe the possibility for it to ever fail with zero argument.

src/analyses/base.ml
```diff
@@ -2501,9 +2501,27 @@ struct
...
+        let include_pointer = (is_zero <> `Eq || get_string "sem.malloc.zero" <> "null") in
```
sim642: All these negations make this very difficult to follow. Especially because both things have values where some are included in another.
Maybe less negations or pattern matching would make this easier to understand.

src/analyses/base.ml
michael-schwarz: The C99 draft says 7.20.3 (https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf)

> If the space cannot be allocated, a null pointer is returned. If the size of the space requested is zero, the behavior is implementation defined: either a null pointer is returned, or the behavior is as if the size were some nonzero value, except that the returned pointer shall not be used to access an object.

I'm not sure I agree that this states that allocating zero bytes always succeeds. It would seem the first sentence says that it can also fail?

tests/regression/11-heap/19-malloc-zero.c
michael-schwarz: I don't think I quite understand what you mean here?

src/analyses/base.ml
michael-schwarz: I tried to make more explicit what happens.

src/analyses/base.ml
sim642: The more flexible option for us would be to decouple the options for zero bytes still. Then one can configure it both ways:
1. If `malloc(0)` can fail, then set `sem.malloc.zero` to `either`.
2. If `malloc(0)` cannot fail, then set `sem.malloc.zero` to `pointer`. This still allows `sem.malloc.fail` to be enabled.

The latter interpretation isn't currently configurable.

Given that it's perhaps not clear from the standard it would make sense to allow both. That's the point of configuring the `sem` options: to define what assumptions to make about the implementation. If the two are coupled, then we're making an additional assumption about the implementation.

michael-schwarz: I have now separated the options as you suggested.

tests/regression/11-heap/19-malloc-zero.c
sim642: I mean adding something like
```c
else {
  *ptr = 1; // WARN
}
```
where the warning is supposed to be out of bounds access (because the size is 0). Maybe it needs memOutOfBounds activated as well?

michael-schwarz: I have added this now.

src/analyses/base.ml
```diff
@@ -2501,9 +2501,33 @@ struct
...
+        | "either" -> true, true
+        | _ -> failwith "Invalid value for sem.malloc.zero."
+      in
+      let bytes = eval_int ~man st size in
+      let cmp_bytes_with_zero = ID.equal_to Z.zero bytes in
+      let bytes_may_be_zero = cmp_bytes_with_zero <> `Neq in
+      let bytes_may_be_nonzero = cmp_bytes_with_zero <> `Eq in
+      let heap_var =
+        let include_null = (bytes_may_be_nonzero && get_bool "sem.malloc.fail") || (bytes_may_be_zero && malloc_zero_null) in
+        let include_pointer = bytes_may_be_nonzero || malloc_zero_pointer in
```
sim642: Shouldn't the second half also check `bytes_may_be_zero`?

src/analyses/base.ml
```diff
@@ -2723,7 +2747,7 @@ struct
     | Alloca size, _ -> begin
         match lv with
         | Some lv ->
-          let heap_var = AD.of_var (heap_var true man) in
+          let heap_var = alloc Queries.AllocationLocation.Stack size in
```
sim642: Since `alloca` isn't standard, it's behavior isn't too precisely described.
For example, whether it can fail for non-zero. The man page only says stack overflow is impossible to detect, so will it always appear to succeed?
Also the man page says nothing about zero argument. Seems like it could be a no-op and not bump the stack pointer at all. Surprisingly, a quick GCC Godbolt check reveals that `alloca(0)` compiles to a bit more than a no-op, but much less than `alloca(1)`.

The previous behavior here was to always succeed, so there will be a difference but I'm not sure how much it will matter. Maybe some sv-benchmarks use `alloca`.

Anyway, I think I'm fine with sharing the options for now, but this is something to keep in mind and maybe add a comment about.

src/analyses/base.ml
```diff
@@ -2732,11 +2756,7 @@ struct
     | Malloc size, _ -> begin
         match lv with
         | Some lv ->
-          let heap_var =
-            if (get_bool "sem.malloc.fail")
-            then AD.join (AD.of_var (heap_var false man)) AD.null_ptr
-            else AD.of_var (heap_var false man)
-          in
+          let heap_var = alloc Queries.AllocationLocation.Heap size in
```
sim642: I now realized something else about this allocation code: even in the case where this _only_ returns a `NULL` pointer, we put a blob into the local state.
But this blob isn't pointed at by anything, so it just clutters the local state.

Ideally we wouldn't have such an unreachable blob, but this can only happen under rare conditions, so it should barely matter. But it might also be worth commenting about.

src/analyses/base.ml
michael-schwarz: If `bytes_may_be_nonzero` is false, `bytes_may_be_zero` is always true.

src/analyses/base.ml
michael-schwarz: I don't think that's true after these changes though? The call to `alloced_var` is guarded by `include_pointer`.

sim642: I mean two lines below is
https://github.com/goblint/analyzer/blob/1014dcda13a8254dc1e01d51bc6a4a0f50b246e0/src/analyses/base.ml#L2761
which puts the blob into the local state and isn't conditional on anything.

sim642: Or hmm, I guess `heap_var` will just be `NullPtr` then and we're writing the blob into that. That indeed can't put such unreachable garbage into the local state.
But it's also a bit weird thing to do. Then we're writing into `NULL` but that isn't producing any warning?
We probably should push #1531 which makes that a bit more explicit, although still no warning I guess.

src/analyses/base.ml
michael-schwarz:
```
   As a special case, alloca(0) reclaims storage without
   allocating any.  It is a good idea to use alloca(0) in
   your main control loop, etc. to force garbage collection.
```

(from https://github.com/gcc-mirror/gcc/blob/a10723efcd611977c5e4a875b9cb1de920732d42/libiberty/alloca.c#L20-L22). So at least for GCC it does interesting things.

michael-schwarz: And in some cases, the null pointer is actually returned.

https://github.com/gcc-mirror/gcc/blob/a10723efcd611977c5e4a875b9cb1de920732d42/libiberty/alloca.c#L194-L196

michael-schwarz: I added a comment now.

src/analyses/base.ml
michael-schwarz: > Then we're writing into NULL but that isn't producing any warning?

I'm not sure if it is the `set` that should produce the warning here either way?

michael-schwarz: I have now refactored the code such that we do not explicitly write to null-pointers. This is maybe the smarter move to avoid creating such strange situations in the first place.

sim642 (approved):

PR #1784, opened 2025-07-17
Add semgrep rule for `BatEnum`

sim642:
Relates to issue #1437. PR #1443 didn't fix everything, possibly because new instances were introduced over the year.

Somehow the semgrep CI job passes, although locally `semgrep scan --config .semgrep/` does not. Maybe because this PR doesn't touch that code, so the semgrep bot cannot comment on it in this PR.

### DIFF
--- a/src/cdomains/congruenceClosure.ml
+++ b/src/cdomains/congruenceClosure.ml
@@ -2,7 +2,6 @@
     It is used by the C-2PO Analysis and based on the UnionFind implementation.
 *)
 include UnionFind
-open Batteries
 open GoblintCil
 open DuplicateVars
 module M = Messages
@@ -25,7 +24,7 @@ module BlDis = struct
   let empty = TMap.empty
   let is_empty = TMap.is_empty
 
-  let to_conj bldiseq = List.fold
+  let to_conj bldiseq = List.fold_left
       (fun list (t1, tset) ->
          TSet.fold (fun t2 bldiseqs -> BlNequal(t1, t2)::bldiseqs) tset [] @ list
       ) [] (bindings bldiseq)
@@ -52,11 +51,11 @@ module BlDis = struct
     let add_block_diseq_t1 bldiseq t2 =
       add_block_diseq bldiseq (t1, t2)
     in
-    List.fold add_block_diseq_t1 bldiseq tlist
+    List.fold_left add_block_diseq_t1 bldiseq tlist
 
   (** For each block disequality bl(t1) != bl(t2) we add all disequalities
       that follow from equalities. I.e., if t1 = z1 + t1' and t2 = z2 + t2',
-      then we add the disequaity  bl(t1') != bl(t2').
+      then we add the disequality  bl(t1') != bl(t2').
   *)
   let element_closure bldis cmap =
     let comp_closure = function
@@ -65,7 +64,7 @@ module BlDis = struct
         let eq_class2 = LMap.comp_t_cmap_repr cmap r2 in
         let terms1 = List.map snd eq_class1 in
         let terms2 = List.map snd eq_class2 in
-        List.cartesian_product terms1 terms2
+        BatList.cartesian_product terms1 terms2
       | _ -> []
     in
     List.concat_map comp_closure bldis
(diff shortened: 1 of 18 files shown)


src/config/options.ml
```diff
@@ -66,7 +66,7 @@
   | Boolean ->
     [("", ["false"; "true"])]
   | String string_specs ->
-    begin match element.enum with
+    begin match element.enum with  (* nosemgrep: batenum-enum *)
```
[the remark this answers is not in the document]
sim642: I'll make the semgrep rule more precise to not complain on these non-module `.enum` lookups.

src/cdomains/congruenceClosure.ml
```diff
@@ -127,15 +127,15 @@ module Disequalities = struct
...
+  let bindings x =
+    List.flatten @@
     List.concat_map
       (fun (t, smap) ->
          List.map (fun (z, tset) ->
              List.map (fun term ->
                  (t,z,term)) (TSet.elements tset))
            (ZMap.bindings smap)
-      ) % TMap.bindings
+      ) @@ TMap.bindings x
```
sim642: Given that we're already touching this code, this seems like a place where `Seq` should really be used instead of all the `List` business: this is basically a nesting of 3 `map`s and 2 `flatten`s (with `concat_map` being both).

src/cdomains/congruenceClosure.ml
```diff
@@ -269,7 +269,7 @@ module Disequalities = struct
       let l2 = map_find_all t2 arg in
       fold_left2 add_diseq acc l1 l2
     in
-    List.fold f acc (TSet.to_list tset)
+    List.fold_left f acc (TSet.elements tset)
```
sim642: Why is this (and maybe `check_neq` above) working on lists at all? It's creating a list from the set and then folding the list. Obviously this could be on `Seq`, but there's possibly an even more direct way: just `TSet.fold`.

src/cdomains/unionFind.ml
```diff
@@ -1055,8 +1056,8 @@ module LookupMap = struct
       [Z.zero, t]
     | Some zmap ->
       let offset_term_product (z, term_set) =
-        let terms = TSet.to_list term_set in
-        List.cartesian_product [z] terms
+        let terms = TSet.elements term_set in
+        BatList.cartesian_product [z] terms
```
sim642: `cartesian_product` with a singleton list? This could just be a normal `map`.

src/domain/hoareDomain.ml
```diff
@@ -46,7 +46,8 @@ struct
       else Map.add i b m
   end
 
-  let elements m = Map.values m |> List.of_enum |> List.flatten
+  let elements m =
+    m |> Map.to_seq |> Seq.map snd |> List.of_seq |> List.flatten
```
sim642, line 50: This could be wrong, but it seems like the flattening could be done on `Seq` even:
```suggestion
    m |> Map.to_seq |> Seq.concat_map snd |> List.of_seq
```

src/cdomains/congruenceClosure.ml
michael-schwarz: Done!

src/cdomains/congruenceClosure.ml
michael-schwarz: I changed it here. For `check_neq`, we're missing a `ZMap.fold2`, so it's not so clear how to do it. I'd suggest leaving it as-is.

src/domain/hoareDomain.ml
michael-schwarz: That has type `'a Seq.t Map.t -> 'a list` instead of `'a list Map.t -> 'a list`.

It would have to be

~~~OCaml
m |> Map.to_seq |> Seq.concat_map (fun x -> List.to_seq @@ snd x) |> List.of_seq
~~~

but round-tripping through Seq for no reason doesn't help here.

src/cdomains/unionFind.ml
michael-schwarz: I have now simplified it.

michael-schwarz:
@sim642: I have now implemented your suggestions, feel free to have another look!

PR #1785, opened 2025-07-17
Turn `malloc_null` into an analysis serving as path-sens hint only & Fix enter

### LINKED ISSUE #1691: `malloc_null`: Interprocedural analysis unsound
While fixing the bug that surfaced in #1688, I realized that the interprocedural handling of the `malloc_null` analysis is completely broken.
Even without any top pointers involved, callees always seem to get the empty set of addresses.

~~~C
// PARAM: --set ana.activated[+] malloc_null --set ana.path_sens[-] malloc_null
#include <stdlib.h>
#include <goblint.h>

void fun(int* a) {
        *a = 5; //WARN
}


int main(void) {
        int *x;

        x = malloc(sizeof(int));
        fun(x);
        *x =8; //WARN

        return 0;
}
~~~

In this example, we produce warnings at both program points. However, at the dereference of `a` only `base` warns

> [Warning][Behavior > Undefined > NullPointerDereference][CWE-476] May dereference NULL pointer (tests/regression/08-malloc_null/03-more.c:6:9-6:15)

while at the dereference of `x`, we also get a warning from `malloc_null`:

> [Warning][Behavior > Undefined > NullPointerDereference][CWE-476] May dereference NULL pointer (tests/regression/08-malloc_null/03-more.c:15:9-15:15)
> [Warning][Behavior > Undefined > NullPointerDereference] Possible dereferencing of null on variable 'x'. (tests/regression/08-malloc_null/03-more.c:15:9-15:15)

I traced the problem part of the way already:

https://github.com/goblint/analyzer/blob/746d436855efd9c1ef82f4507cbb76f7896eb20d/src/analyses/malloc_null.ml#L95-L118

Here, `vars` correctly contains `x`, but it is somehow filtered out in the last step even though `st` supposedly also contains it.

The reason we have not seen any issues with this likely is that base now seems to duplicate the warnings of `malloc_nulll` anyway, which I guess means `malloc_null` is only useful for the fact that it causes path splitting.

What do we want to do here? Do we repair it? Or remove the warning facilities from the analysis and trim it down to a helper analysis that causes additional path-splitting?

michael-schwarz:
- Removes redundant warning facilities from `malloc_null`
- Fixes interprocedural aspect:
   - If a function `f(int x)` is called as `f(y)`, nullness information should be transferred from  `y` in the callee state to `x` in the caller state
   - Adds a test where contexts are disabled, and one relies only on `malloc_null` 
   
Closes #1691.

### DIFF
--- a/src/analyses/malloc_null.ml
+++ b/src/analyses/malloc_null.ml
@@ -1,4 +1,4 @@
-(** Path-sensitive analysis of failed dynamic memory allocations ([malloc_null]). *)
+(** Helper analysis to be path-sensitive in failed dynamic memory allocations ([malloc_null]). *)
 
 module AD = ValueDomain.AD
 module IdxDom = ValueDomain.IndexDomain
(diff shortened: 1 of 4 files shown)


sim642:
I added a few more `NOWARN`s to the old tests and it revealed even more weirdness with the malloc_null analysis:
1. It warned on one `sizeof(*v)` although that doesn't actually evaluate the argument, it just gets the type, which is statically known. The base analysis uses `Cil.constFold` which takes care of it for us so the base analysis never even sees it as evaluation of `*v`.
2. If `sizeof(*v)` was done with `v` being uninitialized (maybe `NULL`, maybe not), then the same thing did not warn. So I think the analysis is weird and possibly wrong in other ways as well: because `v` is possibly `NULL` initially, malloc_null doesn't have it as such in its state and doesn't handle it soundly even. But this becomes irrelevant when it's only for path-sensitivity anyway: it just doesn't induce a path-split on the entry state immediately.

michael-schwarz:
`malloc_null` doesn't have any warning facilities after this PR, so the weirdness must come from the base analysis (maybe exposed by doing more path-splitting)?

sim642 (approved):

PR #1792, opened 2025-07-29
`BranchSet` analysis tracking set of taken branches

michael-schwarz:
Helper analysis to be path-sensitive in the set(!) of taken branches. This is a lightweight version of #1791 that has some hope of terminating.

As @sim642 remarked somewhere else, this is an instance of trace partitioning (or of ego-lane derived digests :wink:) and many other abstractions are also conceivable.

### DIFF
--- a/src/analyses/branchSet.ml
+++ b/src/analyses/branchSet.ml
@@ -7,7 +7,7 @@ module Spec =
 struct
   include Analyses.IdentitySpec
 
-  module Branch = Printable.ProdSimple(BoolDomain.Bool)(CilType.Location)
+  module Branch = Printable.ProdSimple(BoolDomain.Bool)(Node)
   module BranchSet = SetDomain.Make(Branch)
 
   module D = BranchSet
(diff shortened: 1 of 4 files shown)


src/goblint_lib.ml
```diff
@@ -131,6 +131,7 @@ module PthreadSignals = PthreadSignals
 module PthreadBarriers = PthreadBarriers
 module ExtractPthread = ExtractPthread
 module PthreadOnce = PthreadOnce
+module BranchSet = BranchSet
```
sim642: I think this is the wrong place for it. This section is for other concurrency analyses.
A better place would be where `Expsplit` also is.

src/analyses/branchSet.ml
```diff
@@ -0,0 +1,41 @@
...
+  let combine_env man lval fexp f args fc au f_ask =
+    au
+
+  let combine_assign man (lval:lval option) fexp (f:fundec) (args:exp list) fc (au:D.t) (f_ask: Queries.ask) : D.t =
+    man.local
+
+
+  let branch man (exp:exp) (tv:bool) : D.t =
+    let loc = Node.location man.node in
+    BranchSet.add (tv, loc) man.local
```
sim642: I think this should use the nodes themselves instead of their locations because due to CIL transformations there can be many branching nodes with the same location, but each one is a separate `branch` for the analysis.

DrMichaelPetter (commented):
LGTM

PR #1797, opened 2025-07-30
Replace eager `show` with lazy `pretty` during analysis

### LINKED ISSUE #1795: Avoid eager string conversions during analysis
> > But the other worrying point is that we're apparently printing something during solving itself. That's unusual, because usually it's just one solver stats printout after another. Perhaps there's some internal unnecessary string conversion taking place as well.
> 
> The analyses should not be using `Pretty` to convert things to `string`s during the analysis (unless `warn_at` is `early` at least). I may be wrong, but I can't currently think of a legitimate reason to do that.
> It might help to do an sv-benchmarks run with `-v` to get the backtrace at that point. I think the backtrace should still contain the entire non-signal-handler stack, so it would expose where are we using `Pretty` during analysis.
> 
> Of course avoiding that isn't a particularly reliable fix, but it would make sense to avoid it anyway for performance reasons.

 _Originally posted by @sim642 in [#1781](https://github.com/goblint/analyzer/issues/1781#issuecomment-3103979092)_

sim642:
Closes #1795.

This probably isn't all of it, but hopefully is most that I quickly found with some simple searches.


sim642:
I brought this up to date and I'm just merging this to get this over with. This just affects tracing output. And if I've accidentally made it worse somewhere, it can be fixed when someone actually needs to trace these relational analyses again.

PR #1800, opened 2025-08-01
Forward Constraint System

michael-schwarz:
This generates an equivalent forward propagating constraint system:

- Uses `ForwardCFG` with all updates happening via `sidel`
- Instantiates globals as spec globals plus with auxiliary globals for returns per fundec and context
- In `tf_proc` use `getg` to get end state of procedure
- In `tf_ret` side effect to appropriate helper global

It also presents Helmut's new forward solver.

### DIFF
--- a/lib/goblint/runtime/src/dune
+++ b/lib/goblint/runtime/src/dune
@@ -2,4 +2,5 @@
  (archive_name goblint)
  (language c)
  (names goblint)
+ (flags :standard -std=c11) ; workaround for https://github.com/goblint/analyzer/issues/1779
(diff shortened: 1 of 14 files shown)


michael-schwarz:
Might make sense to periodically merge master into this so it doesn't diverge too badly and the PR CI passes.

michael-schwarz:
@arkocal: I guess this is subsumed by `fwd_constrsys_new` and can be closed?

michael-schwarz:
Subsumed by a to-be-opened PR by @arkocal.

PR #1804, opened 2025-08-05
Fix interval division

### LINKED ISSUE #1803: Missing overflow warning on division by -1
`INT_MIN / -1` is the same as `-INT_MIN`, which is an overflow.

Goblint warns about it if the expression is constant, but is missing a warning when numerator may be `INT_MIN` and denominator may be -1:
```c
#include <limits.h>

int main() {
    int bad = INT_MIN / -1; // WARN
    int x, y;
    if (y != 0) {
        bad = x / y; // TODO WARN
    }
    return 0;
}
```

sim642:
Closes #1803.

### TODO
- [ ] Clean up.
- [ ] Confirm with interval set as well. (Or perhaps even split interval at 0).
- [ ] Find out why domain tests couldn't find unsoundness.

### DIFF
--- a/src/cdomain/value/cdomains/int/intervalDomain.ml
+++ b/src/cdomain/value/cdomains/int/intervalDomain.ml
@@ -294,13 +294,23 @@ struct
     | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
     | (Some (x1,x2) as x), (Some (y1,y2) as y) ->
       begin
-        let is_zero v = Ints_t.compare v Ints_t.zero = 0 in
+        let (r, ov) =
+          try
+            binary_op_with_norm IArith.div ik x y
+          with Division_by_zero ->
+            (top_of ik, {underflow=false; overflow=false})
+        in
+        if leq (of_int ik (Ints_t.zero) |> fst) (Some (y1,y2)) then
+          (top_of ik, ov)
+        else
+          (r, ov)
+        (* let is_zero v = Ints_t.compare v Ints_t.zero = 0 in
         match y1, y2 with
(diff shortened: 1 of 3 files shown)


src/cdomain/value/cdomains/int/intervalDomain.ml
```diff
@@ -292,16 +294,13 @@ struct
...
     | None, None -> (bot (),{underflow=false; overflow=false})
     | None, _ | _, None -> raise (ArithmeticOnIntegerBot (Printf.sprintf "%s op %s" (show x) (show y)))
-    | (Some (x1,x2) as x), (Some (y1,y2) as y) ->
-      begin
-        let is_zero v = Ints_t.compare v Ints_t.zero = 0 in
-        match y1, y2 with
-        | l, u when is_zero l && is_zero u -> (top_of ik,{underflow=false; overflow=false})
-        | l, _ when is_zero l              -> div ik (Some (x1,x2)) (Some (Ints_t.one,y2))
-        | _, u when is_zero u              -> div ik (Some (x1,x2)) (Some (y1, Ints_t.(neg one)))
-        | _ when leq (of_int ik (Ints_t.zero) |> fst) (Some (y1,y2)) -> (top_of ik,{underflow=false; overflow=false})
```
sim642, line 302: By the way, I'm not sure why, but the old handling of division by 0 is quite inconsistent (i.e. not monotone in the second argument):
* `x / [0, 0]` returns top (for division by 0).
* `x / [0, 42]` delegates to `x / [1, 42]`, which doesn't return top.
* `x / [-42, 0]` delegates to `x / [-42, -1]`, which doesn't return top.
* `x / [-42, 42]` again returns top (for division by 0).

The new handling consistently returns top in all of the above cases.
Maybe the goal was to imitate the kind of split at 0 which is now part of the general logic in `IArith.div`?

PR #1818, opened 2025-09-04
Fix some questionable invariants in SV-COMP 2025 witnesses

sim642:
This addresses instances from #1722.

### DIFF
--- a/src/analyses/varEq.ml
+++ b/src/analyses/varEq.ml
@@ -17,6 +17,10 @@ struct
   struct
     include PartitionDomain.ExpPartitions
 
+    let is_str_constant = function
+      | Const (CStr _ | CWStr _) -> true
+      | _ -> false
+
     let invariant ~scope ss =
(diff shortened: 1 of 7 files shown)


sim642:
I now did additional fast runs with Goblint Validator also included: https://goblint.cs.ut.ee/results/282-all-validate-pr-1818-after/table-generator-cmp2.table.html#/.

Overall, the number of invariants for the validator changed as follows:
|  | Confirmed | Unconfirmed | Refuted | Error |
|--------|--------|--------|--------|--------|
| Before | 875324 | 107439 | 5923 | 11330 |
| After | 835291 | 64189 | 662 | 9466 | 

The number of unconfirmed and refuted invariants decreased significantly. The decrease in confirmed invariants should not be problematic: if we no longer generate reflexive equalities, then we cannot trivially validate them either.
The number of validated witnesses actually increased from 9304 to 9427.

Issue #1832, opened 2025-09-28
Autotuner only enables relational analysis for at most 2 globals


sim642:
While checking https://github.com/goblint/analyzer/pull/1821#issuecomment-3342492386, I found out the following surprising thing: the autotuner only enables relational analysis for _at most 2_ globals:
https://github.com/goblint/analyzer/blob/529013137ad4de9d020b47dcfc7a553d1e677766/src/autoTune.ml#L435-L447

We should really look into this because we could be shooting ourselves in the foot with such a low limit. SV-COMP benchmarks often use global variables for no good reason (over locals) because code quality doesn't matter.
This also makes relational Mutex-Meet mostly useless.

sim642:
I tried just hard-coding it to 100 to see what happens, but just doing that loses points. It seems that by allowing more globals to be picked, the cost in autotuner goes up and the autotuner doesn't activate relational analysis at all.
So improving our results requires more careful tuning.

sim642:
Yeah, that's a good idea. I now did that and the diff table is here: https://goblint.cs.ut.ee/results/275-all-autotune-globals-100-2/table-generator-cmp.diff.html#/. The summary is:
1. 13 `true`s turned into `TIMEOUT`/`OUT OF MEMORY`: some LDV ones and some pthread-wmm ones.
2. 118 `unknown`s turned into `true`: many of them are eca-rers2012, hardness-nfm22 and product-lines, but surprisingly a couple of goblint-regression ones as well!

So there is something to gain!

Randomly looking into one, we also seem to do some weird things because the list includes (even builtin) function names:
```
[Info] Enabled octagon domain ONLY for:
[Info] pthread_mutex_lock, pthread_mutex_unlock, __VERIFIER_assert, pthread_create, g, h, abort, __assert_fail, reach_error, [...]
```
I don't think the attribute does anything on functions: the autotuner specifically looks for locals inside functions to annotate.
I'll try a small fix to filter functions because it might also be hurting us to pick a function over an actual global.

PR #1834, opened 2025-10-02
Fix octagon autotuner choosing non-integer variables

sim642:
Fixes the issue I accidentally found in https://github.com/goblint/analyzer/issues/1832#issuecomment-3359501789:

> Randomly looking into one, we also seem to do some weird things because the list includes (even builtin) function names:
> ```
> [Info] Enabled octagon domain ONLY for:
> [Info] pthread_mutex_lock, pthread_mutex_unlock, __VERIFIER_assert, pthread_create, g, h, abort, __assert_fail, reach_error, [...]
> ```
> I don't think the attribute does anything on functions: the autotuner specifically looks for locals inside functions to annotate.
> I'll try a small fix to filter functions because it might also be hurting us to pick a function over an actual global.

This introduces a new `RelationCil` module to share the notion of tracked variables/types between the actual relational analyses and the autotuner, so the logic doesn't go out of sync. It's a separate module which doesn't depend on Apron because the autotuner doesn't.

Out of curiosity, I'll do some sv-benchmarks runs to see if this was actually a problem.

### DIFF
--- a/src/autoTune.ml
+++ b/src/autoTune.ml
@@ -334,17 +334,16 @@ type option = {
 (*Option for activating the octagon apron domain on selected vars*)
 module VariableMap = Map.Make(CilType.Varinfo)
 module VariableSet = Set.Make(CilType.Varinfo)
+module OctagonTracked = RelationCil.AutotuneTracked
 
 let isComparison = function
   | Lt | Gt |	Le | Ge | Ne | Eq -> true
   | _ -> false
 
-let isGoblintStub v = List.exists (fun (Attr(s,_)) -> s = "goblint_stub") v.vattr
-
 let rec extractVar = function
   | UnOp (Neg, e, _)
   | CastE (_, e) -> extractVar e
-  | Lval ((Var info),_) when not (isGoblintStub info) -> Some info
+  | Lval ((Var info),_) when OctagonTracked.varinfo_tracked info -> Some info
   | _ -> None
 
 let extractBinOpVars e1 e2 =
@@ -376,7 +375,7 @@ class octagonVariableVisitor(varMap, globals) = object
         List.iter (fun var -> addOrCreateVarMapping varMap var 5 globals) (extractOctagonVars e2);
         DoChildren
       )
-    | Lval ((Var info),_) when not (isGoblintStub info) -> addOrCreateVarMapping varMap info 1 globals; SkipChildren
+    | Lval ((Var info),_) when OctagonTracked.varinfo_tracked info -> addOrCreateVarMapping varMap info 1 globals; SkipChildren
     (*Traverse down only operations fitting for linear equations*)
     | UnOp (LNot, _,_)
     | UnOp (Neg, _,_)
--- a/src/util/relationCil.ml
+++ b/src/util/relationCil.ml
@@ -8,7 +8,7 @@ sig
   val varinfo_tracked: varinfo -> bool
 end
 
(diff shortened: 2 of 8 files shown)


src/util/relationCil.ml
```diff
@@ -0,0 +1,40 @@
...
+
+(** To be used in relational analyses. *)
+module Tracked: Tracked =
+struct
+  include TypeTracked
+
+  let varinfo_tracked vi =
+    (* no vglob check here, because globals are allowed in relation, but just have to be handled separately *)
+    let hasTrackAttribute = List.exists (fun (Attr(s,_)) -> s = "goblint_relation_track") in
+    type_tracked vi.vtype && (not @@ GobConfig.get_bool "annotation.goblint_relation_track" || hasTrackAttribute vi.vattr)
```
michael-schwarz: Does this parse as intended? Even if it does, maybe doing

```suggestion
    type_tracked vi.vtype && ((not @@ GobConfig.get_bool "annotation.goblint_relation_track") || hasTrackAttribute vi.vattr)
```

is the saner thing.

src/autoTune.ml
```diff
@@ -334,17 +334,16 @@ type option = {
 (*Option for activating the octagon apron domain on selected vars*)
 module VariableMap = Map.Make(CilType.Varinfo)
 module VariableSet = Set.Make(CilType.Varinfo)
+module OctagonTracked = RelationCil.AutotuneTracked
```
michael-schwarz, line 337: Also, maybe one needs to rename the option in autotune, because I think it applies globally to all relational domains? (Or at least those built around the shared `CilExp <-> ApronExp` layer?)

sim642:
> Out of curiosity, I'll do some sv-benchmarks runs to see if this was actually a problem.

There's 19 `unknown` → `true` changes:
* 3 in goblint-regression/:
  * goblint-regression/28-race_reach_11-ptr_racing (no-overflow)
  * goblint-regression/13-privatized_34-traces-minepp-L-needs-to-be-um_true (unreach-call)
  * goblint-regression/36-apron_21-traces-cluster-based_true (unreach-call)
* 15 in in product-lines/minepump* (unreach-call)
* ldv-commit-tester/m0_drivers-hwmon-s3c-hwmon-ko--130_7a--af3071a (unreach-call)

The goblint-regression ones are the same as in https://github.com/goblint/analyzer/issues/1832#issuecomment-3359501789, so they don't really need more globals, just not choosing untrackable globals.

There's also one `true` → `unknown`:
* ldv-consumption/32_7a_cilled_linux-3.8-rc1-drivers--staging--comedi--drivers--ni_660x.ko-main.cil.out (unreach-call)

The latter is curious because before octagons weren't activated by the autotuner at all, while now they are, causing a timeout. But that's still with our current variable limits (#1832).

src/autoTune.ml
sim642: The attributes it adds would apply to all of those relational domains, but the autotuner only enabled octagon and only looks for variables in octagon-like expressions I think. So it's a bit of a confusing mix.

src/util/relationCil.ml
sim642: I think it did the right thing before but indeed it's not obvious. I changed it slightly differently: if there's going to be parentheses, then the `@@` isn't really needed anymore.

michael-schwarz (approved):

PR #1838, opened 2025-10-14
Add Mopsa checks JSON output for dashboard

Robotechnic:
The PR adds a new kind of output inspired by this paper from R. Monat & al: https://inria.hal.science/hal-04652657v2. This output contains all the checks performed by the analyzer and their result of three types:
- Error: There will be an error in any possible cases
- Warning: There might be an error in some cases
- Safe: There can't be an error in any possible cases

This output only takes into account the following checks:
- Integer overflow/underflow
- Invalid array access
- Division by zero errors
- Assersion violations
- Unknown function ptr
- Double free
- Writing to string literals

### DIFF
--- a/src/common/util/checks.ml
+++ b/src/common/util/checks.ml
@@ -40,7 +40,7 @@ module Category = struct
     | DoubleFree
     | NegativeArraySize
     | StubCondition
-  [@@deriving hash, eq, show]
+  [@@deriving hash, eq]
 
   let to_yojson x = `String (match x with
       | AssertionFailure -> "Assertion failure"
@@ -102,12 +102,7 @@ module Check = struct
       (Format.pp_print_option CilType.Location.pp) check.range
 end
 
-module CheckMap = Hashtbl.Make (struct
-    type t = Check.t
-    let equal = Check.equal
-    let hash = Check.hash
-  end)
-
+module CheckMap = Hashtbl.Make (Check)
 
 module CategoryLocationMap = Hashtbl.Make (struct
     type t = Category.t * CilType.Location.t [@@deriving hash, eq]


src/common/util/checks.ml
```diff
@@ -0,0 +1,172 @@
...
+
+let error category = Check.check Kind.Error category
+
+let warn category = Check.check Kind.Warning category
+
+let safe ?(message = "") category = 
+  match !Node0.current_node with
+  | Some (Statement _) ->
+    Check.check Kind.Safe category "%s" message
+  | _ -> ()
```
sim642: Why does `safe` have this conditional, but not `warn` or `error`?

src/analyses/assert.ml
```diff
@@ -35,15 +35,19 @@ struct
     match Queries.eval_bool (Analyses.ask_of_man man) e with
     | `Lifted false ->
       warn (M.error ~category:Assert "%s") ~annot:"FAIL" ("Assertion \"" ^ expr ^ "\" will fail.");
+      Checks.error Checks.Category.AssersionFaillure "Assertion \"%s\" will fail." expr;
       if refine then raise Analyses.Deadcode else man.local
     | `Lifted true ->
       warn (M.success ~category:Assert "%s") ("Assertion \"" ^ expr ^ "\" will succeed");
+      Checks.safe Checks.Category.AssersionFaillure;
```
sim642: No message here?

src/analyses/base.ml
```diff
@@ -1176,10 +1187,13 @@ struct
...
-        M.warn ~category:Imprecise ~tags:[Category Call] "Unknown call to function %a." d_exp fval
-      else
-        M.warn ~category:Imprecise ~tags:[Category Call] "Function pointer %a may contain unknown functions." d_exp fval
+      if AD.cardinal fp = 1 then (
+        M.warn ~category:Imprecise ~tags:[Category Call] "Unknown call to function %a." d_exp fval;
+        Checks.warn Checks.Category.InvalidMemoryAccess "Unknown call to function %a." d_exp fval
+      ) else (
+        M.warn ~category:Imprecise ~tags:[Category Call] "Function pointer %a may contain unknown functions." d_exp fval;
+        Checks.warn Checks.Category.InvalidMemoryAccess "Function pointer %a may contain unknown functions." d_exp fval
+      )
```
sim642, line 1194: Why does this go under `InvalidMemoryAccess`?

src/analyses/base.ml
```diff
@@ -2223,7 +2237,10 @@ struct
     end
 
   let special_unknown_invalidate man f args =
-    (if CilType.Varinfo.equal f dummyFunDec.svar then M.warn ~category:Imprecise ~tags:[Category Call] "Unknown function ptr called");
+    (if CilType.Varinfo.equal f dummyFunDec.svar then (
+        M.warn ~category:Imprecise ~tags:[Category Call] "Unknown function ptr called";
+        Checks.warn Checks.Category.InvalidMemoryAccess "Unknown function ptr called"
+      ));
```
sim642, line 2240: Same here: why does this go under `InvalidMemoryAccess`?

src/common/util/checks.ml
```diff
@@ -0,0 +1,172 @@
...
+  let of_yojson = function
+    | `String "error" -> Ok Error
+    | `String "warning" -> Ok Warning
+    | `String "safe" -> Ok Safe
+    | kind -> Error ("Checks.Kind.of_yojson: Invalid kind :" ^ Yojson.Safe.to_string kind)
+end
+
+module Category = struct
+  type t =
+    | AssersionFaillure
```
sim642:
```suggestion
    | AssersionFailure
```
And everywhere else.

src/analyses/baseInvariant.ml
```diff
@@ -293,9 +293,11 @@ struct
       let warn_and_top_on_zero x =
         if GobOption.exists (Z.equal Z.zero) (ID.to_int x) then
           (M.error ~category:M.Category.Integer.div_by_zero ~tags:[CWE 369] "Must Undefined Behavior: Second argument of div or mod is 0, continuing with top";
+           Checks.error Checks.Category.DivisionByZero "Must Undefined Behavior: Second argument of div or mod is 0, continuing with top";
            ID.top_of ikind)
-        else
-          x
+        else (
+          Checks.safe Checks.Category.DivisionByZero;
+          x)
```
sim642, line 300: Hmm, it's strange that we have error output in the middle of conditional refinement. I suspect it's redundant due to #1764.

src/cdomain/value/cdomains/arrayDomain.ml
```diff
@@ -992,7 +997,9 @@
   type substr = IsNotSubstr | IsSubstrAtIndex0 | IsMaybeSubstr
 
   module ArrayOobMessage = M.Category.Behavior.Undefined.ArrayOutOfBounds
-  let warn_past_end = M.error ~category:ArrayOobMessage.past_end
+  let warn_past_end ?loc ?tags fmt = 
+    let () = Checks.error Checks.Category.InvalidMemoryAccess fmt |> ignore in
+    M.error ~category:ArrayOobMessage.past_end ?loc ?tags fmt
```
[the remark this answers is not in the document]
sim642: There's more instances of it in this file at least. But I guess semgrep is being useless at finding them again...

src/cdomain/value/cdomains/int/intDomTuple.ml
```diff
@@ -82,7 +86,7 @@ module IntDomTupleImpl = struct
     let intv =  f p2 @@ r.fi2_ovc (module I2) in
     let intv_set = f p5 @@ r.fi2_ovc (module I5) in
     let bf = f p6 @@ r.fi2_ovc (module I6) in
-    ignore (check_ov ~suppress_ovwarn ~cast:false ik intv intv_set bf);
+    ignore (check_ov ~suppress_ovwarn ~no_safe:true ~cast:false ik intv intv_set bf);
```
sim642: Was this only to suppress safe checks on integer literals? if so, then this might be redundant after #1794.

src/common/util/checks.ml
```diff
@@ -0,0 +1,172 @@
...
+    | InvalidMemoryAccess
+    | DivisionByZero
+    | IntegerOverflow
+    | InvalidPointerComparison
+    | InvalidPointerSubtraction
```
sim642, line 39: Are these used anywhere?

src/common/util/checks.ml
```diff
@@ -0,0 +1,172 @@
...
+  let pp fmt check =
+    Format.fprintf fmt "Check: %a: %s at %a"
+      Kind.pp check.kind
+      check.messages
+      (Format.pp_print_option CilType.Location.pp) check.range
+
+  type check_t = t
+  let check_t_to_yojson = to_yojson
+  module CheckMap = Hashtbl.Make (struct
+      type t = check_t
```
sim642: I'm not sure `check_t` is necessary. You could use `type nonrec t = t` to refer to the outer type instead of making it recursive.

sim642: Actually, this and the following code in the `Check` submodule might make more sense to be moved to the top level of `checks.ml` because these are not part of a single check anymore and end up as `Checks.Check.CheckMap`.

src/common/util/checks.ml
```diff
@@ -0,0 +1,172 @@
...
+      let equal = equal
+      let hash = hash
+    end)
+
+
+  module CategoryLocationMap = Hashtbl.Make (struct
+      type t = Category.t * CilType.Location.t [@@deriving hash, eq]
+    end)
+
+  type key = Category.t * CilType.Location.t option [@@deriving yojson]
```
sim642: This seems unused?

src/common/util/checks.ml
```diff
@@ -0,0 +1,172 @@
...
+
+  let checks_list : (bool ref * unit CheckMap.t) CategoryLocationMap.t = CategoryLocationMap.create 113
```
sim642: What is this `bool ref`? Would be good to document it here at least.

src/common/util/checks.ml
```diff
@@ -0,0 +1,172 @@
...
+        (* Mark all ranges as synthetic for hash purposes *)
+        let range = { range with synthetic = true } in
+        let check = { check with range = Some range } in
+        let check_key = (check.title, range) in
+        match CategoryLocationMap.find_opt checks_list check_key with
+        | Some (safe, existing_checks) ->
+          if !safe && Kind.is_safe check.kind then
+            CheckMap.replace existing_checks check ()
+          else if not @@ Kind.is_safe check.kind then (
+            if !safe then CheckMap.clear existing_checks;
```
sim642: Why are some previously added checks being forgotten here?

src/common/util/checks.ml
```diff
@@ -0,0 +1,172 @@
...
+  let to_yojson x = `String (match x with
+      | Error -> "error"
+      | Warning -> "warning"
+      | Safe -> "safe")
+
+  let of_yojson = function
+    | `String "error" -> Ok Error
+    | `String "warning" -> Ok Warning
+    | `String "safe" -> Ok Safe
+    | kind -> Error ("Checks.Kind.of_yojson: Invalid kind :" ^ Yojson.Safe.to_string kind)
```
sim642: I think these could be still derived using this: https://github.com/ocaml-ppx/ppx_deriving_yojson?tab=readme-ov-file#name.

src/common/util/checks.ml
Robotechnic: Because `warn` and `error` always have a message attached, while most of the time `safe` doesn't

src/analyses/assert.ml
Robotechnic: We can add a message, but Mopsa doesn't, so I didn't.

src/analyses/base.ml
Robotechnic: Because it follows the same categorisation as Mopsa:
```c
int (*funcPtr)(int) = NULL;
funcPtr(10);
```
raises:
```
error: Invalid memory access
  
  7:      return funcPtr(10);
                 ^^^^^^^     
  pointer 'funcPtr' may be null
  Callstack:
        from ./test.c:5.4-8: main
```
And the same goes for any invalid function pointer manipulation.

src/common/util/checks.ml
Robotechnic: No, that's right, failure only takes one 'l.'

src/analyses/baseInvariant.ml
Robotechnic: It seems like it, but I'm not sure

src/cdomain/value/cdomains/int/intDomTuple.ml
Robotechnic: Yes and yes, probably, I'll check

src/common/util/checks.ml
Robotechnic: Not in Goblint, but they are part of the possible Mopsa output (I, however, haven't yet encountered them). This is why they are here.

src/common/util/checks.ml
Robotechnic: yojson converts those to `["error"]` but I want those to be converted to `"error"`. As far as I know, there are no rules to change this behaviour.

src/common/util/checks.ml
sim642: Sure, but this isn't conditional on whether there is a `message` (or really, if it's the empty string), but rather `Node0.current_node`. These two things aren't related:
1. If there's a `current_node` but the message is empty, then the safe check with empty message is still recorded.
2. If there's no `current_node` but there is a safe check (regardless of message), it is just forgotten.

I suspect the `current_node` check is a roundabout way to prevent safe checks from something that shouldn't output safe checks, but I don't know what that would be.

src/analyses/base.ml
sim642: These warnings aren't about that though. They're whether the points-to-set of `funcPtr` contains an unknown non-`NULL` pointer.
For that example code Goblint outputs
```
[Error][Behavior > Undefined > NullPointerDereference][CWE-476] Must dereference NULL pointer (call-null-ptr.c:5:3-5:14)
[Warning][Unsound][Call] No suitable function to be called at call site. Continuing with state before call. (call-null-ptr.c:5:3-5:14)
```
Neither message comes from here.

src/common/util/checks.ml
Robotechnic: The match is here because the safe messages can be sometimes in places where `Node0` is None. In this cases we can't have a location and the safe is hence invalid. The check isn't in the `warn` and `error` checks because this `None` case never happen.

src/analyses/base.ml
Robotechnic:
```
Assertion failure
Stub condition
Invalid memory access
Division by zero
Integer overflow
Invalid shift
Invalid pointer comparison
Invalid pointer subtraction
Double free
Insufficient variadic arguments
Insufficient format arguments
Invalid type of format argument
Invalid floating-point operation
Floating-point division by zero
Floating-point overflow
```
Ok, I agree. However, it seems like it is the most appropriate category among the available ones.

src/common/util/checks.ml
Robotechnic: It's to have the ability to mutate it instead of having to use `replace` to update the value.

src/common/util/checks.ml
Robotechnic: This prevents goblint from emitting safe and unsafe reports at the same time. Safe are emitted for different reasons, and sometimes, a range is safe and unsafe for two different reasons but still under the same category. This filters those cases.

src/cdomain/value/cdomains/int/intDomTuple.ml
Robotechnic: It doesn't seem to work with the last Goblint version:
```c
int main() {
	return 10;
}
```
Yields a safe integer overflow.

src/analyses/base.ml
sim642: I think it makes more sense to not have the checks here at all because they're not about possible runtime erros, but just messages about Goblint warning about being too imprecise.

There are plenty of other messages which don't have corresponding checks either.

src/common/util/checks.ml
sim642: Do you have examples of this were removing the extra condition would be a problem?

because inside `Check.check` there's a condition on `AnalysisState.should_warn` anyway, which is the same thing that control's Goblint's existing message output. It could be that there are places where `should_warn` is currently incorrectly (un)set and should be fixed, instead of covering up the issue here.

The condition on `current_node` also makes checks kind of asymmetric: warnings/errors from such places would still be added.

src/common/util/checks.ml
sim642: Exactly, but currently there's two.

src/analyses/baseInvariant.ml
sim642: We can let it be here for now, I opened #1892 to possibly remove this.

src/cdomain/value/cdomains/int/intDomTuple.ml
sim642: But that's on this branch with the `~no_safe:true` in place? Because it would mean that the `~no_safe` isn't even silencing what it is trying to silence.

src/common/util/checks.ml
sim642: Oh, I see, argument-less constructors still get the list: https://github.com/ocaml-ppx/ppx_deriving_yojson/issues/32. So this will have to remain indeed, which is a bit unfortunate.

src/cdomain/value/cdomains/arrayDomain.ml
```diff
@@ -992,7 +997,9 @@ struct
   type substr = IsNotSubstr | IsSubstrAtIndex0 | IsMaybeSubstr
 
   module ArrayOobMessage = M.Category.Behavior.Undefined.ArrayOutOfBounds
-  let warn_past_end = M.error ~category:ArrayOobMessage.past_end
+  let warn_past_end ?loc ?tags fmt = 
+    Checks.error Checks.Category.InvalidMemoryAccess fmt |> ignore;
+    M.error ~category:ArrayOobMessage.past_end ?loc ?tags fmt
```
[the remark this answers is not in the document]
Robotechnic: This can't work because the fmt is used as a format string in some places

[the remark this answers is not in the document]
sim642: I pushed a proper fix to this. It's quite tricky to achieve the right thing here.

PR #1840, opened 2025-10-15
Add semgrep rule let-unit-in

sim642:
This adds a semgrep rule which suggests to replace `let () = e in ...` just with `e; ...`.

I noticed a bunch being introduced in #1838, so we should have an automatic rule towards simpler code.

### DIFF
--- a/.semgrep/let.yml
+++ b/.semgrep/let.yml
@@ -1,6 +1,6 @@
 rules:
   - id: let-unit-in
     pattern: let () = $E in ...
-    message: use ; instead
+    message: use ; instead (and, if needed, add surrounding parentheses to preserve precedence)
     languages: [ocaml]
     severity: WARNING


.semgrep/let.yml
```diff
@@ -0,0 +1,6 @@
+rules:
+  - id: let-unit-in
+    pattern: let () = $E in ...
+    message: use ; instead
```
michael-schwarz: Does naivley following this advice sometimes lead to bugs? Thinking of

~~~Ocaml
if true then
   let () = do_something () in
   more_stuff()
~~~


vs

~~~OCaml
if true then
    do_something();
    more_stuff()
~~~


Maybe add a caution here ("carefully checking if parens are needed").

sim642: Indeed, good point!

michael-schwarz (approved):

PR #1845, opened 2025-10-21
Executing a Portfolio of Runs via Script

DrMichaelPetter:
in SVCOMP, we want to start our analysis with a cheap config, and refine with increasingly more expensive but more capable configurations until we solve this task. This PR adds a python script to iterate through a list of configurations, until we either run out of options, encounter an error or find a verdict different from ```unknown```.

In detail, we get the script ```goblint_runner.py``` , which delegates control to ```goblint``` with all of its parameters. Only as soon as we start using parameter ```--portfolio-conf [FILE]``` the script starts its iteration through the lines of ```[FILE]```. Each line contains a set of parameters for a goblint run. In particular, we would expect to see a value parameter ```--conf``` per line.


conf/svcomp26/seq.txt
```diff
@@ -0,0 +1,4 @@
+# standard run based on svcomp 25 settings
+svcomp26/level00.json
```
sim642: I'm wondering if it'd make sense to even make this more general: just a list of Goblint arguments, so `--conf conf/svcomp26/level00.json`.

That would make it easier to reuse confs for Goblint-Par with multishot by writing something like `--conf conf/svcomp26/level00.json --set solver td_something`. Instead of having to make copies of all confs with a single line changed.

goblint_runner.py
```diff
@@ -0,0 +1,107 @@
...
+                exit(1)
+            with open(conf_args.portfolio, "r") as conflist_file:
+                self.configs = [path.join(self.current_path, "conf", c.strip())
+                                for c in conflist_file.readlines() if not c.strip().startswith("#")]
+            logger.info(f"Loaded goblint configs: {", ".join(self.configs)}")
+            
+    def run_with_config(self, config_path):
+        args = ["--conf", config_path] + self.other_args
+        self.logger.info(f"Running next shot: ./goblint {" ".join(args)}")
+        output = subprocess.check_output([self.goblint_executable_path, *args]
```
sim642: Does this also show Goblint output on in the terminal or will it just swallow everything?
The latter would be bad because it would make it impossible to see any actual Goblint output, making SV-COMP runs hard to debug.

goblint_runner.py
```diff
@@ -0,0 +1,107 @@
...
+                                        ).decode("utf-8")        
+
+        return ExecutionResult(output, self.has_verdict(output))
+
+    def run_without_config(self):
+        subprocess.run([self.goblint_executable_path, *self.other_args])
+
+    @staticmethod
+    def has_verdict(output):
+        return any("SV-COMP result: true" in line for line in output.splitlines())
```
sim642: Currently if Goblint returns an `ERROR (something)` result, this would just cause the next conf to be run, right?

I think it would be better if such errors also immediately stopped the portfolio with that result. So I guess only `SV-COMP result: unknown` would trigger the next conf.

This also relates to the previous comment about Goblint output visibility because the BenchExec tool info module parses errors into different statuses here: https://github.com/sosy-lab/benchexec/blob/4970b1aef889c9069e68989caa82591737ec098a/benchexec/tools/goblint.py#L89-L100.
We should still be able to distinguish those through the wrapper, e.g. distinguish various uncaught exceptions by their type in the BenchExec results table.

goblint_runner.py
```diff
@@ -0,0 +1,107 @@
...
+            return
+
+        result = None
+        for config in self.configs:
+            if not path.exists(config):
+                logger.warning(f"Config file {config} not found, skipping.")
+                continue
+            if not path.isfile(config):
+                logger.warning(f"Config file {config} is not a file, skipping.")
+                continue
```
sim642: These don't seem like a good idea: if you make a typo in the portfolio, you might never notice it because we just skip over it to the next conf.
It should be fine to just not check these at all and try running Goblint with the conf file. If Goblint cannot read it for whatever reason, it will crash. And the crash should stop the portfolio immediately instead of jumping to the next conf.

conf/svcomp26/level01.json
```diff
@@ -0,0 +1,133 @@
...
+        "memsafetySpecification",
+        "noOverflows",
+        "termination",
+        "tmpSpecialAnalysis"
+      ]
+    }
+  },
+  "exp": {
+    "region-offsets": true,
+    "unrolling-factor": 128
```
sim642: Have you checked how this plays together with the autotuner which also sets per-loop unrolling factors?
Maybe the loop unrolling autotuner should just be disabled if we want a constant global value.

conf/svcomp26/seq.txt
DrMichaelPetter: hm, I actually like this idea.

conf/svcomp26/level01.json
DrMichaelPetter: for now, level01 has not much thought behind it, this is just a dummy to test the sequence of portfolios. But I will take that into account when we do the evaluation of the particular configurations!

goblint_runner.py
DrMichaelPetter: This currently is a compromise: it shows stderrr output, but not stdout output, for purposes of debugging. Is that enough to debug in your opinion?

sim642: Probably not, most of Goblint output (messages) are still on stdout.

DrMichaelPetter: ok, stdout and stderr are now there

goblint_runner.py
```diff
@@ -0,0 +1,105 @@
...
+        self.logger.info(f"Running next shot: ./goblint {" ".join(args)}")
+        process = subprocess.Popen([self.goblint_executable_path, *args],stdout=subprocess.PIPE,stderr=subprocess.STDOUT)
+        output = []
+        for line in process.stdout:
+            decoded_line = line.decode("utf-8")
+            print(decoded_line, end="")
+            output.append(decoded_line)
+        process.wait()
+        output = "".join(output)            
+        return ExecutionResult(output, self.verdict_true_or_error(output))
```
sim642: This constructs a large list and an large string of the entire output but that's actually excessive. We could just be checking if each line has the verdict without accumulating the output itself.

DrMichaelPetter: Yeah, we do not need the output buffer any more now

conf/svcomp26/level01.json
DrMichaelPetter: A first quick glance at a comparison does not show very promising changes anyway, to manually fiddle around with unrolling. More to come, when we finish our comparison of all config variations, the coming days.

goblint_runner.py
```diff
@@ -0,0 +1,109 @@
...
+    def run_with_config(self, config_str):
+        config_args = config_str.split(" ")
+        # in config_args, replace any relative paths with absolute paths
+        pathparameters=["--conf"] # add more if needed
+        for i in range(len(config_args)):
+            if config_args[i] in pathparameters and i + 1 < len(config_args):
+                conf_path = config_args[i + 1]
+                if not path.isabs(conf_path):
+                    abs_conf_path = path.join(self.current_path, conf_path)
+                    config_args[i + 1] = abs_conf_path
```
sim642: This shouldn't be necessary: Goblint should already find confs relative to itself.

scripts/sv-comp/goblint_runner.py
sim642, line 1: Also, I think this should go into `scripts/sv-comp/` in order to not clutter the repository root.
I will make #1841 copy it still next to `goblint` binary for the SV-COMP archive, so the script itself shouldn't need any change to work there.

For easier in-repository testing it might be convenient for the script (in `scripts/sv-comp/`) to find the `goblint` binary two levels above, but this would not be used in SV-COMP.

PR #1862, opened 2025-11-04
Use-after-free Analysis: Fix warning generation for lvalues

### LINKED ISSUE #1861: double free not detected in SV-COMP 26
A wrong verdict true in svcomp as observable by
```
./goblint --set ana.sv-comp.enabled true --set ana.sv-comp.functions true --set ana.path_sens[+] base --sets ana.specification /home/goblint/sv-benchmarks/c/properties/valid-memsafety.prp --sets exp.architecture 32bit /home/goblint/sv-benchmarks/c/ldv-memsafety/memleaks_test19-2.i --set ana.autotune.enabled false --set ana.malloc.unique_address_count 5
```
gives rise to the following snipped, simplified by @michael-schwarz 
```c
#include <stdlib.h>
#include <stddef.h>
#include <stdio.h>
#include <string.h>
#include <wchar.h>

struct A19 {
 int *p;
};

void init(struct A19 *a) {
 a->p = (int *)malloc(sizeof(int));

 free(a->p);
 free(a->p); //WARN
}


void main(void) {
 struct A19 a19;
 a19.p = NULL;
 init(&a19);
}
```
Here, we miss the warning as marked in the code.

jerhard:
This fixes the handling of lvals in the warning generation of the use-after-free analysis in `warn_lval_might_contain_freed`.

### Background
Given, e.g., a statement
    
    lval = rval;
    
 There may be two types of use-after-free errors that can occur for the lval:
    
 1. The memory location to which lval evaluates has been freed already.
 2. The memory location(s) inside lval that are dereferenced have been freed already.
    
 Previously, the check implemented for lvals did not consistently check for both types of errors.
 In case a dereference was at the top-level of the lval, only the second type of error was checked; also this was not check recursively for sub-lvals.
 
### Changes 

This is changed with this PR.
The check for (1.) is now performed whether the lhost is a variable or a memory dereference.
The check for (2.) is performed, in case that the lhost is a dereference, and uses a recursive call of the function `warn_exp_might_contain_freed` that is mutually recursive with `warn_lval_might_contain_freed`.

Fixes #1861


src/analyses/useAfterFree.ml
```diff
@@ -58,76 +58,88 @@ struct
...
-              ) ad []
-          in
+      let (lval_host, o) = lval in
+      offset_might_contain_freed o; (* Check the lval's offset *)
+
+      let freed_heap_vars = snd man.local in
+      let undefined_behavior = if is_free then Undefined DoubleFree else Undefined UseAfterFree in
+      let cwe_number = if is_free then 415 else 416 in
+
+      let check_adress adresses =
```
sim642, line 112:
```suggestion
      let check_address addresses =
```
Also usages of both below.

src/analyses/useAfterFree.ml
```diff
@@ -58,76 +58,88 @@ struct
...
+          end
+        in
+        let pointed_to_heap_vars =
+          Queries.AD.fold (fun addr vars ->
+              match addr with
+              | Queries.AD.Addr.Addr (v,_) when man.ask (Queries.IsAllocVar v) -> v :: vars
+              | _ -> vars
+            ) adresses []
+        in
+        if not (Queries.AD.is_top adresses) then begin
```
sim642, line 126: The `pointed_to_heap_vars` above could be moved into this `if`, right?

Issue #1805, opened 2025-08-08
Consider some more interactions between thread creation, joins, and mutexes


michael-schwarz:
**Segment-Based May-Happen-in-Parallel Analysis for C Programs** by Fu et al. has some interesting cases why some things cannot race, e.g.,
- the parent thread holding some mutex throughout the entire execution of the child, or
- the parent thread starting while holding the same mutex the child will acquire, which yields ordering information.

<img width="318" height="392" alt="Image" src="https://github.com/user-attachments/assets/7e6adc0e-e3af-479b-b026-6e611c659323" />

This may be interesting for, e.g., a Bachelor's student to implement.

----

Paper: https://onlinelibrary.wiley.com/doi/10.1002/cpe.70203

PR #1865, opened 2025-11-05
Improve MHP precision using ancestor locksets

dabund24:
First part of #1805.
Second case is handled in #1913.

# To be handled
## Non-transitive version
When creating $t_1$, $t_0$ must hold a lock $l$. If $l$ is not released before $t_1$ is definitely joined into $t_0$, $t_1$ is protected by $l$.

### Examples

```mermaid
graph TB;
subgraph t1;
    E["..."]-->F["return;"];
end;
subgraph t0;
    A["lock(l);"]-->B;
    B["create(t1);"]-->C;
    C["join(t1);"]-->D["unlock(l);"]
end;
B-.->E
F-.->C
```

```mermaid
graph TB;
subgraph t1;
    E["..."]-->F["return;"];
end;
subgraph t0;
    A["lock(l);"]-->B;
    B["create(t1);"]-->C[return;];
end;
B-.->E
```

## General version

Let $t_1$ be a must-ancestor of $t_0$. When creating $t_1$, $t_0$ must hold a lock $l$. If $l$ is not released before $t_d$ is definitely joined into $t_0$, $t_d$ is protected by $l$.

### Example

```mermaid
graph TB;
subgraph td;
    G["..."]-->H["return;"];
end;
subgraph t1;
    E["create(td);"]-->F["return;"];
end;
subgraph t0;
    A["lock(l);"]-->B;
    B["create(t1);"]-->C;
    C["join(td);"]-->D["unlock(l);"]
end;
B-.->E
E-.->G
H-.->C
```
# Dependency Analyses
- $\mathcal T$: Ego Thread Id at program point 
- $\mathcal L$: Must-Lockset at program point
- $\mathcal C$: May-Creates of ego thread before program point
- $\mathcal J$: Transitive Must-Joins of ego thread before program point
- $\mathcal{DES}\ t$: Descendant threads of $t$ (implemented in this PR)
- $\mathcal A\ t$: Must-ancestors of $t$

# Conditions to satisfy

1. $t_0\in\mathcal A\ t_1\land (t_1=t_d\lor t_1\in\mathcal A\ t_d)$
2. maybe $\exists$ `create(t1)` in $t_0$ with $l\in\mathcal L$
3. $\forall$ `create(t1)` in $t_0:l\in\mathcal L$
4. $\forall$ `unlock(l)` in $t_0:t_d\notin\left(\mathcal C\cup\bigcup_{c\in\mathcal C}\mathcal{DES}\ c\right)\setminus\mathcal J$

# Analysis

- $\mathcal{CL}\subseteq T\to T\to 2^L$
- $T\to 2^L$ is `MapBot`
- $2^L$ is Must-Set
- Flow-Insensitive
- $(t_1\mapsto\\{t_0\mapsto L\\})\in\mathcal{CL}$ means " $t_1$ is protected by all mutexes in $L$ locked in $t_0$ and by nothing else".

## Contributions
- `create(t1)`:
  $\forall t\in t_1\cup\mathcal{DES}\ t_1:$
  $$\mathcal{CL}\ t\sqsupseteq\\{\mathcal T\mapsto\mathcal L\\}$$

- `unlock(l)`:
  $\forall t\in \left(\mathcal C\cup\bigcup_{c\in\mathcal C}\mathcal{DES}\ c \right)\setminus\mathcal J:$
  $\mathcal{CL}\ t\sqsupseteq \\{\mathcal T\mapsto(\mathcal{CL}\ t\ \mathcal T)\setminus \\{l\\}\\}$

- `unlock` of unknown mutex:
  $\forall t\in \left(\mathcal C\cup\bigcup_{c\in\mathcal C}\mathcal{DES}\ c \right)\setminus\mathcal J:$
  $$\mathcal{CL}\ t\sqsupseteq \\{\mathcal T\mapsto\emptyset\\}$$

# Rules for MHP exclusion
Program points $s_1$ with $\mathcal T_1$, $\mathcal L_1$ and $\mathcal{CL}_1$ and $s_2$ with $\mathcal T_2$, $\mathcal L_2$ and $\mathcal{CL}_2$ cannot happen in parallel if at least one condition holds:
- $\exists (t_a\mapsto L_a)\in\mathcal{CL}_1:L_a\cap\mathcal L_2\neq\emptyset,t_a\neq \mathcal T_2$
- $\exists (t_a\mapsto L_a)\in\mathcal{CL}_2:L_a\cap\mathcal L_1\neq\emptyset,t_a\neq \mathcal T_1$
- $\exists(t_{a1}\mapsto L_{a1})\in\mathcal{CL}_ 1,(t_{a2}\mapsto L_{a2})\in\mathcal {CL}_ 2: L_{a1}\cap L_{a2}\neq\emptyset\land t_{a1}\neq t_{a2}$

### DIFF
--- a/.semgrep/let.yml
+++ b/.semgrep/let.yml
@@ -1,6 +1,8 @@
 rules:
-  - id: let-unit-in
-    pattern: let () = $E in ...
-    message: use ; instead (and, if needed, add surrounding parentheses to preserve precedence)
-    languages: [ocaml]
-    severity: WARNING
+  []
+  # TODO: disabled because semgrep cannot distinguish from let@ which is fine (https://github.com/semgrep/semgrep/issues/11432)
+  # - id: let-unit-in
+  #   pattern: let () = $E in ...
+  #   message: use ; instead (and, if needed, add surrounding parentheses to preserve precedence)
+  #   languages: [ocaml]
+  #   severity: WARNING
--- a/scripts/bash-completion.sh
+++ b/scripts/bash-completion.sh
@@ -25,7 +25,7 @@ _goblint ()
     IFS=$'\n'
     local words cword cur
     _get_comp_words_by_ref -n = cur words cword # Bypass = in COMP_WORDBREAKS (https://stackoverflow.com/a/57437406/854540)
-    COMPREPLY=($(${words[0]} --complete "${words[@]:1:cword}"))
+    COMPREPLY=($(${words[0]/#\~/$HOME} --complete "${words[@]:1:cword}"))
     __ltrim_equal_completions "$cur" # Bypass = in COMP_WORDBREAKS (https://stackoverflow.com/a/57437406/854540)
 }
 
@@ -45,7 +45,7 @@ _regtest ()
         *)
             local words cword cur
             _get_comp_words_by_ref -n = cur words cword # Bypass = in COMP_WORDBREAKS (https://stackoverflow.com/a/57437406/854540)
-            COMPREPLY=($($(dirname ${words[0]})/goblint --complete "${words[@]:3:cword}"))
+            COMPREPLY=($($(dirname ${words[0]/#\~/$HOME})/goblint --complete "${words[@]:3:cword}"))
             __ltrim_equal_completions "$cur" # Bypass = in COMP_WORDBREAKS (https://stackoverflow.com/a/57437406/854540)
             ;;
     esac
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -169,6 +169,19 @@ struct
 
   let iDtoIdx x = ID.cast_to ~kind:Internal (Cilfacade.ptrdiff_ikind ()) x (* TODO: proper castkind *)
 
+  (** Unary float predicates return non-zero for [true].
+      @see C11 7.12.3 *)
+  let fd_unary_pred = function
+    | Some true -> ID.of_excl_list IInt [Z.zero]
+    | Some false -> ID.of_int IInt Z.zero
+    | None -> ID.top_of IInt
+
+  (** Binary float predicates return one for [true].
+      @see C11 7.12.14, 6.5.8, 6.5.9 *)
+  let fd_binary_pred = function
+    | Some b -> ID.of_bool IInt b
+    | None -> ID.top_of IInt
+
   let unop_ID = function
     | Neg  -> ID.neg
     | BNot -> ID.lognot
@@ -176,7 +189,7 @@ struct
 
   let unop_FD = function
     | Neg  -> (fun v -> (Float (FD.neg v):value))
-    | LNot -> (fun c -> Int (FD.eq c (FD.of_const (FD.get_fkind c) 0.)))
+    | LNot -> (fun c -> Int (fd_unary_pred (FD.eq c (FD.of_const (FD.get_fkind c) 0.))))
     | BNot -> failwith "BNot on a value of type float!"
 
 
@@ -256,7 +269,7 @@ struct
     | Ge -> FD.ge
     | Eq -> FD.eq
     | Ne -> FD.ne
-    | _ -> (fun _ _ -> ID.top ())
+    | _ -> (fun _ _ -> None)
 
   let is_int_returning_binop_FD = function
     | Lt | Gt | Le | Ge | Eq | Ne -> true
@@ -280,7 +293,7 @@ struct
       let rec addToOffset n (t:typ option) = function
         | `Index (i, `NoOffset) ->
           (* Binary operations on pointer types should not generate warnings in SV-COMP *)
-          GobRef.wrap AnalysisState.executing_speculative_computations true @@ fun () ->
+          let@ () = GobRef.wrap AnalysisState.executing_speculative_computations true in
           (* If we have arrived at the last Offset and it is an Index, we add our integer to it *)
           `Index(IdxDom.add i (iDtoIdx n), `NoOffset)
         | `Field (f, `NoOffset) ->
@@ -346,7 +359,8 @@ struct
     (* For the float values, we apply the float domain operators *)
     | Float v1, Float v2 when is_int_returning_binop_FD op ->
       let result_ik = Cilfacade.get_ikind t in
-      Int (ID.cast_to ~kind:Internal result_ik (int_returning_binop_FD op v1 v2)) (* TODO: proper castkind *)
+      assert (result_ik = IInt);
+      Int (fd_binary_pred (int_returning_binop_FD op v1 v2))
     | Float v1, Float v2 -> Float (binop_FD (Cilfacade.get_fkind t) op v1 v2)
     (* For address +/- value, we try to do some elementary ptr arithmetic *)
     | Address p, Int n
@@ -2661,11 +2675,11 @@ struct
           | Nan (fk, str) when Cil.isPointerType (Cilfacade.typeOf str) -> Float (FD.nan_of fk)
           | Nan _ -> failwith ("non-pointer argument in call to function "^f.vname)
           | Inf fk -> Float (FD.inf_of fk)
-          | Isfinite x -> Int (ID.cast_to ~kind:Internal IInt (apply_unary FDouble FD.isfinite x)) (* TODO: proper castkind *)
-          | Isinf x -> Int (ID.cast_to ~kind:Internal IInt (apply_unary FDouble FD.isinf x)) (* TODO: proper castkind *)
-          | Isnan x -> Int (ID.cast_to ~kind:Internal IInt (apply_unary FDouble FD.isnan x)) (* TODO: proper castkind *)
-          | Isnormal x -> Int (ID.cast_to ~kind:Internal IInt (apply_unary FDouble FD.isnormal x)) (* TODO: proper castkind *)
-          | Signbit x -> Int (ID.cast_to ~kind:Internal IInt (apply_unary FDouble FD.signbit x)) (* TODO: proper castkind *)
+          | Isfinite x -> Int (fd_unary_pred (apply_unary FDouble FD.isfinite x))
+          | Isinf x -> Int (fd_unary_pred (apply_unary FDouble FD.isinf x))
+          | Isnan x -> Int (fd_unary_pred (apply_unary FDouble FD.isnan x))
+          | Isnormal x -> Int (fd_unary_pred (apply_unary FDouble FD.isnormal x))
+          | Signbit x -> Int (fd_unary_pred (apply_unary FDouble FD.signbit x))
           | Ceil (fk,x) -> Float (apply_unary fk FD.ceil x)
           | Floor (fk,x) -> Float (apply_unary fk FD.floor x)
           | Fabs (fk, x) -> Float (apply_unary fk FD.fabs x)
@@ -2676,12 +2690,12 @@ struct
           | Cos (fk, x) -> Float (apply_unary fk FD.cos x)
           | Sin (fk, x) -> Float (apply_unary fk FD.sin x)
           | Tan (fk, x) -> Float (apply_unary fk FD.tan x)
-          | Isgreater (x,y) -> Int(ID.cast_to ~kind:Internal IInt (apply_binary FDouble FD.gt x y)) (* TODO: proper castkind *)
-          | Isgreaterequal (x,y) -> Int(ID.cast_to ~kind:Internal IInt (apply_binary FDouble FD.ge x y)) (* TODO: proper castkind *)
-          | Isless (x,y) -> Int(ID.cast_to ~kind:Internal IInt (apply_binary FDouble FD.lt x y)) (* TODO: proper castkind *)
-          | Islessequal (x,y) -> Int(ID.cast_to ~kind:Internal IInt (apply_binary FDouble FD.le x y)) (* TODO: proper castkind *)
-          | Islessgreater (x,y) -> Int(ID.c_logor (ID.cast_to ~kind:Internal IInt (apply_binary FDouble FD.lt x y)) (ID.cast_to ~kind:Internal IInt (apply_binary FDouble FD.gt x y))) (* TODO: proper castkind *)
-          | Isunordered (x,y) -> Int(ID.cast_to ~kind:Internal IInt (apply_binary FDouble FD.unordered x y)) (* TODO: proper castkind *)
+          | Isgreater (x,y) -> Int(fd_binary_pred (apply_binary FDouble FD.gt x y))
+          | Isgreaterequal (x,y) -> Int(fd_binary_pred (apply_binary FDouble FD.ge x y))
+          | Isless (x,y) -> Int(fd_binary_pred (apply_binary FDouble FD.lt x y))
+          | Islessequal (x,y) -> Int(fd_binary_pred (apply_binary FDouble FD.le x y))
+          | Islessgreater (x,y) -> Int(ID.c_logor (fd_binary_pred (apply_binary FDouble FD.lt x y)) (fd_binary_pred (apply_binary FDouble FD.gt x y)))
+          | Isunordered (x,y) -> Int(fd_binary_pred (apply_binary FDouble FD.unordered x y))
           | Fmax (fd, x ,y) -> Float (apply_binary fd FD.fmax x y)
           | Fmin (fd, x ,y) -> Float (apply_binary fd FD.fmin x y)
           | Sqrt (fk, x) -> Float (apply_unary fk FD.sqrt x)
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -890,6 +890,6 @@ struct
   let invariant man st exp tv =
     (* The computations that happen here are not computations that happen in the programs *)
     (* Any overflow during the forward evaluation will already have been flagged here *)
-    GobRef.wrap AnalysisState.executing_speculative_computations true
-    @@ fun () -> invariant man st exp tv
+    let@ () = GobRef.wrap AnalysisState.executing_speculative_computations true in
+    invariant man st exp tv
 end
--- a/src/analyses/basePriv.ml
+++ b/src/analyses/basePriv.ml
@@ -2031,7 +2031,7 @@ struct
     v
 
   let dump () =
-    Out_channel.with_open_bin (get_string "exp.priv-prec-dump") @@ fun f ->
+    let@ f = Out_channel.with_open_bin (get_string "exp.priv-prec-dump") in
     (* LVH.iter (fun (l, x) v ->
         Logs.debug "%a %a = %a" CilType.Location.pretty l CilType.Varinfo.pretty x VD.pretty v
       ) lvh; *)
--- a/src/autoTune.ml
+++ b/src/autoTune.ml
@@ -62,7 +62,7 @@ class findAllocsInLoops = object
 
   method! vinst = function
     | Call (_, Lval (Var f, NoOffset), args,_,_) when LibraryFunctions.is_special f ->
-      Goblint_backtrace.protect ~mark:(fun () -> Cilfacade.FunVarinfo f) ~finally:Fun.id @@ fun () ->
+      let@ () = Goblint_backtrace.protect ~mark:(fun () -> Cilfacade.FunVarinfo f) ~finally:Fun.id in
       let desc = LibraryFunctions.find f in
       begin match desc.special args with
         | Malloc _
@@ -96,7 +96,7 @@ let functionArgs fd = (ResettableLazy.force functionCallMaps).argLists |> Functi
 
 let findMallocWrappers () =
   let isMalloc f =
-    Goblint_backtrace.wrap_val ~mark:(Cilfacade.FunVarinfo f) @@ fun () ->
+    let@ () = Goblint_backtrace.wrap_val ~mark:(Cilfacade.FunVarinfo f) in
     if LibraryFunctions.is_special f then
       let desc = LibraryFunctions.find f in
       let args = functionArgs f in
@@ -177,7 +177,7 @@ let disableIntervalContextsInRecursiveFunctions () =
 
 let hasFunction pred =
   let relevant_static var =
-    Goblint_backtrace.wrap_val ~mark:(Cilfacade.FunVarinfo var) @@ fun () ->
+    let@ () = Goblint_backtrace.wrap_val ~mark:(Cilfacade.FunVarinfo var) in
     if LibraryFunctions.is_special var then
       let desc = LibraryFunctions.find var in
       GobOption.exists (fun args -> pred desc args) (functionArgs var)
@@ -185,7 +185,7 @@ let hasFunction pred =
       false
   in
   let relevant_dynamic var =
-    Goblint_backtrace.wrap_val ~mark:(Cilfacade.FunVarinfo var) @@ fun () ->
+    let@ () = Goblint_backtrace.wrap_val ~mark:(Cilfacade.FunVarinfo var) in
     if LibraryFunctions.is_special var then
       let desc = LibraryFunctions.find var in
       (* We don't really have arguments at hand, so we cheat and just feed it a list of MyCFG.unknown_exp of appropriate length *)
--- a/src/cdomain/value/cdomains/floatDomain.ml
+++ b/src/cdomain/value/cdomains/floatDomain.ml
@@ -69,31 +69,31 @@ module type FloatArith = sig
   (** (inv_fabs z -> x) if (z = fabs(x)) *)
 
   (** {b Comparison operators} *)
-  val lt : t -> t -> IntDomain.IntDomTuple.t
+  val lt : t -> t -> bool option
   (** Less than: [x < y] *)
-  val gt : t -> t -> IntDomain.IntDomTuple.t
+  val gt : t -> t -> bool option
   (** Greater than: [x > y] *)
-  val le : t -> t -> IntDomain.IntDomTuple.t
+  val le : t -> t -> bool option
   (** Less than or equal: [x <= y] *)
-  val ge : t -> t -> IntDomain.IntDomTuple.t
+  val ge : t -> t -> bool option
   (** Greater than or equal: [x >= y] *)
-  val eq : t -> t -> IntDomain.IntDomTuple.t
+  val eq : t -> t -> bool option
   (** Equal to: [x == y] *)
-  val ne : t -> t -> IntDomain.IntDomTuple.t
+  val ne : t -> t -> bool option
   (** Not equal to: [x != y] *)
-  val unordered: t -> t -> IntDomain.IntDomTuple.t
+  val unordered: t -> t -> bool option
   (** Unordered *)
 
   (** {unary functions returning int} *)
-  val isfinite : t -> IntDomain.IntDomTuple.t
+  val isfinite : t -> bool option
   (** __builtin_isfinite(x) *)
-  val isinf : t -> IntDomain.IntDomTuple.t
+  val isinf : t -> bool option
   (** __builtin_isinf(x) *)
-  val isnan : t -> IntDomain.IntDomTuple.t
+  val isnan : t -> bool option
   (** __builtin_isnan(x) *)
-  val isnormal : t -> IntDomain.IntDomTuple.t
+  val isnormal : t -> bool option
   (** __builtin_isnormal(x) *)
-  val signbit : t -> IntDomain.IntDomTuple.t
+  val signbit : t -> bool option
   (** __builtin_signbit(x) *)
 end
 
@@ -369,34 +369,32 @@ module FloatIntervalImpl(Float_t : CFloatType) = struct
 
   let eval_comparison_binop min max reflexive eval_operation (op1: t) op2 =
     warn_on_specials_comparison op1 op2;
-    let a, b =
-      match (op1, op2) with
-      | Bot, _ | _, Bot -> raise (ArithmeticOnFloatBot (Printf.sprintf "%s op %s" (show op1) (show op2)))
-      | Interval v1, Interval v2 -> eval_operation v1 v2
-      | NaN, _ | _, NaN -> (0,0) (* comparisons involving NaN always return false *)
-      | Top, _ | _, Top -> (0,1) (* comparisons with Top yield top *)
-      (* neither of the arguments below is Top/Bot/NaN *)
-      | v1, v2 when v1 = min ->
-        (* v1 is the minimal element w.r.t. the order *)
-        if v2 <> min || reflexive then
-          (* v2 is different, i.e., greater or the relation is reflexive *)
-          (1,1)
-        else
-          (0,0)
-      | _, v2 when v2 = min ->
-        (* second argument is minimal, first argument cannot be *)
-        (0,0)
-      | v1, v2 when v1 = max ->
-        (* v1 is maximal element w.r.t. the order *)
-        if v2 = max && reflexive then
-          (* v2 is also maximal and the relation is reflexive *)
-          (1,1)
-        else
-          (0,0)
-      | _, v2 when v2 = max -> (1,1) (* first argument cannot be max *)
-      | _ -> (0, 1)
-    in
-    IntDomain.IntDomTuple.of_interval IBool (Z.of_int a, Z.of_int b)
+    (* TODO: don't use polymorphic = and <> here *)
+    match op1, op2 with
+    | Bot, _ | _, Bot -> raise (ArithmeticOnFloatBot (Printf.sprintf "%s op %s" (show op1) (show op2)))
+    | Interval v1, Interval v2 -> eval_operation v1 v2
+    | NaN, _ | _, NaN -> Some false (* comparisons involving NaN always return false *)
+    | Top, _ | _, Top -> None (* comparisons with Top yield top *)
+    (* neither of the arguments below is Top/Bot/NaN *)
+    | v1, v2 when v1 = min ->
+      (* v1 is the minimal element w.r.t. the order *)
+      if v2 <> min || reflexive then
+        (* v2 is different, i.e., greater or the relation is reflexive *)
+        Some true
+      else
+        Some false
+    | _, v2 when v2 = min ->
+      (* second argument is minimal, first argument cannot be *)
+      Some false
+    | v1, v2 when v1 = max ->
+      (* v1 is maximal element w.r.t. the order *)
+      if v2 = max && reflexive then
+        (* v2 is also maximal and the relation is reflexive *)
+        Some true
+      else
+        Some false
+    | _, v2 when v2 = max -> Some true (* first argument cannot be max *)
+    | _ -> None
 
   let eval_neg = function
     | (low, high) -> Interval (Float_t.neg high, Float_t.neg low)
@@ -440,35 +438,37 @@ module FloatIntervalImpl(Float_t : CFloatType) = struct
       Interval (low, high)
 
 
+  (* TODO: don't use polymorphic comparisons here *)
+
   let eval_lt (l1, h1) (l2, h2) =
-    if h1 < l2 then (1, 1)
-    else if l1 >= h2 then (0, 0)
-    else (0, 1)
+    if h1 < l2 then Some true
+    else if l1 >= h2 then Some false
+    else None
 
   let eval_gt (l1, h1) (l2, h2) =
-    if l1 > h2 then (1, 1)
-    else if h1 <= l2 then (0, 0)
-    else (0, 1)
+    if l1 > h2 then Some true
+    else if h1 <= l2 then Some false
+    else None
 
   let eval_le (l1, h1) (l2, h2) =
-    if h1 <= l2 then (1, 1)
-    else if l1 > h2 then (0, 0)
-    else (0, 1)
+    if h1 <= l2 then Some true
+    else if l1 > h2 then Some false
+    else None
 
   let eval_ge (l1, h1) (l2, h2) =
-    if l1 >= h2 then (1, 1)
-    else if h1 < l2 then (0, 0)
-    else (0, 1)
+    if l1 >= h2 then Some true
+    else if h1 < l2 then Some false
+    else None
 
   let eval_eq (l1, h1) (l2, h2) =
-    if h1 < l2 || h2 < l1 then (0, 0)
-    else if h1 = l1 && h2 = l2 && l1 = l2 then (1, 1)
-    else (0, 1)
+    if h1 < l2 || h2 < l1 then Some false
+    else if h1 = l1 && h2 = l2 && l1 = l2 then Some true
+    else None
 
   let eval_ne (l1, h1) (l2, h2) =
-    if h1 < l2 || h2 < l1 then (1, 1)
-    else if h1 = l1 && h2 = l2 && l1 = l2 then (0, 0)
-    else (0, 1)
+    if h1 < l2 || h2 < l1 then Some true
+    else if h1 = l1 && h2 = l2 && l1 = l2 then Some false
+    else None
 
   let eval_fmax (l1, h1) (l2, h2) = (max l1 l2, max h1 h2)
   let eval_fmin (l1, h1) (l2, h2) = (min l1 l2, min h2 h2)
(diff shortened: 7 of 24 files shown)


src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,217 @@
...
+    let must_joined_tids = ask.f Queries.MustJoinedThreads in
+    TIDs.diff may_transitively_created_tids must_joined_tids
+  ;;
+
+  (** stolen from mutexGhost.ml. TODO Maybe add to utils? *)
+  let mustlock_of_addr (addr : LockDomain.Addr.t) : LID.t option =
+    match addr with
+    | Addr mv when LockDomain.Mval.is_definite mv -> Some (LID.of_mval mv)
+    | _ -> None
+  ;;
```
dabund24: I copied this from the mutex ghost analysis. Is this a reasonable way to get the lock of an unlock event if it is known and if so, would it make sense to expose this function globally at some other place?

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,217 @@
...
+    let tid_lifted = ask.f Queries.CurrentThreadId in
+    let child_ask = Analyses.ask_of_man fman in
+    let child_tid_lifted = child_ask.f Queries.CurrentThreadId in
+    match tid_lifted, child_tid_lifted with
+    | `Lifted tid, `Lifted child_tid ->
+      let descendants = descendants_closure child_ask child_tid in
+      let lockset = ask.f Queries.MustLockset in
+      let to_contribute = cartesian_prod (TIDs.singleton tid) lockset in
+      TIDs.iter (contribute_lock man to_contribute) descendants
+    | _ -> (* TODO deal with top or bottom? *) ()
```
dabund24: What does it mean for a thread id to be top or bottom? Not sure how to deal with this here and in some other places

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,217 @@
...
+open Analyses
+module LF = LibraryFunctions
+module TID = ThreadIdDomain.Thread
+module TIDs = ConcDomain.ThreadSet
+module LID = LockDomain.MustLock
+module LIDs = LockDomain.MustLockset
+
+(** common base for [CreationLocksetSpec] and [TaintedCreationLocksetSpec] *)
+module AncestorLocksetSpec = struct
+  include IdentityUnitContextsSpec (* no context necessary(?) *)
```
dabund24: Is this fine for now or do you have ideas on how the analyses could be improved with contexts? I thought that maybe having (regular) locksets may be beneficial in cases where create statements happen in a function, which is called multiple times with different locksets 🤔

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,217 @@
...
+  module D = Lattice.Unit
+
+  module V = struct
+    include TID
+    include StdV
+  end
+
+  (** 2 ^ { [TID] \times [LID] } *)
+  module G = Queries.ALS
```
dabund24: defining the domain here instead of in queries would introduce a dependency conflict. Is this fine?

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,217 @@
...
+    ;;
+
+    let may_race (t1, ls1, itls1) (t2, ls2, itls2) =
+      not
+        (both_protected_inter_threaded itls1 itls2
+         || one_protected_inter_threaded_other_intra_threaded itls1 t2 ls2
+         || one_protected_inter_threaded_other_intra_threaded itls2 t1 ls1)
+    ;;
+
+    let should_print _ = true
```
dabund24: are there some guidelines for when this should be true?

src/domains/queries.ml
```diff
@@ -472,6 +481,8 @@ struct
     | Any (MaySignedOverflow e) -> CilType.Exp.hash e
     | Any (GasExhausted f) -> CilType.Fundec.hash f
     | Any (GhostVarAvailable v) -> WitnessGhostVar.hash v
+    | Any (DescendantThreads t) -> ThreadIdDomain.hash_thread t (* TODO is this fine? *)
```
dabund24: does `ThreadIdDomain.hash_thread` actually do what I think it does/is this the right hash call for this place?

dabund24 (commented):
Some general questions:
- Is there a way to enforce configuration settings to be set to certain values in order for the analysis to work (i.e. similar to how analyses can be declared as dependencies)? This would be necessary for the settings revolving around thread ids as described in the pr summary. Checking settings using `GobConfig.get_string` at the start of transfer functions doesn't really seem ideal to me
- is there an ideal amount of tests I should write or can I write as many as I feel to be necessary?

src/analyses/creationLockset.ml
sim642: It should be `true` when the argument contains anything that can possibly rule out a race (make `may_race` return `false`).

It can be `false` when there's nothing that can rule out a race. I haven't looked at the details of this analysis, but the normal mutex analysis returns `false` when the lockset is empty. Maybe there's something similar here?

src/analyses/creationLockset.ml
sim642: It should be fine because it's just a one-liner in `Queries`.
Theoretically it could be in a separate domain module file, but that's overkill for a single line.

src/analyses/creationLockset.ml
sim642: These analyses seem to have `module D = Lattice.Unit`, so I'm not sure what even could go into contexts.
Maybe it doesn't make sense for these to be context-sensitive but rather the other analyses whose data is being queried.

src/domains/queries.ml
sim642: `ThreadIdDomain.Thread.hash` would be more consistent with the type used in the definition of `DescendantThreads`, but it's the same function under the hood.

sim642:
> * Is there a way to enforce configuration settings to be set to certain values in order for the analysis to work (i.e. similar to how analyses can be declared as dependencies)? This would be necessary for the settings revolving around thread ids as described in the pr summary. Checking settings using `GobConfig.get_string` at the start of transfer functions doesn't really seem ideal to me

`Maingoblint.check_arguments` has a bunch of ad-hoc checks like this.

src/analyses/creationLockset.ml
sim642: I suppose it could be moved to `LockDomain.MustLock.of_addr`. the `of_mval` function there currently implicitly assumes `is_definite` anyway and that's checked before every usage.

src/analyses/creationLockset.ml
dabund24: I see, then this would be out of scope for this pr I guess. Removed the comment in https://github.com/goblint/analyzer/pull/1865/commits/dbf73f10374987c11e73b75ceffab150d442200f

src/domains/queries.ml
dabund24: addressed in https://github.com/goblint/analyzer/pull/1865/commits/e1299eabe7bdc62918dc7e6fcc535134d9ad8b92

src/analyses/creationLockset.ml
dabund24: I'm liking this solution :D
Changed in https://github.com/goblint/analyzer/pull/1865/commits/87dde0543230507166bcd47ac1514a0d1090f022

src/analyses/creationLockset.ml
dabund24: In this case, the analogous coundition would be $\mathcal{IL}\ne\emptyset$. Updated that in https://github.com/goblint/analyzer/pull/1865/commits/b8416efeeffa72a5312b789652cbff17168dc28a

src/analyses/creationLockset.ml
michael-schwarz: It should usually not happen. If it does, I think the best is to assume that all bets are off.

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,210 @@
...
+      @param man man at program point
+      @param to_contribute new edges from [child_tid] to register
+      @param child_tid
+  *)
+  let contribute_lock man to_contribute child_tid = man.sideg child_tid to_contribute
+
+  (** compute [tids] \times \{[lock]\} *)
+  let singleton_cartesian_prod tids lock =
+    TIDs.fold (fun tid acc -> G.add (tid, lock) acc) tids (G.empty ())
+  ;;
```
michael-schwarz:
```suggestion
  
```

I know we sometimes show the `;;` in FPV, but they are not needed. (Also in a few other places).

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,210 @@
+(** Ancestor-lock analysis. See https://github.com/goblint/analyzer/pull/1865 *)
+
+open Analyses
+module LF = LibraryFunctions
+module TID = ThreadIdDomain.Thread
+module TIDs = ConcDomain.ThreadSet
```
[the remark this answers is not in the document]
sim642: Don't commit this directly though, it duplicates the `TIDs` definition...

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,212 @@
...
+      let ask = Analyses.ask_of_man man in
+      let tid_lifted = ask.f Queries.CurrentThreadId in
+      (match tid_lifted with
+       | `Top | `Bot -> ()
+       | `Lifted tid ->
+         let possibly_running_tids = get_possibly_running_tids ask in
+         let lock_opt = LockDomain.MustLock.of_addr addr in
+         (match lock_opt with
+          | Some lock ->
+            let _ = M.tracel "lol" "%a" LID.pretty lock in
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/6c2c849eca494b3b0d3a1d818693794846169de7

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,212 @@
...
+            (* contribute for all possibly_running_tids: (tid, lock) *)
+            let to_contribute = G.singleton (tid, lock) in
+            TIDs.iter (contribute_lock man to_contribute) possibly_running_tids
+          | None ->
+            let _ = M.tracel "lol" "idk" in
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/6c2c849eca494b3b0d3a1d818693794846169de7

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,212 @@
...
+            (* any lock could have been unlocked. Contribute for all possibly_running_tids all members of their CreationLocksets with the ego thread to invalidate them!! *)
+            let contribute_creation_lockset des_tid =
+              let _ = M.tracel "lol" "des tid %a" TID.pretty des_tid in
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/6c2c849eca494b3b0d3a1d818693794846169de7

src/analyses/creationLockset.ml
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/695081c8299f3d87b03eb5051597befdac649713

src/domains/queries.ml
```diff
@@ -540,6 +551,8 @@ struct
     | Any (GasExhausted f) -> Pretty.dprintf "GasExhausted %a" CilType.Fundec.pretty f
     | Any (GhostVarAvailable v) -> Pretty.dprintf "GhostVarAvailable %a" WitnessGhostVar.pretty v
     | Any InvariantGlobalNodes -> Pretty.dprintf "InvariantGlobalNodes"
+    | Any (DescendantThreads t) -> Pretty.dprintf "DescendantThreads"
+    | Any (MayCreationLockset t) -> Pretty.dprintf "MayCreationLockset"
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/5bf35caae0996d42cd187593f1b32f0b7b4f3633

tests/regression/53-races-mhp/12-lockset_inter_threaded_lock_transitive_racefree.c
```diff
@@ -0,0 +1,32 @@
...
+
+void *t1(void *arg) {
+  pthread_mutex_lock(&mutex);
+  global++; // NORACE
+  pthread_mutex_unlock(&mutex);
+  return NULL;
+}
+
+void* t3(void* arg) {
+  global++; // NORACE
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/8d5928e993e320de26669487b9780b45ec3cdf88

src/analyses/creationLockset.ml
dabund24: Would you suggest emitting a warning or something similar here or can I keep everything the way it is? (except for removing the comments of course)

src/analyses/creationLockset.ml
dabund24: addressed in https://github.com/goblint/analyzer/pull/1865/commits/873ff0c7692cc1f23fb84efd02e4bf189c7e4208 and https://github.com/goblint/analyzer/pull/1865/commits/3c45dff510193de6e9e0d00c6cfe3a1df83ee9a3.
The commits are a little chaotic, sorry for that.
And I'm using `ocamlformat`, which put the `;;` there, so those are not your fault :D

michael-schwarz: Ah, I didn't know ocamlformat adds these, normally its defaults are sensible.

dabund24: By default, it actually doesn't add them, so you can keep your positive view on ocamlformat :)
However, I was using the `janestreet` profile, since the code formatted with the default profile doesn't pass the pre-commit hook

michael-schwarz:
> If you want to review it already, you can do so. 

I probably won't get around to it until some time next week, but I added it to my TODO (list / stack / multiset).

src/cdomain/value/cdomains/threadIdDomain.ml
```diff
@@ -17,6 +17,9 @@ sig
 
   (** Is the first TID a must ancestor of the second thread. Always false if the first TID is not unique *)
   val must_be_ancestor: t -> t -> bool
+
+  (** Every TID of the returned list must be an ancestor of the passed TID. Return None if the result cannot be determined *)
+  val must_ancestors: t -> t list
```
dabund24: > Such a function must_ancestors: `TID -> TID list option` could then be added to the generic thread id interface, and just return `None`, i.e., "information not available" for the thread ids which don't allow this trick.

I omitted the `option` for the return type, since the result of the function is an approximation anyway and other functions like `must_be_ancestor` also don't return options, so I don't see a case where `None` and `[]` would be handled differently. Is this fine or is there a reason for the option in your suggestion?

michael-schwarz: Does an **underapproximation** of the must ancestors suffice for your setting? I thought you need to check that none of the must-ancestor unlocked it, so you would need **all** must-ancestors?

If no, then this is fine, but you should document it to clearly say that it is an **underapproximation** of must-ancestors.

michael-schwarz: This also affects the way you construct them, currently you're constructing an underapproximation (by not giving up in the case the set is non-empty).

src/cdomain/value/cdomains/threadIdDomain.ml
```diff
@@ -169,6 +173,16 @@ struct
         false (* composing cannot fix incompatibility there *)
     )
 
+  let must_ancestors (p, s) =
+    let rec build_ancestors acc p =
+      match p with
+      | [] -> acc
+      | h :: t -> build_ancestors ((h :: t, S.empty ()) :: acc) t
```
michael-schwarz:
```suggestion
    let rec build_ancestors acc = function
      | [] -> acc
      | (h :: t as current) -> build_ancestors ((current, S.empty ()) :: acc) t
```

Naming it may make the intention clearer.

src/analyses/creationLocksetAlternative.ml
```diff
@@ -0,0 +1,210 @@
...
+          | Some lock -> unlock man tid created_tids joined_tids lock
+          | None -> unknown_unlock man tid created_tids joined_tids)
+       | _ -> ())
+    | _ -> ()
+
+  module A = struct
+    (** ego tid * must-lockset * creation-lockset *)
+    include Printable.Prod3 (TID) (LIDs) (Queries.CLS)
+
+    let name () = "creationLockset"
```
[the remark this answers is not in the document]
michael-schwarz: Obviously, only one will remain long-term.

src/analyses/threadDescendants.ml
```diff
@@ -0,0 +1,40 @@
...
+
+  let threadspawn man ~multiple lval f args fman =
+    let ask = ask_of_man man in
+    let tid_lifted = ask.f Queries.CurrentThreadId in
+    let child_ask = ask_of_man fman in
+    let child_tid_lifted = child_ask.f Queries.CurrentThreadId in
+    match tid_lifted, child_tid_lifted with
+    | `Lifted tid, `Lifted child_tid ->
+      let to_contribute = G.add child_tid (man.global child_tid) in
+      man.sideg tid to_contribute
```
michael-schwarz: What are upsides/downsides of closing this under the threads created by the child here vs. doing this closure at the point where you do the `Queries.DescendantThreads` query?

Not sure if there are any tbh, as both happen during the analysis and not only in the post-processing. Just wanna make sure we think about it.

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,163 @@
...
+    (t_1) ---L--> (t_0) means that t_1 is protected by all members of L from t_0
+
+    @see https://github.com/goblint/analyzer/pull/1865
+*)
+
+open Analyses
+module TID = ThreadIdDomain.Thread
+module TIDs = ConcDomain.ThreadSet
+module LID = LockDomain.MustLock
+module LIDs = LockDomain.MustLockset
```
michael-schwarz: Why `LID` and `LIDs`? And not `Lock` and  `Lockset`

src/analyses/creationLocksetAlternative.ml
```diff
@@ -0,0 +1,210 @@
+(** alternative creation lockset analysis
```
michael-schwarz: At least looking at LOCs, this one does not seem much more complicated than the first option.

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,163 @@
...
+        such that [ls1] and [ls2] are not disjoint and [tp1] != [tp2]
+        @param cl1 creation-lockset of first thread [t1]
+        @param cl2 creation-lockset of second thread [t2]
+        @returns whether [t1] and [t2] must be running mutually exclusive
+    *)
+    let both_protected_inter_threaded cl1 cl2 =
+      let cl2_has_same_lock_other_tid tp1 ls1 =
+        G.exists (fun tp2 ls2 -> not (LIDs.disjoint ls1 ls2 || TID.equal tp1 tp2)) cl2
+      in
+      G.exists cl2_has_same_lock_other_tid cl1
```
michael-schwarz: Nice, this was a case I hadn't even considered!

src/analyses/creationLocksetAlternative.ml
```diff
@@ -0,0 +1,210 @@
...
+    let ask = ask_of_man man in
+    let tid_lifted = ask.f Queries.CurrentThreadId in
+    let child_ask = ask_of_man fman in
+    let child_tid_lifted = child_ask.f Queries.CurrentThreadId in
+    match tid_lifted, child_tid_lifted with
+    | `Lifted tid, `Lifted child_tid when TID.must_be_ancestor tid child_tid ->
+      let lockset = ask.f Queries.MustLockset in
+      let to_contribute = G.singleton tid lockset in
+      man.sideg child_tid to_contribute
+    | _ -> ()
```
michael-schwarz: Does this not always collapse to just be a singleton? There is exactly one must ancestor?

michael-schwarz: Then, `G` can maybe just be a must lockset. Or am I missing something?

src/analyses/creationLocksetAlternative.ml
```diff
@@ -0,0 +1,210 @@
...
+    let may_race (t1, ls1, il1) (t2, ls2, il2) =
+      not
+        (both_protected_inter_threaded il1 il2
+         || one_protected_inter_threaded_other_intra_threaded il1 t2 ls2
+         || one_protected_inter_threaded_other_intra_threaded il2 t1 ls1)
+
+    let should_print (_t, _ls, cl) = not @@ Queries.CLS.is_empty cl
+  end
+
+  let access man _ =
```
michael-schwarz: @sim642: When is this called? Can we somehow ensure all this stuff happens only when postsolving?

src/analyses/creationLocksetAlternative.ml
```diff
@@ -0,0 +1,210 @@
...
+         let created_tids = get_unique_created_children tid ask in
+         let joined_tids = ask.f Queries.MustJoinedThreads in
+         let lock_opt = LockDomain.MustLock.of_addr addr in
+         (match lock_opt with
+          | Some lock -> unlock man tid created_tids joined_tids lock
+          | None -> unknown_unlock man tid created_tids joined_tids)
+       | _ -> ())
+    | _ -> ()
+
+  module A = struct
```
michael-schwarz: I assume this identical to this module in the other option? If not, let me know and I'll take another look.

src/analyses/creationLocksetAlternative.ml
```diff
@@ -0,0 +1,210 @@
...
+    let created_threads = ask.f Queries.CreatedThreads in
+    TIDs.filter (TID.must_be_ancestor tid) created_threads
+
+  (** handle unlock of mutex [lock] *)
+  let unlock man tid created_tids joined_tids lock =
+    let contribute_lock child_tid =
+      let to_contribute = G.singleton tid (LockToThreads.singleton lock joined_tids) in
+      man.sideg child_tid to_contribute
+    in
+    TIDs.iter contribute_lock created_tids
```
michael-schwarz: Why does this need to be propagated to the children? Can I not just collect this map at the unknown for the current thread id, and then check this when I come up the chain later anyway?

src/analyses/creationLocksetAlternative.ml
```diff
@@ -0,0 +1,210 @@
...
+        (both_protected_inter_threaded il1 il2
+         || one_protected_inter_threaded_other_intra_threaded il1 t2 ls2
+         || one_protected_inter_threaded_other_intra_threaded il2 t1 ls1)
+
+    let should_print (_t, _ls, cl) = not @@ Queries.CLS.is_empty cl
+  end
+
+  let access man _ =
+    let ask = Analyses.ask_of_man man in
+    let tid_lifted = ask.f Queries.CurrentThreadId in
```
michael-schwarz: Haven't considered the rest of this yet, it seems that if the remark on L78 is realized this may change?

src/analyses/creationLocksetAlternative.ml
michael-schwarz: Or maybe I'm just confused (very possible!)

michael-schwarz (commented):
Nice, this is really impressive! The code quality is really high, there's not many of the usual cheap "please format like this or that"-type comments to make here.

I have some conceptual questions about how the alternative approach can be simplified to do even less side-effects and reconstruct even more things on the fly, but I'm not sure if it actually works like that, so please consider it critically.

(**Disclaimer on the scope:** Have not looked at the tests due to lack of time, and also did not look at `access` in the alternative yet in detail, as I suspect it might still change.)

src/analyses/creationLocksetAlternative.ml
dabund24: That is true 👍

src/analyses/creationLockset.ml
dabund24: We save a few chars :3

Changed this in https://github.com/goblint/analyzer/pull/1865/commits/71e0299db3c5fbddb383901df6040b36493eee47

src/analyses/creationLocksetAlternative.ml
dabund24: True, I was surprised by this ngl. Using maps instead of sets with a custom domain took away a good part of the complexity I initially anticipated

src/cdomain/value/cdomains/threadIdDomain.ml
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/6e31237a11b2308c3aa01463e0f985d0a64d70ad

src/analyses/creationLocksetAlternative.ml
sim642: This should be called by the race analysis when reporting races at the end, so it should be enough. Although I'd have to carefully check when this does get called: could be that the access-related data is computed always, but just the warnings are not collected during analysis.

Is this very expensive or why are you asking?

src/cdomain/value/cdomains/threadIdDomain.ml
dabund24: > Does an underapproximation of the must ancestors suffice for your setting? I thought you need to check that none of the must-ancestor unlocked it, so you would need all must-ancestors?

If `must_be_ancestor` works the way I think it does, then I believe this to work. However, a hidden assumption for this is needed, which I had overlooked.

It should work, since we do not need to check that none of the must-ancestor unlocked the mutex, but rather that none of the must-ancestors *from the `creationLockset`* unlocked the mutex. In order to be in the `creationLockset`, `must_be_ancestor` needed to have verified that the creation happened indeed by a must-ancestor.

So the problematic case would be, if `must_be_ancestor` $t_0$ $t_1$ returns `true` and $t_0\notin$ `must_ancestors` $t_1$, which cannot happen if I'm not mistaken? Thus, the assumption is
`must_be_ancestor` $t_0$ $t_1 \implies t_0\in$ `must_ancestors` $t_1$

I guess we could go many ways here:
1. document this assumption in the code (maybe even with an $\iff$) and leave the analysis untouched
2. replace `must_be_ancestor` $t_0$ $t_1$ with $t_0\in$ `must_ancestors` $t_1$ in the `creationLockset` analysis
3. turn the return type of `must_ancestors` into an option again and handle `None` in the analysis

Whilst 1. and 2. could be considered as somewhat dirty (not sure to which extent, though), Option 3 comes with a precision penalty, so we need to pick our poison here. What are your thoughts on that matter?

src/analyses/creationLocksetAlternative.ml
dabund24: True, I hadn't realized that, so implementing `G` as a must-lockset should be viable.
To be precise, there is one nearest must-ancestor, which we would need to access. We could access it by either adding one more function `nearest_must_ancestor` to `threadIdDomain` or documenting the order of the return value of `must_ancestors`.

I guess this may depend on our decision for https://github.com/goblint/analyzer/pull/1865#discussion_r2620776329

src/analyses/threadDescendants.ml
dabund24: The only "clear" downside of the current approach I can think of is the fact that we wouldn't be able to query all child threads if we needed that for future analyses.

Other than that, to me this seems to be a tradeoff between memory usage and performance, where one option re-uses the stored closures of its children whereas the other one needs to compute its own closure from scratch every time (however, only if it is really needed).

I'm wondering if implementing the closure in `query` by also querying the closures of the children (as opposed to using `global`) would be a good middle ground, as I saw that queries also cache stuff.

src/analyses/creationLocksetAlternative.ml
dabund24: > Can I not just collect this map at the unknown for the current thread id, and then check this when I come up the chain later anyway

Not sure if I understand this correctly, but do you suggest storing the maps the other way round (parent to children instead of child to parents)?

We could definitely try that if that could improve the performance (and this *could* even be beneficial for the implementation of the solution for second case from #1805, but I'm not 100% sure about that, yet)

michael-schwarz: > Not sure if I understand this correctly, but do you suggest storing the maps the other way round (parent to children instead of child to parents)?

No, I was wondering if I need to store a list of ancestors instead of only storing the direct parent and then looking up the parent of that at the appropriate unknown (and so on)...

src/analyses/creationLocksetAlternative.ml
michael-schwarz: It could potentially query lots of unknowns that only need to be queried during postsolving (if my suggestions are implemented), and if this is computed all the time, moving the queries here does not reduce re-evaluations.

src/analyses/threadDescendants.ml
michael-schwarz: > I'm wondering if implementing the closure in query by also querying the closures of the children (as opposed to using global) would be a good middle ground, as I saw that queries also cache stuff.

That sounds like a good idea!

dabund24: Implemented this in https://github.com/goblint/analyzer/pull/1865/commits/315dceb3422e1ad268a1692360ac3f95d1c5fdf6.

Performance-wise, this had no negative consequences (on the benchmarks in the `pthread` folder) :D
(the second to last column is what we care about here)

# Closure computation in `threadspawn`
image

# Closure computation in `query`
image

src/analyses/creationLocksetAlternative.ml
dabund24: I'm still not _completely_ sure if I correctly got what you are suggesting, but here is how I understand it now with this example:
```mermaid
graph TB;
subgraph t1;
    E["..."];
end;
subgraph t0;
    A["lock(l);"]-->B;
    B["create(t1);"]-->C;
    C["unlock(l);"]
end;
B-.->E
```
With the current implementation, we contribute
$\mathcal{TCL}\ t_1\sqsupseteq\\{t_0\mapsto\\{l\mapsto\emptyset\\}\\}$
at `unlock(l)`.

> Can I not just collect this map at the unknown for the current thread id, and then check this when I come up the chain later anyway?

This sounds like we would contribute
$\mathcal{TCL}\ t_0\sqsupseteq \\{l\mapsto\emptyset\\}$
instead, which wouldn't work I think. (see the counter-example below, where we would register the same contribution despite $t_1$ being fully protected by $t_0$ with $l$)

```mermaid
graph TB;
subgraph t1;
    E["..."];
end;
subgraph t0;
    A["lock(l);"]-->B1;
    B1["unlock(l);"]-->B2;
    B2["lock(l);"]-->B3;
    B3["create(t1);"]-->C;
    C["return;"]
end;
B3-.->E
```


However, what we could do instead (for the first example) is
$\mathcal{TCL}\ \textcolor{red}{t_1}\sqsupseteq \\{l\mapsto\emptyset\\}$
analogous to your comment in https://github.com/goblint/analyzer/pull/1865#discussion_r2626409696

michael-schwarz: Ah, I see, you're correct my initial idea wouldnt' work.

src/cdomain/value/cdomains/threadIdDomain.ml
michael-schwarz: I think 2 plus a comment that this is like this on purpose might be the most reasonable compromise here.

dabund24: applied this in https://github.com/goblint/analyzer/pull/1865/commits/d71f08ec5028ec8853e80e491d833df225decf80 and https://github.com/goblint/analyzer/pull/1865/commits/3355c938f25ca7f50e0de02084c7209b08103b48
(and https://github.com/goblint/analyzer/pull/1865/commits/16d5318a6368aee8209e5a68a9eabda42fb56355; yikes)

src/analyses/creationLocksetAlternative.ml
dabund24: Opted for a `parent` function in https://github.com/goblint/analyzer/commit/bf0cf5d7ce7ba38fbd3c7f7d044366f60ff49abc. This makes it impossible for thread $\langle l,\emptyset\rangle$ to protect $\langle l,S\rangle$ with $S\ne\emptyset$, but I'm guessing those cases are kind of rare and implementing `parent` such that this case is taken into account would feel like an "over-specialization".

Adapted the analyses in https://github.com/goblint/analyzer/commit/7720f5c65abebcbc9e6b72b1f46335e32028bfba

src/analyses/creationLocksetAlternative.ml
dabund24: > However, what we could do instead (for the first example) is
$\mathcal{TCL}\ \textcolor{red}{t_1}\sqsupseteq \\{l\mapsto\emptyset\\}$
analogous to your comment in https://github.com/goblint/analyzer/pull/1865#discussion_r2626409696

Implemented this in https://github.com/goblint/analyzer/commit/7720f5c65abebcbc9e6b72b1f46335e32028bfba

tests/regression/90-races_ancestor_locksets/17-ambiguous_context_racing.c
```diff
@@ -0,0 +1,38 @@
+// PARAM: --set ana.activated[+] threadJoins --set ana.activated[+] threadDescendants --set ana.activated[+] creationLockset
+#include <pthread.h>
+
+int global = 0;
+pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
+pthread_t id1, id2;
+
+// both accesses are protected by same mutex **locked in same thread**
```
sim642: This comment seems to be copied and not matching the test case.

tests/regression/53-races-mhp/10-lockset_inter_threaded_lock_racefree.c
sim642, line 1: There's already a test group 53-races-mhp. I think these tests should be moved there.

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,163 @@
...
+module Spec = struct
+  include IdentityUnitContextsSpec
+  module D = Lattice.Unit
+
+  module V = struct
+    include TID
+    include StdV
+  end
+
+  module G = MapDomain.MapBot (TID) (Lockset)
```
sim642, line 23: The module comment only talks about a graph but there are no graphs directly in Goblint analyses. Here should probably be a comment about the graph representation.
Is it that `(t_1) ---L--> (t_0)` is represented such that `global t_1 t_0 = L`, or the other way around?

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,163 @@
...
+  let contribute_locks man to_contribute child_tid = man.sideg child_tid to_contribute
+
+  (** reflexive-transitive closure of child relation applied to [tid]
+      and filtered to only include threads, where [tid] is a must-ancestor
+      @param ask any ask
+      @param tid
+  *)
+  let unique_descendants_closure (ask : Queries.ask) tid =
+    let descendants = ask.f @@ Queries.DescendantThreads tid in
+    let unique_descendants = TIDs.filter (TID.must_be_ancestor tid) descendants in
```
sim642: `unique_descendants` sounds like it should be `TIDs.filter TID.is_unique descendants`. Maybe `must_ancestor_descendants` instead?

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,163 @@
...
+    TIDs.add tid unique_descendants
+
+  let threadspawn man ~multiple lval f args fman =
+    let ask = ask_of_man man in
+    let tid_lifted = ask.f Queries.CurrentThreadId in
```
sim642: The queries in this function (and in others below) are a bit roundabout: just `man.ask` works.

src/analyses/creationLockset.ml
```diff
@@ -0,0 +1,163 @@
...
+
+    (** checks if [cl1] has a mapping ([tp1] |-> [ls1])
+        such that [ls1] and [ls2] are not disjoint and [tp1] != [t2]
+        @param cl1 creation-lockset of thread [t1] at first program point
+        @param t2 thread id at second program point
+        @param ls2 lockset at second program point
+        @returns whether [t1] must be running mutually exclusive with second program point
+    *)
+    let one_protected_inter_threaded_other_intra_threaded cl1 t2 ls2 =
+      G.exists (fun tp1 ls1 -> not (Lockset.disjoint ls1 ls2 || TID.equal tp1 t2)) cl1
```
sim642, line 123: Isn't this the same as `cl2_has_same_lock_other_tid` (modulo renaming)?
Maybe this could be defined before and just reused in `both_protected_inter_threaded`.

src/analyses/threadDescendants.ml
```diff
@@ -0,0 +1,45 @@
...
+  let startstate _ = D.bot ()
+  let exitstate _ = D.bot ()
+
+  let query man (type a) (x : a Queries.t) : a Queries.result =
+    match x with
+    | Queries.DescendantThreads t ->
+      let children = man.global t in
+      let ask = ask_of_man man in
+      (G.fold
+         (fun e acc -> G.union (ask.f @@ Queries.DescendantThreads e) acc)
```
sim642: Also roundabout asks here, could just use `man.ask`.

src/analyses/creationLocksetAlternative.ml
sim642: Is this still relevant? There's no `creationLocksetAlternative.ml` in the PR anymore, but was the concerning version removed or did it replace `creationLockset`?

tests/regression/90-races_ancestor_locksets/17-ambiguous_context_racing.c
dabund24: removed it in https://github.com/goblint/analyzer/pull/1865/commits/80c2b9ad152882ec4650a2748a3f4313fe778de0

tests/regression/53-races-mhp/10-lockset_inter_threaded_lock_racefree.c
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/08012f48176b7d7ea9f8f3ef54fa777bd689239a

src/analyses/creationLockset.ml
dabund24: added an explanation for the representation in https://github.com/goblint/analyzer/pull/1865/commits/7d1c0e940837c62f3e63c83358a6077dfe9f65a2

src/analyses/creationLockset.ml
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/0e09637dd5af90e658bc3e9e6ce578c47d4034ab

src/analyses/creationLockset.ml
dabund24: that is a lot nicer. Changed in https://github.com/goblint/analyzer/pull/1865/commits/a07b1d42dd4b28404db85983fb724b65bcf882ee

src/analyses/threadDescendants.ml
dabund24: fixed in https://github.com/goblint/analyzer/pull/1865/commits/a07b1d42dd4b28404db85983fb724b65bcf882ee

src/analyses/creationLockset.ml
dabund24: pretty cool, hadn't noticed this. Changed in https://github.com/goblint/analyzer/pull/1865/commits/90cd210f560d9cd73ceee57d820d95ba723cd11a

src/analyses/creationLocksetAlternative.ml
dabund24: The alternative version was indeed removed

src/analyses/threadDescendants.ml
sim642: That only fixed the instances in `creationLockset.ml`, not `threadDescendants.ml`.

sim642:
> I can go for the interactive rebase. Would it be enough if I remove all commits concerning the alternative version, which is completely removed now or should I also take care of commits fixing some minor mistakes I made in previous commits like [97a6967](https://github.com/goblint/analyzer/commit/97a69679af6ee6742c1be960bdd147a601bfb5fa) ?

That I would move to be a  `fixup` after the commit which introduced it, such that this change is completely cancelled from the history. Other minor reformatting or comment changes could be `fixup`-ed similarly.
Other more semantic bug fixes could remain as separate commits though, as indication for future git blamers to document that this has intentionally been changed from something incorrect.

But it depends on how deep you want/can be bothered to go with the history cleanup. With 121 commits, it could also quickly get very tricky because dropping an early commit ends up making a lot of the remaining patches not apply cleanly.
You could also do it in multiple passes: first drop all the alternative stuff and then interactive rebase again to clean the rest.

src/analyses/threadDescendants.ml
dabund24: yikes, fixed this in https://github.com/goblint/analyzer/pull/1865/commits/0325421bb791c1295dad4420bfbe6a2696fb7381

sim642:
Since this history contains some merges from `master`, the rebase also starts rebasing some merged things to linearize the history.
Using `--rebase-merges` should help, one just needs to not touch those commits that are on `master` to not try to rewrite them.

It might also make sense to rebase `--onto master` to get rid of the old merges entirely, but I don't know if this would make things easier or harder in this case.

sim642 (approved):
Looks fine by me as well, but @DrMichaelPetter should also have a look before we merge this.

PR #1879, opened 2025-11-21
Narrowing gas

### LINKED ISSUE #1494: Narrowing by fixed number of `meet`s
Based on the delayed widening from #1483, we could easily implement a domain lifter for narrowing that applies `meet` at most a fixed number of times. This makes it possible to have a terminating `narrow` for domains that do not have narrowings. It is a standard technique.

Our current analysis for the polyhedral example that #1484 is fixing is even more imprecise than the old literature suggests. This is because Apron doesn't provide narrowing for polyhedra and thus we cannot make any decreasing iterations. But the standard implementation of narrowing for polyhedra is to do a fixed number of `meet`s.

- [ ] Maybe this benefits #1102?

DrMichaelPetter:
Closes  #1494 by providing a domain lifter, that applies ```meet``` until the gas depletes. In a first application, we get a new parameter ```ana.apron.narrowing_gas``` to specify the amount of gas per state, which comes in handy for poyhedra-based analysis.

```c
// SKIP PARAM: --set ana.apron.narrowing_gas 1 --set ana.activated[+] apron --set ana.apron.domain polyhedra
int loop(){
  int i, j, k;
  int a = 0;
  for (i = 500; i >= 1; i--)
  {
    a++;
    __goblint_check(a + i - 501 == 0); // needs 1x narrowing or octagons
  }
  return 0;
}
```

### DIFF
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -15,7 +15,24 @@ let spec_module: (module MCPSpec) Lazy.t =
       let name () = "apron"
     end
     in
-    (module Spec)
+    let narrowing_gas = GobConfig.get_int "ana.apron.narrowing_gas" in
+    if (narrowing_gas > 0) then
+      let module Narrowed =
+      struct
+        module PolyhedraChainParams: Printable.ChainParams =
+        struct
+          let n () = narrowing_gas
+          let names = string_of_int
+        end
+        include NarrowingGas.DLifter (Spec) (PolyhedraChainParams)
+        module A = Spec.A
+        let access = Spec.access
+        let name () = "apron"
+      end
+      in
+      (module Narrowed)
+    else
+      (module Spec)
   )
 
 let get_spec (): (module MCPSpec) =
--- a/src/analyses/apron/apronAnalysis.apron.ml
+++ b/src/analyses/apron/apronAnalysis.apron.ml
@@ -15,7 +15,8 @@ let spec_module: (module MCPSpec) Lazy.t =
       let name () = "apron"
     end
     in
-    let narrowing_gas = GobConfig.get_int "ana.apron.narrowing_gas" in
+    let narrowing_gas = GobConfig.get_int "ana.apron.narrowing_gas"
+    in
     if (narrowing_gas > 0) then
       let module Narrowed =
       struct
@@ -27,7 +28,7 @@ let spec_module: (module MCPSpec) Lazy.t =
         include NarrowingGas.DLifter (Spec) (PolyhedraChainParams)
         module A = Spec.A
         let access = Spec.access
-        let name () = "apron"
+        let name = Spec.name
       end
       in
       (module Narrowed)
--- a/src/config/options.schema.json
+++ b/src/config/options.schema.json
@@ -932,6 +932,12 @@
               "enum": ["octagon", "interval", "polyhedra", "affeq"],
               "default": "octagon"
             },
+            "narrowing_gas": {
+              "title": "ana.apron.narrowing_gas",
+              "description": "Per state perform narrowing via meet as long as gas is left (0 to disable)",
+              "type": "integer",
+              "default": 0
+            },
             "threshold_widening": {
               "title": "ana.apron.threshold_widening",
               "description":
--- a/src/lifters/narrowingGas.ml
+++ b/src/lifters/narrowingGas.ml
@@ -40,7 +40,7 @@ struct
   let widen (b1, i1) (b2, i2) = (Base.widen b1 b2, max i1 i2) 
   let narrow (b1, i1) (b2, i2) = 
     let i' = max i1 i2 in
-    if i' <= ChainParams.n () then
+    if i' < ChainParams.n () then
       (Base.meet b1 b2, i' + 1)  (* Stop narrowing when counter exceeds limit. *)
     else
       (Base.narrow b1 b2, i')
--- a/src/lifters/narrowingGas.ml
+++ b/src/lifters/narrowingGas.ml
@@ -1,0 +1,96 @@
+(** narrowing delay with counters.
+
+    Abstract elements are paired with an integer counter, indicating how many times narrowing has been delayed.
+    Lifted abstract elements are only narrowed if the counter exceeds a predefined limit (gas). *)
+
+open Batteries
+open Lattice
+open Analyses
+
+module LocalChainParams: Printable.ChainParams =
+struct
+  let n () = GobConfig.get_int "ana.narrowing.gas"
+  let names = string_of_int
+end
+
+module GlobalChainParams: Printable.ChainParams =
+struct
+  let n () = GobConfig.get_int "ana.narrowing.gas"
+  let names = string_of_int
+end
+
+module Dom (Base: S) (ChainParams: Printable.ChainParams) =
+struct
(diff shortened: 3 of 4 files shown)


src/analyses/apron/apronAnalysis.apron.ml
```diff
@@ -15,7 +15,25 @@ let spec_module: (module MCPSpec) Lazy.t =
...
+        end
+        include NarrowingGas.DLifter (Spec) (PolyhedraChainParams)
+        module A = Spec.A
+        let access = Spec.access
+        let name = Spec.name
+      end
+      in
+      (module Narrowed)
+    else
+      (module Spec)
```
sim642, line 36: This is in a slightly less ideal place than I imagined in #1494, but I can see why and it is fine.

Ideally, I'd like to just wrap this lifting around polyhedra, but not other Apron domains, somewhere in `ApronDomain`, and have the analysis be completely oblivious to the fact.
Because in some sense it's "just" a lifting of any `Lattice.S → Lattice.S`. But that's only so if things are fully properly abstract, but our `RelationDomain`/`ApronDomain` setup is far from that. So it'd require a large cleanup/refactoring to be able to do it in my dream way.

DrMichaelPetter:
There is some strange influence between apron w. narrow gas 1 and def_exc going on, when the loopUnrollHeuristics is on:
```c
//PARAMS:  --set ana.apron.narrowing_gas 1 --set ana.activated[+] apron --set ana.apron.domain polyhedra --set ana.autotune.enabled true --set ana.autotune.activated '["loopUnrollHeuristic"]' 
int main() {
    int x;
    int y;
    int xtmp;
    if (y >= 0) 
    {
        xtmp = x;
        while(xtmp>y) {
            xtmp = xtmp - y;
        }
        y = xtmp;
    }
    return 0;
}
```

Goblint does not reach a fixpoint in this configuration, with a diff for 
```xtmp = (Unknown int([-31,31])) instead of (Not {0}([0,31]))``` at the inner loop guard.

**Comment by sim642** [comment]
With `--trace sol --trace sol2 --trace solchange` you can see the last iteration (immediately after the `narrow` by `meet`) of that unknown:
```
%%% sol: Var: L:node 27 "xtmp > y" on narrowing-gas-fp-error.c:9:15-9:21 (wp: true)
Old value: HConsed lifted PathSensitive (ProjectiveSet (MCP.D * map)):PathSensitive (ProjectiveSet (MCP.D * map)):{(MCP.D:[expRelation:(),
                                                                                                                           mallocWrapper:(wrapper call:Unknown node, unique calls:{}),
                                                                                                                           base:({
                                                                                                                                   Local {
                                                                                                                                     x ->   ⊤
                                                                                                                                     y ->   (Unknown int([0,31]))
                                                                                                                                     xtmp ->   (Not {0}([0,31]))
                                                                                                                                   }
                                                                                                                                 }, {}, {}, {}),
                                                                                                                           threadid:(wrapper call:unknown node, Thread:[main], created:(current function:bot, callees:bot)),
                                                                                                                           threadflag:Singlethreaded,
                                                                                                                           threadreturn:true,
                                                                                                                           escape:{},
                                                                                                                           mutexEvents:(),
                                                                                                                           access:(),
                                                                                                                           mutex:(lockset:{}, multiplicity:{}),
                                                                                                                           race:(),
                                                                                                                           mhp:(),
                                                                                                                           assert:(),
                                                                                                                           pthreadMutexType:(),
                                                                                                                           apron:(Apron Polyhedra * Unit:([|-x#297+2147483647>=0; -xtmp#299+2147483647>=0; xtmp#299-1>=0; xtmp#299+y#298-1>=0|] (env: [|0> x#297:int; 1> xtmp#299:int; 2> y#298:int|]), ()), chain:1)], map:{})}
Eqd: HConsed lifted PathSensitive (ProjectiveSet (MCP.D * map)):PathSensitive (ProjectiveSet (MCP.D * map)):{(MCP.D:[expRelation:(),
                                                                                                                     mallocWrapper:(wrapper call:Unknown node, unique calls:{}),
                                                                                                                     base:({
                                                                                                                             Local {
                                                                                                                               x ->   ⊤
                                                                                                                               y ->   (Unknown int([0,31]))
                                                                                                                               xtmp ->   ⊤
                                                                                                                             }
                                                                                                                           }, {}, {}, {}),
                                                                                                                     threadid:(wrapper call:unknown node, Thread:[main], created:(current function:bot, callees:bot)),
                                                                                                                     threadflag:Singlethreaded,
                                                                                                                     threadreturn:true,
                                                                                                                     escape:{},
                                                                                                                     mutexEvents:(),
                                                                                                                     access:(),
                                                                                                                     mutex:(lockset:{}, multiplicity:{}),
                                                                                                                     race:(),
                                                                                                                     mhp:(),
                                                                                                                     assert:(),
                                                                                                                     pthreadMutexType:(),
                                                                                                                     apron:(Apron Polyhedra * Unit:([|-x#297+2147483647>=0; -xtmp#299-y#298+2147483647>=0; -xtmp#299+2147483647>=0; xtmp#299-1>=0; xtmp#299+2y#298-1>=0|] (env: [|0> x#297:int; 1> xtmp#299:int; 2> y#298:int|]), ()), chain:0)], map:{})}
```

For some reason the right-hand side for the loop head is not monotonic? Or somehow with improved `apron` abstraction, there's less precise `xtmp` in `base` when reevaluated.

File: src/analyses/apron/apronAnalysis.apron.ml (hunk 1 above)
Comment (michael-schwarz) [reply src/analyses/apron/apronAnalysis.apron.ml:36]: I don't think that's is purely a lattice -> lattice functor, as the gas needs to be reset at every transfer function, no?

Comment (sim642) [reply src/analyses/apron/apronAnalysis.apron.ml:36]: Actually the `NarrowingGas.Dom` functor here (based on `WideningDelay`, etc) is essentially just that. It's the act of `unlift` and `lift` to do the operation that leads to the reset. And that's what `NarrowingGas.DLifter` used here exactly does.

It additionally needs all relational domain operations (instead of `Spec` operations) lifted, which needs more boilerplate. And it won't really gain anything from Goblint usage perspective.

**Comment by DrMichaelPetter** [comment]
Here is a more simplified version of the same thing without autotuner -- so definitely a pure apron/polyhedra and narrowing thing:

```c
// SKIP PARAM --set ana.apron.narrowing_gas 1 --set ana.activated[+] apron --set ana.apron.domain polyhedra
int main(void)
{
  int x;
  int y;
  int xtmp;
  if (y < 0) return 0;
  if (x <= y) return 0;
  xtmp = x-y;
  while (xtmp > y)
  {
      xtmp -= y;
  }
  return 0;
}
```

michael-schwarz:
What's our plan with this?

sim642:
I think it's a standard and straightforward thing worth having. Maybe it makes sense to just get this merged and not block it behind some other unresolved issue that this seems to trigger.

michael-schwarz:
@DrMichaelPetter: it might make sense to do as @sim642  suggests and bring this PR up-to-date, merge it, and pull out the fixpoint issue.

That one seems unrelated to the core thing this PR does.

DrMichaelPetter:
> @DrMichaelPetter: it might make sense to do as @sim642  suggests and bring this PR up-to-date, merge it, and pull out the fixpoint issue.
> 
> That one seems unrelated to the core thing this PR does.

I agree - next week should be light on tasks, I can have a look at it

PR #1887, opened 2025-11-28
Make coverage generation include results from cram tests

karoliineh:


.github/workflows/coverage.yml
```diff
@@ -59,7 +59,10 @@ jobs:
         run: ./make.sh headers
 
       - name: Test
-        run: opam exec -- dune runtest --instrument-with bisect_ppx
+        env:
+          BISECT_ENABLE: YES
+          BISECT_FILE: ${{ github.workspace }}/_build/bisect
```
sim642: This path is different everywhere, there's
1. _build/bisect
2. _build/coverage/bisect
3. _build/coverage

.github/workflows/coverage.yml
```diff
@@ -59,7 +59,10 @@ jobs:
         run: ./make.sh headers
 
       - name: Test
-        run: opam exec -- dune runtest --instrument-with bisect_ppx
+        env:
+          BISECT_ENABLE: YES
```
sim642: I think this isn't necessary since bisect_ppx 2.5.0, which is 5 years old: https://github.com/aantron/bisect_ppx/releases/tag/2.5.0.

sim642 (approved):

PR #1892, opened 2025-12-03
Remove division by zero error message during base `invariant`

sim642:
Discovered in https://github.com/goblint/analyzer/pull/1838#discussion_r2571447038. Should be redundant due to #1764.
Especially because this only trigger for must division by zero, not may.

It's strange because it's the only warning/error emitted during base `invariant`.

### TODO
- [ ] Is the top-ification even necessary/useful then?


karoliineh:
> What would you propose instead of `top`? Raising deadcode? These seem to be the only sensible things we can do.

This is a philosophical question I also raised with @vesalvojdani a few days ago, it seems that the uncertainty it is rooted throughout many decisions about different analyses.

To the question of "what do `bot` and `top` mean", I got an answer from @sim642 that `bot` is no possible values and `top` is any value possible. @vesalvojdani added, that the behavior of programs in real-world should in principle correspond to these values, e.g., when there's division by 0, the program will crash and there will be no possible values: `bot`. However, I then noted that we handle division by 0 to `top` in the code:

https://github.com/goblint/analyzer/blob/55c8b0d9e3310d6eaa9014382f69387399e93c96/src/analyses/baseInvariant.ml#L294-L302

Then we also discussed about overflows, and with an overflow, the program does not crash. However, we have these options `assume_top` and `assume_none`. With `assume_top` the behavior for overflows is logical to me, but with `assume_none`, if there is a definite overflow (i.e., all recorded values are out of range), we result in `bot` for the variable, and then crash every time we reach an arithmetic operation with this variable later. When I tried allowing arithmetic operations with bottoms so that they also result in `bot`, I got `[Unsound] both branches dead` for most of the cases, which is not good.

So, in a sense, `bot` if program would actually crash seems logical, but implementing it that way leads to not desired results (not more helpful than just crashing the analysis with exception). `top` seems to behave normally in most cases but then not here and if used during `assume_none`, the option would become in some way same with `assume_top`. So I am confused about the theoretical meaning and the actual behavior that happens in our analyses.

michael-schwarz:
`bot` in the sense of "no values" should be propagated to the outside by raising dead code, which gets you somewhat consistent behavior I guess.

sim642:
I'll merge this PR which is just about removing the redundant message and move the top-ification aspect to a new PR.

PR #1895, opened 2025-12-10
Add operation names to integer overflow messages and checks

sim642:
### TODO
- [x] Extend for #1894: #1896.


michael-schwarz (commented):
Why introduce polymorphic variants here? In other places we got rid of them for performance reasons.

PR #1897, opened 2025-12-10
Simplify witness invariants for bitfield domain

sim642:
I just happened to look into a witness we now produce with the portfolio at a level which enables the bitfield domain and saw lots of trivial invariants like `(i & 0) == 0`.
The non-relational witness invariant simplification (#1517) happened before the bitfield domain was introduced. The latter wasn't added to witness invariant tests, so this went unnoticed.

This is a bit unfortunate because we now have a bunch of logic for simplifying non-relational and relational witness invariants, but then we also added these, so the SV-COMP 2026 witnesses might again be unnecessarily large...

### DIFF
--- a/src/cdomain/value/cdomains/int/bitfieldDomain.ml
+++ b/src/cdomain/value/cdomains/int/bitfieldDomain.ml
@@ -625,41 +625,38 @@ module BitfieldFunctor (Ints_t : IntOps.IntOps): Bitfield_SOverflow with type in
   (* Invariant *)
 
   let invariant_ikind e ik (z, o) =
-    if BArith.is_invalid (z, o) then
-      Invariant.none (* TODO: should this ever even happen? *)
-    else (
-      let open GoblintCil.Cil in
-      let ik_type = TInt (ik, []) in
-      let i1 =
-        let def0 = z &: (!: o) in
-        if def0 =: BArith.zero_mask then
-          Invariant.none
-        else (
-          let def0 = Ints_t.to_bigint def0 in
-          if fitsInInt ik def0 then ( (* At least for _Bool and unsigned types, kintegerCilint can give an incorrect mask if doesn't fit. See https://github.com/goblint/analyzer/pull/1897/changes#r2610251390. *)
-            let def0 = kintegerCilint ik def0 in
-            Invariant.of_exp (BinOp (Eq, (BinOp (BAnd, e, def0, ik_type)), kintegerCilint ik Z.zero, intType))
-          )
-          else
-            Invariant.none
+    assert (not (BArith.is_invalid (z, o)));
+    let open GoblintCil.Cil in
+    let ik_type = TInt (ik, []) in
+    let i1 =
+      let def0 = z &: (!: o) in
+      if def0 =: BArith.zero_mask then
+        Invariant.none
+      else (
+        let def0 = Ints_t.to_bigint def0 in
+        if fitsInInt ik def0 then ( (* At least for _Bool and unsigned types, kintegerCilint can give an incorrect mask if doesn't fit. See https://github.com/goblint/analyzer/pull/1897/changes#r2610251390. *)
+          let def0 = kintegerCilint ik def0 in
+          Invariant.of_exp (BinOp (Eq, (BinOp (BAnd, e, def0, ik_type)), kintegerCilint ik Z.zero, intType))
         )
-      in
-      let i2 =
-        let def1 = o &: (!: z) in
-        if def1 =: BArith.zero_mask then
+        else
           Invariant.none
-        else (
-          let def1 = Ints_t.to_bigint def1 in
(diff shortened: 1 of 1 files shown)


tests/regression/56-witness/46-top-bool-invariant.t
```diff
@@ -146,9 +146,53 @@ interval_set only:
...
+        value: (_Bool)0 <= x
+        format: c_expression
+    - invariant:
+        type: location_invariant
+        location:
+          file_name: 46-top-bool-invariant.c
+          line: 5
+          column: 3
+          function: main
+        value: (x & (_Bool)1) == (_Bool)0
```
sim642: This is wrong for a top `_Bool`, but I'm having hard time understanding why.

The representation of top `int` is `0b?...?, (zs:-1, os:-1)`, but the representation of top `_Bool` is `0b0...0?, (zs:-1, os:1)`. These are not consistent: for `int` it has an infinite sequence of unknown bits (including higher bits that fit into `int`), but for `_Bool` it's one unknown bit with infinitely many definite zero bits in front.
I suspect this causes something to go wrong because the invariant generation is made to believe that there are all these definite 0 bits, but they're outside of the type range.

What are the bits outside of type size supposed to be?
I'm not sure who supervised #1623, @michael-schwarz, @jerhard? Does anyone know

michael-schwarz: I think it was @jerhard and @DrMichaelPetter who supervised it, I only sat in on some meetings, so this may be wrong:

Iirc, the idea is that it's mathematical integers and the two's complement just proceeds infinitely long (which is consistent with what happens on upcast iirc)? Not sure if this helps clarify?

sim642: For mathematical integers I don't see an issue: they have infinitely many bits and it makes sense to speak about all of them, whether they can be zero or one. And it's very convenient that Zarith has this neat representation with infinite leading ones.

The problem is that this doesn't define what the behavior is supposed to be on machine integers.
Can the 1000th bit of `int` be a zero or a one? (My comment implies says it may be either.)
What about the 1000th bit of `_Bool`? (My comment implies it must be zero.)

What does this mean? I don't know. Neither type actually has such bit, so it's not like the concrete semantics forces such behavior. One could even argue that all bits above the type size should be `bot` because they can't be either because they don't exist.
(In which case both the zeros and ones mask have leading 0s. So none of the Zarith two's complement trickery would even be needed.)

Here's what we currently do for top values:
https://github.com/goblint/analyzer/blob/3988737be1a56f3f960e92f7a4a8293644502c92/src/cdomain/value/cdomains/int/bitfieldDomain.ml#L196-L198
Signed types give an abstraction which allows infinitely many zeros and ones. For unsigned types, it may be infinitely may zeros, but only ones up to the type size (which means higher bits will definitely be zero).

If I change that function to return always `top ()`, no test fails. So there's no clear reason for the distinction.
I tried looking at the cited Miné paper, but I didn't immediately see an answer: it defines abstract domains on mathematical integers and then just slaps a `wrap` operation around every operation or something.
However, we carry the `ikind` together with the abstract value, instead of building abstract domains on mathematical integers, so it's not exactly the same anyway.

michael-schwarz: > If I change that function to return always `top ()`, no test fails. 

I think this may just be because our tests are poor. I think the case where it would make a difference is if you do an upcast from one unsigned type to a bigger one. There, because of our definition hopefully the upper bound would survive?

----

But overall, I'm also not sure what the best / most helpful thing would be.

sim642: For now, I just wrapped things into a `Cil.fitsInInt`. Apparently, it wasn't just problematic for `_Bool`, but unsigned types in general, where the mask for definite leading zeroes (represented by some negative integer) doesn't really fit the type itself and truncates/wraps wrongly.
It's a bit similar to the issues around https://github.com/goblint/cil/pull/111.

src/cdomain/value/cdomains/int/bitfieldDomain.ml
```diff
@@ -624,19 +624,42 @@ module BitfieldFunctor (Ints_t : IntOps.IntOps): Bitfield_SOverflow with type in
...
 
-  let invariant_ikind e ik (z,o) =
-    if z =: BArith.one_mask && o =: BArith.one_mask then
-      Invariant.top ()
-    else if  BArith.is_invalid (z,o) then
-      Invariant.none
-    else
+  let invariant_ikind e ik (z, o) =
+    if BArith.is_invalid (z, o) then
+      Invariant.none (* TODO: should this ever even happen? *)
```
michael-schwarz: Maybe raise? Then we would see if this ever happens.

sim642: I did a run with that replaced by a custom exception and it never triggered. So I'll replace this by just an assertion.

PR #1900, opened 2025-12-12
Fix implicit `pthread_exit` on `main` thread return

### LINKED ISSUE #1767: Thread return value cast causes weird unsigned integer underflow
I stumbled accross a wierd overflow that souldn't happen.
With this code:
```c
int main() {
	int *buff = 0;
	return buff[5];
}
```
And those parameters:
```sh
--ana.arrayoob true --ana.int.interval_set true --ana.base.arrays.domain partitioned --ana.base.arrays.nullbytes true --sem.malloc.fail true --set "ana.activated[+]" memOutOfBounds --set "ana.activated[+]"  useAfterFree
```
I get the following error message:
```
[Warning][Integer > Overflow][CWE-190][CWE-191] Unsigned integer overflow and underflow (... .c:6:9-6:16)
```
But for me, it shouldn't happen as a possibly random value set by the OS, can't be an overflow value. Let alone for an unsigned integer.

sim642:
Closes #1767.

A normal return from `main` does not constitute an implicit `pthread_exit` which would cast the return value to `void*`: https://github.com/goblint/analyzer/issues/1767#issuecomment-3642590227.

### DIFF
--- a/tests/regression/56-witness/30-base-unassume-inc-dec.t
+++ b/tests/regression/56-witness/30-base-unassume-inc-dec.t
@@ -24,7 +24,7 @@ Assertions not proven.
 
 Easiest to run again to get evals.
   $ goblint --enable ana.int.interval --set solvers.td3.side_widen always --set ana.base.privatization vojdani 30-base-unassume-inc-dec.c -v 2>&1 | grep 'evals'
-  [Info] vars = 39    evals = 73    narrow_reuses = 3
+  [Info] vars = 38    evals = 73    narrow_reuses = 3
 
 ## Unassume after lock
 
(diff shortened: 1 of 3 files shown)


src/analyses/base.ml
```diff
@@ -2058,10 +2058,12 @@ struct
           let st' = set ~man ~t_override nst (return_var ()) t_override rv in
           match ThreadId.get_current ask with
           | `Lifted tid when ThreadReturn.is_current ask ->
-            (* Evaluate exp and cast the resulting value to the void-pointer-type.
-                Casting to the right type here avoids precision loss on joins. *)
-            let rv = VD.cast ~torg:(Cilfacade.typeOf exp) Cil.voidPtrType rv in
-            man.sideg (V.thread tid) (G.create_thread rv);
+            if not (ThreadIdDomain.Thread.is_main tid) then (
```
michael-schwarz: Maybe leave a comment why it's different for main?

Issue #1610, opened 2024-10-30
Apron analysis does not support trivial assertions


sim642:
The output of
```console
./regtest.sh 78 02 --trace apron
```
contains
```
%%% apron: assert_constraint 1 > 0 unsupported: Exp_not_supported
%%% apron: assert_constraint 1 < 0 unsupported: Exp_not_supported
```
This is odd because there's nothing complicated or problematic (e.g. overflows) with them.
Asserting the first one is of course useless (it refines nothing), which is also what currently happens with unsupported expressions (but it would be nice to not have spurious messages like this).
Asserting the second one should refine to bottom, instead of remaining unchanged (we're missing the most precise possible result!).

michael-schwarz:
My guess is that our expression translation does simply not translate constants to assertions. I also don't know what we expect to find here, given that we almost never run Goblint without at least constant propagation enabled, and the refinement to bottom happens at the analysis product level.

sim642:
The program doesn't contain expressions like `1 > 0` though. These are something the relational analysis itself conjures up somewhere.

michael-schwarz:
I guess it's what happens with `Pos(1)` internally.

PR #1904, opened 2025-12-29
Fix `SharedFunctions.AssertionModule.exp_is_constraint`

sim642:
`LAnd`, `LOr` and `LNot` are constraint regardless of what's in their arguments (they don't need a `!= 0` added outside if something is non-constraint inside).

This caused us to not validate some witnesses containing trivial invariants in https://github.com/goblint/analyzer/pull/1897#issuecomment-3696001176.
For example `x >= 0 && 1` does not need to be turned into `(x >= 0 && 1) != 0` because one argument of `&&` contains a non-constraint. Such doubly-constraint expression is not supported by our Apron conversion.
Instead, it can stay as it is, so the `x >= 0` part is supported, but just the `1` part will be unsupported (#1610), which is irrelevant.

### DIFF
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -870,15 +870,9 @@ struct
     in
     if eval_bool exp st = Some (not tv) then contra st (* we already know that the branch is dead *)
     else
-      (* C11 6.5.13, 6.5.14, 6.5.3.3: LAnd, LOr and LNot also return 0 or 1 *)
-      let is_cmp = function
-        | UnOp (LNot, _, _)
-        | BinOp ((Lt | Gt | Le | Ge | Eq | Ne | LAnd | LOr), _, _, _) -> true
-        | _ -> false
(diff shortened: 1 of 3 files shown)


sim642:
I did verify-validate sv-benchmarks runs on level02.

There's an increase in the number of confirmed invariants:
image

There are now 50 more self-validated witnesses: https://goblint.cs.ut.ee/results/307-all-validate-level02-pr-1904-after/table-generator-cmp.table.html#/table?filter=1(0*status*(status(notIn(true)),category(notIn()))),3(0*status*(status(in(true)),category(notIn()))).
These mostly overlap with https://github.com/goblint/analyzer/pull/1897#issuecomment-3696001176, because it's a different way to deal with the same issue: instead of not generating trivial invariants, don't be bothered about triviality during validation. But we should have both PRs regardless because this also helps with validation of other tools' witnesses because there are others who also have generated trivial invariants.

PR #1906, opened 2025-12-30
Simplify return type of `FloatDomain` predicates

sim642:
While looking at all the `Unknown` cast kinds in #1896, I realized that all the `int` casts of `FloatDomain` predicate results are pointless because they are already `int` by construction.

In order to enforce this on the OCaml type level, it's better to have those predicates return `bool option` instead of `ID.t`, which is implicitly assumed/required to correspond to a boolean (but in `int`, not `_Bool`).
In fact this revealed that some of those operations actually were incorrectly returning `_Bool`, so the cast was necessary before.

This simplifies their implementations and reduces the `IntDomain` dependency of `FloatDomain`.
I've been thinking about a similar simplification in `IntDomain`, but it's a bit more involved due to the many different module interfaces.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -358,8 +358,9 @@ struct
       Int (ID.cast_to ~kind:Internal result_ik (binop_ID result_ik op v1 v2)) (* TODO: proper castkind *)
     (* For the float values, we apply the float domain operators *)
     | Float v1, Float v2 when is_int_returning_binop_FD op ->
-      let result_ik = Cilfacade.get_ikind t in (* TODO: isn't this always IInt? *)
-      Int (ID.cast_to ~kind:Internal result_ik (fd_binary_pred (int_returning_binop_FD op v1 v2))) (* TODO: cast redundant/proper castkind *)
+      let result_ik = Cilfacade.get_ikind t in
+      assert (result_ik = IInt);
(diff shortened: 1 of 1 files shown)


src/analyses/base.ml
```diff
@@ -345,8 +358,8 @@ struct
       Int (ID.cast_to ~kind:Internal result_ik (binop_ID result_ik op v1 v2)) (* TODO: proper castkind *)
     (* For the float values, we apply the float domain operators *)
     | Float v1, Float v2 when is_int_returning_binop_FD op ->
-      let result_ik = Cilfacade.get_ikind t in
-      Int (ID.cast_to ~kind:Internal result_ik (int_returning_binop_FD op v1 v2)) (* TODO: proper castkind *)
+      let result_ik = Cilfacade.get_ikind t in (* TODO: isn't this always IInt? *)
+      Int (ID.cast_to ~kind:Internal result_ik (fd_binary_pred (int_returning_binop_FD op v1 v2))) (* TODO: cast redundant/proper castkind *)
```
sim642: I'm doing an sv-benchmarks run with an `assert (result_ik = IInt);` here to see if it ever fails.

sim642: It never did, so I'm leaving the assertion and removing the cast. If CIL typing is correct, this should be the case.

PR #1911, opened 2026-01-16
Add warnings about missing function declarations

sim642:
This is allowed by older versions of C.
These functions are assumed to return `int`, etc, which can cause internal type mismatches for standard functions.
For example, it's a problem if `llabs` gets return type `int` by CIL, but Goblint assumes it to be `long long`. This causes missing/extra/wrong implicit casts to be inserted, leading to ikind errors and whatnot.


src/framework/control.ml
```diff
@@ -274,7 +274,8 @@ struct
         | {vname = "getdate_err"; _} (* unix time.h, but somehow always in MacOS even without include *)
         | {vname = ("stdin" | "stdout" | "stderr"); _} (* standard stdio.h *)
         | {vname = ("optarg" | "optind" | "opterr" | "optopt" ); _} (* unix unistd.h *)
-        | {vname = ("__environ"); _} -> (* Linux Standard Base Core Specification *)
+        | {vname = ("__environ"); _} (* Linux Standard Base Core Specification *)
+        | {vname = ("__mb_cur_max"); _} -> (* something on MacOS *)
```
michael-schwarz: `something on MacOS` is a little uninformative here.

michael-schwarz (approved):
Other than this nitpick, this should be good to go!

PR #1912, opened 2026-01-16
Extract subfunctions from base analysis `get` and `set`

sim642:
This shortcuts some `AddressSet` creation and immediate evaluation back to singleton.

The diff is probably best viewed with whitespace ignored: most of the changed parts just had indentation decreased.

This is just a refactoring, but something I did to move towards being able to remove some pointless casts: currently every local variable initialization involves an internal cast (even though `VD.init_value` should already construct the value of correct type), which has the location range of the entire function. This completely screws up location matching in the dashboard.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2044,11 +2044,10 @@ struct
       refine ()
 
   let body man f =
-    (* First we create a variable-initvalue pair for each variable *)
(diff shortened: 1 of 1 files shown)


michael-schwarz:
Iirc the problem is that `VD.init_value` may be `bot` and it is only with the first access that we generate something of the proper type?

PR #1913, opened 2026-01-17
HB-relationship involving thread creations while mutexes are held

### LINKED ISSUE #1805: Consider some more interactions between thread creation, joins, and mutexes

dabund24:
second part of #1805. The first half was implemented in #1865.
closes #1805.

# Summary

Simplest case: After creating $t_1$ in $t_0$ with mutex $l$ held, succeeding statements until maybe unlocking in $t_0$ must happen before everything after definitely locking $l$ in $t_1$.

generalizations: 
- $t_1$ can be any descendant of $t_0$ as long as $t_0$ is a must-ancestor.
- It doesn't matter if locking happens in $t_1$ or a must-ancestor as long as that thread is also a must-ancestor of the thread created in $t_0$

# Examples

In the following examples, `A` must happen before `B`.

## Simple example

```mermaid
graph TB;
subgraph t1;
    E["lock(l);"]-->F;
    F["unlock(l);"]-->G;
    G((B))
end;
subgraph t0;
    A["lock(l);"]-->B;
    B["create(t1);"]-->C;
    C((A))-->D;
    D["unlock(l);"];
end;
B-.->E
```

## `B` in a descendant of $t_1$

```mermaid
graph TB;
subgraph t2;
    H((B))
end;
subgraph t1;
    E["lock(l);"]-->F;
    F["create(t2);"]
end;
subgraph t0;
    A["lock(l);"]-->B;
    B["create(t1);"]-->C;
    C((A))-->D;
    D["unlock(l);"];
end;
B-.->E
F-.->H
```

## `A` in a descendant of $t_0$

```mermaid
graph TB;
subgraph t1;
    E["lock(l);"]-->I;
    I["unlock(l);"]-->F;
    F((B));
end;
subgraph t2;
    H((A))
end;
subgraph t0;
    A["lock(l);"]-->B;
    B["create(t1);"]-->C;
    C["create(t2);"]-->D;
    D["join(t2);"]-->G;
    G["unlock(l);"]
end;
B-.->E
C-.->H
H-.->D
```

Here, it is important that no unlock happens in $t_0$ before $t_2$ is joined into $t_0$, which was computed in #1865.

# Dependency Analyses
- $t_{\mathrm{ego}}$: Ego Thread Id at program point 
- $\mathcal L$: Must-Lockset at program point
- $\mathcal C$: May-Creates of ego thread before program point
- $\mathcal J$: Transitive Must-Joins of ego thread before program point
- $\mathcal{DES}\ t$: Descendant threads of $t$ (implemented in this PR)
- $\mathcal{ANC}\ t$: Must-ancestors of $t$

From these analyses, we compute:
- Given a statement `create(t)` all threads transitively created, for which $t_{\mathrm{ego}}$ is a must-ancestor:
$$c^* \ t:= \set{t _ d\mid t _ d\in \set{t} \cup \mathcal{DES}\ t, t_{\mathrm{ego}}\in\mathcal{ANC}\ t_d}$$
- All possibly running descendants, for which $t_{\mathrm{ego}}$ is a must-ancestor:
$$\mathcal{R}:= \set{t_ r\mid t_ r\in \left(\mathcal{C}\cup \bigcup{\set{\mathcal{DES} c\mid c\in\mathcal{C}}}\right)\setminus \mathcal J, t_{\mathrm{ego}}\in\mathcal{ANC}\ t_r}$$

# Analyses

## Descendant Locksets $\mathcal{DL}$

- flow-sensitive
- Domain: $T\to 2^L$
- $T\to 2^L$ is `MapBot`
- $2^L$ is Must-Set
- $\mathcal{DL}=\set{t_1\mapsto \set{l}}$ means

- There must have existed at least one `create(`$t_c$`)` statement in $t_A$ with $t_B\in c^*\ t_c$.
- For all of those `create(`$t_c$`)` statements, $\mathit{l}\in\mathcal{L}$.
- We must not have encountered an `unlock(`$l$`)` statement after having detected the thread creation.

### Transfer functions

- $\mathsf{init}^\sharp = \emptyset$
- $\mathsf{new}^\sharp\ X = \emptyset$
- $[[\mathsf{create}(t)]]^\sharp\ X=X\sqcup\set{t_d\mapsto \mathcal{L}\mid t_d\in c^*\ t}$
- $[[\mathsf{unlock}(l)]]^\sharp\ X=\set{t\mapsto L\setminus\set{l}\mid t\mapsto L\in X}$
- $[[\mathsf{unlock}(?)]]^\sharp\ X=\set{t\mapsto \emptyset\mid t\mapsto L\in X}$

## Mustlock History $\mathcal{LH}$
- flow-sensitive
- Domain: $L\to 2^T$
- $T\to 2^T$ is `MapTop`
- $2^T$ is Must-Set
- $\mathcal{LH}=\set{l\mapsto \set{t}}$ means "before the next operation, mutex $l$ must have been locked in $t$"

### Transfer functions

- $\mathsf{init}^\sharp=\emptyset$
- $\mathsf{new}^\sharp\ X=X$
- $[[\mathsf{lock}(l)]]^\sharp\ X=X\oplus\set{l\mapsto (X\ l)\cup \set{t_{\mathrm{ego}}}}$

## Global descendant lockset $\mathcal{DL}_g\ t$

- flow-insensitive with `V`$=T$
- Domain: $T\to T\to 2^L$
- $T\to T\to 2^L$ and $T\to 2^L$ are `MapBot`
- $\mathcal{DL}_G\ t=\set{t_{\mathrm{anc}}\mapsto DL}$ means "throughout the entire execution of $t$, the descendant lockset $DL$ is valid in $t_{\mathrm{anc}}$".

### Contributions

We only contribute at `create(`$t_c$`)` statements for all $t_d\in t^*\ t_c$:
$$DL_{t_d}:=\set{t\mapsto (\mathcal{DL}\ t)\cap (\mathcal{CL}\ t_d\ t_{\mathrm{ego}})\mid t\in T}$$
$$\mathcal{DL}_ g\ t_ d\sqsupseteq \set{t_ {\mathrm{ego}} \mapsto DL_ {t_d}}$$

# Happened-Before rules

Statement `s2` with $\mathcal{LH}_ 2, t_2$ must happen after `s1` with $\mathcal{DL}_ 1, \mathcal{LH}_ 1, t_1$, if:
- $\exists t_2\mapsto L_a\in\mathcal{DL}_ 1, l_ {LH}\mapsto T_ {LH}\in \mathcal{LH}_ 1, t_ {LH}\in T_ {LH}:$
$l_{LH}\in L_a\land t_1\in \mathcal{ANC}\ t_{LH}\land(t_{LH}\in\mathcal{ANC}\ \mathcal t_2\lor t_{LH}=t_2)$ or
- $\exists (t_X\mapsto DL_{t_1})\in\mathcal{DL}_ g\ t_1$ such that the rule above holds replacing $t_ 1$ by $t_ X$ and $\mathcal{DL}_ 1$ by $DL_ {t_1}$.


tests/regression/53-races-mhp/59-dl_multiple_mutexes_racing.c
```diff
@@ -0,0 +1,22 @@
+// PARAM: --set ana.activated[+] threadJoins --set ana.activated[+] threadDescendants --set ana.activated[+] mustlockHistory --set ana.activated[+] descendantLockset
+#include <pthread.h>
+
+int global = 0;
+pthread_mutex_t mutex1, mutex2 = PTHREAD_MUTEX_INITIALIZER;
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1913/commits/6020267c773a38d0552302b87d8f9036813c6264

tests/regression/53-races-mhp/45-dl_multiple_mutexes_racefree.c
```diff
@@ -0,0 +1,22 @@
+// PARAM: --set ana.activated[+] threadJoins --set ana.activated[+] threadDescendants --set ana.activated[+] mustlockHistory --set ana.activated[+] descendantLockset
+#include <pthread.h>
+
+int global = 0;
+pthread_mutex_t mutex1, mutex2 = PTHREAD_MUTEX_INITIALIZER;
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1913/commits/6020267c773a38d0552302b87d8f9036813c6264

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,156 @@
...
+    TIDs.fold
+      (fun tid ->
+         let old_value = D.find tid man.local in
+         let new_value = Lockset.remove lock old_value in
+         D.add tid new_value)
+      possibly_running_tids
+      (D.empty ())
+
+  let unknown_unlock man tid possibly_running_tids =
+    TIDs.fold (fun tid -> D.add tid (Lockset.empty ())) possibly_running_tids (D.empty ())
```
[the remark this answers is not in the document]
dabund24: The outer `tid` is completely redundant. Fixed in https://github.com/goblint/analyzer/pull/1913/commits/bd388a2638837025c17a6b06d2ead215e4c4a6f6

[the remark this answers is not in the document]
dabund24: The outer `tid` is completely redundant. Fixed in https://github.com/goblint/analyzer/pull/1913/commits/bd388a2638837025c17a6b06d2ead215e4c4a6f6

src/analyses/creationLockset.ml
```diff
@@ -148,6 +148,11 @@ module Spec = struct
       let creation_lockset = man.global tid in
       tid, lockset, creation_lockset
     | _ -> ThreadIdDomain.UnknownThread, Lockset.empty (), G.empty ()
+
+  let query man (type a) (x : a Queries.t) : a Queries.result =
+    match x with
+    | Queries.CreationLockset t -> (man.global  t : G.t)
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1913/commits/5bf536d573e000c9efe8f30b628d7968e9f4fd6b

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,156 @@
+open Analyses
+module TID = ThreadIdDomain.Thread
+module TIDs = ConcDomain.ThreadSet
+module Lockset = LockDomain.MustLockset
+module TidToLocksetMapTop = MapDomain.MapBot (TID) (Lockset)
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1913/commits/66ad3f139e3f34e7d24113344816738f1cb06e46

However, there still needs to be some thought put into whether this is actually the correct domain.

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,156 @@
...
+  end
+
+  let name () = "descendantLockset"
+  let startstate _ = D.empty ()
+  let exitstate _ = D.empty ()
+
+  (** reflexive-transitive closure of child relation applied to [tid] and
+      filtered to only include threads, where [tid] is a must-ancestor
+      @param man any man
+      @param tid *)
```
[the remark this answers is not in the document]
dabund24: This is some Goblint jargon, so I guess sticking with this comment should be fine.

tests/regression/53-races-mhp/44-dl_cl_transitive_create_racefree.c
```diff
@@ -0,0 +1,35 @@
...
+  return NULL;
+}
+
+void *t3(void *arg) { // t3 is joined into main before unlock happens
+  global++; // NORACE
+  return NULL;
+}
+
+void *t2(void *arg) {
+  pthread_create(&id3, NULL, t3, NULL);
```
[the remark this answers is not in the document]
dabund24: fixed in https://github.com/goblint/analyzer/pull/1913/commits/6c68a6c61ba8fb06ec3da673e007c5e89f5a5a7e

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,156 @@
...
+    let lh = man.ask Queries.MustlockHistory in
+    let tid_lifted = man.ask Queries.CurrentThreadId in
+    match tid_lifted with
+    | `Lifted tid -> tid, lh, (man.local, man.global tid)
+    | _ -> ThreadIdDomain.UnknownThread, Queries.LH.empty (), (D.empty (), G.empty ())
+end
+
+let _ =
+  MCP.register_analysis
+    ~dep:[ "threadid"; "mutex"; "threadJoins"; "threadDescendants"; "mustlockHistory" ]
```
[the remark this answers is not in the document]
dabund24: Does `dep` also play a role in the order the analyses are evaluated and would impact performance this way?
If not, I think we can leave this the way it is, since we just lose some precision that way (i.e. the global analysis won't collect any information, which could be used to exclude races).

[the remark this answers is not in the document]
dabund24: Maybe emitting a warning in `maingoblint.ml` would be a better choice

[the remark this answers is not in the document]
michael-schwarz: > Does `dep` also play a role in the order the analyses are evaluated and would impact performance this way?

No, it doesn't. Does your analysis do anything useful without `creationLockset`? If not, it would be good to add it to `deps` so people get a warning if they mis-configured things.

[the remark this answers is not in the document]
dabund24: Without `creationLockset`, it cannot handle cases like the third example from the PR summary (i.e. when `A` is not in $t_0$, but one of its descendants). For all other cases, the precision stays the same.

The performance gain from omitting `creationLockset` is comparatively subtle for the (admittedly rare) case that users want goblint to catch the simpler case, but not the more complicated one. However, it is better than nothing I guess.
image

(the configs used above are the minimal configs necessary to make the analyses work; in `dl-cl`, both `creationLockset` and `descendantLockset` are enabled)

michael-schwarz:
>  I think the analyses remain sound, but get less precise. 

👍 Could you add tests here and maybe also include some tests for your first analysis (potentially in a separate PR)?

src/analyses/descendantLockset.ml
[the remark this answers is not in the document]
michael-schwarz: 👍 Ok, the. It's best not add the dependency to allow for this configuration.

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,190 @@
...
+  let exitstate _ = D.empty ()
+
+  (** reflexive-transitive closure of child relation applied to [tid] and
+      filtered to only include threads, where [tid] is a must-ancestor
+      @param man any man
+      @param tid *)
+  let must_ancestor_descendants_closure man tid =
+    let descendants = man.ask @@ Queries.DescendantThreads tid in
+    let must_ancestors_descendants = TIDs.filter (TID.must_be_ancestor tid) descendants in
+    TIDs.add tid must_ancestors_descendants
```
dabund24: I copied this directly from the `creationLockset` analysis. Do you prefer having this twice, but independent from another in the codebase or should I call the function from the other analysis (or maybe even put this in a third place)?

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,190 @@
...
+    in
+    let may_transitively_created_tids =
+      TIDs.fold
+        (fun child_tid acc ->
+           TIDs.union acc (must_ancestor_descendants_closure man child_tid))
+        may_must_ancestor_created_tids
+        (TIDs.empty ())
+    in
+    let must_joined_tids = man.ask Queries.MustJoinedThreads in
+    TIDs.diff may_transitively_created_tids must_joined_tids
```
dabund24: this is also copied directly from `creationLocksets`

src/analyses/mustlockHistory.ml
```diff
@@ -0,0 +1,38 @@
+open Analyses
+module TIDs = SetDomain.Reverse (ConcDomain.ThreadSet)
+module Lock = LockDomain.MustLock
+
+module Spec = struct
```
dabund24, line 12: does it make sense to add `A` just for the sake of printing the abstract values? I figured doing this in `descendantLockset` wouldn't make a lot of sense because of how `should_print` is defined

src/analyses/descendantLockset.ml
sim642: It could live as a top-level function in the analysis which answers `DescendantThreads`, using an `ask` argument (instead of `man`). Then it'd be in a reasonable place without duplication.

This is what has been done with `ThreadID.get_current`, `ThreadEscape.has_escaped`, etc.

src/analyses/mustlockHistory.ml
sim642: There's not much reason to add an `A` module just for printing things at accesses without having any way to exclude races.
If `descendantLockset` queries this and stores in its `A`, then it'd be fine for `descendantLockset`'s `A` module to also print this component because this analysis isn't doing it on its own.

The idea is that if there's some information being used to rule out races, then it should be visible with `--enable allglobs` for safe accesses.
For racy accesses there should be as much as Goblint knows, but it's just not enough.
Trivial information which cannot rule out races (like empty lockset) should not clutter the output.

dabund24: I'm still not completely sure if also printing the `mustlockHistory` value along with the `descendantLockset`s is the way to go, but I feel like there is not really a perfect solution here.

If we want to exclude races for two program points, we need the `descendantLockset` value from one program point and the `mustlockHistory` information from the other.
So if we just additionally print `mustlockHistory` and leave `should_print` the way it is, we will print irrelevant `mustlockHistory` values and not print `mustlockHistory` values that actually helped us assert 'norace'.
We can fix the second issue by also checking for good `mustlockHistory` values in `should_print`, but this would leave the first issue and also print irrelevant `descendantLockset` values. If this is fine, then I will go for this approach.

src/analyses/descendantLockset.ml
dabund24: fixed in https://github.com/goblint/analyzer/pull/1913/commits/a71d29e18af53e3f9af903f1d7cb3089914f6e85

src/analyses/descendantLockset.ml
dabund24: fixed in https://github.com/goblint/analyzer/pull/1913/commits/a71d29e18af53e3f9af903f1d7cb3089914f6e85

src/analyses/mustlockHistory.ml
sim642: I think it's fine to print more information over less. `should_print` is a quite simple mechanism that worked well enough for basic locksets, but perhaps it's too limited for various MHP abstractions. It was a way to avoid cluttering the output with too much irrelevant details.

For one, it's independent of what other accesses are out there and what information from where is actually useful for excluding a race and what isn't. For optimal output, more information is probably needed to make the best decisions.

dabund24: Thanks for the insights. Fixed this in https://github.com/goblint/analyzer/pull/1913/commits/624254b9217ffd34d5db2b33125091cbd5d28711

tests/regression/53-races-mhp/42-dl_cl_simple_racefree.c
```diff
@@ -0,0 +1,31 @@
+// PARAM: --set ana.activated[+] threadJoins --set ana.activated[+] threadDescendants --set ana.activated[+] mustlockHistory --set ana.activated[+] descendantLockset --set "ana.activated[+]" creationLockset
```
michael-schwarz:
```suggestion
// PARAM: --set ana.activated[+] threadJoins --set ana.activated[+] threadDescendants --set ana.activated[+] mustlockHistory --set ana.activated[+] descendantLockset --set ana.activated[+] creationLockset
```

tests/regression/53-races-mhp/44-dl_cl_transitive_create_racefree.c
```diff
@@ -0,0 +1,36 @@
+// PARAM: --set ana.activated[+] threadJoins --set ana.activated[+] threadDescendants --set ana.activated[+] mustlockHistory --set ana.activated[+] descendantLockset --set "ana.activated[+]" creationLockset
```
michael-schwarz: `"ana.activated[+]" creationLockset` should be `ana.activated[+] creationLockset` here as well. Probably easiest to do with a search & replace.

tests/regression/53-races-mhp/60-dl_cl_multiple_creates_racing.c
```diff
@@ -0,0 +1,44 @@
...
+}
+
+void *t2(void *arg) {
+  global++; // RACE!
+  return NULL;
+}
+
+int main(void) {
+
+  int maybe;
```
michael-schwarz: With the goal of turning these into SV-COMP tests later, it may make sense to replace all occurences where we rely on uninitialized locals for non-determinism with the proper nondet functions.

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,168 @@
...
+  (** [{ t_d |-> L }]
+
+      [t_d] was transitively created with all members of [L] held.
+      Additionally, no member of [L] could have been unlocked after the creation of [t_d]
+  *)
+  module D = MapDomain.MapBot (TID) (Lockset)
+
+  (** [{ t_0 |-> { t_d |-> L } }]
+
+      [{ t_d |-> L }] is the descendant lockset valid for the [V] value,
```
michael-schwarz, line 27: What is `[V]` supposed to mean here?

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,168 @@
...
+      because [t_d] was created in [t_0] with the lockset being a superset of L.
+  *)
+  module G = MapDomain.MapBot (TID) (D)
+
+  module V = struct
+    include TID
+    include StdV
+  end
```
michael-schwarz: I think we have four instances of this now, maybe add a `TIDV` to `analyses.ml`?

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,168 @@
...
+          Lockset.fold
+            (fun lock -> TIDs.union (Queries.LH.find lock lh2))
+            locks_held_creating_t2
+            (TIDs.empty ())
+        in
+        TIDs.exists
+          (fun t_lh ->
+             TID.must_be_ancestor t1 t_lh
+             && (TID.equal t_lh t2 || TID.must_be_ancestor t_lh t2))
+          relevant_lh2_threads)
```
michael-schwarz: The indentation is a little strange here. Also, I don't think you need the `( )` around the else case.

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,168 @@
+(** descendant lockset analysis [descendantLockset]
+    analyzes a happened-before relationship related to thread creations with mutexes held.
+
+    Enabling [creationLockset] may improve the precision of this analysis.
+
+    @see https://github.com/goblint/analyzer/pull/1923
+*)
```
michael-schwarz, line 7: Maybe you should push your thesis to https://github.com/goblint/theses and then reference it here (and add something like "available upon request") until there's a formal publication.

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,168 @@
...
+    TIDs.fold
+      (fun t_d -> D.join (D.singleton t_d lockset))
+      must_ancestor_descendants
+      man.local
+
+  let threadspawn man ~multiple lval f args fman =
+    let tid_lifted = man.ask Queries.CurrentThreadId in
+    let child_tid_lifted = fman.ask Queries.CurrentThreadId in
+    match tid_lifted, child_tid_lifted with
+    | `Lifted tid, `Lifted child_tid when TID.must_be_ancestor tid child_tid ->
```
michael-schwarz, line 75: This implicitly checks uniqueness of `tid`. Could you elaborate why such a check is not needed for the child?

I was wondering about the case where two threads get the same non-unique id, but for one of the creations no mutex is held while one is held for the other?

```
main

for(int i = 0; i < 2; i++) {
   if(i == 1) { lock(m) }
   create(thread1)  // Both copies receive same TID
}

g = 42; //RACE!
unlock(m)
```

```
thread1:

lock(m)
g = 8; // RACE!
unlock(m)
```

Is this handled by the lockset being a must lockset? Do you have a test for this?

michael-schwarz (changes requested):
Sorry for the stall here. I think we should now try to get this merged so it evolves with the rest of the system. Can you merge master into this (or rebase, whatever you  prefer) and address the comments?

Then it should be good to merge!

src/analyses/descendantLockset.ml
michael-schwarz: Same for the other analyses from your thesis.

src/analyses/descendantLockset.ml
dabund24: It's supposed to be the global, which the value $\\{t_0\mapsto\\{t_d\mapsto L\\}\\}$ belongs to. I agree that this is not explained too well and will give it a shot in case you think this should be improved 🙂

tests/regression/53-races-mhp/42-dl_cl_simple_racefree.c
dabund24: Fixed in https://github.com/goblint/analyzer/pull/1913/commits/f855edafaa1d8fb865a7ba7943393f1d892cc27d

tests/regression/53-races-mhp/44-dl_cl_transitive_create_racefree.c
dabund24: Fixed in https://github.com/goblint/analyzer/pull/1913/commits/f855edafaa1d8fb865a7ba7943393f1d892cc27d

tests/regression/53-races-mhp/60-dl_cl_multiple_creates_racing.c
dabund24: Fixed in https://github.com/goblint/analyzer/pull/1913/commits/3d401db304277c71ad10fa2d4ef943d2a9985b87 and https://github.com/goblint/analyzer/pull/1913/commits/e05ae7b1e83287669956340e67bd0176fe169b97. Hope this is the correct function

Interestingly, this worsens the precision of the `threadJoins` analysis: 53/48 fails now, because t2 in no longer must-joined in line 40. In case this cannot be fixed easily: Should I remove the test or change the comments to `RACE` (without the exclam)?

src/analyses/descendantLockset.ml
dabund24: Sounds reasonable. Fixed in https://github.com/goblint/analyzer/pull/1913/commits/0f525341152ebd224f451fe4232f9527eae8bed6

src/analyses/descendantLockset.ml
dabund24: The parentheses are indeed redundant. Removed them in https://github.com/goblint/analyzer/pull/1913/commits/c6a43d7e2fa59030e56c44145db929f6dcaf1a29

How would you change the indentation here? (To me, it looks fine, though that's just bad taste most likely)

src/analyses/descendantLockset.ml
dabund24: Added the references in https://github.com/goblint/analyzer/pull/1913/commits/e69cfd8d50830ce17b1e0ab95d891143d36f2811
I also created a [PR for the thesis](https://github.com/goblint/theses/pull/3) instead of pushing it directly just in case.

src/analyses/descendantLockset.ml
dabund24: > Is this handled by the lockset being a must lockset?

Yes, pretty much. The reasoning behind this works (at least for me; hope this is sensible) like so:
If no path-splitting is done, $m$ won't be in the must-lockset for any creation and we are good.
If some path-splitting is done (which happens in the test), there will be a path where $m$ is not in the must-lockset for the corresponding creation. For this path, a race will be reported.

> Do you have a test for this?

Now I have one :). Added it in https://github.com/goblint/analyzer/pull/1913/commits/b395b79bcaf6b1365783f79999908f18644370c0.

michael-schwarz: Sounds good!

src/analyses/creationLockset.ml
```diff
@@ -2,7 +2,7 @@
     constructs edges on the graph over all threads, where the edges are labelled with must-locksets:
     (t_1) ---L--> (t_0) is represented by global t_1 t_0 = L and means that t_1 is protected by all members of L from t_0
 
-    @see https://github.com/goblint/analyzer/pull/1865
+    @see Bachelor's Thesis "Leveraging th Potential of Inter-Threaded Locksets in Abstract Interpretation". Available upon request.
```
michael-schwarz:
```suggestion
    @see Daniel Bund. "Leveraging the Potential of Inter-Threaded Locksets in Abstract Interpretation", B.Sc. thesis at TUM. Available upon request.
```

And probably change to this everywhere.

src/analyses/descendantLockset.ml
michael-schwarz: Ah, no, makes sense. Didn't remember that the globals are from `V` here.

tests/regression/53-races-mhp/60-dl_cl_multiple_creates_racing.c
michael-schwarz: I think it may just be about missing `--enable ana.sv-comp.functions`. Without this, we consider the functions unknown and they spawn all possible threads. In fact, this should be added to all tests that use the nondet functions.

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,165 @@
...
+          Lockset.fold
+            (fun lock -> TIDs.union (Queries.LH.find lock lh2))
+            locks_held_creating_t2
+            (TIDs.empty ())
+        in
+        TIDs.exists
+          (fun t_lh ->
+             TID.must_be_ancestor t1 t_lh
+             && (TID.equal t_lh t2 || TID.must_be_ancestor t_lh t2))
+          relevant_lh2_threads
```
michael-schwarz:
```suggestion
      if Lockset.is_bot locks_held_creating_t2 then
         false
      else
        let relevant_lh2_threads =
          Lockset.fold
            (fun lock -> TIDs.union (Queries.LH.find lock lh2))
            locks_held_creating_t2
            (TIDs.empty ())
        in
        TIDs.exists
          (fun t_lh ->
             TID.must_be_ancestor t1 t_lh
             && (TID.equal t_lh t2 || TID.must_be_ancestor t_lh t2))
          relevant_lh2_threads
```

I think the thing that threw me off was the `then false` on a new line. Other than that, the indentation indeed looks good!

tests/regression/53-races-mhp/46-dl_recursive_mutex.c
```diff
@@ -0,0 +1,30 @@
...
+pthread_mutex_t mutex = PTHREAD_RECURSIVE_MUTEX_INITIALIZER;
+#else
+pthread_mutex_t mutex = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
+#endif
+pthread_t id1;
+
+void *t1(void *arg) {
+  pthread_mutex_lock(&mutex);
+  pthread_mutex_unlock(&mutex);
+  global++;
```
sim642: Should there be a race annotation here?

tests/regression/53-races-mhp/47-dl_create_before_second_lock_of_recursive_mutex.c
```diff
@@ -0,0 +1,31 @@
...
+pthread_mutex_t mutex = PTHREAD_RECURSIVE_MUTEX_INITIALIZER;
+#else
+pthread_mutex_t mutex = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
+#endif
+pthread_t id1;
+
+void *t1(void *arg) {
+  pthread_mutex_lock(&mutex);
+  pthread_mutex_unlock(&mutex);
+  global++;
```
sim642: Should there be a race annotation here?

tests/regression/53-races-mhp/53-dl_maybe_unlock_parent_racing.c
```diff
@@ -0,0 +1,25 @@
...
+  global++; // RACE!
+  return NULL;
+}
+
+int main(void) {
+  pthread_mutex_lock(&mutex);
+  pthread_create(&id1, NULL, t1, NULL);
+  int maybe = __VERIFIER_nondet_int();
+  if (maybe) {
+  pthread_mutex_unlock(&mutex);
```
sim642:
```suggestion
    pthread_mutex_unlock(&mutex);
```

tests/regression/53-races-mhp/57-dl_multiple_creates_conditional_racing.c
```diff
@@ -0,0 +1,29 @@
+// PARAM: --set ana.activated[+] threadJoins --set ana.activated[+] threadDescendants --set ana.activated[+] mustlockHistory --set ana.activated[+] descendantLockset --disable ana.thread.include-node
+extern int __VERIFIER_nondet_int();
+
+#include <pthread.h>
+
+int global = 0;
+pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
+pthread_t id1, id2, id3;
```
sim642: Should this be enough?
```suggestion
pthread_t id1;
```

tests/regression/53-races-mhp/58-dl_cl_unlock_before_join_racing.c
```diff
@@ -0,0 +1,30 @@
+// PARAM: --set ana.activated[+] threadJoins --set ana.activated[+] threadDescendants --set ana.activated[+] mustlockHistory --set ana.activated[+] descendantLockset --set ana.activated[+] creationLockset
+#include <pthread.h>
+
+int global = 0;
+pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
+pthread_t id1, id2;
+
+void *t1(void *arg) {
+  pthread_mutex_lock(&mutex);
+  // everything from here must happen after unlock in main
```
sim642: This comment seems copied and wrong?

tests/regression/53-races-mhp/58-dl_cl_unlock_before_join_racing.c
```diff
@@ -0,0 +1,30 @@
...
+  pthread_mutex_unlock(&mutex);
+  global++; // RACE!
+  return NULL;
+}
+
+void *t2(void *arg) { // t2 is joined into main before unlock happens
```
sim642: This comment seems copied and wrong?

tests/regression/53-races-mhp/61-dl_sometimes_creation_without_lock_racing.c
```diff
@@ -0,0 +1,25 @@
...
+  global++; // RACE!
+  return NULL;
+}
+
+int main(void) {
+  for (int i = 0; i < 2; i++) {
+    if (i == 1) { 
+      pthread_mutex_lock(&mutex);
+    }
+    pthread_create(&id1, NULL, t1, NULL);  // Both copies receive same TID
```
sim642: Is this comment right? The first iteration should create a unique thread and the second one a non-unique one.

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,165 @@
+(** descendant lockset analysis [descendantLockset]
+    analyzes a happened-before relationship related to thread creations with mutexes held.
+
+    Enabling [creationLockset] may improve the precision of this analysis.
+
+    @see Bachelor's Thesis "Leveraging th Potential of Inter-Threaded Locksets in Abstract Interpretation". Available upon request.
```
sim642:
```suggestion
    @see Bachelor's Thesis "Leveraging the Potential of Inter-Threaded Locksets in Abstract Interpretation". Available upon request.
```

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,165 @@
...
+
+  let threadenter man ~multiple lval f args = [ D.empty () ]
+
+  let threadspawn_contribute_globals man tid must_ancestor_descendants =
+    let descendant_lockset = man.local in
+
+    (* intersect locksets, but return bot if any arg is bot *)
+    let lockset_inter_sticky_bot = function
+      | `Top, _ | _, `Top -> Lockset.bot ()
+      | ls1, ls2 -> Lockset.inter ls1 ls2
```
sim642, line 46: Is `Lockset.inter` somehow incorrect or why is this necessary?

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,165 @@
...
+    D.map (fun _ -> Lockset.empty ()) man.local
+
+  let event man e _ =
+    match e with
+    | Events.Unlock addr ->
+      let tid_lifted = man.ask Queries.CurrentThreadId in
+      let lock_opt = LockDomain.MustLock.of_addr addr in
+      (match tid_lifted, lock_opt with
+       | `Lifted tid, Some lock -> unlock man lock
+       | `Lifted tid, None -> unknown_unlock man
```
sim642: Why is the `tid` matched if it's not used at all?

src/analyses/descendantLockset.ml
```diff
@@ -0,0 +1,165 @@
...
+      D.exists ls_not_empty dl
+      || G.exists (fun _ -> D.exists ls_not_empty) dlg
+      || Queries.LH.exists (fun l tids -> not @@ TIDs.is_empty tids) lh
+  end
+
+  let access man _ =
+    let lh = man.ask Queries.MustlockHistory in
+    let tid_lifted = man.ask Queries.CurrentThreadId in
+    match tid_lifted with
+    | `Lifted tid -> tid, (man.local, man.global tid, lh)
```
sim642: The query is unnecessary for the other case:
```suggestion
    let tid_lifted = man.ask Queries.CurrentThreadId in
    match tid_lifted with
    | `Lifted tid -> tid, (man.local, man.global tid, man.ask Queries.MustlockHistory)
```

src/analyses/threadDescendants.ml
```diff
@@ -4,16 +4,40 @@
...
+(** reflexive-transitive closure of child relation applied to [tid]
+    and filtered to only include threads, where [tid] is a must-ancestor
+    @param man any man
+    @param tid
+*)
+let must_ancestor_descendants_closure man tid =
+  let descendants = man.ask @@ Queries.DescendantThreads tid in
+  let must_ancestors_descendants = TIDs.filter (TID.must_be_ancestor tid) descendants in
+  TIDs.add tid must_ancestors_descendants
+(** compute all descendant threads that may run along with the ego thread at a program point.
```
sim642, line 19:
```suggestion
  TIDs.add tid must_ancestors_descendants

(** compute all descendant threads that may run along with the ego thread at a program point.
```

src/analyses/creationLockset.ml
dabund24: Fixed in https://github.com/goblint/analyzer/pull/1913/commits/dfb2eba00682632981eab060ce748c2a5ba874d2

tests/regression/53-races-mhp/60-dl_cl_multiple_creates_racing.c
dabund24: Ah, that was it! Added this in https://github.com/goblint/analyzer/pull/1913/commits/5ec0f126dec9babbf44a8866ea5d9641162abee0

tests/regression/53-races-mhp/58-dl_cl_unlock_before_join_racing.c
dabund24: yes and yes, fixed in https://github.com/goblint/analyzer/pull/1913/commits/31f5f3fe6dab70b4fdd2ca8951a51b9a12ccee3b

tests/regression/53-races-mhp/58-dl_cl_unlock_before_join_racing.c
dabund24: Technically, the comment is correct. However, some framing that we don't benefit from this information should definitely be added. Did this in https://github.com/goblint/analyzer/pull/1913/commits/31f5f3fe6dab70b4fdd2ca8951a51b9a12ccee3b

tests/regression/53-races-mhp/46-dl_recursive_mutex.c
dabund24: Yes, added the annotation in https://github.com/goblint/analyzer/pull/1913/changes/bc552edc4e90ec7a5b5ce34a436f20a4d4274971

tests/regression/53-races-mhp/47-dl_create_before_second_lock_of_recursive_mutex.c
dabund24: Yes, added the annotation in https://github.com/goblint/analyzer/pull/1913/changes/bc552edc4e90ec7a5b5ce34a436f20a4d4274971

tests/regression/53-races-mhp/53-dl_maybe_unlock_parent_racing.c
dabund24: Fixed in https://github.com/goblint/analyzer/pull/1913/commits/ab3b3c6d9dc6bfc815f82842aecbc3f96283df45

tests/regression/53-races-mhp/57-dl_multiple_creates_conditional_racing.c
dabund24: Only `id1` and `id2` are used. I removed `id3` in 81d3ac9529f445c52fbf7f28c367292fbb9d7e68

tests/regression/53-races-mhp/61-dl_sometimes_creation_without_lock_racing.c
dabund24: Added a third loop iteration for this case in 76c1aa89a72544796f8a978ac9ba1d4997d299da

src/analyses/descendantLockset.ml
dabund24: Fixed in https://github.com/goblint/analyzer/pull/1913/commits/dfb2eba00682632981eab060ce748c2a5ba874d2

src/analyses/descendantLockset.ml
dabund24: Nice catch. This is indeed not needed. Fixed in 425d0d43732bcfaaa9b08a9b9b6d6aa446fc8b40

src/analyses/descendantLockset.ml
dabund24: True. Fixed in ec2b4f5d6f7a1b6d766eb25715f77218a1777407

src/analyses/threadDescendants.ml
dabund24: Fixed in 6d501eb43c37d6a2dc2018088124877bc1c263cb

src/analyses/descendantLockset.ml
dabund24: This is a really interesting point. What I wanted to achieve here is explicitly checking which mutexes are included in both locksets (where besides $\emptyset$, $\bot$ also means that no mutex is included). So in theory, this expresses what we actually want here.

In practice, though, this will never happen (at least with how the analyses work right now), as in both cases, $\bot$ means that a thread has never been created at all from some other thread (meaning $\mathcal{CL}\ t_1\ t_{\mathrm{ego}} = \bot\iff \mathcal{DL}\ t_1 =\bot$). So `Lockset.inter` should also work here, even if a little by chance.

Not sure how to proceed here. If you strongly lean towards using `Lockset.inter`, I can change this.

michael-schwarz: Those are must mutexes, right? So 

~~~OCaml
`Top
~~~

which is the same as `bot ()` is actually the **full** set of mutexes, not $\emptyset$.

dabund24: I phrased this in a somewhat sloppy way. In my analyses, $\bot$ actually does not represent "all mutexes", but "no mutexes" is not 100% spot on either. Rather, it means that some "may/ $\exists$ -condition" is not satisfied, which is the bare minimum for *any* mutex to be included in the lockset.
For instance, in the case of the creation lockset, a mutex cannot possibly protect a thread $t_1$ from another thread $t_0$ if $t_0$ never creates $t_1$ (transitively). However, we cannot encode this with $\emptyset$, because we want to be able to "eliminate" this state by joining it with something. In must-locksets this is not possible. Using $\bot$ for this works perfectly, though

dabund24: The only issue I see these two representations to possibly clash is when reading out the must-lockset query. This would fortunately only be a precision problem, however, I admit that I have not given this a lot of thought

michael-schwarz: I think for the implementation this is fine for now, for the writeup we may have to think how to separate those things cleanly.

sim642: So I guess it is a bit similar to #1977. I think it would be good to have some of this explanation as a code comment.

dabund24: I added some brief explanations in https://github.com/goblint/analyzer/pull/1913/commits/2cb2bd5055f119efdb26d6d8659dedaa1c6f248c. Feel free to resolve this if you find this sufficient/let me know in the other case :)

Nevertheless, thanks for pointing out that this admittedly *really* important piece of information is documented nowhere in the code. I did not at all have this on my radar for some reason 😅

michael-schwarz (approved):

sim642 (approved):

PR #1931, opened 2026-01-23
Deduplicate TID and lockset in access output with creationLockset

sim642:
While reviewing #1928 I inspected the current access output of the creationLockset analysis to understand why it wasn't precise enough.
While doing so, I noticed that the TID and lockset are printed twice in access output because they are queried from other analyses which are already responsible for outputting those parts.

On a side note, I'm wondering if `should_print` could be refined further. For example,
```
creationLockset:{
                [main] -> {}
              }
```
is a non-empty map, but only containing empty locksets. That can't really rule out any races, can it?

### DIFF
--- a/src/analyses/creationLockset.ml
+++ b/src/analyses/creationLockset.ml
@@ -143,7 +143,7 @@ module Spec = struct
     let to_yojson (_, _, cl) = G.to_yojson cl
     let printXml f (_, _, cl) = G.printXml f cl
 
-    let should_print (_t, _ls, cl) = not @@ G.is_empty cl
+    let should_print (_t, _ls, cl) = G.exists (fun _ ls -> not @@ Lockset.is_empty ls) cl
   end
 
   let access man _ =


src/analyses/creationLockset.ml
```diff
@@ -137,6 +137,12 @@ module Spec = struct
          || one_protected_inter_threaded_other_intra_threaded cl1 t2 ls2
          || one_protected_inter_threaded_other_intra_threaded cl2 t1 ls1)
 
+    (* TID and Lockset components are directly queried from other analyses and already are printed by them *)
+    let pretty () (_, _, cl) = G.pretty () cl
+    let show (_, _, cl) = G.show cl
+    let to_yojson (_, _, cl) = G.to_yojson cl
+    let printXml f (_, _, cl) = G.printXml f cl
+
     let should_print (_t, _ls, cl) = not @@ G.is_empty cl
```
dabund24: > On a side note, I'm wondering if `should_print` could be refined further. For example,
> ```
> creationLockset:{
>                 [main] -> {}
>               }
> ```
> is a non-empty map, but only containing empty locksets. That can't really rule out any races, can it?

True. I guess this would be a better fit for `should_print`.

```suggestion
    let should_print (_t, _ls, cl) = G.exists (fun _ ls -> not @@ Lockset.is_empty ls) cl
```

sim642: Perhaps this could even go further: never have such empty values in there at all and consider missing keys to have empty values.
Since these are `MustLockset`-s where `top` is the empty one, this kind of suggests that there should be a `MapTop` instead.
It wouldn't actually automatically get rid of them, because I guess these empty sets can arise through `join`-s in global unknowns and `MapTop` wouldn't trim such values. Probably not a worthwhile optimization anyway.

dabund24: I think that this would break the analysis, since it heavily relies on the fact that ascending chains like the one below are possible, where the values satisfying the condition we want to assess can be "sandwiched" by some where this is not the case:

$\emptyset \sqsubset \\{t\mapsto \\{l\\}\\} \sqsubset \\{t\mapsto \emptyset\\}$

We go from the missing key to having it by seeing `create(t)` and shrink the lockset if we encounter other creates with smaller locksets or unlocks before t is joined.

PR #1933, opened 2026-01-28
Disable constant folding to not miss checks in constant expressions

sim642:
When CIL folds constant expressions, Goblint never sees the operations, and thus does not appear to perform all the necessary overflow checks. This used to be optional for our first SV-COMP no-overflow implementation in #149.

This reveals new things which probably should be evaluated speculatively.

### TODO
- [x] Fix tests.
- [ ] Merge https://github.com/goblint/cil/pull/215.
- [ ] Add missing cases to base `eval_rv_base`.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -960,9 +960,13 @@ struct
           | bytes -> Int (ID.of_int !Cil.kindOfSizeOf (Z.of_int bytes))
           | exception SizeOfError _ -> Int (ID.top_of !Cil.kindOfSizeOf) (* TODO: does this ever happen? *)
         end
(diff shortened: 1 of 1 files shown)


sim642:
I was worried this might hurt performance, but sv-benchmarks level01 evaluation even suggests the opposite: OLS regression shows 5% speedup overall.
I don't know if this is due to disabling `lowerConstants` on the CIL side or not doing `constFold` in base, or both. I'll try to do an extra run to determine that.

The reason I was worried about performance is related to programs with many or large constant-foldable expressions. In theory, it might be cheaper to fold them directly, instead of using multiple of our int domains to reach the same constant result.
However, I now also see one way the performance could've improved: most expressions are not foldable, so `constFold` will just make an unnecessary copy on each evaluation (of each subexpression!).

PR #1941, opened 2026-02-05
Bachelor's Thesis "Synthesizing Ranking Functions for LASW programs with Goblint"

alexandermazur1:
In this thesis, I'm adding another algorithm for termination proving.
It is based on the synthesis of linear ranking functions suggested by Podelski and Rybalchenko in https://www.cs.cmu.edu/~hzarnani/fm-rg/cmuonly/podelski04complete.pdf.
The implementation uses Apron's convex polyhedra and linear programming.

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -76,7 +76,7 @@ struct
             if VH.mem v_ins v then
               VH.find v_ins v
             else
-              let v_in = Cilfacade.create_var @@ makeVarinfo false (v.vname ^ "#in") v.vtype in (* temporary local g#in for global g *)
+              let v_in = Cilfacade.create_var @@ makeVarinfo false (v.vname ^  "#in") v.vtype in (* temporary local g#in for global g *)
               v_in.vattr <- v.vattr; (* preserve goblint_relation_track attribute *)
               VH.replace v_ins v v_in;
               v_in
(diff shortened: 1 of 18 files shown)


src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -613,7 +613,7 @@ struct
         if (one_var || GobApron.Lincons1.num_vars lincons1 >= 2) && (exact || Apron.Lincons1.get_typ lincons1 <> EQ) then
           RD.cil_exp_of_lincons1 lincons1
           |> Option.map e_inv
-          |> Option.filter (InvariantCil.exp_is_suitable ~scope)
+          (*|> Option.filter (InvariantCil.exp_is_suitable ~scope)*)
```
michael-schwarz: This will break the framework in unexpected places, I'd recommend getting rid of it soon, so you don't go chasing failures from other analyses that are caused by this. You can, e.g., add a custom query for now that doesn't do this and then get rid of it later.

src/dune
```diff
@@ -92,7 +92,7 @@
   (public_names goblint)
   (modes byte native) ; https://dune.readthedocs.io/en/stable/dune-files.html#linking-modes
   (modules goblint)
-  (libraries goblint.lib goblint_std)
+  (libraries goblint.lib goblint_std ocplib-simplex)
```
michael-schwarz, line 95: To get the CI working (highly recommended), `ocplib-simplex` needs to be added to the `opam` and `opam.locked` files.

src/analyses/apron/relationAnalysis.apron.ml
sim642: I'm guessing this is using the witness invariant query internally somewhere. It's not really intended to be used this way:
https://github.com/goblint/analyzer/blob/d2d183c64565cd5818e8aab45eaefb99f265df73/src/cdomain/value/domains/invariant.mli#L9-L10
I didn't look closely but it seems to me that `exp_list_of_constraints` is then trying to convert that invariant back into constraints. It's a very roundabout way to get constraints from one analysis to the other, going back and forth via lossy translations that aren't intended for this.

I think the same matter came up a while ago, maybe in relation to #1635 or something similar.

PR #1944, opened 2026-02-10
Master's Thesis "Garbage-Collector-Aware Static Analysis of OCaml C-stubs"

RonaldJudin:
This analysis checks that OCaml objects in C programs are not deleted by the garbage collector before usage.
It is based on OCaml's documentation at https://ocaml.org/manual/5.4/intfc.html.
The implementation uses the variable set domain.


src/analyses/base.ml
```diff
@@ -2747,22 +2736,44 @@ struct
...
+          let loc = Heap in
+          let (heap_var, addr) = alloc loc wosize in
+          (* ignore @@ printf "alloca will allocate %a bytes\n" ID.pretty (eval_int ~man size); *)
+          let blob_set = Option.map_default (fun heap_var -> [(heap_var, TVoid [], VD.Blob (VD.bot (), eval_int ~man st wosize, ZeroInit.malloc))]) [] heap_var in
+          set_many ~man st ((eval_lv ~man st lv, (Cilfacade.typeOfLval lv), VD.Address addr) :: blob_set)
+        | _ -> st
+      end
+    (* TODO: Rethink OCamlParam's placement *)
+    | OCamlParam params, _ ->
+      List.fold_left (fun acc param -> let _ = eval_rv ~man acc param in acc) st params
```
sim642, line 2764: This should be unnecessary since #1761.

src/analyses/ocaml2.ml
sim642, line 1: There should be just one version of the analysis. It could be the more advanced stack-based version, although it might be even better to combine the two such that we could detect if old- and new-style macros are being mixed within a function.
To do so, it's probably needed to use `Lattice.Lift2` for the domain to combine the two domains into one.

src/analyses/ocaml2.ml
```diff
@@ -0,0 +1,314 @@
...
+(* TODO: Write tests and test them with `ruby scripts/update_suite.rb group ocaml2` *)
+(* after removing the `SKIP` from the beginning of the tests in tests/regression/90-ocaml/{01-bagnall.c,04-o_inter.c} *)
+
+open GoblintCil
+open Analyses
+
+module VarinfoSet = SetDomain.Make(CilType.Varinfo)
+
+(** "Fake" variable to handle returning from a function *)
+let return_varinfo = dummyFunDec.svar
```
sim642, line 14: This should just reuse the one from `ReturnUtil`.

src/analyses/ocaml2.ml
```diff
@@ -0,0 +1,314 @@
...
+(** Flag for first function entered *)
+let first_function = (emptyFunction "@first").svar
+(** Flag for deregistering at return *)
+let to_deregister = (emptyFunction "@dereg").svar
```
sim642, line 18: These varinfo flag hacks should be replaced by something less hacky, e.g. extending the domain with some booleans, possibly via `BoolDomain`.

src/cdomain/value/cdomains/valueDomain.ml
sim642, line 524: I was confused why this change is here because it has been separately already done on master in the same way.
I guess there might be a merge conflict and resolving it should hide the change from this PR.

src/common/util/cilfacade.ml
```diff
@@ -232,6 +232,7 @@ let getFuns fileAST : startfuns =
       Logs.debug "Cleanup function: %s" mn; set_string "exitfun[+]" mn; add_exit def acc
     | GFun ({svar={vstorage=NoStorage; vattr; _}; _} as def, _) when get_bool "nonstatic" && not (Cil.hasAttribute "goblint_stub" vattr) -> add_other def acc
     | GFun ({svar={vattr; _}; _} as def, _) when get_bool "allfuns" && not (Cil.hasAttribute "goblint_stub" vattr) ->  add_other def  acc
+    | GFun({svar={vname=mn; vattr=attr; vtype= TFun (t, _, _, _); _}; _} as def, _) when Cil.hasAttribute "c_stub" attr || Cil.hasAttribute "c_stub" (Cil.typeAttrs(t)) -> add_main def acc
```
sim642, line 235: It might be good to rename the attribute to something like `goblint_caml_prim` to match the scheme for other redefinitions of OCaml macros and make it clear that this is something we've invented.

src/util/library/libraryDesc.ml
```diff
@@ -47,6 +47,12 @@ type math =
 type special =
   | Alloca of Cil.exp
   | Malloc of Cil.exp
+  | OCamlAlloc of Cil.exp
+  (* TODO: Rethink OCamlParam's placement. *)
+  | OCamlParam0
+  | OCamlParam of Cil.exp list
```
sim642, line 53: I might've forgotten but this `OCamlParam0` seems strange. It's unclear how it's different from `OCamlParam []`.

src/util/library/libraryFunctions.ml
```diff
@@ -1242,6 +1243,26 @@ let legacy_libs_misc_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
+    ("caml_alloc_5", special [__ "tag" []; __ "arg1" [r]; __ "arg2" [r]; __ "arg3" [r]; __ "arg4" [r]; __ "arg5" [r]] @@ fun tag _arg1 _arg2 _arg3 _arg4 _arg5 -> OCamlAlloc (GoblintCil.integer 5));
+    ("__goblint_caml_param0", special [] @@ OCamlParam0);
+    ("__goblint_caml_param1", special [__ "param" []] @@ fun param -> OCamlParam [param]);
+    ("__goblint_caml_param2", special [__ "param1" []; __ "param2" []] @@ fun param1 param2 -> OCamlParam [param1; param2]);
+    ("__goblint_caml_param3", special [__ "param1" []; __ "param2" []; __ "param3" []] @@ fun param1 param2 param3 -> OCamlParam [param1; param2; param3]);
+    ("__goblint_caml_param4", special [__ "param1" []; __ "param2" []; __ "param3" []; __ "param4" []] @@ fun param1 param2 param3 param4 -> OCamlParam [param1; param2; param3; param4]);
+    ("__goblint_caml_param5", special [__ "param1" []; __ "param2" []; __ "param3" []; __ "param4" []; __ "param5" []] @@ fun param1 param2 param3 param4 param5 -> OCamlParam [param1; param2; param3; param4; param5]);
```
sim642, line 1260: Perhaps we could get away with varargs for a single `__goblint_caml_param`.

sim642 (changes requested):

src/util/library/libraryFunctions.ml
```diff
@@ -1266,6 +1243,23 @@ let legacy_libs_misc_list: (string * LibraryDesc.t) list = LibraryDsl.[
...
 
+let ocaml_descs_list: (string * LibraryDesc.t) list = LibraryDsl.[
+    ("caml_copy_double", special [drop "nptr" []] (OCamlAlloc (GoblintCil.integer 1)));
+    (* Example: ("malloc", special [__ "size" []] @@ fun size -> Malloc size); *)
+    ("caml_alloc_small", special [__ "wosize" []; __ "tag" []] @@ fun wosize tag -> OCamlAlloc wosize);
+    ("camlidl_apron_policy_ptr_c2ml", special [drop "nptr" []] (OCamlAlloc (GoblintCil.integer 1)));
+    ("caml_alloc_2", special [__ "tag" []; __ "arg1" [r]; __ "arg2" [r]] @@ fun tag _arg1 _arg2 -> OCamlAlloc (GoblintCil.integer 2));
+    ("caml_alloc_3", special [__ "tag" []; __ "arg1" [r]; __ "arg2" [r]; __ "arg3" [r]] @@ fun tag _arg1 _arg2 _arg3 -> OCamlAlloc (GoblintCil.integer 3));
+    ("caml_alloc_4", special [__ "tag" []; __ "arg1" [r]; __ "arg2" [r]; __ "arg3" [r]; __ "arg4" [r]] @@ fun tag _arg1 _arg2 _arg3 _arg4 -> OCamlAlloc (GoblintCil.integer 4));
+    ("caml_alloc_5", special [__ "tag" []; __ "arg1" [r]; __ "arg2" [r]; __ "arg3" [r]; __ "arg4" [r]; __ "arg5" [r]] @@ fun tag _arg1 _arg2 _arg3 _arg4 _arg5 -> OCamlAlloc (GoblintCil.integer 5));
```
sim642, line 1254: It might be possible to combine these with `VarArgs` as well, like `OCamlParam`.

PR #1946, opened 2026-02-28
Bachelor's Thesis – tnilboG: Extending Goblint Analyzer with Weakest Preconditions

emwint:
This is my bachelor's thesis project with the goal of adding weakest precondition analyses ("backward analyses") functionality to Goblint and implementing a concrete "proof of concept" liveness analysis.

# Core Concepts:
Instead of only having a forward transfer function semantic, this pull request adds the possibilty of defining weakest-precondition analyses with backward transfer functions which can query and use information from the forward analyses (including globals). This means that the resulting constraint system has two variables per program point and context – a forward and a backward one:

context: $\alpha$

```mermaid
graph TD;
    u@{ shape: circle, label: "u, α" }
    v@{ shape: circle, label: "v, α" }
   u -- operation --> v
```

$`  [v, \alpha]_{\text{forward}} \sqsupseteq\ [[  \text{operation} ]] _{ \text{forward}} \space get[u, \alpha]_{\text{forward}}   `$
$`  [u, \alpha]_{\text{backward}} \sqsupseteq\ [[  \text{operation} ]] _{ \text{backward}} \space get[v, \alpha]_{\text{backward}}   `$

### Function Calls and Contexts
It is important that the backwards analysis and its variables use the context from the forward analyses, s.t. no new parts of the constraint system are explored where there is no calculated corresponding forward variable. Therefore, the enter function of the forward analyses is used (together with the state of calling program point) to calculate the context:

```mermaid
graph TD;
    u@{ shape: circle, label: "u, α" }
    v@{ shape: circle, label: "v, α" }
    f_start@{ shape: circle, label: "f_start, β" }
    f_return@{ shape: circle, label: "f_return, β" }
    
   u -- f()--> v
   u -.-> f_start
   f_start -- operations in f --> f_return
   f_return -.-> v
```

$` [u, \alpha]_{\text{backward}}  \sqsupseteq `$
$`  \quad let \space (\beta, \_)= \text{enter} _{ \text{forward}} (get[u, \alpha]_{\text{forward}}) \space in \quad `$ //using the enter function of the forward analyses do determine the context
$` \quad   let \space d = \text{enter} _{ \text{backward}} (get[v, \alpha]_{\text{backward}})  \space in `$ 
$`  \quad let () = \text{side} [\text{f\_return}, \beta]_{\text{backward}} d \space in  `$
$`  \quad \text{combine}_{\text{backward}}  (get[v, \alpha]_{\text{backward}})   (get[\text{f\_start}, \beta]_{\text{backward}})`$

Also, if the called function is given as a pointer, the information from the forward analysis is used to resolve the pointer in the backward analsis.

### Liveness Analysis
It's just a liveness analysis. Not much to say (at this point).

# Added Modules:

### `BackwSpecFunctor`  (in `framework/backwAnalyses.ml`)
   - Used to specify an analysis, the same `Spec`.
   - It is a functor of `Spec` in order to guarantee that the contex-type is the same as in the forward-analysis. This also guarantees the types of the fields of the forward-manager (containing the information of the forwards-analyses) passed to the transfer functions.

### `BackwSpec` (in `framework/backwAnalyses.ml`)
   - This is yielded when `BackwSpecSpec`  is applied to a  `Spec`.

###  `BidirFromSpec` (in `framework/bidirConstraints.ml`)
   - Is functor of forward-`Spec`, `BackwSpecSpec` , `CfgBidir ` and (unused) `Increment`
   - This works in the same way as `FromSpec`, but the resulting `DemandGlobConstrSys` includes the forward- and backward-variables. 
   - Since the solver (and the type `DemandGlobConstrSys`) expect a single type of local and global variables, the forward- and backward-variable types are lifted. 
   - The forward-transfer-functions use the existing `FromSpec` instantiated with the forward-`Spec`.
   - The backward-transfer-functions get `getl_forw` and `getg_forw` as paramter s.t. they can construct a forwards-manager to pass to the `BackwSpec`-transfer-functions.
   - here, the discussed handling of function calls (i.e. context creation and resolving function pointers) is implemented

### `AnalyzeCFG_bidir` (in `framework/control.ml`)
   - Conceptually the same as `AnalyzeCFG`, but uses `BidirFromSpec` and initializes forward- and backward-variables. 
   - Does not have the full functionality of  `AnalyzeCFG` regarding e.g. result output, optimizations, increment or transformations.

### liveness analysis (in `analyses/wp_analyses/liveness.ml`)
  - Basic liveness analysis specification with  `BackwSpecFunctor` signature using information from forward-analyses, e.g. in the branch transfer-function.

### Miscellaneous
- added tests ( in `tests/regression/99-tutorials`. Why there? No reason, should not stay there.)
- added command line option/ parameter to activate wp-analysis
- (hastily) changed xsl template for storing result

# Remaining ToDos: 
(updated 22.08.26)
- [ ] `MCP` allowing for running and choosing between different backward analyses 
- [ ] **Tests**:
   - [ ] move Tests to more appropriate place
- [ ] `AnalyzeCFG_bidir`:
   - [ ] better result output (as in `AnalyzeCFG`)
 
# Caveats and Limitations:
-  A nice result output is not that high a priority right now and I can't just easily use the existing modules since they rely on a (single) `Spec`
- neither made for nor tested with multiple threads
- Many modules depend on having only one `Spec`. Especially everything that depends on the `SpecSys` module in `AnalyzeCFG` is a real hassle, since I cannot just bundle my forward `Spec` with my combined `EQSys` or combined hash tables. This means I cannot easily use the existing `AnalysisResult` or `ResultQuery` modules...

# Usage:
- use `--enable ana.wp_run`
- view result using `--html` and the html-server


michael-schwarz:
@emwint: Was is the state here?

emwint:
> @emwint: Was is the state here?

I am currently traveling and will have the time to give a more detailed update in three weeks. Very briefly, the extension is working as intended and it should be possible to merge it without any problems, since it only extends the current functionality without interfering with the existing parts. To make this addition really useful, it is primarily missing the ability to run several backward analyses in parallel, so i.e. a backward version of / addition to `MCP`.

Issue #1449, opened 2024-05-08
Split octagon autotuner to separate octagonAnalysis and octagonVars autotuners


sim642:
Turns out when the octagon autotuner says `Enabled octagon domain for: ...`, it _implicitly_ means "disabling octagon analysis for all other variables, _even if explicitly enabled by the user".
This massively shoot us in the foot for #1394 benchmarks, where explicitly enabling the apron analysis on top of the normal svcomp configuration, which intuitively should only make things more precise, only produced 1 relational invariant. Actually the autotuner more-or-less overrode that, disabling the autotuner gives us 1896 invariants!

We're really selling our novel relational analyses short by disabling them even when explicitly requested. If it's counterintuitive for us, it will definitely be counterintuitive for anyone else evaluating Goblint.

Much less confusing behavior would be to split the autotuner into two:
1. octagonAnalysis, which is just in charge of enabling the octagon analysis, but _not choosing a subset of the variables_.
2. octagonVars, which is in charge of choosing the variables, i.e. _disabling_ octagon analysis for all other variables (because that's what it's really doing, not enabling for additional variables).

michael-schwarz:
Yes, that's a good idea! I was aware that this is the behavior of the autotuner, but then I was also the one advising the thesis which cooked up the autotuners.
If this knowledge has not even propagated in the team, then, as you say, any external people trying it will be confused for sure!

PR #1952, opened 2026-03-05
Split octagon autotuner to separate octagonAnalysis and octagonVars autotuners

karoliineh:
Solves #1449


michael-schwarz (commented):
We should start to document what these autotuners do somehow. Maybe we can have some way they are registered and a command line option to get the description of all autotuners like you can get one for the analyses?

tests/regression/29-svcomp/35-nla-sqrt.c
```diff
@@ -1,4 +1,4 @@
-// SKIP PARAM: --enable ana.sv-comp.functions --enable ana.autotune.enabled --set ana.autotune.activated[+] octagon
+// SKIP PARAM: --enable ana.sv-comp.functions --enable ana.autotune.enabled --set ana.autotune.activated[+] octagonVars
```
sim642, line 1: Why just `octagonVars`? I'd think that `octagonAnalysis` also needs to be activated.

And perhaps `octagonVars` isn't even necessary? Depends on what the test was added to test: the activation of the analysis or the selection of the variables.

tests/regression/29-svcomp/38-autotune-octagon-fun.t
```diff
@@ -1,10 +1,17 @@
...
-  [Info] i, count, tmp, count, i, j, i___0, j___0, k, size, r
+  $ goblint --enable ana.autotune.enabled --set ana.autotune.activated[*] octagonAnalysis 38-autotune-octagon-fun.c
+  [Info] Enabled octagon domain.
   [Info][Deadcode] Logical lines of code (LLoC) summary:
     live: 5
     dead: 0
     total lines: 5
 
+  $ goblint --enable ana.autotune.enabled --set ana.autotune.activated[*] octagonVars 38-autotune-octagon-fun.c
+  [Info] Enabled octagon domain.
```
sim642, line 11: So `octagonVars` also enables `octagonAnalysis` autotuner implicitly?

I think for #1449 they should be independent.

src/autoTune.ml
```diff
@@ -423,7 +423,14 @@ let congruenceOption factors file =
     activate;
   }
 
-let apronOctagonOption factors file =
+let activateOctagonAnalysis  () =
+  set_string "ana.apron.domain" "octagon";
+  set_auto "ana.activated[+]" "apron";
+  set_bool "ana.apron.threshold_widening" true;
+  set_string "ana.apron.threshold_widening_constants" "comparisons";
+  Logs.info "Enabled octagon domain."
```
sim642, line 431: This doesn't seem right without the costs thing. It will always enable octagons if `octagonAnalysis` autotuner is activated.
That's not much of an autotuner if it just unconditionally activates things which could be activated directly.

I think I wasn't aware of this dependency when I opened #1449. Maybe it's actually impossible to separate them then?

karoliineh: So, what I did was to just implement the requirements from the issue:

> split the autotuner into two:
> 1. octagonAnalysis, which is just in charge of enabling the octagon analysis, but not choosing a subset of the variables.
> 2. octagonVars, which is in charge of choosing the variables, i.e. disabling octagon analysis for all other variables (because that's what it's really doing, not enabling for additional variables).

where the main problem was that the option `octagon`:
> implicitly means "disabling octagon analysis for all other variables, _even_ if explicitly enabled by the user".

Due to the separation, the confusing behavior is gone as the option names now better reflect what they actually do.

However, it is true that I did not really consider that the autotuner should actually _tune_ something with each option, which is not the case for `octagonAnalysis` now. 

The reason I separated it the way I did is that the cost calculation depends on the number of selected variables. At first glance, this made it seem like the cost belonged purely to the variable selection step. Because of that, I overlooked the fact that the cost value is later used.

Looking at this again from the perspective of autotuning, I see two possible approaches:
1. **Base the cost on all program variables.** The cost calculation for `octagonAnalysis` could consider all variables present in the program, but I am not sure if we can count these easily with some visitor. I am also not sure if the cost will then ever stay under some threshold that the analysis would even be considered (I haven't looked into how the cost is used).
2. **Actually tune the activation of `octagonAnalysis` meaningfully** Instead of simply enabling it through a portfolio configuration, the tuner could decide whether activating it makes sense based not only on the number of variables but also on their structure, e.g., the presence of patterns that would require relational analyses like `y = x + 1;`.

It definitely is not _impossible_ to separate them cleanly. The question is whether this change is already worthwhile because it makes the options less confusing, even though one of them does not actually tune anything yet, or whether we should wait and introduce that separation only once we also improve the autotuning of enabling `octagon` in some way.

PR #1955, opened 2026-03-06
Treat string function non-address inputs as top values

### LINKED ISSUE #1934: Fatal error in `strcmp` in callbacks passing through a function with missing definition
Goblint version: [nightly sha256 68ef67776c7533bf4e89251efc5855c313dc92d7c257a7f357791595c8a8d7db](https://github.com/goblint/analyzer/actions/runs/21464729039)

When analyzing the following snippet with Goblint in default configuration:
```c
extern int strcmp(const char *, const char *);

static int hello_read(const char *path) {
    return strcmp(path, "Hello");
}

static const struct ops {
	int (*read)(const char *);
} hello_oper = {
	.read = hello_read,
};

extern int main_real(const struct ops *);

int main() {
	return main_real(&hello_oper);
}
```
it fails with
```
Fatal error: exception Failure("String function: not an address")
```
where one would expect `Function definition missing for ...`. The issue is configuration-dependent. In particular with `conf/svcomp25.json`, the error message becomes
```
Fatal error: exception Invalid_argument("option is None")
```

Changing `strcmp` to non-special function or providing a body for `main_real` mitigates the issue. The issue occurred when trying to analyze [libfuse](https://github.com/libfuse/libfuse) examples.

jprotopopov-ut:
Fixes #1934. The same issue also appears in rsync analysis.

The change might be naive, and instead upstream problem that generates non-address strings might need to get fixed. However, I am not sure where that would be, so let me know if that's the case.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2434,7 +2434,7 @@ struct
           | Addr.Addr (v, o) -> Addr.Addr (v, lo o)
           | other -> other in
         AD.map rmLastOffset a
-      | _ -> raise (Failure "String function: not an address")
+      | _ -> AD.top ()
     in
     let string_manipulation s1 s2 lv all op_addr op_array =
       let s1_v = eval_rv ~man st s1 in
--- a/tests/regression/73-strings/11-not-an-address.c
+++ b/tests/regression/73-strings/11-not-an-address.c
@@ -1,0 +1,18 @@
+// NOCRASH
+extern int strcmp(const char *, const char *);
+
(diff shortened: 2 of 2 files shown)


jprotopopov-ut:
Although, I must note that enabling `ana.sv-comp.enabled` still crashes the analyzer, just with `Invalid_argument("option is None")` error. I presume, non-address strings end up in some other analysis that is not covered by my fix.
As mentioned, more general change might be needed elsewhere to prevent such values from being created altogether.

PR #1965, opened 2026-03-20
Make signal handler threads unique

sim642:
While thinking about the signal handler analysis for AnalyzeThat, I realized that we spawn the signal handlers as non-unique threads.
This PR currently adds separate handling for signal handlers to spawn them as unique. Intuitively this seems right but I haven't thought too much about it.

It makes very little difference though: it prevents signal handlers from racing with themselves. Maybe it can also help some privatizations with TIDs.

This might also allow joining the signal handler threads with some extra work: if a new signal handler is installed, then the previous one could be joined.

### TODO
- [ ] `sigaction`?
- [ ] Should signal handlers be analyzed as self-looping because they may be recalled? Could simulate by side-effecting from signal handler return to entry? Or does it not matter for any of our thread-modular analyses?

### DIFF
--- a/tests/regression/03-practical/39-signal-arg.c
+++ b/tests/regression/03-practical/39-signal-arg.c
@@ -1,0 +1,11 @@
+#include <signal.h>
+#include <goblint.h>
+
+void handler(int sig) {
+  __goblint_check(sig == SIGTERM);
+}
+
+int main() {
+  signal(SIGTERM, handler);
+  return 0;
+}


sim642:
If I remember correctly, then signals in a multi-threaded program are handled by one arbitrarily chosen (by OS) thread.
But we should check this properly before going for this. There isn't a particular rush with this PR because on its own it's a very small step towards proper signal handler analysis.

Limiting this to single-threaded programs might be kind of tricky exactly because we still analyze signal handlers as threads. Threadflag, threadid, etc. don't distinguish them enough to make queries specific enough (when are signal handlers considered threads and when not). I suspect a proper handling probably needs to distinguish them more deeply.

PR #1981, opened 2026-04-08
Improve `exp.single-threaded` for analysis without threadflag and escape

sim642:
So far, `exp.single-threaded` only prevented new threads from being spawned but did not override multi-threadedness related queries.
Now, using the queries through the corresponding helper functions, the option has stronger effect.

In particular, this allows analyzing a single-threaded program without the threadflag and escape analyses enabled, without immediately becoming flow-insensitive for globals (like `exp.earlyglobs`) and assuming all _locals_ may have escaped (and become globals).

There is a performance benefit for having less analyses activated, even if they don't compute anything for single-threaded programs.
This is also what motivated the single-threaded autotuner for SV-COMP. This PR would allow pushing that autotuner even further to disable threadflag and escape analyses for such programs.


sim642:
In my rsync runs (https://github.com/goblint/bench/commit/58495e9504f4bbc8d56a7fa90722068a07e42992), something strange happens with this change. When deactivating threadflag and escape in favor of `exp.single-threaded` the runtime increases from 2h 30min to 11h 13min. The number of vars slightly decreases and the number of evals more than doubles.

I thought that maybe the `GobConfig.get_bool` is causing an overhead because these are probably called _very_ often (probably every access to a global at least), and should be cached into a `ref`. But that inefficiency wouldn't increase evals, so something else must be happening as well.

sim642:
> In my rsync runs ([goblint/bench@58495e9](https://github.com/goblint/bench/commit/58495e9504f4bbc8d56a7fa90722068a07e42992)), something strange happens with this change. When deactivating threadflag and escape in favor of `exp.single-threaded` the runtime increases from 2h 30min to 11h 13min. The number of vars slightly decreases and the number of evals more than doubles.

Apparently our `rsync-v3.4.1-gcc-nosignal.i` still has some signal handlers left and still was making the analysis multi-threaded. The error improvements for "Thread not spawned" in this PR already paid off.

michael-schwarz (changes requested):
Great PR, I think not using the query with the `since_start` flag directly makes the intent easier to follow.

The warning message in `maingoblint.ml` about not enabling `escape` should be adapted so it is not produced if `exp.single-threaded` is on.

Also, should we maybe make an issue for the improved autotuner (or ask AI to fix the autotuner right away)?

sim642:
> Apparently our `rsync-v3.4.1-gcc-nosignal.i` still has some signal handlers left and still was making the analysis multi-threaded. The error improvements for "Thread not spawned" in this PR already paid off.

I redid the rsync comparison after removing the remaining signal handlers in https://github.com/goblint/bench/commit/9192101db3c86cce4a59ebdd3cebb13f6538b778.
It reduced the analysis time of one configuration from ~10h 27min to ~9h 23min, which is ~10% reduction.

> The warning message in `maingoblint.ml` about not enabling `escape` should be adapted so it is not produced if `exp.single-threaded` is on.

Indeed, I'll update it.

> Also, should we maybe make an issue for the improved autotuner (or ask AI to fix the autotuner right away)?

I'll open a separate issue about it because unlike the existing single-threaded SV-COMP autotuner, this is more aggressive and requires a bit more care. This can become unsound if there are other things which we also treat as thread-like (signal handlers, `atexit`, etc.). Not sure if any of those are covered by sv-benchmarks though.
And improvements like https://github.com/goblint/analyzer/pull/1965 which change our approach to thread-like things (e.g. making them non-threads in some sense, but not other) can lead to new problematic interactions.

I'd like to have it for SV-COMP 2027 and it's easy to do closer to it happening. I want to see where we can get with the signal handlers (like https://github.com/goblint/analyzer/pull/1966) before needing to worry about the SV-COMP interactions.

michael-schwarz (approved):

Issue #1967, opened 2026-03-23
Excessive allocations during map `join` in refinement by ambiguous pointer


sim642:
I reported this originally just on Zulip, so I was confused when I started looking for this non-existent issue on GitHub.

---

> When I turned hashconsing off, then that disappeared and a lot of time was spent on Ocaml GC internals instead.

So I tried looking into the large amount of garbage using Memtrace and found out that the `branch` of base analysis was doing over 90% of the allocations. The issue there seems to be the refinement via ambiguous pointers which was implemented in https://github.com/goblint/analyzer/pull/1659.

The problem is that this `join`s base domain maps which will lead to a lot of allocations for the join of each of the values. But in that particular case, each joinee only differs from the original in 1-2 variable keys, so the entire operation should not depend on the size of the map. This could possibly be implemented more efficiently by joining some partial maps or so. Seems like we never benchmarked this precision improvement.

This would also be a problem for https://github.com/goblint/analyzer/pull/1531. The idea there was to make the writing of ambiguous pointers more intuitive (a join over all the cases) but it also introduces joins on the entire maps. So perhaps it's not a good idea to be so naive.

This is particularly wasteful because usually the refining of ambiguous pointers means nothing is refined. And the joins are a roundabout way to refine the original state in various ways and join them back together into the original state...

sim642:
Reopening because majority of time is still spent in `join` and `widen` for rsync (at the solver level, not refinement) due to the inner operators: https://github.com/goblint/analyzer/pull/2002#issuecomment-4257920213.

PR #1982, opened 2026-04-08
Add option `ana.base.branch.refine-pointer-by-pointee`

sim642:
Adds an option to disable #1659.
This can be used to work around #1967.

### DIFF
--- a/conf/examples/large-program.json
+++ b/conf/examples/large-program.json
@@ -7,6 +7,9 @@
         "privatization": "none",
         "context": {
           "non-ptr": false
+        },
+        "branch": {
+          "refine-pointer-by-pointee": false
         }
       },
       "thread": {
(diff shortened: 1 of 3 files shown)


michael-schwarz (approved):

sim642:
In my rsync runs (https://github.com/goblint/bench/commit/58495e9504f4bbc8d56a7fa90722068a07e42992), the runtime decreases from 2h 30min to 1h 34min when disabling this option. The number of vars does not change and the number of evals goes down by 7 (out of 6.6M).

The warnings don't show a precision difference. Only a bunch of `[Info][Imprecise] Trying to read a field, but the struct is unknown` warnings disappear. Maybe these are happening during this logic.
I think at some point I found a possible type mismatch problem in this logic but I'll have to revive some WIP branch.

PR #1984, opened 2026-04-09
Remove type argument from `do_update_offset`, fix calls with incorrect offset type

sim642:
This is a continuation of #1914.
There I noticed that `do_update_offset` recursively may change the type, whereas `do_eval_offset` did not. This mismatch of strange because both should recurse into base value domain abstractions in the same manner.
Thus, this PR removes that argument to make things more consistent.

Simply removing the argument breaks a lot of tests. Turns out that `VD.update_offset` is being called in multiple places with the incorrect type which isn't the one corresponding to the offset being updated.
By fixing these (explanations in commit messages), all tests again pass.

The type-changing aspect of `do_update_offset` was just covering up these root issues by managing to get the type from somewhere else.

### TODO
- [x] sv-benchmarks runs.
- [x] Check if this fixes something about https://github.com/goblint/analyzer/pull/1982#issuecomment-4212180407. Doesn't seem to.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2002,7 +2002,7 @@ struct
               let t = v.vtype in
               let iv = VD.bot_value ~varAttr:v.vattr t in (* correct bottom value for top level variable *)
               if M.tracing then M.tracel "set" "init bot value (%a): %a" d_plaintype t VD.pretty iv;
-              let nv = VD.update_offset (Queries.to_value_domain_ask (Analyses.ask_of_man man)) iv offs rval_val (Some  (Lval lval)) lval t in (* do desired update to value *)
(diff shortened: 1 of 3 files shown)


michael-schwarz (commented):
Looks reasonable, but I think this is worth an SV-COMP run before merging, a lot of the weirdness this code works around may not show up in our benchmark suite.

sim642:
I did 60s sv-benchmarks runs with level01 (https://goblint.cs.ut.ee/results/326-all-level01-pr-1984-after/table-generator-cmp.diff.html). The only real difference with before is that this makes Goblint solve one more task (aws-c-common/aws_priority_queue_clean_up_harness).
It used to have "Trying to update an index, but was not given an array" in the logs but these now disappeared.

michael-schwarz (approved):

PR #2002, opened 2026-04-15
Use Codex patricia-tree library for base map domain

### LINKED ISSUE #1967: Excessive allocations during map `join` in refinement by ambiguous pointer

sim642:
Closes #1967.

During the AnalyzeThat 2026 workshop the Codex team told us that they have specifically published on opam the patricia-tree library which provides efficient merging with idempotent operators. This is an attempt to use it for the map domain in the base analysis.
I noticed that some part of Frama-C is also possibly switching to this library: https://git.frama-c.com/pub/frama-c/-/merge_requests/18.

Note that the library doesn't completely replace `Stdlib.Map.Make`: patricia-tree requires keys to have unique `int` values.
Luckily, base analysis map domain has `varinfo`s which have `vid`, so it should work.
But we have plenty of other usages of `MapDomain` with more complex keys, so this PR tries to generalize `MapDomain` to allow both to be used.

Apparently this causes one test with int domain refinement and narrowing to fail again because the `narrow` is not idempotent.

### TODO
- [x] rsync benchmark.
- [x] sv-benchmarks.
- [x] Look through code TODOs.
- [x] Figure out what to do with int domain refinement narrowing test. Using non-idempotent merging with `meet` and `narrow` due to possible int domain refinement issues for now.

### DIFF
--- a/src/domain/mapDomain.ml
+++ b/src/domain/mapDomain.ml
@@ -28,8 +28,8 @@ sig
   val add_list_fun: key list -> (key -> value) -> t -> t
 
   val for_all: (key -> value -> bool) -> t -> bool
-  val map2: (value -> value -> value) -> t -> t -> t
-  val long_map2: (value -> value -> value) -> t -> t -> t
+  val nonidempotent_inter: (value -> value -> value) -> t -> t -> t
+  val nonidempotent_union: (value -> value -> value) -> t -> t -> t
   val merge : (key -> value option -> value option -> value option) -> t -> t -> t (* TODO: unused, remove? *)
 
   val cardinal: t -> int
@@ -210,7 +210,7 @@ struct
   let add_list_fun keys f m =
     List.fold_left (fun acc key -> add key (f key) acc) m keys
 
-  let long_map2 op =
+  let nonidempotent_union op =
     let f k v1 v2 =
       match v1, v2 with
       | Some v1, Some v2 -> Some (op v1 v2)
@@ -220,7 +220,7 @@ struct
     in
     M.merge f
 
-  let map2 op =
+  let nonidempotent_inter op =
     (* Similar to the previous, except we ignore elements that only occur in one
      * of the mappings, so we start from an empty map *)
     let f k v1 v2 =
@@ -289,9 +289,9 @@ struct
 
   let add_list_fun keys f = lift_f' (M.add_list_fun keys f)
 
-  let long_map2 op = lift_f2' (M.long_map2 op)
+  let nonidempotent_union op = lift_f2' (M.nonidempotent_union op)
 
-  let map2 op = lift_f2' (M.map2 op)
+  let nonidempotent_inter op = lift_f2' (M.nonidempotent_inter op)
 
   let leq_with_fct f = lift_f2 (M.leq_with_fct f)
   let join_with_fct f = lift_f2' (M.join_with_fct f)
@@ -342,9 +342,9 @@ struct
 
   let add_list_fun keys f = lift_f' (M.add_list_fun keys f)
 
-  let long_map2 op = lift_f2' (M.long_map2 op)
+  let nonidempotent_union op = lift_f2' (M.nonidempotent_union op)
 
-  let map2 op = lift_f2' (M.map2 op)
+  let nonidempotent_inter op = lift_f2' (M.nonidempotent_inter op)
 
   let leq_with_fct f = lift_f2 (M.leq_with_fct f)
   let join_with_fct f = lift_f2' (M.join_with_fct f)
@@ -414,9 +414,9 @@ struct
   let add_list_set ks v x = time "add_list_set" (M.add_list_set ks v) x
   let add_list_fun ks f x = time "add_list_fun" (M.add_list_fun ks f) x
 
-  let long_map2 f x y = time "long_map2" (M.long_map2 f x) y
+  let nonidempotent_union f x y = time "nonidempotent_union" (M.nonidempotent_union f x) y
 
-  let map2 f x y = time "map2" (M.map2 f x) y
+  let nonidempotent_inter f x y = time "nonidempotent_inter" (M.nonidempotent_inter f x) y
 
   let leq_with_fct f x y = time "leq_with_fct" (M.leq_with_fct f x) y
   let join_with_fct f x y = time "join_with_fct" (M.join_with_fct f x) y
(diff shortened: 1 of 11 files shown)


sim642:
For flow-sensitive analysis of rsync, this reduces the analysis time from ~9h 23min to ~7h 17min, which is about -22% (https://github.com/goblint/bench/commit/7f22f21f075e2f4a2aaa38e83a741d24b2c0e2df).

Around 20% of the time is still spent on GC. A lot of time is still spent in the inner `join`, etc. operations which can be optimized separately. My guess is that they haven't been written to carefully preserve physical equality but rather allocate new copies for identical arguments. The patricia-tree library exploits physical equality extensively for efficiency.

I'm now also doing a run on sv-benchmarks although I don't expect much change there.

michael-schwarz:
I was confused about what Codex had to do with it, but then realized it's the Lemerre et al. Codex. For anyone also confused: https://codex.top/.

sim642:
> I'm now also doing a run on sv-benchmarks although I don't expect much change there.

There's a lot more than I expected. I did before vs after runs with level01 with 60s and 1GB. There's ~53 new trues: https://goblint.cs.ut.ee/results/328-all-level01-pr-2002-before/table-generator-cmp.diff.html#/table?filter=0(0*status*(status(notIn()),category(notIn()))),1(0*status*(status(notIn()),category(in(correct)))). Some are borderline noise flips, but some are due to significant differences.

### CPU time
Average 9% improvement, up to 2×:
image

(https://goblint.cs.ut.ee/results/328-all-level01-pr-2002-before/table-generator-cmp.table.html#/scatter?toolX=0&columnX=2&toolY=1&columnY=2&filter=0(0*status*(status(notIn()),category(notIn(error)))),1(0*status*(status(notIn()),category(notIn(error))))&results=All&regression=Linear&scaling=Logarithmic&line=2)

### Memory
Average 15% improvement, up to 3×:
image

(https://goblint.cs.ut.ee/results/328-all-level01-pr-2002-before/table-generator-cmp.table.html#/scatter?toolX=0&columnX=4&toolY=1&columnY=4&filter=0(0*status*(status(notIn()),category(notIn(error)))),1(0*status*(status(notIn()),category(notIn(error))))&results=All&regression=Linear&scaling=Logarithmic&line=3)

michael-schwarz:
There's also one we lose that is unrelated to timing changes. Maybe worth taking a look, I guess this should behave exactly as the old one apart from runtime / memory concerns.

sim642:
> There's also one we lose that is unrelated to timing changes. Maybe worth taking a look, I guess this should behave exactly as the old one apart from runtime / memory concerns.

I also noticed that. It's from https://github.com/goblint/analyzer/pull/1984#issuecomment-4241950713.

sim642:
Because our `meet` and `narrow` are not guaranteed to be idempotent, I reverted them to the unoptimized operations in 1afd7c8ce48d9de5c876067f0ddc24ef2b74220a.
I also reran the sv-benchmarks and this revert has a downside: 23 trues go back to timeouts.
I'm rerunning the rsync benchmark as well to see how it affects the improvements there.

So we're leaving some performance on the table but that's the best we can do right now. I think we should rethink where to apply the reductions such that `meet` and `narrow` can still be idempotent.

src/domain/mapDomain.ml
```diff
@@ -135,12 +138,98 @@ struct
...
+module PatriciaMap (K: PatriciaTree.KEY): MapS with type key = K.t =
+struct
+  include PatriciaTree.MakeMap (K)
+
+  let merge = slow_merge (* TODO: get rid of this *)
+  let idempotent_union f = idempotent_union (fun _ v v' -> f v v')
+  let nonidempotent_union f = nonidempotent_union (fun _ v v' -> f v v')
+  let idempotent_inter f = idempotent_inter (fun _ v v' -> f v v')
+  let nonidempotent_inter f = nonidempotent_inter_no_share (fun _ v v' -> f v v')
+  let reflexive_subset_domain_for_all2 f = reflexive_subset_domain_for_all2 (fun _ v v' -> f v v')
```
michael-schwarz, line 226: `(fun _ -> f)` ?

src/domain/mapDomain.ml
```diff
@@ -758,8 +858,10 @@ struct
       er
   let map f (e, r) = (e, f r)
   let mapi f (e, r) = (e, f e r)
-  let map2 f (e, r) (e', r') = (E.meet e e', f r r')
-  let long_map2 f (e, r) (e', r') = (E.join e e', f r r')
+  let idempotent_inter f (e, r) (e', r') = (E.meet e e', f r r') (* TODO: does this make sense? *)
+  let nonidempotent_inter f (e, r) (e', r') = (E.meet e e', f r r') (* TODO: does this make sense? *)
+  let idempotent_union f (e, r) (e', r') = (E.join e e', f r r') (* TODO: does this make sense? *)
+  let nonidempotent_union f (e, r) (e', r') = (E.join e e', f r r') (* TODO: does this make sense? *)
   let merge f m1 m2 = failwith "MapDomain.Joined.merge" (* TODO: ? *)
```
michael-schwarz: Does it? 😉

michael-schwarz (approved):

src/domain/mapDomain.ml
sim642: It should also work but I was afraid that the additional partial application causes more closure allocations and double the number of (partial) function applications under the hood. But I have no benchmarks for this.

michael-schwarz: Ok, then maybe leave a comment here that it's like this for a reason, and should not be changed without further benchmarking?

sim642:
> I'm rerunning the rsync benchmark as well to see how it affects the improvements there.

Luckily using the non-idempotent versions of `meet` and `narrow` doesn't hurt rsync analysis at all.

sim642:
Annoyingly, this seems to have broken GobView: https://github.com/codex-semantics-library/patricia-tree/issues/33.

PR #2015, opened 2026-05-04
Use patricia-tree library more

sim642:
This is on top of #2002.

### Changes
1. Remove slow `merge` from `MapDomain` interface altogether. All of its uses have possibly more efficient alternatives in the patricia-tree library.
2. Use `PatriciaMapBot` for other domains than base values.
3. Add lifted versions of `Patricia` functors to `MapDomain`.

### TODO
- [ ] Use patricia-tree also in `SetDomain`? **Not sure what this would improve, sets don't do such value operations.**

### DIFF
--- a/src/analyses/commonPriv.ml
+++ b/src/analyses/commonPriv.ml
@@ -169,7 +169,7 @@ struct
 
   module W =
   struct
-    include MapDomain.MapBot_LiftTop (Basetype.Variables) (MinLocksets)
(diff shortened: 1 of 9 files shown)


sim642:
Unlike #2002, this doesn't seem to have notable impact. On sv-benchmarks there's no difference. On rsync this maybe improves by ~100s which is ~0.4%, but that could also be up to some noise.

Still, I think this is a good idea because it eliminates the inefficient `merge` from `MapDomain` interface, making it much harder to carelessly introduce similar bottlenecks.

PR #2016, opened 2026-05-05
Remove `warn_and_top_on_zero` from `BaseInvariant` for division by zero

sim642:
This is the continuation of #1892 regarding the top-ification aspect.

I tried to just remove it and no tests fail. But I'll see on sv-benchmarks if this actually matters sometime.

### Question

In https://github.com/goblint/analyzer/issues/1892#issuecomment-4054491792:
> Okay, so here specifically the question is what the backward transfer function should be for `c = a / [0;0]`, given some value for `c`, how can one refine `a`? And what about `c = a / [0;1]`...

The strange thing is that the `warn_and_top_on_zero` code is about the second argument, not the first one.

And it looks like it just tries to avoid a definite division by zero in the calculations that follow, e.g. `ID.div a b`. But doing so isn't actually a problem: once we're doing the refinement, we've already forward-evaluated that exact calculation with the zero divisor without issues. I don't think the abstract division operator in int domains should fail because of that.
Also note that the second argument (without top-ification) would still be, e.g., `[0,0]`, not bottom. So it's not triggering `ArithmeticOnIntegerBot` by doing that.

Maybe some git archeology is needed for this code. Perhaps it was added at a time when the division operators in int domains had problems with doing this and `warn_and_top_on_zero` was added to work around them.

### TODO
- [x] sv-benchmarks

### DIFF
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -291,12 +291,6 @@ struct
     (* inverse values for binary operation a `op` b == c *)
     (* ikind is the type of a for limiting ranges of the operands a, b. The only binops which can have different types for a, b are Shiftlt, Shiftrt (not handled below; don't use ikind to limit b there). *)
     let inv_bin_int (a, b) ikind c op =
-      let warn_and_top_on_zero x =
-        if ID.equal_to Z.zero x = `Eq then
-          ID.top_of ikind
-        else
-          x
-      in
       let meet_bin a' b'  = id_meet_down ~old:a ~c:a', id_meet_down ~old:b ~c:b' in
       let meet_com oi = (* commutative *)
         try
@@ -320,8 +314,6 @@ struct
         (refine_by a b, refine_by b a)
       | MinusA -> meet_non ID.add ID.sub
       | Div    ->
-        (* If b must be zero, we have must UB *)
-        let b = warn_and_top_on_zero b in
         (* Integer division means we need to add the remainder, so instead of just `a = c*b` we have `a = c*b + a%b`.
          * However, a%b will give [-b+1, b-1] for a=top, but we only want the positive/negative side depending on the sign of c*b.
          * If c*b = 0 or it can be positive or negative, we need the full range for the remainder. *)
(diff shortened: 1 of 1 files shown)


sim642:
> Maybe some git archeology is needed for this code. Perhaps it was added at a time when the division operators in int domains had problems with doing this and `warn_and_top_on_zero` was added to work around them.

This is essentially reverting ae01aba44cd4167900cb4ec761883cfd3d9fe2e6. Its previous commit added the following test:
https://github.com/goblint/analyzer/blob/df5edaa7537db5084a20e1a205fb06f5c5789f0f/tests/regression/27-inv_invariants/02-bot-during-condition.c#L4-L11
The test doesn't contain any checks though. But it's also not crashing after the revert.

sim642:
On sv-benchmarks this only changes two unknowns into `ArithmeticOnIntegerBot`: https://goblint.cs.ut.ee/results/332-all-level01-pr-2016-after/table-generator-cmp.diff.html#/table?filter=1(0*status*(status(in(EXCEPTION%20%28IntDomain0.ArithmeticOnIntegerBot%29)),category(notIn()))).
The benchmarks seem to trigger `0 / 0` which in itself is a problem with the benchmarks maybe.

I'm not sure it's worth keeping this strange handling for that.
I think the broader issue is what's discussed in #1892: what is the result of dividing by 0? Is it top or is it assumed to not have happened?
Currently, some int domains do one, some do the other. So it matters on which int domains are active. I think it even differs by operation in the same domain (`div` and `rem` in intervals).

To properly fix this, I think we need an option like `sem.int.div-by-zero` with values `assume_top` and `assume_none`. Then all int domains should follow this option and we can be consistent and flexible.

michael-schwarz:
> ArithmeticOnIntegerBot

I think we should aim to get rid of this as much as possible, it is not a sensible message to report to users. Also divisions such as `0/0` can always happen due to over approximation in contexts / digests / ..., so if we accept those we are potentially blocking our path towards analyzing larger programs as that error cannot be recovered from.

sim642:
I agree that `ArithmeticOnIntegerBot` indicates an internal problem that a user should never see.

Since there's no direct benefit from this PR, there's no rush to merge it. If the division by zero story is made consistent in the int domains, this should no longer cause any new exceptions. Because really, if division by zero is assumed to not have happened, then branching on it should be dead and no refinement is necessary.

PR #2017, opened 2026-05-05
Fix `memOutOfBounds` handling of negative pointer offsets

karoliineh:
This PR fixes a bug in `memOutOfBounds` where before-start accesses were classified as safe when the offset was negative and precise enough under `ana.int.interval`, i.e., the bug only manifested when interval domain was turned on. With `ana.int.interval`, Goblint can compute offsets precisely enough that negative-offset accesses no longer stay at top. This issue was found using the dashboard comparison between Goblint and Mopsa and was exposed by the SV-COMP task `list-ext-properties/960521-1_1-1.i`. 

* Added regressions for negative-offset `memOutOfBounds` cases in `tests/regression/74-invalid_deref`.
* Fixed `memOutOfBounds` to warn when the computed byte offset is negative, in addition to the existing past-the-end check.
* Made the `ptr_size_lt_offs` comparison sites handle `IntDomain.ArithmeticOnIntegerBot` consistently by falling back to the existing conservative “could not compare” behavior.
* Refactored the duplicated offset-bound comparison logic into top-level helpers, while preserving the original distinction between:
    * dereference-offset checks using `(size - 1) < offset`
    * pointer-offset checks using `size < offset`

### DIFF
--- a/src/analyses/memOutOfBounds.ml
+++ b/src/analyses/memOutOfBounds.ml
@@ -41,9 +41,7 @@ struct
   let check_deref_offset_bounds ptr_size offs =
     let ptr_size_lt_offs =
       try
-        let one = intdom_of_int 1 in
-        let max_valid_offs = ID.sub ptr_size one in
-        ID.lt max_valid_offs offs
+        ID.le ptr_size offs
       with IntDomain.ArithmeticOnIntegerBot _ -> None
     in
     offs_lt_zero offs, ptr_size_lt_offs
--- a/src/analyses/memOutOfBounds.ml
+++ b/src/analyses/memOutOfBounds.ml
@@ -39,12 +39,11 @@ struct
     with IntDomain.ArithmeticOnIntegerBot _ -> None
 
   let check_deref_offset_bounds ptr_size offs =
-    let ptr_size_lt_offs =
-      try
-        ID.le ptr_size offs
+    let ptr_size_le_offs =
+      try ID.le ptr_size offs
       with IntDomain.ArithmeticOnIntegerBot _ -> None
     in
-    offs_lt_zero offs, ptr_size_lt_offs
+    offs_lt_zero offs, ptr_size_le_offs
 
   let check_ptr_offset_bounds ptr_size offs =
     let ptr_size_lt_offs =
--- a/src/analyses/memOutOfBounds.ml
+++ b/src/analyses/memOutOfBounds.ml
@@ -34,6 +34,27 @@ struct
   let size_of_type_in_bytes typ =
     intdom_of_int (Cilfacade.bytesSizeOf typ)
 
+  let offs_lt_zero offs =
+    try ID.lt offs (intdom_of_int 0)
+    with IntDomain.ArithmeticOnIntegerBot _ -> None
+
+  let check_deref_offset_bounds ptr_size offs =
+    let ptr_size_lt_offs =
+      try
+        let one = intdom_of_int 1 in
+        let max_valid_offs = ID.sub ptr_size one in
+        ID.lt max_valid_offs offs
+      with IntDomain.ArithmeticOnIntegerBot _ -> None
+    in
+    offs_lt_zero offs, ptr_size_lt_offs
+
+  let check_ptr_offset_bounds ptr_size offs =
+    let ptr_size_lt_offs =
+      try ID.lt ptr_size offs
+      with IntDomain.ArithmeticOnIntegerBot _ -> None
+    in
+    offs_lt_zero offs, ptr_size_lt_offs
+
   let rec exp_contains_a_ptr (exp:exp) =
     match exp with
     | Const _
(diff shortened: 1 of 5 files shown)


src/analyses/memOutOfBounds.ml
```diff
@@ -34,6 +34,27 @@ struct
...
+  let offs_lt_zero offs =
+    try ID.lt offs (intdom_of_int 0)
+    with IntDomain.ArithmeticOnIntegerBot _ -> None
+
+  let check_deref_offset_bounds ptr_size offs =
+    let ptr_size_lt_offs =
+      try
+        let one = intdom_of_int 1 in
+        let max_valid_offs = ID.sub ptr_size one in
+        ID.lt max_valid_offs offs
```
sim642: Isn't this the same?
```suggestion
        ID.le ptr_size offs
```

michael-schwarz: Mathematically yes. We might be careful with naming though: `ID.le ptr_size offs` in `ptr_size_lt_offs` is prone to someone "fixing" the typo.

src/analyses/memOutOfBounds.ml
```diff
@@ -34,6 +34,27 @@ struct
...
+      with IntDomain.ArithmeticOnIntegerBot _ -> None
+    in
+    offs_lt_zero offs, ptr_size_lt_offs
+
+  let check_ptr_offset_bounds ptr_size offs =
+    let ptr_size_lt_offs =
+      try ID.lt ptr_size offs
+      with IntDomain.ArithmeticOnIntegerBot _ -> None
+    in
+    offs_lt_zero offs, ptr_size_lt_offs
```
michael-schwarz, line 53: Could you elaborate on why this needs to be handled differently from the things above?

karoliineh: First I didn't give it much thought, I just refactored out the parts that were in the original code, but with the help of Codex, I now know the difference and we also added regtests that would catch a change in these in both directions:

> This follows directly from the C standard distinction between pointer values and dereferenceable positions. Per C11 §6.5.6, pointer arithmetic may produce a pointer one past the end of an object (this is well-defined), but such a pointer must not be dereferenced. So:
>
> For dereferencing, the valid range is [0, size), hence ID.le ptr_size offs (reject offs == size).
> For pointer values, the valid range is [0, size], hence ID.lt ptr_size offs (allow offs == size, reject only offs > size).
>
> and changing either side breaks soundness/precision, I added two regression tests to make this concrete:
>
> 74/36: valid one-past pointer `(buf + 4)` → fails if we switch to `le` in the pointer case (false positive)
> 74/37: dereference at end → missed if we switch to `lt` in the deref case (unsound)
>
> So the difference encodes the standard’s [0, size] vs [0, size) distinction and is intentional.

The unsoundness by changing the deref case to `lt` is also caught by an existing test 74/13, but the other way around was not detectable by the current tests.

tests/regression/74-invalid_deref/36-one-past-pointer.c
```diff
@@ -0,0 +1,12 @@
+// PARAM: --set ana.activated[+] memOutOfBounds --enable ana.int.interval
+#include <stdlib.h>
+#include <stdio.h>
+
+int main(void) {
+  char *buf = malloc(4);
+  char *end;
+  end = buf + 4; //NOWARN
+  printf("%p", (void *) end); //NOWARN
```
michael-schwarz, line 9:
```suggestion
  printf("%p", (void *) end); //NOWARN
  printf("%c", *end); //WARN
```

So we also see that we do indeed warn if this pointer is dereferenced.

michael-schwarz (commented):
Other than the suggestion, LGTM!

src/analyses/memOutOfBounds.ml
michael-schwarz: Thanks for checking!

michael-schwarz (approved):

tests/regression/74-invalid_deref/36-one-past-pointer.c
karoliineh: ha! it seems you just uncovered an unrelated unsoundness with this ;) I suggest we merge this as is and I will add this suggestion on another PR with an appropriate fix.

michael-schwarz: Yeah, sounds good!

src/analyses/memOutOfBounds.ml
sim642: But this is still strange. The memOutOfBounds analysis should only consider dereferencing. If a pointer points one or more past the end should make no difference if it's never dereferenced. It would only matter for the valid-memtrack subproperty which is handled by memLeak (which has many other problems).

sim642: I'll merge this so we can see if it makes any difference in the knightly.
This strangeness already existed before and could be fixed separately.

sim642: This is now delegated to #2029.

Issue #1766, opened 2025-06-25
Spurious warning due to wrong array size from `calloc`


Robotechnic:
I just found out that this program:
```c
#include <assert.h>
#include <stdlib.h>

int main() {
	int *buff = calloc(50, sizeof(int));
	if (buff == NULL) {
		return 1;
	}
	return buff[5];
}
```
Causes goblint to raise an error while it shouldn't. I get:
```
[Error][Behavior > Undefined > ArrayOutOfBounds > PastEnd] Must access array past end
```

It has been mentioned in #702 and #135, but this is not really a precision problem.

Changing the array size to the number of items and the blob size to the size of each item seems to fix the issue with this particular problem. However, it breaks soundness, so it isn't a suitable solution.

PR #2018, opened 2026-05-05
Infer effective length for single-blob arrays

karoliineh:
This PR fixes array out-of-bounds handling for dynamically allocated arrays whose abstract array length is initially represented as a single blob. This was found using the dashboard comparison between Goblint and Mopsa and was exposed by the SV-COMP task `memsafety-cve/libredwg.i`.

Goblint output spurious error messages for the program:
```shell
[Error][Behavior > Undefined > ArrayOutOfBounds > PastEnd] Must access array past end (/home/holterka/sv-benchmarks/c/memsafety-cve/libredwg.i:425:5-425:48)
[Error][Behavior > Undefined > ArrayOutOfBounds > PastEnd] Must access array past end (/home/holterka/sv-benchmarks/c/memsafety-cve/libredwg.i:425:5-425:48)
[Error][Behavior > Undefined > ArrayOutOfBounds > PastEnd] Must access array past end (/home/holterka/sv-benchmarks/c/memsafety-cve/libredwg.i:425:5-425:48)
[Error][Behavior > Undefined > ArrayOutOfBounds > PastEnd] Must access array past end (/home/holterka/sv-benchmarks/c/memsafety-cve/libredwg.i:425:5-425:48)
```

With the offending code being:
```c
unsigned char *getRandomByteStream(int size) {
  unsigned char *randomString = (unsigned char *)calloc(size, sizeof(unsigned char));
  if (randomString == ((void *)0)) {
    printf("Out of memory\n");
    exit(1);
  }
  for (int i = 0; i < size; i++) {
    randomString[i] = __VERIFIER_nondet_uchar();
  }
  return randomString;
```

where `size` was a constant of 50 during the function call, but Goblint's `array_oob_check` considered the array size as 1.

The change computes an effective array length from the blob size divided by the element size, then uses that length before indexed reads and writes. This prevents valid accesses into calloc-allocated arrays from being treated imprecisely.

- Added a regression test for a valid access into a `calloc`-allocated `unsigned char` array.
- Added `effective_array_length` to derive array length from blob size when possible.
- Apply the effective length before array indexed reads.
- Apply the effective length before array indexed writes.

### DIFF
--- a/src/cdomain/value/cdomains/valueDomain.ml
+++ b/src/cdomain/value/cdomains/valueDomain.ml
@@ -900,6 +900,28 @@ struct
     else
       x (* This already contains some value *)
 
+  let effective_array_length (ask: VDQ.t) (array: CArrays.t) (elem_typ: typ) =
+    let ik = Cilfacade.ptrdiff_ikind () in
+    let elem_typ =
+      match Cil.unrollType elem_typ with
+      | TArray (elem_typ, _, _) -> elem_typ
+      | TPtr (elem_typ, _) -> elem_typ
+      | elem_typ -> elem_typ
+    in
+    match CArrays.length array with
+    | Some l when not (ID.is_bot l) && ID.equal_to Z.one l = `Eq -> begin
+        match CArrays.get ~checkBounds:false ask array (None, IndexDomain.of_int ik Z.zero) with
+        | Blob (_, size, _) -> begin
+            try
+              let elem_size = ID.of_int ik (Z.of_int (Cilfacade.bytesSizeOf elem_typ)) in
+              ID.div size elem_size
+            with Cilfacade.TypeOfError _ -> l
+          end
+        | _ -> l
+      end
+    | Some l -> l
+    | None -> ID.top_of ik
+
   (* Funny, this does not compile without the final type annotation! *)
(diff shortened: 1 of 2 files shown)


sim642:
This seems a bit roundabout to try to recover the length after the fact if it's 1. Currently it's being set as 1 at `calloc`:
https://github.com/goblint/analyzer/blob/729eb71c68dfa7300b89d935f8c5b7169f665932/src/analyses/base.ml#L2785

I don't really know why it's 1 there though, instead of the actual length given by the argument to `calloc`. I think this has come up before and for some reason we never changed it.

michael-schwarz:
I think this should be fixed together with https://github.com/goblint/analyzer/issues/1766, if we are already touching this part.

michael-schwarz:
Related: https://github.com/goblint/analyzer/issues/702 https://github.com/goblint/analyzer/issues/135 https://github.com/goblint/analyzer/pull/145

karoliineh:
After spending time investigating the related issues and trying a few alternative approaches, I agree that this PR proposal is only a hacky remedy for one part of a broader problem rather than the right direction. A proper fix would require rethinking how `calloc` allocations, blob/array representations, and allocation-size reasoning interact throughout the base analysis, which makes solving this issue quite complex and is more work than I can prioritize right now. Closing this for the time being.

PR #2022, opened 2026-05-06
Disable OOB checks during reachability calculations

karoliineh:
Reachability computation in `Base.reachable_from_addr` uses the normal value read path via `get_addr -> get_mval -> VD.eval_offset`. For array-backed addresses, this goes through `CArrays.get`, which performs array OOB checks.

That is the wrong behavior for reachability: the analysis is only trying to inspect the immediate contents of a cell to discover embedded pointers, not to report an actual dereference. On zero-length VLAs this triggered `array_oob_check` during function-entry reachability handling. This was found using the dashboard comparison between Goblint and Mopsa and was exposed by the SV-COMP task `ldv-memsafety/ArraysOfVariableLength2.c`

Changes:
* Added a regression extracted from `ldv-memsafety/ArraysOfVariableLength2.c`
* Added an optional `checkBounds` flag to `VD.eval_offset`, threads it through `get_mval` and `get_addr`, and uses `checkBounds=false` specifically in `reachable_from_addr`. Normal semantic reads still keep bounds checks enabled.
* Unrelated to the discovered bug, also added `~checkBounds:false` to all `ValueDomain.CArrays.get` calls within the functions that calculate reachability.

### DIFF
--- a/tests/regression/25-vla/07-zero-arg-reach.c
+++ b/tests/regression/25-vla/07-zero-arg-reach.c
@@ -1,0 +1,16 @@
+// PARAM: --enable ana.arrayoob
+// extracted from SV-COMP task ldv-memsafety/ArraysOfVariableLength2.c
+
+int bar(int b[]) {
+  return 0;
+}
+
+void foo(int n) {
+  int a[n];
+  bar(a); //NOWARN
+}
+
+int main(void) {
+  foo(0);
+  return 0;
+}
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -613,10 +613,15 @@ struct
     | JmpBuf _ -> empty (* Jump buffers are abstract and nothing known can be reached from them *)
     | Mutex -> empty (* mutexes are abstract and nothing known can be reached from them *)
 
+  let reachable_from_value ask (value: value) (t: typ) (description: string) =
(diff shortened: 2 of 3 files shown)


sim642:
I wonder if the whole reachability calculation should just be under `executing_speculative_computations` and the OOB warnings suppressed when speculating, similarly to the overflow ones.

michael-schwarz:
> I wonder if the whole reachability calculation should just be under `executing_speculative_computations` and the OOB warnings suppressed when speculating, similarly to the overflow ones.

That sounds like a good idea!

tests/regression/25-vla/07-zero-arg-reach.c
```diff
@@ -0,0 +1,16 @@
+// PARAM: --enable ana.sv-comp.enabled --enable ana.sv-comp.functions --set ana.specification "CHECK( init(main()), LTL(G valid-deref) )"
```
sim642: The SV-COMP mode does a lot of stuff. Wouldn't it be sufficient to just use `--enable ana.arrayoob` or something?

sim642: I tried it out and it's enough to test the regression, so I changed it.
I also squashed some commits in this PR now to make the history cleaner.

sim642 (approved):

PR #2024, opened 2026-05-07
Account for stored pointer offsets in dereference checks

karoliineh:
This is on top of https://github.com/goblint/analyzer/pull/2017 based on https://github.com/goblint/analyzer/pull/2017#discussion_r3200648384.

Changes:
* Adds the regression from https://github.com/goblint/analyzer/pull/2017#discussion_r3200648384 that exposed the unsoundness
* Fixes dereference bound checks to include offsets already encoded in the pointer's points-to address, instead of only looking at the offset introduced by the current dereference expression
* Refactors the new offset-handling logic into helper functions for readability
* Simplifies `BinOp` handling by passing the original expression through the checks instead of destructuring and reconstructing it
    * I initially removed the match on `BinOp` from `check_binop_exp` altogether because the only place where it is called from is within a match case on `BinOp ((PlusPI | MinusPI | IndexPI), ...)`, but decided to keep it just in case for future, so that it is not called from a wrong place and if it is, would notify with an error message.

This is still not an ideal code structure, as evidenced by the need for comments like:
```
(* Explicit pointer arithmetic is checked separately by check_binop_exp; do not add it here again. *)
```

Obviously the overall analysis workflow is a bit non-straightforward if we have to leave comments like this. But that is an issue for another cleanup PR, at least this should fix the unsoundness itself.

### DIFF
--- a/src/analyses/memOutOfBounds.ml
+++ b/src/analyses/memOutOfBounds.ml
@@ -56,14 +56,6 @@ struct
     try ID.add x y
     with IntDomain.ArithmeticOnIntegerBot _ -> ID.bot_of @@ Cilfacade.ptrdiff_ikind ()
 
-  let is_ptr_arith_binop = function
-    | PlusPI | MinusPI | IndexPI -> true
-    | _ -> false
-
-  let exp_is_explicit_ptr_arith = function
-    | BinOp (binop, _, _, _) -> is_ptr_arith_binop binop
-    | _ -> false
-
   let rec exp_contains_a_ptr (exp:exp) =
     match exp with
     | Const _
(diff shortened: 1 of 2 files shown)


karoliineh:
I've been going back-and-forth with marking this as a draft and ready because I was unsure if this should be merged before the cleanup or after. For now, the fix will be applied during a cleanup that will be similar to the PR https://github.com/goblint/analyzer/pull/1864, using access events for the `memOutOfBounds` analysis as was done for the `useAfterFree` analysis. So, most likely this PR will be replaced by a more extensive one and thus, review effort should not be wasted on this for now.

PR #2026, opened 2026-05-08
`protection-atomic`: Properly handle unlock of atomicity mutex

### LINKED ISSUE #2025: `protection-atomic` Privatization: Unsound because `P` is not shrunk on `__VERIFIER_atomic_end()`
During work on ghost witness validation, I noticed some inconsistency where the P set and local map in the protection-based analysis are not properly cleared after an `__VERIFIER_atomic_end`. This affects `master` too.

We then end up with a state where `g` is `P` so we only consult this value, but do not record writes in the local component, and only side-effect to the `unprotected` invariant. Thus, the assert below succeeds.

~~~C
// PARAM: --disable warn.race --enable ana.sv-comp.functions --set ana.path_sens[+] threadflag --set lib.activated[+] sv-comp --set ana.base.privatization protection-atomic
#include<pthread.h>
#include<stdlib.h>
#include<goblint.h>
extern void __VERIFIER_atomic_begin();
extern void __VERIFIER_atomic_end();

int g;

void fun() {
}

int main(void) {
    pthread_t thread;
    pthread_create(&thread, NULL, (void*)fun, NULL);

    __VERIFIER_atomic_begin();
    g = 1;
    __VERIFIER_atomic_end();

    g = 5;
    __goblint_check(g <= 1); //UNKNOWN!

    return 0;
}
~~~

I'll make a fix on our ghost branch and then see about porting this back. It's quite amazing this has never caused us to be unsound....

michael-schwarz:
On unlock of atomicity mutex, iterate through local state to ensure things that are no longer protected are removed from `P` and `st.priv`.

Closes #2025 via cherry-pick from https://github.com/goblint/analyzer/commits/ghosts/.

### DIFF
--- a/src/analyses/basePriv.ml
+++ b/src/analyses/basePriv.ml
@@ -980,7 +980,7 @@ struct
     let atomic = Param.handle_atomic && LockDomain.MustLock.equal m (LockDomain.MustLock.of_var LibraryFunctions.verifier_atomic_var) in
     (* TODO: what about G_m globals in cpa that weren't actually written? *)
     CPA.fold (fun x v (st: BaseComponents (D).t) ->
-        if is_protected_by ask m x then ( (* is_in_Gm *)
+        if (atomic && is_global ask x && not (VD.is_immediate_type x.vtype)) || is_protected_by ask m x then ( (* is_in_Gm *)
           (* Only apply sides for values that were actually written to globals!
              This excludes invariants inferred through guards. *)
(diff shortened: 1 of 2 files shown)


sim642 (approved):
It looks like at some point `is_protected_by` included the atomicity mutex check in itself but this was again removed in 696a35ff9a232bb48353e6ddefeca5398e92d6a2.

PR #2027, opened 2026-05-08
Use access events in MemOutOfBounds analysis

karoliineh:
This PR refactors `memOutOfBounds` to use access events, similarly to the `useAfterFree` refactoring in #1864.

Instead of recursively traversing expressions inside `memOutOfBounds` to rediscover what is accessed, the analysis now consumes `Events.Access` emitted by `accessAnalysis`. This makes the analysis depend on the shared access infrastructure and removes the local implicit/explicit dereference traversal logic. It also refactors and simplifies the code in other aspects.

This PR replaces #2024. The stored-pointer-offset dereference fix (https://github.com/goblint/analyzer/pull/2017#discussion_r3200648384) from #2024 is included here as part of the event-based cleanup instead of being merged separately.

Changes:
- Register `memOutOfBounds` as depending on `access`.
- Enable single-threaded access event emission when `memOutOfBounds` is active.
- Replace expression/lvalue traversal in `memOutOfBounds` with handling of `Events.Access`.
- Split OOB checks into clearer cases:
  - pointer dereference checks
  - pointer value checks
  - counted library accesses such as `memset`/`memcpy` (kept close to as was originally)
- Account for stored pointer offsets when checking dereference accesses.
- Clean up diagnostic reporting through a shared OOB reporting helper.
- Update dune test outputs for changed diagnostic wording.

I ran the portfolio on 074e4d7db486eea10b8427dfc1e2150150ee3b5e for `memsafety` tasks with 15s timeouts (as there were only 5 tasks that took > 12s to solve in the C.MemSafety category from SV-COMP 2026). The results were sound.

### Open questions

2. Check whether access-event address information can be used more directly in dereference checks without double-counting offsets.
3. Figure out if and why the `check_ptr_value_access` case is even needed to get rid of the `lt` and `le` things (https://github.com/goblint/analyzer/pull/2017#discussion_r3201502823)
1. Investigate whether the local offset conversion can be replaced by `PreValueDomain.Offs.to_index` without losing precision. Currently, the index case is less precise in the version `Offsets.to_index` and using this, some OOB tests start to fail due to lack of precision, but adding a case of
```ocaml
| Some typ -> (None, idx_of_int (Cilfacade.bytesSizeOf typ))
```
in between the 
```ocaml
| Some TArray(item_typ, _, _) -> ...
| _ -> (None, Idx.top ())
```
makes some other (unrelated) tests fail

I also didn't resolve the TODOs that were already there about deduplicating some functions with base, namely:
- `get_size_of_ptr_target`  (* TODO: deduplicate with base *)
- `convert_offset` (* TODO: Some duplication with convert_offset in base.ml, unclear how to immediately get more reuse *)


src/analyses/accessAnalysis.ml
```diff
@@ -29,7 +29,7 @@ struct
   let init _ =
     collect_local := get_bool "witness.yaml.enabled" && get_bool "witness.invariant.accessed";
     let activated = get_string_list "ana.activated" in
-    emit_single_threaded := List.mem (ModifiedSinceSetjmp.Spec.name ()) activated || List.mem (PoisonVariables.Spec.name ()) activated || List.mem (UseAfterFree.Spec.name ()) activated (* TODO: some of these don't have access as dependency *)
+    emit_single_threaded := List.mem (ModifiedSinceSetjmp.Spec.name ()) activated || List.mem (PoisonVariables.Spec.name ()) activated || List.mem (UseAfterFree.Spec.name ()) activated || List.mem (MemOutOfBounds.Spec.name ()) activated (* TODO: some of these don't have access as dependency *)
```
michael-schwarz, line 32: Can we do this TODO here? Seems simple.

PR #2029, opened 2026-05-12
Remove incorrect one-past-end distinction in memOutOfBounds

sim642:
This addresses the finding from https://github.com/goblint/analyzer/pull/2017#discussion_r3199758804.

The memOutOfBounds analysis warns about invalid dereferences. Dereferencing one past end is never valid.

While adding the test I discovered so much more wrong about memOutOfBounds which in some cases completely ignores the offset altogether. I found this because initially I had hard time constructing a test where the off-by-one aspect even made a difference.

There's also a whole other rabbit hole that comes up with this test: `%p` prints the pointer itself (not what it points to), so there's never a need to warn about out of bounds for that because it's not even dereferenced. Thus, the added tests use `%s` which actually dereferences. However, we don't have an analysis for the format strings, so we won't be able to not warn for `%p`. Currently the analysis did warn there but for three wrong reasons combined.

This is orthogonal to #2027: the point here is to simplify the existing thing (and make it correct!) such that the refactoring there doesn't need to preserve an artificial and incorrect distinction.

### DIFF
--- a/src/analyses/memOutOfBounds.ml
+++ b/src/analyses/memOutOfBounds.ml
@@ -38,19 +38,12 @@ struct
     try ID.lt offs (intdom_of_int 0)
     with IntDomain.ArithmeticOnIntegerBot _ -> None
 
-  let check_deref_offset_bounds ptr_size offs =
+  let check_offset_bounds ptr_size offs =
     let ptr_size_le_offs =
       try ID.le ptr_size offs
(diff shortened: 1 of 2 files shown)


sim642:
According to an sv-benchmarks run with level01, 60s and 1GB, this (surprisingly) makes no difference to verdicts. This probably explains why this unnecessary distinction flew under the radar.

PR #2031, opened 2026-05-13
Force dune version 3.19.1 in make dev

arkocal:
Installing `ocaml-lsp-server` will pull the dependency `dune-rcp`, which in turn (when unconstrainted) upgrades dune, leading to compatibility issues. The easy temporary fix is to force the version here.


sim642:
Closing in favor of #2036.

PR #2037, opened 2026-05-19
Fix flaky 56-witness/66-ghost-alloc-lock on MacOS

sim642:
Recently some nightly unlocked MacOS jobs (macos-14, macos-15-intel) have started to fail because of this test.

The problem is that even with the ghost stripping script, there's flakiness in the order of Goblint output. Depending on the IDs/names of the variables the YAML witness entries can be in different order.
The normalization done by `yamlWitnessStrip` is not enough for this because it matters which `ALLOC_VARn_LOCKED` variable gets related to which `gn` variable.

This changes the helper script to pick the variables based on which `gn` variable they relate to in the invariant.
Passing it through `yamlWitnessStrip` after that should make it even more stable.


sim642:
This fixed the nightly MacOS unlocked CIs but https://github.com/goblint/analyzer/pull/2002#issuecomment-4487313768 still causes a failure.

PR #2038, opened 2026-05-19
Consider `Calloc` and `Realloc` for autotuners

sim642:
While debugging https://github.com/goblint/analyzer/pull/2030#issuecomment-4448496014 I noticed that the mallocWrappers autotuner doesn't pick up on the wrapper of `calloc`. Despite the name, it makes sense to do so (and with `realloc`). The mallocWrapper analysis also already considers them.

After reviewing all pattern matchings for these allocation kinds, I also found the same being missing from the allocation loop unrolling autotuner.

### TODO
- [x] sv-benchmarks

### DIFF
--- a/src/analyses/c2poAnalysis.ml
+++ b/src/analyses/c2poAnalysis.ml
@@ -212,7 +212,7 @@ struct
               let cc = begin match desc.special exprs with
                 | Malloc _
                 | Calloc _
-                | Alloca _ -> (* TODO: Why not Realloc? *)
+                | Alloca _ -> (* No Realloc because it may return the same pointer. *)
                   add_block_diseqs cc lterm
                 | _ -> cc
               end
--- a/src/analyses/c2poAnalysis.ml
+++ b/src/analyses/c2poAnalysis.ml
@@ -212,7 +212,7 @@ struct
               let cc = begin match desc.special exprs with
                 | Malloc _
                 | Calloc _
-                | Alloca _ ->
+                | Alloca _ -> (* TODO: Why not Realloc? *)
                   add_block_diseqs cc lterm
                 | _ -> cc
               end
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2833,6 +2833,7 @@ struct
           let lv_addr = eval_lv ~man st lv in
           let blob_set = Option.map_default (fun heap_addr -> [heap_addr, TVoid [], heap_val]) [] heap_var in
           (* TODO: free (i.e. invalidate) old blob if successful? *)
+          (* TODO: realloc may return the same pointer, is this handled correctly? *)
           set_many ~man st ((lv_addr, Cilfacade.typeOfLval lv, Address addr) :: blob_set)
         ) st lv
     | Free ptr, _ ->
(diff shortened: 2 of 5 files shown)


src/analyses/c2poAnalysis.ml
```diff
@@ -212,7 +212,7 @@ struct
               let cc = begin match desc.special exprs with
                 | Malloc _
                 | Calloc _
-                | Alloca _ ->
+                | Alloca _ -> (* TODO: Why not Realloc? *)
```
sim642: @michael-schwarz Is there a reason C2PO doesn't account for `realloc` or is this probably just an oversight?

sim642:
On sv-benchmarks with level01, 60s and 1GB this already has interesting effects: https://goblint.cs.ut.ee/results/340-all-level01-pr-2038-after/table-generator-cmp.diff.html#/table?filter=0(0*status*(status(notIn(FAILED)),category(notIn()))).
To summarize:
* 8 trues go to timeout/OOM
* 10 timeouts/OOMs go to true, most by significant margin
* 3 unknowns go to true
  * 1 of them is **unsound**, which suggests an underlying issue because more malloc wrappers should not have that effect on its own.

src/analyses/c2poAnalysis.ml
michael-schwarz: So here we add block disequalities that say that things from different blocks may never have the same address (see https://dl.acm.org/doi/epdf/10.1145/3689609.3689994).

I have not thought about happens when someone does `realloc` (specifically for the case the memory is kept in-place, e.g., as the new size was smaller). Are all old pointers invalid then?

So this is something one would have to consider before adding block disequalities to C2PO.

michael-schwarz (approved):
Looks reasonable!

sim642:
> * 1 of them is **unsound**, which suggests an underlying issue because more malloc wrappers should not have that effect on its own.

I looked into it and according to manual inspection the branches that became dead due to this change seemed to actually be dead. I then looked at SV-COMP 2026 results and found out that the task was invalidated. Turns out Mopsa already had an issue with this:
https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/a45aa8f2eae1325c6c555c945ea721c023dde6f3/c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-32_7a-drivers--pcmcia--pcmcia_rsrc.ko-entry_point.cil.out.yml. Apparently, I never updated sv-benchmarks on our server. So there should be no actual problem in Goblint.

src/analyses/c2poAnalysis.ml
sim642: I guess not adding the disequalities would be the right thing, although it creeps into some gray area. The standard says that the given pointer is deallocated, so it shouldn't be dereferenced any more but on the level of pointer equalities (which don't dereference that), there's also no guarantee for the new pointer to be distinct. If one checks for their equality and then dereferences the old one, I guess technically that'd be OK but that's not a particularly useful thing to do.

I'll update the comment to say that this is should be intentional. And it's definitely sound to not add the disequalities.

PR #2050, opened 2026-06-01
Pin batteries with hashconsing performance improvements

sim642:
This pins https://github.com/ocaml-batteries-team/batteries-included/pull/1149.

The performance improvement is quite significant: https://github.com/ocaml/ocaml/issues/13733#issuecomment-4572552390 (first plot). That is on OCaml 5.5, but the improvements are very similar on OCaml 4.14 as well.


michael-schwarz (approved):
The AI comments look meaningful, I think if we add those, we're good to merge!

sim642:
Out of interest I also tried this on rsync because hashconsing was a performance problem there. However, this doesn't noticeably improve the situation there: after ~12 hours things hadn't finished and according to `perf` the dominating thing is still `equal` under `hashcons`.

PR #2054, opened 2026-06-09
Remove option `dbg.dump`

sim642:
It's described as "Dumps the results to the given path" and it is some kind of strange override for `outfile` ("File to print output to.").


michael-schwarz (approved):
First I heard of this feature. If no one knows or uses it, I think it's safe to remove.

PR #2055, opened 2026-06-15
Mutex ghost

Ang9876:
- Mutex Ghost Checking
- Invariant Validation using Mutex Ghost
- Test Cases


tests/regression/19-ghost-witness/50-mutex-ghost-valid-basic.c
```diff
@@ -0,0 +1,45 @@
+#include <pthread.h>
```
michael-schwarz, line 2: The system checks that each test contains some check. As here all tests are done via CRAM, we need to add a line at the top

```
// CRAM
```

to make these scripts happy. Probably best to do this with AI, as it will affect all tests and all line numbers (but it in very simple way, namely just increasing them).

 After adding these, we will see if the cram tests also succeeds in the CI.

src/analyses/mutexGhost.ml
```diff
@@ -0,0 +1,181 @@
...
+
+    let isGhostAtomic (x, _, _) = x
+    let mutex (_, y, _) = y
+    let ghosts (_, _, z) = z
+    let init = (false, Mutex.bot(), MayGhostVarSet.bot())
+    let create_annotation = (true, Mutex.bot(), MayGhostVarSet.bot())
+    let create_lock_mutex l = (true, `Lifted (l, true), MayGhostVarSet.bot())
+    let create_unlock_mutex l = (true, `Lifted (l, false), MayGhostVarSet.bot())
+    let create_ghosts v = (true, Mutex.bot(), MayGhostVarSet.singleton v)
+  end
```
michael-schwarz, line 33: It may be helpful to have some comments of what these components are here.

src/analyses/mutexGhost.ml
```diff
@@ -0,0 +1,181 @@
...
+
+  let eval_const man e =
+    match man.ask (Queries.EvalInt e) with 
+    | `Lifted n -> IntDomain.IntDomTuple.to_int n
+    | _ -> None
+
+  let is_one man rval = 
+    match eval_const man rval with 
+    | Some value -> Z.equal value (Z.one)
+    | _ -> false
```
michael-schwarz:
```suggestion
  let is_one man rval = GobOption.exists (Z.equal Z.one) (eval_const man rval)
```

src/analyses/mutexGhost.ml
```diff
@@ -0,0 +1,181 @@
...
+
+  let is_zero man rval = 
+    match eval_const man rval with 
+    | Some value -> Z.equal value (Z.zero)
+    | _ -> false
+
+  let event (man : (D.t, G.t, C.t, V.t) man) e oman : D.t =
+    let verifier_atomic_instrument_addr = LockDomain.Addr.of_var LibraryFunctions.verifier_atomic_instrument_var in
+    (* TODO: MustBeGhostAtomic *)
```
michael-schwarz: What does this TODO mean? Is it done now?

src/analyses/mutexGhost.ml
```diff
@@ -0,0 +1,181 @@
...
+      )
+    | _ ->
+      man.local
+
+  let assign (man :(D.t, G.t, C.t, V.t) man) (lval : lval) (rval : exp) : D.t =
+    if !AnalysisState.global_initialization then
+      man.local
+    else
+      match lval with
+      | Var var, NoOffset when YamlWitness.VarSet.mem var !(YamlWitness.ghostVars) ->
```
michael-schwarz, line 128: Good to leave a comment here that these ghostVars only appear in this shape.

src/analyses/mutexGhost.ml
```diff
@@ -0,0 +1,181 @@
...
+        let is_one = is_one man rval in 
+        let is_zero = is_zero man rval in
+        let lock, is_lock, is_unlock = match (D.mutex man.local) with 
+          | `Lifted (lock, true) -> Some lock, true, false 
+          | `Lifted (lock, false) -> Some lock, false, true
+          | _ -> None, false, false
+        in
+        let valid = (D.isGhostAtomic man.local) && (is_one || is_zero) &&
```
michael-schwarz:
``` 
(is_one || is_zero)
```

looks like it is subsumed by the other checks?

src/analyses/mutexGhost.ml
```diff
@@ -0,0 +1,181 @@
...
+                    ((is_zero && is_unlock) || (is_one && is_lock)) in 
+        if valid then 
+          ( (match lock with 
+                | Some lock -> 
+                  man.sideg (V.ghost var) (G.create_mutex (`Lifted lock))
+                | None -> ());
+            (D.join man.local (D.create_ghosts var)))
```
michael-schwarz: Can something be valid and not a lock associated with it?

michael-schwarz: It looks like by matching on the non-lifted cases first, the code could be made easier maybe?

michael-schwarz (changes requested):
Thanks for the PR, this looks like it's 95% of the way there!
I left some comments on how to get the CI to run on this in a meaningful way and some questions about where code could be simplified. 

Hopefully once this is done we can run this on the whole SV-COMP suite to see if we spot any issues.

src/analyses/mutexGhost.ml
Ang9876: Yes, it is resolved by the new `verifier_atomic_instrument_var`

src/analyses/mutexGhost.ml
Ang9876: Yes. I have removed it.

src/analyses/mutexGhost.ml
Ang9876: A ghost update is valid only when a mutex operation is associated.

src/dune
```diff
@@ -6,7 +6,7 @@
 (library
   (name goblint_lib)
   (public_name goblint.lib)
-  (modules :standard \ goblint goblint_memtrace privPrecCompare apronPrecCompare messagesCompare)
+  (modules :standard \ goblint privPrecCompare apronPrecCompare messagesCompare)
```
michael-schwarz: Removing this means the CI will fail, it needs to stay here.

michael-schwarz:
The failing indentation CI job is most likely due to force-pushes.

PR #2057, opened 2026-06-16
 Improve memOutOfBounds precision by querying size/length of particular variable

sim642:
This is a continuation of #2030.

It is necessary to make the analysis relational between the address base size and the offset (PR to come) but already on its own reveals some issues.


sim642:
On sv-benchmarks with level02 and 90s, this makes no difference to verdicts.
Also this makes no difference to the dashboard checks comparison.
But that's expected because the values are joined currently anyway. This just avoids joining the sizes multiple times with themselves.

PR #2060, opened 2026-06-18
Remove `EvalValue` queries from memOutOfBounds

sim642:
Doing an `EvalValue` query to look at the `Address` case is just a roundabout way of querying `MayPointTo`.

`check_unknown_addr_deref` is redundant because if it can be an unknown pointer, then the general per-addr logic already checks it anyway.

### DIFF
--- a/src/analyses/memOutOfBounds.ml
+++ b/src/analyses/memOutOfBounds.ml
@@ -167,24 +167,6 @@ struct
     PreValueDomain.Offs.to_index (convert_offset offs)
 
 
-  let check_unknown_addr_deref man ptr =
-    let may_contain_unknown_addr =
-      match man.ask (Queries.EvalValue ptr) with
-      | a when not (Queries.VD.is_top a) ->
-        begin match a with
-          | Address a -> ValueDomain.AD.may_be_unknown a
-          | _ -> false
-        end
-      (* Intuition: if ptr evaluates to top, it could potentially evaluate to the unknown address *)
-      | _ -> true
-    in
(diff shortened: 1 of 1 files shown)


src/analyses/memOutOfBounds.ml
```diff
@@ -204,7 +179,7 @@ struct
 
   let rec check_lval_for_oob_access man ?(is_implicitly_derefed = false) lval =
     (* If the lval does not contain a pointer or if it does contain a pointer, but only points to string addresses, then no need to WARN *)
-    if (not @@ lval_contains_a_ptr lval) || ptr_only_has_str_addr man (Lval lval) then ()
+    if (not @@ lval_contains_a_ptr lval) || ptr_only_has_str_addr man (Lval lval) then () (* TODO: why are StrPtrs special? *)
```
sim642, line 182: This is strange in the same way as #2030: if there's a mix of string pointers and something else, then the string pointers are checked below in the general logic.
Since #2058 that happens per-addr, so the same handling should probably happen below. Although I'm a bit confused about the complete ignoring.

sim642: This is delegated to #2062 as it requires replacing the current unsound behavior with something sufficiently precise to not lose most SV-COMP valid-memsafety points.

sim642:
On sv-benchmarks with level02 and 90s, this makes no difference to verdicts or dashboard checks comparison.
This confirms that the removed top pointer warnings were redundant.

PR #2061, opened 2026-06-18
BSc thesis "Coloring Data Race Interference Graphs"

sim642:
This is the implementation for an older BSc thesis in Tartu (in Estonian): https://thesis.cs.ut.ee/353b5994-7442-4681-989f-ea3e62ac21c2?language=en

I haven't reviewed the code myself, so this probably needs some cleanup first. I'm opening the PR such that this isn't completely forgotten.

### TODO
- [x] Figure out self-race situation.
- [x] Test coloring algorithms.
- [x] Clean up/optimize coloring algorithms?


src/analyses/raceAnalysis.ml
```diff
@@ -400,6 +391,9 @@ struct
         (Pretty.dprintf "total memory locations: %d" total, None);
       ];
     )
+    else
+      ();
+    ()
```
michael-schwarz: This should not be needed.

src/domains/access.ml
```diff
@@ -594,16 +644,54 @@ let incr_summary ~safe ~vulnerable ~unsafe grouped_accs =
...
+      Some (InterferenceGraph.Coloring.color_with (module InterferenceGraph.Coloring.Dsatur) graph)
+    | Some "rlf" ->
+      let graph = InterferenceGraph.of_accesses accs in
+      Some (InterferenceGraph.Coloring.color_with (module InterferenceGraph.Coloring.Rlf) graph)
+    | Some "optimal" ->
+      let graph = InterferenceGraph.of_accesses accs in
+      Some (InterferenceGraph.Coloring.color_with (module InterferenceGraph.Coloring.Optimal) graph)
+    | Some _ -> None
+  in
+  let msgs ?coloring race_accs =
```
michael-schwarz: There should be a way to do this once instead of matching for every `accs`?

src/config/options.schema.json
```diff
@@ -2463,6 +2456,13 @@
           "type": "integer",
           "default": 0
         },
+        "race-coloring": {
+          "title": "warn.race-coloring",
+          "description": "Group race warnings by graph coloring (none, greedy, dsatur, rlf, optimal).",
+          "type": "string",
+          "enum": ["none", "greedy", "dsatur", "rlf", "optimal"],
+          "default": "none"
+        },
```
michael-schwarz, line 2472: There should be some explanation of what this would achieve from a user-level perspective.

src/domains/access.ml
```diff
@@ -484,6 +485,55 @@ struct
...
 end
 
+module InterferenceGraph = struct
+  module Vertex = struct
+    type t = A.t
+
+    let compare = A.compare
+    let hash = A.hash
+    let equal = A.equal
+  end
```
michael-schwarz: Why can vertices not be `A` directly?

sim642:
@michael-schwarz Are you interested in reviewing this again or could this just be merged after my cleanups?

This still includes all the graph coloring algorithms from the student (which according to his experiments perform very similarly) but they're all in a completely separate module so this shouldn't be a burden.
I'd like to do updated experiments on my own (especially because I added handling of self-races and possibly fixed small issues in the algorithms) before possibly getting rid of some of the algorithms.

michael-schwarz:
I'll try to take another look tomorrow if it can wait until then.

src/domains/access.ml
```diff
@@ -594,16 +594,83 @@ let incr_summary ~safe ~vulnerable ~unsafe grouped_accs =
...
+    let accs_list = AS.elements accs in
+    let rec loop = function
+      | [] -> ()
+      | a :: rest as l ->
+        List.iter (fun b ->
+            if may_race a b then
+              add_edge graph a b
+          ) l;
+        loop rest
+    in
```
michael-schwarz, line 613: Why not formulate the outer thing as a `List.iter` too?

src/domains/access.ml
```diff
@@ -594,16 +594,83 @@ let incr_summary ~safe ~vulnerable ~unsafe grouped_accs =
...
+    let acc_msg ?(indent="") A.{conf; kind; node; exp; acc} =
+      let doc = dprintf "%s%a with %a (conf. %d)  (exp: %a)" indent AccessKind.pretty kind MCPAccess.A.pretty acc conf d_exp exp in
       (doc, Some (Messages.Location.Node node))
     in
-    AS.elements race_accs
-    |> List.map h
+    match coloring_module with
+    | lazy None ->
+      AS.elements race_accs
+      |> List.map acc_msg
```
michael-schwarz, line 643: Interesting, this feature has apparently been in OCaml sind way back when, but I have thus far never come across it.

tests/unit/util/ocamlgraph/coloringTest.ml
```diff
@@ -0,0 +1,106 @@
+open OUnit2
```
michael-schwarz, line 1: Should those not live in `util/ocamlgraph` given how that is structured as more or less its own package.

michael-schwarz (approved):
LGTM!

Question: Should there not also be some cram(?) tests to actually test this feature end-to-end? Currently there are only tests that test the generic coloring algorithm but none that apply it to the problem of races.

src/domains/access.ml
sim642: The inner one iterates over the tail of the list (with the current element included for self-races) at each outer iteration. `List.iter` doesn't give access to all the intermediate tails.
This avoids checking each pair twice because `may_race` is symmetric.

sim642:
> Question: Should there not also be some cram(?) tests to actually test this feature end-to-end? Currently there are only tests that test the generic coloring algorithm but none that apply it to the problem of races.

I added a bunch of cram tests for existing regression test programs where the coloring does something possibly interesting.

sim642:
> I added a bunch of cram tests for existing regression test programs where the coloring does something possibly interesting.

Hmm, the problem is now that they don't pass on MacOS, probably due to the `vid`s being somehow different. The graph coloring algorithms are not randomized but they're also not fully stable in that sense.

sim642:
> Hmm, the problem is now that they don't pass on MacOS, probably due to the `vid`s being somehow different. The graph coloring algorithms are not randomized but they're also not fully stable in that sense.

Even though `vid`s are different because the stuff in the headers is different, the relative order of non-header `vid`s should be the same. So `compare` and `Set` shouldn't be affected.
The problem was that `hash` and thus `Hashtbl` is.

I now fixed the algorithms to make them deterministic: they all compare vertices/accesses at some point to pick the preferred one based on some criteria. If all the criteria are the same, there's now a final comparison on the `compare` of the vertices themselves, which should make them deterministic in this sense.

I also scattered `--enable warn.deterministic` into the new tests for good measure.

PR #2075, opened 2026-07-16
Remove unused StackTrace Analysis

michael-schwarz:


src/analyses/stackTrace.ml
```diff
@@ -1,56 +0,0 @@
...
-  let startstate v = D.bot ()
-  let exitstate  v = D.top ()
-
-  let threadenter man ~multiple lval f args =
-    [D.push !Goblint_tracing.current_loc man.local]
-end
-
-
-module Spec1 = Spec (StackDomain.Dom1) (struct let name = "stack_trace" end)
-module Spec2 = Spec (StackDomain.Dom2) (struct let name = "stack_trace_set" end)
```
sim642, line 52: I guess this also makes `StackDomain` redundant.

src/config/options.schema.json
```diff
@@ -360,7 +360,7 @@
           "description": "List of context-insensitive analyses. This setting is ignored if `ana.ctx_sens` contains elements.",
           "type": "array",
           "items": { "type": "string" },
-          "default": [ "stack_loc", "stack_trace_set" ]
```
sim642, line 363: Not sure if I ever realized, but having the old stack trace domains in `ctx_insens` seems strange but whatever.

sim642 (commented):
Just for reference: the newer versions are `call_string` and `call_site`.

src/config/options.schema.json
michael-schwarz: I think the idea was that the path-sensitivity already does the heavy lifting?

sim642: But these analyses didn't define any path-sensitivity. But at this point it doesn't matter anymore.

sim642 (approved):

Issue #1801, opened 2025-08-04
memOutOfBounds analysis causes integer overflow


Robotechnic:
On the following code with the last version of Goblint and `--ana.int.interval_set true`:

```c
#include <stdlib.h>

int main() {
	int **buffer;
	buffer = malloc(5 * sizeof(int *));
	
	for (int i = 0; i < 5; i++) {
		for (int j = 0; j < 5; j++) {
			int x = i * 5 + j;
			buffer[i][j] = x;
		}
	}
	return 0;
}
```
I don't encounter any issues apart from a `May dereference NULL pointer` warning. However, if I add `--set "ana.activated[+]" memOutOfBounds`, then I get integer overflow errors on line 11 (`buffer[i][j] = x;`).

This is not the same issue as in #1767 because here it was a return value problem.

sim642:
We've added `executing_speculative_computations` in many places, but not to `memOutOfBounds`, which unfortunately duplicates a lot of the logic on pointer arithmetic, etc.
However, I think this shouldn't be a problem for us in SV-COMP because valid-memsafety and no-overflow are separate properties. And we shouldn't be activating `memOutOfBounds` for no-overflow analysis anyway. This is probably why we haven't noticed the spurious warnings in SV-COMP runs.

PR #2077, opened 2026-07-21
Fix memOutOfBounds address offset calculation

### LINKED ISSUE #2069: Strange out of bounds messages in sv-benchmarks/c/memsafety-cve/ppp
The dashboard comparisons for #2066 revealed the PR having the following difference:
```json
      {
        "kind": "PrecisionAsymmetry",
        "title": "Invalid memory access",
        "range": {
          "file": "/mnt/goblint-svcomp/benchexec/sv-benchmarks/c/memsafety-cve/ppp.i",
          "start": { "line": 492, "column": 16 },
          "end": { "line": 492, "column": 37 }
        },
        "from_po1": [
          {
            "kind": "safe",
            "messages": "",
            "range": {
              "file": "/mnt/goblint-svcomp/benchexec/sv-benchmarks/c/memsafety-cve/ppp.i",
              "start": { "line": 492, "column": 16 },
              "end": { "line": 492, "column": 37 }
            }
          }
        ],
        "verdict_po1": "safe",
        "from_po2": [
          {
            "kind": "warning",
            "messages": "Size of pointer is 136 (in bytes). It is offset by 136 (in bytes) due to pointer arithmetic. Memory out-of-bounds access must occur",
            "range": {
              "file": "/mnt/goblint-svcomp/benchexec/sv-benchmarks/c/memsafety-cve/ppp.i",
              "start": { "line": 492, "column": 16 },
              "end": { "line": 492, "column": 37 }
            }
          }
        ],
        "verdict_po2": "warning"
      },
```

The new warning indicates a definite (_must_) out of bounds access which doesn't seem right, I think it could also be in bounds. We might be comparing something incorrectly because I don't see how we could have a definite offset 136 there.
The old safe result might also be unsound because the program is supposed to have out of bounds accesses. I don't think we should be able to prove that safe.

sim642:
This is on top of #2066.
Closes #2069.

The old `offs_to_idx` in memOutOfBounds used the size of the entire array as the size of one element, causing it to massively overestimate.
This PR delegates to `Offset.to_index` which already handles arrays correctly, also reducing code duplication.

### TODO
- [x] Renumber test.
- [x] New OVD comparison. (Unchanged)


sim642:
This changes one overall verdict in sv-benchmarks: sv-benchmarks/c/ldv-regression/test23-2 goes from `unknown` to `true`.
The change to OVD checks:
| Before\After | Warning | Error |  Safe | Unreached |
| --------- | ------: | ----: | ----: | --------: |
| Warning   |    5845 |     0 |     2 |         0 |
| Error     |       0 |    68 |     0 |         0 |
| Safe      |       0 |     0 | 17322 |         0 |
| Unreached |       0 |     0 |     0 |         0 |

The two checks it changes are exactly that and the ppp one from #2069.
That's surprisingly few considering it's a huge imprecision in array indexing. I guess in many cases it's covered up by other imprecision, e.g., the offset range still going to "infinity".

`Offset.to_index` has another difference as well: it properly speculates for the offset calculations. I should check the overflow checks comparison as well.

sim642:
> `Offset.to_index` has another difference as well: it properly speculates for the offset calculations. I should check the overflow checks comparison as well.

For reference:

| Before\After | Warning | Error |  Safe | Unreached |
| --------- | ------: | ----: | ----: | --------: |
| Warning   |     741 |     0 |    37 |        84 |
| Error     |       0 |     0 |     0 |         0 |
| Safe      |       0 |     0 |  5799 |      3917 |
| Unreached |       0 |     0 |     0 |         0 |

All of these spurious overflow checks manifested as warnings but didn't affect us in SV-COMP: #1801.

PR #2095, opened 2026-08-08
Allow $schema in configuration files

### LINKED ISSUE #1011: Allow $schema field in config file
Lots of text editors support adding a `$schema` field to a JSON file to define the JSON schema that should be used to validate the file (see for example https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings).
However, it seems that currently Goblint refuses to parse goblint.json config files that have a `$schema` field. It would be nice if Goblint just ignored the `$schema` field so that it can be safely used.

ColumbusLabs:
Closes #1011.

Root-level `$schema` fields are editor annotations rather than Goblint options. This removes them before validating and merging configuration input, while leaving validation of every real option unchanged.

The unit test also verifies that the annotation is not retained in Goblint's active configuration.

Tested with `opam exec -- dune runtest tests/unit` (3,379 tests).

### DIFF
--- a/src/config/gobConfig.ml
+++ b/src/config/gobConfig.ml
@@ -395,6 +395,11 @@ struct
       Printexc.raise_with_backtrace e bt
 
   let merge json =
+    (* $schema is an editor annotation, not a Goblint option. *)
+    let json = match json with
+      | `Assoc fields -> `Assoc (List.filter (fun (name, _) -> name <> "$schema") fields)
+      | json -> json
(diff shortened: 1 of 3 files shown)


sim642:
So there are now agents trying to fix random "good first issues" to advertise their own product?

sim642 (changes requested):
The "Contributors" list of Goblint is not for advertisement.

PR #2096, opened 2026-08-10
Fix `alloca` returning `NULL`

### LINKED ISSUE #2092: `sem.malloc.fail` affects `alloca`
While reviewing https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1686, I did runs with `sem.malloc.fail` enabled and found differences in tasks using `alloca`.

According to the man page, `alloca` can never return `NULL`:
> #### RETURN VALUE
> The `alloca()` function returns a pointer to the beginning of the allocated space.  If the allocation causes stack overflow, program behavior is undefined.

sim642:
Closes #2092.

### TODO
- [x] sv-benchmarks.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -2511,21 +2511,26 @@ struct
           set ~man st lv_a lv_typ (VD.top_value lv_typ)
     in
     (* Returns a tuple, the first is the address of the blob if one was allocated, the second is the returned address (may contain null pointer or be only null-pointer) *)
-    let alloc loc size =
+    let alloc (loc: Q.AllocationLocation.t) size =
+      let zero_option, fail_option =
+        match loc with
+        | Heap -> "sem.malloc.zero", "sem.malloc.fail"
+        | Stack -> "sem.alloca.zero", "sem.alloca.fail"
+      in
       (* Whether malloc(0) is assumed to return the null pointer, a valid pointer, or both cases need to be considered. *)
-      let malloc_zero_null, malloc_zero_pointer =
-        match get_string "sem.malloc.zero" with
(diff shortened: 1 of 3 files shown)


michael-schwarz (approved):

sim642:
With level01, 60s and 1GB, and `sem.malloc.fail` enabled, this gets us back 107 tasks on sv-benchmarks: https://goblint.cs.ut.ee/results/mr-1686-03-all-level01-main-fail-pr-2096/table-generator-cmp.diff.html.

sim642:
I did some Googling, though, and since `alloca` is so non-standard, there might be implementations that do return `NULL`.
For example, this MacOS man page explicitly mentions the possibility: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/alloca.3.html

> If the allocation failed, a `NULL` pointer is returned.

And GCC apparently has a strange non-stack garbage-collecting implementation: https://stackoverflow.com/a/41954776/854540

> By the way in this implementation `alloca(0)` returns `NULL`.

There was never any SV-COMP discussion about this (because `malloc` assumptions weren't even changed yet), but maybe there will still be.
Perhaps we should have a separate `sem.alloca.fail` option to quickly flip?

michael-schwarz:
> Perhaps we should have a separate `sem.alloca.fail` option to quickly flip?

Makes sense!

sim642:
I also added a `sem.alloca.zero` option now, making it symmetric with `sem.malloc.*`. A separate option for `alloca(0)` allows configuring the GCC special behavior (or the opposite of it).

PR #2097, opened 2026-08-10
Split memLeak paths with `sem.malloc.fail`

### LINKED ISSUE #2093: `sem.malloc.fail` makes memory leak analysis less precise
While reviewing https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1686, I did runs with `sem.malloc.fail` enabled and found differences in memory leak analysis. For example, the result for sv-benchmarks/c/ldv-memsafety/memleaks_test1-3.i changes.

In particular, it seems that calling `free` on a pointer, which can be known or `NULL` is less precise than just with the known pointer in the memLeak analysis. However, that shouldn't be the case: `free(NULL)` is fine and a no-op.

This also seems to affect at least **1175** tasks in sv-benchmarks/Juliet_Test/.

sim642:
Closes #2093.

### TODO
- [x] sv-benchmarks.

### DIFF
--- a/src/analyses/memLeak.ml
+++ b/src/analyses/memLeak.ml
@@ -199,8 +199,14 @@ struct
     | Calloc _
     | Realloc _ ->
       man.sideg () true;
-      begin match man.ask (Queries.AllocVar Heap) with
-        | `Lifted var ->
+      begin match man.ask (Queries.AllocVar Heap), lval with
+        | `Lifted var, Some lval ->
+          man.split (ToppedVarInfoSet.add var state) [SplitBranch (Lval lval, true)];
+          if GobConfig.get_bool "sem.malloc.fail" then
+            man.split state [SplitBranch (Lval lval, false)];
+          raise Analyses.Deadcode
+        | `Lifted var, None ->
+          (* No reason to split without lval: the untracked memory cannot be freed anyway. *)
           ToppedVarInfoSet.add var state
         | _ -> state
       end
--- a/tests/regression/76-memleak/33-malloc-fail-no-mem-leak.c
+++ b/tests/regression/76-memleak/33-malloc-fail-no-mem-leak.c
@@ -1,0 +1,9 @@
+// PARAM: --set ana.malloc.unique_address_count 1 --set ana.activated[+] memLeak --enable sem.malloc.fail
+
+#include <stdlib.h>
+
+int main() {
+  int *p = malloc(sizeof(int));
+  free(p);
+  return 0; // NOWARN
+}


michael-schwarz:
Could you explain why path-sensitivity is the right tool here?

sim642:
Sure, it wasn't my first idea either.

Initially I thought I'd just replace the handling of `free` here: https://github.com/goblint/analyzer/blob/3062271a29bc1f2476d5993af7719d29991cdf16/src/analyses/memLeak.ml#L207-L217
In particular, to do the natural thing of joining states over all points-to targets. But the sound thing to do in the `NullPtr` case is to remove nothing, so that wouldn't help in such a case.

And matching 2-element points-to sets which contain such a removable pointer and `NullPtr` to remove the variable also wouldn't be sound. The `NullPtr` in the points-to set might've come from somewhere else, not the failing allocation. So `NullPtr` might still be in there if the allocation succeeded and there's memory that still needs to be cleaned up.

The memLeak analysis is already path-sensitive w.r.t. its whole local state of allocated variables that still need to be freed, so this additional splitting seems to be the right way. Hopefully, the additional paths with `NULL`s are very short-lived: existing and adapted sv-benchmarks do an `abort()` close to the failed allocation.

sim642:
With level01, 60s and 1GB, and `sem.malloc.fail` enabled, this gets us back 1188 tasks on sv-benchmarks: https://goblint.cs.ut.ee/results/mr-1686-04-all-level01-main-fail-pr-2096-pr-2097/table-generator-cmp.diff.html.

The path-sensitivity seems to have a significant cost in some cases. There are ~170 tasks that go to `TIMEOUT`. None from `true` though, but many from `unknown`, especially in uthash and busybox.

On the `true` tasks, there doesn't seem to be a performance difference though:
image
(https://goblint.cs.ut.ee/results/mr-1686-04-all-level01-main-fail-pr-2096-pr-2097/table-generator-cmp.table.html)

But all these comparisons have still been on the main branch of sv-benchmarks. I'm in the process of repeating them on https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1686.
This could matter for the `TIMEOUT`s because that MR is what would add the quick `abort`s on allocation failure. The main branch benchmarks don't have that and could be pushing the large number of paths through the entire program.

sim642:
> But all these comparisons have still been on the main branch of sv-benchmarks. I'm in the process of repeating them on https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1686.
> This could matter for the `TIMEOUT`s because that MR is what would add the quick `abort`s on allocation failure. The main branch benchmarks don't have that and could be pushing the large number of paths through the entire program.

I've now done that in https://goblint.cs.ut.ee/results/mr-1686-08-all-level01-mr-fail-pr-2096-pr-2097/table-generator-cmp.diff.html.
That leaves ~46 going from `unknown` to `TIMEOUT`. Most of them are from uthash.

---

As the final comparison, I compared Goblint master and sv-benchmarks main with Goblint after #2096+#2097 and sv-benchmarks MR 1686: https://goblint.cs.ut.ee/results/mr-1686-08-all-level01-mr-fail-pr-2096-pr-2097/table-generator-cmp3.diff.html.
Besides the oddities remaining from #2102, there's no other real differences. So #2096 and #2097 are enough to fix our main problems with failing `malloc`, and the sv-benchmarks MR seems good.

PR #2099, opened 2026-08-12
Fix unknown thread invalidation ordering

### LINKED ISSUE #2098: `46/94`: Invalidation on thread creation broken
#1646 did not really fix the issue:

~~~C
#include<pthread.h>
pthread_mutex_t c;
int d, e, f;

void b(void* arg);

void main(int argc, char *argv) {
  pthread_t t;
  e = pthread_create(&t, 0, b, &f);
  pthread_mutex_lock(&c);
  d++;
  pthread_mutex_unlock(&c);
  pthread_mutex_lock(&c);
}
~~~

Here, when disabling widening, the invariant `(long long )f + 2147483648LL >= (long long )e` is still confirmed despite `f` and `e` being invalidated. This was just hidden by widening previosuly.

michael-schwarz:
## Summary

Fix relational invalidation for unknown spawned functions.

The thread-create result and the unknown thread argument were invalidated in separate thread initialization snapshots. Joining those relational snapshots could introduce a spurious relation between the independently invalidated globals, causing an invalid witness invariant to be confirmed when side widening was disabled.

Apply the unknown function invalidation in `threadspawn`, where the caller state already contains the invalidated thread-create result. For the first spawn, publish this combined state as the initialization snapshot; later spawns use ordinary privatized writes. `threadenter` now only initializes the child privatization state.

The regression test disables side widening and independently checks both invalidated globals in addition to the previously spurious relational invariant.

Closes #2098

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -675,10 +675,22 @@ struct
       let new_rel = make_callee_rel ~thread:true man fd args in
       [{st' with rel = new_rel}]
     | exception Not_found ->
-      [special_unknown_invalidate man f args]
+      [Priv.threadenter (Analyses.ask_of_man man) man.global st]
 
   let threadspawn man ~multiple lval f args fman =
-    man.local
+    match Cilfacade.find_varinfo_fundec f with
+    | _ -> man.local
+    | exception Not_found ->
+      (* [man.local] already contains the invalidated thread-create result.
+         Add the unknown thread's invalidations to that same state. When this is
+         the first spawn, publish the combined state as the initialization
+         snapshot; later spawns use ordinary privatized writes. *)
+      let st = special_unknown_invalidate man f args in
+      let ask = Analyses.ask_of_man man in
+      if ThreadFlag.has_ever_been_multi ask then
+        st
+      else
+        Priv.enter_multithreaded ask man.global man.sideg st
 
   let event man e oman =
     let ask = Analyses.ask_of_man man in
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -559,6 +559,8 @@ struct
          let st = invalidate_one ask man st lv in
          assert_fn {man with local = st} (BinOp (Ge, Lval lv, zero, intType)) true
         ) st r
+    | ThreadCreate _, _ ->
+      st (* result is invalidated by the framework after threadspawn *)
     | _, _ ->
       let st' = special_unknown_invalidate man f args in
       (* invalidate lval if present *)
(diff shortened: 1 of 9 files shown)


sim642:
> The thread-create result and the unknown thread argument were invalidated in separate thread initialization snapshots.

What is a "thread-create result"? Is it the resulting abstract state, or is it the lvalue of `pthread_create`?
**EDIT:** Or the thread ID assignment?

And what are "thread initialization snapshots"?

> Joining those relational snapshots could introduce a spurious relation between the independently invalidated globals, causing an invalid witness invariant to be confirmed when side widening was disabled.

Does "joining" mean something other than `join` here?  There's no `join` in this code.

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -675,10 +675,22 @@ struct
...
+      (* [man.local] already contains the invalidated thread-create result.
+         Add the unknown thread's invalidations to that same state. When this is
+         the first spawn, publish the combined state as the initialization
+         snapshot; later spawns use ordinary privatized writes. *)
+      let st = special_unknown_invalidate man f args in
+      let ask = Analyses.ask_of_man man in
+      if ThreadFlag.has_ever_been_multi ask then
+        st
+      else
+        Priv.enter_multithreaded ask man.global man.sideg st
```
sim642: The previous fix in #1646 made both base and relation analyses essentially use the same code. This makes them out of sync again.

Isn't there a similar issue in base then? The invariants added to the test are non-relational.

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -675,10 +675,22 @@ struct
...
   let threadspawn man ~multiple lval f args fman =
-    man.local
+    match Cilfacade.find_varinfo_fundec f with
+    | _ -> man.local
+    | exception Not_found ->
+      (* [man.local] already contains the invalidated thread-create result.
+         Add the unknown thread's invalidations to that same state. When this is
+         the first spawn, publish the combined state as the initialization
+         snapshot; later spawns use ordinary privatized writes. *)
+      let st = special_unknown_invalidate man f args in
```
sim642: I'm not so sure about doing this in `threadspawn`, which is working with the creator's manager.
For thread ID aware analyses, this would make the writes happen in the creator, but they actually are made by the (unknown) created thread, which has a different ID.

sim642: Actually, that's already an issue in `threadenter` as well. Especially because at that point the created thread ID hasn't even been created (it happens in the `threadenter` of threadid analysis).
So maybe it would be correct to do these invalidations here, but on `fman` instead because that should have the new ID.

michael-schwarz:
> What is a "thread-create result"? Is it the resulting abstract state, or is it the lvalue of pthread_create?

It's the `lvalue` of the `pthread create`, so in `g = pthread_create(...)` it is the `g`.

michael-schwarz:
> And what are "thread initialization snapshots"?

It's the values published to `V.mutex_inits` in `Priv.enter_multithreaded`. 

> Does "joining" mean something other than join here? There's no join in this code.

It's about the `LRD.join` that happens for values at `V.mutex_init`.

The problem is that the invalidations of the things reachable via the argument and the invalidation of the `lval` happened separately. So in one, you have

```  
A: e ∈ [INT_MIN, INT_MAX], f = 0
B: e = 0,                 f ∈ [INT_MIN, INT_MAX]
```

The join then gives you `f + 2147483648 >= e` which holds in both.

michael-schwarz:
For

```c
e = pthread_create(&t, NULL, b, &f);
```

the current code works as follows:

1. `threadenter` constructs the complete child state, including its thread ID.
2. For unknown `b`, `S.special` runs on that child state and invalidates memory reachable from `&f`.
3. The child state is synchronized, publishing the invalidation under the child's thread digest.
4. The call-result lvalue `e` is passed through `man.spawn` and stored in the framework's pending-spawn record as `result_lval`. It is kept separate from the thread-handle lvalue `t`.
5. `S.threadspawn` runs for the creator. The thread-flag analysis emits `EnterMultiThreaded`, which invokes each analysis's normal event handler and therefore its `Priv.enter_multithreaded`.
6. After `S.threadspawn` returns, the framework applies an ordinary `Events.Assign` to the retained `result_lval`, invalidating `e` under the creator's digest.

`Priv.enter_multithreaded` is no longer called explicitly by Base or Apron because the existing `EnterMultiThreaded` event now occurs at the correct point and performs that transition uniformly for all analyses.

This ensures that `f` is invalidated using the completed child context, while `e` is invalidated only after the creator has entered multithreaded mode. The invalidations are ordinary writes attributed to the correct threads, avoiding the separate initialization snapshots whose relational join previously retained the spurious invariant.

src/analyses/apron/relationAnalysis.apron.ml
michael-schwarz: I have re-thought this now.

src/analyses/apron/relationAnalysis.apron.ml
michael-schwarz: Back to both being the same again.

PR #2100, opened 2026-08-14
Fix Apron handling of escaped locals

### LINKED ISSUE #2091: Apron unsoundly retains relations for locals modified through an escaped pointer
Apron unsoundly retains relations for locals modified through an escaped pointer. Tested with Goblint nightly Docker image (digest `sha256:74da0d7a666e94b972508c074c97f66d70639d3c4da433b57fffc2e1f71fb208`).

The following test case unsoundly produces `[Success][Assert] Assertion "left == right" will succeed`:
```c
// --set 'ana.activated[+]' apron --enable ana.sv-comp.functions
static int *target;

static void modify(void) {
  *target = __VERIFIER_nondet_int();
}

int main(void) {
  int left = __VERIFIER_nondet_int();
  int right = left;
  target = &right;

  modify();
  __goblint_check(left == right); // incorrectly succeeds
}
```

michael-schwarz:
Fix Apron’s handling of locals whose addresses escape through global pointers.

In single-threaded execution, escaped locals were read using their local relational variable but written using a separate global relational variable. Calls could also omit escaped locals from callee states and restore stale pre-call relations afterward. This allowed Apron to retain relations invalidated by indirect writes.

This change:

- consistently updates escaped locals using their local relational variables before entering multithreaded mode;
- retains escaped locals in callee relational states;
- avoids restoring stale pre-call relations for escaped locals;
- adds regressions for both entirely single-threaded execution and entering multithreaded mode inside the callee.

Tested with:

- `opam exec -- scripts/update_suite.rb group apron3 -q`
- `opam exec -- dune build @runaprontest --display quiet`

Closes #2091.

### DIFF
--- a/src/analyses/apron/relationAnalysis.apron.ml
+++ b/src/analyses/apron/relationAnalysis.apron.ml
@@ -134,7 +134,9 @@ struct
       Priv.write_global ask getg sideg st g x
     else (
       let rel = st.rel in
-      let g_var = RV.global g in
+      (* Escaped locals remain represented by their local variable until the
+         program becomes multithreaded, just like in [read_global]. *)
+      let g_var = if g.vglob then RV.global g else RV.local g in
       let x_var = RV.local x in
       let rel' = RD.add_vars rel [g_var] in
       let rel' = RD.assign_var rel' g_var x_var in
@@ -327,7 +329,7 @@ struct
     let any_local_reachable = any_local_reachable fundec reachable_from_args in
     RD.remove_filter_with new_rel (fun var ->
         match RV.find_metadata var with
-        | Some (Local _) when not (belongs_to_fundec fundec var || any_local_reachable) -> true (* remove caller locals provided they are unreachable *)
+        | Some (Local v) when not (belongs_to_fundec fundec var || any_local_reachable || ThreadEscape.has_escaped (Analyses.ask_of_man man) v) -> true (* remove caller locals provided they are unreachable *)
         | Some (Arg _) when not (List.mem_cmp Apron.Var.compare var arg_vars) -> true (* remove caller args, but keep just added args *)
         | _ -> false (* keep everything else (just added args, globals, global privs) *)
       );
@@ -422,6 +424,7 @@ struct
     let tainted_vars = TaintPartialContexts.conv_varset tainted in
     let new_rel = RD.keep_filter st.rel (fun var ->
         match RV.find_metadata var with
+        | Some (Local v) when ThreadEscape.has_escaped ask v -> false (* escaped locals may be modified through globals *)
         | Some (Local _) when not (belongs_to_fundec fundec var || any_local_reachable) -> true (* keep caller locals, provided they were not passed to the function *)
         | Some (Arg _) -> true (* keep caller args *)
         | Some ((Local _ | Global _)) when not (RD.mem_var new_fun_rel var) -> false (* remove locals and globals, for which no record exists in the new_fun_apr *)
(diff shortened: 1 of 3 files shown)


jprotopopov-ut:
Locally, the change appears fine, although to me it looks like a whack-a-mole fix. For what it's worth, I tried to prompt AI to poke holes in the code for counterexamples, and (before the session got derailed by cybersecurity guardrails) it got me the following adjacent case:
```c
// PARAM: --set ana.activated[+] apron --enable ana.sv-comp.functions
#include <goblint.h>
#include <string.h>

static int *target;

static void modify(void) {
  *target = __VERIFIER_nondet_int();
}

int main(void) {
  int left = __VERIFIER_nondet_int();
  int right = left;
  int *source = &right;
  memcpy(&target, &source, sizeof(target));

  modify();
  __goblint_check(left == right); // UNKNOWN!
}
```
which crashes Goblint with
```
Fatal error: exception Apron.Manager.Error
 {  exn = Exc_invalid_argument; funid = Funid_assign_texpr_array;
  msg = unknown variable right#671 in the environment; }
```

Given that the failure is different from the original one, I am approving this pull request.

jprotopopov-ut (approved):

michael-schwarz:
> Although to me it looks like a whack-a-mole fix.

I took a look at your example, thank you! I think the issue there is fundamentally different. The problem is that the escape analysis does not consider the escape via `memcpy`. This also leads to an unsoundness in the base issue.

~~~C
// PARAM: --enable ana.sv-comp.functions
#include <goblint.h>
#include <pthread.h>
#include <string.h>

static int *target;

static void *modify(void *arg) {
  *target = 7;
  return NULL;
}

int main(void) {
  int right = 42;
  int *source = &right;
  memcpy(&target, &source, sizeof(target));

  pthread_t thread;
  pthread_create(&thread, NULL, modify, NULL);
  pthread_join(thread, NULL);

  __goblint_check(right == 42); // Incorrectly succeeds
}
~~~

Given that there it is an issue with `threadEscape` affecting different analyses, I'll move it to a separate issue.

src/analyses/apron/relationAnalysis.apron.ml
```diff
@@ -422,6 +424,7 @@ struct
     let tainted_vars = TaintPartialContexts.conv_varset tainted in
     let new_rel = RD.keep_filter st.rel (fun var ->
         match RV.find_metadata var with
+        | Some (Local v) when ThreadEscape.has_escaped ask v -> false (* escaped locals may be modified through globals *)
         | Some (Local _) when not (belongs_to_fundec fundec var || any_local_reachable) -> true (* keep caller locals, provided they were not passed to the function *)
```
sim642: How come this is different from the `make_callee_rel` change above?
Previously, both used the exact same condition (negated) to either remove or keep the corresponding locals.
Now, they are no longer symmetric like that.

michael-schwarz: It is not semantically different here, it's just that I broke the symmetry in formatting. I have restored the symmetry now.

sim642: Initially I thought so too that it's just different formatting, but that's not true.
For an escaped local:
* Before the symmetry fix, this immediately goes to `false`.
* After the symmetry fix, all the negated escape condition does is not allow it to go to _that_ `true`. But there are other `Local` cases below with different conditions that may still return `true`.

michael-schwarz: That's true, I hadn't thought about this. Looking at it, the case where this can now actually go to true is

```ocaml
| Some ((Local v | Global v)) when not (TaintPartialContexts.VS.mem v tainted_vars) -> true (* keep locals and globals, which have not been touched by the call *)
```

This is correct: If the escaped variable was definitely not modified in the callee, we can preserve the caller state.

michael-schwarz:
@sim642: Can you have another look so we can merge? Or is this subsumed by one of your larger redsigns, in which cas we shoudl close this?

sim642 (approved):

PR #2105, opened 2026-08-17
fix unwrapping of VarVarinfoMap into normal variable info to avoid mismatches in c2po

### LINKED ISSUE #2104: c2po does not invalidate pointer equality after the callee modifies global pointer
Goblint unsoundly claims that the assertion holds in the program below

```c
// PARAM: --set ana.activated[+] c2po --set ana.activated[+] startState --set ana.activated[+] taintPartialContexts

#include <goblint.h>

int a, b;
int *p, *q;

// Writes q, so q is tainted by this call.
void repoint(int c) {
  if (c)
    q = &a;
  else
    q = &b;
}

int main(int argc, char **argv) {
  q = &a;
  p = q; // establishes p == q

  repoint(argc); // q may now be &b, so the equality no longer holds

  __goblint_check(p == q); // UNKNOWN!
  return 0;
}
```

The issue found by @vesalvojdani coding agent during the discussion of one of effects-related pull requests.

Note: the test case hinges on `argc == 0` possibility (which is possible), but more conventional non-determinism as a condition seems to work too

jprotopopov-ut:
This attempts to fix #2104
The underlying reason seems to be that rich variable info was not correctly unwrapped, thus causing mismatches in c2po which then unsoundly ignored taint

### DIFF
--- a/src/cdomains/duplicateVars.ml
+++ b/src/cdomains/duplicateVars.ml
@@ -117,7 +117,10 @@ struct
     VarVarinfoMap.to_varinfo (ReturnAux typ)
 
   let to_varinfo v =
-    let res = VarVarinfoMap.to_varinfo v in
+    let res = match v with
+      | NormalVar v -> v
+      | v -> VarVarinfoMap.to_varinfo v
+    in
     if M.tracing then M.trace "c2po-varinfo" "to_varinfo: %a -> %a" d_type (get_type v) d_type res.vtype;
     res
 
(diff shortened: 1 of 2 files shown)


michael-schwarz (approved):
Looks good! There should be some comment how for normal variables the original varinfo is re-used, but other than that it's good to merge.

(Btw, it seemed like such a handling as there at some point, but was then removed in a re-factoring https://github.com/goblint/analyzer/commit/cb9dc2feafbf3e600e5bb1787ed97d9e448746da#diff-e69cd4e781fc2670f4640699a0f21eae4ac42764b0a0e6c0baee62a361464011)

PR #2109, opened 2026-08-18
Try to make `graph-easy` deterministic in general

sim642:
In https://github.com/goblint/analyzer/pull/2067#issuecomment-5328148703 I found out that `graph-easy` isn't deterministic.
In that PR the env variables are added to the single test, but if they work then they should be applied to all tests.

I did this PR separately because I was afraid this might cause changes to existing tests with `graph-easy`. Luckily that's not the case, which matches with the fact that this issue never came up before.

### DIFF
--- a/tests/dune
+++ b/tests/dune
@@ -1,0 +1,7 @@
+(env
+ (_
+  (env-vars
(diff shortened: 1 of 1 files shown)


michael-schwarz (approved):
LGTM!

Issue #2110, opened 2026-08-19
Invalid widen exception due to `reflexive_compare` of patricia-tree


jprotopopov-ut:
Goblint (tested with 1daf1de9c) crashes with `Lattice.Invalid_widen` exception upon analyzing the following test case.
```c
// PARAM: --set ana.path_sens[+] base
#include <pthread.h>

int g;

void *t_fun(void *arg) { return NULL; }

int main() {
  pthread_t id;
  int i;
  for (i = 0; i < 3; i++)
    pthread_create(&id, NULL, t_fun, NULL);
  return 0;
}
```

The failure has been discovered during effect-related implementation, triaged and minimized by @vesalvojdani coding agent. The failure only happens on Linux x86_64, not reproducible on arm-based mac's.

Below follow AI triaging conclusions (copy-paste from PR in the effects repo):
> The crash is not in the effect lifter: patricia-tree 0.14.0's reflexive_compare is not antisymmetric, so Map.Make over a base state kept the same key twice and the path-sensitive join dropped a bucket the solver then widened against.

> The exception is Lattice.Invalid_widen from DisjointDomain.ProjectiveSet.widen, whose contract is leq old (join old eqd). Dumping both maps at the failure shows old with 3 buckets and join old eqd with 5, and the bucket old needs is missing from the join. The buckets are a Map.Make keyed by MCP.P, so a missing key after a pointwise join means the map held two entries for one key. Comparing the keys pairwise confirms it: for the two base states involved, compare returns -1 in both directions.
> The order is BaseComponents.compare, first component CPA.compare, which is patricia-tree's reflexive_compare. In the branch where one trie's branch contains the other's, it picks the subtree to descend into with (mb :> int) land (pa :> int) == 0. That is the test for the mirrored case one line above; this case needs ma land pb (functors.ml:656-658). So two maps can each compare smaller than the other, Map.Make keeps both, and any join over that map silently drops one.
> PatriciaMap.reflexive_compare in src/domain/mapDomain.ml now compares the bindings in to_seq order, which is unsigned key order, with the keys compared in that same order — a total order by construction. equal is a different library function (reflexive_same_domain_for_all2) and is unaffected; so is leq.
> Nothing before this branch hit it because packets carried no frame variables. Test 50's two path buckets are two main states that differ by one variable being bound in one and absent from the other, and that variable is worker's parameter, which is only in the packet state now that a suspended frame is retained instead of returned from.
[...]

PR #2111, opened 2026-08-19
domains: compare Patricia maps in an order that is antisymmetric

jprotopopov-ut:
Workaround mitigating #2110 for mapDomain. The code has been cherry-picked "as-is" from the effects repo, and probably does not provide general fix. Therefore, draft PR is created just for reference. Validated to fix the issue with example in #2110

### DIFF
--- a/src/domain/mapDomain.ml
+++ b/src/domain/mapDomain.ml
@@ -252,6 +252,37 @@ struct
   let exists f m = not (for_all (fun k v -> not (f k v)) m)
   let bindings = to_list
   let choose = unsigned_min_binding
+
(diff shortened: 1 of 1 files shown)


sim642:
This essentially replaces the patricia tree `compare` with the inefficient one that's in `Map.Make().compare`, so this likely is a performance-killer. I'm hoping to have the upstream issue (https://github.com/codex-semantics-library/patricia-tree/issues/39) fixed soon enough that this workaround will not be necessary.

sim642:
There's now a working upstream fix in https://github.com/codex-semantics-library/patricia-tree/pull/40, so this will not be necessary.

PR #2112, opened 2026-08-19
Fix `ArrayDomain.Partitioned.make` returning non-normalized values

sim642:
In https://github.com/goblint/analyzer/pull/2081#issuecomment-5341778576, I noticed the following partitioned array abstract value
```
Array (part. by 0): (Unknown -- Unknown -- Unknown)
```
This is strange because it should be normalized away by
https://github.com/goblint/analyzer/blob/1daf1de9c543f34727e84a8584dd7dd42242332f/src/cdomain/value/cdomains/arrayDomain.ml#L331-L336

I looked through all the `Partitioned`-creating places and found that `make` is the only one which doesn't call `normalize`.
So this essentially also applies `normalize` to `make` but simplifies away all the unnecessary back-and-forth.

According to git blame, this strange all-same partitioned value originates from 577ac0ddfbfbe562f6af9526664378b61bb95497. Prior to that the surrounding elements were `Val.bot ()`.
I'm not sure why the special case is there to begin with. Perhaps it should still be there but something from 577ac0ddfbfbe562f6af9526664378b61bb95497 should be reverted instead.
So I don't know if this is necessarily the right solution. But it is one way to ensure that only normalized abstract values are ever used.

### DIFF
--- a/src/cdomain/value/cdomains/arrayDomain.ml
+++ b/src/cdomain/value/cdomains/arrayDomain.ml
@@ -642,12 +642,7 @@ struct
 
 
   let make ?(varAttr=[]) ?(typAttr=[]) i v:t =
-    if Idx.to_int i = Some Z.one  then
-      Partitioned ((Cil.integer 0), (v, v, v))
-    else if Val.is_bot v then
-      Joint (Val.bot ())
-    else
-      Joint v
+    Joint v
 
   let length _ = None
 


sim642:
Although it is perhaps also alarming that the `join`/`widen` doesn't behave the same with `Joint v` and `Partitioned (_, (v, v, v))`.
I think `leq` returns `true` for them both ways, so it seems like they should.

src/cdomain/value/cdomains/arrayDomain.ml
```diff
@@ -642,12 +642,7 @@ struct
 
 
   let make ?(varAttr=[]) ?(typAttr=[]) i v:t =
-    if Idx.to_int i = Some Z.one  then
-      Partitioned ((Cil.integer 0), (v, v, v))
-    else if Val.is_bot v then
-      Joint (Val.bot ())
-    else
-      Joint v
+    Joint v
```
michael-schwarz, line 645: I don't think those are the same, `Val.is_bot()` can be true for things that are not `bot ()`. There presumably was some reason for this.

I think the idea behind the construction with `Partitioned ((Cil.integer 0), (v, v, v))` for arrays of length one was probably to allow destructive updates. But hopefully these will automatically arise again when the array is accessed?

michael-schwarz: Ah, nevermind, I was thinking of `is_bot_value`.

michael-schwarz:
Hopefully the normal join is never actually called, but `smart_join_with_length` is, which behaves in a more meaningful way.

The distinction is that with a partitioned array there is very little that can be done if no function to determine things about the partitioning expression is available.

michael-schwarz:
I think we should do an SV-COMP run for this, the partitioned arrays are famously finicky.

sim642:
> Hopefully the normal join is never actually called, but `smart_join_with_length` is, which behaves in a more meaningful way.

I think in https://github.com/goblint/analyzer/pull/2081#issuecomment-5341778576, it is `smart_join_with_length` that's really being used. The normal `join` looks like it might behave equivalently with an unnormalized element.

src/cdomain/value/cdomains/arrayDomain.ml
sim642: The length 1 special case was added for `calloc` in 5b9b4a331195e07113716a963fad1563aee51683. It matches with this strangeness on the allocation side: #1766.
But I don't know that this length 1 case is exploited anywhere to do strong updates. Given that it was somehow related to just `calloc`, that doesn't seem to have been the motivation.

sim642:
> I think we should do an SV-COMP run for this, the partitioned arrays are famously finicky.

At least on level01 with 60s and 1GB, this PR makes no difference: https://goblint.cs.ut.ee/results/382-all-level01-pr-2112-after/table-generator-cmp.diff.html.

michael-schwarz (approved):
Might want to keep an eye on the weekly this week to see if there's any other impact.

PR #2115, opened 2026-08-25
Use `ValueDomainQueries` for `ArrayDomain` smart ops

sim642:
While working on #2081, especially the `smart_*` ops, I noticed that they could also use `ValueDomainQueries`, exists for other `ArrayDomain` functions.


michael-schwarz (commented):
Looks reasonable. I don't understand how this is an improvement though?

sim642:
AFAIK this doesn't make the current analysis more precise or whatnot. Conceptually, this can be more general because it's not limited to constant results (`Z.t option`). But all of the uses in these functions only do `must_be_equal` which should be equivalent to comparison of constants.

So the improvement is mainly consistency in using the same helpers for everything in partitioned arrays.

michael-schwarz (approved):

PR #2116, opened 2026-08-25
Fix inverted zero-divisor check in base MaySignedOverflow

karoliineh:
The division case in `MaySignedOverflow` computed whether the divisor may contain zero with inverted logic. This caused nonzero divisors to be treated as possibly zero and vice versa.

Found when looking through sv-comp tasks, but I couldn't construct a good regression to expose it.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1396,7 +1396,7 @@ struct
           | `Bot, _ -> false
           | _, `Bot -> false
           | `Lifted i1, `Lifted i2 ->
-            ( let divisor_contains_zero = (ID.is_bot @@ ID.meet i2 (ID.of_int ik Z.zero))  in
+            ( let divisor_contains_zero = not (ID.is_bot @@ ID.meet i2 (ID.of_int ik Z.zero)) in
               if divisor_contains_zero then true else
                 ( let (min_ik, max_ik) = IntDomain.Size.range ik in
                   let (min_i1, max_i1) = (IntDomain.IntDomTuple.minimal i1, IntDomain.IntDomTuple.maximal i1) in


michael-schwarz:
This could potentially have a non-trivial impact on the relational analyses that only keep relationships where things don't overflow.

src/analyses/base.ml
```diff
@@ -1396,7 +1396,7 @@ struct
           | `Bot, _ -> false
           | _, `Bot -> false
           | `Lifted i1, `Lifted i2 ->
-            ( let divisor_contains_zero = (ID.is_bot @@ ID.meet i2 (ID.of_int ik Z.zero))  in
+            ( let divisor_contains_zero = not (ID.is_bot @@ ID.meet i2 (ID.of_int ik Z.zero)) in
```
sim642, line 1399: I think something like this should also work:
```suggestion
            ( let divisor_contains_zero = ID.equal_to Z.zero i2 <> `Neq in
```

sim642:
> Found when looking through sv-comp tasks, but I couldn't construct a good regression to expose it.

> This could potentially have a non-trivial impact on the relational analyses that only keep relationships where things don't overflow.

Indeed, I think the query is just used for helper information for relational analyses, not for the no-overflow verdicts (which are set directly within the int domains).
Our SV-COMP conf uses `assume_none`, so I'm not sure it matters there, but in general it might.

If I were to guess, the reason this hasn't come up before is because the relational analyses probably can't handle division anyway. Or maybe divisions by non-zero constants at best.

sim642 (commented):
I thought I left a comment in code as well but apparently I didn't submit the review and only commented in the PR.

PR #2122, opened 2026-08-27
Upgrade to patricia-tree 0.15.0

### LINKED ISSUE #2110: Invalid widen exception due to `reflexive_compare` of patricia-tree

sim642:
Closes #2110.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -40,7 +40,7 @@ Goblint includes analyses for assertions, overflows, deadlocks, etc and can be e
     (ocaml (>= 4.14))
     (goblint-cil (>= 2.1.0)) ; TODO no way to define as pin-depends? Used goblint.opam.template to add it for now. https://github.com/ocaml/dune/issues/3231. Alternatively, removing this line and adding cil as a git submodule and `(vendored_dirs cil)` as ./dune also works. This way, no more need to reinstall the pinned cil opam package on changes. However, then cil is cleaned and has to be rebuild together with goblint.
     (batteries (>= 3.9.0))
-    (patricia-tree (>= 0.14.0))
+    (patricia-tree (>= 0.15.0))
     (zarith (>= 1.12))
     (yojson (and (>= 2.0.0) (< 3))) ; json-data-encoding has incompatible yojson representation for yojson 3
     (qcheck-core (>= 0.90))
--- a/goblint.opam
+++ b/goblint.opam
@@ -41,7 +41,7 @@ depends: [
   "ocaml" {>= "4.14"}
   "goblint-cil" {>= "2.1.0"}
   "batteries" {>= "3.9.0"}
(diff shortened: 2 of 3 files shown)


michael-schwarz (approved):
Unless you want to do extensive benchmarking, I think we can go ahead and merge (while keeping an eye on nightly and weekly)?

PR #2123, opened 2026-08-27
Simplify descendantLockset access output, fix ambiguous history TID printing

sim642:
This simplifies access output of the descendantLockset analysis from #1913. See 793242c0a43aa985a21726769174784b568fc8fd for the previous unwieldy output.

It adds names to the components (instead of the default `map`) and omits subcomponents which are uninteresting (similarly to `MHP`). See individual commits for changes in output.

It also fixes an old issue about ambiguous history TIDs being printed weirdly, especially in sets:
```diff
- {[main], [main], {t1}}
+ {[main], [main, {t1}]}
```
The set doesn't have two unique `main` threads, although the old output looked like that.

I came across this when looking for examples for #2061 in our existing tests.

### DIFF
--- a/src/analyses/descendantLockset.ml
+++ b/src/analyses/descendantLockset.ml
@@ -100,13 +100,54 @@ module Spec = struct
     struct
       include D
       let name () = "local"
+      let should_print dl =
+        let ls_not_empty _ ls = not @@ Lockset.is_empty ls in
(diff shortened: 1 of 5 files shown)


dabund24 (approved):
LGTM!

Apologies for the slow reply

PR #2124, opened 2026-08-31
Mod bounds unsoundness

vesalvojdani:
This fixes unsound invariant refinement for the mod operation. The issue I ran into was when having intervals and congruences enabled, there was some weird unsoundness that Claude had to debug. I don't even remember what I was doing... But the code it isolated and rewrote is so clearly an improvement that I think it's worth taking. The previous code is very confusing, even to understand what the intention was, and it's also very wrong.

So the main unsoundness was actually that it would just brutally meet with a refined bound, which is very destructive for the congruence domain itself, since it just results in bot. This now relies on meeting with the half-planes ID.starting/ending, so that it is up to the domain to give its best sound representation of that, which is top for congruences. This fixes the unsoundness in 27/27 (mod-bounds-congruence).

The other things were discovered by accident, but is almost more important since it's pure interval unsoundness where some values are refined too aggressively; tests in 27/26 (mod-bounds) also show precision improvements.

### DIFF
--- a/src/analyses/baseInvariant.ml
+++ b/src/analyses/baseInvariant.ml
@@ -342,19 +342,24 @@ struct
         let a' = ID.add (ID.mul (ID.div a b) b) c in
         let b' = ID.div (ID.sub a c) (ID.div a b) in
         (* However, for [2,4]%2 == 1 this only gives [3,4].
-         * If the upper bound of a is divisible by b, we can also meet with the result of a/b*b - c to get the precise [3,3].
-         * If b is negative we have to look at the lower bound. *)
-        let is_divisible bound =
-          (* TODO: could check divisibility directly, instead of via ID? *)
-          GobOption.exists (fun ba -> ID.equal_to Z.zero (ID.rem (ID.of_int ikind ba) b) = `Eq) (bound a)
-        in
-        let max_pos = match ID.maximal b with None -> true | Some x -> Z.compare x Z.zero >= 0 in
-        let min_neg = match ID.minimal b with None -> true | Some x -> Z.compare x Z.zero < 0 in
-        let implies a b = not a || b in
+         * Since [a % b == c] implies that [a] is congruent to [c] modulo [b], every known
+         * bound of [a] can be moved inwards to the closest value congruent to [c] modulo [b].
+         * Here the upper bound 4 moves to 3, which gives the precise [3,3]. *)
         let a'' =
-          if implies max_pos (is_divisible ID.maximal) && implies min_neg (is_divisible ID.minimal) then
-            ID.meet a' (ID.sub (ID.mul (ID.div a b) b) c)
-          else a'
+          match ID.to_int b, ID.to_int c with
+          | Some b, Some c when not (Z.equal b Z.zero) ->
+            (* To handle signs uniformly, we tighten the upper bound m by subtracting the
+             * always positive Euclidean remainder of (m - c) modulo b, which is the precise
+             * amount that m is too high. And symmetrically for the lower bound. *)
+            let a' = match ID.maximal a with
+              | Some m -> ID.meet a' (ID.ending ~suppress_ovwarn:true ikind (Z.sub m (Z.erem (Z.sub m c) b)))
+              | None -> a'
+            in
+            begin match ID.minimal a with
+              | Some m -> ID.meet a' (ID.starting ~suppress_ovwarn:true ikind (Z.add m (Z.erem (Z.sub c m) b)))
+              | None -> a'
+            end
+          | _, _ -> a'
         in
         let a''' =
           (* if both b and c are definite, we can get a precise value in the congruence domain *)
--- a/tests/regression/27-inv_invariants/26-mod-bounds.c
+++ b/tests/regression/27-inv_invariants/26-mod-bounds.c
@@ -1,0 +1,47 @@
+// PARAM: --enable ana.int.interval
+// Refining the interval of a in "a % b == c" must only move the bounds of a to values
+// that are congruent to c modulo b, otherwise satisfying values are cut away.
+#include <goblint.h>
+
+int main() {
+  int x;
+  if (x > 1 && x < 5 && x % 2 == 1) // x = [2,4] && x % 2 == 1 => x = 3
+    __goblint_check(x == 3);
+
+  int y;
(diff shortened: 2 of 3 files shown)


tests/regression/27-inv_invariants/26-mod-bounds.c
```diff
@@ -0,0 +1,47 @@
+// PARAM: --enable ana.int.interval
+// Refining the interval of a in "a % b == c" must only move the bounds of a to values
+// that are congruent to c modulo b, otherwise satisfying values are cut away.
```
sim642, line 3: This comment is somehow specific to a bug that gets fixed here, but is confusing/wrong in general: if `a` is not refined at all, then no satisfying values are cut away.

Maybe something like:
> Refining the interval of a in "a % b == c" should tighten the bounds of a to closest already-contained values that are congruent to c modulo b.

tests/regression/27-inv_invariants/26-mod-bounds.c
```diff
@@ -0,0 +1,47 @@
...
+  int n;
+  if (n >= -28 && n <= -22 && n % -7 == -5)
+    __goblint_check(n == -26); // TODO
+
+  // Test to show both refinements are needed. The congruence shift accepts 1,
+  // but the original mod refinement restricts the sign to negative, so <= -2.
+  int s;
+  if (s >= -12 && s <= 1 && s % 3 == -2) { // s = -11, -8, -5, -2
+    __goblint_check(s >= -11); // TODO
+    __goblint_check(s <= -2);
```
sim642, line 31: The "both" is meaningless just looking at the test. I guess the point is that the interval should be refined on both sides? But the previous checks already had to do it also, so what's different about this?

Also "original mod refinement" is meaningless when we fix this.
And what is the second comment line saying anyway? The new one also restricts to `<= -2` so the check passes, so the old one wasn't wrong about that.

src/analyses/baseInvariant.ml
```diff
@@ -342,19 +342,24 @@ struct
...
-        let implies a b = not a || b in
+         * Since [a % b == c] implies that [a] is congruent to [c] modulo [b], every known
+         * bound of [a] can be moved inwards to the closest value congruent to [c] modulo [b].
+         * Here the upper bound 4 moves to 3, which gives the precise [3,3]. *)
         let a'' =
-          if implies max_pos (is_divisible ID.maximal) && implies min_neg (is_divisible ID.minimal) then
-            ID.meet a' (ID.sub (ID.mul (ID.div a b) b) c)
-          else a'
+          match ID.to_int b, ID.to_int c with
+          | Some b, Some c when not (Z.equal b Z.zero) ->
```
sim642, line 350: This only works for definite values of `b` and `c`, while the old code seemed to be more general, only looking at `minimal` and `maximal` of `b` and nothing about `c`.
I wonder if the old code could soundly do some refinement that the new one doesn't, or not. All of the tests reveal unsoundness already with the definite values.

I guess there are cases where some refinement is still possible with non-definite values, e.g. `[0,8] % 4 = [1,2]` could refine the first argument to `[1,6]`. But I don't know if the old code even did that and if it'd be even worth it.

tests/regression/27-inv_invariants/26-mod-bounds.c
vesalvojdani: This isn't history, but it's very code specific indeed, maybe we don't need a comment at all, why this test was added. But specifically this test kills the "remove the meet on line whatever mutation" that I was wondering about. The original here refers to the previous refinement in the current code that is more precise than the congruence shift.

sim642:
I did the usual sv-benchmarks run with level01, 60s and 1GB. This solves two more tasks: https://goblint.cs.ut.ee/results/385-all-level01-pr-2124-after/table-generator-cmp.diff.html#/table?filter=0(0*status*(status(notIn()),category(in(unknown)))),1(0*status*(status(notIn()),category(in(correct))))
1. loop-industry-pattern/mod3.c.v+lhb-reducer
2. loop-industry-pattern/mod3.c.v+sep-reducer

Somehow we now solve some transformed tasks, but not the loop-industry-pattern/mod3 itself.

src/analyses/baseInvariant.ml
vesalvojdani: I've committed a version now that has a bit more of the generality, but remains very simple. It is often better than what was here before, but it is incomparable. Being general w.r.t. b gets very complicated, but there is a simple generalization of what I've committed for a fixed `b` while keeping c general. It fixes the main unsoundness here that is when the results of the divisor can be both positive and negative, the sound approach is to split this and consider them separately. That's about ten more lines than what is currently committed, and I haven't currently the time to dig into it. Maybe a separate pull request or I dump the code here as a comment if you want to take a look.

vesalvojdani:
Handling the actual benchmark would be nice, but the mod refinements are not the problem: the original program requires a path splitting on the variable involved there. With some brief experiments, I was not able to find a good auto-tuner heuristics to detect these kinds of things. The fully path sensitive configurations time out on that example. Anyway, I think trying to figure out how to efficiently path split for an example like this would be more worthwhile than getting the most precise mod refinement.

src/analyses/baseInvariant.ml
```diff
@@ -337,35 +337,49 @@ struct
...
+            | _, _ ->
+              (* The identity a/b * b + c == a solves to:
+               *      a == (a / b) * b + c
+               *      b == (a - c) / (a / b). *)
+              let a_identity = ID.meet a_sign (ID.add (ID.mul (ID.div a_sign b) b) c) in
+              let b_identity = ID.div (ID.sub a_sign c) (ID.div a_sign b) in
+              meet_bin a_identity b_identity
+          with IntDomain.ArithmeticOnIntegerBot _ -> raise Analyses.Deadcode
         in
-        let a,b = meet_bin a''' b' in
```
sim642, line 368: Such a `meet_bin` should maybe remain: it uses `id_meet_down` which allows all of this to be reused for unassume without actually meeting with the old value (and just using the new range arising from the refinement).
For that to work, I think the new stack of refinements also shouldn't start refining `a`, but should start from `top`.

**EDIT:** I now see that each of the match cases ends with `meet_bin`. Still, the stack of refinements shouldn't start from `a` then.

PR #2127, opened 2026-09-03
Extract weakUpdates analysis from base

sim642:
Following some effects escaping discussions, I tried some things and looked how the `WeakUpdates` handling works.
It's currently part of the base domain but I realized that there's no real reason it needs to be, so this is an attempt to extract it into a separate analysis.

### TODO
- [x] Extract fixes to newly revealed oddities to separate PRs.
   - [x] Merge #2128?
   - [x] Merge #2129?
   - [ ] ~~Fix #2130?~~ No, separate issue.
- [x] Remove `WeakUpdates` from base.

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -25,7 +25,6 @@ module PU = PrecisionUtil
 module VD     = BaseDomain.VD
 module CPA    = BaseDomain.CPA
 module Dep    = BaseDomain.PartDeps
-module WeakUpdates   = BaseDomain.WeakUpdates
 module BaseComponents = BaseDomain.BaseComponents
 
 
@@ -80,8 +79,8 @@ struct
 
   let name () = "base"
 
-  let startstate v: store = { cpa = CPA.bot (); deps = Dep.bot (); weak = WeakUpdates.bot (); priv = Priv.startstate ()}
-  let exitstate  v: store = { cpa = CPA.bot (); deps = Dep.bot (); weak = WeakUpdates.bot (); priv = Priv.startstate ()}
+  let startstate v: store = { cpa = CPA.bot (); deps = Dep.bot (); priv = Priv.startstate ()}
+  let exitstate  v: store = { cpa = CPA.bot (); deps = Dep.bot (); priv = Priv.startstate ()}
 
   (**************************************************************************
    * Helpers
@@ -1643,8 +1642,6 @@ struct
           (* ignore @@ printf "EvalStr Unknown: %a -> %s\n" d_plainexp e (VD.short 80 x); *)
           Queries.Result.top q
       end
-    | Q.IsMultiple v -> WeakUpdates.mem v man.local.weak ||
-                        (hasAttribute "thread" v.vattr && v.vaddrof) (* thread-local variables if they have their address taken, as one could then compare several such variables *)
     | Q.IterSysVars (vq, vf) ->
       let vf' x = vf (Obj.repr (V.priv x)) in
       Priv.iter_sys_vars (priv_getg man.global) vq vf'
@@ -1713,7 +1710,6 @@ struct
       (* Projection globals to highest Precision *)
       let projected_value = project_val (Queries.to_value_domain_ask ask) None None value (is_global ask x) in
       let new_value = VD.update_offset ~blob_destructive (Queries.to_value_domain_ask ask) old_value offs projected_value lval_raw ((Var x), cil_offset) t in
-      (* if WeakUpdates.mem x st.weak then *)
       if man.ask (IsMultiple x) then (
         if invariant then
           old_value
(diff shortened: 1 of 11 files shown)


src/analyses/base.ml
```diff
@@ -1713,8 +1710,12 @@ struct
       (* Projection globals to highest Precision *)
       let projected_value = project_val (Queries.to_value_domain_ask ask) None None value (is_global ask x) in
       let new_value = VD.update_offset ~blob_destructive (Queries.to_value_domain_ask ask) old_value offs projected_value lval_raw ((Var x), cil_offset) t in
-      if WeakUpdates.mem x st.weak then
-        VD.join old_value new_value
+      if man.ask (IsMultiple x) then (
```
sim642, line 1713: This is actually a change: previously we only joined old with new for setting the locals which recursively escaped via arguments. But the `IsMultiple` query also changes it to be done with allocated variables.

Of course for allocated variables we've so far achieved the same through some blob business in `ValueDomain`.
It might be that the latter could be simplified away now even, but I'm not sure.

src/analyses/base.ml
```diff
@@ -2084,7 +2085,7 @@ struct
       (* TODO: move into sync `Init *)
       Priv.enter_multithreaded ask (priv_getg man.global) (priv_sideg man.sideg) st
     | _ ->
-      let locals = List.filter (fun v -> not (WeakUpdates.mem v st.weak)) (fundec.sformals @ fundec.slocals) in
+      let locals = List.filter (fun v -> not (man.ask (IsMultiple v))) (fundec.sformals @ fundec.slocals) in
```
sim642, line 2088: Similarly here, there's a slight apparent generalization. But because `v` is from formals/locals, it cannot be an alloc variable, so this shouldn't actually be a change.

src/analyses/base.ml
sim642: > Of course for allocated variables we've so far achieved the same through some blob business in `ValueDomain`.
> It might be that the latter could be simplified away now even, but I'm not sure.

Not trivially at least: I tried removing the 5 joins on blobs in `update_offset` and a bunch of tests fail.

tests/regression/68-longjmp/47-more-reduced.c
```diff
@@ -24,6 +24,9 @@ main() {
     case 1:
       if (*t->g) { g=1; }
       // This refinement somehow adds bottom for keyword as an explicit binding (???)
+      // weakUpdates finds keyword to be reachable from argument t (via global)
+      // and that leaks to taintPC, modifiedsinceLongJmp, etc, causing a new warning
+      // base's version filters reachable with current CPA on entry and didn't consider it as weakly updatable
       set_key(t);
       longjmp(env_buffer, 2); //NOWARN
```
sim642, line 31: Here `t->g` points to `keyword` of `set_key`, which is also modified there. So this test doesn't now pass due to
```
[Info][Behavior > Undefined > Other] Since setjmp at node 9 "tmp = _setjmp((struct __jmp_buf_tag *)(env_buffer));", locals {keyword} were modified! Reading them will yield Undefined Behavior. (tests/regression/68-longjmp/47-more-reduced.c:31:7-31:29)
```

This is the only check in this test and it seems to have been added to specifically not get this warning here. But I'm not sure what's even the right thing here because `keyword` is out of scope at this point anyway, so it's UB. It was also modified since the `setjmp`, so I don't understand why there shouldn't be this warning.

There's a strange asymmetry in this test, because there's another
```c
set_key(t);
g = 4;
longjmp(env_buffer, 1);
```
(but without `NOWARN`) in this program. The situation there is exactly the same.

sim642: The only difference between the two is that on this first call, `keyword` hasn't even been seen yet, so it doesn't become weakly updatable. But on the second call, `keyword` is pointed to via a global, so it does become weakly updatable.
So it also boils down to: should locals of returned functions even be considered weakly updatable?

it's not so easy to exclude them from weak updates though because there's no way to tell if a pointer to `keyword` in globals points to something up or down the call stack. We don't have any analyses for properly distinguishing them.

The previous fix here had nothing to do with `longjmp` as well: it was just about refining `*t->g` which points to the out-of-scope `keyword`. Since it's missing from base's state, an implicit bottom was refined to an explicit one. The "fix" was just to not do that very specific thing.

sim642: @michael-schwarz You added/fixed this test originally. Do you have any thoughts on this?
I'm not sure what this test was (more) reduced from that it might've been important to not have the warning here.

It's loosely related to a more general issue (#2155) that doesn't involve `longjmp`.

michael-schwarz: I think the point is that there should not be any warning here, as there isn't any local that has been modified since calling setjmp. The thing being modified here is a global and this global can be modified freely.

The problem with `keyword` is not that is a local that is modified between setjmp and longjmp, but that it is local of an entirely different function. As such, I think the warning we produce here is somewhat contradictory (it has nothing to do with the actual problem).

michael-schwarz: The actual warning should not be here, but when `g` is dereferenced.

sim642: But the modifiedSinceSetjmp analysis is explicitly collecting the locals modified also in called functions during `combine`: https://github.com/goblint/analyzer/blob/4dfda097b6fd953e43df197436238ee72d96b773/src/analyses/modifiedSinceSetjmp.ml#L36-L38
where https://github.com/goblint/analyzer/blob/4dfda097b6fd953e43df197436238ee72d96b773/src/analyses/modifiedSinceSetjmp.ml#L17-L19

michael-schwarz: I think that's not really necessary, it would suffice to keep track of the locals of the caller that have been modified.

michael-schwarz: I can never jump back into a frame that has been left. Or am I missing something?

PR #2130, opened 2026-09-04
Add unsound test 01-cpa/63-escaping-recursion-global

sim642:
While debugging the last failing test from #2127, I realized that the current approach in base is wildly unsound.
It only considers locals escaping to callees via argument reachability, but not via globals.

### DIFF
--- a/tests/regression/01-cpa/63-escaping-recursion-global.c
+++ b/tests/regression/01-cpa/63-escaping-recursion-global.c
@@ -1,0 +1,57 @@
+// Copied from 01-cpa/52-escaping-recursion.
+// Modified to escape local via global, not recursive argument.
+#include <goblint.h>
+
+int *ptr;
+
+int rec(int i) {
+    int top;
+    int x = 17;
+
+    if(i == 0) {
+        ptr = &x;
+        rec(5);
+        // Recursive call may have modified x
+        __goblint_check(x == 17); //UNKNOWN!
+
+        // If we analyse this with int contexts, there is no other x that is reachable, so this
+        // update is strong
+        x = 17;
+        __goblint_check(x == 17);
+    } else {
+        x = 31;
+
+        // ptr points to the outer x, it is unaffected by this assignment
+        // and should be 17
+        __goblint_check(*ptr == 31); //UNKNOWN!
+
+        if(top) {
+            ptr = &x;
+        }
+
+        // ptr may now point to both the inner and the outer x
+        *ptr = 12;
+        __goblint_check(*ptr == 12); //UNKNOWN!
+        __goblint_check(x == 12); //UNKNOWN!
+
+        if(*ptr == 12) {
+            __goblint_check(x == 12); //UNKNOWN!
+        }
+
+        // ptr may still point to the outer instance
+        __goblint_check(ptr == &x); //UNKNOWN!
+
+        // Another copy of x is reachable, so we are conservative and do a weak update
+        x = 31;
+        __goblint_check(x == 31); // UNKNOWN
+    }
+    return 0;
+}
+
+
+int main() {
+    int t;
+    ptr = &t;
+    rec(0);
+    return 0;
+}


vesalvojdani:
This is also something the AI agents ran into independently, while mutating some of our benchmarks. The real benchmarks only do escapes through locals, but another variation Opus probed was when addresses of formals escape. The following two tests expose current unsoundness in Goblint. The vibe-fixes touch the same reachable list, and the second example below is meant to show that if one naively make more things reachable, you can get more formals overwritten.

[This is Claude Opus 5.5.]

### A formal reached through an argument

```c
#include <goblint.h>

static void rec(int x, int *p) {
  if (!p) {
    rec(5, &x);
    __goblint_check(x == 2); //UNKNOWN (the inner activation writes only its own x, but one cell carries both)
  } else {
    // x is the argument 5; p points to the outer x, which is 2.
    __goblint_check(x == 5); //UNKNOWN (one cell carries both copies of x)
    __goblint_check(*p == 2); //UNKNOWN (the same cell)
  }
}

int main(void) {
  rec(2, 0);
  return 0;
}
```

```console
$ ./goblint formal.c
[Error][Assert] Assertion "x == 5" will fail. (formal.c:9:5-9:28)
[Success][Assert] Assertion "*p == 2" will succeed (formal.c:10:5-10:29)
[Success][Assert] Assertion "x == 2" will succeed (formal.c:6:5-6:28)
```

`make_entry` binds the formal `x` to 5 at [line 2201](https://github.com/goblint/analyzer/blob/4dfda097b6fd953e43df197436238ee72d96b773/src/analyses/base.ml#L2201), `CPA.add_list pa st'.cpa`, and then copies the caller's binding of every variable the arguments reach at [line 2205](https://github.com/goblint/analyzer/blob/4dfda097b6fd953e43df197436238ee72d96b773/src/analyses/base.ml#L2205), `CPA.add_list_fun reachable (fun v -> CPA.find v st.cpa)`. `&x` reaches the outer `x`, which is the same `varinfo`, so the outer 2 replaces the 5. `x` then goes into the weak set, but its one cell holds 2 alone, where it has to hold both 2 and 5.

### A formal reached through a global

The formal version of `63`:

```c
#include <goblint.h>

static int *outer;

static void f(int n) {
  if (!outer) {
    outer = &n;
    f(1);
    __goblint_check(n == 2); //UNKNOWN (the inner activation writes only its own n, but one cell carries both)
  } else {
    // n is the argument 1; outer points to the outer n, which is 2.
    __goblint_check(n == 1); //UNKNOWN (one cell carries both copies of n)
    __goblint_check(*outer == 2); //UNKNOWN (the same cell)
  }
}

int main(void) {
  f(2);
  return 0;
}
```

On master this is the defect of this PR, on a formal:

```console
$ ./goblint global-formal.c
[Success][Assert] Assertion "n == 1" will succeed (global-formal.c:12:5-12:28)
[Error][Assert] Assertion "*outer == 2" will fail. (global-formal.c:13:5-13:33)
[Warning][Assert] Assertion "n == 2" is unknown. (global-formal.c:9:5-9:28)
```

With the globals added to the roots `make_entry` passes to `reachable_vars`, and nothing else changed, `n` becomes reachable through `outer`. Its outer binding 2 then replaces the argument 1 by the same mechanism as in the first program, and the wrong verdict moves to the other check:

```console
[Error][Assert] Assertion "n == 1" will fail. (global-formal.c:12:5-12:28)
[Success][Assert] Assertion "*outer == 2" will succeed (global-formal.c:13:5-13:33)
[Success][Assert] Assertion "n == 2" will succeed (global-formal.c:9:5-9:28)
```

With that change `63` passes as annotated, so a fix that makes `63` pass can still leave this program failing.

### What was run

Goblint master at 4dfda097b, default options. The second output of the global program, and `63` passing, are from master with only the roots in `make_entry` changed, built from the working tree rather than a commit. Compiled with gcc and `__goblint_check` defined as `assert`, every check in both programs holds at `-O0` and at `-O2`, and under ASan and UBSan.

#2127 moves the weak set into `weakUpdates.make_entry`, but by its diff the binding of formals and the copying of reachable bindings stay in `Base.make_entry` unchanged, so both programs should behave the same on that branch; that was not run.

PR #2133, opened 2026-09-04
Add used assumptions declarations

sim642:
This is a very simple implementation of `Assumptions.add` for declaring which assumptions have been used by the analysis.

Unlike the vibed version for effects, this doesn't implement a system for dynamically registering assumption types (separately from their message). That seems overengineering given that we don't have a standard numeration of all warning types even.

### TODO
- [x] Output assumptions to checks file like Mopsa.
- [x] Add possibility of specifying separate final assumption message (for cases where location-message has analysis-specific details).

### DIFF
--- a/src/analyses/base.ml
+++ b/src/analyses/base.ml
@@ -1033,7 +1033,7 @@ struct
                 if contains_vla t || contains_vla (Addr.Mval.type_of (x, o)) then
                   begin
                     (* TODO: Is this ok? *)
-                    M.info ~category:Unsound "Casting involving a VLA is assumed to work";
+                    M.info ~category:Unsound "Casting involving a VLA is assumed to work"; (* TODO: assume *)
                     true
                   end
                 else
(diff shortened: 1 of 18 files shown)


src/analyses/tutorials/taint.ml
```diff
@@ -59,6 +59,7 @@ struct
       false
     | _ ->
       (* We assume using a tainted offset does not taint the expression, and that our language has no pointers *)
+      (* TODO: assume? *)
```
sim642, line 62: I wonder if we should also already show this interface in this tutorial.

src/analyses/base.ml
```diff
@@ -1521,7 +1520,8 @@ struct
     | Q.EvalMutexAttr e -> begin
         match eval_rv_address ~man man.local e with
         | Address a ->
-          let default = `Lifted MutexAttrDomain.MutexKind.NonRec in (* Goblint assumption *)
+          let default = `Lifted MutexAttrDomain.MutexKind.NonRec in
+          Assumptions.add "Mutexes are non-recursive by default";
```
sim642, line 1524: This gets triggered quite a lot even in our existing tests. Basically, it happens on every `pthread_mutex_init(&m, NULL)`, not when something is done with the mutex where the (non-)recursiveness actually matters.

This causes test 71/17 default-null-dyn to fail.

vesalvojdani:
So I think there is some benefit in the assumptions being predefined, but spreading them out to allow that dynamic registration was indeed a strange counter-productive choice. I think naming the assumptions, and having e.g. global static constants in `Assumptions.ml` would be useful. We haven't standardized warnings, but these assumptions are things that I don't think we should allow people to introduce without having a very good reason, so forcing them to be added to a central file would be a good discipline.

PR #2138, opened 2026-09-08
Remove ikind argument from `IndexDomain` functions

sim642:
`IndexDomain` defaults the ikind to `ptrdiff_t`: https://github.com/goblint/analyzer/blob/1ed379e2db8f8df4df204f46a989a761be644110/src/cdomain/value/cdomains/preValueDomain.ml#L3

However, currently `IntDomWithDefaultIkind` doesn't pass the default ikind to all functions, but only some.
This means that it's possible to misuse `IndexDomain`, e.g. `IndexDomain.of_int IInt Z.zero`, which is fine on its own but can lead to `IncompatibleIKinds` down the line, when operations on `IndexDomain` try to use `ptrdiff_t` on it instead.
So the default ikind should really be used across the entire interface, preventing such misuses altogether.

In most cases `(Cilfacade.ptrdiff_ikind ())` was passed anyway, so this just simplifies code.
There are two major misuses though:
- [x] For some reason the gStoreWidening tutorial defaults intervals to `ptrdiff_t` as well: https://github.com/goblint/analyzer/blob/1ed379e2db8f8df4df204f46a989a761be644110/src/analyses/tutorials/gStoreWideningHelper.ml#L4-L5
    However, later it still creates values with other ikinds. Fixed by #2139.
- [ ] `ArrayDomain.NullByte` uses a wild mix of `IInt`, `ILong` and `!Cil.kindOfSizeOf`.
- [x] ~~The latter probably causes~~ 73/07 larger_example and 73/10 char_arrays to fail now. Fixed by #2140.
- [ ] Do casts in `IntDomWithDefaultIkind`? Shouldn't be needed for `of_*` operations. But `IndexDomain` should have abstract type and conversion to-from `ID` cast.

This came up in https://github.com/goblint/analyzer/pull/1987#discussion_r3915030799.

### DIFF
--- a/src/cdomain/value/cdomains/arrayDomain.ml
+++ b/src/cdomain/value/cdomains/arrayDomain.ml
@@ -1182,7 +1182,7 @@ struct
         Checks.safe Checks.Category.NegativeArraySize;
         Z.zero, None
     in
-    let size = BatOption.map_default (fun max -> Idx.of_interval ILong (min_i, max)) (Idx.starting ILong min_i) max_i in
+    let size = BatOption.map_default (fun max -> Idx.of_interval (min_i, max)) (Idx.starting ILong min_i) max_i in (* TODO: used ILong *)
     match Val.is_null v with
     | Null -> (Nulls.make_all_must (), size)
     | NotNull -> (Nulls.empty (), size)
@@ -1338,7 +1338,7 @@ struct
       (* else return interval [minimal may null, minimal must null] *)
     else (
       Checks.safe Checks.Category.InvalidMemoryAccess;
-      Idx.of_interval !Cil.kindOfSizeOf (Nulls.min_elem Possibly nulls, Nulls.min_elem Definitely nulls))
+      Idx.of_interval (Nulls.min_elem Possibly nulls, Nulls.min_elem Definitely nulls)) (* TODO: used !Cil.kindOfSizeOf *)
 
   let string_copy (dstnulls, dstsize) ((srcnulls, srcsize) as src) n =
     let must_nulls_set1, may_nulls_set1 = dstnulls in
--- a/src/cdomain/value/cdomains/arrayDomain.ml
+++ b/src/cdomain/value/cdomains/arrayDomain.ml
@@ -1182,7 +1182,7 @@ struct
         Checks.safe Checks.Category.NegativeArraySize;
         Z.zero, None
     in
-    let size = BatOption.map_default (fun max -> Idx.of_interval (min_i, max)) (Idx.starting ILong min_i) max_i in (* TODO: used ILong *)
+    let size = BatOption.map_default (fun max -> Idx.of_interval (min_i, max)) (Idx.starting min_i) max_i in (* TODO: used ILong *)
     match Val.is_null v with
     | Null -> (Nulls.make_all_must (), size)
     | NotNull -> (Nulls.empty (), size)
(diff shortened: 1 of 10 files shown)


src/cdomain/value/cdomains/valueDomain.ml
```diff
@@ -1134,10 +1134,10 @@ struct
                           match Cil.unrollType fld.ftype with
                           | TArray(_, l, _) ->
                             let len = Cil.lenOfArray l in (* LenOfArray exception will not happen, VLA not allowed in union and struct *)
-                            Array(CArrays.make (IndexDomain.of_int (Cilfacade.ptrdiff_ikind ()) (Z.of_int len)) Top), offs
+                            Array(CArrays.make (IndexDomain.of_int (Z.of_int len)) Top), offs
                           | _ -> top (), offs (* will not happen*)
                         end
-                      | `Index (idx, _) when IndexDomain.equal idx (IndexDomain.of_int (Cilfacade.ptrdiff_ikind ()) Z.zero) ->
+                      | `Index (idx, _) when IndexDomain.equal idx (IndexDomain.of_int Z.zero) ->
```
sim642, line 1140: On an unrelated note: this is slightly strange, using `equal_to` would be the more correct thing to do. I guess this usually works the same way, although it's possible that one int domain can prove the equality while another cannot, then this wouldn't work.

src/cdomain/value/cdomains/offset.ml
```diff
@@ -205,23 +205,23 @@ struct
      element) answers with any_index_exp on all non-concrete cases. *)
   let rec of_exp: exp offs -> t = function (* TODO: Poly.map_indices *)
     | `NoOffset    -> `NoOffset
-    | `Index (Const (CInt (i,ik,s)),o) -> `Index (Idx.of_int ik i, of_exp o)
+    | `Index (Const (CInt (i,ik,s)),o) -> `Index (Idx.of_int i, of_exp o)
```
sim642, line 208: I think this is the root of all the confusion I had in https://github.com/goblint/analyzer/pull/1987#discussion_r3915030799.
When a constant index offset expression is converted to the integer abstraction, it should _not_ use the original ikind if we want the index abstractions to be compatible for computations.
Sure, it could be cast to `ptrdiff_t` like in #1987 but just using `of_int` with `ptrdiff_t` (defaulted) should do the same thing anyway. The cast achieves nothing: if `i` doesn't fit `ptrdiff_t`, then `of_int` produces the overflow warning, just like `cast_to` would.

src/cdomain/value/cdomains/arrayDomain.ml
```diff
@@ -888,7 +888,7 @@ struct
 end
 
 
-module PartitionedWithLength (Val: LatticeWithSmartOps) (Idx: IntDomain.Z): S with type value = Val.t and type idx = Idx.t =
+module PartitionedWithLength (Val: LatticeWithSmartOps) (Idx: IntDomain.ZDefault): S with type value = Val.t and type idx = Idx.t =
 struct
   module Base = Partitioned (Val) (Idx)
   include Lattice.Prod (Base) (Idx)
```
sim642, line 894: I think the mixture of `IInt`, `ILong` and `!Cil.kindOfSizeOf` is due to also using `Idx` for the array/string length. This is more sane than #2140 at least.

Ideally, the length abstraction would default the ikind to `!Cil.kindOfSizeOf` (`size_t`?) or something (unsigned!) instead.
That would require some casting for comparing indexes with lengths, which might be more annoying.

**EDIT:** I tried this in #2142.

PR #2141, opened 2026-09-08
Remove polymorphic `BatSet` usage in old solvers and `MCP`

sim642:
@kalmera used the `slr3` solver for some evaluations and it sometimes crashed with
```
Fatal error: exception Invalid_argument("compare: abstract value")
```
For example:
```
./goblint --conf conf/svcomp26/common.json --conf conf/svcomp26/verify.json --conf conf/svcomp26/level01.json --set solver slr3 -v --sets ana.specification ../sv-benchmarks/c/properties/termination.prp --sets exp.architecture 64bit ../sv-benchmarks/c/termination-memory-alloca/Urban-alloca-2.i
```

This exception is caused by using polymorphic comparison on some abstract values (in the OCaml sense), possibly Apron elements.
This PR removes usages of the polymorphic `BatSet` in the old solvers and `MCP` (where it was actually fine due to `int` elements).

### DIFF
--- a/src/analyses/mCP.ml
+++ b/src/analyses/mCP.ml
@@ -36,7 +36,7 @@ struct
   let name () = "MCP2"
 
   let path_sens = ref []
-  let act_cont_sens = ref Set.empty
+  let act_cont_sens = ref Set.Int.empty
   let base_id   = ref (-1)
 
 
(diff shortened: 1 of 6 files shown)


kalmera (approved):
I confirm that it solves the issues in C.valid-memsafety.Other and in no-data-race that I tested.

PR #2142, opened 2026-09-09
Add `SizeDomain` for length in `ArrayDomain`

sim642:
This is a quick attempt at trying to implement my idea from https://github.com/goblint/analyzer/pull/2138#discussion_r3966176111.
Basically, to use a new `SizeDomain`, which is `IntDomain` with default `size_t` ikind, for the array length abstraction.

This `SizeDomain` is global (like `ValueDomain.IndexDomain`), but the array domains are functorized over `Idx`.
We could add another functor argument for `Size` but that just makes things more complicated. Actually I might even want to get rid of the `Idx` argument if possible (I think we don't need to instantiate arrays with different index abstractions), but not here.

### DIFF
--- a/src/cdomain/value/cdomains/arrayDomain.ml
+++ b/src/cdomain/value/cdomains/arrayDomain.ml
@@ -7,6 +7,7 @@ module M = Messages
 module A = Array
 module VDQ = ValueDomainQueries
 module ID = PreValueDomain.ID
+module SizeDomain = PreValueDomain.SizeDomain
 
 type domain = TrivialDomain | PartitionedDomain | UnrolledDomain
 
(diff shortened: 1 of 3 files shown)


src/cdomain/value/cdomains/arrayDomain.ml
```diff
@@ -816,7 +817,7 @@ let array_oob_check ( type a ) (module Idx: IntDomain.ZDefault with type t = a)
   if !AnalysisState.executing_speculative_computations then
     ()
   else if GobConfig.get_bool "ana.arrayoob" then (* The purpose of the following 2 lines is to give the user extra info about the array oob *)
-    let idx_before_end = Idx.lt v l in (* check whether index is before the end of the array *)
+    let idx_before_end = None in (* check whether index is before the end of the array *)
```
sim642, line 820: This bounds check needs to be fixed. This requires conversions between `Idx` and `SizeDomain`, which are integer domains with different default ikinds.

PR #2144, opened 2026-09-09
Uses modulo constants as extra widening thresholds

### LINKED ISSUE #2135: Use modulo constants as widening thresholds
Astrée apparently found it to be useful to use the constants (minus 1) in assignments such as `x = x % 10` (here: `10`) as additional widening thresholds. We should consider this also in our autotuner. 

They did not mention a specific example, but here is a simple example:

~~~C
int x = 0; 

while (nondet()) {
    x = (x + 1) % 10;
}

assert(x <= 10);
~~~ 

In the example above, narrowing recovers precision, but here it wouldn't:

~~~C
int x = 1;

while (nondet()) {
    if (x % 10 == 0)
        x = 1;
    else
        x++;
}

assert(x <= 10);
~~~

In more realistic examples, this would probably be related to something such as a ring buffer maintaining the last k entries. Given this is only one additional line of code, I think it's worth trying.

bernalalexis-try:
In wideningThresholds.ml, extractThresholdsFromConditionsVisitor now matches BinOp (Mod, _, Const (CInt c), _) and contributes |c|-1 as an upper threshold and -(|c|-1) as a lower one, matching C semantics where the result takes the dividend's sign, plus the four octagon thresholds following the existing cases. 

Zero divisors are guarded, negative ones normalised with Z.abs.

autoTune.ml needs no change: wideningOption derives value and cost from upper_thresholds' cardinality, so modulo constants feed the heuristic already.

Closes #2135.

### DIFF
--- a/tests/regression/03-practical/37-threshold-modulo.c
+++ b/tests/regression/03-practical/37-threshold-modulo.c
@@ -1,0 +1,25 @@
+// PARAM: --enable ana.int.interval --enable ana.int.interval_threshold_widening --set ana.int.interval_threshold_widening_constants comparisons --enable exp.no-narrow
+#include <goblint.h>
+
+int nondet();
+
+int main() {
+  int x = 0;
+
+  while (nondet()) {
+    x = (x + 1) % 10;
+  }
+
+  __goblint_check(x >= 0);
+  __goblint_check(x <= 9);
+
+  int y = 0;
+
+  while (nondet()) {
+    y = (y - 1) % 10;
+  }
+
+  __goblint_check(y >= -9);
+
+  return 0;
+}
--- a/src/cdomain/value/util/wideningThresholds.ml
+++ b/src/cdomain/value/util/wideningThresholds.ml
@@ -56,6 +56,20 @@ class extractThresholdsFromConditionsVisitor(upper_thresholds,lower_thresholds, 
       addThreshold octagon_thresholds @@ doubleI;
       addThreshold octagon_thresholds @@ Z.neg doubleI;
       DoChildren
+
+    (* Modulo by a constant: expr % 10 *)
+    (* The result of expr % c is contained in [-(|c|-1), |c|-1], so |c|-1 is the tightest bound *)
+    | BinOp (Mod, _, (Const (CInt(i,_,_))), _) when not (Z.equal i Z.zero) ->
+      let i = Z.pred @@ Z.abs i in
+      addThreshold upper_thresholds @@ i;
+      addThreshold lower_thresholds @@ Z.neg i;
+
+      addThreshold octagon_thresholds @@ i;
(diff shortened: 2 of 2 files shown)


tests/regression/03-practical/37-threshold-modulo.c
```diff
@@ -0,0 +1,25 @@
+// PARAM: --enable ana.int.interval --enable ana.int.interval_threshold_widening --set ana.int.interval_threshold_widening_constants comparisons --enable exp.no-narrow
+#include <goblint.h>
+
+int nondet();
```
sim642, line 4: It would be better for consistency with SV-COMP to use the following instead:
```suggestion
extern _Bool __VERIFIER_nondet_bool();
```
Then `--enable ana.sv-comp.functions` should also be added above.

tests/regression/03-practical/37-threshold-modulo.c
```diff
@@ -0,0 +1,25 @@
...
+
+  int y = 0;
+
+  while (nondet()) {
+    y = (y - 1) % 10;
+  }
+
+  __goblint_check(y >= -9);
+
+  return 0;
```
sim642, line 24: It would be good to also have the second example from #2135 here.

sim642 (changes requested):

bernalalexis-try:
Thanks, both addressed in 186acb0:

- The loops now use `extern _Bool __VERIFIER_nondet_bool();`, with `--enable ana.sv-comp.functions` added to the `PARAM` line.
- The second example from #2135 is in the test as the `z` loop, checking `z <= 10`.

I ran the test with and without the `wideningThresholds.ml` change, same options:

| check | without the change | with it |
| --- | --- | --- |
| `x <= 9` | unknown | success |
| `y >= -9` | unknown | success |
| `z <= 10` | success | success |

So the first two cover the modulo thresholds, but the second example already passes without them under these options, so it doesn't demonstrate the change here. If you had a different configuration in mind for it, one where it only passes with the modulo thresholds, I'm happy to adjust or move it to its own test.

tests/regression/03-practical/37-threshold-modulo.c
```diff
@@ -0,0 +1,38 @@
...
+
+int main() {
+  int x = 0;
+
+  while (__VERIFIER_nondet_bool()) {
+    x = (x + 1) % 10;
+  }
+
+  __goblint_check(x >= 0);
+  __goblint_check(x <= 9);
```
sim642, line 14: Actually the widening threshold finder is supposed to find the threshold 9 from this check, without needing this PR at all. The only reason it currently doesn't is silliness: #2147.

The act of checking is influencing the test itself ;)
Perhaps I should just deal with #2147 first...

PR #2145, opened 2026-09-10
Remove unnecessary code from region analysis

sim642:
Due to https://github.com/goblint/analyzer/issues/118#issuecomment-5614353021 I had a quick look at the region analysis.
I don't have a solution for that issue but I did some very obvious cleanup while at it:

1. Remove unnecessary `Lift` wrapper on the domain. The `` `Bot `` and `` `Top`` values weren't constructed anywhere.
2. Remove unnecessary `leq` checks before `sideg`. No other analysis does this and the solvers should optimally handle the subsumed case anyway. I think this is a leftover from #124.

This PR is probably best viewed with whitespace ignored.

### DIFF
--- a/src/analyses/region.ml
+++ b/src/analyses/region.ml
@@ -14,7 +14,7 @@ module Spec =
 struct
   include Analyses.DefaultSpec
 
-  module D = RegionDomain.RegionDom
+  module D = RegMap
   module G = RegPart
   include Analyses.ValueContexts(D)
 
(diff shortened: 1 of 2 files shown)


michael-schwarz (approved):
Code-wise this makes sense, the behavior is unaffected. I have however not taken a look into what the region analysis does. 

So my approval basically is to be read as: The code still does the same thing it did before, but looks nicer. I don't claim that it does the right thing.

PR #2147, opened 2026-09-11
Find widening thresholds from reset conditions

sim642:
The existing logic for widening thresholds considers the case where the expression is used as the loop condition.
However, there are other examples like the second one in #2135 or an existing Miné one copies here as a test, where the condition is inside the loop and used to reset a counter to be back in bounds.
This essentially means that the same expression is used in the opposite role w.r.t. widening thresholds, so this adds both upper and lower thresholds in all cases.

I've intentionally not added a helper to do both (even though all cases now do) because the reasons in each case are different and worth documenting.
Moreover, this doesn't undo the splitting/refactoring of #2146 (the test didn't work before it either). Prior to #2146 the opposite bound added from the reset condition was off by one and, thus, didn't help.

### DIFF
--- a/tests/regression/03-practical/38-mine-tutorial-ex4.7-threshold.c
+++ b/tests/regression/03-practical/38-mine-tutorial-ex4.7-threshold.c
@@ -1,0 +1,17 @@
+// PARAM: --enable ana.int.interval --enable ana.sv-comp.functions --enable ana.int.interval_threshold_widening --set ana.int.interval_threshold_widening_constants comparisons
+// Copy of 56-witness/27-mine-tutorial-ex4.7 without witness.
+#include <goblint.h>
+extern _Bool __VERIFIER_nondet_bool();
+int main() {
+  int x = 0;
+  while (__VERIFIER_nondet_bool() == 0) {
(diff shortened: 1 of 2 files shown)


tests/regression/03-practical/38-mine-tutorial-ex4.7-threshold.c
```diff
@@ -0,0 +1,17 @@
+// PARAM: --enable ana.int.interval --enable ana.sv-comp.functions --enable ana.int.interval_threshold_widening --set ana.int.interval_threshold_widening_constants comparisons
+// Copy of 56-witness/27-mine-tutorial-ex4.7 without witness.
+#include <goblint.h>
+extern _Bool __VERIFIER_nondet_bool();
+int main() {
+  int x = 0;
+  while (__VERIFIER_nondet_bool() == 0) {
+    __goblint_check(0 <= x);
+    __goblint_check(x <= 40);
```
sim642, line 9: The presence of this check should already be making the addition unnecessary: https://github.com/goblint/analyzer/pull/2144#discussion_r4081991073.
It's just another case that's missing here.

PR #2152, opened 2026-09-15
Autotuner: barriers, `pthread_once`, complicated thread creation and lockset interactions

dabund24:
Autotuning for analyses published in Aplas paper:

# `pthread_barrier` analysis

Disabled if no barrier is initialized

# `pthread_once` analysis

Disabled if `pthread_once` is used nowhere

# locked creation analyses

enabled/disabled alongside other `notNeccessaryThreadAnalyses`. More 'precise' autotuning (i.e. something like don't disable only if thread is created and mutex is used and either `data-race` svcomp property or no svcomp property is used) could be done in a later PR. Did not do this here as similar decisions would need to be done for other `notNeccessaryThreadAnalyses`(?, see comment [`TODO: what's the exact relation between thread analyses enabled in conf, the ones we disable in reduceAnalyses and the ones we enable here?`](https://github.com/goblint/analyzer/pull/2152/changes#diff-ef9dbea6cca7d31e1e1853acf1be347f9db38947c32d709f8e8c8b4771951d01L262))


sim642 (approved):

sim642:
I've added a TODO #2131 to also add the new autotuners to the upcoming svcomp27 confs, if there are new tasks needing them.

sim642 (changes requested):
I just noticed that the CI fails on MacOS. Some of the `grep` business might not be entirely standard.

michael-schwarz:
MacOS does not have barriers, I'd guess that is the problem. All barrier tests are annotated with `//NOMAC`.

src/autoTune.ml
```diff
@@ -216,7 +216,7 @@ let enableAnalyses reason description analyses =
 (*escape is also still enabled, because otherwise we get a warning*)
 (*does not consider dynamic calls!*)
 let notNeccessaryRaceAnalyses = ["race"; "symb_locks"; "region"]
```
michael-schwarz: Should they not be above in `notNeccessaryRaceAnalyses`.

dabund24:
https://github.com/goblint/analyzer/pull/2152/commits/194a69296a064a767c05d3435fd427d6c8cae0ae should fix the MACOS issue

src/autoTune.ml
dabund24: Right, that seems reasonable. (The only difference would be that the analyses are additionally disabled if the `data-race` prop is not used, which makes sense)

Changed this in https://github.com/goblint/analyzer/pull/2152/commits/184de985242ef6ed876f1e5a140784d2c75c4686

michael-schwarz (approved):

sim642 (approved):

PR #2153, opened 2026-09-16
Interleave contexts with paths in XSLT output

sim642:
While debugging some branchSet stuff (#2150), I got really confused for a while by seeing multiple paths with the same representative, which makes no sense.
Turns out those paths were in different contexts but the current XSLT output makes it very difficult to tell:
* All contexts are pulled together under a collapsed "context" section.
* In that section there are no delimiters between the individual contexts.
* All paths from all contexts are below the "context" section, making it impossible to tell which path corresponds to which context.

This PR changes it to the following:
* Each context is a separate collapsed "context" section.
* The context sections are interleaved with the path sections.
* Each context section is followed by the paths corresponding to that context. This may be one or multiple.


michael-schwarz:
This is something that has been annoying me basically forever, nice that there is now a fix!

PR #2157, opened 2026-09-22
TD Parallel

arkocal:
Details to follow


src/common/framework/analysisState.ml
```diff
@@ -2,13 +2,13 @@
 
 (** If this is true we output messages and collect accesses.
     This is set to true in control.ml before we verify the result (or already before solving if warn = 'early') *)
-let should_warn = ref false
+let should_warn = Domain.DLS.new_key (fun () -> false)
```
sim642, line 5: I don't think this should be domain-local. It's a very global centrally-set thing and all solver threads should follow that.
I think this shouldn't even change while the solver is running.

PR #2158, opened 2026-09-22
Perform `ana.unassume.precheck` by disjunct

sim642:
This should make unassuming of disjunctive invariants (from loop unrolling) more precise,
avoiding unassuming the invariant for the entire loop already for the first iteration.

### TODO
- [ ] Check if helps in any actual SV-COMP cases.
- [ ] Can this prevent useful unassuming where disjunctive joining is beneficial? If so, then maybe this could be only applied at unrolled loop nodes?
- [ ] Evaluate sv-benchmarks?
- [ ] Enable in svcomp confs?


karoliineh (approved):
Out of the ~2k tasks in `C.no-overflow.Main` category, using a 30s for verification and a 60s for validation, the validation results on `master` and `unassume-precheck-disj` differed for 772 tasks:
* 494 `error`/`timeout`/`unknown` => `true`
* 13 `true` => `failed` due to some issue in benchexec
* 59 `timeout`/`error` => `failed` due to some issue in benchexec
* other diffs were just one non-solve reason to another, and there were no cases of `true` => not-solved other than the benchexec failures.

